PR #4, opened 2019-11-05
Fix bugs in database3 pack/unpackSpecialData

### LINKED ISSUE #2: Asymmetry in pack/unpackSpecialData in Database3
When packing special data in `Database3`, jagged arrays with `None`s apply the `"nones"` attribute, as well as the `noneLocations` attribute, which is needed for unpacking `None`s in a jagged array. However, when unpacking the same data, if the "nones" attribute is set, it just unpacks it as though it is a non-jagged array with `None`s, and immediately returns the result. This leaves all of the jagged data as a big, flat array of the wrong size. Need to check that the "jagged" attribute isn't set when unpacking `None`s, or not set the "nones" attribute when packing jagged arrays. Also need to get jagged arrays with `None`s under test.

youngmit:
This fixes #2

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -34,6 +34,7 @@ the database file, the ``Layout`` class is used to help map the hierarchical Com
 Reactor Model to the flat representation in the database.
 """
 import collections
+import copy
 import io
 import itertools
 import os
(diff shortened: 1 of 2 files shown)


armi/bookkeeping/db/database3.py
```diff
@@ -1611,8 +1612,16 @@ def packSpecialData(
 
     """
 
+    # Check to make sure that we even need to do this. If the numpy data type is
+    # not "O", chances are we have nice, clean data.
+    if data.dtype != "O":
+        return data, {}
```
ntouran: slightly out of scope but the return values for this method aren't super well defined in the docstring.

armi/bookkeeping/db/tests/test_database3.py
```diff
@@ -14,6 +14,7 @@
 
 import unittest
 import numpy
+import numpy.testing
 import h5py
 import os
```
ntouran: sidenote/pylint: out of order imports

PR #7, opened 2019-11-06
Travis cov

youngmit:


youngmit:
I looks like this branch was actually able to submit coverage results through Travis. Should be good to land.

Issue #15, opened 2019-11-16
tox fails when run locally on coveralls


ntouran:
As part of the testing process, I run `tox` locally after doing some changes. It runs tests three times:

* py36
* py37
* py37 again but this time with coverage

On the third item, it fails with the traceback below. I know it work on Travis but we should get it to also work locally. Also also, maybe just doing py37 with coverage instead of running it twice would be nice, but that's kind of a separate issue. 

Workaround for developers: just use `pytest -n6 armi` directly before submitting a PR.

```
Not on Travis or CircleCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var.
Traceback (most recent call last):
[...snipped...]
coveralls.exception.CoverallsException: Not on Travis or CircleCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var.
ERROR: InvocationError for command '.tox/report/bin/coveralls' (exited with code 1)
________________________________________________ summary ________________________________________________
  py36: commands succeeded
  py37: commands succeeded
  lint: commands succeeded
  cov: commands succeeded
ERROR:   report: commands failed
```

youngmit:
I can fix this. Issue is I shouldn't have put `report` in the `envlist`; should probably just call it explicitly from the travis.yaml

PR #17, opened 2019-11-19
Improve tox/travis config

youngmit:
- Remove redundant py3.7 tests; coverage does these
 - Remove report env from envlist in tox.ini


ntouran:
Fixes #15

PR #21, opened 2020-01-16
Replace object references with soft links

youngmit:
This replaces the use of object references to handle HDF5 attributes
that are too large to store naturally. This is because object references
break when copying an object that contains them to a new HDF5 file, as
is done in mergeHistory(). Instead, we use a simple string to refer to
the targed dataset like "@[path/to/dataset]".

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -77,7 +77,9 @@ from armi.reactor import geometry
 from armi.utils.textProcessors import resolveMarkupInclusions
 
 ORDER = interfaces.STACK_ORDER.BOOKKEEPING
-DB_VERSION = "3.1"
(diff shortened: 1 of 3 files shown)


armi/bookkeeping/db/database3.py
```diff
@@ -77,7 +77,9 @@
 from armi.utils.textProcessors import resolveMarkupInclusions
 
 ORDER = interfaces.STACK_ORDER.BOOKKEEPING
-DB_VERSION = "3.1"
+DB_VERSION = "3.2"
```
ntouran: (your idea): Time for a version log?

youngmit: Added some docstring talk about versions, and a little changelog

PR #27, opened 2020-02-04
Add default value for compiledExclusions

youngmit:


armi/bookkeeping/db/compareDB3.py
```diff
@@ -161,6 +161,7 @@ def compareDatabases(
     tolerance: float = 0.0,
 ) -> Optional[DiffResults]:
 
+    compiledExclusions = None
```
ntouran: would these be better from a typing standpoint as an empty list?

youngmit: I'd say six of one, half-dozen the other, but leaning towards `None`. Really, this is propagating/translating optional kwargs. While making `compiledExclusions` an empty list when `exclusions` is `None` would be correct (i.e., not fall into the classic mutable-default-argument-value Python trap, since you get a new `[]` each time), it would be sort of confusing from the callee's standpoint. To have an empty list passed to you because the caller's kwarg was default/`None`, when your default for a very similar kwarg is also `None` is a bit odd. You sort of lose the ability to propagate default-ness, if that makes sense.

PR #29, opened 2020-02-21
Fix to timesteps list changing in database interface

sammiller11235:
When database interface removes the `now` timestep, it would alter the list that was passed in from the caller's scope. Bad manners.


youngmit:
Thanks, @sammiller11235 good find

PR #31, opened 2020-02-27
Remove Duplicate Time Reporting on Case Initialization

jakehader:
An extra date/time message was being printed within the `writeWelcomeHeaders`
function and reporting two different times. The first time reflected the
time that the function was called, whereas the second time reflected the
start time for ARMI. The first time was removed.


jakehader:
Fixes #32

PR #34, opened 2020-03-06
Improve standardization in global flux calculations

ntouran:
Many global flux plugins had a similar run sequence.
This combines them into one to allow better inter-operability.

Also:
- Add better error handling to directory changers
- Add better testing for directory changers
- Remove some DIF3D-specific auxiliary code
- Defer error handling to the directory changer
- Improve debug statements in directorychangers
- Reduce parallel test collisions
- Remove complex cs dependency as much as possible.

### DIFF
--- a/armi/physics/executers.py
+++ b/armi/physics/executers.py
@@ -1,0 +1,177 @@
+"""
+Executors are useful for having a standard way to run physics calculations.
+
+They may involve external codes (with inputs/execution/output) or in-memory
+data pathways.
+"""
+
+import os
+
+from armi.utils import directoryChangers
+from armi import runLog
+
+
+class ExecutionOptions:
+    """
+    A data structure representing all options needed for a physics kernel.
+
+    Attributes
+    ----------
+    inputFile : str
+        Name of main input file. Often passed to stdin of external code.
+    outputFile : str
+        Name of main output file. Often the stdout of external code.
+    extraInputFiles : list of tuples
+        (sourceName, destName) pairs of file names that will be brought from the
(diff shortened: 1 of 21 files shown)


armi/physics/executers.py
```diff
@@ -0,0 +1,177 @@
...
+data pathways.
+"""
+
+import os
+
+from armi.utils import directoryChangers
+from armi import runLog
+
+
+class ExecutionOptions:
```
youngmit: Perhaps not now, but something to think about: I think Options stand a better chance of becoming part of various Interface APIs, and therefore may make sense to live in a module of their own.

armi/reactor/tests/test_zones.py
```diff
@@ -35,8 +35,8 @@
 class Zone_TestCase(unittest.TestCase):
     def setUp(self):
         bp = blueprints.Blueprints()
-        r = reactors.Reactor(settings.getMasterCs(), bp)
-        r.add(reactors.Core("Core", settings.getMasterCs()))
+        r = reactors.Reactor("zonetest", bp)
+        r.add(reactors.Core("Core"))
```
youngmit: Mmmm

armi/tests/__init__.py
```diff
@@ -42,8 +42,8 @@ def getEmptyHexReactor(cs=None):
 
     cs = cs or settings.getMasterCs()
```
youngmit: Perfect opportunity to get rig of a `getMasterCs()` call

armi/physics/executers.py
ntouran: not a bad idea

PR #35, opened 2020-03-12
Small improvements to new global flux code.

ntouran:
A few universal options weren't being read in at the
framework level, so this adds them. We also sorted
the options dump during debugging and decreased
surprising nuclide expansion behavior when running
without lattice physics.

### DIFF
--- a/armi/physics/executers.py
+++ b/armi/physics/executers.py
@@ -70,7 +70,7 @@ class ExecutionOptions:
     def describe(self):
         """Make a string summary of all options."""
         lines = ["Options summary:", "----------------"]
-        for key, val in self.__dict__.items():
+        for key, val in sorted(self.__dict__.items()):
             if not key.startswith("_"):
                 lines.append(f"  {key:40s}{str(val)[:80]:80s}")
         return "\n".join(lines)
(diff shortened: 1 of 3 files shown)


armi/physics/executers.py
```diff
@@ -70,7 +70,7 @@ def resolveDerivedOptions(self):
     def describe(self):
```
youngmit: Still wondering why this isn't just `__str__()`

armi/reactor/blueprints/isotopicOptions.py
```diff
@@ -524,7 +509,7 @@ def autoSelectElementsToKeepFromSettings(cs):
                 )
             )
 
-    elif cs["xsKernel"] in ["SERPENT", "MC2v3", "MC2v3-PARTISN"]:
+    elif cs["xsKernel"] in ["", "SERPENT", "MC2v3", "MC2v3-PARTISN"]:
```
youngmit: Yikes. Maybe this should be a plugin thing? No need to do it now though

ntouran: Absolutely. There are commends above in the docstring from earlier stating that this needs to be migrated into the plugins. The behavior is surprising and complicated here.

armi/physics/executers.py
ntouran: It's just so big! I'm always slightly surprised when objects str() is a gargantuan printout. But I guess that may be a me problem.

PR #37, opened 2020-03-17
Improve coverage tracking behavior

ntouran:
If you have just one coverage results file
in a folder and you call cov.combine(), the one
file gets deleted. This prevents this from happening
in most cases used in ARMI testing.

### DIFF
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -52,18 +52,25 @@ _HASH_BUFFER_SIZE = (
 
 def coverageReportHelper(config, dataPaths):
     """
-    Small utility function to generate converage reports.
+    Small utility function to generate coverage reports.
 
     This was created to side-step the difficulties in submitting multi-line python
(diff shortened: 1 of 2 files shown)


armi/utils/__init__.py
```diff
@@ -52,18 +52,25 @@
...
     """
     from coverage import Coverage
     import coverage
 
     try:
         cov = Coverage(config_file=config)
-        cov.combine(data_paths=dataPaths)
-        cov.save()
+        if dataPaths:
+            # fun fact: if you combine when there's only one file, it gets deleted.
```
youngmit: Is this worth submitting an issue upstream?

ntouran: It's at least related to this one: https://github.com/nedbat/coveragepy/issues/516

I guess their solution was to add the `strict` option... ok. Our solution here still makes sense for our usage.

PR #39, opened 2020-03-18
Create a generic fuel performance plugin

sammiller11235:
Similar to the other generic plugins, this will be a central place
to put settings, parameters, executers, etc. that are expected to
be shared across a variety of fuel performance plugins.


armi/physics/fuelPerformance/tests/test_fuelPerformancePlugin.py
```diff
@@ -0,0 +1,15 @@
+"""
```
youngmit: license header

PR #42, opened 2020-03-21
Minor reorg and fixes to docs

ntouran:
* This makes a new section describing the detail assembly settings.
  It was moved with an anchor so plugins that use it can easily
  reference it.

* Bug in RTD theme related to table widths was fixed. Now the
  table of all settings renders in a readable way.

* Empty units column of settings table removed

### DIFF
--- a/doc/developer/documenting.rst
+++ b/doc/developer/documenting.rst
@@ -38,14 +38,15 @@ all output cells in the notebook web interface itself before committing the file
 
 Documentation for ARMI plugins
 ------------------------------
+The following subsections apply to documentation for ARMI plugins.
 
 Linking to ARMI documentation from plugins
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 ARMI plugin documentation can feature rich hyperlinks to the ARMI API documentation 
-with the help of the 
-`intersphinx plugin <http://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html>`_.
-The plugin documentation config file simply should add ``"sphinx.ext.intersphinx",`` to
-its active plugin list, and change the default config to read::
+with the help of the
+`intersphinx Sphinx plugin <http://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html>`_.
+The ARMI plugin documentation config file should add ``"sphinx.ext.intersphinx",`` to
+its active Sphinx plugin list, and change the default config to read::
 
     intersphinx_mapping = {
         "python": ("https://docs.python.org/3", None),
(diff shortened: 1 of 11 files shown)


doc/developer/documenting.rst
```diff
@@ -34,4 +34,34 @@ documentation is built. To do this, you can clean the output with::
 	jupyter nbconvert --ClearOutputPreprocessor.enabled=True --inplace mynotebook.ipynb
 
 This should clear the output and overwrite the file. If this doesn't work, you can clear
-all output cells in the notebook web interface itself before committing the file.
\ No newline at end of file
+all output cells in the notebook web interface itself before committing the file.
+
+Documentation for ARMI plugins
+------------------------------
```
youngmit: Did you have something in mind to go here?

doc/developer/documenting.rst
```diff
@@ -34,4 +34,34 @@ documentation is built. To do this, you can clean the output with::
...
+
+Linking to ARMI documentation from plugins
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+ARMI plugin documentation can feature rich hyperlinks to the ARMI API documentation 
+with the help of the 
+`intersphinx plugin <http://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html>`_.
+The plugin documentation config file simply should add ``"sphinx.ext.intersphinx",`` to
```
youngmit: Consider removing "simply" (A little opinionated, but I do agree with the sentiment: https://www.knowledgeowl.com/home/dont-say-simply-jim-fisher)

doc/developer/documenting.rst
```diff
@@ -34,4 +34,34 @@ documentation is built. To do this, you can clean the output with::
...
+
+    :doc:`armi:developer/documenting`
+    :py:mod:`armi.physics.executers`
+
+
+Automatically building apidocs of namespace packages
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Activating the ``"sphinxcontrib.apidoc",`` `plugin <https://github.com/sphinx-contrib/apidoc>`_
+enables plugin API documentation to be built with the standard ``make html`` sphinx workflow. If 
+your plugin is a namespace package, the following extra config is required::
```
youngmit: Might be useful to provide distinction between Sphinx plugins and ARMI plugins here.

doc/user/inputs/settings.rst
```diff
@@ -61,3 +61,34 @@ to update this setting.
...
+
+Detail Assembly Locations BOL
+    The ``detailAssemLocationsBOL`` setting is a list of assembly location strings 
+    (e.g. ``A4003`` for ring 4, position3). Assemblies that are in these locations at the 
+    beginning-of-life will be activated as detail assemblies.
+
+Detail assembly numbers
+    The ``detailAssemNums`` setting is a list of``assemNum``s that can be inferred from a previous
+    case and specified, regardless of when the assemblies enter the core. This is useful for
+    activating detailed treatment of assemblies that enter the core at a later cycle.
```
youngmit: Does the double-backtick followed by "s" render correctly? At least with the GH hilighting, it seems off. I also seem to remember having trouble with pluralizing things in the past.

youngmit: I think "detail assemblies" is getting perhaps a little overgrown. It can be surprising when certain extra analyses are performed that aren't expected, and there is no way of controlling what is done and what isn't, given a collection of enabled plugins, without extra plugin-specific settings. Consider adding a note that we plan on re-thinking it at some point.

doc/developer/documenting.rst
ntouran: It's just a section for subsections. Will add a line

doc/developer/documenting.rst
ntouran: Done. thanks. haha

doc/developer/documenting.rst
ntouran: Great point. done

doc/user/inputs/settings.rst
ntouran: adding note. Also, you have to use `\ s` to pluralize things like that apparently. good catch!

doc/developer/documenting.rst
ntouran: awww man you can't even say "boom, you're done"!?

PR #44, opened 2020-03-24
Remove one level of the call stack for ndens.

ntouran:
Getting number densities is a performance-critical activity. This small
change just accesses the number density params directly rather than
using the method call just above. This won't change things by much but
just acknowledges that performance really matters in these methods.

### DIFF
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -594,7 +594,7 @@ class Component(composites.Composite, metaclass=ComponentType):
 
     def getNuclideNumberDensities(self, nucNames):
         """Return a list of number densities for the nuc names requested."""
-        return [self.getNumberDensity(nucName) for nucName in nucNames]
+        return [self.p.numberDensities.get(nucName, 0.0) for nucName in nucNames]
 
     def _getNdensHelper(self):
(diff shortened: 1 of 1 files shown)


armi/reactor/components/component.py
```diff
@@ -594,7 +594,7 @@ def getNumberDensity(self, nucName):
 
     def getNuclideNumberDensities(self, nucNames):
         """Return a list of number densities for the nuc names requested."""
-        return [self.getNumberDensity(nucName) for nucName in nucNames]
+        return [self.p.numberDensities.get(nucName, 0.0) for nucName in nucNames]
```
youngmit: If calling `getNumberDensity` is discouraged, and `getNumberDensity` does the same as this line, should we just remove it?

ntouran: `getNumberDensity` here is a specific implementation of the `composite` pattern's broader one so it has to stay so that a component looks like a block/assembly/core/reactor/whatever via the api

ntouran:
> 
> 
> Did you do any measurement on this? Always fun to see how things affect performance in practice

I did not measure it, it was just a flyby change. I would be nice to get some legit profiling going as part of the CI tests.

PR #48, opened 2020-03-31
Updates to Hastelloy N and Incoloy 800 materials

sammiller11235:
* Changes to composition based on open literature
* Adding/Updating references
* Implementing thermal expansion based on open source references
* Changing Hastelloy N thermal conductivity and heat capacity to match fits to open literature data

### DIFF
--- a/armi/materials/hastelloyN.py
+++ b/armi/materials/hastelloyN.py
@@ -16,11 +16,22 @@
 Hastelloy-N is a high-nickel structural material invented by ORNL for handling molten fluoride salts.
 """
 
-from armi.utils.units import getTk
+from armi.utils.units import getTk, getTc
 from armi.materials.material import Material
 
 
 class HastelloyN(Material):
+    r"""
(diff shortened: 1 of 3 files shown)


sammiller11235:
The Hastelloy N material gives negative values for heat capacity above the temperature range which isn't physical. There's a warning currently, but maybe it should raise an error if requested. The `material.checkTempRange` method has a hard-coded "FAIL_ON_RANGE" variable which is set to false, but maybe there could be an added argument to the `checkTempRange` method to raise an error if the property values are unphysical when extrapolated. Just a thought.

sammiller11235:
I was following precedent to check a range of values.

ntouran:
Good idea on having better control over when to fail on range. 

Yeah we have tons of precedent of checks like that and that's why I worry about it. They have historically been a pain to update when material properties change. I should probably just put comment/warnings in all the tests that do it so future developers know the story. Sorry about that.

PR #49, opened 2020-04-01
Re-implement the 1/3 hex asciimap reader

youngmit:
The old one was not interpreting the leading hyphens properly. The new
ones is a more reliable inversion of the writing process, and handles
spacing between elements better.

### DIFF
--- a/armi/utils/asciimaps.py
+++ b/armi/utils/asciimaps.py
@@ -33,6 +33,9 @@ armi.reactor.geometry : a specific usage of lattices, for core maps
 import itertools
 
 
+PLACEHOLDER = "-"
+
+
 class AsciiMap(object):
     """
     Base class for maps.
@@ -71,7 +74,7 @@ class AsciiMap(object):
                 i, j = self._getIndices(base, col)
                 self.lattice[i, j] = char
 
-        self.lattice = {k: v for k, v in self.lattice.items() if v != "-"}
+        self.lattice = {k: v for k, v in self.lattice.items() if v != PLACEHOLDER}
         return self.lattice
 
     def writeMap(self, stream):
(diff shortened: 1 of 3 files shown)


armi/utils/asciimaps.py
```diff
@@ -69,6 +70,8 @@ def readMap(self, text):
             for col, char in enumerate(line.split()):
                 i, j = self._getIndices(base, col)
                 self.lattice[i, j] = char
+
+        self.lattice = {k: v for k, v in self.lattice.items() if v != "-"}
```
ntouran, line 74: we should probably constantize this with 

`PLACEHOLDER = "-"` and then refer to `PLACEHOLDER` here.

armi/utils/asciimaps.py
```diff
@@ -180,22 +183,44 @@ def _getIndices(base, col):  # pylint: disable=no-self-use
         return iBase + col * 2, jBase - col
 
     def writeMap(self, stream):
-        """Writing this is easiest if we just make a big ascii map and fill it in."""
+        """Write the lattice to a text stream"""
         grid = {}
+        # The length of the largest string contents. Useful for spacing everything out
+        # nicely
+        maxSize = len(str(next(iter(self.lattice.values()))))
```
ntouran, line 190: groan is that really what we have to do to get the length of the first value!? Python 3 whyyyy

armi/utils/asciimaps.py
```diff
@@ -180,22 +183,44 @@ def _getIndices(base, col):  # pylint: disable=no-self-use
...
+            for col in itertools.count(0):
+                i, j = self._getIndices(base, col)
+
+                contents = self.lattice.get((i, j), None)
+                if contents is None and foundContents:
+                    # we have gotten to the end of the contents for this row; break out
+                    # of our otherwise infinite loop
+                    break
+                if contents is None:
+                    contents = '-'
```
ntouran, line 217: `PLACEHOLDER`

armi/utils/asciimaps.py
youngmit: Yeah, I had a private little hissy of my own when I found out that the various dict view classes arent _themselves_ iterators. It feels like the `iter()` call should be unnecessary.

PR #50, opened 2020-04-02
Small performance enhancement with str-to-float.

ntouran:


armi/utils/__init__.py
```diff
@@ -149,9 +151,10 @@ def fixThreeDigitExp(strToFloat):
     just fine, and they are valid floating point numbers.  It would not be a useful effort,
     in terms of time, trying to get FORTRAN to behave differently.
     The approach has been to write a routine in the reading code which will interpret these.
+
+    This helps when the scientific number exponent does not fit.
     """
-    scipat = r"([+-]?\d*\.\d+)[eEdD]?([+-]\d+)"
-    match = re.match(scipat, strToFloat)
+    match = re.match(SCIPAT_SPECIAL, strToFloat)
```
youngmit, line 157: Did you measure the results of this change? I'm used to using pre-compiled regex something like this:
```
_RE_SCIPAT_SPECIAL = re.compile(...)
...
_RE_SCIPAT_SPECIAL.match(candidate)
```

Looking at the docs for the re module, I'm not really sure what happens when you pass a compiled pattern to the module-level `match()` function. I would certainly hope that it doesn't recompile...

Either way, could you change this to `SCIPAT_SPECIAL.match()`. At the very least it is more obvious that `SCIPAT_SPECIAL` is a precompiled expression, rather than a string containing one.

ntouran:
No I did not measure it. I was working on a different bug and just did this as an auto-cleanup. 

I am nearly 100% sure re.match does not recompile when passed a compiled pattern. I got big performance improvements with compiling REs way back in the early REBUS-reading days by doing it like this before I knew `compiledPattern.match(line)` was even possible.

armi/utils/__init__.py
ntouran: No I did not measure it. I was working on a different bug and just did this as an auto-cleanup. 

I am nearly 100% sure re.match does not recompile when passed a compiled pattern. I got big performance improvements with compiling REs way back in the early REBUS-reading days by doing it like this before I knew `compiledPattern.match(line)` was even possible.

PR #51, opened 2020-04-02
Uniform mesh conversion fixes

ntouran:
Fixes issues where multiple uniform mesh conversions happen in the same timestep (e.g. gamma-transport corrected global flux calculations with detailed axial expansion and critical control rod insertion).

### DIFF
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -16,8 +16,12 @@
 Converts reactor with arbitrary axial meshing (e.g. multiple assemblies with different
 axial meshes) to one with a global uniform axial mesh.
 
-Useful for preparing DIF3D, etc. structured mesh neutronics runs from flexible ARMI
-reactor
+Useful for preparing inputs for physics codes that require structured meshes
+from a more flexible ARMI reactor mesh.
+
+This is implemented generically but includes a concrete subclass for
+neutronics-specific parameters. This is used for build input files
+for codes like DIF3D which require axially uniform meshes.
 
 Requirements
 ------------
@@ -29,8 +33,7 @@ Requirements
 
 Examples
 --------
-
-converter = uniformMesh.UniformMeshGeometryConverter()
+converter = uniformMesh.NeutronicsUniformMeshConverter()
 uniformReactor = converter.convert(reactor)
 # do calcs, then:
 converter.applyStateToOriginal()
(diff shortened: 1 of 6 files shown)


ntouran:
Ah dang the commit from the other PR was not supposed to be in here. Will remove.

armi/reactor/converters/uniformMesh.py
```diff
@@ -61,21 +61,54 @@ class UniformMeshGeometryConverter(GeometryConverter):
...
+        """Create a new reactor with a uniform mesh."""
         runLog.extra("Building copy of {} with a uniform axial mesh".format(r))
         self._sourceReactor = r
         self.convReactor = self.initNewReactor(r)
+        self._setParamsToUpdate()
         self._computeAverageAxialMesh()
-        self._convertNumberDensities()
+        self._buildAllUniformAssemblies()
+        self._clearStateOnReactor(self.convReactor)
+        self._mapStateFromReactorToOther(self._sourceReactor, self.convReactor)
```
ntouran, line 76: FYI: this is the key to the bugfix

armi/reactor/converters/uniformMesh.py
```diff
@@ -188,11 +232,14 @@ def plotConvertedReactor(self):
...
+        """Gather a list of parameters that will be mapped between reactors."""
         b = self._sourceReactor.core.getFirstBlock()
 
-        self.blockParamNames = b.p.paramDefs.inCategory("detailedAxialExpansion").names
+        self.blockParamNames = b.p.paramDefs.inCategory(
+            parameters.Category.detailedAxialExpansion
+        ).names
         self.reactorParamNames = self._sourceReactor.core.p.paramDefs.inCategory(
-            "neutronics"
+            parameters.Category.neutronics
```
youngmit, line 242: It seems odd to use a category dedicated to this purpose in the block params, but to reach for the oddly-specific "neutronics" category for the reactor params. This is nothing new, so shouldn't hold up the show, but maybe we should consider adding the neutronics reactor params to the detailedAxialExpansion category?

armi/reactor/converters/uniformMesh.py
```diff
@@ -105,7 +138,20 @@ def _createNewAssembly(sourceAssembly):
...
-    def applyUniformMesh(sourceAssem, newMesh):
+    def makeAssemWithUniformMesh(sourceAssem, newMesh):
+        """
+        Build new assembly based on a source assembly but apply the uniform mesh.
+
+        The new assemblies must have appropriately mapped number densities as
+        input for a neutronics solve. They must also have other relevant
+        neutronics parameters for follow-on neutronics steps (e.g. a fixed
+        source gamma transport step that needs appropriate fission rates).
+        Thus, this maps many parameters from the ARMI mesh to the uniform mesh.
```
youngmit, line 149: As below this seems pretty neurtronics-y in a context where it shouldnt have to be. If this converter really is intended for neutronics-type stuff, perhaps we should move into that package.

armi/reactor/converters/uniformMesh.py
ntouran: good point. I'm thinking we should make the a `UniformMeshConverterForNeutronics` subclass in the neutronics package to really separate it. The conversion itself is generic, but the list of params that get translated could be different for different clients/subclasses.

ntouran:
K I made a subclass at least to acknowledge the fact that that one was really neutronics specific.

armi/reactor/converters/uniformMesh.py
ntouran: moved to a subclass at least for now. if this were  a plugin I'd namespace package it into `plugins.reactor.converters` but it'd feel a little weird to buy this down in physics.neutronics to me at the moment. Maybe you're right though

PR #52, opened 2020-04-06
Improve handling of settings and blueprints files

youngmit:
This makes a handful of changes that improve how settings and blueprints files
are handled:
 - The mechanism for including external YAML files no longer uses the YAML
   parser itself. This was to support things like anchors and aliases across
   files without having to resort to too much dark magic. Instead, the the
   mechanism underpinning !include is as close to a textual include as
   reasonably possible. The only "fancy" thing that is done when including files
   is maintaining consistent indentation, relative to the !include directive.
 - Update calls to `resolveMarkupInclusions()` to point to the appropriate root
   directory.
 - Implement proper cloning of included YAML files in the case/case suite
   system.
 - Update numerous uses of DirectoryChanger to not dumpOnException; most of
   these are in the case/case suite system, which really shouldnt making dump
   directories when things fail.


youngmit:
Been testing this out on more complicated case suite scenarios, and it still needs some work. Mostly related to how blueprints snippets get copied around as part of a Suite, if the paths aren't relative to the settings file that uses them. I think I just got it sorted out though, will update PR soon if things look like they're working.

PR #58, opened 2020-04-16
Remove deprecated settings implementation

youngmit:
This removes the old settings module and renames setting2.py to
setting.py. This is a breaking change for plugins that were doing
something like `from armi.settings import setting2 as setting`, but
trivial to refactor around.


armi/settings/setting.py
```diff
@@ -98,431 +175,132 @@ def value(self):
...
         """
-        attrib = Setting.getDefaultAttributes(self)
-        attrib["containedType"] = self.containedType
-        attrib["options"] = self.options
-        attrib["enforcedOptions"] = self.enforcedOptions
-        return attrib
-
-
-# help direct python types to the respective setting type
-SUBSETTING_MAP = {
```
ntouran, line 520: ugh this diff renders weirdly rather than showing a rename.

youngmit:
Yeah, should have looked at that coverage stuff earlier. I'll comb through the coveralls stuff and see what I can do

armi/settings/tests/test_settingsIO.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2019 TerraPower, LLC
```
ntouran, line 1: :open_mouth:

youngmit: Eyes like hawk

youngmit:
I think 0.08% is pretty good

PR #59, opened 2020-04-23
Allow square components to be used for Cartesian lattice

onufer:


armi/reactor/blocks.py
```diff
@@ -2909,8 +2911,11 @@ def getPitch(self, returnComp=False):
         # block as opposed to initializing _pitchDefiningComponent to (None, None)
         if c is None:
             raise ValueError("{} has no valid pitch".format(self))
+
+        # ask component for dimensions, since they could have changed
+        elif isinstance(c, components.Square):
+            maxLength = maxWidth = c.getDimension("widthOuter")
```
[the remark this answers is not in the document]
ntouran: This kind of class-based specialization hints to me that we should refactor this need down into the components more and do the specialization via subclasses.

armi/reactor/tests/test_blocks.py
```diff
@@ -1118,20 +1118,6 @@ def test102_setPitch(self):
             self.Block.getComponent(Flags.INTERCOOLANT).getDimension("op"), pitch
         )
 
-    def test_UnshapedGetPitch(self):
-        """
-        Test that a homogenous block can be created with a specific pitch.
-        This functionality is necessary for making simple homogenous reactors.
```
onufer, line 1124: This seemed pretty hacky/unneeded. You can always make it a hex shape or whatever shape you expect your homog reactor to be made of. Fundementally UnshapedComponent having an outer pitch is really smelling, and if none of our testing needs it I think we should not support that feature.

ntouran: This seems to have been necessary for when people wanted to perform basic benchmark problems that only defined fundamental volume fractions. In these scenarios, if you have 25% coolant, 25% steel, and 50% fuel, you don't usually want to have to back-calculate some fake geometry that would have those fractions. This particular test is holding up that capability. 

If we want to remove it entirely we should also remove op from the unshaped component constructor. I never really was a fan of that but do like the feature.

armi/reactor/tests/test_blocks.py
```diff
@@ -1522,13 +1509,196 @@ def test_getPinCoords(self):
...
+
+        # We haven't set a pitch defining component this time so set it now with 0 area.
+        pitchDefiningComponent = components.Rectangle(
+            "pitchComp",
+            "Void",
+            lengthOuter=desiredPitch[0],
+            lengthInner=desiredPitch[0],
+            widthOuter=desiredPitch[1],
+            widthInner=desiredPitch[1],
+            mult=1,
```
ntouran, line 1695: I recommend just leaving this 2nd way test out to keep the testmaintenance burden low. You demonstrated this approach fine in the hex test.

ntouran:
Oh also there should be a slightly better commit message with this.

armi/tests/refSmallReactor.yaml
```diff
@@ -358,14 +358,6 @@ blocks:
         coolant: *component_fuel_coolant
         duct: *component_fuel_duct
         intercoolant: *component_fuel_intercoolant
-    homog fuel:
```
onufer, line 361: This was unused, and no longer supported.

doc/user/inputs/blueprints.rst
```diff
@@ -597,11 +597,85 @@ Here is a complete fuel block definition::
...
+                    shape: Rectangle
+                    material: Void
+                    lengthOuter: 3.0
+                    lengthInner: 3.0
+                    widthOuter: 2.0
+                    widthInner: 2.0
+                    mult:1.0
+
+
+This can similarly be done for hex geometry and and a hexagon with Outer Pitch (`op`).
```
ntouran, line 678: should have double backticks

PR #61, opened 2020-04-24
Boost global flux interface testing coverage

ntouran:
The generic standalone global flux interface didn't have any
tests. This code has historically been covered by plugin tests
that are not in the framework repo. As part of testing it, a
few minor changes were made and adjustments to methodology.

* Basic unit tests that run the majority of the globalFluxInterface.py
  were added. Mostly they just make sure it runs.

* Some dpa-related settings are used in the framework that were
  not defined in the framework. Though they are a bit design-specific, I
  decided to bring the settings into the framework for now. The framework
  code otherwise simply cannot run.

* The neutronics restart setting is now in the framework

* The power normalization was assuming that powers/flux ratios
  were reasonable. In the dummy tests I made this was certainly
  not the case. I decided that this method should snap the power
  to something consistent with the flux because it has all the
  information needed anyway.

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -317,8 +317,6 @@ class GlobalFluxOptions(executers.ExecutionOptions):
         self.runDir = os.path.join(
             armi.FAST_PATH, f"{cs.caseTitle}-{self.label}-{armi.context.MPI_RANK}"
         )
-        if cs["restartNeutronics"]:
-            self.isRestart = True
         self.isRestart = cs["restartNeutronics"]
         self.adjoint = neutronics.adjointCalculationRequested(cs)
         self.real = neutronics.realCalculationRequested(cs)
@@ -506,11 +504,11 @@ class GlobalFluxResultMapper(interfaces.OutputReader):
         --------
         getTotalEnergyGenerationConstants
         """
-        # The multi-group flux is volume integrated, so J/cm * n-cm/s gives units of Watts
         # update the block power param here as well so
         # the ratio/multiplications below are consistent
         currentCorePower = 0.0
         for b in self.r.core.getBlocks():
+            # The multi-group flux is volume integrated, so J/cm * n-cm/s gives units of Watts
             b.p.power = numpy.dot(
                 b.getTotalEnergyGenerationConstants(), b.getIntegratedMgFlux()
             )
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -547,6 +547,7 @@ class GlobalFluxResultMapper(interfaces.OutputReader):
                 b.p.arealPd = b.p.power / area * conversion
             a.p.arealPd = a.calcTotalParam("arealPd")
         self.r.core.p.maxPD = self.r.core.getMaxParam("arealPd")
+        self._updateAssemblyLevelParams()
 
     def updateDpaRate(self, blockList=None):
         """
@@ -596,15 +597,11 @@ class GlobalFluxResultMapper(interfaces.OutputReader):
         maxGridDose = self.r.core.getMaxBlockParam("detailedDpaPeak", Flags.GRID_PLATE)
         self.r.core.p.maxGridDpa = maxGridDose
 
-    def updateAssemblyLevelParams(self, excludedBlockTypes=None):
+    def _updateAssemblyLevelParams(self):
         for a in self.r.core.getAssemblies():
-            if excludedBlockTypes is None:
-                excludedBlockTypes = []
             totalAbs = 0.0  # for calculating assembly average k-inf
             totalSrc = 0.0
             for b in a:
-                if b.getType() in excludedBlockTypes:
-                    continue
                 totalAbs += b.p.rateAbs
                 totalSrc += b.p.rateProdNet
 
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -16,6 +16,7 @@
 The Global flux interface provide a base class for all neutronics tools that compute the neutron and/or photon flux.
 """
 import math
+import os
 
 
 import numpy
@@ -46,7 +47,7 @@ class GlobalFluxInterface(interfaces.Interface):
     Should be subclassed by more specific implementations.
     """
 
-    name = None  # make sure to set this in subclasses
+    name = "GlobalFlux"  # make sure to set this in subclasses
     function = "globalFlux"
     _ENERGY_BALANCE_REL_TOL = 1e-5
 
@@ -313,6 +314,11 @@ class GlobalFluxOptions(executers.ExecutionOptions):
         This is not required; these options can alternatively be set programmatically.
         """
         self.kernelName = cs["neutronicsKernel"]
+        self.runDir = os.path.join(
+            armi.FAST_PATH, f"{cs.caseTitle}-{self.label}-{armi.context.MPI_RANK}"
+        )
+        if cs["restartNeutronics"]:
+            self.isRestart = True
         self.isRestart = cs["restartNeutronics"]
         self.adjoint = neutronics.adjointCalculationRequested(cs)
         self.real = neutronics.realCalculationRequested(cs)
@@ -489,7 +495,7 @@ class GlobalFluxResultMapper(interfaces.OutputReader):
     def _renormalizeNeutronFluxByBlock(self, renormalizationCorePower):
         """
         Normalize the neutron flux within each block to meet the renormalization power.
-        
+
         Parameters
         ----------
         renormalizationCorePower: float
@@ -501,14 +507,16 @@ class GlobalFluxResultMapper(interfaces.OutputReader):
         getTotalEnergyGenerationConstants
         """
         # The multi-group flux is volume integrated, so J/cm * n-cm/s gives units of Watts
-        currentCorePower = sum(
-            [
-                numpy.dot(
-                    b.getTotalEnergyGenerationConstants(), b.getIntegratedMgFlux()
-                )
-                for b in self.r.core.getBlocks()
-            ]
-        )
+        # update the block power param here as well so
+        # the ratio/multiplications below are consistent
+        currentCorePower = 0.0
+        for b in self.r.core.getBlocks():
+            b.p.power = numpy.dot(
+                b.getTotalEnergyGenerationConstants(), b.getIntegratedMgFlux()
+            )
+            b.p.flux = sum(b.getMgFlux())
+            currentCorePower += b.p.power
+
         powerRatio = renormalizationCorePower / currentCorePower
         runLog.info(
             "Renormalizing the neutron flux in {:<s} by a factor of {:<8.5e}, "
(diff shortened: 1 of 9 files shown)


armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -46,7 +47,7 @@ class GlobalFluxInterface(interfaces.Interface):
     Should be subclassed by more specific implementations.
     """
 
-    name = None  # make sure to set this in subclasses
+    name = "GlobalFlux"  # make sure to set this in subclasses
```
onufer, line 50: Is the comment `make sure to set this in subclasses` still relevent?

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -313,6 +314,11 @@ def fromUserSettings(self, cs):
         This is not required; these options can alternatively be set programmatically.
         """
         self.kernelName = cs["neutronicsKernel"]
+        self.runDir = os.path.join(
+            armi.FAST_PATH, f"{cs.caseTitle}-{self.label}-{armi.context.MPI_RANK}"
+        )
+        if cs["restartNeutronics"]:
+            self.isRestart = True
```
onufer, line 321: What doe these lines doe that the line below does not?

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -501,14 +507,16 @@ def _renormalizeNeutronFluxByBlock(self, renormalizationCorePower):
         getTotalEnergyGenerationConstants
         """
         # The multi-group flux is volume integrated, so J/cm * n-cm/s gives units of Watts
```
onufer, line 509: This comment might belong better above the numpy dot in the for loop.

armi/physics/neutronics/settings.py
```diff
@@ -28,14 +28,28 @@
...
 CONF_EPS_EIG = "epsEig"
 CONF_EPS_FSAVG = "epsFSAvg"
 CONF_EPS_FSPOINT = "epsFSPoint"
 
-# used for dpa/dose analysis
+# used for dpa/dose analysis. These should be relocated to more
+# design-specific places
 CONF_LOAD_PAD_ELEVATION = "loadPadElevation"
 CONF_LOAD_PAD_LENGTH = "loadPadLength"
+CONF_ACLP_DOSE_LIMIT = "aclpDoseLimit"
```
onufer, line 41: consider defining acronym

armi/utils/directoryChangers.py
```diff
@@ -220,6 +220,8 @@ def __init__(
         dumpOnException=True,
         clean=False,
     ):
+        if not destination:
+            raise ValueError("Non-empty destination directory must be provided.")
```
onufer, line 224: This doesn't appear to check if its empty or not. Consider checking if its empty or changing wording to `valid destination directory must be provided`

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -0,0 +1,150 @@
...
+Tests for generic global flux interface.
+"""
+import unittest
+
+import numpy
+
+from armi import settings
+from armi.reactor import reactors
+
+from armi.physics.neutronics.globalFlux import globalFluxInterface as gi
```
onufer, line 11: Consider capitalizing, and or renaming GF or GFI. global interface is not super clear to me

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -0,0 +1,150 @@
...
+        gfi.interactEveryNode(0, 0)
+        r.p.timeNode += 1
+        gfi.interactEveryNode(0, 1)
+        gfi.interactEOC()
+        self.assertAlmostEqual(r.core.p.rxSwing, (1.02 - 1.01) / 1.01 * 1e5)
+
+
+class TestGlobalFluxResultMapper(unittest.TestCase):
+    def test_mapper(self):
+        MCC2_SETTINGS = {"xsKernel": "MC2v2"}  # compat with ISOAA test lib
```
onufer, line 118: compat  --> compatible

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -0,0 +1,150 @@
...
+from armi.nuclearDataIO import isotxs
+
+
+class MockParams:
+    pass
+
+
+class MockCore:
+    def __init__(self):
+        self.geomType = "spiral"
```
onufer, line 23: Either change or add comment explaining that this is not currently a supported geom type.
I understand we want to keep things light, but clarity is also important.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -0,0 +1,150 @@
...
+    def interactBOC(self, cycle=None):
+        gi.GlobalFluxInterface.interactBOC(self, cycle=cycle)
+        self.r.core.p.keff = 1.00
+
+    def interactEveryNode(self, cycle, node):
+        gi.GlobalFluxInterface.interactEveryNode(self, cycle, node)
+        self.r.core.p.keff = 1.01
+
+
+class MockGlobalFluxWithExecuters(gi.GlobalFluxInterfaceUsingExecuters):
```
onufer, line 50: Is this a Mock? It uses a mock for an exectutor, but I am not sure it itself is a mock.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -0,0 +1,150 @@
...
+    def getExecuterCls(self):
+        return MockGlobalFluxExecuter
+
+
+class MockGlobalFluxExecuter(gi.GlobalFluxExecuter):
+    def _readOutput(self):
+        class MockOutputReader:
```
onufer, line 57: It seems like there are 2 ways to change self.r.core.p.keff , one here, and one in interact every node. Can you add a comment explaining how they relate.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -0,0 +1,150 @@
...
+
+        Check that a 1000 pcm rx swing is observed due to the mock.
+        """
+        cs = settings.Settings()
+        o, r = test_reactors.loadTestReactor()
+        gfi = MockGlobalFluxInterface(r, cs)
+        gfi.interactBOC()
+        gfi.interactEveryNode(0, 0)
+        gfi.interactEOC()
+        self.assertAlmostEqual(r.core.p.rxSwing, 1000)
```
onufer, line 91: Consider also testing the initial or final keff.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -0,0 +1,150 @@
...
+        self.assertAlmostEqual(r.core.calcTotalParam("power", generationNum=2), 100)
+
+        mapper._updateDerivedParams()
+        self.assertGreater(r.core.p.maxPD, 0.0)
+        self.assertGreater(r.core.p.maxFlux, 0.0)
+
+        mapper.updateDpaRate()
+        self.assertGreater(r.core.getFirstBlock().p.detailedDpaRate, 0)
+
+        self.assertEqual(r.core.getFirstBlock().p.detailedDpa, 0)
```
onufer, line 134: consider adding comment saying it is expected this is zero, since zero time has passed.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -0,0 +1,150 @@
...
+        _o, r = test_reactors.loadTestReactor(customSettings=MCC2_SETTINGS)
+        applyDummyFlux(r)
+        r.core.lib = isotxs.readBinary(ISOAA_PATH)
+        mapper = gi.GlobalFluxResultMapper()
+        mapper.r = r
+        mapper._renormalizeNeutronFluxByBlock(100)
+        self.assertAlmostEqual(r.core.calcTotalParam("power", generationNum=2), 100)
+
+        mapper._updateDerivedParams()
+        self.assertGreater(r.core.p.maxPD, 0.0)
```
onufer, line 128: You should be able to calculate the expected values for all this stuff given the knowledge of the number of blocks in the core (using len(core.getBlocks()) so there is no maintenance burden as the test reactor changes). Testing assertGreater doesn't really prove much, just that it is set to a float > 0. 

For example, I could change _renormalizeNeutronFluxByBlock to set everything to 1 or divide by powerRatio and all these test would likely still pass.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
youngmit: Also, this should probably be `function`, rather than name

ntouran: * Yes it has to be set in subclasses to get the proper name in file labels and printouts
* No it shouldn't be function because it needs to change for different particular implementations of the `globalFlux` function

armi/physics/neutronics/globalFlux/globalFluxInterface.py
ntouran: lol not much!

armi/utils/directoryChangers.py
ntouran: I meant "empty" as in `""`, not that the folder itself is empty. Will clarify.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
ntouran: I'll just remove the rename completely.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
ntouran: Added docstring along those lines.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
ntouran: This is just a test that the mapping process operates, not that it is right. These tests do not intend to test physics of any kind. Areal pd here is a summation over blocks of a param set by physics modules that do not exist in the framework. The summation of params is tested elsewhere.

armi/physics/neutronics/settings.py
```diff
@@ -170,6 +184,32 @@ def defineSettings():
...
+            description="The cross sections to use for grid plate blocks DPA when computing displacements per atom.",
+            options=CONF_OPT_DPA,
+        ),
+        setting.Setting(
+            CONF_DPA_XS_SET,
+            default="dpa_EBRII_HT9",
+            label="DPA Cross Sections",
+            description="The cross sections to use when computing displacements per atom.",
+            options=CONF_OPT_DPA,
+        ),
```
youngmit: These seem like pretty un-framework-y settings to me...

ntouran: Agreed for sure, but they are required for the framework to run at the moment. (see PR description). This PR fixes broken framework code. Follow-up work should file these somewhere else.

PR #63, opened 2020-04-29
Add optional syncDbAfterWrite behavior and setting

ntouran:
The ARMI output database is created by the master node in parallel cases
and is actually written on a local scratch space during the run (for i/o
performance in large sweeps). There are some scenarios, however, where a
DB is needed by other MPI processes as they're doing their work (for
example, if a DB-derived history is needed). This feature just copies
the whole database from the local drive into the run's working directory
(often on a shared network drive) after every write.

This feature can be activted by setting the new syncDbAfterWrite user
setting to `true`.

Relatedly, MPI worker nodes now open the database with read-only
permissions after every DistributeState. This allows them to perform
DB-reading tasks during a run, now that the DB is available to them.

Also added optional Executer-level way to set a runPath.

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -206,6 +206,8 @@ class DatabaseInterface(interfaces.Interface):
                 time.time() - self.r.core.timeOfStart
             ) / 60.0
             self._db.writeToDB(self.r)
+            if self.cs["syncDbAfterWrite"]:
+                self._db.syncToSharedFolder()
 
     def interactEOC(self, cycle=None):
         """In case anything changed since last cycle (e.g. rxSwing), update DB. """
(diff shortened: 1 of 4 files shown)


armi/bookkeeping/db/database3.py
```diff
@@ -206,6 +206,8 @@ def interactEveryNode(self, cycle, node):
                 time.time() - self.r.core.timeOfStart
             ) / 60.0
             self._db.writeToDB(self.r)
+            if self.cs["syncDbAfterWrite"]:
```
youngmit, line 209: Could use the CONF_SYNC_AFTER_WRITE constant 🦆

ntouran:
tested manually. Travis will catch up soon.

PR #64, opened 2020-04-30
Fix bug when specifying GRID_PLATE

youngmit:
This fixes a bug preventing users from specifying `grid_plate` in blueprints. The implicit conversion would turn "grid" into "grid_plate", leaving "grid_plate_plate", which then fails. Using regular expressions to be more precise in the conversions resolves this

### DIFF
--- a/armi/reactor/flags.py
+++ b/armi/reactor/flags.py
@@ -80,7 +80,8 @@ True
 [<component fuel1>, <component fuel2>, ...]
 
 """
-from typing import Optional, Sequence, Union
+import re
+from typing import Any, Dict, Optional, Sequence, Union
 
 from armi.utils.flags import Flag, auto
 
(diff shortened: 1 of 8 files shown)


armi/reactor/tests/test_flags.py
```diff
@@ -53,9 +58,17 @@ def test_lookup(self):
...
-            if len(conv.split()) > 1:
-                self.assertEqual(flags.Flags.fromString(conv), flag)
+        for conv, flag in flags._CONVERSIONS.items():
+            # the conversions are specified as RE patterns, so we need to do a little
+            # work to get them into something that can serve as candidate input. Since
+            # we are only using \b and \s+, this is pretty straightforward. If any more
+            # complicated patterns work their way in there, this will need to become
+            # more sophisticated. One might be tempted to bake the plain-text versions
+            # of the conversions in the collection in the flags module, but this is
+            # pretty much only needed for testing, so that wouldn't be appropriate.
```
ntouran, line 68: Just so I understand, you're wanting to take the conversions and recover the possible user input that might lead to those conversions for test coverage?

armi/reactor/flags.py
```diff
@@ -95,10 +97,11 @@ def __fromStringGeneral(cls, typeSpec, updateMethod):
...
     result = cls(0)
     typeSpec = typeSpec.upper()
-    for conversion in CONVERSIONS:
-        if conversion in typeSpec:
-            typeSpec = typeSpec.replace(conversion, "")
-            result |= CONVERSIONS[conversion]
+    for conversion in _CONVERSIONS:
+        m = conversion.search(typeSpec)
+        if m:
+            typeSpec = conversion.sub("", typeSpec)
```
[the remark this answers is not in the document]
youngmit: Welcome to the awful mess that is the python standard library! It replaces the first instance of the regular expression in `conversion` in the second argument with the contents of the first argument, which feels like the opposite of the order in `str.replace()`, and feels unnatural even without the juxstaposition. Also it should //be called replace//!!! Anyways, I'll add a comment.

armi/reactor/tests/test_flags.py
youngmit: Yeah, just taking the regular expression string and converting it to something it would `match`

PR #66, opened 2020-05-05
Add support for supplying explicit flags in blueprints

youngmit:
This adds a `flags:` attribute to Assembly, Block, and Component
blueprints, allowing users to override the flags that would otherwise be
inferred from the blueprint `type`.

### DIFF
--- a/armi/reactor/blueprints/tests/test_componentBlueprint.py
+++ b/armi/reactor/blueprints/tests/test_componentBlueprint.py
@@ -21,6 +21,7 @@ import unittest
 
 from armi import settings
 from armi.reactor import blueprints
+from armi.reactor.flags import Flags
 from armi.nucDirectory import nuclideBases
 
 
@@ -30,6 +31,7 @@ class TestComponentBlueprint(unittest.TestCase):
 blocks:
     block: &block
         component:
+            flags: fuel test
             shape: Hexagon
             material: {material} # This is being used to format a string to allow for different materials to be added
             {isotopics} # This is being used to format a string to allow for different isotopics to be added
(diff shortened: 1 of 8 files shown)


armi/reactor/blueprints/tests/test_componentBlueprint.py
```diff
@@ -110,6 +112,12 @@ def test_componentInitializationControlCustomIsotopics(self):
         for nuc in unexpectedNuclides:
             self.assertNotIn(nuc, a[0][0].getNuclides())
 
+        c = a[0][0]
+        # Watch out, depletion is adding DEPLETABLE, so this can be a bit brittle
```
ntouran, line 116: is "depletion" going to make sense in this context?

armi/reactor/composites.py
```diff
@@ -647,10 +647,15 @@ def getType(self):
         """Return the object type."""
         return self.p.type
 
-    def setType(self, newType):
-        """Set the object type."""
-        self.p.flags = Flags.fromStringIgnoreErrors(newType)
-        self.p.type = newType
+    def setType(self, typ, flags: Optional[Flags] = None):
+        """
+        Set the object type.
```
ntouran, line 652: say what the type is and means in this docstring

armi/reactor/blueprints/tests/test_componentBlueprint.py
youngmit: not really... almost as if we shouldn't be sneakily adding the DEPLETABLE flag in the component blueprint in the first place, because it forces us to know about it in this context. Cat's kind of out of the bag, though. The purpose of the comment is to provide a hint to a future developer that tries to change that behavior as to why the test would start breaking.

PR #68, opened 2020-05-06
Add glob option to directory changer file retrieval.

ntouran:
Now you can pass wildcards in when you're not sure exactly
which output files will be generated.

### DIFF
--- a/armi/utils/directoryChangers.py
+++ b/armi/utils/directoryChangers.py
@@ -16,6 +16,7 @@ import os
 import random
 import shutil
 import string
+import glob
 
 import armi
 from armi import runLog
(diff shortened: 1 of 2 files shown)


armi/utils/directoryChangers.py
```diff
@@ -133,6 +134,19 @@ def _retrieveEntireFolder(self):
...
+        initialPath : str
+            Path to the folder to find files in.
+        destinationPath: str
+            Path to the folder to move file to.
+        fileList : list of str or list of tuple
+            File names to move from initial to destination. If this is a 
+            simple list of strings, the files will be transferred. Alternatively
+            tuples of (initialName, finalName) are allowed if you want the file
+            renamed during transit. In the non-tuple option, globs/wildcards
+            are allowed.
```
youngmit, line 149: Wildly pedantic, but what we really have here is a `List[Union[Tuple[str, str], str]]`. Is "list of str and/or tuple" valid numpydoc?

ntouran: I think it's at least close https://numpydoc.readthedocs.io/en/latest/format.html#sections

PR #71, opened 2020-05-08
Plot cartesian

onufer:
Holding off on removing capability of Cartesian to have two pitches for the moment. Will make ticket.


armi/reactor/tests/test_reactors.py
```diff
@@ -44,26 +44,54 @@
 TEST_REACTOR = None  # pickled string of test reactor (for fast caching)
 
 
-def buildOperatorOfEmptyBlocks(customSettings=None):
+def buildOperatorOfEmptyBlocks(customSettings=None, reactorType="Hex"):
```
ntouran, line 47: good to use `CONSTANTS` here, like `RX_TYPE_HEX`.

armi/reactor/tests/test_reactors.py
```diff
@@ -44,26 +44,54 @@
...
+        r = tests.getEmptyCartesianReactor()
+        AssemClass = assemblies.CartesianAssembly
+        BlockClass = blocks.CartesianBlock
+        CompClass = Rectangle
+        dims = {
+            "widthOuter": 16.0,
+            "lengthOuter": 10.0,
+            "widthInner": 1,
+            "lengthInner": 1,
+        }
```
ntouran, line 84: I think it might make sense to just make these totally independent without the input conditional/switch, even though they share some code.

armi/utils/plotting.py
```diff
@@ -38,3 +53,724 @@ def colorGenerator(skippedColors=10):
...
+
+    maxAssems: integer
+        maximum number of assemblies to plot in the assems list.
+
+    showBlockAxMesh: bool
+        if true, the axial mesh information will be displayed on the right side of the assembly plot.
+    """
+
+    if assems is None:
+        assems = list(core.parent.blueprints.assemblies.values())
```
youngmit, line 614: If the only thing that we are using the core for is to get the blueprints, then we should probably just pass the blueprints. Looking at the call site higher in the diff, I was confused to see that core was being passed at all, prompting me to dig in farther to see this.

armi/utils/plotting.py
```diff
@@ -38,3 +53,724 @@ def colorGenerator(skippedColors=10):
...
+    ax.set_yticks([0.0] + list(set(numpy.cumsum(yBlockHeightDiffs))))
+    ax.xaxis.set_visible(False)
+
+    ax.set_title("Assembly Designs for {}".format(core.name), y=1.03)
+    ax.set_ylabel("Thermally Expanded Axial Heights (cm)".upper(), labelpad=20)
+    ax.set_xlim([0.0, 0.5 + maxAssems * (assemWidth + assemSeparation)])
+
+    # Plot and save figure
+    ax.plot()
+    figName = core.name + "AssemblyTypes{}.png".format(plotNumber)
```
youngmit, line 678: Also probably not worth passing core in just for the name. Might make sense to pass name in on its own

armi/utils/plotting.py
```diff
@@ -38,3 +53,724 @@ def colorGenerator(skippedColors=10):
...
+    depthIndex=0,
+):
+    """
+    Plot a param distribution in xy space with the ability to page through depth.
+
+    Notes
+    -----
+    This is useful for visualizing the spatial distribution of a param through the core.
+    Blocks could possibly not be in alignment between assemblies, but the depths
+    viewable are based on the first fuel assembly.
```
youngmit, line 85: It may be worth exposing this through a simple script somewhere as an entry point. If this gets powerful enough, it may even be worth creating a proper entry point for it. At the very least, a usage example would be nice. It's just a bit of a shame to have a tool like this buried so deep

armi/reactor/tests/test_reactors.py
onufer: yeah, i had that thought after I got near the end.

armi/reactor/reactors.py
```diff
@@ -1811,547 +1815,6 @@ def createAssemblyOfType(self, assemType=None, enrichList=None, cs=None):
...
-                    runLog.error(
-                        "Cannot determine type letter of {0} assemblies. Are there more "
-                        "than 26 assembly types? If so, upgrade assembly type ID system."
-                        "Done letters are: {1}".format(aType, doneLetters)
-                    )
-                    raise RuntimeError("Failed to determine assembly type letter.")
-                assemTypeLetters[aType] = candidate
-
-        return assemTypeLetters
-
```
youngmit, line 2354: Mmmmmmmmm

armi/utils/plotting.py
onufer: Good point. I just moved this out of reactors and didn't change anything, so I didn't spend much time reading it.

armi/utils/plotting.py
onufer: like I said. its the same code as before, but I can look into how many cases need to be changed

armi/utils/plotting.py
onufer: Yeah, I wrote a test for demonstration, but I was also thinking of adding it to the ipython tutorial instead of one of the get face map calls. I'll definitely be using it when I put together more C5G7 stuff.

armi/reactor/reactors.py
onufer: yeah, everything was basically just moved (with some refactor on plot face map). the only thing that I actually got to delete was getAssemTypeLetters

armi/utils/plotting.py
onufer: added to data_model.ipynb

armi/utils/plotting.py
youngmit: fair enough

PR #73, opened 2020-05-14
Reduce Database Overwriting 

onufer:
DatabaseInterface now checks if the database file that it is
creating already exists and raises an error if it does. This
change is to prevent unfortunate overwrite of an old database.

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -175,13 +175,14 @@ class DatabaseInterface(interfaces.Interface):
         interface stack (so that all the parameters have been updated) while the Main
         Interface interacts first.
         """
-        if self.cs["reloadDBName"].lower() == (self.cs.caseTitle + ".h5").lower():
+        databaseName = self.cs.caseTitle + ".h5"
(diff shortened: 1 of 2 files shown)


youngmit:
How do we protect users from overwriting a database that they didn't intend to, without preventing them from overwriting a database that they expected to? One option would be to make this a setting query, which would be silent when submitting a job, or when using `--batch`, but prompt the user otherwise. Another option is to rename the old database. It is still not immediately obvious to me if that is the right behavior though. I can simply think of too many situations where re-running a case, overwriting a database is the expected behavior.

onufer:
yeah, Nick and I talked about it a little. One idea I had was when you are calling run or submit (and have db on), ARMI could delete the database since you are running ARMI in the traditional way. You could still still avoiding having interface.interactWhatever methods deleting your stuff when you are poking around with ipython.

Either way Nick said this should go to back burner for now.

jakehader:
This PR has been open for sometime and does not look like it's going any further (or maybe the issue was addressed separately). I suggest that you open an issue for the problem and we can revisit later as you said.

PR #74, opened 2020-05-15
Loading Unstable Nuclides from the RIPL-3 Database using the ARMI_RIPL_PATH Environment Variable

jakehader:
Adding an `ARMI_RIPL_PATH` environment variable to determine if
nuclide bases should read in unstable nuclide data from the RIPL-3
data files during the `factory` function call. These data files are not 
distributed with the framework, so a user will have to download them 
separately from `https://www-nds.iaea.org/RIPL-3/levels/`.

### DIFF
--- a/armi/nucDirectory/nuclideBases.py
+++ b/armi/nucDirectory/nuclideBases.py
@@ -16,6 +16,12 @@ r"""
 The nuclideBases module classes for providing *base* nuclide information, such as
 Z, A, state and energy release.
 
+By default the data used to create the nuclides comes from a subset of the RIPL-3
+database, which ships with ARMI. The entire RIPL-3 database can be downloaded from
+`<https://www-nds.iaea.org/RIPL-3/levels/>`_. If the ``ARMI_RIPL_PATH`` environment
+is set to a location containing the full RIPL-3 database, all nuclides will be read
+from there.
+
 The nuclide class structure is outlined in :ref:`nuclide-bases-class-diagram`.
 
 .. _nuclide-bases-class-diagram:
@@ -101,8 +107,10 @@ Table of All nuclides:
 """
 
 import os
+import pathlib
+import hashlib
+
 import yaml
-import collections
 
 import armi
 from armi.nucDirectory import elements
@@ -115,6 +123,7 @@ from armi.utils.units import HEAVY_METAL_CUTOFF_Z
 # unphysically. This is a bit of a crutch for the global state that is the nuclide
 # directory.
 _burnChainImposed = False
+_burnChainHash = None
 
 instances = []
 
@@ -128,7 +137,6 @@ byLabel = {}
 byMccId = {}
 
 byMcnpId = {}
-
 
 byAAAZZZSId = {}
 
@@ -147,6 +155,8 @@ BASE_ENDFB7_MAT_NUM = {
     "CF": 240,
     "TC": 99,
 }
+
+RIPL_PATH = None
 
 
 def isotopes(z):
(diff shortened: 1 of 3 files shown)


armi/nucDirectory/nuclideBases.py
```diff
@@ -378,15 +388,25 @@ def imposeBurnChain(burnChainStream):
...
-        # unphysical traits in the nuclide directory (e.g., duplicate decays and
-        # transmutations)
-        runLog.warning(
-            "Applying a burn chain when one has already been applied; "
-            "resetting the nuclide directory to it's default state first."
-        )
-        factory(True)
+        streamHash = hashlib.sha1()
+        streamHash.update(burnChainStream.read().encode())
+        burnChainStream.seek(0)
```
ntouran, line 395: 1. Add comments about what's going on here with cryptographic hashing
2. Consider using a faster non-cryptographic hash like https://docs.python.org/3/library/zlib.html#zlib.crc32

armi/nucDirectory/nuclideBases.py
```diff
@@ -440,6 +460,46 @@ def factory(force=False):
...
+        if riplDataPath is not None:
+            from armi.nuclearDataIO import ripl
+
+            ripl.makeDecayConstantTable(directory=riplDataPath)
+
+
+def _findRiplData():
+    """
+    Validates that the ``ARMI_RIPL_PATH`` environment variable has been set
+    and points to a valid directory for reading in the RIPL database files.
```
ntouran, line 473: docstring format should be 1 line summary, blank line, details

ntouran:
I'm excited about this feature. This will help anyone who wants to deal with all fission products. I have the in-line comments above, plus:

* Add documentation to the installation section of the user manual telling people about setting this environmental variable

armi/nucDirectory/nuclideBases.py
```diff
@@ -440,6 +460,46 @@ def factory(force=False):
...
+    """
+    global RIPL_PATH
+
+    riplEnvironVariable = "ARMI_RIPL_PATH"
+    if riplEnvironVariable not in os.environ:
+        return None
```
youngmit, line 479: You can still use the `get(..., None)` idiom, if you want. The problem from before is that we were doing `.get(..., "")` and passing that directly to `Path()`, which led to a `Path` pointing to the cwd. So:
```
riplEnvPath = os.environ.get("ARMI_RIPL_PATH", None)
if riplEnvPath is None:
    return None
path = pathlib.Path(riplEnvPath)
# etcetera
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -16,6 +16,12 @@
 The nuclideBases module classes for providing *base* nuclide information, such as
 Z, A, state and energy release.
 
+By default the data used to create the nuclides comes from a subset of the RIPL-3
+database, which ships with ARMI. The entire RIPL-3 database can be downloaded from
+`<https://www-nds.iaea.org/RIPL-3/levels/>`_. If the ``ARMI_RIPL_PATH`` environment
+is set to a location containing the full RIPL-3 database, all nuclides will be read
+from there.
+
```
youngmit, line 24: Now that I think about it, some justification as to why we are operating by default with a pruned set of nuclides, and/or why one would want the full data set could be useful. Something to the tune of "For most reactor calculations, it is typical to reduce the dimensionality of depletion by ignoring or lumping together nuclides that aren't important. For other tasks, like detailed decay or activation analysis, a full set of nuclides may be more important."

jakehader:
To explicitly highlight what this feature adds, i've written the nuclides that are processed within `nuclideBases.py` with and without the RIPL-3 data files. Working on addressing the other comments now.

[nuclides-with-ripl-3.txt](https://github.com/terrapower/armi/files/4636097/nuclides-with-ripl-3.txt)
[nuclides-without-ripl-3.txt](https://github.com/terrapower/armi/files/4636096/nuclides-without-ripl-3.txt)

armi/nucDirectory/nuclideBases.py
jakehader: Added this to the `user_install.rst` file and added a link to that documentation here.

ntouran:
looks good. in the squash commit msg be sure to note why you eliminated mc2 labels from the big table. Fine by me, just mention why.

PR #79, opened 2020-05-29
Fixes to Circular Ring HEX to RZ Conversion and Preloading ISOTXS on Core

jakehader:
A bug was found using the circular ring mode when converting between
the hex reactor to the rz reactor where total, fissile, and heavy metal
masses were not always being preserved. The circular ring mode conversions
were never used in practice, but having this available as an option could
lead to future issues for users.

Additionally, the `hexRingGeometryConversion` setting is being removed
since this is implied as the default behavior in `getAssembliesInRing`
on the core. This default is overwritten using the `circularRingMode`
setting.

Finally, the core.lib was assuming the existence of an `ISOTXS` file
in the working directory if it was not explicitly assigned. This is
generally not good practice as a cross section set from previous core
state could be accidently read in without the user realizing it. This is
being changed to happen explicitly only when `genXS` is not set versus
always implicitly being loaded. The library loading still occurs when `core.lib`
is called.

### DIFF
--- a/armi/physics/neutronics/settings.py
+++ b/armi/physics/neutronics/settings.py
@@ -88,7 +88,8 @@ def defineSettings():
             label="Multigroup Cross Sections Generation",
             description="Generate multigroup cross sections for the selected particle "
             "type(s) using the specified lattice physics kernel (see Lattice Physics "
-            "tab).",
+            "tab). When not set, the XS library will be auto-loaded from an existing ISOTXS "
+            "within then working directory and fail if the ISOTXS does not exist.",
             options=["", "Neutron", "Neutron and Gamma"],
         ),
         setting.Setting(
@@ -125,7 +126,7 @@ def defineSettings():
             default="DIF3D-Nodal",
             label="Neutronics Kernel",
             description="The neutronics / depletion solver for global flux solve.",
-            options=["DIF3D-Nodal", "DIF3D-FD", "VARIANT", "DIFNT",],  # lol
+            options=["DIF3D-Nodal", "DIF3D-FD", "VARIANT", "DIFNT"],
             enforcedOptions=True,
         ),
         setting.Setting(
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -224,7 +224,7 @@ class Core(composites.Composite):
         self._circularRingPitch = cs["circularRingPitch"]
         self._automaticVariableMesh = cs["automaticVariableMesh"]
         self._minMeshSizeRatio = cs["minMeshSizeRatio"]
-        self._preloadCoreXS = cs["preloadCoreXS"]
+        self._preloadCoreXS = True if not cs["genXS"] else False
 
     def __getstate__(self):
         """Applies a settings and parent to the core and components. """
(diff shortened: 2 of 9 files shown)


armi/reactor/reactors.py
```diff
@@ -292,8 +293,8 @@ def powerMultiplier(self):
     @property
     def lib(self):
         """"Get the microscopic cross section library."""
-        if self._lib is None:
-            runLog.info("Loading microscopic cross section library ISOTXS")
+        if self._lib is None and self._preloadCoreXS:
+            runLog.info(f"Loading existing ISOTXS microscopic cross section library")
```
ntouran, line 297: Should be a message here if lib is None and not preload, there should be a meaningful error pointing users to the preload setting when this starts failing on their previously-fine cases.

armi/reactor/reactors.py
```diff
@@ -293,8 +293,9 @@ def powerMultiplier(self):
     def lib(self):
         """"Get the microscopic cross section library."""
         if self._lib is None:
-            runLog.info("Loading microscopic cross section library ISOTXS")
-            self._lib = nuclearDataIO.ISOTXS()
+            runLog.warning(
+                f"A cross section library has not yet been assigned to {self}."
+            )
```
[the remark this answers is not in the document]
onufer: I agree that this is an important capability, but attaching the behavior loading an ISOTXS to a getLib function is a bit much.  Just pointing to that nuclearDataIO.ISOTXS() is a function that reads the file name "ISOTXS" unless a different kwarg is passed.

There are also other types of lib formats possible besides isotxs. I really like that Jake is making the getter a simple getter rather than a getter + read if it is None.

Maybe the warning message should suggest a way for the user to set up lib?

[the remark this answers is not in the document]
jakehader: @onufer, I think @ntouran, @youngmit and I have decided to maintain the functionality but I am trying to make it more explicit from a user perspective versus an automatic ISOTXS load when it's "needed". 

@ntouran also mentioned to me that ISOTXS is the "defacto standard" of ARMI, so I think that this should also be documented.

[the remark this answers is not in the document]
onufer: Yeah, I get that loading and ISOTXS should be easy, but I have heard Nick say methods should do one thing. I am not suggesting removing the functionality, but just separating it from the getter. Anyway, it sounds like you have already decided.

armi/reactor/reactors.py
jakehader: Added the check and added a `ValueError`.

armi/physics/neutronics/settings.py
```diff
@@ -125,7 +126,7 @@ def defineSettings():
             default="DIF3D-Nodal",
             label="Neutronics Kernel",
             description="The neutronics / depletion solver for global flux solve.",
-            options=["DIF3D-Nodal", "DIF3D-FD", "VARIANT", "DIFNT",],  # lol
```
ntouran, line 128: nice catch

armi/reactor/reactors.py
```diff
@@ -224,7 +224,7 @@ def setOptionsFromCs(self, cs):
         self._circularRingPitch = cs["circularRingPitch"]
         self._automaticVariableMesh = cs["automaticVariableMesh"]
         self._minMeshSizeRatio = cs["minMeshSizeRatio"]
-        self._preloadCoreXS = cs["preloadCoreXS"]
+        self._preloadCoreXS = True if not cs["genXS"] else False
```
ntouran, line 227: alternatively: `bool(cs["genXS"])`

but whatever

PR #83, opened 2020-06-09
Move T/H params to new T/H plugin

ntouran:
This is the start of a generic framework-level T/H plugin. We will move
param definitions, settings, validations, etc. to this


doc/conf.py
```diff
@@ -303,7 +303,7 @@ def setup(app):
 napoleon_use_param = True
 napoleon_use_rtype = True
 
-nbsphinx_kernel_name = "python"
+nbsphinx_kernel_name = "python3"
```
youngmit, line 306: AAAARRGGHHH python2, die already!!!!

PR #87, opened 2020-06-16
Add run-suite entry point for local param sweeps

ntouran:
A run-suite entry point is now provided, similar to the submit-suite
entry point available in some apps. This entry point runs all cases of a
suite one after the other, on the local machine. It is useful for
situations where a HPC is not available but a big workstation is.

Relatedly, some issues with paths in the parameter sweep system were
dealt with by allowing plugins to provide relative glob patterns that
are expanded at a higher level. This simplifies some plugin code (e.g.
when lattice physics specifies ISOTXS files) and makes the param sweep
system more capable.

### DIFF
--- a/armi/cli/runSuite.py
+++ b/armi/cli/runSuite.py
@@ -1,0 +1,98 @@
+# Copyright 2020 TerraPower, LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


armi/cli/runSuite.py
```diff
@@ -0,0 +1,98 @@
...
+            type=str,
+            default=[],
+            help="Pattern to search for inputs to ignore.",
+        )
+        self.parser.add_argument(
+            "--list",
+            "-l",
+            action="store_true",
+            default=False,
+            help="Just list the settings files found, don't actually submit them.",
```
youngmit, line 58: ~~submit~~ run them

armi/cli/runSuite.py
```diff
@@ -0,0 +1,98 @@
...
+                            case.cs.path,
+                            ",".join(pp.title for pp in case.dependencies),
+                        )
+                    )
+                runLog.important(
+                    tabulate.tabulate(
+                        tableData,
+                        headers=["Case title", "Settings path", "Parent tasks"],
+                    )
+                )
```
youngmit, line 92: Any reason to not just call `suite.echoConfiguration()`?

ntouran: looks like no

PR #91, opened 2020-06-30
Fix bug related to database file path

youngmit:
The path is not guaranteed to be set when interactDistributeState is
called, so it is important to check that there is a non-None path before
asking if the file exists.

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -275,7 +275,7 @@ class DatabaseInterface(interfaces.Interface):
         """
         if armi.MPI_RANK > 0:
             # DB may not exist if distribute state is called early.
-            if os.path.exists(self._dbPath):
+            if self._dbPath is not None and os.path.exists(self._dbPath):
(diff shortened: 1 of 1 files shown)


armi/bookkeeping/db/database3.py
```diff
@@ -275,7 +275,7 @@ def interactDistributeState(self):
         """
         if armi.MPI_RANK > 0:
             # DB may not exist if distribute state is called early.
-            if os.path.exists(self._dbPath):
+            if self._dbPath is not None and os.path.exists(self._dbPath):
```
ntouran, line 278: Almost feels like we could have a Database3.dbPathIsValid() class method or something in case this chain gets longer. But it probably won't get much longer...

youngmit: Perhaps. Fact of the matter is `_dbPath` is an `Optional[Pathlike]`, so it's not safe to use it as a path-like without checking for None. I'm not sure why mypy isn't catching this one, to be honest....

PR #92, opened 2020-06-30
Remove straggling references to external codes

youngmit:
This cleans up some reference to DIF3D, VARIANT, MCNP, etc. which were not resolved when ARMI was initially released. Some factory methods that weren't really being used, and are difficult to implement from the framework itself were also removed.


armi/physics/neutronics/latticePhysics/__init__.py
```diff
@@ -22,12 +22,3 @@
...
 ORDER = interfaces.STACK_ORDER.CROSS_SECTIONS
-
-
-def neutronicsKernelNeedsLatticePhysics(cs):
-    """Return True if the neutronics kernel requires cross sections from lattice physics"""
-    return cs["neutronicsKernel"] not in [neutronics.MCNP, neutronics.MCNP_DISTORTED]
-
-
-def latticePhysicsActive(cs):
-    return cs["globalFluxActive"] and neutronicsKernelNeedsLatticePhysics(cs)
```
ntouran, line 33: this is the one thing in here that feels pretty frameworky (except the mcnp specifics). I guess your point is that each individual set of plugins will have to make this decision on its own and there's not much value in sharing that code. Ok that makes sense actually... moving on.

youngmit: Yeah this sort of gets wrapped up in the need for some extra stuff on the neutronics interfaces to mediate this kind of thing. We can't just squint at the settings from lattice physics point of view and decide if we are needed.

PR #96, opened 2020-07-09
Improve setting rename/migration functionality

youngmit:
The previous mechanism for implicitly renaming old setting names to new
ones lived in settingsRules, which are difficult to extend from plugins.
As a result there were still a bunch of RENAMES for settings that have
since been removed from ARMI altogether. This includes a handful of
changes that allow these renames to live closer to the Settings to which
they apply, and therefore accessible to plugins. Changes include:

 - Introduce ArmiPluginManager. This is a subclass of the pluggy
 version, which adds a counter that is incremented any time the set of
 registered plugins changes. This allows clients to safely cache things
 derived from the plugins, and invalidate those caches as necessary. We
 may instead want to go the route of simply locking the plugin manager
 and preventing future modification, but this should work well for now,
 and is more flexible.

 - Add getSettings() method to the App class. The corresponding logic
 used to be called through the Settings.__init__() method, which was
 pretty odd.

 - Add the oldNames instance attribute to the Setting class. This is
 used to handle setting renames in lieu of settingsRules.RENAMES.

 - Add the SettingRenamer class to settingsIO, which distills the
 oldNames from the Setting instances and helps settingsIO perform
 migrations.

 - Merge the protected and public sections of SettingsReader. This was a
 holdover from when the SettingsReader doubled as a setting definition
 reader as well, and now only served to complicate matters.

 - Migrate the contents of RENAMES to their respective settings.

### DIFF
--- a/armi/apps.py
+++ b/armi/apps.py
@@ -177,7 +177,10 @@ class App:
         Return the parameter renames from all registered plugins.
 
         This renders a merged dictionary containing all parameter renames from all of
-        the registered plugins. It also performs simple error checking.
+        the registered plugins. It also performs simple error checking. The result of
+        this operation is cached, since it is somewhat expensive to perform. If the App
+        detects that its plugin manager's set of registered plugins has changed, the
+        cache will be invalidated and recomputed.
         """
         cacheInvalid = False
         if self._paramRenames is not None:
--- a/armi/apps.py
+++ b/armi/apps.py
@@ -28,10 +28,16 @@ customizing much of the Framework's behavior.
     object. We are planning to do this, but for now this App class is somewhat
     rudimentary.
 """
-from typing import Dict, Optional
+from typing import Dict, Optional, Tuple, List
+import collections
 
+import armi
 from armi import plugins
+from armi import pluginManager
 from armi import meta
+
+from armi.settings import Setting
+from armi.settings import fwSettings
 from armi.reactor import parameters
 
 
@@ -49,12 +55,7 @@ class App:
     data out of the Plugin API; calling the ``pluggy`` hooks directly can sometimes be a
     pain, as the results returned by the individual plugins may need to be merged and/or
     checked for errors. Adding that logic here reduces boilerplate throughout the rest
-    of the code. Another place where such code could go would be an ARMI subclass of the
-    ``pluggy.PluginManager`` class. For the time being, we are doing okay without having
-    to specialize the PluginManager, and we already have an App class to work with, so
-    this is a pretty good place. If at some point it makes sense to introduce stateful
-    plugins, or a more specialized PluginManager, then these methods should be migrated
-    there.
+    of the code.
     """
 
     name = "ARMI"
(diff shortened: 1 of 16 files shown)


armi/settings/settingsIO.py
```diff
@@ -16,67 +16,155 @@
...
+        """
+        Attempt to rename a candidate setting.
+
+        Renaming follows these rules:
+         - If the ``name`` corresponds to a current setting name, do not attempt to
+           rename it.
+         - If the ``name`` does not correspond to a current setting name, but is one of
+           the active renames, return the corresponding active rename.
+         - If the ``name`` does not correspond to a current setting name, but is one of
+           the expired renamse, produce a warning and do not rename it.
```
ntouran, line 92: spelling: renames

armi/settings/setting.py
```diff
@@ -110,7 +112,11 @@ def __init__(
             Whether this should be considered an "environment" setting. These can be
             used by the Case system to propagate environment options through
             command-line flags.
-
+        oldNames : list of tuple, optional
+            List of previous names that this setting used to have, along with optional
+            expiration dates. These can aid in automatic migration of old inputs. When
+            provided, if it is appears that the expiration date has passed, old names
+            will result in errors.
```
ntouran, line 119: maybe clarify this is expected only if an input file is loaded that has to be updated by the user to proceed with this version of the code.

armi/settings/settingsRules.py
```diff
@@ -12,7 +12,20 @@
...
+This module contains rules supporting custom actions taken on the reading of setting inputs
+
+Note
+----
+The ``settingsRules`` system was developed before the pluginification of ARMI, and
+doesn't play very nicely with the post-plugin world. This is mainly because
+registration of new rules happens at import time of the rule itself, which is
+unreliable and difficult to control, and even sort of implicit and sneaky. They are
+also somewhate redundant with the
+:py:class:`armi.operators.settingsValidation.Query`/:py:class:`armi.operators.settingsValidation.Inspector`
```
ntouran, line 25: might need spaces by `/` for sphinx to handle, not sure.

armi/apps.py
```diff
@@ -90,26 +91,106 @@ def __init__(self):
...
+
+        # The optionsCache stores options that may have come from a plugin before the
+        # setting to which they apply. Whenever a new setting is added, we check to see
+        # if there are any options in the cache, popping them out and adding them to the
+        # setting.  If all plugins' settings have been processed and the cache is not
+        # empty, that's an error, because a plugin must have provided options to a
+        # setting that doesn't exist.
+        optionsCache: Dict[str, List[armi.settings.Option]] = collections.defaultdict(
+            list
+        )
```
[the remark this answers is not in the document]
youngmit: perhaps, but less accurate. It's not actually a stack, and I'm not pushing/popping. Candidate options can be added and removed in any order 😝

armi/settings/settingsRules.py
youngmit: Seems happy with it

PR #98, opened 2020-07-12
GAMISO/PMATRX merging treatment with voided cross section sets

jakehader:
Updating the merging of GAMISO and PMATRX files when both gamma transport
and reactivity coefficients are enabled within the same run. This is
needed when considering a reactivity coefficient like Voided Doppler where
some cross section groups are updated to have a cross section library
suffix and others do not. This mirrors the behavior of the ISOTXS file
treatment.

### DIFF
--- a/armi/nuclearDataIO/__init__.py
+++ b/armi/nuclearDataIO/__init__.py
@@ -173,7 +173,9 @@ def _getGammaKeywords(cycle, suffix, xsID):
         elif xsID is not None:
             keywords = [xsID]
             if suffix not in [None, ""]:
-                keywords.append("-" + suffix)
+                if not suffix.startswith("-"):
+                    suffix = "-" + suffix
+                keywords.append(suffix)
         else:
             raise ValueError("The cycle or XS ID must be specified.")
         keywords.append(".")
--- a/armi/nuclearDataIO/xsLibraries.py
+++ b/armi/nuclearDataIO/xsLibraries.py
@@ -200,8 +200,22 @@ def mergeXSLibrariesInWorkingDirectory(lib, xsLibrarySuffix="", mergeGammaLibs=F
                 for nuc in neutronLibrary.nuclides
                 if isinstance(nuc._base, nuclideBases.DummyNuclideBase)
             ]
+
+            gamisoLibraryPath = nuclearDataIO.getExpectedGAMISOFileName(suffix=xsLibrarySuffix, xsID=xsID)
+            pmatrxLibraryPath = nuclearDataIO.getExpectedPMATRXFileName(suffix=xsLibrarySuffix, xsID=xsID)
+
+            # Check if the gamiso and pmatrx data paths exist with the xs library suffix so that
+            # these are merged in. If they don't both exist then that is OK and we can just
+            # revert back to expecting the files just based on the XS ID.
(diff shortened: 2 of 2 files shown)


armi/nuclearDataIO/xsLibraries.py
```diff
@@ -200,8 +200,22 @@ def mergeXSLibrariesInWorkingDirectory(lib, xsLibrarySuffix="", mergeGammaLibs=F
...
             ]
+
+            gamisoLibraryPath = nuclearDataIO.getExpectedGAMISOFileName(suffix=xsLibrarySuffix, xsID=xsID)
+            pmatrxLibraryPath = nuclearDataIO.getExpectedPMATRXFileName(suffix=xsLibrarySuffix, xsID=xsID)
+
+            # Check if the gamiso and pmatrx data paths exist with the xs library suffix so that
+            # these are merged in. If they don't both exist then that is OK and we can just
+            # revert back to expecting the files just based on the XS ID.
+            if not (os.path.exists(gamisoLibraryPath) and os.path.exists(pmatrxLibraryPath)):
+                runLog.warning(f"Neither the GAMISO or PMATRX data exist for "
```
youngmit, line 211: "One of GAMISO and/or PMATRIX data does not exist for..."

"Neither" would apply to logic like `if not os.path.exists(gamiso) and not os.path.exists(pmatrx)`

armi/nuclearDataIO/__init__.py
```diff
@@ -173,7 +173,10 @@ def _getGammaKeywords(cycle, suffix, xsID):
         elif xsID is not None:
             keywords = [xsID]
             if suffix not in [None, ""]:
-                keywords.append("-" + suffix)
+                if "-" in suffix:
```
youngmit, line 176: I had to do some head scratching to figure out what this was trying to do. Maybe
```
if not suffix.startswith("-"):
    suffix = "-" + suffix
```
would be clearer

youngmit: GH isnt letting me comment above your changes, but even though this is private, I think a short docstring or comment describing what this function is supposed to do would help a lot

armi/nuclearDataIO/xsLibraries.py
```diff
@@ -200,8 +200,22 @@ def mergeXSLibrariesInWorkingDirectory(lib, xsLibrarySuffix="", mergeGammaLibs=F
...
+                               f"XS ID {xsID} with suffix {xsLibrarySuffix}. "
+                               f"Attempting to find GAMISO/PMATRX data with "
+                               f"only XS ID {xsID} instead.")
+                gamisoLibraryPath = nuclearDataIO.getExpectedGAMISOFileName(xsID=xsID)
+                pmatrxLibraryPath = nuclearDataIO.getExpectedPMATRXFileName(xsID=xsID)
```
youngmit, line 216: Do we not want to check if these files exist, too?

jakehader: We could, but currently the ISOTXS files try to get merged and then the global flux solver just fails later on if something wasn't found. I kind of like that approach, but there is there weird "do ISOTXS and then do GAMISO/PMATRX similarly, but the code is a bit different to treat the file names and such."

armi/nuclearDataIO/xsLibraries.py
jakehader: Updating now.

PR #99, opened 2020-07-15
Move XSsettings

ZanderUF:
Moving XSsettings so they live within the neutronics settings and removed the XSsettings file


armi/settings/tests/test_writeYamltest_setting_io.yaml
```diff
@@ -0,0 +1,4 @@
+metadata:
+  version: 0.1.4
+settings:
+  nCycles: 55
```
youngmit, line 4: Oops, looks like you did a `git add .`

PR #108, opened 2020-07-30
Add basic fuel management demo with dummy burnup on a LWR core

ntouran:
Images look like: 

image

image


doc/gallery-src/framework/run_fuelManagement.py
```diff
@@ -0,0 +1,72 @@
...
+exclusions = []
+for num in range(12):
+    # find the 12 highest burnup fuel assemblies...
+    high = fuelHandler.findAssembly(
+        param="percentBu", compareTo=100, exclusions=exclusions, blockLevelMax=True
+    )
+    # and the 12 lowest burnup assemblies...
+    low = fuelHandler.findAssembly(
+        param="percentBu", compareTo=0, exclusions=exclusions, blockLevelMax=True
+    )
```
youngmit, line 55: `findAssembly()` is a pretty large, complex function with lots of kwargs that aren't super well-documented. As a result, it's not entirely clear what this is doing, or how it works, or where to go next. Perhaps we could identify the high/low-burnup assemblies with ARMI's rich tapestry of Composite model primitives? Something like `core.getAssemblies(Flags.FUEL)`, then sort those somehow by percentBu? Each of those little operations are simpler and easier to understand and better-documented

ntouran: good point. thanks for this suggestion and your offline recommendations. way better and more flexible now

PR #110, opened 2020-07-31
Add transmutation and decay plot to gallery

ntouran:
Make a nice plot of the nuclear transformations that comes with the nuclide library.

Found a alpha decay but along the way and fixed it.

image

### DIFF
--- a/armi/resources/burn-chain.yaml
+++ b/armi/resources/burn-chain.yaml
@@ -657,7 +657,7 @@ NP237:
     branch: 1.0
     halfLifeInSeconds: 67531821640000.0
     products:
-    - TH233
+    - PA233
     - DUMP2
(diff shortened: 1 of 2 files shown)


armi/resources/burn-chain.yaml
```diff
@@ -657,7 +659,7 @@ NP237:
     branch: 1.0
     halfLifeInSeconds: 67531821640000.0
     products:
-    - TH233
+    - PA233
```
youngmit, line 662: :+1:

ntouran: good news on this is that it doesn't affect anything except cases that have Np237 and a thorium decay chain because non-thorium cases just go straight to dump.

PR #116, opened 2020-08-05
Change assembly ordering to be i,j based instead of ring,pos

youngmit:
Ring and position are not always available, and therefore aren't a good
candidate about which to build fundamental concepts like object
ordering. This will unfortunately break unit tests and other code that
are sensitive to assembly (and therefore Block) ordering within a
reactor model.

### DIFF
--- a/armi/apps.py
+++ b/armi/apps.py
@@ -153,7 +153,9 @@ class App:
                         defaultsCache[pluginSetting.settingName] = pluginSetting
                 else:
                     raise TypeError(
-                        f"Invalid setting definition found: {pluginSetting}"
+                        "Invalid setting definition found: {} ({})".format(
+                            pluginSetting, type(pluginSetting)
+                        )
                     )
 
         if optionsCache:
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -129,12 +129,21 @@ class Assembly(composites.Composite):
         """
         Compare two assemblies by location.
 
-        This allows for consistent sorting of assemblies based on location. If either
-        assembly is not in the core (or more accurately a container), returns False.
-        This behavior may lead to some strange sorting behavior when two or more
-        Assemblies are being compared that do not live in the same grid. It may be
-        beneficial in the future to maintain the more strict behavior of ArmiObject's
-        ``__lt__`` implementation.
+        Notes
+        -----
+        As with other ArmiObjects, Assemblies are sorted based on location. Assemblies
+        are more permissive in the grid consistency checks to accomodate situations
+        where assemblies might be children of the same Core, but not in the same grid as
+        each other (as can be the case in the spent fuel or charge fuel pools). In these
+        situations, the operator returns ``False``.  This behavior may lead to some
+        strange sorting behavior when two or more Assemblies are being compared that do
+        not live in the same grid. It may be beneficial in the future to maintain the
+        more strict behavior of ArmiObject's ``__lt__`` implementation once the SFP/CFP
+        situation is cleared up.
+
+        See also
+        --------
+        armi.reactor.composites.ArmiObject.__lt__
         """
         try:
             return composites.ArmiObject.__lt__(self, other)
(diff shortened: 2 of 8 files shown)


armi/apps.py
```diff
@@ -153,7 +153,9 @@ def getSettings(self) -> Dict[str, Setting]:
                         defaultsCache[pluginSetting.settingName] = pluginSetting
                 else:
                     raise TypeError(
-                        f"Invalid setting definition found: {pluginSetting}"
+                        "Invalid setting definition found: {} ({})".format(
+                            pluginSetting, type(pluginSetting)
```
ntouran, line 157: I kind of prefer the f-string. 
```
f"Invalid setting definition found: {pluginSetting}\n"
f"({type(pluginSetting)})"
```

armi/reactor/assemblies.py
```diff
@@ -129,12 +129,16 @@ def __lt__(self, other):
         """
         Compare two assemblies by location.
 
-        This allow consistent sorting of assemblies based on location. If either
```
ntouran, line 132: probably should all be in a `Notes` thing talking about implementation

armi/reactor/assemblies.py
```diff
@@ -200,7 +204,14 @@ def getNum(self):
         return int(self.p.assemNum)
 
     def getLocation(self):
-        """Get string label representing this object's location."""
+        """
+        Get string label representing this object's location.
+
+        This function (and its friends) were created before the advent of both the
```
ntouran, line 210:
```
Notes
-----
```

armi/reactor/reactors.py
```diff
@@ -1160,6 +1160,15 @@ def getFirstAssembly(self, typeSpec=None, exact=False):
         """
         Gets the first assembly in the reactor.
 
+        Warning
+        -------
```
ntouran, line 1164: `Warnings` is a [valid numpydoc section heading](https://numpydoc.readthedocs.io/en/latest/format.html#method-docstrings) but `Warning` is not. 

Could also use `.. warning::`

armi/reactor/tests/test_zones.py
```diff
@@ -203,7 +203,11 @@ def test_createHotZones(self):
         # Create a single assembly zone to verify that it will not create a hot zone
         single = zones.Zone("single")
         daZones.add(single)
-        aLoc = r.core.getFirstAssembly(Flags.FUEL).getLocation()
+        aLoc = next(
+            a
+            for a in r.core.getAssemblies(Flags.FUEL)
+            if a.spatialLocator.getRingPos() == (1, 1)
+        ).getLocation()
```
ntouran, line 210: pretty sure there's a better way to get assemblies by ring and position. 

Hey that'd be a good gallery example... ;)

armi/apps.py
youngmit: It wasn't working for some reason... didn't want to spend too much time trying to figure it out.

armi/reactor/assemblies.py
youngmit: Good point. Should be better now.

armi/reactor/assemblies.py
youngmit: :+1:

armi/reactor/reactors.py
youngmit: Strange, it built just fine... Updating anyways

armi/reactor/tests/test_zones.py
youngmit: You mean a Composite method like `getChildAtLocation(loc: grids.SpatialLocator)`? That would be a wonderful idea :P Closest thing we have is the rather peculiar `getLocationContents()` on the Core, which uses the to-be defunct location labels. Unless I'm missing something...

armi/apps.py
ntouran: I bet you forgot the second `f`

PR #121, opened 2020-08-12
Add `doTH` T/H activation setting.

ntouran:
This setting can be used to activate any thermal/hydraulics plugin that
is incorporated into an ARMI app. It is being promoted from internal
plugins to a framework setting so that it can be shared across various
T/H plugins without coupling to any particular one.

### DIFF
--- a/armi/physics/thermalHydraulics/settings.py
+++ b/armi/physics/thermalHydraulics/settings.py
@@ -17,10 +17,19 @@
 from armi.settings import setting
 from armi.operators.settingsValidation import Query
 
+CONF_DO_TH = "doTH"
+
 
 def defineSettings():
     """Define generic thermal/hydraulic settings."""
-    settings = []
+    settings = [
+        setting.Setting(
+            CONF_DO_TH,
+            default=False,
+            label="Run Thermal Hydraulics",
+            description="Run thermal hydraulics",
+        ),
+    ]
     return settings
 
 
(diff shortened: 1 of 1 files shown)


armi/physics/thermalHydraulics/settings.py
```diff
@@ -17,10 +17,19 @@
...
 
 def defineSettings():
     """Define generic thermal/hydraulic settings."""
-    settings = []
+    settings = [
+        setting.Setting(
+            CONF_DO_TH,
+            default=False,
+            label="Run Thermal Hydraulics",
+            description="Run thermal hydraulics",
```
youngmit, line 30: 🦆 Might need a better description. In the global ARMI scope, what does this really mean in a context where any plugin might respond to it? Should we jump straight to a `thKernel` setting instead?

ntouran:
see also: #122 122

armi/physics/thermalHydraulics/settings.py
```diff
@@ -27,7 +27,16 @@ def defineSettings():
...
             label="Run Thermal Hydraulics",
-            description="Run thermal hydraulics",
+            description=(
+                f"Activate thermal hydraulics calculations using the physics module defined in "
+                f"`{CONF_TH_KERNEL}`"
+            ),
+        ),
+        setting.Setting(
+            CONF_TH_KERNEL,
+            default=False,
```
youngmit, line 37: Should this have options, default None?

youngmit: or default="", rather? Whatever appeases the settings validation. `neutronicsKernel` uses options=[] and default="".

PR #125, opened 2020-08-14
Fix some lint errors

ntouran:
This is broad set of fixes found by running pylint showing
just errors besides no-member errors.

Also, fix unit error in graphite thermal expansion.


armi/utils/tests/test_triangle.py
```diff
@@ -91,67 +91,6 @@ def test_checkIfPointIsInTriangle(self):
         # generalTriangleInOrOut = triangle.checkIfPointIsInTriangle(xT1, yT1, xT2, yT2, xT3, yT3, xP, yP)
         self.assertFalse(generalTriangleInOrOut)
 
-    def test_checkIfPointIsInTriangle(self):
```
youngmit, line 94: This does appear to be slightly different from the test above

PR #126, opened 2020-08-14
Implement proper CartesianGrid

youngmit:
This makes a new Grid subclass, imbued with meaningful concepts of
symmetry and ring/position.

### DIFF
--- a/armi/reactor/grids.py
+++ b/armi/reactor/grids.py
@@ -761,7 +761,6 @@ class Grid:
         For instance, if we have a 1/3rd core hex grid, this would return False for
         locators that are outside of the first third of the grid.
         """
-        return True
         raise NotImplementedError("Not implemented on base Grid type.")
 
     def _evaluateMesh(self, indices, stepOperator, boundsOperator):
@@ -972,11 +971,22 @@ class CartesianGrid(Grid):
         """
         Build a finite step-based 2-D Cartesian grid based on a width and height in cm.
 
+        Parameters
+        ----------
+        width : float
+            Width of the unit rectangle
+        width : float
+            Height of the unit rectangle
+        numRings : int
+            Number of rings that the grid should span
+        symmetry : str
+            The symmetry condition (see :py:mod:`armi.reactor.geometry`)
         isOffset : bool
             If True, the origin of the Grid's coordinate system will be placed at the
             bottom-left corner of the center-most cell. Otherwise, the origin will be
             placed at the center of the center-most cell.
-
+        armiObject : ArmiObject
+            An object in a Composite model that the Grid should be bound to.
         """
         unitSteps = ((width, 0.0, 0.0), (0.0, height, 0.0), (0, 0, 0))
         offset = numpy.array((width / 2.0, height / 2.0, 0.0)) if isOffset else None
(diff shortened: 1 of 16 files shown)


armi/reactor/grids.py
```diff
@@ -779,6 +761,7 @@ def locatorInDomain(self, locator: LocationBase) -> bool:
         For instance, if we have a 1/3rd core hex grid, this would return False for
         locators that are outside of the first third of the grid.
         """
+        return True
         raise NotImplementedError("Not implemented on base Grid type.")
```
ntouran, line 765: get rid of one or the othre

armi/reactor/grids.py
```diff
@@ -952,6 +933,112 @@ def getRingPos(self, indices) -> Tuple[int, int]:
...
+    def fromRectangle(
+        cls, width, height, numRings=5, symmetry=None, isOffset=False, armiObject=None
+    ):
+        """
+        Build a finite step-based 2-D Cartesian grid based on a width and height in cm.
+
+        isOffset : bool
+            If True, the origin of the Grid's coordinate system will be placed at the
+            bottom-left corner of the center-most cell. Otherwise, the origin will be
+            placed at the center of the center-most cell.
```
ntouran, line 978: add rest of 
```
Parameters
------------
```
to express at least symmetry and armiObject

armi/reactor/grids.py
```diff
@@ -952,6 +933,112 @@ def getRingPos(self, indices) -> Tuple[int, int]:
...
+            pos = 3 * ring - j
+        elif j == -ring:
+            # region 3
+            pos = 5 * ring + i
+        else:
+            # region 4
+            pos = 7 * ring + j
+        return (int(ring) + 1, int(pos) + 1)
+
+    def locatorInDomain(self, locator):
```
ntouran, line 1036: docstring for this would look good in sphinx, though if you want to rely o the docstring of the parent class that should be rendered appropriately using default sphinx options

armi/reactor/grids.py
```diff
@@ -967,17 +1054,65 @@ def changePitch(self, xw, yw):
         newOffsetY = self._offset[1] * yw / ywOld
         self._offset = numpy.array((newOffsetX, newOffsetY, 0.0))
 
-    @property
-    def pitch(self):
-        """
-        The pitch of the grid.
+    def getSymmetricEquivalents(self, indices):
```
ntouran, line 1057: same about docstring. I'm ok with this if the parent has a good one.

armi/reactor/grids.py
```diff
@@ -967,17 +1054,65 @@ def changePitch(self, xw, yw):
...
+                    #        QII           QIII          QIV
+                    return [(-i - 1, j), (-i - 1, -j - 1), (i, -j - 1)]
+
+        elif geometry.EIGHTH_CORE in self.symmetry:
+            raise NotImplementedError(
+                "Eighth-core symmetry isn't fully implemented " " for grids yet!"
+            )
+        else:
+            raise NotImplementedError(
+                "Unhandled symmetry condition for CartesianGrid: {}".format(
```
ntouran, line 1108: maybe `self.__class__.__name__` just in case?

armi/reactor/grids.py
youngmit: Oopsie, missed that!

armi/reactor/grids.py
youngmit: sphinx is re-using the base class docstring

armi/reactor/grids.py
youngmit: sphinx is re-using the base class docstring

PR #127, opened 2020-08-17
Remove enrichment method redundant with composite method.

ntouran:


ntouran:
@youngmit raised an important issue during discussions that ended up in #129.

Issue #129, opened 2020-08-18
Define and uniformly implement a typeSpec filter on composite methods as appropriate


ntouran:
In #127, @youngmit asked if we should add a Flags/typeSpec filter to the composite method that is being applied as the replacement for the outgoing `getEnrichment` block method. This is a good point and the answer is: most likely. 

* Methods on the composite tree that scan over children should offer some consistent and uniform method for filtering the children (e.g. the flags mechanism)

* We need a consistent definition of how the typespec would be applied. If you ask a core for fuel and it has fuel components in blocks that are marked "heat exchanger", would those pass the filter? this must be defined and documented. 

* which methods should we apply this kind of filtration to?

* Other issues?

ntouran:
Does anyone like the idea of embedding the `exact=True` metadata directly into the Flags? as like a special flag flag `Flags.EXACT`?

onufer:
Yeah I like that a lot. No longer being slightly uncertain about what the default is for a specific method.

youngmit:
Exact should not be a flag, but it would make a lot of sense to ship flags, plus exactness, and perhaps even depth semantics into a TypeSpec class of some sort. This would make consistent method signatures easier to maintain, and make it easy to document how filtering works

PR #138, opened 2020-08-25
Move many block methods to ArmiObject (temporarily).

ntouran:
Various block methods that do make logical sense at any level
of the reactor data model were moved to the ArmiObject. On one
hand this exasperates the fact that ArmiObject has way too
many public methods. On the other, it consolidates our tree-
traversing methods that are still in use into one place and
relieves some of the "specialness" of the Block. This is a
useful step in designing a better ArmiObject that will likely
rely more on composition of other functional classes which
will take these methods with them. A warning was added
to the code warning users that the ArmiObject API should
be considered quite unstable as we go through this.

* Remove the Batch class and its tests.

  The functionality of a Batch was useful but will be built in
  to the updated composite model once it's done. In the meantime
  it is not being used and is safe to remove.

  The motivation for removing it came when I noticed a discrepancy
  in the various set/add mass/masses methods.

* Move crossSectionTable attribute and methods out of data model

  These moved into the neutronics plugin. Caching may be reduced but
  can be re-implemented/optimized as necessary.


armi/reactor/composites.py
```diff
@@ -2300,6 +2321,254 @@ def getBoundingIndices(self):
...
+                )
+            return None
+        else:
+            raise ValueError(
+                "Multiple components match in {} match typeSpec {}: {}".format(
+                    self, typeSpec, results
+                )
+            )
+
+    def iterComponents(self, typeSpec: TypeSpec = None, exact=False):
```
onufer, line 2486: Surprised that get component is defined, but iter component is not. Might be worthwhile to add a comment describing why.

armi/reactor/composites.py
```diff
@@ -2300,6 +2321,254 @@ def getBoundingIndices(self):
...
+        raise NotImplementedError()
+
+    def getNumComponents(self, typeSpec):
```
onufer, line 2489: consider also an exact=False kwarg

armi/reactor/composites.py
```diff
@@ -2300,6 +2321,254 @@ def getBoundingIndices(self):
...
+                material is None
+            ), "Cannot call with more than one selector. Choose one or the other."
+
+        componentsWithThisMat = []
+        for c in self.getComponents():
+            if c.getProperties().getName() == materialName:
+                componentsWithThisMat.append(c)
+        return componentsWithThisMat
+
+    def hasComponents(self, typeSpec):
```
onufer, line 2404: exact=False could be good.

armi/reactor/composites.py
ntouran: good point, and this goes well with #129.

armi/reactor/composites.py
ntouran: The reason is because the implementation of getComponent is identical whether you're on a Composite or a Leaf whereas iterComponents is different for composite (loop over children) vs. leaf (yield self). I will try to make that more clear in the class docstring but I'm doing that to follow what (I think) the design pattern suggests.

onufer: Okay cool. I didn't realize that getComponent will actually raise not implemented since getComponents is not implemented. Its just a thin wrapper.

Might be worth considering moving iterComponents up to the other notImplemented methods like getComponents.

armi/physics/neutronics/isotopicDepletion/crossSectionTable.py
```diff
@@ -154,3 +154,68 @@ def getXsecTable(
                 dataToWrite["mcnpId"] = mcnpNucName
                 output.append(tableFormat.format(**dataToWrite))
         return output
+
+
+def getCrossSectionTable(obj, nuclides=None):
```
onufer, line 159: Since this method's main purpose was previously caching,  it might make sense to deprecate or delete. 

Right now it doesn't do anything besides provide a default kwarg.

armi/reactor/composites.py
ntouran: good idea on the move. Done.

ntouran:
At your convenience @youngmit can you audit this?

PR #139, opened 2020-08-25
Consolidate getDominantMaterial methods into one

ntouran:
There were quasi-independent implementations of getDominantMaterial
on three separate objects. Now they are all unified into the same
implementation. The special needs of various clients have been handled
by refactoring the functionality into two functions at the composite
module level. A warning was added indicating the high likelihood of these
functions being moved to something more topical of composition in the
near future.

In addition, several calls to `getComponents` were changed to `iterComponents`
mostly as an indication of where it's appropriate to use one over the other.
The performance enhancement of this is not expected to be large.

### DIFF
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -3224,7 +3224,7 @@ def gatherMaterialsByVolume(
             volumes[matName] = volumes.get(matName, 0.0) + vol
             if matName not in samples:
                 samples[matName] = c.material
-        return volumes, samples
+    return volumes, samples
 
 
 def getDominantMaterial(
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -874,7 +874,9 @@ class ArmiObject(metaclass=CompositeModelType):
         mass : float
             The mass in grams.
         """
-        return sum([c.getMass(nuclideNames=nuclideNames) for c in self.getComponents()])
+        return sum(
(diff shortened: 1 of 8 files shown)


armi/reactor/composites.py
```diff
@@ -2568,6 +2572,30 @@ def getAverageTempInC(self, typeSpec: TypeSpec = None, exact=False):
...
+        ----------
+        typeSpec : Flags or iterable of Flags
+            The types of components to consider (e.g. [Flags.FUEL, Flags.CONTROL])
+
+        Returns
+        -------
+        mats : dict
+            keys are material names, values are the total volume of this material in cm*2
+        samples : dict
+            keys are material names, values are the first Material of this name in the tree
```
ntouran, line 2589: oops I have to update this

armi/reactor/composites.py
```diff
@@ -3157,3 +3185,65 @@ def _enterExitHelper(self, func):
...
+    volumes = {}
+    samples = {}
+    for obj in objects:
+        for c in obj.iterComponents(typeSpec, exact):
+            vol = c.getVolume()
+            matName = c.material.getName()
+            volumes[matName] = volumes.get(matName, 0.0) + vol
+            if matName not in samples:
+                samples[matName] = c.material
+        return volumes, samples
```
ntouran, line 3227: gah indentation fail

ntouran:
I'm gonna send this one too and will need a post-push audit from @youngmit  for cleanups and whatnot.

PR #142, opened 2020-09-03
Add support for exporting VTK vis files 

youngmit:
This adds an ARMI entry point for converting a database file into a VTK file for visualization in tools like ParaView or Visit. Supports Hex, Cartesian and ThRZ meshes! Try it out with something like `armi vis-file [databasefile]


armi/bookkeeping/visualization/dumper.py
```diff
@@ -0,0 +1,48 @@
...
+        """Dump a single reactor state to the vis file."""
+
+    @abstractmethod
+    def initialize(self):
+        """
+        Prepare to write one or more reactor states. Called from __enter__.
+        """
+
+    @abstractmethod
+    def finalize(self):
```
ntouran, line 33: should be `_finalize`?

armi/bookkeeping/visualization/dumper.py
```diff
@@ -0,0 +1,48 @@
...
+
+from armi.reactor import reactors
+
+class VisFileDumper(ABC):
+    @abstractmethod
+    def dumpState(self, r: reactors.Reactor):
+        """Dump a single reactor state to the vis file."""
+
+    @abstractmethod
+    def initialize(self):
```
ntouran, line 27: Since called by the context manager should this be `_initialize`? Or do you expect clients to sometimes call it directly? If so then the `__enter__` thing in the docstring may be confusing

armi/bookkeeping/visualization/entryPoint.py
```diff
@@ -0,0 +1,113 @@
...
+from armi import runLog
+from armi.cli import entryPoint
+
+
+class VisFileEntryPoint(entryPoint.EntryPoint):
+    """
+    Create visualization files from database files.
+    """
+
+    name = "vis-file"
```
ntouran, line 27: is this using `vis` as the verb: _to visualize_ ? I guess that makes sense. At first I read it as a noun

armi/bookkeeping/visualization/entryPoint.py
```diff
@@ -0,0 +1,113 @@
...
+    description = "Convert ARMI databases in to visualization files"
+
+    _FORMAT_VTK = "vtk"
+    _SUPPORTED_FORMATS = {_FORMAT_VTK}
+
+    def __init__(self):
+        entryPoint.EntryPoint.__init__(self)
+
+        self.nodes = None
```
ntouran, line 36: nodes here without a docstring is confusing and I had to scroll down to figure out what it was. Can you add a 

```
self.nodes = None
"""Data base nodes to visualize"""
``` 
kinda docstring?

armi/bookkeeping/visualization/entryPoint.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def addOptions(self):
+        self.parser.add_argument("h5db", help="Input database path", type=str)
+        self.parser.add_argument(
+            "--output-name",
+            "-o",
+            help="Base to use output file name(s)",
```
ntouran, line 43: help message grammar issue?

`Base name for output files. Will have suffixes appended for timesteps`

maybe?

armi/bookkeeping/visualization/entryPoint.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if self.args.output_name is None:
+            # infer name from input
+            inp = pathlib.Path(self.args.h5db)
+            self.args.output_name = inp.stem
+
+    def invoke(self):
+        # late imports so that we dont have to import the world to do anything
+        from . import vtk
+        from armi.bookkeeping.db import databaseFactory
```
ntouran, line 86: add `pylint-disable` comments to get 10/10 pylint

armi/bookkeeping/visualization/entryPoint.py
```diff
@@ -0,0 +1,113 @@
...
+                raise RuntimeError(
+                    "Some of the requested nodes are not in the source database.\n"
+                    "Requested: {}\n"
+                    "Present: {}".format(nodes, dbNodes)
+                )
+
+            with dumper:
+                for cycle, node in dbNodes:
+                    runLog.info(
+                        "Dumping cycle {}, time node {}".format(cycle, node)
```
ntouran, line 108: Maybe more useful: `Creating visualization file for cycle ...`

armi/bookkeeping/visualization/tests/test_vtk.py
```diff
@@ -0,0 +1,98 @@
...
+from pyevtk.vtk import VtkTetra
+
+from armi.reactor.tests import test_reactors
+from armi.reactor import components
+from armi.reactor import blocks
+from armi import settings
+from armi.bookkeeping.visualization import vtk
+
+
+class TestVtkMesh(unittest.TestCase):
```
ntouran, line 27: quick desc of what this tests useful but not required. is this basically a test of the vtk system?

armi/bookkeeping/visualization/vtk.py
```diff
@@ -0,0 +1,363 @@
...
+
+"""
+Visualization implementation for VTK files.
+
+Limitations
+-----------
+This version of the VTK file writer comes with a number of limitations and/or aspects
+that can be improved upon. For instance
+ - Only the Block mesh and related parameters are exported to the VTK file. Adding
+   Assembly and Core meshes is totally doable, and will be the product of future work.
```
ntouran, line 23: not to mention components! A super epic later development might be plotting e.g pin-reconstructed fluxes in a whole core.

armi/bookkeeping/visualization/vtk.py
```diff
@@ -0,0 +1,363 @@
...
+
+                if len(nones) == data.shape[0]:
+                    # all Nones, so give up
+                    continue
+
+                if len(nones) == 0:
+                    # looks like Nones had nothing to do with it. bail
+                    continue
+
+                data = database3.replaceNonesWithNonsense(data, pDef.name, nones=nones)
```
ntouran, line 197: I still always 🔎 at this method name

ntouran (approved):
I love how compact the geometry specific code at the end is. 

You sure you don't wanna slap a screenshot in this PR?

armi/bookkeeping/visualization/dumper.py
youngmit: Something just feels weird to me about requiring context manageryness from an ABC. Maybe its not that crazy. I think ill just get rid of initialize and finalize, and implementers will just have to define __enter__ and __exit__ 🤷‍♀️

armi/bookkeeping/visualization/entryPoint.py
youngmit: i think of it as shorthand for "visualization file", but that isnt really a verb i suppose. "make-vis-file" is too long. "vis-file" just seems so _right_

PR #146, opened 2020-09-09
Changed functions that iterate and call getNumberDensity many times to more optimal methods 

ZanderUF:
Changed calls to getNumberDensity in loops to instead use  getNumberDensities or getNuclideNumberDensities.  This avoids some overhead in getNumberDensity that does not need to be repeated during iterative processes.


armi/reactor/blocks.py
```diff
@@ -757,8 +757,11 @@ def adjustDensity(self, frac, adjustList, returnMass=False):
         if returnMass:
             # do this with a flag to enable faster operation when mass is not needed.
             volume = self.getVolume()
-        for nuclideName in adjustList:
-            dens = self.getNumberDensity(nuclideName)
+
+        numDensities = self.getNuclideNumberDensities(adjustList)
+
+        for nuclideName,dens in zip(adjustList,numDensities):
```
ntouran, line 763: Lack of space after `,` indicates that you should run `black` on this file to get formatting perfect.

armi/reactor/blocks.py
```diff
@@ -1212,15 +1215,17 @@ def mergeWithBlock(self, otherBlock, fraction):
         into a full block.
         """
 
+        numDensities = self.getNumberDensities()
```
ntouran, line 1218: Going even further, this method and the `self.getNuclides()` call both have to run `getNuclides()` which is redundant. I recommend:

```
nuclideNames = self.getNuclides();
numDensities = self.getNuclideNumberDensities()

for nucName, ndens in zip(nucNames, nucDensities):
   set.setNum...
```

Oohh and if you want to get crazy you can use `setNumberDensities` to speed things up on the setting side. Though really that could be a different diff.

armi/reactor/blocks.py
```diff
@@ -1212,15 +1215,17 @@ def mergeWithBlock(self, otherBlock, fraction):
...
         for nuc in self.getNuclides():
-            self.setNumberDensity(nuc, (1.0 - fraction) * self.getNumberDensity(nuc))
+            self.setNumberDensity(nuc, (1.0 - fraction) * numDensities[nuc])
 
         # now add the other blocks densities.
         for nuc in otherBlock.getNuclides():
             self.setNumberDensity(
                 nuc,
-                self.getNumberDensity(nuc)
+                numDensities[nuc]
```
ntouran, line 1228: similar zip down here

armi/reactor/blocks.py
```diff
@@ -1462,9 +1467,11 @@ def getMfp(self, gamma=False):
         mfpNumerator = numpy.zeros(len(flux))
         absMfpNumerator = numpy.zeros(len(flux))
         transportNumerator = numpy.zeros(len(flux))
+
+        numDensities = self.getNumberDensities()
         # vol = self.getVolume()
         for nuc in self.getNuclides():
```
ntouran, line 1473: can do similar single `getNuclides` and then use `getNuclideNumberDensities` as recommended above.

armi/reactor/composites.py
```diff
@@ -2174,9 +2174,12 @@ def getAtomicWeight(self):
         """
         numerator = 0.0
         denominator = 0.0
+
+        numDensities = self.getNumberDensities()
+
         for nucName in self.getNuclides():
             atomicWeight = nuclideBases.byName[nucName].weight
```
ntouran, line 2181: and again!

armi/reactor/converters/geometryConverters.py
```diff
@@ -910,9 +910,12 @@ def createHomogenizedRZTBlock(
                     )
                 homBlockVolume += blockVolumeHere
                 homBlockTemperature += b.getAverageTempInC() * blockVolumeHere
-                for nucName in b.getNuclides():
+
+                numDensities = b.getNuclideNumberDensities(b.getNuclides())
+
+                for nucName, numDen in zip(b.getNuclides(), numDensities):
```
ntouran, line 916: recommending reuse the `getNuclides` resturn from above

ntouran (changes requested):
Just a few small requests, plus run the code formatter

armi/reactor/blocks.py
ZanderUF: Right, forgot about that.  Should be done.

ntouran (commented):
wonderful, thank you!!

ntouran (approved):

PR #147, opened 2020-09-09
Pull access of refAssem out of loop for speed

ntouran:
refAssem is a property now that is relatively expensive to access. This
method was unnecessarily accessing it N times. Now it is only accessed
once. This will speed up initialization of ARMI cases.


youngmit (approved):
Good find!

PR #155, opened 2020-09-16
Fix and improve mergeBlock

ntouran:
We recently broke mergeBlock a little but only saw it in the larger-scope in-use test suite. 
This fixes it and also goes a little farther in the optimization of getting/setting number densities.

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -1216,17 +1216,16 @@ class Block(composites.Composite):
         """
 
         numDensities = self.getNumberDensities()
+        otherBlockDensities = otherBlock.getNumberDensities()
+        newDensities = {}
 
-        # reduce this block's number densities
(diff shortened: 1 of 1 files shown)


armi/reactor/blocks.py
```diff
@@ -1216,17 +1216,16 @@ def mergeWithBlock(self, otherBlock, fraction):
         """
 
         numDensities = self.getNumberDensities()
+        otherBlockDensities = otherBlock.getNumberDensities()
+        newDensities = {}
 
-        # reduce this block's number densities
-        for nucName, nDen in numDensities.items():
-            self.setNumberDensity(nucName, (1.0 - fraction) * nDen)
+        # Make sure to hit all nuclides in union of blocks
```
ZanderUF, line 1222: This looks good and avoids repeating the same logic for the other block.

PR #157, opened 2020-09-19
Convert app tutorial to use submodule

ntouran:
This allows for better testing. 

Also added vtk image for output processing.


youngmit (approved):
I think this is okay to land for now, but I don't really like the idea of having a submodule just to support building docs. We should probably move pretty aggressively towards a separate docs/website repository for this kind of stuff, while leaving the in-code docs on the main ARMI repo

PR #158, opened 2020-09-19
Remove unused geomType from the constructAssem constructor

jakehader:
Remove unused geometry type from the blueprints constructAssem constructor.


youngmit (approved):
I don't think this will bump into my SFP work at all; I'm happy to merge this. @ntouran, did you have any concerns about this other than conflicts?

ntouran:
nope that was all. happy for the cleanup otherwise. go ahead and merge.

PR #159, opened 2020-09-19
Allow setting block heights to zero with DerivedShape components

jakehader:
A zero block height is supported in the framework when it is assigned in the blueprints and when there are no `DerivedShape`
components defined within the block. This aims to fix this issue to allow a user to specify a zero height block within an assembly even with a `DerivedShape` being used.

Additionally, there was an issue where while the blueprints could support adding a zero height block to an assembly, the `setHeight` method on blocks would not allow this. 

* The `setHeight` method on blocks now allows the user to set a zero height block and then return it to a non-zero height later on. 
* A unit test was added to check that setting a non-zero height block to zero and then back to the original height preserves the volume, area, number densities, and height.

The assignment of the `DerivedShape` component area was determined based on the computed volume and the height of the parent block. When the height was set to zero, this resulted in a `ZeroDivisionError`. Now, this exception is caught and the area of the `DerivedShape` component is instead set directly.

### DIFF
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -519,16 +519,15 @@ class Component(composites.Composite, metaclass=ComponentType):
         return self.p.volume
 
     def setVolume(self, volume):
+        self.p.volume = volume
         if not self.is3D:
             try:
                 self.setArea(volume / self.parent.getHeight())
             except ZeroDivisionError:
-                runLog.error(
-                    "The parent object {0} of {1} has no height. Cannot compute volume."
-                    "Check input.".format(self.parent, self)
-                )
-                raise
-        self.p.volume = volume
+                msg = (f"The parent object {self.parent} of {self} has no defined height and "
+                       f"therefore the area cannot be computed during the volume assignment. "
+                       f"The assigned volume is {volume} cc.")
+                raise ZeroDivisionError(msg)
 
     def setArea(self, val):
         raise NotImplementedError
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -521,13 +521,13 @@ class Component(composites.Composite, metaclass=ComponentType):
     def setVolume(self, volume):
         self.p.volume = volume
         if not self.is3D:
-            try:
-                self.setArea(volume / self.parent.getHeight())
-            except ZeroDivisionError:
+            if not self.parent.getHeight():
                 msg = (f"The parent object {self.parent} of {self} has no defined height and "
                        f"therefore the area cannot be computed during the volume assignment. "
                        f"The assigned volume is {volume} cc.")
-                raise ZeroDivisionError(msg)
+                raise ValueError(msg)
+            self.setArea(volume / self.parent.getHeight())
+
 
     def setArea(self, val):
         raise NotImplementedError
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -404,12 +404,15 @@ class Component(composites.Composite, metaclass=ComponentType):
 
     def getVolume(self):
         """
-        Return volume in cm**3.
-
-        Returns
-        -------
-        volume : float
-            volume in cm3
+        Return the volume [cm^3] of the component.
+        
+        Notes
+        -----
+        ``self.p.volume`` is not set until this method is called,
+        so under most circumstances it is probably not safe to
+        access ``self.p.volume`` directly. This is because not
+        all components (e.g., ``DerivedShape``) can compute
+        their volume during initialization.
         """
         if self.p.volume is None:
             self._updateVolume()
@@ -518,17 +521,9 @@ class Component(composites.Composite, metaclass=ComponentType):
     def getComponentVolume(self):
         return self.p.volume
 
-    def setVolume(self, volume):
-        self.p.volume = volume
-        if not self.is3D:
-            if not self.parent.getHeight():
-                msg = (f"The parent object {self.parent} of {self} has no defined height and "
-                       f"therefore the area cannot be computed during the volume assignment. "
-                       f"The assigned volume is {volume} cc.")
-                raise ValueError(msg)
-            self.setArea(volume / self.parent.getHeight())
-
-
+    def setVolume(self, val):
+        raise NotImplementedError
+    
     def setArea(self, val):
         raise NotImplementedError
 
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -758,9 +758,9 @@ class ArmiObject(metaclass=CompositeModelType):
             if child is self:
                 return frac
 
-    def _deriveUndefinedVolume(self):
+    def _deriveUndefinedVolumeAndArea(self):
         """
-        Determine the volume of any DerivedShapes (e.g. coolant components).
+        Determine the volume and area of any DerivedShapes (e.g. coolant components).
 
         When a coolant component first gets loaded, it has no area. But after that, it
         does have area so we must detect purely derived components by the fact that they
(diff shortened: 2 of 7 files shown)


armi/reactor/composites.py
```diff
@@ -750,17 +750,24 @@ def getVolumeFractions(self):
...
+            volFracs = [(ci, vi / vol) for ci, vi in zip(children, volumes)]
+        else:
+            volFracs = [(ci, 0.0) for ci in children]
+        return volFracs
 
     def getVolumeFraction(self):
         """Return the volume fraction that this object takes up in its parent."""
+        if self.parent is None:
+            return 0.0
+
```
ntouran, line 763: undefined or None or error is more appropriate here. Let's talk more about how to justify this or accomplish what you need less surprisingly

armi/reactor/composites.py
```diff
@@ -771,66 +778,80 @@ def _deriveUndefinedVolume(self):
...
-        return derivedVolume
+        # Determine the number of `DerivedShape` components and non-derived shape components
+        derivedShapeComps = [c for c in self.getChildren() if c.__class__ is components.DerivedShape]
+        otherComps = [c for c in self.getChildren() if c not in derivedShapeComps]
+
+        otherCompAreas = sum([c.getArea() for c in otherComps])
+        otherCompVolumes = sum([c.getVolume() for c in otherComps])
+
+        remainingArea = self.getMaxArea() - otherCompAreas
+        remainingVolume = self.getMaxVolume() - otherCompVolumes
```
ntouran, line 792: I think this is a potentially serious performance hit for blocks that do not have DerivedShapes in them, right? Can you try to keep this in a conditional that only happens when there's a derived shape?

(BTW we've been thinking of making the "DerivedShape" behavior specific to a block subclass that has that behavior so that may be the future state of this).

armi/reactor/composites.py
```diff
@@ -771,66 +778,80 @@ def _deriveUndefinedVolume(self):
...
+                c.setArea(remainingArea)
+
+        elif derivedShapeComps > 1:
+            msg = (f"The number of `DerivedShape` children exceeds 1 in {self} "
+                   f"and it is not possible to set the volume/area of these children.\n"
+                   f"Remove the multiple instances of these components to fully define the "
+                   f"geometry.\n"
+                   f"Derived shapes: {derivedShapeComps}")
+            runLog.error(msg)
+            raise ValueError(f"Multiple `DerivedShape` children exceed in {self}. Check log for errors.")
```
ntouran, line 852: `exceed` → `exist`

armi/reactor/composites.py
```diff
@@ -771,66 +778,80 @@ def _deriveUndefinedVolume(self):
...
+                c.setVolume(remainingVolume)
+            except ValueError:
+                c.setArea(remainingArea)
+
+        elif derivedShapeComps > 1:
+            msg = (f"The number of `DerivedShape` children exceeds 1 in {self} "
+                   f"and it is not possible to set the volume/area of these children.\n"
+                   f"Remove the multiple instances of these components to fully define the "
+                   f"geometry.\n"
+                   f"Derived shapes: {derivedShapeComps}")
```
ntouran, line 850: I think `"More than 1 DerivedShapes ({derivedShapeComps}) in {self} is not allowed"` is basically sufficient here.

armi/reactor/composites.py
```diff
@@ -771,66 +778,80 @@ def _deriveUndefinedVolume(self):
...
+            # A ZeroDivisonError can occur when the height of the component's
+            # parent is zero, thus the volume may be set within `setVolume`
+            # (using the derived volume calculated here), but the component's
+            # area cannot be determined by directly using this volume. In this
+            # case, the area can be computed by subtracting the area of all
+            # other components from the parent's maximum area.
+            try:
+                c.setVolume(remainingVolume)
+            except ValueError:
+                c.setArea(remainingArea)
```
ntouran, line 843: This is a little surprising at this level of abstraction and seems fragile or unpredictable. I think we may be able to make this cleaner by changing the framework behavior at some other level (e.g. within `c.setVolume`). 

We need to make `components` have height regardless of their parent.

ntouran (changes requested):

armi/reactor/composites.py
jakehader: That makes sense. Probably should just raise an error.

armi/reactor/composites.py
jakehader: So my understanding is that this only gets called on the `DerivedShape` class, but I agree that this can be put underneath the check that there is only 1 `DerivedShape` in the block.

armi/reactor/composites.py
jakehader: Alright, we can talk about this more.

armi/reactor/composites.py
```diff
@@ -771,66 +778,80 @@ def _deriveUndefinedVolume(self):
...
+                   f"geometry is defined correctly.\n"
+                   f"Maximum allowable volume: {self.getMaxVolume()} cc\n"
+                   f"Volume of all defined components: {otherCompVolumes} cc\n"
+                   f"Breakdown: {componentVols}")
+            runLog.error(msg)
+
+        if geometryError:
+            raise ValueError(f"Negative area/volume errors occurred for {self}. "
+                             "Check log for errors.")
+
```
youngmit, line 830: Handle the error if the number of derived shapes != 0 up here, so that the happy path doesnt need to be in a branch, and so that it matches the error-handling pattern above

armi/reactor/composites.py
```diff
@@ -771,66 +778,80 @@ def _deriveUndefinedVolume(self):
...
+            # (using the derived volume calculated here), but the component's
+            # area cannot be determined by directly using this volume. In this
+            # case, the area can be computed by subtracting the area of all
+            # other components from the parent's maximum area.
+            try:
+                c.setVolume(remainingVolume)
+            except ValueError:
+                c.setArea(remainingArea)
+
+        elif derivedShapeComps > 1:
```
youngmit, line 845: Move this case above with the other error checks. That it isnt checked with the other error conditions above makes me wonder "is this going to do something to handle more than one derived shape somehow?"

armi/reactor/composites.py
youngmit: If this only exists to make DerivedShape possible, then this logic should live in the DerivedShape class. This whole method is implemented using public data and interfaces.

youngmit: wait.... and its _private_?? Yeah, definitely just move this whole thing over to the DerivedShape class.

youngmit (changes requested):
I've got a handful of recommendations. The biggest one is that `_deriveUndefinedVolumeAndArea()` should just be `_deriveVolumeAndArea()` on the `DerivedShape` class. It probably shouldn't have been on the `Composite` in the first place.

armi/reactor/composites.py
jakehader: Moved up.

armi/reactor/composites.py
```diff
@@ -771,66 +778,80 @@ def _deriveUndefinedVolume(self):
...
-                )
-            runLog.error(
-                "These are leftovers: {0}\n"
-                " Cannot deduce area of more than one component".format(leftover)
-            )
-            raise RuntimeError("Gluttony Error too many leftovers.")
-        return derivedVolume
+        # Determine the number of `DerivedShape` components and non-derived shape components
+        derivedShapeComps = [c for c in self.getChildren() if c.__class__ is components.DerivedShape]
+        otherComps = [c for c in self.getChildren() if c not in derivedShapeComps]
```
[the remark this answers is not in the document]
jakehader: Okay, fixed.

armi/reactor/composites.py
```diff
@@ -771,66 +778,80 @@ def _deriveUndefinedVolume(self):
...
-                    "".format(c.getVolume(), c.p, c.getName())
-                )
-            runLog.error(
-                "These are leftovers: {0}\n"
-                " Cannot deduce area of more than one component".format(leftover)
-            )
-            raise RuntimeError("Gluttony Error too many leftovers.")
-        return derivedVolume
+        # Determine the number of `DerivedShape` components and non-derived shape components
+        derivedShapeComps = [c for c in self.getChildren() if c.__class__ is components.DerivedShape]
```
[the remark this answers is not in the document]
jakehader: Fixed.

armi/reactor/components/component.py
```diff
@@ -519,16 +519,15 @@ def getComponentVolume(self):
...
-                self.setArea(volume / self.parent.getHeight())
-            except ZeroDivisionError:
-                runLog.error(
-                    "The parent object {0} of {1} has no height. Cannot compute volume."
-                    "Check input.".format(self.parent, self)
-                )
-                raise
         self.p.volume = volume
+        if not self.is3D:
+            if not self.parent.getHeight():
```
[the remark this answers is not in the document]
jakehader: Hm. looks like we can probably just call setArea first and then call setVolume afterwards. Good call, ill try that.

armi/reactor/composites.py
```diff
@@ -750,17 +750,24 @@ def getVolumeFractions(self):
         children = self.getChildren()
         volumes = [c.getVolume() for c in children]
         vol = sum(volumes)
-        return [(ci, vi / vol) for ci, vi in zip(children, volumes)]
+        if vol > 0.0:
+            volFracs = [(ci, vi / vol) for ci, vi in zip(children, volumes)]
+        else:
+            volFracs = [(ci, 0.0) for ci in children]
+        return volFracs
```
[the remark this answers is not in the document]
jakehader: The components will have area fractions, but not volume fractions. This can call getAreaFractions() when volume is zero?

jakehader:
@ntouran, @youngmit - This is going to need a re-review since I changed a lot based on your comments. Essentially, the main difference is that I fixed up the hacky `setVolume` calls (removed them) because it doesn't make sense to set the volume of a component when it can be either calculated directly based on its dimensions, or if it can be derived as in the case with the `DerivedShape` component. 

This simplified the logic in my mind. Also, based on the comment from @youngmit, the volume/area derivation was removed from composites and moved into the `DerivedShape` class where it belongs. Now we no longer have to call `setArea` as `volume/self.parent.getHeight()`. Instead, the area is just set on the `DerivedShape` when the volume is determined so that they stay consistent. Additionally, this allows a `DerivedShape` to have an area even though the volume may be zero.

Finally, I updated the `getVolumeFractions` to return area fractions if the volume is zero. This was a comment from @youngmit that I think makes a lot of sense. Essentially, the volume fractions and area fractions should be identical for a block containing 2D extruded shapes, unless the block height is zero.

youngmit (approved):
Looks good! big step in the right direction

PR #160, opened 2020-09-21
Improve XDMF implementation

youngmit:
- Add time de-duplication. ParaView chokes on duplicate time entriees
 - Add error when absolute path is provided. Paraview doesn't like
   absolute paths for XDMF data sources. At least on Windows.


jakehader:
Added some suggestions.

jakehader (approved):

PR #162, opened 2020-10-02
Add flag list setting class

jakehader:
Adding a subclass to Setting to handle flag inputs. This automatically converts a list of strings to a list of flags 
when applying the value to a setting. Additionally, this will convert back to a list of strings
when writing to the Settings YAML or XML file.


youngmit (approved):
🌟

PR #165, opened 2020-10-05
Add support spent fuel pool in system blueprints

youngmit:
This adds rudimentary support for adding systems of the SpentFuelPool class in system blueprints. There is more work needed before this is line of work is complete. Not least of which will be the ability to subclass SystemBlueprint (which is itself a misnomer, because it's really more of a CoreBlueprint), which may require doing away with yamlize. SFP systems can be added now, but they will not be targeted by the fuel handler yet. That will be a future enhancement. Next steps:
 - Make SystemBlueprints subclassable; this will allow arbitrary collections of Systems definitions in the `systems:` blueprints section
 - Refactor the fuel handler code to locate a spent fuel pool off of the reactor, rather than using `core.sfp`

### DIFF
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -1369,7 +1369,7 @@ class FuelHandler:
 
     def dischargeSwap(self, incoming, outgoing):
         r"""
-        Removes one assembly from the core and replace it with another assembly
+        Removes one assembly from the core and replace it with another assembly.
 
         See Also
         --------
@@ -1397,10 +1397,16 @@ class FuelHandler:
         # which, coincidentally is the same time we're at right now at BOC.
         self.r.core.removeAssembly(outgoing)
 
+        # adjust the assembly multiplicity so that it doesnt forget how many it really
+        # represents
+        outgoing.p.multiplicity = len(loc.getSymmetricEquivalents()) + 1
+
         if incoming in self.r.core.sfp.getChildren():
             # pull it out of the sfp if it's in there.
             runLog.extra("removing {0} from the sfp".format(incoming))
             self.r.core.sfp.remove(incoming)
+
+        incoming.p.multiplicity = 1
         self.r.core.add(incoming, loc)
 
         self._swapFluxParam(incoming, outgoing)
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -1398,7 +1398,10 @@ class FuelHandler:
         self.r.core.removeAssembly(outgoing)
 
         # adjust the assembly multiplicity so that it doesnt forget how many it really
-        # represents
+        # represents. This allows us to discharge an assembly from any location in
+        # fractional-core models where the central location may only be one assembly,
+        # whereas other locations are more, and keep proper track of things. In the
+        # future, this mechanism may be used to handle symmetry in general.
         outgoing.p.multiplicity = len(loc.getSymmetricEquivalents()) + 1
 
         if incoming in self.r.core.sfp.getChildren():
--- a/armi/reactor/assemblyLists.py
+++ b/armi/reactor/assemblyLists.py
@@ -38,7 +38,8 @@ from armi.reactor.flags import Flags
 
 class AutoFiller(abc.ABC):
     """
-    Class for governing automatic insertion of Assemblies.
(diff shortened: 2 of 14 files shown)


armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1397,10 +1397,16 @@ def dischargeSwap(self, incoming, outgoing):
         # which, coincidentally is the same time we're at right now at BOC.
         self.r.core.removeAssembly(outgoing)
 
+        # adjust the assembly multiplicity so that it doesnt forget how many it really
+        # represents
+        outgoing.p.multiplicity = len(loc.getSymmetricEquivalents()) + 1
```
ntouran, line 1402: that's kind of a lot of work just to get the number. Maybe add explanation for why this was the best choice (related to center assem vs. edge assem vs. normal assem?

armi/reactor/assemblyLists.py
```diff
@@ -25,35 +36,136 @@
 from armi.reactor.flags import Flags
 
 
+class AutoFiller(abc.ABC):
+    """
+    Class for governing automatic insertion of Assemblies.
```
ntouran, line 41: be more specific: insertion of assemblies without specifying exactly which location of a container they go in.

armi/reactor/blueprints/reactorBlueprint.py
```diff
@@ -70,40 +73,88 @@ def __init__(self, name=None, gridName=None, origin=None):
...
-        container.spatialGrid = spatialGrid
-        container.spatialGrid.armiObject = container
-        reactor.add(container)  # need parent before loading assemblies
+            gridDesign = bp.gridDesigns.get(self.gridName, None)
+
+        system = self._resolveSystemType(self.typ)(self.name)
+
+        # TODO: obviously move someplace better. If system blueprints could be
+        # subclassed or whatever, this could live in the CoreBlueprint, but we aren't
+        # super wild about setOptionsFromCS() to begin with.
```
ntouran, line 119: would feel more proud of slightly more formal language here, but NBD.

armi/reactor/blueprints/reactorBlueprint.py
```diff
@@ -70,40 +73,88 @@ def __init__(self, name=None, gridName=None, origin=None):
...
+        if isinstance(system, reactors.Core):
+            system.setOptionsFromCs(cs)
+
+        # Some systems may not require a prescribed grid design. Only try to use on if
```
ntouran, line 123: to use *one*?

armi/reactor/blueprints/reactorBlueprint.py
```diff
@@ -116,12 +167,11 @@ def _loadAssemblies(self, cs, container, gridDesign, gridContents, bp):
 
             i, j = locationInfo
             loc = container.spatialGrid[i, j, 0]
-            if (
-                container.symmetry == geometry.THIRD_CORE + geometry.PERIODIC
-                and not container.spatialGrid.isInFirstThird(loc, includeTopEdge=True)
-            ):
+            try:
+                container.add(newAssembly, loc)
+            except exceptions.SymmetryError:
```
ntouran, line 172: will this  handle geoms that had the extra edge assemblies differently? If so we should add to list of changes that will eventually become release notes

armi/reactor/converters/geometryConverters.py
```diff
@@ -1288,7 +1296,7 @@ def convert(self, r=None):
                 self._newAssembliesAdded.append(newAssem)
 
         # set symmetry after expanding, because it isnt actually full core until it's
-        # full core.
+        # full core; setting the symmetry causes the core to rejigger itself.
```
ntouran, line 1299: "rejigger": be more specific.

ntouran (changes requested):

armi/reactor/blueprints/reactorBlueprint.py
youngmit: No, I had considered making this more strict, but decided to make the symmetry line part of the domain, so the behavior should be unchanged. The issue comes down to edge assems specifically; since they dont fill their respective grid locations, grid locations arent as useful as a determination of whether something is or isnt in the domain. we may need to make the idea of shared edges somewhat explicit to support split assems

youngmit: Alright, i bubbled the idea of `symmetryOverlap` up to the `Grid`; let me know what you think

ntouran (approved):

Issue #163, opened 2020-10-02
Document Flags


youngmit:
We have sort of been firing from the hip on what `reactor.Flags` really mean, and how they ought to be used. I think it's time to get a little more formal/precise with their definition. Flags settings from #162 is getting us closer to the vision of Flags-as-description with settings to map Flags to specific behavior. We should codify this in the docs.

jakehader:
We've had some discussions on this and the intent is that `Flags` can be used to define attributes of a composite, like `DUCT`, `FUEL`, `CLAD`, etc. as a way to differentiate between what the intended use of the composite will be within either the framework and/or plugins. 

When it comes to using `Flags` within the plugins, we are _not_ intending for them to be used to describe modeling behavior,  unless there is a specific setting that allows for the user to configure that behavior based on flags. One exception to this rule may be something like a `FuelHandler` plugin, where it may be obvious by the name of the plugin that the intent is to find the composites with `Flags.FUEL` and move them. Whereas if had a `Shuffler` plugin, we might want to have a setting to that describes the things can be shuffled/moved:

```
> cs["assemblyShuffleFlags"] = ["fuel", "control"]
> print(cs["assemblyShuffleFlags"]) 
> [Flags.FUEL, Flags.CONTROL]
```

PR #166, opened 2020-10-05
Expand on the definition of Flags

youngmit:
This is intended to address #163

### DIFF
--- a/armi/reactor/flags.py
+++ b/armi/reactor/flags.py
@@ -13,57 +13,82 @@
 # limitations under the License.
 
 """
-Handles *flags* that trigger behaviors related to reactor parts.
+Handles *flags* that formally bind certain categories to reactor parts.
 
-*Flags* are used to trigger certain treatments by the various modules. For instance, if
-a module needs to separate **reflector** assemblies from **fuel** assemblies, it can use
-flags.
+``Flags`` are used to formally categorize the various ``ArmiObject`` objects that make
+up a reactor model. These categories allow parts of the ARMI system treat different
+Assemblies, Blocks, Components, etc. differently.
 
-Flags are derived from the user-input name. If a valid flag name is included in a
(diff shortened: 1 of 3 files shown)


armi/reactor/flags.py
```diff
@@ -13,57 +13,82 @@
...
-ignored to avoid defining hundreds of flags for assemblies with many pins in them (e.g.
-in pin-level depletion cases). Use ``FUEL A`` or ``FUEL B`` to distinguish if necessary.
-
-Code modules that check the names of objects must use valid flags only.
-
-Things that flags are used for include:
+Handles *flags* that formally bind certain categories to reactor parts.
+
+``Flags`` are used to formally categorize the various ``ArmiObject`` objects that make
+up a reactor model. These categories allow parts of the ARMI system treat different
```
jakehader, line 19: `to treat`

armi/reactor/flags.py
```diff
@@ -13,57 +13,82 @@
...
+
+    Migrating the `meaning` of Flags into settings will allow us to better document how
+    those Flags/settings affect ARMI's behavior.
+
+    As mentioned above, plenty of code still hard-codes Flag ``TypeSpecs``, and certain
+    Flags do not follow the `what something is` convention. Future work should improve
+    upon this as possible.
+
+
+Things that Flags are used for include:
```
jakehader, line 69: Should we consider adding some discussion around the use of settings to define the behavior? Also, this is good documentation, but it seems like it should be discoverable in the docs as well. Should we link to this discussion?

youngmit: Yeah there should probably be something in the main doc/ tree pointing to this...

jakehader (changes requested):
Add links to the flags description in the main docs.

jakehader (approved):
🔥

youngmit:
Fixes #163

PR #167, opened 2020-10-06
Add support for ex-core structures to VTK files

youngmit:
Huzzah!

image

This used #165 to create the input database.


ntouran (approved):
works for xdmf too right?

youngmit:
> 
> 
> works for xdmf too right?

Yup, already did!

PR #168, opened 2020-10-06
Add support for reading/writing GEODST CCCC files

ntouran:
GEODST is a geometry specification file defined on the CCCC standard
interface committee. This adds the ability to read and write these files
in ASCII and binary format.

This also moves NHFLUX support out from a __init__ into its own module
with no changes.


armi/nuclearDataIO/geodst.py
```diff
@@ -0,0 +1,298 @@
...
+                # pylint: disable=protected-access
+                self._geom.fineMeshRegions[:, :, ki] = record._rwMatrix(
+                    self._geom.fineMeshRegions[:, :, ki],
+                    record.rwInt,
+                    self._metadata["NINTI"],
+                    self._metadata["NINTJ"],
+                )
+
+
+readBinary = GeodstStream.readBinary  # pylint: disable=invalid-name
```
youngmit, line 295: What's pylint's complaint here? Does it not know to look at the base class?

youngmit (changes requested):
The new fixture needs added to manifest.in. Other than that, looks good

armi/nuclearDataIO/geodst.py
ntouran: actually it has no problem!

youngmit (approved):

PR #170, opened 2020-10-06
Upgrade RTFLUX reader to be a reader/writer and add tests

ntouran:
This updates the RTFLUX reader to be more consistent with the other CCCC reader/writers.


armi/nuclearDataIO/rtflux.py
```diff
@@ -24,180 +41,172 @@
...
-        Read multigroup fluxes from the original DIF3D FD triangular-z mesh points.
-        RTFLUX contains the real fluxes, while ATFLUX contains the adjoint fluxes.
-
-        """
-
-        # Read basic data parameters (number of energy groups, assemblies, axial nodes, etc.)
-        self.read1D()
+class RtfluxData:
+    """
+    Data structure that is read from or written to a RTFLUX file.
```
youngmit, line 64: Its really nice to have separated this out from the act of reading, but now the question is... what is it?

armi/nuclearDataIO/rtflux.py
```diff
@@ -24,180 +41,172 @@
...
-        ]  # number of triangular mesh cells in "i" direction (rhombus or rectangle cells)
-        jmax = self.fc[
-            "nintj"
-        ]  # number of triangular mesh cells in "j" direction (rhombus or rectangle cells)
-        zmax = self.fc[
-            "nintk"
-        ]  # number of axial nodes (same for each assembly in DIF3D)
+    def __init__(self):
+        # Need Metadata subclass for default keys
+        self.metadata = nuclearFileMetadata._Metadata()
```
youngmit, line 69: Assuming that a user is intended to interact with this metadata, it probably shouldnt be a private thing.

armi/nuclearDataIO/rtflux.py
```diff
@@ -24,180 +41,172 @@
...
 
-        self.triangleFluxes = numpy.zeros((imax, jmax, zmax, ng))
+        self.groupFluxes = None
```
youngmit, line 71: What is this? An array of MG flux for a single region? A 2-D array by region/group?

armi/nuclearDataIO/rtflux.py
```diff
@@ -24,180 +41,172 @@
...
-        See Also
-        --------
-        rtflux.RTFLUX.__init__
-        nuclearDataIO.ISOTXS.read1D
-        nuclearDataIO.SPECTR.read1D
+    @classmethod
+    def _readWrite(cls, flux: RtfluxData, fileName: str, fileMode: str) -> RtfluxData:
+        with RtfluxStream(flux, fileName, fileMode) as rw:
+            rw.readWrite()
+        return flux
```
youngmit, line 116: Pretty squirrely, but 🤷‍♂️. How would I know how I am supposed to use this class? Seems like all of the client-facing stuff is actually on the `cccc.Stream` class? An example in the class docstring would be nice. Or an example in the `RtFluxData` of how to read one of those from one of these would be useful

youngmit: I think we might be shooting ourselves in the foot with the DRY-cannon with the structure of the cccc classes. Not worth fixing in this PR, but something to think about.

youngmit (changes requested):
Could mostly use docstrings for the client-facing things.

armi/nuclearDataIO/rtflux.py
ntouran: probably true. Can do a cccc cleanup pr at some point

armi/nuclearDataIO/rtflux.py
ntouran: I'm sure we can reduce the boilerplate but I am just copy/pasting it rather than DRYing it. I could have yet another parent class for all these CCCC rw-ers that specializes their data container. 

I used to think these were really complex but once you see the light it's quite glorious to be able to define the file structure once and get free reading AND writing!

ntouran:
ok I think I handled some of the suggestions

armi/nuclearDataIO/rtflux.py
youngmit: Sure, but that's an *implementer's* paradise, not a client's. Especially in the context of a file format reader/writer, these classes will be the subject of someone trying to figure out how to use them considerably more often than someone trying to create/modify one. Classic "you write code once, but read it many times" situation. Someone that thinks to themselves "ooh look, an RTFLUX reader! let's read me an RTFLUX," then proceeding to try to figure out how will probably struggle to, because the methods that they actually want to use are defined two classes away.

armi/nuclearDataIO/rtflux.py
```diff
@@ -61,14 +61,19 @@
...
+    This is the data structure that is read from or written to a RTFLUX file.
     """
 
     def __init__(self):
         # Need Metadata subclass for default keys
         self.metadata = nuclearFileMetadata._Metadata()
 
-        self.groupFluxes = None
+        self.groupFluxes: numpy.ndarray = numpy.array([])
+        """Maps i,j,k,g indices to total real or adjoint flux in n/cm^2-s"""
```
youngmit, line 76: do these indices correspond to our grid indexing?

ntouran: in some rare cases, maybe. But almost certainly not in most cases. for example in Finite difference cases there will be 6 or more of these per axial submesh of each block, and so on.

We could reconsider our grid indices based on these if we want since these have a longer legacy and there were smarter people doing nuclear models back then

youngmit (approved):

armi/nuclearDataIO/rtflux.py
youngmit: Probably worth mentioning whatever class could be used to make sense of the i, j, k indices

PR #172, opened 2020-10-06
Add read/writing of PWDINT power density CCCC file

ntouran:
This file contains the power density in W/cc vs. i,j,k mesh point for
any code that uses the CCCC standard interface files.

Did some minor refactoring for the block bandwidth code as well.


armi/nuclearDataIO/cccc.py
```diff
@@ -671,3 +671,23 @@ def writeAscii(cls, lib, fileName):
...
+    Return block bandwidth JL, JU from CCCC interface files.
+
+    It is common for CCCC files to block data in various record with
+    a description along the lines of::
+
+        WITH M AS THE BLOCK INDEX, JL=(M-1)*((NINTJ-1)/NBLOK +1)+1
+        AND JU=MIN0(NINTJ,JUP) WHERE JUP=M*((NINTJ-1)/NBLOK +1)
+
+    This function computes JL and JU for these purposes. It also converts
+    JL and JU to zero based indices rather than 1 based ones, as is almost
```
youngmit, line 687: Is this inconsistent with other interface files? I remember another one of these having a comment warning that the indices are one-based. Does this replace those? They should all be consistent across the board.

armi/nuclearDataIO/cccc.py
```diff
@@ -671,3 +671,23 @@ def writeAscii(cls, lib, fileName):
     @classmethod
     def _write(cls, lib, fileName, fileMode):
         raise NotImplementedError()
+
+
+def getBlockBandwidth(m, nintj, nblok):
+    """
+    Return block bandwidth JL, JU from CCCC interface files.
```
youngmit, line 678: Im not sure i understand what "bandwidth" means... is it like a stride of some sort?

youngmit (changes requested):
Just need more clarity on the "bandwidth" thing. Also, probably best to keep things 1-based unless everything is zero-based.

armi/nuclearDataIO/cccc.py
ntouran: yeah... I'll clarify

armi/nuclearDataIO/cccc.py
```diff
@@ -686,6 +686,10 @@ def getBlockBandwidth(m, nintj, nblok):
     This function computes JL and JU for these purposes. It also converts
     JL and JU to zero based indices rather than 1 based ones, as is almost
     always wanted when dealing with python/numpy matrices.
+
+    The term *bandwidth* refers to a kind of sparse matrix representation.
+    Some rows only have columns JL to JH in them rather than 0 to JMAX.
+    The non-zero band from JL to JH is what we're talking about here.
```
youngmit, line 692: Ooooh, so legit diagonal matrix bandwidth (https://en.wikipedia.org/wiki/Band_matrix#Bandwidth)?

youngmit (approved):

PR #173, opened 2020-10-08
Several fixes to the CCCC reader/writers

ntouran:
This fixes some issues related to the recent upgrades/changes to the CCCC reader/writers

* Fixes a class inheritance propagation bug affecting the new adjoint RTFLUX reader/writer
* Propagates that fix to other related classes in case they get subclassed later
* Fixes an i/j order bug on the new GEODST reader
* Implements a rwIntMatrix method and applies use of the related public methods 
   rather than the private _rwMatrix
* Adds new tests covering these bugs

### DIFF
--- a/armi/nuclearDataIO/cccc.py
+++ b/armi/nuclearDataIO/cccc.py
@@ -356,6 +356,12 @@ class IORecord(object):
         """
         return self._rwMatrix(contents, self.rwDouble, *shape)
 
+    def rwIntMatrix(self, contents, *shape):
+        """
+        Read or write a matrix of int values.
+        """
+        return self._rwMatrix(contents, self.rwInt, *shape)
+
     def _rwMatrix(self, contents, func, *shape):
         fortranShape = list(reversed(shape))
         if contents is None or contents.size == 0:
--- a/armi/nuclearDataIO/cccc.py
+++ b/armi/nuclearDataIO/cccc.py
@@ -362,7 +362,17 @@ class IORecord(object):
         """
         return self._rwMatrix(contents, self.rwInt, *shape)
 
-    def _rwMatrix(self, contents, func, *shape):
+    @staticmethod
+    def _rwMatrix(contents, func, *shape):
+        """
+        Read/write a matrix.
+
+        Notes
+        -----
+        This can be important for performance when reading large matrices (e.g. scatter
+        matrices). It may be worth investigating ``numpy.frombuffer`` on read and
+        something similar on write.
+        """
         fortranShape = list(reversed(shape))
         if contents is None or contents.size == 0:
             contents = numpy.empty(fortranShape)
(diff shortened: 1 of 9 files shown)


armi/nuclearDataIO/cccc.py
```diff
@@ -356,6 +356,12 @@ def rwDoubleMatrix(self, contents, *shape):
         """
         return self._rwMatrix(contents, self.rwDouble, *shape)
 
+    def rwIntMatrix(self, contents, *shape):
+        """
+        Read or write a matrix of int values.
+        """
+        return self._rwMatrix(contents, self.rwInt, *shape)
+
```
youngmit, line 364: While we're lookin at them, these could both be static methods

youngmit: My guess is there's a numpy function that could take a type and a data buffer and some size/stride info and turn it into an array. If we ever care enough about speed, might be worth thinking about using that below

armi/nuclearDataIO/rzflux.py
```diff
@@ -146,7 +146,7 @@ def _rw2DRecord(self):
         Zones are blocked into multiple records so we have to block or unblock
         them.
 
-        _rwMatrix reverses the indices into FORTRAN data order so be
+        rwMatrix reverses the indices into FORTRAN data order so be
```
youngmit, line 149: Worth considering https://numpy.org/doc/stable/reference/generated/numpy.frombuffer.html
That would read into a 1-D, then https://numpy.org/doc/stable/reference/generated/numpy.reshape.html#numpy.reshape to reshape. It even takes an arg for column-major or row major.

armi/nuclearDataIO/isotxs.py
```diff
@@ -175,7 +175,7 @@ def _write(cls, lib, fileName, fileMode):
 
     @classmethod
     def _readWrite(cls, lib, fileName, fileMode, getNuclideFunc):
-        with _IsotxsIO(fileName, lib, fileMode, getNuclideFunc) as rw:
+        with cls(fileName, lib, fileMode, getNuclideFunc) as rw:
```
youngmit, line 178: Still seems kind of squirrely

youngmit (approved):
Just a couple of recommendations. In general, I think the way all of these classes interact is pretty confusing. But if they work it really doesnt matter as long as the public interface is small and well-documented. The structure kind of makes that difficult, though...

armi/nuclearDataIO/isotxs.py
ntouran: 🐿️  
was squirrelier before when it was demolishing any inheritance by having the class be hard-coded!

armi/nuclearDataIO/rzflux.py
ntouran: great idea. Speed does kinda matter when we're reading big scatter matrices and stuff.

PR #174, opened 2020-10-09
Optimizations and refactoring in mesh and cccc

ntouran:
* Add a implicit-typed list reader to CCCC utils since we have 3
  implementations, and growing

* Make CCCC file names into module-level constants to reduce strings

* Reduce pointless mesh computations during reactor init

* Promote coverage to main requirements file since it is required if you
  activate the framework "coverage" setting.


armi/nuclearDataIO/cccc.py
```diff
@@ -689,6 +706,57 @@ def _write(cls, lib, fileName, fileMode):
         raise NotImplementedError()
 
 
+class DataContainer:
+    """
+    Data representation that can be read/written to/from with a Stream.
```
youngmit, line 711: from a `CCCC.Stream`. We are overloading a very meaningful term here with Stream, so should be careful.

armi/nuclearDataIO/cccc.py
```diff
@@ -689,6 +706,57 @@ def _write(cls, lib, fileName, fileMode):
...
+        # Need Metadata subclass for default keys
+        self.metadata = nuclearFileMetadata._Metadata()
+
+
+class StreamWithDataContainer(Stream):
+    """
+    A stream that reads/writes to a specialized data container.
+
+    This is a relatively common pattern so some of the boilerplate
+    is handled here.
```
youngmit, line 727: Might be worth discussing why not all CCCC.Streams use this pattern. Havent refactored yet?

armi/nuclearDataIO/cccc.py
```diff
@@ -689,6 +706,57 @@ def _write(cls, lib, fileName, fileMode):
...
+    """
+
+    def __init__(self):
+        # Need Metadata subclass for default keys
+        self.metadata = nuclearFileMetadata._Metadata()
+
+
+class StreamWithDataContainer(Stream):
+    """
+    A stream that reads/writes to a specialized data container.
```
youngmit, line 724: CCCC.Stream

armi/reactor/reactors.py
```diff
@@ -618,7 +618,7 @@ def getNumAssembliesWithAllRingsFilledOut(self, nRings):
             return nRings * (nRings - 1) + (nRings + 1) // 2
 
     def getNumEnergyGroups(self):
-        """"
+        """ "
```
youngmit, line 621: extra "

armi/reactor/reactors.py
```diff
@@ -1983,23 +1983,27 @@ def findAllAxialMeshPoints(self, assems=None, applySubMesh=True):
...
+        processLoading : sets up the master mesh that this perturbs.
         """
         # most of the time, we want fuel, but they should mostly have the same number of blocks
         # if this becomes a problem, we might find either the
         #  1. mode: (len(a) for a in self).mode(), or
         #  2. max: max(len(a) for a in self)
         # depending on what makes the most sense
         refAssem = self.refAssem
-
+        refMesh = self.findAllAxialMeshPoints([refAssem])
```
youngmit, line 2006: This is fine for now, but i honestly think we should consider a moratorium on `refAssem` until we figure out how we really want to handle it.

armi/reactor/reactors.py
```diff
@@ -2009,8 +2013,7 @@ def updateAxialMesh(self):
                         for h in [(b.p.ztop - b.p.zbottom) / b.p.axMesh] * b.p.axMesh
                     ]
                     for a in self
-                    if self.findAllAxialMeshPoints([a])
-                    == self.findAllAxialMeshPoints([refAssem])
+                    if self.findAllAxialMeshPoints([a]) == refMesh
```
youngmit, line 2016: Also pretty sketchy to be comparing a list of floats like this. Nothing new though

youngmit (changes requested):
This cuts down on some boiler plate, but I'm worried that CCCC is getting out of hand. It isn't clear where implementation details stop and client-facing APIs begin. The implementation is perfectly fine, but it is rather baroque, and exactly the kind of thing that "hide implementation details" is all about. We should probably push a lot of this stuff down into a sub-package of nuclearDataIO, and expose the public bits in its `__init__.py`

Seems to me like the only thing a user should see is a bunch of data container classes (one per file type), which have class methods for reading/writing from file/stream, plus whatever bespoke stuff each file should have based on its intended use.

armi/nuclearDataIO/cccc/__init__.py
```diff
@@ -0,0 +1,72 @@
+"""
+This subpackage reads and writes CCCC standard interface files for reactor physics codes.
+
+Starting in the late 1960s, the computational nuclear analysis community recognized a need to
+establish some standard file formats to exchange reactor descriptions and reactor physics
+quantities. They formed the Committee on Computer Code Coordination (CCCC) and issued
+several versions of their standards. The latest was issued in 1977 as [CCCC-IV]_. Many
+reactor codes to this day use these files. This package provides a Python abstraction to
+read many (though not necessarily all) of these files, manipulate the data writing, and
```
youngmit, line 9: manipulate the data ~~writing~~

armi/nuclearDataIO/cccc/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+While loading from stream classmethods is explicit and nice and all, there has been some
+talk about moving the read/write ascii/binary methods to the data classes for 
+implementations that use data structures. This would hide the Stream subclasses from
+users, which may be appropriate. On the other hand, logic to select which stream
+subclass to user (e.g. adjoint vs. real) will have to be moved into the
+data classes.
+
+Notes
+-----
+A CCCC record consists of a leading and ending integer, which indicates the size of the record in
```
youngmit, line 57: Pretty sure these are plain-old Fortran binary format. If true, might be nice to mention. Or not. Fun little ancient computer history tidbit that folks might delight in learning...

youngmit: Deets: https://gcc.gnu.org/onlinedocs/gfortran/File-format-of-unformatted-sequential-files.html

armi/nuclearDataIO/nuclearFileMetadata.py
```diff
@@ -68,6 +68,9 @@ def keys(self):
         """Returns keys similar to the dict implementation."""
         return self._data.keys()
 
+    def values(self):
+        return self._data.values()
+
```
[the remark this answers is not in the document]
ntouran: I sort of agree that it seems like a dict could do it, but there are complexities with gamsor vs. pmatrix vs. isotxs metadata that may have led to this. do you mind if I don't bite this one off at the moment?

armi/nuclearDataIO/cccc/cccc.py
```diff
@@ -14,159 +14,30 @@
...
+from typing import List
 
 import numpy
 
 from armi.localization import exceptions
 from armi import runLog
 
+from .. import nuclearFileMetadata
+
 IMPLICIT_INT = "IJKLMN"
```
youngmit, line 32: hahaha, no way... does CCCC use the implicit type rules from fortran?

armi/nuclearDataIO/cccc.py
```diff
@@ -689,6 +706,57 @@ def _write(cls, lib, fileName, fileMode):
...
+    """
+    Data representation that can be read/written to/from with a Stream.
+
+    This is an optional convenience class expected to be used in
+    concert with :py:class:`StreamWithDataStructure`.
+    """
+
+    def __init__(self):
+        # Need Metadata subclass for default keys
+        self.metadata = nuclearFileMetadata._Metadata()
```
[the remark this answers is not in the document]
ntouran: good point. probably could use .get() in all the `rw` things. not sure why this wasn't done.

armi/nuclearDataIO/cccc.py
```diff
@@ -689,6 +706,57 @@ def _write(cls, lib, fileName, fileMode):
...
+
+
+class StreamWithDataContainer(Stream):
```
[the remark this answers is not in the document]
ntouran: it's nearly possible to merge these into one class but for some complexities in the big boys: isotxs, gamsor, pmatrx

armi/nuclearDataIO/cccc/cccc.py
```diff
@@ -372,6 +243,11 @@ def _rwMatrix(contents, func, *shape):
         This can be important for performance when reading large matrices (e.g. scatter
         matrices). It may be worth investigating ``numpy.frombuffer`` on read and
         something similar on write.
+
+        It seems that with shape, the first shape argument should be the outermost loop.
+
+        So if you have ``((MR(I,J),I=1,NCINTI),J=1,NCINTJ)`` you would pass in
+        the shape as (NCINTJ, NCINTI).
```
youngmit, line 250: I think you can flip numpy.ndarray to use column major. No biggie though

armi/reactor/reactors.py
ntouran: huh. good old new black version?

armi/reactor/reactors.py
ntouran: sure but this just pulls it out of a loop.

armi/nuclearDataIO/cccc/cccc.py
```diff
@@ -381,6 +257,20 @@ def _rwMatrix(contents, func, *shape):
...
+    def rwImplicitlyTypedMap(self, keys: List[str], contents) -> dict:
+        """
+        Read a dict of floats and/or ints with FORTRAN77-style implicit typing.
+
+        Length of list is determined by length of list of keys passed in.
+        """
+        for key in keys:
+            # ready for some implicit madness from the FORTRAN 77 days?
+            if key[0].upper() in IMPLICIT_INT:
+                contents[key] = self.rwInt(contents[key])
```
youngmit, line 269: BONKERS!

armi/nuclearDataIO/cccc/cccc.py
```diff
@@ -665,30 +568,72 @@ def createRecord(self, hasRecordBoundaries=True):
...
     @classmethod
-    def readBinary(cls, fileName):
+    def readBinary(cls, fileName: str):
+        """Read data from a binary file into a data structure"""
         return cls._read(fileName, "rb")
 
     @classmethod
-    def readAscii(cls, fileName):
+    def readAscii(cls, fileName: str):
+        """Read data from an ASCII file into a data structure"""
```
youngmit, line 581: Still seems a shame that the user need to know up front whether theyre dealing with ascii or binary...

armi/physics/neutronics/const.py
```diff
@@ -52,3 +55,28 @@
...
 INPUTOUTPUT = "Input/Output"
 FLUXFILES = "Flux files"
+
+
+class SolutionType(Enum):
+    """Type of flux computed."""
+
+    REAL = 0
+    ADJOINT = 1
+    REAL_AND_ADJOINT = 2
```
youngmit, line 65: can you make these 1, 2 and 3? It would let us transition a bitfield and literally treat REAL_AND_ADJOINT as REAL & ADJOINT, as in Flags. Imagine a code that doesnt know how to do both at the same time, it wouldnt need a special case to do both, it could do if REAL in solType, do real solve; if ADJOINT in solType do adjoint solve;

youngmit: or make this a flags, with auto()?

armi/physics/neutronics/const.py
```diff
@@ -52,3 +55,28 @@
...
+
+    @staticmethod
+    def fromGlobalFluxOpts(opts):
+        """Infer solution type from GlobalFluxOptions object."""
+        if opts.real and opts.adjoint:
+            return SolutionType.REAL_AND_ADJOINT
+        elif opts.real:
+            return SolutionType.REAL
+        else:
+            return SolutionType.ADJOINT
```
youngmit, line 75: Flags would make this easier, too. `if opts.real: typ |= REAL`, `if opts.adjoint: typ |= ADJOINT`, no special case for both

armi/nuclearDataIO/cccc/nhflux.py
```diff
@@ -0,0 +1,477 @@
...
+        self.partialCurrentsZ: numpy.ndarray = numpy.array([])
+        self.incomingPointersToAllAssemblies: numpy.ndarray = numpy.array([])
+
+
+class NhfluxStream(cccc.StreamWithDataContainer):
+    @staticmethod
+    def _getDataContainer() -> NHFLUX:
+        return NHFLUX()
+
+    def readWrite(self):
```
youngmit, line 136: ugg and you cant make this private without having to re-do ISOTXS?

armi/nuclearDataIO/cccc/nhflux.py
```diff
@@ -0,0 +1,477 @@
...
+        r"""
+        Read everything from the DIF3D binary file NHFLUX that is necessary for pin flux and power reconstruction.
+
+        Read all surface-averaged partial currents, all planar moments,
+        and the DIF3D "four color" nodal coordinate mapping system.
+
+        Notes
+        -----
+        This method should be private but conflicts with ``_readWrite`` so we need a better name.
```
youngmit, line 145: `_readWriteImpl()`?

youngmit: impl, short for implementation; common idiom, in other languages anyways

armi/nuclearDataIO/cccc/cccc.py
youngmit: could even just do a `read()` that `try:`s to read binary, then falls back to ascii if that fails?

armi/nuclearDataIO/cccc/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+This is the original source of the code. TerraPower authorized the publication
+of some of the CCCC code to the PyNE project way back in the 2011 era. 
+This code has since been updated significantly to both read and 
+write the files.
+
+This was originally inspired by Prof. James Paul Holloway's alpha
+release of ccccutils written in c++ from 2001. 
+
+"""
+from .cccc import *
```
youngmit, line 72: We can tighten this up later, but this is the perfect place to serve as an aperture to control what a client sees within the package. Lots of the submodules themselves could be private (e.g. _nhflux.py), with the necessary (public) NHFLUX classes being imported/reexported here. Future work

youngmit (changes requested):
Looking better I think. Just a couple of things left and I'll be happy

armi/nuclearDataIO/cccc/__init__.py
ntouran: added link to discussion

armi/nuclearDataIO/cccc/cccc.py
ntouran: you can use `order="F"` to change how they're laid out in memory but the indexing stays constant I believe. But you're probably right that there's a better way to do this using column major fortran order. I'll at least babble about it in comments.

armi/nuclearDataIO/cccc/cccc.py
ntouran: THROWBACK!

armi/physics/neutronics/const.py
ntouran: I made a few code-specifics ones of these in an internal library and they correspond to specific values. Maybe I could move this back out of the framework because you're right about it not being great for flags.

armi/nuclearDataIO/cccc/nhflux.py
ntouran: this `readWrite` thing being public is one current bane of the cccc subpackags. It needs a new name and it needs to be private. It can't be `_readWrite` because we already have one. I was considering `_readWriteMain`, etc. for a while. 

Redoing isotxs won't be *too* hard.  It's pretty close.

armi/nuclearDataIO/cccc/nhflux.py
ntouran: ah yes I wrote that. Uh sure that'd be fine. I want to change all of them at once to do that so that m ight be in phase 5 cccc cleanup pr if that's ok

armi/nuclearDataIO/cccc/cccc.py
jakehader: Not sure if it does actually use them, but it does seem that way based on all the naming + types so far

armi/physics/neutronics/const.py
youngmit: Pick one. If we do it on the framework side, we need to do it right. Plugins can do what they want. Flags/bitfield is 100% the way to go. Maybe we can address this in a separate PR?

armi/nuclearDataIO/cccc/nhflux.py
youngmit: okay

armi/reactor/reactors.py
youngmit: an extra quote is an extra quote. my guess is black is hesitant to actually remove it

armi/physics/neutronics/const.py
```diff
@@ -55,28 +55,3 @@
 RESTARTFILES = "Restart files"
 INPUTOUTPUT = "Input/Output"
 FLUXFILES = "Flux files"
-
-
-class SolutionType(Enum):
```
youngmit: Thanks!

armi/__init__.py
```diff
@@ -400,6 +400,26 @@ def configure(app: Optional[apps.App] = None):
         runLog.raw(app.splashText)
 
 
+def applyAsyncioWindowsWorkaround():
```
youngmit, line 403: Ought to be a separate PR, but I'll allow it

youngmit (approved):
Looking good. So next steps:

- make stream classes private, 
- put public interface on data containers
- rename readWrite and make it private

PR #177, opened 2020-10-13
Allow more customization of entry point settings

ntouran:
This is useful, e.g. for making an entry point that has app-specific hard-coded settings that it wants to set at runtime of the entrypoint (e.g. allowing a user to add a few settings while hard-coding some others).

### DIFF
--- a/armi/cli/entryPoint.py
+++ b/armi/cli/entryPoint.py
@@ -90,9 +90,19 @@ class EntryPoint:
 
         self.args = argparse.Namespace()
 
-        self.cs = settings.Settings()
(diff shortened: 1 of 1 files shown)


armi/cli/entryPoint.py
```diff
@@ -90,10 +90,20 @@ def __init__(self):
 
         self.args = argparse.Namespace()
 
-        self.cs = settings.Settings()
+        self.cs = self._initSettings()
```
youngmit, line 93: one step closer to killing masterCs

youngmit (approved):

PR #192, opened 2020-10-23
Allow module path to be specified on input for material names

### LINKED ISSUE #114: Make the test reactor reliably use material definitions that ship with ARMI
There is no specification of the material namespace search order (`materialNamespaceOrder`) in the settings file for the test reactor, upon which lots of unit testing is performed. This leaves the actual state of that reactor sensitive to which plugins are active, potentially using plugin-provided material definitions. We should update the settings to explicitly request `armi.materials` so that the various physical properties can be relied upon in peoples' tests.

ntouran:
Fixes #114


ntouran:
I'm going to leave this feature but walk back the changes to the test inputs for now.

youngmit (approved):

PR #195, opened 2020-10-24
Remove deprecated Location objects

youngmit:
These have been removed in favor of the Grid and Locator classes in the `grids` module.
The Location classes stuck around for a long while due to the difficulty of refactoring certain code away from them. This does all of that work and removes the classes entirely, as well as various other framework sections that were defined around them (such as ArmiObject.LOCATION_CLASS, etc.)


armi/reactor/assemblies.py
```diff
@@ -257,7 +245,7 @@ def location(self):
 
         Just creates a new location object based on current spatialLocator.
         """
-        return self.getLocationObject()
+        raise NotImplementedError("NO MORE!!")
```
youngmit, line 248: looks like i can remove these now

armi/bookkeeping/historyTracker.py
```diff
@@ -331,8 +332,7 @@ def getTimeIndices(self, a=None, boc=False, moc=False, eoc=False):
                     )
                 # only add this timestep if it's around for this assembly.
                 if blockLocationLabel is not None:
-                    bLoc = locations.Location(label=blockLocationLabel)
-                    if bLoc.isInCore():
+                    if a.spatialLocator.grid is coreGrid:
```
ntouran, line 335: This condition can be checked at the top of the method right? rather than way down here?

armi/bookkeeping/report/reportingUtils.py
```diff
@@ -264,7 +264,8 @@ def writeAssemblyMassSummary(r):
         thisTypeList = core.getChildrenOfType(a.getType())
         count = 0
         for t in thisTypeList:
-            if t.getLocationObject().i1 == 1:
+            ring = t.spatialLocator.getRingPos()[0]
```
ntouran, line 267: I find `ring, _pos = t.spatialLocator.getRingPos()` to be slightly more self-documenting. nbd though

armi/bookkeeping/snapshotInterface.py
```diff
@@ -55,8 +55,6 @@ def interactEveryNode(self, cycle, node):
         snapText = getCycleNodeStamp(cycle, node)  # CCCNNN
         if self.cs["dumpSnapshot"] and snapText in self.cs["dumpSnapshot"]:
             self.o.snapshotRequest(cycle, node)
-        if self.cs["dumpLocationSnapshot"]:
-            self._dumpLocationSnapshot(cycle, node)
```
ntouran, line 59: you should add "Removed the location snapshot dump feature and setting" to release notes right now before we forget.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1966,6 +1969,16 @@ def buildEqRingSchedule(self, ringSchedule):
         locationSchedule : list
 
         """
+
+        def distanceFromOrigin(a):
+            origin = numpy.array([0.0, 0.0, 0.0])
+            p = numpy.array(a.spatialLocator.getLocalCoordinates())
+            return math.sqrt(((p - origin) ** 2).sum())
```
ntouran, line 1976: can almost definitely skip the sqrt and still be proportional to distance from origin. need new name then though.

armi/reactor/converters/blockConverters.py
```diff
@@ -354,7 +355,8 @@ def _addBlockRings(
             tempInput = tempHot = blockToAdd.getAverageTempInC()
 
         for ringNum in range(firstRing, firstRing + numRingsToAdd):
-            numFuelBlocksInRing = blockToAdd.location.getNumPosInRing(ringNum)
+            numFuelBlocksInRing = locations.numPositionsInRing(blockToAdd, ringNum)
```
ntouran, line 358: I thought this module was deleted. will read on...

armi/reactor/grids.py
```diff
@@ -390,6 +387,20 @@ def getSymmetricEquivalents(self):
         """
         return self.grid.getSymmetricEquivalents(self.indices)
 
+    def distanceTo(self, other) -> float:
+        """
+        Return the distance from this locator to another.
```
ntouran, line 392: worth mentioning that units will be cm.

armi/reactor/grids.py
```diff
@@ -983,6 +1008,7 @@ def pitch(self):
 
         Assumes 2-d unit-step defined (works for cartesian)
         """
+        #TODO move this to the CartesianGrid
```
ntouran, line 1011: did you mean to do this now?

armi/reactor/grids.py
```diff
@@ -1073,6 +1099,35 @@ def getRingPos(self, indices):
...
+        for ring in itertools.count(1):
+            ringPositions = self._getPositionsInRing(ring, throughCenter)
+            numPositions += ringPositions
+            if numPositions >= n:
+                return ring
+
+    @staticmethod
+    def _getPositionsInRing(ring, throughCenter):
+        # Keep private because this is really here to support old locations stuff that
+        # hasn't been removed yet.
```
ntouran, line 1122: not sure I understand that comment

ntouran (changes requested):
On big request: can you add a section to the user documentation showing the preferred ways to locate assemblies and blocks using locators? We will point to it when we push this out.

armi/bookkeeping/historyTracker.py
youngmit: I... don't know. I actually think this is broken in general, but only updated it so that it will be less confusing to the next person to come along. If you look at `_blockLocationAtTimenode()`, its making pretty bold assumptions about the database structure:
```
    def _blockLocationAtTimenode(self, block, timeNode):
        """Find block location label at a specific timenode."""
        dbi = self.getInterface("database")
        ids = dbi.database.readBlockParam("id", timeNode)
        locs = dbi.database.lookupGeometry()
```.

With the database-based history tracking, a lot of this can maybe go anyways, but that should be future work; I will add some comments to this effect.

File: armi/reactor/grids.py (hunk 8 above)
Comment (youngmit) [reply armi/reactor/grids.py:1011]: I was afraid.

File: armi/reactor/grids.py (hunk 9 above)
Comment (youngmit) [reply armi/reactor/grids.py:1122]: This is pretty weird and nuanced, so yeah probably needs a better comment. More context:
Cartesian grids need state (throughCenter) in order to reason about ring/pos. So for ring/pos to be part of the base Grid abstraction, we need to assume that _any_ Grid can only cough up ring/pos stuff with an instance method. HexGrid is sort of the special case, in that it _could_ do ring/pos stuff as a static method (notice that it punts the actual implementation to hexagon.py). There is no similar `rectangle.py` to hold the concept of ring/pos for the CartesianGrid case, and to do such a thing would be extra weird, because a throughCenter doesn't really mean anything to a simple rectangle; the grid adds the meaning to throughCenter.

File: armi/reactor/converters/blockConverters.py (hunk 6 above)
Comment (youngmit) [reply armi/reactor/converters/blockConverters.py:358]: still holds some legacy stuff that was hard to remove. Determination of ring/pos stuff without access to a grid instance, mostly. I added a bunch of context the the module level docstring of locations.py. let me know of that helps

File: armi/physics/fuelCycle/fuelHandlers.py [hunk 10]
```diff
@@ -1970,7 +1970,7 @@ def buildEqRingSchedule(self, ringSchedule):
 
         """
 
-        def distanceFromOrigin(a):
+        def squaredDistanceFromOrigin(a):
             origin = numpy.array([0.0, 0.0, 0.0])
             p = numpy.array(a.spatialLocator.getLocalCoordinates())
             return math.sqrt(((p - origin) ** 2).sum())
```
ntouran, line 1976: but still sqrt!

youngmit: 🤦‍♂️🤦‍♂️🤦‍♂️

ntouran (approved):
is that sqrt still in that distance squared function?

youngmit:
Not anymore! Sorry

PR #198, opened 2020-10-29
Checking ISOTXS existence before attempting autoload onto the core state

jakehader:
Checking for the existence of the `ISOTXS` file within the working when attempting to set 
the library if one does not already exist. Additionally, adding a check for an existing library 
within the uniform mesh converter before attempting calculation block reaction rates.

### DIFF
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -390,12 +390,13 @@ class NeutronicsUniformMeshConverter(UniformMeshGeometryConverter):
             _setStateFromOverlaps(
                 aSource, aDest, paramSetter, paramGetter, self.blockParamNames
             )
-            # Now recalculate derived params with the mapped flux to minimize
-            # potential numerical diffusion (e.g. control rod tip into large coolant)
-            for b in aDest:
-                globalFluxInterface.calcReactionRates(
-                    b, destReactor.core.p.keff, destReactor.core.lib
-                )
+            if destReactor.core.lib is not None:
+                # Now recalculate derived params with the mapped flux to minimize
+                # potential numerical diffusion (e.g. control rod tip into large coolant)
+                for b in aDest:
+                    globalFluxInterface.calcReactionRates(
+                        b, destReactor.core.p.keff, destReactor.core.lib
+                    )
 
 
 def _setNumberDensitiesFromOverlaps(block, overlappingBlockInfo):
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -285,6 +285,21 @@ class NeutronicsUniformMeshConverter(UniformMeshGeometryConverter):
     back to the source reactor.
     """
 
+    def __init__(self, cs=None, calcReactionRates=True):
+        """
+        Parameters
+        ----------
+        cs : obj, optional
+            Case settings object.
+
+        calcReactionRates : bool, optional
+            Set to True by default, but if set to False the reaction
+            rate calculation after the neutron flux is remapped will
+            not be calculated.
+        """
+        UniformMeshGeometryConverter.__init__(self, cs)
+        self.calcReactionRates = calcReactionRates
+
     def _checkConversion(self):
         """
         Make sure both reactors have the same power and that it's equal to user-input.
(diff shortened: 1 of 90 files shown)


armi/reactor/reactors.py
```diff
@@ -293,15 +292,26 @@ def powerMultiplier(self):
...
+        This returns the current library object (default as ``ISOTXS``) stored on
+        the core state. If no library has been set on the core yet, this will
+        load in an existing ``ISOTXS`` file from the working directory if one
+        exists. If an ``ISOTXS`` file does not exist in the working directory
+        no library will be loaded.
+        """
+        isotxsFileName = nuclearDataIO.getExpectedISOTXSFileName()
+        if self._lib is None and os.path.exists(isotxsFileName):
+            runLog.info(f"Loading microscopic cross section library `{isotxsFileName}`")
             self._lib = nuclearDataIO.ISOTXS()
```
ntouran, line 309: this needs to use the expected name

armi/reactor/converters/uniformMesh.py
```diff
@@ -390,12 +390,13 @@ def adjointFluxGetter(block, _paramNames):
             _setStateFromOverlaps(
                 aSource, aDest, paramSetter, paramGetter, self.blockParamNames
             )
-            # Now recalculate derived params with the mapped flux to minimize
-            # potential numerical diffusion (e.g. control rod tip into large coolant)
-            for b in aDest:
-                globalFluxInterface.calcReactionRates(
-                    b, destReactor.core.p.keff, destReactor.core.lib
-                )
+            if destReactor.core.lib is not None:
```
ntouran, line 393: needs explanation. It's unclear from reading the code here why this would happen and also why it's ok to just skip it if the library is none.

armi/reactor/reactors.py
ntouran: I also feel like this should crash if the file isn't there. Can you explain what scenario needs this?

armi/reactor/reactors.py
```diff
@@ -293,15 +292,26 @@ def powerMultiplier(self):
 
     @property
     def lib(self):
```
youngmit, line 294: -> Optional[XSLibraryOrWhateverBaseClass]

armi/reactor/reactors.py
```diff
@@ -293,15 +292,26 @@ def powerMultiplier(self):
...
-        """"Get the microscopic cross section library."""
-        if self._lib is None:
-            runLog.info("Loading microscopic cross section library ISOTXS")
+        """
+        Return the microscopic cross section library.
```
youngmit, line 296: ...if there is one

armi/reactor/reactors.py
```diff
@@ -293,15 +292,26 @@ def powerMultiplier(self):
...
+
+        Notes
+        -----
```
youngmit, line 299: Probably shouldnt be in Notes, as it is fundamental to what this is doing. I think this would be better as a bulleted list; something like:
 - if there is a library currently associated with the Core, it will be returned
 - otherwise, an "ISOTXS" file will be be searched for in the working directory, opened as an ISOTXS object and returned,
 - otherwise, None will be returned

armi/reactor/reactors.py
youngmit: There simply is no guarantee that a Core has a lib, or can get one. If this just blows up when anyone asks, then the only way to check _whether_ it has one is in a try/except, which is not ideal. Code that needs a lib is welcome to pitch a fit, and code that wants one can check `is not None:` and roll with it if it doesn't exist.

armi/reactor/converters/uniformMesh.py
youngmit: That a lib might not exist should be made clear from the Core class itself.

armi/reactor/converters/uniformMesh.py
```diff
@@ -390,12 +390,13 @@ def adjointFluxGetter(block, _paramNames):
...
+                # Now recalculate derived params with the mapped flux to minimize
+                # potential numerical diffusion (e.g. control rod tip into large coolant)
```
youngmit, line 395: Why it's okay not to do this if there is no lib does make sense to explain. "if we don't have a lib, then XS haven't been provided or calculated, stands to reason that no reaction rates/flux has been calculated either"

youngmit (changes requested):

armi/reactor/reactors.py
```diff
@@ -291,22 +292,21 @@ def powerMultiplier(self):
         return geometry.SYMMETRY_FACTORS[self.symmetry]
 
     @property
-    def lib(self):
+    def lib(self) -> Optional[xsLibraries.IsotxsLibrary]:
```
youngmit, line 295: Does this have to be IsotxsLibrary specifically? Could it be the base... oh, nvm `_XSLibrary` is priavate?

jakehader: Yes it's private.

youngmit (approved):

PR #201, opened 2020-11-05
Add support for converting old database flags

youngmit:
Before this, if flags are added to an application, databases from a
previous version would be unreadable, because the Flags serializer makes
sure that the meaning of all flags is the same between the reading app,
and the app that wrote the database. Now, it will only check that all
old flags still exist, but permit converting to the new Flag numeric
values.

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -673,13 +673,23 @@ class Database3(database.Database):
 
         cs = settings.Settings()
         cs.caseTitle = os.path.splitext(os.path.basename(self.fileName))[0]
-        cs.loadFromString(self.h5db["inputs/settings"][()])
+        try:
+            cs.loadFromString(self.h5db["inputs/settings"][()])
+        except:
+            pass
+
         return cs
 
     def loadBlueprints(self):
         from armi.reactor import blueprints
 
-        bpString = self.h5db["inputs/blueprints"][()]
+        bpString = None
+
+        try:
+            bpString = self.h5db["inputs/blueprints"][()]
+        except:
+            pass
+
         if not bpString:
             # looks like no blueprints contents
             return None
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -675,7 +675,14 @@ class Database3(database.Database):
         cs.caseTitle = os.path.splitext(os.path.basename(self.fileName))[0]
         try:
             cs.loadFromString(self.h5db["inputs/settings"][()])
-        except:
+        except KeyError:
+            # not all paths to writing a database require inputs to be written to the
+            # database. Technically, settings do affect some of the behavior of database
+            # reading, so not having the settings that made the reactor that went into
+            # the database is not ideal. However, this isn't the right place to crash
+            # into it. Ideally, there would be not way to not have the settings in the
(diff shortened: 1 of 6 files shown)


armi/bookkeeping/db/database3.py
```diff
@@ -673,13 +673,23 @@ def loadCS(self):
 
         cs = settings.Settings()
         cs.caseTitle = os.path.splitext(os.path.basename(self.fileName))[0]
-        cs.loadFromString(self.h5db["inputs/settings"][()])
+        try:
+            cs.loadFromString(self.h5db["inputs/settings"][()])
+        except:
```
ntouran, line 678: make pylint broad exception exception and explain when it's expected and why it's ok

armi/bookkeeping/db/database3.py
```diff
@@ -673,13 +673,23 @@ def loadCS(self):
...
     def loadBlueprints(self):
         from armi.reactor import blueprints
 
-        bpString = self.h5db["inputs/blueprints"][()]
+        bpString = None
+
+        try:
+            bpString = self.h5db["inputs/blueprints"][()]
+        except:
+            pass
```
ntouran, line 691: add explanation and pylint disable

armi/reactor/tests/test_composites.py
```diff
@@ -422,19 +436,36 @@ def test_flagSerialization(self):
...
-            )
+    def test_flagConversion(self):
+        data = [
+            self.TestFlagsA.A,
+            self.TestFlagsA.A | self.TestFlagsA.C,
+            self.TestFlagsA.A | self.TestFlagsA.C | self.TestFlagsA.D,
+        ]
+
+        for f in data:
+            print(int(f))
```
ntouran, line 454: still want this in here?

armi/reactor/tests/test_composites.py
```diff
@@ -422,19 +436,36 @@ def test_flagSerialization(self):
...
+
+        serialized, attrs = composites.FlagSerializer._packImpl(data, self.TestFlagsA)
+
+        data2 = composites.FlagSerializer._unpackImpl(
+            serialized, composites.FlagSerializer.version, attrs, self.TestFlagsB
+        )
+        for f in data2:
+            print(int(f))
+
```
ntouran, line 463: remove?

armi/reactor/tests/test_composites.py
```diff
@@ -422,19 +436,36 @@ def test_flagSerialization(self):
...
+        expected = [
+            self.TestFlagsB.A,
+            self.TestFlagsB.A | self.TestFlagsB.C,
+            self.TestFlagsB.A | self.TestFlagsB.C | self.TestFlagsB.D,
+        ]
```
ntouran, line 468: should be an assert after this checking that result matches

ntouran (changes requested):
* Please add some narrative docs somewhere about how exactly the flag conversion is done.
* Consider updating release notes summarizing the change

armi/bookkeeping/db/database3.py
youngmit: Turns out it should be a KeyError and i was just lazy. Added with comment

armi/bookkeeping/db/database3.py
youngmit: same as above

ntouran (approved):
sweeet

PR #202, opened 2020-11-09
Fix flag string conversions for multi-word combos

youngmit:
Some single-word conversions were taking place before matching multi-word flags (specifically NOZZLE in INNER_NOZZLE leaving INNER_ behind). This adds word-boundary regex to all of the strings to avoid consuming parts of words.


ntouran (approved):
would love a previously failing unit test for this too.

youngmit:
Will make a test!

PR #204, opened 2020-11-10
Fix catch-all try/except in __main__

youngmit:
We were catching all exceptions, including the SystemExit exception
which is nominally raised by sys.exit(). This was leading to unexpected
stack traces.


ntouran (approved):

ntouran (changes requested):
Can you add this unit test as well?

```
import unittest

from armi.__main__ import main


class TestRun(unittest.TestCase):
    def test_runCommand(self):
        """Ensure run entry point with no args completes."""
        main()
```

youngmit:
Were you able to get this to work under pytest? I'm having a heck of a time

armi/tests/tutorials/c5g7-settings.locationHistory.txt
```diff
@@ -0,0 +1 @@
+Assem     0     1     2     3     4     5     6     7     8     9
```
ntouran, line 1: don't want this committed I don't think.

ntouran (changes requested):
I was only able to get it to fail in pytest. I did not try the new version with a success.

ntouran (approved):

PR #205, opened 2020-11-10
Fix ThetaRZGrid ring/pos bug and update ring/pos docs.

onufer:
Update doc strings for getLocatorFromRingAndPos.
When it is hex geometry 0-based indices are not expected.

### DIFF
--- a/armi/reactor/grids.py
+++ b/armi/reactor/grids.py
@@ -1018,6 +1018,15 @@ class Grid:
 class CartesianGrid(Grid):
     """
     Grid class representing a conformal Cartesian mesh.
+
+    Notes
+    -----
+    In Cartesian, (i, j, k) indices map to (x, y, z) coordinates.
+    In an axial plane (i, j) are as follows::
+
+        (-1, 1) (0, 1) (1, 1)
+        (-1, 0) (0, 0) (1, 0)
+        (-1,-1) (0,-1) (1,-1)
     """
 
     @classmethod
@@ -1076,12 +1085,12 @@ class CartesianGrid(Grid):
         Notes
         -----
         This is needed to support GUI, but should not often be used.
-        x, y (0-based) indices are much more useful. For example:
+        i, j (0-based) indices are much more useful. For example:
 
-        >>> indices = core.spatialGrid[x, y, 0] # 3rd index is 0 for assembly
+        >>> indices = core.spatialGrid[i, j, 0] # 3rd index is 0 for assembly
         >>> a = core.childrenByLocator[indices]
 
-        >>> a = core.childrenByLocator[core.spatialGrid[x, y, 0] # one liner
+        >>> a = core.childrenByLocator[core.spatialGrid[i, j, 0]] # one liner
         """
         i, j = indices[0:2]
         split = self._isThroughCenter()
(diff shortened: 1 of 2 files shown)


armi/reactor/grids.py
```diff
@@ -923,9 +923,9 @@ def getLocatorFromRingAndPos(self, ring, pos, k=0):
         Parameters
         ----------
         ring : int
-            Ring index
+            Ring index (or number starting at 1 for hex)
```
youngmit, line 926: Ring and position are __always__ 1-based

youngmit: ...and if this is inconsistent in any geometry, then they are wrong

onufer: rzTheta seems to be wrong then:
```
    def getRingPos(self, indices):
        return (indices[1] + 1, indices[0] + 1)

    @staticmethod
    def getIndicesFromRingAndPos(ring, pos):
        return (pos, ring)
```

We are coding in python, so non-zero based indexing is surprising. Also the argument k is clearly zero based indexing, so I thought it would be helpful to point out that this is 1 based.

It is unclear from your feedback what actions you would like me to take. Can you give me some actionable feedback, or let me know if you are okay with these changes?

onufer: Not really sure what you wanted, so I guessed. Let me know if this satisfies your concerns.

youngmit: rzTheta is wrong. 100% agree that ring and position being 1-based is surprising, but it __is__ 1-based. The greater sin is to be inconsistent within ring/pos. Could you fix rzTheta?

armi/reactor/grids.py
```diff
@@ -1610,7 +1610,7 @@ def getRingPos(self, indices):
 
     @staticmethod
     def getIndicesFromRingAndPos(ring, pos):
-        return (pos, ring)
+        return (pos - 1, ring - 1)
```
youngmit, line 1613: Great, awesome! Thanks!

youngmit (approved):

armi/reactor/grids.py
onufer: I thought I did.

youngmit (changes requested):

armi/reactor/grids.py
```diff
@@ -1072,6 +1081,16 @@ def getRingPos(self, indices):
 
         Grids that split the central locations have 1 location in in inner-most ring,
         whereas grids without split central locations will have 4.
+
+        Notes
+        -----
+        This is needed to support GUI, but should not often be used.
+        i, j (0-based) indices are much more useful. For example:
+
+        >>> indices = core.spatialGrid[i, j, 0] # 3rd index is 0 for assembly
```
youngmit, line 1090: `locator` may be better than `indices`

armi/reactor/grids.py
```diff
@@ -1207,18 +1236,21 @@ def getSymmetricEquivalents(self, indices):
...
+    """
+    Has 6 neighbors in plane.
+
+    Notes
+    -----
+    In an axial plane (i, j) are as follows::
+
+               ( 0, 1) ( 1, 0)
+            (-1, 1) (0, 0) (1,-1)
+               (-1, 0) ( 0,-1)
```
youngmit, line 1252: dig this!

youngmit (approved):
consider changing `indices` -> `locator` in your example. Indices throughout grids.py mean a tuple of (i, j, k)

armi/reactor/grids.py
onufer: done. will push when build completes

PR #208, opened 2020-11-10
Make psutil dependency optional

youngmit:
MacOS users have trouble getting psutil to work, and it is only needed
to do memory profiling, which is not a necessity. This makes the psutil
package optional, and hobbles most of the MemoryProfiler functionality
if it isn't found. Removing the MemoryProfiler entirely in the absence
of psutil might be a better long-term goal, but it is rather intertwined
with reports and the standard operator and difficult to remove.

### DIFF
--- a/armi/bookkeeping/memoryProfiler.py
+++ b/armi/bookkeeping/memoryProfiler.py
@@ -42,10 +42,17 @@ http://packages.python.org/Pympler/index.html
 https://pythonhosted.org/psutil/
 https://docs.python.org/2/library/gc.html#gc.garbage
 """
-import logging
-import psutil
-import tabulate
 import gc
+import logging
+import tabulate
+from typing import Optional
+
+try:
+    import psutil
+
+    _havePsutil = True
+except ImportError:
+    _havePsutil = False
 
 import armi
 from armi import interfaces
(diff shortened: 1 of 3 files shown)


youngmit:
Do we think it's worth it to add test environments for testing without extra features?

armi/bookkeeping/memoryProfiler.py
```diff
@@ -42,10 +42,17 @@
...
 import tabulate
-import gc
+from typing import Optional
+
+try:
+    import psutil
+
+    _havePsutil = True
+except ImportError:
+    _havePsutil = False
```
ntouran, line 55: add docs explaining why this is this way to future code readers

armi/bookkeeping/memoryProfiler.py
```diff
@@ -502,26 +509,31 @@ def invokeHook(self):
 class SystemAndProcessMemoryUsage(object):
     def __init__(self):
         self.nodeName = armi.MPI_NODENAME
-        self.percentNodeRamUsed = psutil.virtual_memory().percent
-        self.processMemoryInMB = psutil.Process().memory_info().rss / (1012.0 ** 2)
+        self.percentNodeRamUsed: Optional[float] = None
+        self.processMemoryInMB: Optional[float] = None
+        if _havePsutil:
```
ntouran, line 514: else: 
warning: you don't have psutil so can't do anything here

ntouran (approved):
pls add comments explaining the situation.

PR #209, opened 2020-11-10
Update HexGrid class doc string

onufer:
* The hex grid class doc string now shows ASCII map for both orientations.


youngmit (approved):
awesome, thanks!

PR #213, opened 2020-11-13
Add permissive option to armi.configure()

youngmit:
This was added to make it easier for tests and other demonstration
examples to run back-to-back without having to restart the python
interpreter.


ntouran (approved):
thank you!

PR #216, opened 2020-11-14
Update README.rst

ZanderUF:
Link to ARMI gallery examples did not point to the documentation. Changed so it points to the hosted doc webpage.


ntouran (approved):
good call, thanks.

PR #221, opened 2020-11-17
Delegate requirements.txt packages to setup.py

youngmit:
This moves the lists of required packages out of requirements.txt and in
to setup.py. This reduces duplication while maintaining flexibility and
staying true to the spirit of where which package requirements should be
expressed. Now, requirements.txt will install the armi package in
"editable" mode, with all extras features enabled. requirements-testinng.txt
will install everything from requirements.txt, plus the `dev` extras.


ntouran (changes requested):
Looks good, and thanks for doing this. I would like to request some narrative description of the logic behind this method so that we can continue to understand the problems that led to this choice and the logic behind why this is the best one. This will help us moving forward as we continue to evolve. Please add to the Developer docs part of the manual in rst.

ntouran (approved):

PR #226, opened 2020-12-16
Programmatic inputs example

### LINKED ISSUE #225: Make example showing how to make blueprints objects in a script
To potentially help with #223, we should make a new example (in the gallery? in blueprints docs?) that shows how to build reactor inputs from a script by using Blueprints objects. This is likely a better option than writing out yaml code directly for script-generated ARMI inputs. 

For example, let's make a script that builds a few basic reactor parts and then either:
* Generates the input files
* Builds the ARMI objects

ntouran:
Closes #225


ntouran:
FYI here's what it ends up looking like in the gallery 
image

ntouran:
The build failure regarding XS settings is a new one and it's caused by a change in a dependency version, probably in voluptuous, which is treating things differently. I can replicate it on a fresh install of ARMI.

youngmit (changes requested):
Ever so minor request

youngmit:
> Yeah this commit added sorted which causes this because there is a `None` in the list getting sorted. [alecthomas/voluptuous@997c8db](https://github.com/alecthomas/voluptuous/commit/997c8dbba92fec3f58b081141c9a4a90e046ada8)

Good find!

youngmit (approved):

youngmit:
Sweet!

PR #229, opened 2020-12-18
Change geomType arguments to prefer GeomType enum

youngmit:
Most code that takes a `geomType` argument used to expect a string, which would
be tested against the string constants defined in the `geometry` module. This
was error-prone, inefficient, and ambiguous. This adds some extra conversion
functionality to the GeomType enumeration class and changes most interfaces that
pass `geomType` around to use instances of the GeomType enum instead. Some
functions will still attempt a conversion from string to aid in backwards
compatibility, but these should be reduced moving forward.

Additionally, to better-organize the code in the geometry module, the
SystemLayoutInput class has been migrated to its own module.


onufer:
can you add a test that confirms `core.geomType` is an enum rather than a string. I was having an issue with that about a month ago.

youngmit:
> 
> 
> can you add a test that confirms `core.geomType` is an enum rather than a string. I was having an issue with that about a month ago.

Done! Yeah your issues were the main reason I made this PR. ALSO, lucky test 1,000 (I think)! 🎉

ntouran (approved):

PR #232, opened 2021-01-12
Changed RIPL reader to correctly identify metastable states by considering the number of decay modes reported

ZanderUF:
When reading the half-life values from the RIPL database, an internal cutoff value for the smallest half-life acceptable was employed. This sometimes selected metastable states that are not well understood (in terms of their decay modes) and have very short half-lives. Furthermore these very short lived states are typically less important for most analysis, as they decay away so quickly it practically doesn't matter, for instance if you care about understanding the dose consequence from an inventory of isotopes. 

For example, there are two metastable states of Ag-110, with half-lives of either 6.6e-07 seconds or ~249 days. The 6.6e-7 state is not well understood and not terribly important for most analysis. Fortunately, the RIPL database gives a flag for each level that indicates the number of decay modes (see `nd` flag in the RIPL data). If `nd` is > 0 it indicates there are 1 or more decay modes OR the isotope decays by gamma emission. If `nd` is greater than 0 and has well understood decay pathways the lines after that level's entry (in the data file) describes the decay mode and probabilities of that decay. This is all described in the RIPL database files in greater detail. 

So now the RIPL reader in ARMI considers a lower cutoff value for the half-life and if there is at least one well defined decay mode for a metastable state. Updating the unit tests for Krypton, it now correctly handles some metastable states that were not being tested for previously.


youngmit (changes requested):
Could you add some comments describing the criteria that are actually applied? Just thinking about future selves wondering why we have all of these tests. Also, a description in the PR would be good so that we can have a complete commit message when this merges. Discuss what it was doing before, why that was a problem, and what it does now. Thanks!

youngmit (approved):

Issue #237, opened 2021-01-16
Add SHEM-based neutron energy group structures


sombrereau:
When using DRAGON to generate multigroup cross sections, one should use neutron energy group structures that snap to the SHEM361 energy structure because that's the "fine group structure" for open source nuclear data for DRAGON.

- There should be an energy spectrum that's analogous the ARMI33, but snapped to the SHEM361 rather than the ANL2081
- Having the SHEM361 structure will be good for modeling thermal spectrum reactors

jakehader:
Closed with #238

PR #238, opened 2021-01-18
Shem energy structure

sombrereau:
This pull adds in the SHEM361 neutron energy group structure, generalizes the function to make a group structure that snaps to the boundaries of the ANL2081 structure to an arbitrary fine group structure and uses that function to make a 33 group structure -- SHEM33_361 -- that is like ANL33, but snapped to the boundaries of SHEM361.

This addresses issue https://github.com/terrapower/armi/issues/237


armi/physics/neutronics/energyGroups.py
```diff
@@ -275,14 +275,391 @@ def getGroupStructureType(neutronEnergyBoundsInEv):
...
-    modifiedEnergyBounds = []
-    for energyBound in group_energy_bounds:
-        modifiedEnergyBounds.append(utils.findNearestValue(ufgEnergies, energyBound))
-    return modifiedEnergyBounds
+    return _create_multigroup_structures_on_finegroup_energies(
+        group_energy_bounds, ufgEnergies
+    )
+
 
+GROUP_STRUCTURE["SHEM361"] = [
```
ntouran, line 300: Can you just give a reference for SHEM361 in a comment here?

armi/physics/neutronics/energyGroups.py
```diff
@@ -275,14 +275,391 @@ def getGroupStructureType(neutronEnergyBoundsInEv):
 )
 
 # fmt: on
+def _create_multigroup_structures_on_finegroup_energies(
```
ntouran, line 278: interesting that we have camel_case functions in here. Anyway you're being consistent so we'll allow it!

ntouran (approved):
approve with inline comment about comments. Can you also add another comment where the methods are called that says that these do run at import time which will eventually be a factor in armi import/startup time and that we probably should consider calling them in a more structured way during the armi bootstrap process?

armi/physics/neutronics/energyGroups.py
sombrereau: @ntouran -- that's a good idea (its mildly tricky to find for free); here you go.

PR #240, opened 2021-01-19
Hide wx import in gui tests behind import_module()

youngmit:
This aviods import errors when pytest is importing/discovering tests and
wxpython (an optional dependency) is not installed.


youngmit:
@scottyak you might find this interesting. TIL about `test.support.import_module()`; super handy!

scottyak:
Nice find!

PR #244, opened 2021-01-26
Troubleshoot GUI test on Travis

scottyak:
We attempt to troubleshoot why the GUI test is failing on Travis by:
- Logging more information about the display and display server
- Take screenshots of the virtual screen, save them as a video, base64 encode, and output to stdout.

### DIFF
--- a/armi/utils/tests/test_gridGui.py
+++ b/armi/utils/tests/test_gridGui.py
@@ -71,26 +71,28 @@ print(f"DISPLAY: {os.environ['DISPLAY']}")
 loginctl = subprocess.Popen(
     ["loginctl", "list-sessions", "--no-legend"], stdout=subprocess.PIPE
 )
-cut = subprocess.Popen(
+cut_output = subprocess.check_output(
     ["cut", "--delimiter= ", "--field=1"],
-    stdin=loginctl.stdout,
-    stdout=subprocess.PIPE,
-)
-display_server_type = (
-    subprocess.check_output(
-        ["xargs", "loginctl", "show-session", "--property=Type", "--value"],
-        stdin=cut.stdout,
+    stdin=loginctl.stdout
+).decode("utf-8").strip("\n")
+if cut_output:
+    print(f"cut_output: {cut_output}")
+    display_server_type = (
+        subprocess.check_output(
+            ["loginctl", "show-session", "--property=Type"] + cut_output.split()
+        )
+        .decode("utf-8")
+        .strip("\n")
     )
-    .decode("utf-8")
-    .strip("\n")
-)
-print(f"display_server_type: {display_server_type}")  # is it "x11"?
+    print(f"display_server_type: {display_server_type}")  # is it "x11"?
+else:
+    print("!!!!No loginctl sessions!!!!")
 
-_SECONDS_PER_TICK = 0.02
+_SECONDS_PER_TICK = 0.05
+_TMP_DIR = '/tmp/armi'
+_TMP_DIR = '/home/travis/build/terrapower/armi'
 
 
-def _wait(num_ticks: int) -> None:
-    time.sleep(num_ticks * _SECONDS_PER_TICK)
 
 
 def _findPointInWindow(
@@ -149,7 +151,12 @@ class GuiTestCase(unittest.TestCase):
         self.imageList = []
         self.inputSimulator = wx.UIActionSimulator()
 
-    def takeScreenshot(self):
+    def wait(self, num_ticks: int) -> None:
+        for _ in range(num_ticks):
+            time.sleep(_SECONDS_PER_TICK)
+            self._takeScreenshot()
+
+    def _takeScreenshot(self):
         bmp = wx.Bitmap(width=1200, height=1050)
         memDC = wx.MemoryDC()
         memDC.SelectObject(bmp)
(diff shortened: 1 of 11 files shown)


scottyak:
Here's what the animation looks like on my local machine.

test_setNumRings

youngmit:
Wonder why it goes black when "Apply" is clicked...

scottyak:
Ahh, the screen didn't actually turn black... I made the animation loop, because otherwise one would have to refresh the window to replay the gif, which is annoying. I added a few black frames at the end so that the viewer knows when the loop ended. That also contained an animation I made with fewer frames to conserve space.

I later added a few still frames before the blacks frames at the end of the animation to make it less confusing -- it would look like this:

test_setNumRings

armi/utils/tests/test_gridGui.py
```diff
@@ -107,8 +149,50 @@ def initializeGui(self):
...
+        memDC = wx.MemoryDC()
+        memDC.SelectObject(bmp)
+        mousePos = wx.GetMouseState().GetPosition()
+        memDC.Blit(0, 0, 1200, 1050, wx.ClientDC(self.frame), 0, 0)
+        memDC.SetPen(wx.Pen(colour=wx.Colour(red=150, green=50, blue=50), width=5))
+        memDC.DrawCircle(x=mousePos.x, y=mousePos.y, radius=7)
+        memDC.SelectObject(wx.NullBitmap)
+        img = bmp.ConvertToImage()
+        self.imageList.append(img)
+
```
youngmit, line 171: Glorious

requirements-testing.txt
```diff
@@ -1,2 +1,5 @@
 numpy<1.19.4
+https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-16.04/wxPython-4.1.0-cp36-cp36m-linux_x86_64.whl
```
youngmit, line 2: This bends thie particular requirements file into a what-Travis-needs spec, which I think is fine. In the past, we have used it to set up local environments, so it was a bit more platform-independent. We recently did some cleanup on setup.py to add the `dev` and `memprof` extras, which made the requirements files pretty much vestigial, so I don't think anyone uses it to set up their own environment anymore. Might be nice to just add a comment to the top that says something like "this requirements file is mostly here to make Travis CI go"

setup.py
```diff
@@ -99,7 +99,14 @@ def collectExtraFiles():
...
     },
-    tests_require=["nbconvert", "jupyter_client", "ipykernel"],
+    tests_require=[
+        "nbconvert",
+        "jupyter_client",
+        "ipykernel",
+        "wxpython",
+        "pytest-xvfb",
+        "screeninfo",
+    ],
```
youngmit, line 109: I'm not sure if we still need `tests_require` at all, to be honest. According to the `setuptools` docs:

```
tests_require

    If your project’s tests need one or more additional packages besides those needed to install it,
you can use this option to specify them. It should be a string or list of strings specifying what
other distributions need to be present for the package’s tests to run. When you run the test
command, setuptools will attempt to obtain these (using pip if available). Note that these
required projects will not be installed on the system where the tests are run, but only
downloaded to the project’s setup directory if they’re not already installed locally.

    New in 41.5.0: Deprecated the test command.

```

Looks like it's only used by the `setup.py test` command, which has been deprecated anyways. It's unclear to me if anything else looks at this (e.g., pip)... I think the `requirements-testing.txt` might be sufficient to specify extra dependencies, without the duplication, but it does seem weird to not express them in the `setup.py`... Thoughts?

youngmit (commented):
Looks like you are still experimenting with tox/coverage, but some early thoughts.
This is looking rad

scottyak:
Here's some stuff I logged from the Travis run:

```
platform linux -- Python 3.6.7, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /home/travis/build/terrapower/armi/.tox/py36/bin/python

cachedir: .tox/py36/.pytest_cache

metadata: {'Python': '3.6.7', 'Platform': 'Linux-4.15.0-1077-gcp-x86_64-with-debian-stretch-sid', 'Packages': {'pytest': '6.2.2', 'py': '1.10.0', 'pluggy': '0.13.1'}, 'Plugins': {'metadata': '1.11.0', 'cov': '2.11.1', 'html': '3.1.1', 'xvfb': '2.0.0', 'forked': '1.3.0', 'xdist': '2.2.0'}}

rootdir: /home/travis/build/terrapower/armi, configfile: pytest.ini

plugins: metadata-1.11.0, cov-2.11.1, html-3.1.1, xvfb-2.0.0, forked-1.3.0, xdist-2.2.0

collecting ... Monitor(x=0, y=0, width=1200, height=1050, width_mm=0, height_mm=0, name='screen')

DISPLAY: :0

!!!!No loginctl sessions!!!!
```

Here's what I got from the base64 -> video conversion:

test_numRings_travis


It appears that the Travis environment doesn't have X11 set up.

armi/utils/tests/test_gridGui.py
scottyak: Thanks!

requirements-testing.txt
scottyak: Ok, I'll add a comment to the top of this file.

john-science:
Sorry, you'll have to forgive how late I am to the party here; I didn't join the project until months after this PR was opened.

It looks like this PR left off with the problem that there is no X11 on the Travis CI servers.  Was that the only outstanding issue?

Since then, the project has moved to GitHub Actions for testing.  Which has the same headless / no-X11 problem. But I wonder if the new testing ecosystem has more tools to solve the problem.  (Also, have we considered [XVFB](https://www.x.org/releases/X11R7.7/doc/man/man1/Xvfb.1.xhtml)?)

scottyak:
Yes, it does use xvfb, but the problem still persists in a weird way -- it
doesn't crash outright, but the grid UI just doesn't appear.

I was only able to figure that out by having the unit test take
screenshots, pack it into a gif, print the base64 encoding of the gif into
the Travis logs, then base64 decode it back into a gif. That was rather
painful, so if browsers would *just work*, that might be worth it.

scottyak:
Working on a notebook-based approach for rewriting the GUI. Stay tuned.

john-science:
I love this feature idea, but the PR was opened 18 months ago, so I'm going to close it.  For now.

If further work is being done towards these goals, we can obviously just re-open the PR.

Thanks for your interest!

PR #248, opened 2021-02-05
Move the isDepletable function off of the composite structure

ntouran:
This moves isDepletable off of the composite and into the framework neutronics plugin. This is part of the overall push to keep the composite structure cleaner. 

Additionally, a situation was discovered and fixed where if the flags were set explicitly, including a depletable flag, the burn chain nuclides were not expanded properly on the composite. This would have been an error condition to any user who attempted it before. 

The annular and depletable flags were explicitly set on a test input to cover situations with annular fuel that is also depletable.


youngmit (approved):
Looks good, thanks for adding more context

PR #249, opened 2021-02-05
Bump h5py version

youngmit:
h5py changed the manner in which they handle strings, requireing some
modifications to the database3 implementation. These were relatively
straightforward, but the new h5py required a newer version of numpy,
which unerthed some new warnings and issues in the tests. This addresses
all of those as well.

Of particular note is the addition of floating-point epsilon-aware
line-by-line file comparisons. This is needed, since newer numpy
versions produce slightly different results in certain scenarios, and we
want to make the tests a little less sensitive to minor differences
between numpy versions. The approach is to (in the event that string
comparison of two lines fails) split each line into words along
whitespace. For each pair of words, attempt parsing them as floats. If
they can both be parsed into floats, do a relative epsilon comparison.
Any trouble in this process and it will declare the strings non-equal.

### DIFF
--- a/armi/tests/__init__.py
+++ b/armi/tests/__init__.py
@@ -283,7 +283,7 @@ class ArmiTestHelper(unittest.TestCase):
         os.remove(actualFilePath)
 
     @staticmethod
-    def compareLines(actual: str, expected: str, eps: Optional[float]=None):
+    def compareLines(actual: str, expected: str, eps: Optional[float] = None):
         """
         Impl of line comparison for compareFilesLineByLine.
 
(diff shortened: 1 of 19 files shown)


armi/tests/__init__.py
```diff
@@ -247,36 +254,90 @@ def compareFilesLineByLine(
...
+            for lineIndex, (expectedLine, actualLine) in enumerate(
+                itertools.zip_longest(expected, actual)
+            ):
+                if expectedLine is None:
+                    raise AssertionError(
+                        "The test-generated file is longer than expected file"
+                    )
+                if actualLine is None:
+                    raise AssertionError(
+                        "The test-generated file is shorter than expected file"
```
ntouran, line 266: that's nicer! thanks.

ntouran (approved):
awesome

PR #253, opened 2021-02-23
Adding several unit tests to improve coverage.

mgjarrett:
I selected files currently in the top 20 of missed lines on coveralls and skimmed through them to determine which had low-hanging fruit for improving test coverage statistics. I added unit tests to the following files:

armi/utils/tests/test_utils.py
armi/reactor/tests/test_reactors.py
armi/reactor/tests/test_blocks.py
armi/reactor/tests/test_zones.py

I tended towards smaller bites: unit tests for small functions with only a couple of relevant lines that could be easily implemented giving the existing test infrastructure and example objects that were already set up for other tests. I considered writing tests for more complicated functions that would have likely required setting up new data (e.g., burnup/power/flux summary functions that require spatial distributions) but generally stayed away from those.

I fixed three small bugs that I discovered while writing the tests:

1) armi/reactor/blocks.py: In the class ThRZBlock, the thetaOuter() method returned the min() of the "outer_theta" dimensions, but it seems like it should return the max() based on the name and comments.
2) armi/reactor/zones.py: The Zones method _buildAssemTypeZones() is supposed to create and name zones based on the assembly type (i.e., "igniter fuel", "lta fuel", etc.) but it was instead using the assembly name attribute, derived from the location ("A0010", "A0016", etc.). As a result, the assemblies were not correctly grouped into zones based on type.
3) armi/reactor/converters/geometryConverters.py: In the class BlockNumberModifier, the convert() method calls SetNumberOfBlocks, and passed it a string blockType (e.g., "clad") where it expects a flag (e.g., "Flags.CLAD"). Fixed with a call to Flags.fromString().

P.S. This is my first pull request. I used black and tried to keep naming and style in line with the surrounding code, but I wouldn't be surprised if I missed some subtleties (or even some not-so-subtleties). The commits are pretty atomic, so I'm hoping it will be easy to go back and revise them as necessary.

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -2216,7 +2216,7 @@ class ThRZBlock(Block):
     def thetaOuter(self):
         """Return a largest theta of all the components."""
         outerTheta = self.getDimensions("outer_theta")
-        largestOuter = min(outerTheta) if outerTheta else None
+        largestOuter = max(outerTheta) if outerTheta else None
         return largestOuter
 
     def axialInner(self):
--- a/armi/reactor/converters/geometryConverters.py
+++ b/armi/reactor/converters/geometryConverters.py
@@ -159,7 +159,7 @@ class BlockNumberModifier(GeometryChanger):
                     self.setNumberOfBlocks(a)
                 else:
                     # non-fuel. Control?
-                    self.setNumberOfBlocks(a, blockType=a[fuelI].getType())
+                    self.setNumberOfBlocks(a, blockFlags=a[fuelI].p.flags)
             else:
                 # radial shields, etc. go here.
                 pass
@@ -173,7 +173,7 @@ class BlockNumberModifier(GeometryChanger):
         # update bookkeeping.
         r.core.regenAssemblyLists()
 
-    def setNumberOfBlocks(self, assem, blockType=Flags.FUEL):
+    def setNumberOfBlocks(self, assem, blockFlags=Flags.FUEL):
         r"""
         Change the region to have a certain number of blocks with uniform height
 
@@ -184,9 +184,9 @@ class BlockNumberModifier(GeometryChanger):
         assem : Assembly
             The assembly to modify
 
-        blockType : str, optional
-            Type of block to change. Default: Fuel. Allows control
-            assemblies, etc. to modified just like fuel assemblies.
+        blockFlags : Flags, optional
+            Type of block to change. Default: Flags.FUEL. Allows control
+            assemblies, etc. to be modified just like fuel assemblies.
 
         Notes
         -----
@@ -198,9 +198,9 @@ class BlockNumberModifier(GeometryChanger):
         if you're tracking history.
 
         """
-        fuelHeight = assem.getTotalHeight(blockType)
+        fuelHeight = assem.getTotalHeight(blockFlags)
         blockHeight = fuelHeight / self.numToAdd
-        fuelBlocks = set(assem.getBlocks(blockType))
+        fuelBlocks = set(assem.getBlocks(blockFlags))
         newBlockStack = []
         numFuelBlocksAdded = 0
         # make a tracker flag that tells us if we're below or above fuel.
--- a/armi/reactor/tests/test_blocks.py
+++ b/armi/reactor/tests/test_blocks.py
@@ -1566,6 +1566,9 @@ class HexBlock_TestCase(unittest.TestCase):
         self.assertEqual(v0 / 3.0, self.HexBlock.getVolume())
         self.assertAlmostEqual(m0 / 3.0, self.HexBlock.getMass())
 
+        symmetryLine = self.HexBlock.isOnWhichSymmetryLine()
+        self.assertEqual(grids.BOUNDARY_CENTER, symmetryLine)
+
     def test_retainState(self):
         """Ensure retainState restores params and spatialGrids."""
         self.HexBlock.spatialGrid = grids.HexGrid.fromPitch(1.0)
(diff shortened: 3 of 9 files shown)


armi/utils/tests/test_utils.py
```diff
@@ -36,6 +36,9 @@ def test_parabola(self):
         self.assertEqual(b, 0.0)
         self.assertEqual(c, 1.0)
 
+        with self.assertRaises(Exception):
```
ntouran, line 41: probably should get more specific than `Exception`. I get a `LinAlgError: Singular matrix`, so would `assertRaises(numpy.linalg.LinAlgError)` be better? Meh, maybe not. We're really just trying to make sure it fails and probably shouldn't be too sensitive to the internals of numpy. Ok you can leave it.

armi/utils/tests/test_utils.py
```diff
@@ -202,6 +202,14 @@ def test_slantSplit(self):
         self.assertListEqual(x1, [1.0, 2.0, 3.0, 4.0])
         self.assertListEqual(x2, [4.0, 3.0, 2.0, 1.0])
 
+    def test_newtonsMethod(self):
```
ntouran, line 216: Good call on leaving the cs out as we are hoping to get rid of most things that need some kind of global CS in our roadmap.

armi/utils/tests/test_utils.py
```diff
@@ -218,6 +220,12 @@ def test_minimizeScalarFunc(self):
         )
         self.assertAlmostEqual(minimum, 0.0, places=3)
 
+    def test_prependToList(self):
+        a = ["Foo", "Bar"]
+        b = [1, 2, 3]
+        utils.prependToList(a, b)
+        self.assertTrue(a == [1, 2, 3, "Foo", "Bar"])
```
ntouran: Does `self.assertEqual(a, [1, 2, 3, "Foo", "Bar"])` work here?

armi/utils/tests/test_utils.py
```diff
@@ -232,22 +232,21 @@ def test_prependToList(self):
         a = ["Foo", "Bar"]
         b = [1, 2, 3]
         utils.prependToList(a, b)
-        self.assertTrue(a == [1, 2, 3, "Foo", "Bar"])
+        self.assertListEqual(a, [1, 2, 3, "Foo", "Bar"])
```
ntouran, line 236: Ah that answers my previous question.

armi/utils/tests/test_utils.py
```diff
@@ -17,6 +17,7 @@
 """
 import unittest
 import math
+import os
 import numpy as np
```
ntouran, line 21: should leave the single blank line between python built in imports and third party ones.

armi/utils/tests/test_utils.py
```diff
@@ -250,6 +251,20 @@ def test_convertToSlice(self):
         with self.assertRaises(Exception):
             slice1 = utils.convertToSlice("slice")
 
+    def test_plotMatrix(self):
+        fname = "F"
```
ntouran, line 255: I recommend using a more unique filename here. We run into trouble in parallel with os.remove locks sometimes in the off chance some other tests makes a file called that. Maybe `test_plotMatrix_testfile.png`?

armi/utils/tests/test_utils.py
```diff
@@ -250,6 +251,20 @@ def test_convertToSlice(self):
...
+        matrix[0,0] = 1
+        matrix[0,1] = 2
+        matrix[1,0] = 3
+        matrix[1,1] = 4
+        utils.plotMatrix(matrix, fname, show=True, title="plot")
+        utils.plotMatrix(matrix, fname, minV=0, maxV=5, figsize=[3, 4])
+        xtick = ([0, 1],["1", "2"])
+        ytick = ([0, 1],["1", "2"])
+        utils.plotMatrix(matrix, fname, xticks=xtick, yticks=ytick)
+        os.remove("F.png")
```
ntouran, line 266: consider `f"{fname}.png"` here

armi/reactor/tests/test_reactors.py
```diff
@@ -537,6 +545,21 @@ def test_removeAssembly(self):
...
 
+    def test_removeAssembliesInRing(self):
+        aLoc = [
+            self.r.core.spatialGrid.getLocatorFromRingAndPos(3, i + 1)
+            for i in range(12)
+        ]
+        assems = {
+            i: self.r.core.childrenByLocator[loc]
+            for i, loc in enumerate(aLoc)
+            if loc in self.r.core.childrenByLocator.keys()
```
ntouran, line 698: .keys() should be optional there.

armi/reactor/tests/test_reactors.py
```diff
@@ -334,6 +334,51 @@ def test_findAllAxialMeshPoints(self):
...
+            21.7588882701,
+            36.2648137835,
+            50.7707392969,
+            65.2766648103,
+            79.7825903236,
+            94.288515837,
+            108.7944413504,
+            123.3003668638,
+        ]
+        self.assertListEqual(expectedPoints, aziPoints)
```
ntouran, line 358: We sometimes use things like `np.testing.assert_almost_equal` for list comparisons of floats, which may twiddle sometimes. 

There's also some ongoing concern about how to handle the maintenance burden of tests like this that change when the test fixtures change. We have lots like this so you're doing as we always do, but it's just something we worry about. We'll probably eventually need to decouple more tests from the "one test reactor" loaded in this case, even though we thought it was convenient at first. Just a fyi.

armi/reactor/tests/test_blocks.py
```diff
@@ -1339,6 +1339,47 @@ def test_getMfp(self):
         assert_allclose(235.0, mfpAbs, rtol=0.1)
         assert_allclose(17.0, diffusionLength, rtol=0.1)
 
+    def test_getMgFlux(self):
+        applyDummyData(self.Block)
+        # These are unverified numbers, just the result of this calculation.
+        flux = self.Block.getMgFlux(average=True)
+        referenceFlux = [
+            3.68362183e07,
+            5.21203125e08,
```
ntouran, line 1348: can these numbers be derived algorithmically from the dummy data loaded up so we don't have to manually rebuild this if that changes? Or at that point are we just re-implementing the function being tested?

armi/reactor/tests/test_blocks.py
```diff
@@ -1607,6 +1607,9 @@ def testSymmetryFactor(self):
         self.assertEqual(v0 / 3.0, self.HexBlock.getVolume())
         self.assertAlmostEqual(m0 / 3.0, self.HexBlock.getMass())
 
+        symmetryLine = self.HexBlock.isOnWhichSymmetryLine()
+        self.assertEqual(4, symmetryLine)
```
ntouran, line 1611: use `grids.BOUNDARY_CENTER` here rather than `4`.

armi/reactor/tests/test_blocks.py
```diff
@@ -1708,6 +1708,10 @@ def test_getDuctPitch(self):
         ductOP = self.HexBlock.getDuctOP()
         self.assertAlmostEqual(70.6, ductOP)
 
+    def test_getPinCenterFlatToFlat(self):
+        f2f = self.HexBlock.getPinCenterFlatToFlat()
+        self.assertAlmostEqual(1.3336791218280355, f2f)
```
ntouran, line 1713: any option to build this from the input data? Again we've just been bitten a lot when the test fixtures change and we have like N-hundred unit tests to go update it can be a pain.

armi/reactor/blocks.py
```diff
@@ -2216,7 +2216,7 @@ def thetaInner(self):
     def thetaOuter(self):
         """Return a largest theta of all the components."""
         outerTheta = self.getDimensions("outer_theta")
-        largestOuter = min(outerTheta) if outerTheta else None
+        largestOuter = max(outerTheta) if outerTheta else None
```
ntouran, line 2219: Nice catch!!

armi/reactor/zones.py
```diff
@@ -599,7 +599,7 @@ def _buildAssemTypeZones(core, cs, typeSpec=None):
         try:
             zone = zones[zoneName]
         except KeyError:
-            zone = Zone(a.name)
+            zone = Zone(zoneName)
```
ntouran, line 602: looks right to me. another great catch. thanks!

ntouran:
Thanks for all these! Almost a full % increase in coverage! That's the best we've had in a long time. And great bug finds too. 

I went through and made a few small comments and observations. I think @youngmit will want to go through some of these as well. 

One thing about the commit messages is that they should all be in the imperative mood, so "Add a unit test for getAllXsSuffixes." instead of "Adding a unit test for getAllXsSuffixes." This is just a style/consistency thing apparently [traceable back to git man pages. ](https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?id=HEAD#n133). Not a huge deal. 

We'll also have to ask if we want to squash these or not. I kind of like them atomic as you have them TBH but will discuss with @youngmit. Anyway great job.

armi/utils/tests/test_utils.py
mgjarrett: I tried to clean up the commit history after the fact but it looks like I missed this.

armi/reactor/tests/test_blocks.py
mgjarrett: Probably? I'll look into it. I didn't dig too deep here, but it looks like it's just multiplying the flux and volume provided in the test.

armi/reactor/tests/test_blocks.py
mgjarrett: To be honest, I wasn't exactly sure what this is calculating, so I just copied the output it produced (fingers crossed that it's already implemented correctly). I'm not sure what "flat-to-flat" means in reference to pins. Is that just the pin pitch? I'll take a look at the numbers and see where this is coming from.

mgjarrett:
I think committing in bits and pieces is good for reviewing, but once they're in the repo it seems unlikely we would ever want to go back to an intermediate state somewhere in the middle of this, so squashing might be the way to go. I'll add a single commit addressing all of your comments.

armi/utils/tests/test_utils.py
mgjarrett: I didn't want to modify the actual code, which just has a "raise" statement with no specific exception/error type given.

armi/reactor/tests/test_blocks.py
mgjarrett: This test was actually not covering the if block that I wanted it to, so I'm going to delete it. We would need a block.p.lastMgFlux defined in order to test the part of getMgFlux() that I was targeting.

armi/reactor/tests/test_blocks.py
mgjarrett: I changed this to calculate from the input data stored on self.HexBlock. It looks a lot like the function it's testing, but still better for the future if the test fixture changes.

armi/utils/tests/test_utils.py
youngmit: Other option is to do the whole thing in a TemporaryDirectoryChanger, then we dont need to worry about removing anything.

youngmit (changes requested):
This all mostly looks great! Just a few things here and there.

youngmit:
> 
> 
> I think committing in bits and pieces is good for reviewing, but once they're in the repo it seems unlikely we would ever want to go back to an intermediate state somewhere in the middle of this, so squashing might be the way to go. I'll add a single commit addressing all of your comments.

Wholeheartedly agree. We can squash on the way in

armi/reactor/converters/geometryConverters.py
```diff
@@ -159,7 +159,8 @@ def convert(self, r=None):
                     self.setNumberOfBlocks(a)
                 else:
                     # non-fuel. Control?
-                    self.setNumberOfBlocks(a, blockType=a[fuelI].getType())
+                    blockType = a[fuelI].getType()
+                    self.setNumberOfBlocks(a, blockType=Flags.fromString(blockType))
```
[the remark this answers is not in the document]
mgjarrett: I made both changes: using `b.p.flags`, and changing the kwarg name to `blockFlags`.

We might want to take a look at what `setNumberOfBlocks` is doing with it. We're passing the flags from the block in the assembly at the index `fuelI`, which is the index of the first fuel block in a reference fuel assembly. `setNumberOfBlocks` then determines the "active" height that needs to be re-meshed:

`fuelHeight = assem.getTotalHeight(blockFlags)`

For a non-fuel assembly, we probably want `blockFlags `to be something like `Flags.CONTROL` or `Flags.SHIELD`. But is it safe to assume `a[fuelI].p.flags` would give us `Flags.CONTROL` and not `Flags.REFLECTOR` or something else "inert"? We're effectively assuming all assemblies are identical up to the beginning of the fuel, right?

armi/utils/tests/test_utils.py
```diff
@@ -161,6 +192,84 @@ def test_list2str(self):
...
+        matrix[1, 1] = 4
+        xtick = ([0, 1], ["1", "2"])
+        ytick = ([0, 1], ["1", "2"])
+        dc = directoryChangers.TemporaryDirectoryChanger()
+        dc.__enter__()
+        fname = "test_plotMatrix_testfile"
+        utils.plotMatrix(matrix, fname, show=True, title="plot")
+        utils.plotMatrix(matrix, fname, minV=0, maxV=5, figsize=[3, 4])
+        utils.plotMatrix(matrix, fname, xticks=xtick, yticks=ytick)
+        dc.__exit__(None, None, None)
```
youngmit, line 267: is this where that os.remove() used to be? the DirectoryChangers are actually [context managers](https://docs.python.org/3/reference/datamodel.html#context-managers), meaning that you can enter them with a `with ...:` statement. Like:
```
with directoryChangers.TemporaryDirectoryChanger(): # __enter__() called implicitly here
    # do stuff
# __exit__() implicitly called here
```

youngmit (changes requested):
One last thing: Use the directory changer as a context manager block

armi/utils/tests/test_utils.py
mgjarrett: got it. my initial attempt at this type of usage didn't work how I wanted it, I think because I was using the wrong method. I copied that usage from a setUp/tearDown in a unit test, where enter/exit are called explicitly because they're in different functions. I'll fix this.

youngmit (approved):

ntouran:
horray that got us rounding up to 80% and therefore out of the red!  THANKS!!
image

PR #255, opened 2021-02-23
Fix incorrect mult for non-grid block components

### LINKED ISSUE #254: Non-grid components have multiplicity of zero
Using the inputs from #251, even if I force the multiplicity to be one with
```yaml
blocks:
    fuel: &block_fuel
        grid name: pins
        duct: &comp_duct
            shape: Hexagon
            material: HT9
            Tinput: 600
            Thot: 600
            ip: 32
            op: 32.2
            mult: 1
```
the duct still has no multiplicity
```python
import armi
armi.configure()
o = armi.init(fName="minicore.yaml")
f = o.r.core.getFirstBlock(Flags.FUEL)
d = f.getComponentByName("duct")
d.getDimension("op")
32.2
d.getDimension("ip")
32.0
d.getDimension("mult")
0
d.getArea()
0.0
```
per https://github.com/terrapower/armi/blob/9e7c2984eae29ad0bb9752f9769cc17a7d12fec0/armi/reactor/components/basicShapes.py#L133-L138

This leads to problems when ARMI is computing volume fractions and can produce divide by zero errors in 
https://github.com/terrapower/armi/blob/9e7c2984eae29ad0bb9752f9769cc17a7d12fec0/armi/reactor/composites.py#L1544

ntouran:
All components in blueprints-defined blocks that had grids assigned but
were not themselves in the grid got a multiplicity of zero, which made
the volume fractions (and therefore everything else) wrong. This fix
detects when a component is not represented in the grid and just leaves
the user-input multiplicity. A unit test covering the situation was also
added.

Fixes #254.

### DIFF
--- a/armi/reactor/blueprints/blockBlueprint.py
+++ b/armi/reactor/blueprints/blockBlueprint.py
@@ -111,19 +111,25 @@ class BlockBlueprint(yamlize.KeyedList):
             c = componentDesign.construct(blueprint, materialInput)
             components[c.name] = c
             if spatialGrid:
-                c.spatialLocator = gridDesign.getMultiLocator(
+                componentLocators = gridDesign.getMultiLocator(
                     spatialGrid, componentDesign.latticeIDs
                 )
-                mult = c.getDimension("mult")
-                if mult and mult != 1.0 and mult != len(c.spatialLocator):
-                    raise ValueError(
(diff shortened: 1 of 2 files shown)


youngmit (approved):
Looks great; good catch!

Issue #257, opened 2021-02-24
python segfaults unless h5py is imported with armi


drewejohnson:
In the process of upgrading ARMI (thanks to all your fantastic PRs 😁) I found myself in a very strange predicament. This doesn't seem to impact any CLI calls like `armi run` which is great. Probably because `h5py` gets picked up when interacting with database files. But I felt like elevating this would hopefully help other people

Using Python 3.7.10 installed in a fresh conda environment, I keep getting segfaults when python (or ipython) exit. I can still work in an ipython or python environment, but the segfault appears at exit, possibly due to https://github.com/terrapower/armi/blob/259c91e200dc0f2a8ddf27258478d96de6412f78/armi/__init__.py#L344-L345 ?

Using `zsh 5.8 (x86_64-apple-darwin20.0)`
```zsh
% python -c "import armi"
zsh: segmentation fault  python -c "import armi"
```
and 
```bash
bash3.2$ bash --version

GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin20)
Copyright (C) 2007 Free Software Foundation, Inc.
bash-3.2$ python -c "import armi"
Segmentation fault: 11
```
```zsh
% python
>>> import armi
>>> print("hello world")
hello world
>>> exit()
zsh: segmentation fault  python
```

`python -c "import h5py; import armi"` and the reverse `python -c import armi; import h5py` work, discovered after digging through #249 https://github.com/terrapower/armi/blob/f702a89174cbe53a2a39343e680fd02ce65b9c60/armi/conftest.py#L38-L42 

## Steps

- Checkout f702a89 or later
- Install new h5py>3 with pip or conda: `conda install numpy\<1.19.4 h5py\>3`
- Upgrade ARMI
- Navigate to a new directory (not really necessary but forces python to use installed armi vs. local)
- `python -c "import armi"`

youngmit:
Thanks for the report! I've actually been looking into this myself. The conftest import was there because i knew it was causing issues with out tests, and i sort of suspected pytest, but turns out it was causing issues everywhere, and on windows it's hard to tell unless you look at `%errorlevel%`.

In my previous explorations into this, I've turned up a handful of other GH issues and the like in other projects where certain versions of numpy and others seem to goof some reference counting, leading to what I imagine are double-frees when the python interpreter cleans up. I am just picking the thread back up on this, so will post things that look relevant as I go along.

[...]

PR #258, opened 2021-02-24
Add early import of h5py in __init__

youngmit:
This is needed to avoid the possibility of the first import of `h5py`
occurring in the cleanup function that gets called by `atexit`.
Importing `h5py` from that context leads to segfaults.

### DIFF
--- a/armi/__init__.py
+++ b/armi/__init__.py
@@ -87,26 +87,6 @@ from armi.reactor import flags
 from armi.reactor import parameters
 from armi.nucDirectory import nuclideBases
 
-# h5py needs to be imported here, so that the disconnectAllHdfDBs() call that gets bound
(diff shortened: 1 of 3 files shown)


youngmit:
Fixes #257

armi/__init__.py
```diff
@@ -87,6 +87,25 @@
...
+# imported in this process, doing so will cause a segfault. The theory here is that this
+# happens because the h5py extension module is not safe to import (for whatever reason)
+# when the python interpreter is in whatever state it's in when the atexit callbacks are
+# being invoked.  Importing early avoids this.
+#
+# Minimal code to reproduce the issue:
+#
+# >>> import atexit
+#
+# >>> def boom():
```
ntouran, line 103: want to make this `willSegFault` ?

ntouran (approved):

PR #264, opened 2021-03-06
Complete removal of locations.py

jakehader:
This eliminates `locations.py` by adding a public `getPositionsInRing` and `getMinimumRings` methods to the HexGrid and CartesianGrid classes. This change now requires that blocks being converted to cylindrical equivalents within the block converters classes have the spatial grids defined.   

This also eliminates the special methods within `HexAssembly` and `CartesianAssembly` making them no different
than the base `Assembly` class. Maybe in the future we should enforce a "spatialGrid" type for these special shaped 
assemblies. If not, there is not much value in keeping these around.


armi/reactor/grids.py
```diff
@@ -1154,28 +1159,30 @@ def getMinimumRings(self, n):
         """
         Return the minimum number of rings needed to fit ``n`` objects.
         """
-        return self._getMinimumRingsStatic(n, throughCenter=self._isThroughCenter())
-
-    @staticmethod
-    def _getMinimumRingsStatic(n, throughCenter):
-        # Keep private because this is really here to support old locations stuff that
-        # hasn't been removed yet. See module docs for locations.py for more details.
```
youngmit, line 1162: We still havent solved the problem that requires this to be here. Again, the Blocks don't necessarily have a grid. If they did have a grid, at least it would let us express the appropriate even/odd through center behavior (the locations.py logic that the geometry converters was using just assumes `True` 😲). One way to get rid of this would be to address some of the stuff that I talk about up there.

armi/reactor/converters/blockConverters.py
```diff
@@ -497,8 +510,8 @@ def convert(self):
             )
         )
         self._dissolveComponents()
-        numRings = locations.minimumRings(
-            self._sourceBlock, self._sourceBlock.getNumPins()
+        numRings = self._sourceBlock.spatialGrid.getMinimumRings(
```
youngmit, line 513: We cannot assume that a block has a spatialGrid in the first place (you set up the test blocks below so that they do, but in general, Blocks dont need to have one). Furthermore, there isn't really any guarantee if a Block _does_ have one, that it will be a type of grid that corresponds in any way to the type of the block. Hex blocks with some sort of ThR grid is something that I've seen in the wild before.

That's not to say the the existing implementation here is handling any of this nuance properly anyhow; it's just stuffing all of those assumptions into the remnants of locations.py and holding on for dear life. I just see reaching for a spatial grid that might not be there, and which this converter is really expecting to correspond to the geometry of the block itself opens us up to lots of new surprises/bugs down the line.

I think that maybe we should just be more honest here in the block converters about their assumptions and limitations and confirm that the input blocks satisfy those assumptions. That might look something like:
 - move the type-to-geom mapping logic out of locations.py and into here; I think this is the only place that's really using it anyway so might as well own it.
 - check if things do have a spatialGrid, and if so, make sure that it matches the geom type of the block. otherwise these converters will be doing something wrong.
 - for the cartesian case (do we really exercise this??), perhaps _require_ that the block has a spatialGrid, so that we can stop assuming that `throughCenter == True`.

youngmit (changes requested):
There is some great progress in here. However, I don't think it's safe for the geom converters to assume that a block has a spatial grid in the first place. If it does, there are further constraints about _which_ grids the conversion algorithms support. We should probably just check this stuff explicitly and interpolate reasonable assumptions when possible (e.g., a HexBlock with no spatialGrid is assumed to have a HexGrid with such-and-such number of rings. CartesianBlock.... 🤷‍♀️)

jakehader:
Thanks for the comments @youngmit. Progress has been slow but I'll look back at your feedback and try to get to it soon.

jakehader:
Hi @youngmit, I've gone ahead and limited the scope of the block converters and documented their assumptions with using the spatial grid. I think that this meets the intent of your comments and will allow us to fully remove `locations.py` without changing the actual (or intended) behavior of the existing block converters. I think the issue is that they seemed to be more general than they actually are.

youngmit (approved):

PR #265, opened 2021-03-08
Force pinPowers parameters to be numpy arrays.

mgjarrett:
setPinPowers() in blocks.py was setting pinPowers as python lists.
It was ambiguous whether these parameters would by python lists 
or numpy arrays, which could cause confusion in other parts of the 
code. Add a setter so that the type is reliably known. As a bonus, 
floating point operations with numpy arrays will likely be faster 
than with basic python lists.

### DIFF
--- a/armi/physics/neutronics/parameters.py
+++ b/armi/physics/neutronics/parameters.py
@@ -197,18 +197,42 @@ def _getNeutronicsBlockParams():
             default=None,
         )
 
+        def pinPowersNeutron(self, value):
+            self._p_pinPowersNeutron = (
+                value
+                if value is None or isinstance(value, numpy.ndarray)
+                else numpy.array(value)
+            )
+
         pb.defParam(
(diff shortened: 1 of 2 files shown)


armi/physics/neutronics/parameters.py
```diff
@@ -197,18 +197,42 @@ def mgFlux(self, value):
...
+
         pb.defParam(
             "pinPowersGamma",
+            setter=pinPowersGamma,
             units="W/cm",
-            description="should be a blank 2-D array, but re-defined later (nPins x nAxialSegments)",
+            description="""
+                The block-level pin linear power densities from gamma heating.
+                pinPowersGamma[i] represents the average linear power density of pin i (from gamma
+                heating). Power units are Watts/cm (Watts produced per cm of pin length).",
```
youngmit, line 234: what a saint

youngmit (approved):
Looks great!

PR #266, opened 2021-03-08
Add unit tests for Material.

mgjarrett:
Needed more test coverage for material.py. Identified one issue, with
the implementation of isBeyondIncubationDose and
updateDeltaDPApastIncubation. The logic tries to evaluate the result
of self.modelConst["Rincu"] as a boolean. If "Rincu" is not in modelConst, 
it will not return False, but rather throw a KeyError. First, we need to check
whether it is in the dict, then we can proceed.

### DIFF
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -585,6 +585,7 @@ class Material(composites.Leaf):
         if not "Rincu" in self.modelConst:
             msg = "Material missing incubation dose"
             runLog.warning(msg, single=True, label="Missing incubation dose")
+            return False
         else:
             if totalDPA > self.modelConst["Rincu"]:
                 return True
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -582,13 +582,14 @@ class Material(composites.Leaf):
 
         """
 
-        if not self.modelConst["Rincu"]:
+        if not "Rincu" in self.modelConst:
             msg = "Material missing incubation dose"
             runLog.warning(msg, single=True, label="Missing incubation dose")
-        elif totalDPA > self.modelConst["Rincu"]:
-            return True
         else:
-            return False
+            if totalDPA > self.modelConst["Rincu"]:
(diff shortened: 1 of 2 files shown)


armi/materials/material.py
```diff
@@ -582,13 +582,14 @@ def isBeyondIncubationDose(self, totalDPA):
...
             msg = "Material missing incubation dose"
             runLog.warning(msg, single=True, label="Missing incubation dose")
-        elif totalDPA > self.modelConst["Rincu"]:
-            return True
         else:
-            return False
+            if totalDPA > self.modelConst["Rincu"]:
+                return True
+            else:
+                return False
```
youngmit, line 592: Simply return the result of the comparison: `return totalDPA > self.modelConst["Rincu"]`

armi/materials/material.py
```diff
@@ -582,13 +582,14 @@ def isBeyondIncubationDose(self, totalDPA):
 
         """
 
-        if not self.modelConst["Rincu"]:
+        if not "Rincu" in self.modelConst:
             msg = "Material missing incubation dose"
             runLog.warning(msg, single=True, label="Missing incubation dose")
```
youngmit, line 587: This will implicitly return `None`. This is not specified in the docstring. Either we should state that this possibly returns a None (in which case we should add a type annotation ` -> Optional[bool]`), or we explicitly return some default value (probably `False`?)

youngmit (changes requested):
Minor changes

armi/materials/material.py
mgjarrett: Good point. The docstring says "Passes if the material does not have an incubation dose assigned..." and in this case "passing" the isBeyondIncuatbionDose test would be getting a `False` back.

armi/materials/material.py
mgjarrett: 👍

PR #267, opened 2021-03-11
Avoid plotting empty vectors in keffVsTime

drewejohnson:
Sometimes a database will be populated with an empty vector for the
uncontrolled multiplication factor. I believe this happens if the
value of keffUnc is not updated during an ARMI run. This leads to a
list of zeros being passed to the bookkeeping keffVsTime graph which
_technically_ evaluates to `True` and gets plotted. Using `any` for
the conditional check skips plotting empty lists and `None`s.

Signed-off-by: Andrew Johnson <drewejohnson@users.noreply.github.com>

### DIFF
--- a/armi/bookkeeping/plotting.py
+++ b/armi/bookkeeping/plotting.py
@@ -156,14 +156,14 @@ def keffVsTime(reactor, time, keff, keffUnc=None, ymin=None):
         Minimum y-axis value to target.
     """
     plt.figure()
-    if keffUnc:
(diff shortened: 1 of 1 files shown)


drewejohnson:
From the CLA and for the record

> I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.

youngmit (approved):

youngmit:
Looks good, thanks!

PR #272, opened 2021-03-29
Add tests for lumpedFissionProducts.

mgjarrett:
Improve the test coverage for lumped fission products. Fixed one bug in iterating through the directory returned by `LumpedFissionProductCollection.getMassFracs()`.

LumpedFissionProduct and LumpedFissionProductCollection have methods to create dictionaries of number fractions or mass fractions. The code creates these dictionaries with `nuclideBase` objects as keys, but in some places the key was used as if the code expects it to be a `str` type name. These commits change the usage of the dictionary to consistently handle these dictionary keys as `nuclideBase` objects.

### DIFF
--- a/armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
+++ b/armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
@@ -259,10 +259,10 @@ class LumpedFissionProduct(object):
 
     def printDensities(self, lfpDens):
         """Print densities of nuclides given a LFP density. """
-        nucs = self.keys()
+        nucs = list(self.keys())
         nucs.sort()
         for n in nucs:
-            runLog.info("{0:6s} {1:.7E}".format(n, lfpDens * self[n]))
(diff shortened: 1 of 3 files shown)


armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -259,10 +259,10 @@ def getLanthanideFraction(self):
 
     def printDensities(self, lfpDens):
         """Print densities of nuclides given a LFP density. """
-        nucs = self.keys()
+        nucs = list(self.keys())
         nucs.sort()
         for n in nucs:
```
ntouran, line 264: I think if you just make this `for n in sorted(self.keys()):` that would also work. But huh TIL that sorted takes more memory (because it makes an additional list apparently) https://blog.usejournal.com/list-sort-vs-sorted-list-aab92c00e17

mgjarrett: I'm going to make this change, but I have to force push because I rebased the branch onto the updated master, so the comment might go away.

Issue #268, opened 2021-03-16
KeyError exporting hex geometry from grid gui


drewejohnson:
Trying to save a hex geometry core using the FFTF blueprints and I kept getting
```
[warn] Cannot write geometry with asciimap. Defaulting to dict
```
messages. I added
```
                except Exception as ee:
                    from traceback import format_exception_only
                    runLog.error("".join(format_exception_only(type(ee), ee)).strip())
 ```
to https://github.com/terrapower/armi/blob/3cc980eb55eae7b6f8bcfcaa1c248d11c1758ca3/armi/utils/gridEditor.py#L1452-L1455 and the message became
```
[err ] KeyError: (<GeomType.HEX: 1>, 'full')
[warn] Cannot write geometry with asciimap. Defaulting to dict
```

Maybe related to #212? Possibly introduced in #229?

Using terrapower/fftf-isothermal-model@793fd8ec1a494b0254bdac61f9b145837c6b41ca and df4a4e27abbf190f975d04bd8eb22f59e98d3f22

**Comment by onufer** [comment]
Yeah, i think you are right about when it was introduced. I think that the issue is that:
 https://github.com/terrapower/armi/blob/3cc980eb55eae7b6f8bcfcaa1c248d11c1758ca3/armi/utils/asciimaps.py#L486-L504 

still expects a string. The `MAP_FROM_GEOM`  dict is:
```
('hex', 'third ')
('hex', 'full')
('hex_corners_up', 'full')
('cartesian', None)
('cartesian', 'full')
('cartesian', 'quarter ')
```

Probably needs to be changed to geometry.GeomType.HEX.

mgjarrett:
Looks like it could mostly be fixed by replacing the `geomType` in the return statement with a call to `geometry.GeomType.fromAny(geomType)`, which would handle `geomType` as a string or as the `GeomType` enum.

`def fromAny(cls, source: Union[str, "GeomType"]) -> "GeomType":`

The only problem is that the enumeration folds `HEX_CORNERS_UP` into `HEX`, so there is no enumeration for `geometry.HEX_CORNERS_UP`. We can check explicitly for the string `hex_corners_up`, it's just a little bit clunkier.

drewejohnson:
#273 has resolved my issue. Thank you @mgjarrett!

PR #273, opened 2021-03-30
Fix KeyError in asciiMapFromGeomAndSym.

mgjarrett:
Use the geometry.GeomType.fromAny() method to pass a string or
enumeration of a geomType and return the enumeration. Check explicitly 
for HEX_CORNERS_UP case because it doesn't exist in the enumeration.


mgjarrett:
This PR addresses  #268

PR #275, opened 2021-04-05
Several documentation updates.

ntouran:
* Removed warning about Python 3.9 since the installation is fairly
  smooth by now (except wxpython on linux is still a pain, but that's an
  extra).

* Improved error handling for pylint errors, preferring to inform the
  person building the docs of errors more than the person reading them.

* Fixed a broken link to a settings snippet

* Pinned docutils at <0.17 to avoid bug

### DIFF
--- a/.github/workflows/black.yaml
+++ b/.github/workflows/black.yaml
@@ -2,10 +2,11 @@ name: black
 
 on: [push, pull_request]
 
+# use workaround due to: https://github.com/psf/black/issues/2079#issuecomment-812359146
 jobs:
     check-formatting:
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-20.04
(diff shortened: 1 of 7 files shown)


youngmit:
Looks like the black action is busted on their end. Since you are already touching docs and tooling, could you try slipping in one of the workarounds here? https://github.com/psf/black/issues/2079#issuecomment-812359146

Also, while we are at it, `ubuntu-latest` seems like kind of a time bomb; should we change that to `ubuntu-20.04` to buy some stability? https://github.com/actions/virtual-environments

ntouran:
should be all good now

ntouran:
@youngmit  ping

youngmit (approved):
Looks good to me!

PR #278, opened 2021-04-09
Fix some clashes between new and old symmetry.

mgjarrett:
#276 updated the representation of symmetry from plain strings to an enumeration. The intent was that the new SymmetryType class could handle string or enumeration representations seamlessly to minimize breakage of old code that used the string representations.

While the strings still worked with little to no modification in most places, some pieces of code with the older format flew under the radar because the problems weren't caught by the unit tests. Also, the `.lower()` was misplaced in the `fromStr()` constructor for `SymmetryType` in a way that would break case insensitivity.

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -474,8 +474,8 @@ class GlobalFluxExecuter(executers.DefaultExecuter):
         """
         return (
             "FD" in self.options.kernelName
-            and self.options.symmetry.domain == geometry.THIRD_CORE
-            and self.options.symmetry.boundary == geometry.PERIODIC
+            and self.options.symmetry.domain == geometry.DomainType.THIRD_CORE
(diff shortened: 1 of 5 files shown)


ntouran:
In the grid editor `armi grids` when I try to make a new 1/3 grid, I get this tb:

```
  File "c:\users\ntouran\codes\armi-framework\armi\utils\gridEditor.py", line 368, in onNew
    self.parent.new(event)
  File "c:\users\ntouran\codes\armi-framework\armi\utils\gridEditor.py", line 1503, in new
    gridBp = dlg.getGridBlueprint()
  File "c:\users\ntouran\codes\armi-framework\armi\utils\gridEditor.py", line 1790, in getGridBlueprint
    symmetry = geometry.SymmetryType.fromStr(
TypeError: fromStr() takes 2 positional arguments but 4 were given
```

mgjarrett:
Thanks Nick. Should be the default constructor for `SymmetryType` there but I accidentally left it as `fromStr()`

ntouran (approved):

Issue #191, opened 2020-10-20
Add more forensic information to the Database metadata


youngmit:
The Database3 class stores some metadata in the HDF5 attributes of the file. Things like user, date, framework version, etc are included, but there is some lacking information for getting a full picture of what the deal was with a given run. At the very least, we need to add Application name and version. Anything else?

theJollySin:
My assumption is that this is the block of code we are talking about, and it looks like it already has the basics:

https://github.com/terrapower/armi/blob/master/armi/bookkeeping/db/database3.py#L563

Though, as you say, we could grab the `App` subclass name starting at:

* `armi.getApp()`

The only things that immediately jump out at me are maybe we could grab information about the OS:

* `sys.platform           #  just "linux"`
* `os.name                 #  posix`
* `platform.system()   # Linux`
* `platform.release()   # 4.15.0-140-generic`

Is this code usually run from within a Git repo?  If so, we can attempt to grab the current hash, in case someone is running off of a real release, with their own modifications:

* `subprocess.check_output(["git", "describe"])`

Is there anything else you can think of you'd like on there?

drewejohnson:
For the git version, would using a package like [`setuptools_scm`](https://pypi.org/project/setuptools-scm/) be appropriate? It embeds the output of `git describe` into the package distribution pretty seamlessly in my experience. Then, the output of `armi.__version__` might be something like `0.1.6-41-g3e7ac5a` indicating the most recent tag (`0.1.6`), number of commits ahead of the tag, and the short hash of the commit (3e7ac5a)

The downsides I see with using this as a way to get the revision are
1. It doesn't seem to provide an easy way to extract just the commit (without parsing the `__version__` string), let alone the full ~40 characters to specify the full commit
2. If you're exactly on a `git tag`, then the version string is just the tag, e.g., `0.1.6` with no commit data

Maybe also the hostname via `os.uid().hostname`?

theJollySin:
@drewejohnson   I like your `hostname` idea.  That's an easy win.


For the commit hash; my original concern was that when someone runs ARMI they might do so on their own dev branch, not a release version. And to debug a run, they will need some way to identify where their workspace was at run time.

Do users typically run the code from within the Git repo folders?

* If Yes, then the above concern is easy: we just the Git Python library to grab the current branch name and commit hash.
* If No, then we have to include the commit hash somewhere in the distribution, like you were saying.

youngmit:
[...]

PR #281, opened 2021-04-17
Adding forensic information to the Database metadata

theJollySin:
This PR is in reference to issue #191.

**Changes:**

1. The important changes are in database3.py.  I added three new blocks of metadata: platform info, plugin info, and Git commit.
2. Unit tests. Also, while running the `Database3` tests I noticed H5 files were being created in the top directory, but not cleaned.
3. `.gitignore`. While running the project unit tests I found many `*.nucdata` files were being created in tests folders.

**Discussion:**

In the #191 PR page we discussed several possibilities. The one I did not implement here was changing the version string to include the commit hash as @drewejohnson suggested (to something like `0.1.6-41-g3e7ac5a`). That is a nice idea, but I felt it would be over-stepping for me to change how the entire project is versioned.

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -62,10 +62,12 @@ import io
 import itertools
 import os
 import pathlib
+from platform import uname
 import re
 import sys
 import time
 import shutil
+from subprocess import check_output
 from typing import (
     Optional,
     Tuple,
@@ -101,7 +103,6 @@ from armi.bookkeeping.db import database
 from armi.reactor import systemLayoutInput
 from armi.utils.textProcessors import resolveMarkupInclusions
 from armi.nucDirectory import nuclideBases
-
 from armi.settings.fwSettings.databaseSettings import CONF_SYNC_AFTER_WRITE
 
 ORDER = interfaces.STACK_ORDER.BOOKKEEPING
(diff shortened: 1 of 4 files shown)


youngmit:
This looks great! The only request that I would make is to try using a `TemporaryDirectoryChanger` (see https://github.com/terrapower/armi/blob/master/armi/utils/directoryChangers.py) in favor of explicit removal of the test artifacts. There is a pretty good example of this here: https://github.com/terrapower/armi/blob/3e7ac5ae41a89c6cb6f11b1996961926c048d996/armi/bookkeeping/tests/test_databaseInterface.py#L79

The rationale is that we usually run our tests in parallel, and sometimes two runners will try to delete the same file at the same time. Sometimes not, depending on how the individual tests get scheduled. Then it's a bit of a time bomb until someone adds a new test and we get a surprise heisenbug.

Thanks!

youngmit (changes requested):
See above. Other than that, this looks great!

theJollySin:
@youngmit That's a cleaner solution, sure.  I'll try it.

theJollySin:
@youngmit  Thanks!  I used the `TemporaryDirectoryChanger()` function and those tests are a lot cleaner.

While I was at it, I also applied the same function to fix a test in `test_blockConverter.py`. It has the same issue with leaving files around after testing.

armi/bookkeeping/db/database3.py
```diff
@@ -568,6 +569,46 @@ def open(self):
         self.h5db.attrs["armiLocation"] = os.path.dirname(armi.ROOT)
         self.h5db.attrs["startTime"] = armi.START_TIME
         self.h5db.attrs["machines"] = numpy.array(armi.MPI_NODENAMES).astype("S")
+        # store platform data
+        platform_data = uname()
+        self.h5db.attrs["platform"] = platform_data[0]
+        self.h5db.attrs["hostname"] = platform_data[1]
+        self.h5db.attrs["platformRelease"] = platform_data[2]
+        self.h5db.attrs["platformVersion"] = platform_data[3]
+        self.h5db.attrs["platformArch"] = platform_data[4]
```
drewejohnson, line 578: (piping in because I was tagged in the PR) this is great! Just learned about this package through this pull request.

One thing that might help readability would be to use the attributes provided via the [`uname` `namedtuple`](https://docs.python.org/3/library/platform.html#platform.uname), e.g.,
```python
platform = platform_data.system
hostname = platform_data.node
...
```

theJollySin: Oh, it's a namedtuple.  When I printed it out, I thought it was just a regular tuple.  Nice catch, I'll make the change.

theJollySin: Cool, I made the change.

youngmit (approved):
Thanks, @theJollySin!

PR #283, opened 2021-04-23
Addressing PyLint Concerns

theJollySin:
This is a low-priority, low-risk PR addressing purely PyLint concerns. It specifically addresses two PyLint warnings:

* **R0201** - method could be function
* **W0614** - unused imports

Also, there were a great many warnings in the unit test files that just don't matter, so I suppressed them.  For instance, do you really need a docstring on a unit test class?

NOTE: Viewing the changes in this PR might be easier if you look at the commits individually.

### DIFF
--- a/armi/_bootstrap.py
+++ b/armi/_bootstrap.py
@@ -17,9 +17,7 @@ Collection of code that needs to be executed before most ARMI components are saf
 import.
 """
 
-import os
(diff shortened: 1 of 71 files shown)


youngmit (approved):
Thanks again! Yeoman's work, this...

PR #284, opened 2021-04-23
Python 2 to 3 Class Definition

theJollySin:
In later versions of Python 2, it became popular to use the "new" class type, which meant switching from this class definition:

    class Thing:

to this:

    class Thing(object):

According to the `setup.py` file in this repo, this project is purely Python3.  In which case, the "new" class type is default, and this `(object)` is unnecessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ doc/tutorials/case-suite
 armi/tests/tutorials/case-suite
 .apidocs/
 doc/gallery
+htmlcov/
 
 # No workspace crumbs.
 *~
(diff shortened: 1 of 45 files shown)


youngmit (approved):
Thanks! I had been using these as a sort of when's-the-last-time-someone-touched-this measure, but I think the time has come. Honestly, fewer than I'd have expected.

theJollySin:
> Thanks! I had been using these as a sort of when's-the-last-time-someone-touched-this measure, but I think the time has come. Honestly, fewer than I'd have expected.

Yeah, it was only about 10% of the classes in the repo. 

Also, if you're interested, I have a little Git/Bash script that you can run on a repo to see which files haven't been touched in the longest time:

https://antineutrino.net/2018/08/30/git-cleanup

PR #286, opened 2021-04-25
Adding the 1D/2D mesh reading/writing capabilities to the GEODST interface file.

jakehader:
This adds the 1D and 2D reading capabilities to the existing `GEODST` interface file. This is useful for obtaining the coarse and mesh points in simple slab, sphere, and cylindrical (rz) geometries .


ntouran (approved):
thanks! For testing, do you happen to have a simple dummy file that can go alongside the hex geodst? Alternatively, we could fill in the data structure and test writing and then reading rather than reading an existing file.

PR #288, opened 2021-04-27
Fix order of operation bug in rotatePins.

mgjarrett:
RotatePins uses modulus (%) to avoid over-rotating, but the order of
operations was wrong:

initialRotation + newRotation % 6

instead of

(initialRotation + newRotation) % 6

In some cases, this could lead to creation of pin index numbers that
were larger than the number of pins.

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -1703,10 +1703,9 @@ class HexBlock(Block):
         numPins = self.getNumComponents(Flags.CLAD)  # number of pins in this assembly
         rotateIndexLookup = dict(zip(range(1, numPins + 1), range(1, numPins + 1)))
 
-        currentRotNum = self.getRotationNum()
         # look up the current orientation and add this to it. The math below just rotates from the
         # reference point so we need a total rotation.
-        rotNum = currentRotNum + rotNum % 6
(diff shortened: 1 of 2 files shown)


jakehader:
@mgjarrett, Can you add a unit test that fails with the previous implementation and that also demonstrates that this fixes the rotation behavior?

mgjarrett:
I added a unit test that would fail with the previous implementation.

I don't know why black is failing. It seems unhappy with some docstrings in files that I didn't touch. I checked and it seems like I'm running the same version of black, but when I run black on these files manually it does nothing. So I'm kind of perplexed.

jakehader:
@mgjarrett make sure you have the latest pip and then try to update it. It seems that this CI is using an updated version.

mgjarrett:
@jakehader I was using 20.8b1, but I think this is out of my control anyway. The affected files are runLog.py, reactors.py, database3.py, mainInterface.py, a bunch of materials files, etc. These files are not modified in the pull request.

youngmit:
Black's GH action is kind of busted in that when you request a specific version, it still uses whatever the latest happens to be. We already merged a PR with different formatting so I'd say we just merge this. There is some ongoing discussion here: https://github.com/psf/black/issues/1939 and here: https://github.com/psf/black/pull/1940

PR #289, opened 2021-04-28
Add testing and validation checks to the cross section settings.

jakehader:
Adding validation checks in `crossSectionSettings.py` to warn the user if extra options are provided in the various geometry options and enforcing that the `fileLocation` and `geometry` cannot both be set simultaneously. Finally, if a value in the `crossSectionControl `setting is set to None or any empty dictionary then it will not be initialized in the setting or serialized when output to a file.


armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -1131,12 +1131,6 @@ def updateNuclideTemperatures(self, blockCollectionByXsGroup=None):
     "FluxWeightedAverage": FluxWeightedAverageBlockCollection,
 }
 
-HOMOGENEOUS_BLOCK_COLLECTIONS = {
-    "Median": MedianBlockCollection,
-    "Average": AverageBlockCollection,
-    "FluxWeightedAverage": FluxWeightedAverageBlockCollection,
-}
-
```
jakehader, line 1139: This is unused and just duplicate information for the normal `BLOCK_COLLECTIONS` above.

armi/tests/armiRun.yaml
```diff
@@ -19,11 +19,6 @@ settings:
       numInternalRings: 1
       numExternalRings: 1
     XA:
-      geometry: 0D
-      externalDriver: true
-      useHomogenizedBlockComposition: false
-      numInternalRings: 1
-      numExternalRings: 1
```
jakehader, line 26: This is removed because we should only be defining `fileLocation` OR `geometry`, not both

armi/physics/neutronics/tests/test_neutronicsPlugin.py
```diff
@@ -35,7 +35,6 @@
     blockRepresentation: Median
 BA:
     geometry: 1D slab
-    criticalBuckling: false
```
jakehader, line 38: critical buckling is removed since it is an unsupported option for `1D slab`.

jakehader:
I must have a different version of black then what is being used for the framework. I will update and resubmit.

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -109,6 +157,14 @@ def __init__(
         self.mergeIntoClad = mergeIntoClad
         self.fileLocation = fileLocation
         self.meshSubdivisionsPerCm = meshSubdivisionsPerCm
+        if self.geometry is not None or self.fileLocation is not None:
```
ntouran, line 160: why only validate in this condition? Why not always validate?

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -125,65 +181,136 @@ def isPregenerated(self):
...
 
+    def _validate(self):
+        """
+        Perform validation checks on the set of inputs.
+
+        This makes some design assumptions about the lattice
+        physics code (i.e., critical buckling search is not
+        a valid option for non-0D problems), so if different
+        behavior is desired then it is recommended to subclass
+        this and to make a new cross section setting.
```
ntouran, line 192: This is validation beyond what voluptuous does because it's extra logic, etc, right?

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -125,65 +181,136 @@ def isPregenerated(self):
...
+        -----
+        This checks for any inconsistencies in the definition of
+        the inputs.
+        """
+        if self.fileLocation is None and self.geometry is None:
+            raise ValueError(f"{self} is missing a geometry input or a file location.")
+
+        if self.fileLocation is not None and self.geometry is not None:
+            raise ValueError(
+                f"The file location and geometry inputs cannot be combined in {self}. "
```
ntouran, line 204: slight clarification to explain what to do could be like: "Either file location or geometry inputs must be given, but not both. Remove one or the other"

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -125,65 +181,136 @@ def isPregenerated(self):
...
+                f"File location set to {self.fileLocation}. "
+                f"Geometry set to {self.geometry}"
+            )
+
+        # Check for valid inputs when the file location is supplied.
+        invalids = []
+        if self.fileLocation is not None:
+            for var, val in self:
+                # Skip the ``xsID`` and ``fileLocation`` attributes.
```
ntouran, line 213: why skip them?

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -125,65 +181,136 @@ def isPregenerated(self):
...
+                if var in [CONF_XSID, CONF_FILE_LOCATION, CONF_BLOCK_REPRESENTATION]:
+                    continue
+                if val is not None:
+                    invalids.append((var, val))
```
ntouran, line 217: Is there a clear reason why the voluptuous "extra keys" feature won't work for this?  That might be useful to document since we're kind of mixing vol validation with manual validation.

https://github.com/alecthomas/voluptuous#extra-dictionary-keys

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -125,65 +181,136 @@ def isPregenerated(self):
...
+            for var, val in invalids:
+                runLog.warning(f"\tAttribute: {var}, Value: {val}")
+
+        # Check for valid inputs when the geometry is supplied.
+        invalids = []
+        if self.geometry is not None:
+            validOptions = VALID_INPUTS_BY_GEOMETRY_TYPE[self.geometry]
+            for var, val in self:
+                if var not in validOptions and val is not None:
+                    invalids.append((var, val))
```
ntouran, line 232: also feel like re-writing some vol capability, right? or am I missing a detail?

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -125,65 +181,136 @@ def isPregenerated(self):
...
-                criticalBuckling=False,
-                mergeIntoClad=[],
-                blockRepresentation=blockRepresentation,
-                numExternalRings=1,
-                validBlockTypes=validBlockTypes,
-            )
+            defaults = {
+                CONF_GEOM: "1D cylinder",
+                CONF_DRIVER: "",
+                CONF_MERGE_INTO_CLAD: ["gap"],
```
ntouran, line 283: this should probably be in a SFR design specific plugin. could be a framework plugin but doesn't seem right here. right?

ntouran (changes requested):
There's a lot of fairly complex stuff going on in here. I just have some general questions but more than anything just want more clarity on why the defaults are set to what they're set to and see if we can try to keep the really design specific stuff packaged in a design-specific module.

armi/physics/neutronics/crossSectionSettings.py
jakehader: I am not looking to refactor this out right now, just trying to clarify which are the actual expected defaults rather than the ones that were just put here before.

armi/physics/neutronics/crossSectionSettings.py
jakehader: Not sure if extra keys would work. Seems like it would make the Schema very complex and probably more difficult to understand.

armi/physics/neutronics/crossSectionSettings.py
jakehader: Yeah

jakehader: Actually, the vol is performed only when an XSModelingOption is created in the `_load` method of XSSettingDef. It seems that the XSModelingOption object itself should be responsible for its our attribute/input validation and the vol. package only helps the user conform to the specific inputs when setting the `crossSectionSettings` in the case settings.

armi/physics/neutronics/crossSectionSettings.py
jakehader: See comment about. The Schema is actually only used for loading the dictionary before it gets loaded into the XSModelingOption class. Even if the Schema handled this, the XSModelingOption could still then be incorrectly set/defined.

armi/physics/neutronics/crossSectionSettings.py
jakehader: See above.

armi/physics/neutronics/crossSectionSettings.py
jakehader: It seems that this implementation of the crossSectionSettings is pretty much all design specific, but without this then the behavior of the `CrossSectionGroupManager` is not well undefined. Other than `fileLocation`, `blockRepresentation`, and `validBlockTypes` probably all other options should be removed. `geometry` probably also makes sense because it feeds into the `latticePhysicsInterface` in `getWriters`.

armi/physics/neutronics/crossSectionSettings.py
jakehader: Updated in revision.

armi/physics/neutronics/crossSectionSettings.py
jakehader: Updated the comment in the revision.

armi/physics/neutronics/crossSectionSettings.py
jakehader: Updated in revision.

armi/physics/neutronics/crossSectionSettings.py
jakehader: Updated the docstring since it's not really validating design specifics here, just warning when things might be over defined. I agree that this stuff is design and application specific (see https://github.com/terrapower/armi/pull/289#discussion_r622737320)

jakehader:
> There's a lot of fairly complex stuff going on in here. I just have some general questions but more than anything just want more clarity on why the defaults are set to what they're set to and see if we can try to keep the really design specific stuff packaged in a design-specific module.

I agree, I think that we should generally think about a refactor here.

armi/physics/neutronics/tests/test_crossSectionSettings.py
```diff
@@ -142,6 +144,113 @@ def test_yamlIO(self):
...
+        # caseSetting using the ``XSModelingOptions`` or using
+        # a dictionary.
+        cs = _setInitialXSSettings()
+        cs[CONF_CROSS_SECTION].update(
+            {"CA": XSModelingOptions("CA", geometry="0D"), "DA": {"geometry": "0D"}}
+        )
+        self.assertIn("AA", cs[CONF_CROSS_SECTION].keys())
+        self.assertIn("BA", cs[CONF_CROSS_SECTION].keys())
+        self.assertIn("CA", cs[CONF_CROSS_SECTION].keys())
+        self.assertIn("DA", cs[CONF_CROSS_SECTION].keys())
```
ntouran, line 193: keys is likely (or at least should be!) unnecessary since cs acts as a dict.

armi/physics/neutronics/crossSectionSettings.py
ntouran: Since vol's job is to "validate data structures", regardless of where they are in their lifecycle, do you think we could use only vol for this if we refactor this system to traffic in more isolated data structures? 

Overall my biggest thought here is that we should add lots of info to the docstring of the module explaining what we think is wrong with this to help guide future framework users and developers because it all must be pretty surprising as is.

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -72,44 +123,246 @@
...
+    Attributes
+    ----------
+    xsID : str
+        Cross section ID that is two characters maximum (i.e., AA).
+
+    geometry: str
+        The geometry modeling approximation for regions of the core with
+        this assigned xsID. This is required if the ``fileLocation``
+        attribute is not provided. This cannot be set if the ``fileLocation``
+        is provided.
```
ntouran, line 243: could `:py:` link to the set above that has the valid options

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -72,44 +123,246 @@
...
+
+    fileLocation: list of str
+        This should be a list of paths where the cross sections for this
+        xsID can be copied from. This is required if the ``geometry``
+        attribute is provided. This cannot be set if the ``geometry``
+        is provided.
```
ntouran, line 249: gonna be hard to satisfy both of those :P

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -72,44 +123,246 @@
...
+
+    validBlockTypes: str or None
+        This is a configuration option for how the cross section group manager
```
ntouran, line 252: could shorten and just say: `Specifies how the cross...`

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -72,44 +123,246 @@
...
+        representative block, the neutron source is taken from a single
+        nuclides fission spectrum (i.e., U235). This is particularly useful
+        for configuring SERPENT 2 lattice physics calculations.
+
+    externalDriver : bool
+        This is a lattice physics configuration option that can be used
+        to determine if the fixed source problem is internally driven
+        or externally driven by the ``driverID`` region. Externally
+        driven means that the region will be placed on the outside of the
+        current xsID block/region.
```
ntouran, line 284: and if it's false it'll be on the inside?

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -72,44 +123,246 @@
...
+
+    useHomogenizedBlockComposition : bool
+        This is a lattice physics configuration option that is useful for
+        modeling spatially dependent problems (i.e., 1D/2D). If this is
+        True then the representative block for the current xsID will be
+        be homogenzied. If this is False then the block will be represented
+        in the geometry type selected. This is mainly used for 1D cylindrical
+        problems.
```
ntouran, line 292: wait but if it's 1D and homogenized at the same time... I'm confused

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -72,44 +123,246 @@
...
+
+    numInternalRings : int
+        This is a lattice physics configuration option that is used to
+        specify the number of rings for the internal driver block/region.
```
ntouran, line 296: mesh rings? composition rings? what kind of rings?

ntouran (changes requested):
excellent cleanup. I've added lots of doc update suggestions. There's at least one that you should fix

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -51,19 +50,71 @@
...
+XS_GEOM_TYPES = {
+    "0D",
+    "1D slab",
+    "1D cylinder",
+    "2D hex",
+}
+
+# This dictionary defines the valid set of inputs based on
+# the geometry type within the ``XSModelingOptions``
+_VALID_INPUTS_BY_GEOMETRY_TYPE = {
```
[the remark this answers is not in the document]
jakehader: Yeah the Schema would be a bit large/confusing i think.

armi/physics/neutronics/crossSectionSettings.py
jakehader: Its the difference between having multiple rings of control, clad, coolant, control, clad, coolant, ... for the representative block and being driven by external "fuel" and just being a homogeneous mixture of "control/clad/coolant" as a region and then "fuel" as the separate region. This allows for an external spectrum source onto the control region, but without accounting for the spatial self-shielding effects.

armi/physics/neutronics/crossSectionSettings.py
jakehader: grid-based rings as in 1 ring is 1 blocks worth of material, 2 rings is 7 blocks worth of material, etc.

ntouran (approved):

PR #298, opened 2021-05-11
Quiet the cross section setting recommendations in the stdout

jakehader:
Remove the verbose cross section setting messages to the user in the stdout and set them to debug only.


partofthething (approved):
Looks fine. Should you explain in the comment that the fact that there are invalids is no big deal in this case?

PR #100, opened 2020-07-20
Add C5G7 compositions and dimensions, updating LWR tutorial to reflect changes.

onufer:

PR #301, opened 2021-05-11
Add C5G7 compositions and dimensions, updating LWR tutorial to reflect changes

jakehader:
This updates the C5G7 tutorial inputs to reflect the actual MOX compositions and includes other features, such
as the fission chamber guide tubes.


jakehader:
@ntouran, this pull request is the same as #100, but was taken over to implement the comments that you had suggested. The documentation for the LWR inputs uses `literalinclude` statements so the docs should remain up-to-date. Additionally, I ran the plotting script in `walkthrough_lwr_inputs.rst` manually to make sure that the case still loads and can plot the pin information.

PR #307, opened 2021-05-26
Make it possible to add interfaces with the same function

chris10mckenz:
Originally addInterface would raise an exception when the Interface
stack already contained an Interface with the same function as the one
passed into it. This was limiting when you had different plug-ins but
wanted the right Interface to take precedence.
So, now it adds an Interface even if it has the same function as one
previously if the new Interface is a subclass of the current Interface
with that function. This keeps the Interface that is more derived and
removes the old Interface. This still throws an exception if the new
Interface has the same function but shares no inheritance with the old
Interface. If the old Interface is instead a subclass of the new
Interface, neither interface is removed or added.

### DIFF
--- a/armi/operators/operator.py
+++ b/armi/operators/operator.py
@@ -548,7 +548,12 @@ class Operator:  # pylint: disable=too-many-public-methods
             self.addInterface(klass(self.r, self.cs), **kwargs)
 
     def addInterface(
-        self, interface, index=None, reverseAtEOL=False, enabled=True, bolForce=False
+        self,
+        interface,
+        index=None,
+        reverseAtEOL=False,
+        enabled=True,
+        bolForce=False,
     ):
         """
         Attach an interface to this operator.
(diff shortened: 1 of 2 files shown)


armi/operators/operator.py
```diff
@@ -590,21 +595,30 @@ def addInterface(
...
-                "function is not supported.".format(
-                    interface, iFunc, interface.function
+            if issubclass(type(iFunc), type(interface)):
+                runLog.info(
+                    "Ignoring Interface {newFunc} because existing interface {old} already "
+                    " more specific".format(newFunc=interface, old=iFunc)
+                )
+                return
+            elif issubclass(type(interface), type(iFunc)):
+                self.removeInterface(iFunc)
```
youngmit, line 607: Which of these branches gets taken is I think just a matter of the order in which a fist full of Interfaces are added. So I think it makes sense to report something to the user either way. Could you add a `runLog` message saying that we are replacing the existing interface with the new one?

armi/operators/tests/test_operators.py
```diff
@@ -122,7 +181,8 @@ def test_workerException(self):
 
 
 if __name__ == "__main__":
-    args = ["mpiexec", "-n", "2", "python", "-m", "unittest"]
+    """args = ["mpiexec", "-n", "2", "python", "-m", "unittest"]
     args += ["armi.tests.test_operators.OperatorTests"]
-    subprocess.call(args)
-    # unittest.main()
+    subprocess.call(args)"""
```
youngmit, line 186: This is a bit confusing as a string literal. Kinda looks like a docstring? Probably better to make it a proper comment.

youngmit (changes requested):
Just a couple of minor things. Mostly that it would be nice to report something to the user for both the more- and less-derived cases.

chris10mckenz:
#307

Issue #311, opened 2021-06-09
Helix diameter incorrect in tutorial documentation


jakehader:
see: https://github.com/terrapower/armi/blob/master/armi/tests/tutorials/anl-afci-177-blueprints.yaml#L12-L23

The helix diameter for the ANL-AFCI-177 inputs should be 0.8888 cm based on the Table 4.4 of https://publications.anl.gov/anlpubs/2008/05/61507.pdf, but currently is set to 0.84825 cm in the blueprints file. The helix diameter should be equivalent to the pin pitch, which is also defined as the **Wire OD** + **Clad OD**. Currently, the 0.84825 cm is calculated incorrectly as **Wire OD / 2.0** + **Clad OD**.

![image](https://user-images.githubusercontent.com/26268658/121441199-32100c80-c93e-11eb-92af-89768ac92f83.png)

ntouran:
Yeah I dunno if it'd be more clear in the helix component as helixdiameter = 2* helixradius = 2*(clad radius + wire radius) = 2*(cladOD/2 + wireOD/2) = cladOD + wireOD. Nice catch.

jakehader:
@ntouran, is there even a reason to have `helixDiameter` as a parameter since it seems over specified? What is it physically mean to have a `helixDiameter` smaller or larger than the **Wire OD** + **Clad OD**? Does larger imply "slop" in the pin lattice?

PR #312, opened 2021-06-10
Fix to ANL-AFCI-177 helix diameter

jakehader:
Updating the ANL-AFCI-177 tutorial wire wrap helix diameter so that the pin pitch aligns with the reference. This
changed the helix diameter from `0.84825` cm to `0.8888 cm` and fixes the issue identified in #311.

### DIFF
--- a/armi/tests/tutorials/anl-afci-177-blueprints.yaml
+++ b/armi/tests/tutorials/anl-afci-177-blueprints.yaml
@@ -18,9 +18,9 @@ blocks:
             Tinput: 25.0
             Thot: 450.0
             axialPitch: 30.0
-            helixDiameter: 0.84825
+            helixDiameter: 0.8888
             id: 0.0
-            od: 0.0805
+            od: 0.0808
             mult: 271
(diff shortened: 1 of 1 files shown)


armi/tests/tutorials/anl-afci-177-blueprints.yaml
```diff
@@ -111,9 +111,9 @@ blocks:
             Tinput: 25.0
             Thot: 450.0
             axialPitch: 30.0
-            helixDiameter: 0.84825
+            helixDiameter: 0.777
```
jakehader, line 114: This is updated to 0.777 cm because the reflector slug OD is 0.6962 cm and not the same as the clad OD dimensions for the fuel. This leads to a smaller pitch.

ntouran: I wonder where that reflector pin OD came from. Maybe later we should just make that equal to the fuel pin OD just to make it less strange.

ntouran (approved):

PR #322, opened 2021-06-30
Add expand-bp entry point

youngmit:
This entry point will process the !include directives in an ARMI YAML
file, emitting the results to standard output. This also adds a class
attribute the the EntryPoint class to control whether a splash text
should be writen to stdout, allowing the output to be safely captured in
a new YAML file.

### DIFF
--- a/armi/cli/checkInputs.py
+++ b/armi/cli/checkInputs.py
@@ -41,6 +41,8 @@ def _runInspectorOnSettings(cs):
 class ExpandBlueprints(EntryPoint):
     """
     Perform expansion of !include directives in a blueprint file.
+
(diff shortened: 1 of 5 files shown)


armi/cli/checkInputs.py
```diff
@@ -36,6 +38,31 @@ def _runInspectorOnSettings(cs):
     return issues
 
 
+class ExpandBlueprints(EntryPoint):
+    """
+    Perform expansion of !include directives in a blueprint file.
```
ntouran, line 43: I'd add more to this for more context to a new user. Maybe:

`Process any !include directives in a blueprint file and print the expanded content to stdout`

ntouran (approved):
approve with one small suggestion to the help text

PR #323, opened 2021-07-06
Create Grid and Locators for a Block

chris10mckenz:
In blocks, creates autoCreateSpatialGrids() that when called upon for
a block with no spatialGrid currently, will create one for it if it
meets certain conditions. These conditions are that among the blocks
components, the multiplicities are either equal to 1 or N. (So there can
be no greater than two multiplicities among the components and one of
the multiplicities must be 1). Things with mult equal to 1 will have a
spatialLocation located at the center of the grid. Those with
multiplicity N will have a MultIndexLocation and appear at every pin
position location in the grid (they will appear N times). The
functionality of this is for the added feature of Block Diagrams.

In blockBlueprints, makes it so that at construction, a block is given a
spatialGrid and its components spatialLocators if certain conditions are
met. These conditions are that it is a HexBlock (meaning the function
autoCreateSpatialGrids within blocks.py is not implemented for non-Hex
Blocks, and an exception is caught and the grid remains none) and it has
basic multiplicities as defined within autoCreateSpatialGrids (mult = 1
or N, only two possible multiplicities). This sets up blocks so that
they may later be used to create basic block diagrams in cases where the spatialGrid was
not originally provided but the component placement could be inferred.

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -327,10 +327,19 @@ class Block(composites.Composite):
 
     def autoCreateSpatialGrids(self):
         """
+        Creates a spatialGrid for a Block.
+
         Blocks do not always have a spatialGrid from Blueprints, but, some Blocks can have their
         spatialGrids inferred based on the multiplicty of their components.
         This would add the ability to create a spatialGrid for a Block and give its children
         the corresponding spatialLocators if certain conditions are met.
+
+        Raises
+        ------
+        ValueError
+            If the multiplicities of the block are not only 1 or N or if generated ringNumber leads to more positions than necessary.
+
+
         """
         raise NotImplementedError()
 
@@ -1941,20 +1950,29 @@ class HexBlock(Block):
     def autoCreateSpatialGrids(self):
         """
         Given a block without a spatialGrid, create a spatialGrid and give its children
-        the corresponding spatialLocators (if it is a simple block). In this case, a simple block would
-        be one that has either multiplicity of components equal to 1 or n but no other multiplicities.
-        Otherwise, raise a ValueError since the number of multiplicities makes it so we cannot generate
-        a grid for this block.
+        the corresponding spatialLocators (if it is a simple block).
 
-        In other words, here we gather all components to either be a multiIndexLocation containing all
+        In this case, a simple block would
+        be one that has either multiplicity of components equal to 1 or N but no other multiplicities. Also, this should only happen when N fits exactly into a given number of hex rings.
+        Otherwise, do not create a grid for this block.
+
+        Notes
+        -----
+        If the block meets all the conditions, we gather all components to either be a multiIndexLocation containing all
         of the pin positions, otherwise, locator is the center (0,0).
+
+        Raises
+        ------
+        ValueError
+            If the multiplicities of the block are not only 1 or N or if generated ringNumber leads to more positions than necessary.
+
+
         """
 
         # Check multiplicities...
         mults = {c.getDimension("mult") for c in self}
 
         if len(mults) != 2 or 1 not in mults:
-            self.spatialGrid = None
             raise ValueError(
                 "Could not create a spatialGrid for block {}, multiplicities are not 1 or N they are {}".format(
                     self.p.type, mults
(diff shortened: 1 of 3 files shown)


armi/reactor/blocks.py
```diff
@@ -1929,6 +1938,48 @@ def _getPinCoordinatesHex(self):
                 coordinates.append(xyz)
         return coordinates
 
+    def autoCreateSpatialGrids(self):
+        """
+        Given a block without a spatialGrid, create a spatialGrid and give its children
+        the corresponding spatialLocators (if it is a simple block). In this case, a simple block would
+        be one that has either multiplicity of components equal to 1 or n but no other multiplicities.
```
youngmit, line 1945: ...where N fits exactly into a given number of hex rings. I think this is pretty critical. We should also be verifying this in the implementation and throw a ValueError if it's not satisfied.

armi/reactor/blocks.py
```diff
@@ -1929,6 +1938,48 @@ def _getPinCoordinatesHex(self):
...
+        if len(mults) != 2 or 1 not in mults:
+            self.spatialGrid = None
+            raise ValueError(
+                "Could not create a spatialGrid for block {}, multiplicities are not 1 or N they are {}".format(
+                    self.p.type, mults
+                )
+            )
+
+        spatialLocators = grids.MultiIndexLocation(grid=self.spatialGrid)
+        ringNumber = hexagon.numRingsToHoldNumCells(self.getNumPins())
```
youngmit, line 1965: Check that the number of locations in `ringNumber` rings is the same as the number of pins

armi/reactor/blocks.py
```diff
@@ -1929,6 +1938,48 @@ def _getPinCoordinatesHex(self):
...
+
+        In other words, here we gather all components to either be a multiIndexLocation containing all
+        of the pin positions, otherwise, locator is the center (0,0).
+        """
+
+        # Check multiplicities...
+        mults = {c.getDimension("mult") for c in self}
+
+        if len(mults) != 2 or 1 not in mults:
+            self.spatialGrid = None
```
youngmit, line 1957: No need to do this. We should be checking that the necessary assumptions are satisfied up-front. Then if not, we shouldn't alter the state of the block at all.

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -162,6 +162,12 @@ def construct(
         b.verifyBlockDims()
         b.spatialGrid = spatialGrid
 
+        if b.spatialGrid is None:
+            try:
+                b.autoCreateSpatialGrids()
+            except (ValueError, NotImplementedError) as e:
+                runLog.warning(str(e), single=True)
+                b.spatialGrid = None
```
youngmit, line 170: No need to do this. If it raised, we should  be able to assume that the block is unmodified.

armi/reactor/blocks.py
```diff
@@ -325,6 +325,15 @@ def getSmearDensity(self, cold=True):
 
         return smearDensity
 
+    def autoCreateSpatialGrids(self):
+        """
+        Blocks do not always have a spatialGrid from Blueprints, but, some Blocks can have their
+        spatialGrids inferred based on the multiplicty of their components.
+        This would add the ability to create a spatialGrid for a Block and give its children
+        the corresponding spatialLocators if certain conditions are met.
```
youngmit, line 333: Follow [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) format for docstrings. So short, one-line description, followed by a longer, more complete description. We have also decided that this can raise a ValueError if the implementation cannot be applied to a given block; there should be a [raises](https://numpydoc.readthedocs.io/en/latest/format.html#raises) section explaining what that means, and reminding the implementer/reassuring the user that in such a case, the block will be left un-modified.

armi/reactor/tests/test_blocks.py
```diff
@@ -1704,6 +1704,40 @@ def test_getPinCenterFlatToFlat(self):
...
+            if c.getDimension("mult", cold=True) == 169:
+                # Then it's spatialLocator must be of size 217
+                locations = c.spatialLocator
+                self.assertEqual(type(locations), grids.MultiIndexLocation)
+                mult = 0
+                for loc in locations:
+                    mult = mult + 1
+                self.assertEqual(mult, 169)
+
+    def test_gridNotCreatedMultipleMultiplicities(self):
```
youngmit, line 1723: Can we get something like this, but to test that we raise when the multiplicity for the pins doesn't fit snugly into the number of rings?

youngmit (changes requested):
Mostly looks great. Just need to check that the pin multiplicity matches the number of rings exactly, a little cleanup, and do some docstring stuff.

youngmit (approved):

PR #324, opened 2021-07-06
Replacing Pathlib Glob with Glob in Case.py

### LINKED ISSUE #321: Obscure python system warning for a bad file name.
A bad file name in an ARMI script generated the obscure python system warning "NotImplementedError: Non-relative patterns are unsupported."  The warning does not reflect the actual problem, and is confusing. 

The error was being thrown from python glob().  Beginning from line #756 of "case.py", the current logic in this case is: 

           for f in files:
                path = pathlib.Path(f)
                if path.is_absolute() and path.exists() and path.is_file():
                    # looks like an extant, absolute path; no need to do anything
                    pass
                else:
                    # relative path/glob. Should be safe to just use glob resolution
                   srcFiles = list(sourceDirPath.glob(f))

In this case, a .py script could not be found under the hard-coded name.  As a result path.exists() and path.is_file() were both false.  The above logic threw the script name into a relative path.  This resulted in the obscure python system warning when an absolute name was provided.  I suggest inserting a warning similar to the logic below immediately following “else:”, above:

                   if not (path.exists() and path.is_file()):
                        runLog.warning(
                              f"Input file named ‘{path}’ not found.  Checking for relative address."
                        ) 

Such a change will likely shorten debugging times when a bad script name is inadvertently passed to ARMI.

jakehader:
Updating the `copyInterfaceInputs` within case.py to use glob.glob rather than pathlib.glob. 
For Windows environments with Python 3.7.2 the pathlib.glob does not preserve the case 
sensitivity of file names when it finds a match. This is problematic when interfaces are copying 
python scripts (like fuel management, or custom control rod logic) since the imports of these 
scripts is case sensitive.

Fixes #321

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -37,6 +37,7 @@ from typing import Dict, Optional, Sequence
 import tabulate
 import six
 import coverage
+import glob
 
 import armi
 from armi import context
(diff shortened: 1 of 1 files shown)


youngmit:
I should mention that while this fixes the issue that ultimately led to the discovery of #321, I wouldn't say that this really fixes it. The logic in the `if` statement is still suspect I.M.O., and still needs to be addressed. This is great though

PR #325, opened 2021-07-06
Improve plotting behavior.

chris10mckenz:
Adjust plotFaceMap where _setPlotValText now takes in the collection
currently used for plotting so that we have a reference to what colors
were used for each Hex. In passing in the collection, we get access to
its color map and norm so we can go back and find what patch color went
with what value originally. With this information we can figure out if
it is a dark or a light color and assign text to be white if a dark
patch and black if a light patch.

In addition to this, also, changed colorbar to be fig.colorbar and not
ax.colorbar since ax.colorbar was not working and was getting
AttributeError: 'AxesSubplot' object has no attribute 'colorbar' so I
looked into it and I think we want to add the colorbar to the fig and
not the ax.

Also, for plotFaceMapLegend, just passed in the collection instead of
the norm and cmap separately and also added a size argument that
defaults to 9 (the original legend size for faceMaps) since this could
make it more generalized for use to create different legends later on that may
need to be a different size.

### DIFF
--- a/armi/utils/plotting.py
+++ b/armi/utils/plotting.py
@@ -33,6 +33,8 @@ from armi.reactor.flags import Flags
 from armi.reactor import grids
 from armi.bookkeeping import report
 
+LUMINANCE_WEIGHTS = numpy.array([0.3, 0.59, 0.11, 0.0])
+
 
 def colorGenerator(skippedColors=10):
     """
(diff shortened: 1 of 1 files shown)


youngmit (approved):
So good. Thanks!

PR #326, opened 2021-07-06
Change how assembly pitch is found

chris10mckenz:
Originally, would take the average of assembly pitches and return that
as the assembly pitch. Ran into a problem when using blueprints from a
Thermal Reactor and it did not like how originally it called 
"pitches = numpy.array([b.getPitch() for b in self.getBlocks(Flags.FUEL)])" 
but in that specific case, no flags were fuel so pitches would be empty. 
But now, since we have spatialGrids, just calling self.spatialGrid.pitch may do 
what we want as well and get over this no
blocks with a certain flag issue.

### DIFF
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -1518,38 +1518,17 @@ class Core(composites.Composite):
 
     def getAssemblyPitch(self):
         """
-        Find the representative assembly pitch for the whole core.
+        Find the assembly pitch for the whole core.
 
-        This loops over all fuel blocks to find the best pitch.
+        This returns the pitch according to the spatialGrid.
 
         Returns
         -------
-        avgPitch : float
-            The average pitch of fuel assems in cm.
+        pitch : float
+            The assembly pitch.
 
         """
-        pitches = numpy.array([b.getPitch() for b in self.getBlocks(Flags.FUEL)])
-
-        # keepdims assures we get array of length one for when there is only one pitch
-        # value (such as in the hex case), so the same code can be used.
-        keepDims = False if pitches.ndim > 1 else True
-
-        # axis=0 will return average of the horizontal and vertical pitch independently
-        # in the case of rectangular lattice.
-        avgPitch = pitches.mean(axis=0, keepdims=keepDims)
-
-        # Check that pitches are consistent
-        minPitch = pitches.min(axis=0, keepdims=keepDims)
-        maxPitch = pitches.max(axis=0, keepdims=keepDims)
-        minDeviatesTooMuch = any(diff > 1e-10 for diff in avgPitch - minPitch)
-        maxDeviatesTooMuch = any(diff > 1e-10 for diff in maxPitch - avgPitch)
-        if minDeviatesTooMuch or maxDeviatesTooMuch:
-            raise RuntimeError(
-                "Not all fuel blocks have the same pitch (in cm). "
-                "Min: {}, Mean: {}, Max: {}".format(minPitch, avgPitch, maxPitch)
-            )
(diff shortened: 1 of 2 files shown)


ntouran:
I think the previous implementation kind of was hoping to have relevance for cases that are coupled to the temperature calculations in a way that might adjust the pitch on a block-by-block level. But that was likely ill-conceived since it really requires fancy core mechanical models to handle properly. With this update we can still adjust the grid's pitch vs. T/H feedback in a T/H module and it will be consistent. 

My only suggestion might be to say something like this in the docstring just as a tip: 

```
To capture any thermal/hydraulic feedback of the core pitch, 
T/H modules will need to modify the grid pitch directly based 
on the relevant mechanical assumptions.
```

youngmit:
To further complicate matters, the Travis CI integration that we had set up to run the unit tests with each PR has stopped working with their .org -> .com migration. Running the tests manually shows one lonely failure on the Cartesian tests. Could you run the tests locally, get that sorted out and update? Thanks!

youngmit (changes requested):
From pytest
```
================================== FAILURES ===================================
_________________ CartesianReactorTests.test_getAssemblyPitch _________________
[gw10] win32 -- Python 3.7.2 \\albert\apps\dev\armi\venvs\develop37\Scripts\python.exe
  File "\\albert\home\myoung\tparmi-clean\framework\armi\reactor\tests\test_reactors.py", line 733, in test_getAssemblyPitch
    assert_equal(self.r.core.getAssemblyPitch(), [10.0, 16.0])
  File "\\albert\apps\dev\armi\venvs\develop37\lib\site-packages\numpy\testing\_private\utils.py", line 337, in assert_equal
    assert_equal(actual[k], desired[k], 'item=%r\n%s' % (k, err_msg), verbose)
  File "\\albert\apps\dev\armi\venvs\develop37\lib\site-packages\numpy\testing\_private\utils.py", line 421, in assert_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
item=0

 ACTUAL: 1.0
 DESIRED: 10.0
```

Could you address this? This isn't really so much an issue with your code per se, but my guess is something isn't being done at the Core/Reactor level to appropriately configure the grid in this case.

youngmit:
Poking around, it seems like what we probably want to do is give the Cartesian test reactor a pitch. The failing test is essentially the only one that uses it, so we can adopt the pitch that it wants as the default, but 10X16cm is kind of weird, and if someone ever does get around to writing more testing for the Cartesian stuff, I imagine that will get annoying. So we probably want the assembly pitch to be configurable from the caller's end. Something like this worked for me:
```diff
diff --git a/armi/tests/__init__.py b/armi/tests/__init__.py
index 2552456..71dba7d 100644
--- a/armi/tests/__init__.py
+++ b/armi/tests/__init__.py
@@ -54,14 +54,14 @@ def getEmptyHexReactor():
     return reactor


-def getEmptyCartesianReactor():
+def getEmptyCartesianReactor(pitch=(10.0, 16.0)):
     """Return an empty Cartesian reactor used in some tests."""
     from armi.reactor import blueprints

     bp = blueprints.Blueprints()
     reactor = reactors.Reactor("Reactor", bp)
     reactor.add(reactors.Core("Core"))
-    reactor.core.spatialGrid = grids.CartesianGrid.fromRectangle(1.0, 1.0)
+    reactor.core.spatialGrid = grids.CartesianGrid.fromRectangle(*pitch)
     reactor.core.spatialGrid.symmetry = geometry.SymmetryType(
         geometry.DomainType.QUARTER_CORE,
         geometry.BoundaryType.REFLECTIVE,
```

PR #328, opened 2021-07-07
Switch from Travis CI to Github actions for building unit tests

ntouran:


youngmit (approved):
Looks good! Failing tests are from recent changes that would have been caught if CI was working. I think I'll just pull the trigger

PR #329, opened 2021-07-08
Add ability to plot Block Diagrams

chris10mckenz:
Create a function plotBlockDiagram in plotting.py in order to plot block
diagrams when a spatial grid is on the block. In addition, the
_makeBlockPinPatches and _makeComponentPatch work to create component
patches as long as the block's grid type is supported.

For each components spatialLocator, it creates a patch for the
components shape and size at the location/locations described by the
locator and colors according to material name the component is made out
of. Also creates a legend saying what color is what material. At the
end, plotBlockDiagram saves the figure as an SVG.

### DIFF
--- a/armi/utils/plotting.py
+++ b/armi/utils/plotting.py
@@ -1168,14 +1168,6 @@ def _makeComponentPatch(component, position, cold):
         cold: boolean
             True if looking for dimension at cold temps
 
-        nSides: int
-            Either 4 or 6, for Thermal vs. Fast Reactors.
-
-        smallestInnerForDerived: int
-                                The DerivedShape is the backdrop of everything else,
-                                so to draw it currently, it uses the smallest innerPitch(for HexBlocks)
-                                or smallest widthInner for Cartesian.
-
     Return
     ------
         blockPatch: List
--- a/armi/utils/plotting.py
+++ b/armi/utils/plotting.py
@@ -32,6 +32,7 @@ from armi import runLog
 from armi.reactor.flags import Flags
 from armi.reactor import grids
 from armi.bookkeeping import report
+from armi.reactor.components.basicShapes import Hexagon, Rectangle, Square
 
 LUMINANCE_WEIGHTS = numpy.array([0.3, 0.59, 0.11, 0.0])
 
@@ -336,7 +337,7 @@ def plotFaceMap(
         colbar.ax.tick_params(labelsize=16)
 
     if legendMap is not None:
-        legend = _createFaceMapLegend(legendMap, collection)
+        legend = _createLegend(legendMap, collection)
 
     else:
         legend = None
@@ -411,7 +412,7 @@ def _makeAssemPatches(core):
     if isinstance(core.spatialGrid, grids.HexGrid):
         nSides = 6
     elif isinstance(core.spatialGrid, grids.ThetaRZGrid):
-        raise ValueError(
+        raise TypeError(
             "This plot function is not currently supported for ThetaRZGrid grids."
         )
     else:
@@ -479,8 +480,8 @@ def _setPlotValText(ax, texts, core, data, labels, labelFmt, fontSize, collectio
         texts.append(text)
 
 
-def _createFaceMapLegend(legendMap, collection, size=9):
-    """Make special assembly-legend for the assembly face map plot with assembly counts."""
+def _createLegend(legendMap, collection, size=9, shape=Hexagon):
+    """Make special legend for the assembly face map plot with assembly counts, and Block Diagrams."""
 
     class AssemblyLegend:
         """
@@ -503,14 +504,30 @@ def _createFaceMapLegend(legendMap, collection, size=9):
             normVal = collection.norm(index)
             cmap = collection.get_cmap()
             colorRgb = cmap(normVal)
-            patch = matplotlib.patches.RegularPolygon(
-                (x, y),
-                6,
-                height,
-                orientation=math.pi / 2.0,
-                facecolor=colorRgb,
-                transform=handlebox.get_transform(),
-            )
+            if shape == Hexagon:
+                patch = matplotlib.patches.RegularPolygon(
+                    (x, y),
+                    6,
+                    height,
+                    orientation=math.pi / 2.0,
+                    facecolor=colorRgb,
+                    transform=handlebox.get_transform(),
+                )
+            elif shape == Rectangle:
+                patch = matplotlib.patches.Rectangle(
+                    (x - height / 2, y - height / 2),
+                    height * 2,
+                    height,
+                    facecolor=colorRgb,
+                    transform=handlebox.get_transform(),
+                )
+            else:
+                patch = matplotlib.patches.Circle(
+                    (x, y),
+                    height,
+                    facecolor=colorRgb,
+                    transform=handlebox.get_transform(),
+                )
 
             luminance = numpy.array(colorRgb).dot(LUMINANCE_WEIGHTS)
             dark = luminance < 0.5
@@ -1048,7 +1065,7 @@ def makeHistogram(x, y):
     return xHistogram, yHistogram
 
 
-def _makeBlockPinPatches(block, temp):
+def _makeBlockPinPatches(block, cold):
     """Return lists of block component patches and corresponding data and names (which relates to material
     of the component for later plot-coloring/legend) for a single block.
 
@@ -1059,10 +1076,10 @@ def _makeBlockPinPatches(block, temp):
 
     Parameters
     ----------
-    block : block
+    block : Block
 
-    temp : boolean
-        cold = true, hot = false
+    cold : boolean
+        true for cold temps, hot = false
 
     Return
     ------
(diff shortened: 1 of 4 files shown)


armi/utils/plotting.py
```diff
@@ -1047,3 +1046,237 @@ def makeHistogram(x, y):
...
+    """
+    patches = []
+    data = []
+    name = []
+    if isinstance(block.spatialGrid, grids.HexGrid):
+        nSides = 6
+        smallestInnerPitch = block.getPitch()
+
+    elif isinstance(block.spatialGrid, grids.ThetaRZGrid):
+        raise ValueError(
```
youngmit, line 1086: This should actually be a TypeError (https://docs.python.org/3/library/exceptions.html#TypeError)

armi/utils/plotting.py
```diff
@@ -1047,3 +1046,237 @@ def makeHistogram(x, y):
...
+                    smallestInnerPitch = component.getDimension("ip", cold=temp)
+        else:
+            if "widthInner" in component.DIMENSION_NAMES:
+                if (
+                    component.getDimension("widthInner", cold=temp) < smallestInnerPitch
+                    and component.getDimension("widthInner", cold=temp) != 0.0
+                ):
+                    smallestInnerPitch = component.getDimension("widthInner", cold=temp)
+
+    for components in sorted(sortedComp):
```
youngmit, line 1113: Draw the DerivedShape up here, by using `b.getPitch(returnComp=True)`, then draw that component using the Material of the DerivedShape. You will probably have to somehow extract the DerivedShape from the collection of `sortedComp`onents

youngmit: You probably wont have to do the smallestInnerPitch search anymore as well.

armi/utils/plotting.py
```diff
@@ -1047,3 +1046,237 @@ def makeHistogram(x, y):
...
+    Takes in a block that must have a spatialGrid attached as well as a variable
+    which signifies whether the dimensions of the components are at hot or cold temps.
+    When cold is set to true, you would get the BOL cold temp dimensions.
+
+    Parameters
+    ----------
+    block : block
+
+    temp : boolean
+        cold = true, hot = false
```
youngmit, line 1065: Change the name to `cold`, which makes the meaning of the boolean obvious

armi/utils/plotting.py
```diff
@@ -1047,3 +1046,237 @@ def makeHistogram(x, y):
...
+    of the component for later plot-coloring/legend) for a single block.
+
+
+    Takes in a block that must have a spatialGrid attached as well as a variable
+    which signifies whether the dimensions of the components are at hot or cold temps.
+    When cold is set to true, you would get the BOL cold temp dimensions.
+
+    Parameters
+    ----------
+    block : block
```
youngmit, line 1062: capital-B Block

armi/utils/plotting.py
```diff
@@ -1047,3 +1046,237 @@ def makeHistogram(x, y):
...
+        temp: boolean
+            True if looking for dimension at cold temps
+
+        nSides: int
+            Either 4 or 6, for Thermal vs. Fast Reactors.
+
+        smallestInnerForDerived: int
+                                The DerivedShape is the backdrop of everything else,
+                                so to draw it currently, it uses the smallest innerPitch(for HexBlocks)
+                                or smallest widthInner for Cartesian.
```
youngmit, line 1153: Shouldn't need these if we handle the derived shape on its own

youngmit (changes requested):
See the comments about making the DerivedShape component special. We can just draw the pitch-defining component (intercoolant or whatever) using the DerivedShape material, then proceed to draw all other components on top of it and get a perfectly valid diagram without having to count sides or find the correct inner pitch to draw it with.

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+                "Shape of the pitch-defining element is not a Square or Hex it is {}, cannot plot for this type of block".format(
+                    comp.shape
+                )
+            )
+        patches.append(derivedPatch)
+        data.append(material)
+        name.append(cName)
+    sortedComp.sort(reverse=True)
+    for components in sortedComp:
+        if components.name is not "pitch" and not isinstance(components, DerivedShape):
```
youngmit, line 1138: Maybe just `.remove()` the DerivedShape component from `sortedComp[s]` up above so that you don't need to keep checking.

Also, why is the "pitch" case special?

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+            loc = components.spatialLocator
+            if not isinstance(loc, grids.MultiIndexLocation):
+                # make a single location a list to iterate.
+                loc = [
+                    loc,
+                ]
```
youngmit, line 1144: Is Black happy with this? The check passed, so i guess so, but seems strange.

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+    elif isinstance(block.spatialGrid, grids.ThetaRZGrid):
+        raise TypeError(
+            "This plot function is not currently supported for ThetaRZGrid grids."
+        )
+    else:
+        largestPitch, comp = block.getPitch(returnComp=True)
+        if block.getPitch()[0] != block.getPitch()[1]:
+            raise ValueError("Only works for blocks with equal length and width.")
+
+    sortedComp = [c for c in block]
```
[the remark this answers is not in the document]
youngmit: Also, could you be a bit more precise in pluralization throughout? Things that are lists, should be plural, single elements not. So `sortedComps` for this.

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+
+    derivedComponent = block.getComponentsOfShape(DerivedShape)
```
youngmit, line 1114: derivedComponents

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+            raise TypeError(
+                "Shape of the pitch-defining element is not a Square or Hex it is {}, cannot plot for this type of block".format(
+                    comp.shape
+                )
+            )
+        patches.append(derivedPatch)
+        data.append(material)
+        name.append(cName)
+    sortedComp.sort(reverse=True)
+    for components in sortedComp:
```
youngmit, line 1137: `component` singular.

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+        if components.name is not "pitch" and not isinstance(components, DerivedShape):
+            loc = components.spatialLocator
+            if not isinstance(loc, grids.MultiIndexLocation):
+                # make a single location a list to iterate.
+                loc = [
```
youngmit, line 1142: `locs`

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+                    loc,
+                ]
+            for local in loc:
```
youngmit, line 1145: Now you can do `for loc in locs:`

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+                x, y, _ = local.getLocalCoordinates()
+                # goes through each location
+                # want to place a patch at that location
+                blockPatch = _makeComponentPatch(components, (x, y), cold)
```
youngmit, line 1149: `blockPatches`

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+        position: tuple
+            (x, y) position
+
+        cold: boolean
+            True if looking for dimension at cold temps
+
+    Return
+    ------
+        blockPatch: List
+            A list that is either one element (a single patch), or 6 in the case of a Hexagon
```
youngmit, line 1174: Probably no need to go to that much detail about how many patches; just say something like "a list of Patch objects representing the component in the diagram."

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+            # Just make it a rectangle...
+            blockPatch = matplotlib.patches.Rectangle(
+                (
+                    x - component.getDimension("widthOuter", cold) / 2,
+                    y - component.getDimension("lengthOuter", cold) / 2,
+                ),
+                component.getDimension("widthOuter", cold),
+                component.getDimension("lengthOuter", cold),
+            )
+    if type(blockPatch) is not list:
```
youngmit, line 1296: pylint is probably complaining about this. Prefer `if isinstance(blockPatch, list):` when possible. The direct type comparison is very similar, but does not test true for sub-classes of `list`. Sometimes, but rarely, this is necessary, so it looks a bit strange to see if that logic isn't needed.

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+    """Given a Block with a spatial Grid, plot the diagram of
+    it with all of its components. (wire, duct, coolant, etc...)
+
+    Parameters
+    ----------
+    block : block object
+    cmapName : String
+        name of a colorMap to use for block colors
+    num : int
+        for making the file name (What number block are we plotting)
```
youngmit, line 1311: Take a full file name.

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+
+def plotBlockDiagram(block, cmapName, num, cold):
+    """Given a Block with a spatial Grid, plot the diagram of
+    it with all of its components. (wire, duct, coolant, etc...)
+
+    Parameters
+    ----------
+    block : block object
+    cmapName : String
+        name of a colorMap to use for block colors
```
youngmit, line 1309: Might be worth providing a default colormap.

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+    num : int
+        for making the file name (What number block are we plotting)
+    cold : boolean
+        true is for cold temps, hot is false.
+    """
+    from collections import defaultdict
+    import copy
```
youngmit, line 1316: Import these at the top of the module

armi/utils/plotting.py
```diff
@@ -1047,3 +1063,304 @@ def makeHistogram(x, y):
...
+
+    fig, ax = plt.subplots(figsize=(50, 50), dpi=100)
+
+    if block.spatialGrid is None:
+        return None
+    patches, data, name = _makeBlockPinPatches(block, cold)
```
youngmit, line 1322: `names`

youngmit (changes requested):
Starting to look really good! Just a smattering of mostly minor things

armi/utils/plotting.py
chris10mckenz: "pitch" is something that is considered a "dummy component for assembly sizing" in the Cartesian grid file I was using. It had a material called Void, but I just noticed Inner and Outer Width are equal for it, so I think I would be able to still draw it and not actually have it show up without checking. But then if I do not have some check, Void,  it's material that may not be used by any other component would still show up in the legend even if it wasn't actually present in the diagram. But would a "dummy component for assembly sizing" always have the same name, or not?

armi/utils/tests/test_plotting.py
```diff
@@ -88,6 +89,26 @@ def test_plotBlockFlux(self):
             os.remove("bList2.txt")  # secondarily created during the call.
             os.remove("bList2.png")  # created during the call.
 
+    def test_plotHexBlock(self):
+        o, r = test_reactors.loadTestReactor()
```
youngmit, line 93: Note that you have `cls.o` and `cls.r` at your disposal.

armi/utils/tests/test_plotting.py
```diff
@@ -88,6 +89,26 @@ def test_plotBlockFlux(self):
...
+        from armi import settings
+        from armi.reactor import blueprints, reactors
+
+        cs = settings.Settings(
+            os.path.join(TEST_ROOT, "tutorials", "c5g7-settings.yaml")
+        )
+        blueprint = blueprints.loadFromCs(cs)
+        r = reactors.factory(cs, blueprint)
+        for b in r.core.getBlocks():
+            plotting.plotBlockDiagram(b, "blockDiagram25.svg", True)
```
youngmit, line 109: Plotting all of the blocks in the core is pretty heavy handed. Since you have the blueprints, try iterating the blockDesigns on those, constructing the blocks, then plotting those.

youngmit (changes requested):

youngmit (changes requested):
One little thing in the C5G7 blueprints and this should be good to go

youngmit (approved):

PR #330, opened 2021-07-08
Add implementation of detachedCopy() to MultiIndexLocation

youngmit:
MIL has a different init signature than its base IndexLocation, so it
needed its own implementation of detachedCopy()

### DIFF
--- a/armi/bookkeeping/tests/test_databaseInterface.py
+++ b/armi/bookkeeping/tests/test_databaseInterface.py
@@ -34,6 +34,7 @@ from armi.cases import case
 from armi.utils import directoryChangers
 from armi import runLog
 from armi.reactor.tests import test_reactors
+from armi.reactor import grids
 
 
 def getSimpleDBOperator(cs):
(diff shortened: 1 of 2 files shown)


armi/bookkeeping/tests/test_databaseInterface.py
```diff
@@ -265,7 +266,14 @@ def test_readWritten(self):
...
                     self.assertEqual(c1.name, c2.name)
-                    assert_equal(c1.spatialLocator.indices, c2.spatialLocator.indices)
+                    if isinstance(c1.spatialLocator, grids.MultiIndexLocation):
+                        assert_equal(
+                            c1.spatialLocator.allIndices, c2.spatialLocator.allIndices
+                        )
+                    else:
+                        assert_equal(
+                            c1.spatialLocator.indices, c2.spatialLocator.indices
+                        )
```
ntouran, line 276: didn't  want to override `__eq__` on `spatialLocator` itself?

armi/reactor/grids.py
```diff
@@ -436,6 +436,11 @@ def __iter__(self):
     def __len__(self):
         return len(self._locations)
 
+    def detachedCopy(self):
+        loc = MultiIndexLocation(None)
+        loc.extend(self._locations)
+        return loc
```
ntouran, line 442: probably needs explanatory docstring?

ntouran (approved):

armi/bookkeeping/tests/test_databaseInterface.py
youngmit: There is an __eq__() impl, but it includes the grid itself...

armi/reactor/grids.py
youngmit: Nothing really to be gained from what's in the base class's docstring

PR #333, opened 2021-07-15
Fix issues with Grids and  multi-index locations

youngmit:
- Add an `associate()` method to LocationBase, allowing to bind a
   location to a new grid. MultiIndexLocation implements this
   appropriately
 - Remove `__hash__()` from MIL, since it can't support it. There are
   still some substitutability issues in Locations, which we should
   address at some point to cut down on how special a case MIL is.
 - Add code to Composite `__setstate__()` to ensure that spatialLocators
   of children are properly associated with their new parent grid
 - Teach the GridBlueprint to infer "through center" based on the number
   of things in the lattice being even or odd.

### DIFF
--- a/armi/reactor/grids.py
+++ b/armi/reactor/grids.py
@@ -813,7 +813,9 @@ class Grid:
             locators = [self[idx] for idx in ijk]
             val.extend(locators)
         else:
-            raise TypeError("Unsupported index type `{}` for `{}`".format(type(ijk), ijk))
+            raise TypeError(
+                "Unsupported index type `{}` for `{}`".format(type(ijk), ijk)
+            )
         return val
 
     def __len__(self):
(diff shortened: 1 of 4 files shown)


armi/reactor/grids.py
```diff
@@ -455,7 +474,11 @@ def pop(self, location: IndexLocation):
 
     @property
     def indices(self):
-        raise NotImplementedError
+        """
+        This should be at least symmetric with the way indices passed to __getitem__ on
+        Grid behaves.
```
ntouran, line 479: minor: should be in a Notes section

ntouran (approved):
looks good

armi/reactor/grids.py
youngmit: Good catch; i had meant when i started staging this stuff yesterday to expound on this a little, but forgot. Will improve

PR #336, opened 2021-07-19
Refactoring runLog to use logging standard library

theJollySin:
This PR is broken up into two commits:

1. Refactor `runLog.py` to use the Python standard library `logging`.
2. Allow module-specific logging (see `reactors.py` for an example).

I tried to keep this PR as minimal as possible, but touching anything in our logging has long fingers. So even without moving code to `context.py` this ended up being a 16-file PR. Changing `runLog.py` meant making significant changes to `mockRunLog.py`, which lead to all kinds of unit testing changes. 

I also found that testing logging changes in Python is exceptionally tedious, as PyTest intercepts *all* logging, for the purposes of display. So you can't use PyTest to test your logging functionality without changing the outputs: see Heisenberg.

I also moved the `prompt()` stuff out of `runLog.py` and into `settingsIO.py`.

### DIFF
--- a/armi/context.py
+++ b/armi/context.py
@@ -20,14 +20,15 @@ running. Things like the MPI environment, executing user, etc. live here. These 
 re-exported by the `armi` package, but live here so that import loops won't lead to
 as many issues.
 """
+from logging import DEBUG
 import datetime
+import enum
+import gc
 import getpass
 import os
-import time
-import sys
-import enum
 import shutil
-import gc
+import sys
+import time
 
 # h5py needs to be imported here, so that the disconnectAllHdfDBs() call that gets bound
 # to atexit below doesn't lead to a segfault on python exit. The Database3 module is
@@ -149,7 +150,7 @@ MPI_DISTRIBUTABLE = MPI_RANK == 0 and MPI_SIZE > 1
 
 _FAST_PATH = os.path.join(os.getcwd())
 """
-A directory available for high-performance I/O  
+A directory available for high-performance I/O
 
 .. warning:: This is not a constant and can change at runtime.
 """
@@ -222,7 +223,7 @@ def cleanTempDirs(olderThanDays=None):
     )  # pylint: disable=import-outside-toplevel # avoid cyclic import
 
     disconnectAllHdfDBs()
-    printMsg = runLog.getVerbosity() <= runLog.getLogVerbosityRank("debug")
+    printMsg = runLog.getVerbosity() <= DEBUG
     if _FAST_PATH_IS_TEMPORARY and os.path.exists(_FAST_PATH):
         if printMsg:
             print(
--- a/armi/context.py
+++ b/armi/context.py
@@ -295,3 +295,32 @@ def disconnectAllHdfDBs():
     h5dbs = [db for db in gc.get_objects() if isinstance(db, Database3)]
     for db in h5dbs:
         db.close()
+
+
+# ============ begin logging support ============
+
+OS_SECONDS_TIMEOUT = 5 * 60
+
+
+def createLogDir(mpiRank=1, logDir="logs"):
+    """A helper method to create the log directory"""
+    # let only do this from one thread
+    if mpiRank == 0:
+        if not os.path.exists(logDir):
+            os.makedirs(logDir)
+
+    # now all ranks should spin until they see the log directory
+    secondsWait = 0.5
+    loopCounter = 0
+    while not os.path.exists(logDir):
+        time.sleep(secondsWait)
+        loopCounter += 1
+        if loopCounter > (OS_SECONDS_TIMEOUT / secondsWait):
+            raise OSError("Was unable to create the log directory: {}".format(logDir))
+
+    # wait until all execution paths can see the log dir
+    if MPI_COMM is not None:
+        MPI_COMM.barrier()
+
+
+createLogDir(MPI_RANK)
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -26,6 +26,13 @@ The default way of using the ARMI runLog is:
     runLog.error('extra error info here')
     raise SomeException  # runLog.error() implies that the code will crash!
 
+You can interact with the logger in much the same way now by doing:
+
+.. code::
+
+    import logging
+    runLog = logging.getLogger('whatever')
+
 """
 from __future__ import print_function
 from glob import glob
@@ -39,7 +46,15 @@ import time
 from armi import context
 
 
-class RunLog:
+# global constants
+_WHITE_SPACE = " " * 6
+_ADD_LOG_METHOD_STR = """def {0}(self, message, *args, **kws):
+    if self.isEnabledFor({1}):
+        self._log({1}, message, args, **kws)
+logging.Logger.{0} = {0}"""
+
+
+class _RunLog:
     """
     Handles all the logging
     For the parent process, things are allowed to print to stdout and stderr,
@@ -64,16 +79,14 @@ class RunLog:
         """
         self._mpiRank = mpiRank
         self._verbosity = logging.INFO
-        self._singleMessageCounts = collections.defaultdict(lambda: 0)
-        self._singleWarningMessageCounts = collections.defaultdict(lambda: 0)
         self.initialErr = None
-        self.logger = None
-        self.stderrLogger = None
         self._logLevels = None
-        self._whitespace = " " * 6
+        self.logger = logging.getLogger("NULL")
+        self.logger.addHandler(logging.NullHandler())
+        self.stderrLogger = logging.getLogger("NULL2")
+        self.stderrLogger.addHandler(logging.NullHandler())
 
         self._setLogLevels()
-        self._createLogDir()
 
     def _setLogLevels(self):
         """Here we fill the logLevels dict with custom strings that depend on the MPI rank"""
@@ -83,7 +96,7 @@ class RunLog:
             [
                 ("debug", (logging.DEBUG, "[dbug{}] ".format(_rank))),
                 ("extra", (15, "[xtra{}] ".format(_rank))),
-                ("info", (logging.DEBUG, "[info{}] ".format(_rank))),
+                ("info", (logging.INFO, "[info{}] ".format(_rank))),
                 ("important", (25, "[impt{}] ".format(_rank))),
                 ("prompt", (27, "[prmt{}] ".format(_rank))),
                 ("warning", (logging.WARNING, "[warn{}] ".format(_rank))),
@@ -91,80 +104,72 @@ class RunLog:
                 ("header", (100, "".format(_rank))),
             ]
         )
-        self._whitespace = " " * len(max([l[1] for l in self._logLevels.values()]))
+        global _WHITE_SPACE
+        _WHITE_SPACE = " " * len(max([l[1] for l in self._logLevels.values()]))
 
         # modify the logging module strings for printing
-        for logValue, shortLogString in self._logLevels.values():
+        for longLogString, (logValue, shortLogString) in self._logLevels.items():
+            # add the log string name (upper and lower) to logging module
+            logging.addLevelName(logValue, shortLogString.upper())
             logging.addLevelName(logValue, shortLogString)
+
+            # ensure that we add any custom logging levels as constants to the module, e.g. logging.HEADER
+            try:
+                getattr(logging, longLogString.upper())
+            except AttributeError:
+                setattr(logging, longLogString.upper(), logValue)
+
+            # Add logging methods for our new custom levels: LOG.extra("message")
+            try:
+                getattr(logging, longLogString)
+            except AttributeError:
+                exec(_ADD_LOG_METHOD_STR.format(longLogString, logValue))
 
     def log(self, msgType, msg, single=False, label=None):
         """
-        Add formatting to a message and handle its singleness, if applicable.
-
         This is a wrapper around logger.log() that does most of the work and is
         used by all message passers (e.g. info, warning, etc.).
+
+        In this situation, we do the mangling needed to get the log level to the correct number.
+        And we do some custom string manipulation so we can handle de-duplicating warnings.
         """
-        # Skip writing the message if it is below the set verbosity
-        msgVerbosity = self._logLevels[msgType][0]
-        if msgVerbosity < self._verbosity:
-            return
+        # Determine the log level: users can optionally pass in custom strings ("debug")
+        msgLevel = msgType if isinstance(msgType, int) else self._logLevels[msgType][0]
 
-        # the message label is only used to determine unique for single-print warnings
-        if label is None:
-            label = msg
+        # If this is a special "don't duplicate me" string, we need to add that info to the msg temporarily
+        msg = str(msg)
 
-        # Skip writing the message if it is single-print warning
-        if single and self._msgHasAlreadyBeenEmitted(label, msgType):
-            return
+        # Do the actual logging
+        self.logger.log(msgLevel, msg, single=single, label=label)
 
-        # Do the actual logging, but add that custom indenting first
-        msg = self._cleanMsg(msg)
-        if self._mpiRank == 0:
-            print(self._logLevels[msgType][1] + msg)
-        else:
-            self.logger.write(msgVerbosity, msg)
+    def getDuplicatesFilter(self):
+        """
+        The top-level ARMI logger should have a no duplicates filter
+        If it exists, find it.
+        """
+        if not self.logger or not isinstance(self.logger, logging.Logger):
+            return None
 
-    def _cleanMsg(self, msg):
-        """Messages need to be strings, and tabbed if multi-line"""
-        return str(msg).rstrip().replace("\n", "\n" + self._whitespace)
-
-    def _msgHasAlreadyBeenEmitted(self, label, msgType=""):
-        """Return True if the count of the label is greater than 1."""
-        if msgType in ("warning", "critical"):
-            self._singleWarningMessageCounts[label] += 1
-            if self._singleWarningMessageCounts[label] > 1:
-                return True
-        else:
-            self._singleMessageCounts[label] += 1
-            if self._singleMessageCounts[label] > 1:
-                return True
-        return False
+        return self.logger.getDuplicatesFilter()
 
     def clearSingleWarnings(self):
         """Reset the single warned list so we get messages again."""
-        self._singleMessageCounts.clear()
+        dupsFilter = self.getDuplicatesFilter()
+        if dupsFilter:
+            dupsFilter.singleMessageCounts.clear()
 
     def warningReport(self):
         """Summarize all warnings for the run."""
-        info("----- Final Warning Count --------")
-        info("  {0:^10s}   {1:^25s}".format("COUNT", "LABEL"))
+        self.logger.warningReport()
 
-        # sort by labcollections.defaultdict(lambda: 1)
-        for label, count in sorted(
-            self._singleWarningMessageCounts.items(), key=operator.itemgetter(1)
-        ):
-            info("  {0:^10s}   {1:^25s}".format(str(count), str(label)))
-        info("------------------------------------")
-
-    def _getLogVerbosityRank(self, lvl):
+    def _getLogVerbosityRank(self, level):
         """Return integer verbosity rank given the string verbosity name."""
-        level = lvl.strip().lower()
-        if level not in self._logLevels:
+        try:
+            return self._logLevels[level][0]
+        except KeyError:
             log_strs = list(self._logLevels.keys())
             raise KeyError(
-                "{} is not a valid verbosity level. Choose from {}".format(
-                    level, log_strs
-                )
+                "{} is not a valid verbosity level: {}".format(level, log_strs)
             )
         return self._logLevels[level][0]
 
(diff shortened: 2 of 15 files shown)


theJollySin:
@youngmit @ntouran 

So, here is the refactor of `runLog.py` to use the Python standard module `logging`.   This also required changing the test-only subclasses in `mockRunLog.py`, so there are several testing changes outside the main file.

I like the changes in general, but I had to do something a bit ugly with `_logLevels` in the global region of `runLog.py`.  Basically, these strings need to be correct before we do basic config on the `logging` library. That basic config has to happen _real_ early in the run time or it gets stomped on and ignored.

Other than that, the only funny thing to note is that this logging change is only half of the functionality in `runLog.py`. The other half is intercepting all of the standard error messages. This intercepting could involve the `logging` module if we like. But it is already intercepting stderr error and pipping it directly to a file, which it does well and has nothing to do with `logging`.

I'm happy for any comments or thoughts.

armi/runLog.py
```diff
@@ -98,82 +108,79 @@ class Log:
...
+        self._mpiSize = mpiSize
+        self._verbosity = logging.INFO
+        self.name = "log"
         self._singleMessageCounts = collections.defaultdict(lambda: 0)
         self._singleWarningMessageCounts = collections.defaultdict(lambda: 0)
-        self._outputStream = None
+        self.loggerName = (
+            "parent" if self._mpiRank == 0 else "{:>03d}".format(self._mpiRank)
+        )
+        self.logger = logging.getLogger(self.loggerName)
```
drewejohnson, line 135: I've been wanting this feature to come in for a while and I'm very happy about this PR 🤩 

I do have one comment that I'd like to raise here: Embedding ARMI as the "root" logger name. The reason is two-fold

1. This allows other applications to grab the logger (if they so choose) with `logger.getLogger("ARMI")`
2. This embeds the string `ARMI` in the log message if someone tweaks their log config, e.g.,
```python
>>> import logging
>>> logging.basicConfig(level=logging.INFO, format="%(name)s::%(message)s")
>>> lg = logging.getLogger("test-app")
>>> lg.info("Hello world")
test-app::Hello world
```

It might be useful to have the application logger "name" for the simply reflect the solver application (e.g., `DIF3D` vs. `ARMI.physics.neutronics.dif3d`). Might be arguments for both (grepping for all messages that relate to physics vs. reporting, bookkeeping, etc.). The "dot separated" structure is intentional as it build a hierarchy of loggers

> Loggers that are further down in the hierarchical list are children of loggers higher up in the list. For example, given a logger with a name of foo, loggers with names of foo.bar, foo.bar.baz, and foo.bam are all descendants of foo. 

- https://docs.python.org/3/library/logging.html#logger-objects

and for `Logger.propagate` - https://docs.python.org/3/library/logging.html#logging.Logger.propagate

> If this attribute evaluates to true, events logged to this logger will be passed to the handlers of higher level (ancestor) loggers, in addition to any handlers attached to this logger. Messages are passed directly to the ancestor loggers’ handlers - neither the level nor filters of the ancestor loggers in question are considered.

If the goal is to have solver-specific loggers, an "approved" way to grab a logger would be great. An interface could do something like
```python
from armi.runLog import buildChildLog

d3dLog = buildChildLog("DIF3D")   
# or
d3dLog = buildChildLog("DIF3D", "physics.neutronics") # to log as armi.physics.neutronics.DIF3D
```

Kind of a rambling mess but I hope something useful can be found 😅

theJollySin: 1. I like `"ARMI"` as a better default than `"parent"`, sure.  That makes sense to me.
2. As far as having things more custom to the module like `"DIF3D"` as the logger name, we love that. The idea is this is the first PR to refactor `runLog.py` to use the `logging` module. Then the _next_ PR would refactor the logging more to allow for various module-specific logging options. The primary one being module-by-module logging levels. But I like the idea of each logger having their own name. (I might have to play with that to make sure it's still convenient with MPI runs a bit.)

I like both of those ideas a lot. I can go change the default logger name, unless someone objects.

armi/runLog.py
```diff
@@ -98,82 +108,79 @@ class Log:
     Abstract class that ARMI code calls to be rendered to some kind of log.
     """
 
-    def __init__(self, verbosity=50):  # pylint: disable=unused-argument
+    def __init__(self, mpiRank=0, mpiSize=1):
```
youngmit, line 111: What real function does having our own Log class really serve at this point? I know we are currently using it to handle concatenation of different rank's log streams, but this could really be handled just about anywhere, and probably just be a straightforward function. We may wish to keep an adapter class around to maintain the API that folks are used to (`warn`, `info`, `extra`, etc.), but I even view that as optional, since currently no one (I don't think, anyway) interacts directly with the `Log` class; they're calling the module-level functions instead.

youngmit (changes requested):
Oddly enough, I think this maybe retains _too_ much of what we had before. In my mind, we would ideally do away with the Log/PrintLog classes entirely, only necessarily preserving:
 - the module-level logging functions (`info()`, `warn()`, `extra()`, etc.)
 - Rank-aware file output (non-zero MPI ranks -> `logs/` directory)
 - The behavior (somewhere) to concatenate higher-rank logs at the end

armi/tests/test_runLog.py
```diff
@@ -43,6 +51,47 @@ def test_invalidSetVerbosityByString(self):
...
+            runLog.important("two two\n")
+            runLog.warning("three three three")
+            runLog.error("four four four four")
+
+        messages = [r.message for r in self._caplog.records]
+        assert len(messages) > 0
+        assert "one" in messages[0]
+        assert "two two" in messages[1]
+        assert "three three three" in messages[2]
+        assert "four four four four" in messages[3]
```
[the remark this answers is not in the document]
theJollySin: Ack! Sorry, yes, `self.assertX`.  I keep forgetting about those. I just finished a big push to move away from those at my last job and keep trying to remember we are using them here. I'll fix those.

(I will take a look at ditching the two log classes in a second. I thought to preserve them because I can't see if they are being used in the plugins. Give me a minute on an opinion there.)

armi/runLog.py
```diff
@@ -33,35 +33,45 @@
...
+        ("header", (100, "".format(_rank))),
     ]
 )
+_whitespace = " " * len(max([l[1] for l in _logLevels.values()]))
 
-_stderrName = "{0}.{1:04d}.stderr"
-_stdoutName = "{0}.{1:04d}.stdout"
+# modify the logging module strings for printing
+for logValue, shortLogString in _logLevels.values():
+    logging.addLevelName(logValue, shortLogString)
```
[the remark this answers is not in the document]
theJollySin: I will take a look at `_bootstrap`! Thanks for that.

armi/runLog.py
theJollySin: I was actually maintaining the class partially _because_ I couldn't guarantee no one else would be subclassing it.

Since we have two different subclasses of it in `mockRunLogs.py` it seemed like the safe assumption would be to preserve the API as it is now.  BUT if the API doesn't involve the class, that's definitely something.

I would be happy to simplify the code, but let me look at it. Right now the class serves two purposes:

1. Easy encapsulation of log concatenation.
2. It handles the piping of all `stderr` messages into files.

Let me look at it.

armi/tests/test_runLog.py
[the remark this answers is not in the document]
youngmit: you're moving away from the unittest assert methods? Why? Do tell!

armi/runLog.py
```diff
@@ -98,82 +108,79 @@ class Log:
...
-            if a msg verbosity is > this, it will be  emitted.
-            The default of 50 means only error messages will be emitted.
-            This usually gets adjusted by user settings quickly after instantiation.
+        mpiRank : int
+            If this is zero, we are in the parent process, otherwise child process.
+            The default of 0 means we assume the parent process.
+            This should not be adjusted after instantiation.
+        mpiSize : int
+            If this is one, we are in the parent process, with child process.
+            This should not be adjusted after instantiation.
```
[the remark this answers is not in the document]
theJollySin: Logger Name.  Sure, I have changed it.

It just seemed natural to have the name be dependent on the child process number because I was trying not to change two current features:

1. Log file names depend on child process (number)
2. Log strings depend on child process (number)

But, as I say, that's an easy change. Done.

armi/runLog.py
theJollySin: Item 1 complete

armi/runLog.py
youngmit: What API we land on is kind of wide open. The only thing that I think we _have_ to support (@drewejohnson let me know if you have anything that is different) are direct calls to `runLog.whichever()`. Beyond that, I'm considering this a clean sheet. Direct access to a `logging.Logger` to modules that opt in doesn't even seem unreasonable to me (so long as the root Logger is already handling streams and the like) and _something_ cats everything together when we turn out the lights.

armi/runLog.py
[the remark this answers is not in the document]
theJollySin: So, `_bootstrap.py` imports both `context.py` and `runLog.py` in their entirety.  So I don't want to be importing things from `_boostrap.py` into `runLog.py`. That's circular.

I *do* like the idea of having something less ugly like `configureLogging()`, even if all it does is fix these `logLevels` strings.

theJollySin:
Looking at this again, I'm not sure the `prompt()` method belongs in `runLog.py`.  It's not related to logging in any way that I can see. And it's only used in `settingsValidation.py` and `settingsIO.py`.  On balance, I would prefer to move `prompt()` out of the logging area.

armi/tests/test_runLog.py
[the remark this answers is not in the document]
theJollySin: Oh, sorry, there was no exciting reason. There were some tools in pytest we wanted to use that didn't play well with `unittest`, so we dropped all usages of `unittest`. That's all.  

Pytest works fine with plain `assert`s, but the `self.assertX` comes from the `unittest.TestCase` class.

armi/runLog.py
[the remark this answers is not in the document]
youngmit: If bootstrap imports runLog, what would be the issue with:
```python
import runLog
#...
runLog.configureLogging()
```

It seems silly/academic (and to a point, it kind of is), but it at least pushes us in the direction of "all of the infernal, unholy import-time magic happens [here]"

armi/runLog.py
theJollySin: @youngmit  @ntouran   Okay, the latest version of the PR allows for people to have module-specific loggers.  Please see `armi/reactor/reactors.py` for an example of the usage.

armi/runLog.py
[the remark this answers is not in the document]
theJollySin: I found a few issues trying to export much of the logging logic from `runLog.py` to `context.py` and `_bootstrap.py`.  It turns out we initialize our area significantly differently in runs versus in pytest.  So, there is some pytest env setup code in the repo that needs to change.

Also, some of the plugins are calling code I wanted to move out of `runLog.py` directly, so we'd have to at least keep stubs there.

I am, however, now fully convinced of Mitch's point of view that we need to move stuff out of `operatorMPI.py` and into `context.py`.  I'd like to have a conversation with him about what I found, and any specifics he has, because I think that is enough material for it's own PR.

armi/runLog.py
theJollySin: Take a look and tell me what you think. We are definitely using the `logging` module far more now.  Which was necessary to get the module-specific logging working.

theJollySin: In the end, I moved as much logging logic out of our old logging class (`PrintLog`) as possible, and into a `logging.Logger` subclass.  I only had to keep something like the old `PrintLog` class in existence because it defines the API that people are using elsewhere in the code.

armi/runLog.py
```diff
@@ -26,188 +26,157 @@
...
+# global constants
+_SINGLE_STR = "|SINGLE|"
+_WHITE_SPACE = " " * 6
+_ADD_LOG_METHOD_STR = """def {0}(self, message, *args, **kws):
+    if self.isEnabledFor({1}):
+        self._log({1}, message, args, **kws)
+logging.Logger.{0} = {0}"""
+
+
+class RunLog:
```
youngmit, line 58: So this class is mostly responsible for standing up and tearing down the logging stuff. It's used by the module-level logging functions (old API), but modules that wish to opt into the `logging.getLogger()` approach (new API) seem not to need any special knowledge of the `RunLog` class. Is this accurate? Should this be private?

armi/runLog.py
```diff
@@ -26,188 +26,157 @@
...
 
 
-class Log:
+# global constants
+_SINGLE_STR = "|SINGLE|"
+_WHITE_SPACE = " " * 6
+_ADD_LOG_METHOD_STR = """def {0}(self, message, *args, **kws):
+    if self.isEnabledFor({1}):
+        self._log({1}, message, args, **kws)
+logging.Logger.{0} = {0}"""
```
youngmit, line 55: Gnarly 🏂

armi/runLog.py
```diff
@@ -226,265 +195,346 @@ def setVerbosity(self, levelInput):
...
+        except IOError as ee:
+            warning("Failed to concatenate logs due to IOError.")
+            error(ee)
+    else:
+        if LOG.stderrLogger:
+            _ = [h.close() for h in LOG.stderrLogger.logger.handlers]
+            LOG.stderrLogger = None
+        if LOG.logger:
+            _ = [h.close() for h in LOG.logger.logger.handlers]
+            LOG.logger = None
```
youngmit, line 261: Do we not need to make sure that all of the >0 ranks have closed their handlers before rank 0 attempts to concatenate? This may require proper MPI comms...

armi/runLog.py
```diff
@@ -226,265 +195,346 @@ def setVerbosity(self, levelInput):
...
+    """
+    Important logging filter
+
+    * allow users to turn off duplicate warnings
+        * This is done via some special string manipulation
+    * handles special indentation rules for our logs
+    """
+
+    singleMessageCounts = {}
+    singleWarningMessageCounts = {}
```
youngmit, line 382: Is there a reason why these are stored as class attributes? If so, it might warrant a comment explaining why

armi/runLog.py
```diff
@@ -226,265 +195,346 @@ def setVerbosity(self, levelInput):
...
+    2. Piping stderr to a log file
+    """
+
+    FMT = "%(levelname)s%(message)s"
+
+    def __init__(self, *args, **kwargs):
+        # optionally, the user can pass in the MPI_RANK
+        if "mpiRank" in kwargs:
+            mpiRank = int(kwargs["mpiRank"])
+            del kwargs["mpiRank"]
```
youngmit, line 422: Try `mpiRank = int(kwargs.pop("mpiRank", context.MPI_RANK))`

armi/runLog.py
```diff
@@ -226,265 +195,346 @@ def setVerbosity(self, levelInput):
...
+        self.addHandler(h)
 
+    def _log(self, *args, **kwargs):
+        """wrapper around the standard library Logger._log() method
 
-def warningReport():
-    LOG.warningReport()
+        The primary goal here is to allow us to support the deduplication of warnings.
+        NOTE: All of the *args and **kwargs logic here are mandatory, as the standard library implementation of this
+        method has been changing the number of kwargs between Python v3.4 and v3.9.
```
youngmit, line 448: Do we mostly just need this as a trap door out of the `single` filter? A perhaps simpler alternative could be the `extra` dict , which can be passed to the various `logging` functions to staple things to the log record's `__dict__`, which can then be used in the `Filter`. So when calling the module-global `runLog.log()`, it can do something like:
```python
    def log(self, msgType, msg, single=False, label=None):
        ...
        # Do the actual logging
        self.logger.log(msgLevel, msg, extra={"single": single})
```

Then on the Filter side, we would peek at the record with `getattr()` to access `single` and handle accordingly.

Thoughts?

armi/runLog.py
```diff
@@ -226,265 +195,346 @@ def setVerbosity(self, levelInput):
...
+        self.handlers.clear()
+        del self
+
+    def getDuplicatesFilter(self):
+        """This object should have a no-duplicates filter. If it exists, find it."""
+        for f in self.filters:
+            if isinstance(f, DeduplicationFilter):
+                return f
+
+        return None
```
youngmit, line 501: If we propagate the `single` argument through `extra`, would we still need this?

youngmit (changes requested):
This looks so good!

I just have a couple of questions throughout, and wonder if we can simplify the logic in and around the DeduplicationFilter with the `extra` strategy (described in https://docs.python.org/3/library/logging.html#logging.Logger.debug)

armi/context.py
```diff
@@ -294,3 +295,27 @@ def disconnectAllHdfDBs():
     h5dbs = [db for db in gc.get_objects() if isinstance(db, Database3)]
     for db in h5dbs:
         db.close()
+
+
+# ============ begin logging support ============
+
+
+def createLogDir(mpiRank=1, log_dir="logs"):
```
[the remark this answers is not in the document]
theJollySin: > `logDir`

Sorry, my last project had the opposite naming rules.  I will see if I can find a linter to help me keep more consistent. I just found another one of these in the new unit test file for `context.py`. Fixing now.

pytest.ini
```diff
@@ -1,7 +1,7 @@
 [pytest]
 python_files=test_*.py
 python_functions=nothing matches this pattern
-addopts = --durations=30 --tb=native
+addopts = --durations=30 --tb=native -n 1 --dist=loadfile
```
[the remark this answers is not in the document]
theJollySin: That's an interesting point.  The `setup.py` requires `xdist`, so I expect it to be there.  Nick also said that it is a requirement of the test env, so I would be okay expecting it.

Because of how PyTest intercepts logging, it's pretty hard to test safely without:
 
1. Using `pytest-xdist` over `pytest-parallel` (I just go lucky there).
2. And forcing `--dist=loadfile`, which has to be done at the `pytest.ini` level, not test-by-test.

I suppose I felt I need the new `reactors.py` test to prove the new system is working. If we don't feel comfortable with this change right here... I guess my suggestion would be to delete that new `reactors.py` test. 

Which path do you like better: (a) keep this line, or (b) remove the one test that demands it?

armi/runLog.py
theJollySin: That's an interesting point.  Yeah, we could make that class private.  The only place it is used is in `armi/test/mockRunLogs.py`, which subclasses `RunLog` for testing purposes. 

I guess it depends on how hard-line you want to be about following the "don't use a private class" rule in `mockRunLogs.py`.  If you don't mind importing the private class there, we can do it.  Your call.

pytest.ini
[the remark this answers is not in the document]
youngmit: Oh, sorry i didnt know that we had `xdist` in the `dev` extra. Disregard!

armi/runLog.py
theJollySin: Well, yeah. Because people can add more `Filter` objects to the `Logger`.  That's part of the API, I see no reason to stop people from doing that.

armi/runLog.py
theJollySin: Okay, I'll need your opinion on this one.

All of these `close()` statements are called inside `operatorMPI.py`: that's where all the MPI comms are being done right now.  (As far as I can tell.)

The logic inside `OperatorMPI.operate()` is clearly setting up the run so the parent process finishes at earliest 1 second after all the child processes are finished.

So... in my mind, this method is safe. Thoughts?

The problem I have is I don't want to move the cross-process communication into the logger code, that doesn't seem like the right place for it. But you haven't given me a lot of confidence in `operatorMPI.py`.

armi/context.py
```diff
@@ -294,3 +295,27 @@ def disconnectAllHdfDBs():
...
+    # make the directory
+    if not os.path.exists(log_dir):
+        try:
+            os.makedirs(log_dir)
+        except FileExistsError:
+            pass
+
+    # stall until it shows up in file system (SMB caching issue?)
+    while not os.path.exists(log_dir):
+        time.sleep(0.5)
```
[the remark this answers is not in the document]
theJollySin: Agreed. And I threw in the hard timeout as well. Good call.

armi/runLog.py
theJollySin: Fixed

armi/runLog.py
theJollySin: Sure, changed.

armi/runLog.py
theJollySin: Yeah, I like your idea of using `extra`.  It's much cleaner.

Nice, man.

theJollySin:
@youngmit  I updated the PR with 5 changes. Tell me what you think!

armi/tests/mockRunLogs.py
```diff
@@ -44,10 +47,55 @@ def __enter__(self):
...
+            label = msg
+
+        # Skip writing the message if it is below the set verbosity
+        msgVerbosity = self._logLevels[msgType][0]  # pylint: disable=protected-access
+        if msgVerbosity < self._verbosity:
+            return
+
+        # Skip writing the message if it is single-print warning
+        if single and self._msgHasAlreadyBeenEmitted(label, msgType):
+            return
```
youngmit, line 69: Is this handled by the dedup Filter now? Happy with this if there is a good reason to have it for testing purposes, but seems like we'd want to keep this as close to the base impl as possible.

youngmit (commented):
With the new deduplication Filter, can we clean up the mock runlog stuff some?

armi/tests/mockRunLogs.py
theJollySin: Naw.  Since the `MockRunLogs.BufferLog` class doesn't actually _use_ the `logging.Logger` objects, it also doesn't use the `DedupFilter`. It uses plain string streams.  

The old `runLog` system wrote text files. Well, those were annoying to parse for testing, so `BufferLog` existed to skip the hard files and write to a string buffer, which could be used much more easily for testing.

The new `runLog` system writes using the `logging` module, which presents it's own problems because PyTest automatically intercepts all `logging` statements.  So now this class is used to bypass Python's standard library `logging` module, and write directly to a in-memory string stream, so we can test against that.

If I were to change anything, I guess I would want to refactor `_RunLog` to make things like this `BufferLog` class shorter/easier to write.

youngmit: sweet, thanks for the context

youngmit (approved):

PR #339, opened 2021-07-22
Removing instances of internal paths.

jakehader:
Removing referenced internal path locations that are not relevant to the broader ARMI framework development and user community.

### DIFF
--- a/armi/settings/tests/old_xml_settings_input.xml
+++ b/armi/settings/tests/old_xml_settings_input.xml
@@ -2,7 +2,6 @@
 
 <settings version="v6.1.5">
 	<aclpMaterial value="SS316" />
-	<armiLocation value="\\albert\apps\dev\armi\master" />
 	<axialExpansion value="True" />
 	<bondRemoval value="True" />
 	<branchVerbosity value="extra" />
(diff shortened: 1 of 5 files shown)


armi/settings/tests/old_xml_settings_input.xml
```diff
@@ -2,7 +2,6 @@
 
 <settings version="v6.1.5">
 	<aclpMaterial value="SS316" />
-	<armiLocation value="\\albert\apps\dev\armi\master" />
 	<axialExpansion value="True" />
 	<bondRemoval value="True" />
 	<branchVerbosity value="extra" />
```
youngmit, line 7: Any reason we can't just delete this? Are any tests still using it? I'd assume not, since it contains settings not defined by ARMI....

youngmit (approved):
This looks good; might be worth deleting the old XML settings file, if it's not being used in the first place.

armi/settings/tests/old_xml_settings_input.xml
jakehader: There is a unit test that still uses this. It is curious that the conversion still worked with these settings there.

PR #340, opened 2021-07-23
Relaxing XS settings validation failure when assigning file location and geometry inputs.

jakehader:
Relaxing cross section settings restrictions so that the case does not fail if the XS are overspecified with setting the file location
and the geometry information.


ntouran:
Can you explain situations where this may be desirable, or is it just something best left up to the app developers?

ntouran (approved):
I like the idea of not crashing but still giving a warning when people have minor inconsistencies. Presumably the file location takes precedence? Might be worth saying that in the warning. 

geometry inputs being ignored!

PR #341, opened 2021-07-23
Add splash text on configure() when in live interpreter

youngmit:
Users feel reassured that stuff is working correctly when they manually configure() ARMI and get a nice splash screen in the terminal. This will do just that when running within an live Python interpreter. Otherwise, we still use the logic in `EntryPoint` to determine whether we should emit splash text.


jakehader:
🔥

jakehader (approved):

PR #343, opened 2021-07-24
Fix to through center symmetry check within the GridBlueprint class 

jakehader:
Applying a fix to skip the "through center" Cartesian symmetry check if no grid contents are provided in the `GridBlueprint` class initialization.


jakehader:
Fixed:

image

PR #344, opened 2021-07-24
Fix copying interface input files with wildcard

jakehader:
An error occured within the `copyInterfaceInputs` function when a file path was provided with a wildcard.

In this case, any file checks using pathlib.Path would crash with an OSError. To circumvent this issue, the path checks are wrapped
in a try/except so that glob function may still be used to copy all files that end up meeting the wildcard criteria.

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -760,10 +760,16 @@ def copyInterfaceInputs(
                     # looks like an extant, absolute path; no need to do anything
                     pass
                 else:
-                    if not (path.exists() and path.is_file()):
-                        runLog.extra(
-                            f"Input file `{f}` not found. Checking for file at path `{sourceDirPath}`"
-                        )
+                    # An OSError can occur if a wildcard is included in the file name so
+                    # this is wrapped in a try/except to circumvent instances where an
+                    # interface requests to copy multiple files based on some prefix/suffix.
+                    try:
+                        if not (path.exists() and path.is_file()):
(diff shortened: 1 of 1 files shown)


jakehader:
This error occured when an interface specifically requested `/path/to/ISOTXS*`.

armi/cases/case.py
```diff
@@ -760,10 +760,16 @@ def copyInterfaceInputs(
...
+                    # An OSError can occur if a wildcard is included in the file name so
+                    # this is wrapped in a try/except to circumvent instances where an
+                    # interface requests to copy multiple files based on some prefix/suffix.
+                    try:
+                        if not (path.exists() and path.is_file()):
+                            runLog.extra(
+                                f"Input file `{f}` not found. Checking for file at path `{sourceDirPath}`"
+                            )
+                    except OSError:
+                        pass
```
youngmit, line 772: We should still *try* to resolve wildcards. I know we were having case-sensitivity issues with pathlib glob, but what about the raw `glob` library?

youngmit (changes requested):
I think we need to keep supporting wildcards. Per your example case, does this even do what we want for `ISOTXS*`?

armi/cases/case.py
jakehader: They get resolved below with glob but this is only for a run log statement. We could remove this but it might be helpful in some cases?

youngmit (approved):

PR #345, opened 2021-07-27
Doc errors

keckler:
Fixed various grammatical and typographical errors that I found while going through the tutorial docs. 

The only change more complicated than simple typos is the addition of the metadata setting to the SFR tutorial. Without this addition, the input that a user builds by following the tutorial will not run. I added a step in the tutorial to put the needed metadata entry into their settings file and added a yaml anchor to the stored input file so that the code renders in the docs.

### DIFF
--- a/doc/tutorials/walkthrough_inputs.rst
+++ b/doc/tutorials/walkthrough_inputs.rst
@@ -353,6 +353,13 @@ Now we need to specify some **settings** that define fundamental reactor
 parameters, as well as modeling approximation options. For this, we make a
 **settings file**, called ``anl-afci-177.yaml``.
 
+First we need to supply some metadata.
+
+.. literalinclude:: ../../armi/tests/tutorials/anl-afci-177.yaml
+    :language: yaml
+    :start-after: begin-metadata
+    :end-before: begin-settings-1
+
 The thermal power in this reference is 1000 MWt. The thermal efficiency isn't
 specified, so let's assume 0.38. From Table 4.8, the cycle length is 370 EFPD.
 Let's also assume a 0.90 capacity factor which will gives full cycles of 411.1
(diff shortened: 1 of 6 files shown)


doc/tutorials/walkthrough_inputs.rst
```diff
@@ -353,6 +353,13 @@ Now we need to specify some **settings** that define fundamental reactor
 parameters, as well as modeling approximation options. For this, we make a
 **settings file**, called ``anl-afci-177.yaml``.
 
+First we need to supply some metadata.
+
+.. literalinclude:: ../../armi/tests/tutorials/anl-afci-177.yaml
+    :language: yaml
+    :start-after: begin-metadata
+    :end-before: begin-settings
+
```
jakehader, line 362: If you try to delete this metadata section will the settings load or is this actually required as a part of the settings file?

jakehader (commented):
Hi Chris,

Please see my comment and make sure to sign the CLA above.

doc/tutorials/walkthrough_inputs.rst
keckler: For me, ARMI will not run without the metadata entry. 

image

Once I put the metadata entry in there, it runs with no problems. Do you think it should not be required? If so, perhaps we should check on another machine as well.

jakehader: Okay, thanks! I forgot about this section in the settings being required so this helps.

jakehader:
@keckler - Looks good! Please sign the license/cla agreement and then I can approve and merge in.

keckler:
It is signed and should be ready now.

jakehader (commented):
Approved.

Issue #332, opened 2021-07-14
ASCII maps for cartesian always behave as though they are reading for quarter-symmetry


youngmit:
Consider this grid blueprint from the C5G7 example:
```
    UO2 grid:
        symmetry: full
        geom: cartesian
        lattice pitch:
            x: 1.26
            y: 1.26
        lattice map: |
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  GT U  U  GT U  U  GT U  U  U  U  U
            U  U  U  GT U  U  U  U  U  U  U  U  U  GT U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  GT U  U  GT U  U  GT U  U  GT U  U  GT U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  GT U  U  GT U  U  FC U  U  GT U  U  GT U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  GT U  U  GT U  U  GT U  U  GT U  U  GT U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  GT U  U  U  U  U  U  U  U  U  GT U  U  U
            U  U  U  U  U  GT U  U  GT U  U  GT U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
```

When the contents of `lattice map` are read in, the `asciimaps` code is treating the bottom-left `U` as (0,0), and placing all of the objects in the first quadrant. This is what you would expect in a quarter-symmetric case. However in a full-symmetry case, one would expect the (0,0) to lie on the `FC` in the center. We need to do a symmetry-aware offset when reading `lattice maps:`

jakehader:
@youngmit, we saw this in a benchmark model of ZPPR as well within the grid editor.

jakehader:
Closing since this was fixed with #346.

PR #346, opened 2021-07-27
Fix centering of full-symmetry cartesian latticeMaps

youngmit:
Asciimaps assume everything starts at (0,0), so grid blueprints need to
offset indices when reading in full geometry inputs. Also, when writing
asciimaps, gridContents need to be offset in the other direction for
things to work out correctly.

This also fixes a bug related to asciimaps writing cartesian maps at
all.

### DIFF
--- a/armi/reactor/blueprints/tests/test_gridBlueprints.py
+++ b/armi/reactor/blueprints/tests/test_gridBlueprints.py
@@ -283,6 +283,8 @@ class TestGridBlueprintsSection(unittest.TestCase):
         self.assertEqual(gridDesign4.gridContents[-1, -1], "2")
         self.assertEqual(gridDesign4.gridContents[2, 2], "2")
         self.assertEqual(gridDesign4.gridContents[-3, -3], "1")
+        with self.assertRaises(KeyError):
+            self.assertEqual(gridDesign4.gridContents[-4, -3], "1")
 
 
 class TestRZTGridBlueprint(unittest.TestCase):
--- a/armi/reactor/blueprints/tests/test_gridBlueprints.py
+++ b/armi/reactor/blueprints/tests/test_gridBlueprints.py
@@ -57,6 +57,27 @@ sfp:
         2 1 3 1 2
         2 3 1 1 2
         2 2 2 2 2
+
+sfp quarter:
+    geom: cartesian
+    symmetry: quarter through center assembly
+    lattice map: |
+        2 2 2 2 2
+        2 1 1 1 2
+        2 1 3 1 2
+        2 3 1 1 2
+        2 2 2 2 2
+
+sfp even:
+    geom: cartesian
+    symmetry: full
+    lattice map: |
+        1 2 2 2 2 2
+        1 2 1 1 1 2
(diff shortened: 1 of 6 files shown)


youngmit:
Fixes #332

jakehader:
The code changes look good. Can you post a screenshot of the before and after for the full core Cartesian case with the `gridEditor` just to show the differences?

armi/reactor/blueprints/tests/test_gridBlueprints.py
```diff
@@ -242,7 +242,9 @@ def test_simple_read(self):
 
         gridDesign2 = self.grids["sfp"]
         _grid = gridDesign2.construct()
-        self.assertEqual(gridDesign2.gridContents[1, 1], "3")
+        self.assertEqual(gridDesign2.gridContents[1, 1], "1")
+        self.assertEqual(gridDesign2.gridContents[0, 0], "3")
+        self.assertEqual(gridDesign2.gridContents[-1, -1], "3")
```
jakehader, line 247: Can you add unit tests to cover the following scenarios:

- Cartesian, Full Core, Even
- Cartesian, Quarter Core, Odd
- Cartesian, Quarter Core, Even

The SFP test is nice, but let's make sure that we have the other variations tested too.

jakehader (changes requested):
Can we improve some of the `grids` documentation to include the images that you provided me with where the origin points are for the different grid types and symmetries? I think that this will be very useful going forward as a visual aid for users and developers.

youngmit:
Before:
image


After:
image

youngmit:
Also, switching to 1/4 symmetry in the blueprints works as expected:
image

armi/reactor/blueprints/tests/test_gridBlueprints.py
youngmit: Added quarter and even cases

armi/reactor/blueprints/tests/test_gridBlueprints.py
```diff
@@ -240,12 +261,29 @@ def test_simple_read(self):
...
+        self.assertEqual(gridDesign3.gridContents[3, 3], "1")
+
+        # Cartesian full, even/odd hybrid
+        gridDesign4 = self.grids["sfp even"]
+        _grid = gridDesign4.construct()
+        self.assertEqual(gridDesign4.gridContents[0, 0], "4")
+        self.assertEqual(gridDesign4.gridContents[-1, -1], "2")
+        self.assertEqual(gridDesign4.gridContents[2, 2], "2")
+        self.assertEqual(gridDesign4.gridContents[-3, -3], "1")
+
```
jakehader, line 286: One last comment, can you add a check for grid contents that are outside the map? I assume these should resolve to None?

jakehader (changes requested):

armi/reactor/blueprints/tests/test_gridBlueprints.py
youngmit: done

jakehader (approved):

PR #350, opened 2021-07-28
Add a setting to control auto-creation of Block grids

youngmit:
These rely pretty heavily on MultiIndexLocation, which still isn't fully
supported across the code. Apps can override the default value of the
setting to opt out of the feature until it is more mature.


jakehader (approved):
Thanks for this feature! The Multi-Index Locations need some more work before they can be universally accepted and used by applications.

PR #353, opened 2021-07-29
Fix links to parameter definition tables in user docs

keckler:
Current links to the pages containing the parameter definitions in the user docs
are not consistent. Some of them go to the tabular listings, while others link
to the API documentation. These links are updated so that all of them go to the
tabular listings.

In addition, a link was missing for the Core Parameters table. This link has
been added.

One exception to the consistency among links is that no page currently exists
with a comprehensive tabulation of the Component Parameters. It will take a good
amount of work to generate such a page, so for now this link points to a partial
tabulation elsewhere in the existing docs. This should be changed in the future,
and this need has been documented in an issue on GitHub.

### DIFF
--- a/doc/user/inputs/blueprints.rst
+++ b/doc/user/inputs/blueprints.rst
@@ -156,6 +156,7 @@ id
 od
     Outer diameter (in cm).
 
+.. _componentTypes:
 
 Component Types
 ---------------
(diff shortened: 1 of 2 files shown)


jakehader (approved):
Thanks for the updates @keckler. Let me know if you're still working on other doc changes or if I can merge in.

keckler:
I am working on another change to the docs, but it is really unrelated and I would say it belongs in its own commit.

jakehader:
Sounds great, i'll merge now.

PR #356, opened 2021-07-30
Increase test coverage of assemblies code

davis8dd:


armi/reactor/tests/test_assemblies.py
```diff
@@ -1137,6 +1137,44 @@ def test_carestianCoordinates(self):
...
+
+    def test_averagePlenumTemperature(self):
+        """
+        Test an assembly's average plenum temperature with a single block outlet.
+        """
+        averagePlenumTemp = 42.0
+        tempParam = (
+            assemblies.assemblyParameters.getAssemblyParameterDefinitions().__getitem__(
+                'THcoolantOutletT'
+                )
```
youngmit, line 1167: Shouldn't need to do this, since you can just index into `.p` below with `"THcoolantOutletT"`

armi/reactor/tests/test_assemblies.py
```diff
@@ -1137,6 +1137,44 @@ def test_carestianCoordinates(self):
...
+            )
+        plenumBlock = makeTestAssembly(
+            1, 2, grids.CartesianGrid.fromRectangle(1.0, 1.0))
+        plenumBlock.setType("plenum", Flags.PLENUM)
+        plenumBlock.p[tempParam.name] = averagePlenumTemp
```
youngmit, line 1172: Even more magical, `ParameterCollection`s have a fancy `__getattr__` implementation, so that you can do `plenumBlock.p.THcoolantOutletT`

armi/reactor/tests/test_assemblies.py
```diff
@@ -1137,6 +1137,44 @@ def test_carestianCoordinates(self):
...
+        """
+        Test the volume of a pin in the assembly's plenum.
+        """
+        assembly = assemblies.Assembly("dummy")
+        self.assertEqual(0.0, assembly.getPinPlenumVolumeInCubicMeters())
+
+        pinPlenumVolume = 7.853981633974482e-06
+        testAssemblies = buildTestAssemblies()
+        plenumBlock = makeTestAssembly(
+            1, 2, grids.CartesianGrid.fromRectangle(2.0, 2.0))
```
youngmit, line 1150: I think `plenumBlock` in here is actually an `Assembly`, rather than a block. Get one of its child `Block`s by indexing into it.

armi/reactor/tests/test_assemblies.py
```diff
@@ -1137,6 +1137,44 @@ def test_carestianCoordinates(self):
...
+        plenumBlock.setType("plenum", Flags.PLENUM)
+        self.Assembly.setBlockMesh([10.0, 20.0, 30.0], conserveMassFlag="auto")
+        plenumBlock.append(testAssemblies[0])
```
youngmit, line 1153: This is pretty interesting; you are appending an assembly to another assembly. I think lots of stuff in the ARMI composite model will sort of "work," but might do some surprising things. Assemblies expect to be collections of Blocks, so another way to go might be to just construct one in a similar fashion as is done in `buildTestAssemblies()`, except with some plenum components, then append that to `testAssemblies[0]` and use it for the test. An example of what a typical plenum block might look like can be found here: https://github.com/terrapower/armi/blob/5700728b129ef87ecbf01f4fef2fa82f582a92ac/armi/tests/tutorials/anl-afci-177-blueprints.yaml#L163

youngmit (changes requested):
First off, thanks for the contribution!

For the most part this looks great, but I have a couple of questions/suggestions pertaining mostly to how to set up an Assembly with a plenum block; let me know what you think. These are probably all due to our docs needing some work, so if there was anything in there that was confusing or could use clarification, let us know and we can try to improve them (curse of knowledge!).

Also, looks like Black is still mad. If you have the latest version of Black installed, you should be able to do something like `black armi/` from the root of the repo and it should format everything for you.

armi/reactor/tests/test_assemblies.py
davis8dd: Thank you! This will be updated in the next commit.

armi/reactor/tests/test_assemblies.py
davis8dd: Thank you, I'm still internalizing Python's special methods.  It will be updated in the next commit.

armi/reactor/tests/test_assemblies.py
youngmit: As in to add an `isinstance(obj, Block)` check on `Assembly.add()`? Might be a pretty good idea. I'm all for firming up/formalizing all of the mysterious assumptions we make. If you were to add such a check, and the tests pass, I'd be happy to adopt it. I would be curious to know if some things stub their toe on such a check...

youngmit (approved):
So good. Thank you!

PR #358, opened 2021-08-03
Add the ANL116 energy group structure.

jakehader:


ntouran (approved):
assuming from mc2 also

Issue #357, opened 2021-08-02
Building Macro XS's without MPI parallelism


keckler:
If the import of `mpi4py` is unsuccessful, `armi.MPI_COMM` is set to `None` within context.py. This leads to issues with the current implementation of `invokeHook` within the `MacroXSGenerator` class, which assumes that `armi.MPI_COMM` is instantiated, and causes the code to fail with an AttributeError.

This means that machines which fail the `mpi4py` load will not be able to run a large class of ARMI inputs, even though `mpi4py` is not required to run the input in serial.

This could be fixed by adding logic to the `invokeHook` method to explicitly handle cases with `armi.MPI_SIZE=1` so that calls to `armi.MPI_COMM` are avoided. This solution has been preliminarily implemented successfully on my machine, but another set of eyes would be useful to determine if there is a better solution to the problem.

jakehader:
@youngmit and I were just talking about this issue so thanks for making a ticket (and also discovering the issue too!). Mitch had some ideas of how to handle this more generically within the framework if a machine did not have MSMPI (or equivalent) installed.

keckler:
I figured that someone might have a more sophisticated idea on how to fix this than me. I'll leave it to the experts.

jakehader:
@keckler, can you just submit the changes that you made and we can review them? Your solution might be a good stop gap for now and we can think about future improvements if needed.

theJollySin:
@youngmit   Is this your 'stub version of MPI' idea?  Or do you favor something more like keckler's idea above (which also sounds quite reasonable).

jakehader:
@theJollySin, yeah the other idea that @youngmit had was the MPI stub, but maybe that's a longer term solution and @keckler's fix is a shorter term solution. We can keep this issue open even after the PR.

keckler:
Sure, I'll put it up later tonight and let you know.

keckler:
The fix that I implemented is up at #359 . @theJollySin @jakehader @youngmit

PR #359, opened 2021-08-03
Fix building of MacroXS tables when MPI import fails #357

keckler:
When mpi4py import fails, `MPI_SIZE` defaults to 1. Use this to disable the
execution of MPI commands. See #357

### DIFF
--- a/armi/physics/neutronics/macroXSGenerationInterface.py
+++ b/armi/physics/neutronics/macroXSGenerationInterface.py
@@ -84,8 +84,8 @@ class MacroXSGenerator(mpiActions.MpiAction):
             allMacros = [mc.createMacrosFromMicros(lib, b, libType=self.libType) for b in allBlocks]
 
         if armi.MPI_RANK == 0:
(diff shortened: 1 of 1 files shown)


keckler:
Eh, wait, I've got a failing test.

theJollySin:
The code coverage just went down because you added lines to the code.  So... by definition the % code coverage went down very slightly for the whole project.  Oh well.  (It would be nice if we could tell it not to fail code coverage unless it dropped by more than 0.01% or something.)

**Fun Fact**: The file `macroXSGenerationInterface.py` has exactly 0% code coverage.

keckler:
Yeah, I saw that. What I meant when I said a test failed was that the check-formatting test failed cause I had code lines longer than 80 characters. So I fixed that and push it back up. It should be ready now.

jakehader:
This was tested manually on a machine that did not have MSMPI and it worked. I will accept and merge in.

jakehader (approved):

PR #361, opened 2021-08-04
Add Report Content Structure

chris10mckenz:
Add a new way of compiling report content. In general, ReportContent is
the way report content is stored. It is made up of ReportNodes(Sections, Tables,
Images, TimeSeries(the storage representation for eventual graph output) and 
HtmlElements. It is converted to an html page output when writeReports()
 is called and renders each of it's items into html elements. In addition, 
getReportContents() is a plugin hook that collects contents from different
 plugins over the course of collecting data for the report. 
This allows for different plugins to either add to an existing Section/Table/etc
 or create new Section/Tables/etc

Currently this collection happens in the entry point for
reports, that allows a user to either provide a settings yaml file or an
h5 file to create report content.

In addition, created developer documentation for these items within
reports.rst. Also, added added a section for accessing the report entry
point within the User Documentation section.


armi/bookkeeping/report/reportingUtils.py
```diff
@@ -693,6 +693,27 @@ def makeCoreDesignReport(core, cs):
     _setGeneralSimulationData(core, cs, coreDesignTable)
 
 
+def makeCoreDesignReport2(core, cs, report):
```
youngmit, line 696: pull this into newReportUtils?

armi/physics/neutronics/__init__.py
```diff
@@ -31,7 +31,9 @@
     plugins outside of ARMI.
 
 """
+
 import os
+import base64
```
youngmit, line 36: not used

armi/physics/neutronics/reportConstants.py
```diff
@@ -0,0 +1,6 @@
+"""
+    Constants
+"""
+NEUTRONICS_SECTION = "Neutronics"
+KEFF_PLOT = "Keff-Plot"
+INITIAL_CORE_FUEL_ASSEMBLY = "Initial Core Fuel Assembly Count"
```
youngmit, line 6: I would just put these straight into neutronics/reports.py

armi/plugins.py
```diff
@@ -121,12 +121,14 @@
    replaced with plugin-based fuel handler logic.
 """
 from typing import Dict, Union
+from h5py._hl.files import File
```
youngmit, line 124: unused

armi/plugins.py
```diff
@@ -536,6 +538,23 @@ def mpiActionRequiresReset(cmd) -> bool:
...
+
+        Parameters
+        ----------
+        r : a reactor
+        cs : case settings
+        report : current report object to add to
+        blueprint : blueprint for a reactor (if None, only partial contents created)
+        stage : begin/standard/or end (stage of the report for
+                when inserting BOL vs. EOL content)
+
```
youngmit, line 555: I would add a link to the developer docs that you wrote.

youngmit (changes requested):
A high-level thing is function naming. There are some functions in newReportUtils.py that serve mostly to insert contents into the passed `report` argument. I would make sure that these are named consistently (I recommend a `insert[Whatever]()` convention). Some of them are returning some ReportNode stuff that is expected to be inserted by the caller. Those I would also name consistently (`get[Whatever]()`?). There are also some that seem to add stuff to the report and return something. I think it probably best to steer clear of that pattern because it's rather confusing to have a function with side effects on an argument, but also produces a value. I suspect the return values in these cases are not used anyways.

There some functions in the old reportUtils module named `whatever2()` is there a reason why they are there, instead of just being in newReportUtils? The ultimate goal will be to remove reportUtils and replace it with newReportUtils, so we probably dont want to mix the two worlds too much, or have anything in old reportUtils that we can't just remove without breaking the new reports.

armi/bookkeeping/newReportUtils.py
```diff
@@ -0,0 +1,831 @@
...
+
+from numpy import ComplexWarning
+from framework.armi import runLog
+from armi.reactor.flags import Flags
+from armi.bookkeeping import newReports
+from armi.reactor import blueprints
+import os
+import matplotlib.pyplot as plt
+import htmltree
+from htmltree import Table as HtmlTable
```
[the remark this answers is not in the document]
chris10mckenz: I moved table of contents into newReports so don't even need to import here anymore. But yes, also only import htmltree in general now.

armi/bookkeeping/newReportUtils.py
```diff
@@ -0,0 +1,711 @@
...
+    """
+
+    for c, frac in block.getVolumeFractions():
+
+        report[COMPREHENSIVE_REPORT][ASSEMBLY_AREA].addRow(
+            [c.getName(), "{0:10f}".format(c.getArea()), "{0:10f}".format(frac)]
+        )
+
+
+def setDimensionReport(comp):
```
youngmit, line 508: If this returns a Table, lets call it, `get`, or `create`, rather than `set`. `set` implies side effects.

armi/bookkeeping/newReportUtils.py
```diff
@@ -0,0 +1,711 @@
...
+        vals = (
+            "Thickness (cm)",
+            "{0:.7f}".format(coldThick),
+            "{0:.7f}".format(hotThick),
+        )
+        reportGroup.addRow([vals[0], vals[1], vals[2]])
+    return reportGroup
+
+
+def makeCoreAndAssemblyMaps(
```
youngmit, line 593: insert?

youngmit (changes requested):
A few more minor things.

armi/cli/reportsEntryPoint.py
```diff
@@ -0,0 +1,170 @@
...
+                        ):
+                            continue
+
+                        r = db.load(cycle, node)
+                        cs = db.loadCS()
+
+                        if i == len(dbNodes) - 1:
+                            stage = ReportStage.End
+                        else:
+                            stage = ReportStage.Standard
```
youngmit, line 157: Is this right? Seems like you still need standard processing for the last node, but extra end processing, too. So always do Standard, then outside the loop do another call to the plugin hood, passing `End`

youngmit (changes requested):

youngmit (approved):

PR #362, opened 2021-08-04
Implement various improvements related to GridBlueprints

youngmit:
This:
 - Restricts the through-center-assembly inferrence to full-symmetry
   cases. Fractional cases nx-ny do not carry enough information on
   their own.
 - Makes "assembly" option to activate "through center" for Symmetry;
   assembly is overly specific and weird, since many lattices are for
   laying out pins.
 - Adds adds a public getter/setter to Grid._offset. This allows client
   code to reach in and change the offset if necessary.

### DIFF
--- a/armi/reactor/geometry.py
+++ b/armi/reactor/geometry.py
@@ -266,8 +266,10 @@ class BoundaryType(enum.Enum):
         elif canonical == REFLECTIVE:
             return cls.REFLECTIVE
 
-        errorMsg = "{} is not a valid boundary option. Valid boundary options are:".format(
-            str(canonical)
+        errorMsg = (
+            "{} is not a valid boundary option. Valid boundary options are:".format(
+                str(canonical)
+            )
         )
         errorMsg += ", ".join([f"{sym}" for sym in boundaryTypes])
         raise ValueError(errorMsg)
@@ -449,7 +451,8 @@ class SymmetryType:
 
 
 def checkValidGeomSymmetryCombo(
-    geomType: Union[str, "GeomType"], symmetryInput: Union[str, "SymmetryType"],
+    geomType: Union[str, "GeomType"],
+    symmetryInput: Union[str, "SymmetryType"],
 ) -> bool:
     """
     Check if the given combination of GeomType and SymmetryType is valid.
--- a/armi/reactor/geometry.py
+++ b/armi/reactor/geometry.py
@@ -266,10 +266,8 @@ class BoundaryType(enum.Enum):
         elif canonical == REFLECTIVE:
             return cls.REFLECTIVE
 
-        errorMsg = (
-            "{} is not a valid boundary option. Valid boundary options are:".format(
-                str(canonical)
-            )
+        errorMsg = "{} is not a valid boundary option. Valid boundary options are:".format(
+            str(canonical)
         )
         errorMsg += ", ".join([f"{sym}" for sym in boundaryTypes])
         raise ValueError(errorMsg)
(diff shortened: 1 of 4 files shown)


armi/reactor/geometry.py
```diff
@@ -359,8 +359,10 @@ def createValidSymmetryStrings(cls):
     def fromStr(cls, symmetryString: str) -> "SymmetryType":
         """Construct a SymmetryType object from a valid string"""
         canonical = symmetryString.lower().strip()
+        # ignore "assembly" since it is unnecessary and overly-verbose and too specific
+        noAssembly = canonical.replace("assembly", "").strip()
         isThroughCenter = cls._checkIfThroughCenter(canonical)
-        coreString = canonical.replace(THROUGH_CENTER_ASSEMBLY, "").strip()
+        coreString = noAssembly.replace(THROUGH_CENTER_ASSEMBLY, "").strip()
```
jakehader, line 365: Is `assembly` going to be in the symmetryString anymore now that the `THROUGH_CENTER_ASSEMBLY ` string below was changed? Also, is there a reason for us keeping the `core` string in symmetryString? This seems a bit fragile long term.

armi/reactor/geometry.py
```diff
@@ -519,9 +521,8 @@ def checkValidGeomSymmetryCombo(
 REFLECTIVE = "reflective"
 PERIODIC = "periodic"
 NO_SYMMETRY = "no symmetry"
-THROUGH_CENTER_ASSEMBLY = (
-    "through center assembly"  # through center assembly applies only to cartesian
-)
+# through center assembly applies only to cartesian
+THROUGH_CENTER_ASSEMBLY = "through center"
```
jakehader, line 525: Should we rename to just `THROUGH_CENTER`? It seems that this symmetry could be applied to block grids as well, no?

jakehader (changes requested):
Can you add a unit test to cover the changes in the `gridBlueprint` for setting `symmetry.isThroughCenterAssembly`?

Also, should we make the `isThroughCenterAssembly` instance attribute on symmetry equal to `isThroughCenter` so it is more generic for non-core grids?

armi/reactor/geometry.py
youngmit: they are both optional. `quarter core through center assembly` is the same as `quarter through center`. That' why we are just stripping the words out; they carry no actual meaning

armi/reactor/geometry.py
youngmit: Yeah, I guess so. Since we no longer store this stuff in strings/do direct string comparisons against them, `THROUGH_CENTER_ASSEMBLY` is not really used anywhere other than in this module. Might as well rename though

youngmit:
>Also, should we make the isThroughCenterAssembly instance attribute on symmetry equal to isThroughCenter so it is more generic for non-core grids?

I don't want to break the API in this PR.

armi/reactor/geometry.py
youngmit: Ugg it is defined in geometry.py, but still being used directly in one place in blueprints. Blueprints shouldn't be using it directly, but im not sure this is the time or place

armi/reactor/geometry.py
jakehader: Oh interesting. Seems like we should just have one way to define this but as you say below you don't want an API breaking change. Maybe a good issue?

armi/reactor/geometry.py
jakehader: Agreed that its not the right time/place, but just something i noticed...

jakehader (approved):

PR #363, opened 2021-08-05
Update to running directory in the generic ExecutionOptions class.

jakehader:
Updating the default run directory within the generic ExecutionOptions to be based on the first eight chateracters of the hash of combining the case title and the label. This helps to create a unique directory name for running the case but also helps to avoid OS related file path length issues (particularly in Windows).

### DIFF
--- a/armi/physics/executers.py
+++ b/armi/physics/executers.py
@@ -6,6 +6,7 @@ data pathways.
 """
 
 import os
+import hashlib
 
 from armi.utils import directoryChangers
 from armi import runLog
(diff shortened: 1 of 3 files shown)


theJollySin:
The change seems straight-forward, and I'm all for it.

I wish that making this change broke a unit test though. What this tells me is there are not tests that cover that code.

jakehader:
Working on some tests now. Thanks for the feedback @theJollySin

youngmit (approved):

PR #365, opened 2021-08-05
Moving to faster YAML reads by default

john-science:
The `yamlize` library defaults to using a read-and-write YAML library to read: `ruamel.yaml.RoundTripLoader`, but this is ~12x slower at reading large YAML files than `ruamel.yaml.CLoader`, which is best for read-only situations.

So, this PR updates our use of the `yamlize` API to default to non-round-trip scenarios. The change is in some ways a slam-dunk, straight-forward speed improvement. But I have two concerns:

1. I am new to the code base, and may have missed a YAML-writing function that needs to be updated for this change.
2. This change will affect plugins and such, so perhaps we need to inform people of this semi-opaque change.

**Speed Comparisons:**

For the first test, I read a very large, example YAML file using various `yamlize`/`ruamel` options:

| YAML Loader | Pure Python implementation? | cProfile time (seconds) |
| :-------- | :-------- | :-------- |
| `RoundTripLoader` | True | 33.040 |
| `RoundTripLoader` | False | 33.270 |
| `CLoader` | True | 26.231 |
| `CLoader` | False | 2.744 |

So, it appears that using `CLoader` is >10x speed improvement over our current pathway. But, of course, loading a YAML blueprint in ARMI involves more than just the YAML string parsing. So, I ran another test where I loaded the above, large blueprint file:

| ARMI command | cProfile time (seconds) |
| :-------- | :-------- |
| `Blueprints().load(file, Loader=RoundTripLoader)` | 38.007 |
| `Blueprints().load(file, Loader=CLoader)` | 7.450 |

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -142,7 +142,7 @@ class Case:
         This property allows lazy loading.
         """
         if self._bp is None:
-            self._bp = blueprints.loadFromCs(self.cs)
+            self._bp = blueprints.loadFromCs(self.cs, roundTrip=True)
         return self._bp
 
     @bp.setter
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -34,10 +34,10 @@ import time
 import textwrap
 import ast
 from typing import Dict, Optional, Sequence
+import glob
 import tabulate
 import six
 import coverage
-import glob
 
 import armi
 from armi import context
@@ -46,7 +46,6 @@ from armi import operators
 from armi import runLog
 from armi import interfaces
 from armi.reactor import blueprints
-from armi.reactor import geometry
 from armi.reactor import systemLayoutInput
 from armi.reactor import reactors
 from armi.bookkeeping import report
@@ -674,8 +673,8 @@ class Case:
         ):
             # trick: these seemingly no-ops load the bp and geom via properties if
             # they are not yet initialized.
-            self.bp
-            self.geom
+            self.bp  # pylint: disable=pointless-statement
+            self.geom  # pylint: disable=pointless-statement
             self.cs["loadingFile"] = self.title + "-blueprints.yaml"
             if self.geom:
                 self.cs["geomFile"] = self.title + "-geom.yaml"
--- a/armi/reactor/blueprints/__init__.py
+++ b/armi/reactor/blueprints/__init__.py
@@ -68,18 +68,19 @@ text representations of input and objects in the code.
 
 
 """
-import copy
-import collections
 from collections import OrderedDict
+import collections
+import copy
 import os
 import pathlib
 import traceback
 import typing
 
+from ruamel.yaml import CLoader, RoundTripLoader
+import ordered_set
 import tabulate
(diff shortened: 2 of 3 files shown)


youngmit:
One thing to keep in mind (and this might not always be important), but one of the reasons for using the round trip loader was to preserve certain aspects of YAML formatting and comments. @ntouran  do you know if we are testing all of the desired behavior of this?

youngmit:
But holy cats; 38 sec to 7.4 sec is a huge get!

theJollySin:
> One thing to keep in mind (and this might not always be important), but one of the reasons for using the round trip loader was to preserve certain aspects of YAML formatting and comments. @ntouran do you know if we are testing all of the desired behavior of this?

Ah ha!  I can't wait to hear the answer to this.   My only real concern about this change is it is easy to make, but also easy to change parts of the code base I haven't poked around in yet.

If there *is* desired behavior we aren't testing, I'd be happy to add some unit tests.

ntouran:
I do not think we have testing for that because the round tripping is usually handled by external param sweep scripts. We could check the in-use tests as submitted to see if comments survive.

armi/cases/case.py
```diff
@@ -142,7 +142,7 @@ def bp(self):
         This property allows lazy loading.
         """
         if self._bp is None:
-            self._bp = blueprints.loadFromCs(self.cs)
+            self._bp = blueprints.loadFromCs(self.cs, roundTrip=True)
```
ntouran, line 145: why true by default here but false by default in the method? Could use a comment explaining

armi/reactor/blueprints/__init__.py
```diff
@@ -508,6 +509,11 @@ def migrate(cls, inp: typing.TextIO):
                 inp = mig.apply()
         return inp
 
+    @classmethod
+    def load(cls, stream, roundTrip=False):
+        loader = RoundTripLoader if roundTrip else CLoader
+        return super().load(stream, Loader=loader)
+
```
ntouran, line 516: Could use an explanatory docstring about how cloader is faster but doesn't allow round tripping. Sidenote, are we sure the cloader still doesn't support roundtripping?

ntouran (approved):

armi/cases/case.py
john-science: The idea here is that non-round-trip is the much more common case in our code base.  Normally, we are just reading data from blueprint YAMLs.  But in the caseSuite situation, we might want to write them back out, so in that one scenario we want 'round trip' as an option.

Am I wrong in my understanding of the situation?

armi/reactor/blueprints/__init__.py
john-science: 1. Sure, I'll add a comment, that's fair.
2. Huh... I guess I assumed since it was call "Round Trip" Loader, and the other loader is WAY faster, there would be no reason to use it unless the other loader didn't do "round trip".  But I suppose I didn't verify that. hmm

Issue #366, opened 2021-08-06
Bugs in updated logger


jakehader:
Recently, #336 was merged in and I have noticed several bugs that are now present when running cases:

1. All outputs are being sent to the standard error (stderr) stream.
2. Some MPI messages appear to be missing from the outputs, even for the head process (MPI_RANK=0)
3. I received the following logger error after a run was "completed" causing the case to fail:

```
[CRIT 004 ] Stack trace: Traceback (most recent call last):
  File "...\home\jhader\natrium\framework\armi\operators\operatorMPI.py", line 103, in operate
    self.workerOperate()
  File "...\home\jhader\natrium\framework\armi\operators\operatorMPI.py", line 149, in workerOperate
    self.workerQuit()
  File "...\home\jhader\natrium\framework\armi\operators\operatorMPI.py", line 230, in workerQuit
    runLog.close()  # no more messages.
  File "...\home\jhader\natrium\framework\armi\runLog.py", line 254, in close
    _ = [h.close() for h in LOG.stderrLogger.logger.handlers]
AttributeError: 'Logger' object has no attribute 'logger'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "...\home\jhader\natrium\nala\__main__.py", line 24, in main
    code = ArmiCLI().run()
  File "...\home\jhader\natrium\framework\armi\cli\__init__.py", line 189, in run
    return self.executeCommand(args.command, args.args)
  File "...\home\jhader\natrium\framework\armi\cli\__init__.py", line 218, in executeCommand
    return cmd.invoke()
  File "...\home\jhader\natrium\framework\armi\cli\run.py", line 33, in invoke
    inputCase.run()
  File "...\home\jhader\natrium\framework\armi\cases\case.py", line 348, in run
    o.operate()
  File "...\home\jhader\natrium\framework\armi\operators\operatorMPI.py", line 111, in operate
    traceback.print_exc()
  File "...\apps\dev\Python\3.7.2\lib\traceback.py", line 163, in print_exc
    print_exception(*sys.exc_info(), limit=limit, file=file, chain=chain)
  File "...\apps\dev\Python\3.7.2\lib\traceback.py", line 105, in print_exception
    print(line, file=file, end="")
AttributeError: 'Logger' object has no attribute 'write'
```

theJollySin:
Darn and/or shucks.

Okay, one at a time:

1. This would be quite surprising to me. Especially since we are literally `print`ing the log messages at the end of the run: 
[...]

PR #367, opened 2021-08-07
Bug Fixes - logging inconsistency

john-science:
This PR fixes the issues @jakehader brought up in this ticket:  https://github.com/terrapower/armi/issues/366

Essentially, it fixes two separate problems:

1. An issue in `_RunLog.close()` where the stderr logger was being called incorrectly in some circumstances.
2. An issue where (I believe) everything was being written purely to stderr, not stdout, in the parent process.

Bonus Features:

1. While I was at it, I removed the PyTest "caplog" dependency. I added it in the original logging PR, and I didn't like it. 
2. Also, there was an issue with logging being called during testing before the `runLog.py` was fully called and instantiated. So I added a new `NullLogger` class to fix that.

### DIFF
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -41,7 +41,6 @@ import logging
 import operator
 import os
 import sys
-import time
 
 from armi import context
 
@@ -171,7 +170,7 @@ class _RunLog:
         """Summarize all warnings for the run."""
         self.logger.warningReport()
 
-    def _getLogVerbosityRank(self, level):
+    def getLogVerbosityRank(self, level):
         """Return integer verbosity rank given the string verbosity name."""
         try:
             return self._logLevels[level][0]
@@ -202,7 +201,7 @@ class _RunLog:
 
         """
         if isinstance(level, str):
-            self._verbosity = self._getLogVerbosityRank(level)
+            self._verbosity = self.getLogVerbosityRank(level)
         elif isinstance(level, int):
             # The logging module does strange things if you set the log level to something other than DEBUG, INFO, etc
             # So, if someone tries, we HAVE to set the log level at a canonical value.
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -47,11 +47,14 @@ from armi import context
 
 
 # global constants
-_WHITE_SPACE = " " * 6
 _ADD_LOG_METHOD_STR = """def {0}(self, message, *args, **kws):
     if self.isEnabledFor({1}):
         self._log({1}, message, args, **kws)
 logging.Logger.{0} = {0}"""
+_WHITE_SPACE = " " * 6
+SEP = "|"
+STDERR_LOGGER_NAME = "ARMI_ERROR"
+STDOUT_LOGGER_NAME = "ARMI"
 
 
 class _RunLog:
@@ -81,12 +84,17 @@ class _RunLog:
         self._verbosity = logging.INFO
         self.initialErr = None
         self._logLevels = None
-        self.logger = logging.getLogger("NULL")
-        self.logger.addHandler(logging.NullHandler())
-        self.stderrLogger = logging.getLogger("NULL2")
-        self.stderrLogger.addHandler(logging.NullHandler())
+        self._logLevelNumbers = []
+        self.logger = None
+        self.stderrLogger = None
 
+        self._setNullLoggers()
         self._setLogLevels()
+
+    def _setNullLoggers(self):
+        """Helper method to set both of our loggers to Null handlers"""
+        self.logger = NullLogger("NULL")
+        self.stderrLogger = NullLogger("NULL2", isStderr=True)
 
     def _setLogLevels(self):
         """Here we fill the logLevels dict with custom strings that depend on the MPI rank"""
@@ -104,6 +112,7 @@ class _RunLog:
                 ("header", (100, "".format(_rank))),
             ]
         )
+        self._logLevelNumbers = sorted([l[0] for l in self._logLevels.values()])
         global _WHITE_SPACE
         _WHITE_SPACE = " " * len(max([l[1] for l in self._logLevels.values()]))
 
@@ -195,18 +204,24 @@ class _RunLog:
         if isinstance(level, str):
             self._verbosity = self._getLogVerbosityRank(level)
         elif isinstance(level, int):
-            if level < 0 or level > 100:
-                raise KeyError(
-                    "Invalid verbosity rank {}. ".format(level)
-                    + "It needs to be in the range [0, 100]."
-                )
-            self._verbosity = level
+            # The logging module does strange things if you set the log level to something other than DEBUG, INFO, etc
+            # So, if someone tries, we HAVE to set the log level at a canonical value.
+            # Otherwise, nearly all log statements will be silently dropped.
+            if level in self._logLevelNumbers:
+                self._verbosity = level
+            elif level < self._logLevelNumbers[0]:
+                self._verbosity = self._logLevelNumbers[0]
+            else:
+                for i in range(len(self._logLevelNumbers) - 1, -1, -1):
+                    if level >= self._logLevelNumbers[i]:
+                        self._verbosity = self._logLevelNumbers[i]
+                        break
         else:
             raise TypeError("Invalid verbosity rank {}.".format(level))
 
         if self.logger is not None:
-            for h in self.logger.handlers:
-                h.setLevel(self._verbosity)
+            for handler in self.logger.handlers:
+                handler.setLevel(self._verbosity)
             self.logger.setLevel(self._verbosity)
 
     def getVerbosity(self):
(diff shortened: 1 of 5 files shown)


jakehader:
Hi @john-science, thanks for the update! I am currently testing this out locally and it looks like I am still seeing that the `verbosity` setting is not being handled correctly. For example, in my case I have `cs["verbosity"]` set to `extra` but I am only getting `important` and `info` statements.

Additionally, one thing that is a bit strange is I am seeing logging information from the `font_manager.py` in `matplotlib` written to the stdout as the run is starting:

```
[info] generated new fontManager
[info] generated new fontManager
```

image

It would be nice if we could send output from other external modules to a separate output file, or at least make it clear that this is coming from an external module.

jakehader:
One more issue that I am seeing is that when using `ipython` the logger does not emit anything to the command line. This is problematic when trying to load cases interactively with `o = armi.init()` because it doesn't display the choices. Additionally, the `runLog.raw` calls for the splash and other high-level information is not emitted either.

jakehader:
hi @john-science, as a follow-up it looks like the case ran to completion.

john-science:
> One more issue that I am seeing is that when using `ipython` the logger does not emit anything to the command line. This is problematic when trying to load cases interactively with `o = armi.init()` because it doesn't display the choices. Additionally, the `runLog.raw` calls for the splash and other high-level information is not emitted either.
> 
> I assume this has something to do with the `NullLogger` being initialized, but I am not sure.

This isn't an error. This is by design.  We should talk about it.

Now that we're not just letting things `print()`, we have to initialize a logger object. Before we do that, some code in `context.py` and even `opeartorsMPI.py` needs to be called. (Though we might remove the operatorsMPI step.)  Anyway, we start with NullLoggers.  

If we want, we can start off with some default logging parameters. But those have to be independent of MPI_RANK (child process number), so it's not obvious what would be safe to do there.  

Obviously, you could `runLog.startLog()`, and all of this would go away. But if we want to support logging before `context.py` is called, what would we want to do for default logging settings.  (@youngmit  Opinions?)

john-science:
> It would be nice if we could send output from other external modules to a separate output file, or at least make it clear that this is coming from an external module.

I asked the team about this when we were designing requirements for this new logging feature, and it was decide it was out of scope.  So I guess it would just be a ticket.

(Off the top of my head, I don't know what I would want the API to look like that would be easy to understand and use. Also, I think it would possibly interact with the `startLog("my_fancy_run_name")` method that we currently have.)

john-science:
> I am currently testing this out locally and it looks like I am still seeing that the `verbosity` setting is not being handled correctly. For example, in my case I have `cs["verbosity"]` set to `extra` but I am only getting `important` and `info` statements.

Sorry, can you show me an example where you are doing this:  `cs["verbosity"]`?   And did that work before the logging change?

In my mind, the only way to change the log level is to do something like: `runLog.setVerbosity(0)` or maybe `logger.setLevel(0)`. I'm trying to imagine what dictionary `cs` is, and where you would be setting it.  There must be a workflow I have not seen before.

jakehader:
My understanding is when `python -m armi run` gets called the `verbosity` setting should be getting propagated to the log configuration. There are many places in the code that call `runLog.extra`. These seem to be not showing up even when the `cs["verbosity"]` setting in a case is set to `extra`. 

The `setVerbosity` call is in the entry point code:

https://github.com/terrapower/armi/blob/master/armi/cli/entryPoint.py#L137

> This isn't an error. This is by design. We should talk about it. 

> Now that we're not just letting things print(), we have to initialize a logger object. Before we do that, some code in context.py and even opeartorsMPI.py needs to be called. (Though we might remove the operatorsMPI step.) Anyway, we start with NullLoggers.

I feel strongly that being able to see the `runLog` statements within an interactive session is a requirement of at least an application, but probably the framework as well. Could this not be configured when calling `armi.configure` from `ipython`?

jakehader:
@john-science - it looks like a simple fix to adding the runLog in the interactive session is to add the `startLog` within the `configure` function of `armi.__init__.py`:

```
    if _liveInterpreter():
        runLog.LOG.startLog(name=f"interactive-{app.name}")
        cli.splash()
```

Doing this it seems that the interactive session does print the `runLog.extra` information. 

The only other instance of `startLog` being called is in `OperatorMPI`'s `__init__`. Is it possible that the `cs['verbosity']` is being set on a different logger before the actual log for the run is initialized?

```
class OperatorMPI(Operator):
    """MPI-aware Operator."""

    def __init__(self, cs):
        runLog.LOG.startLog(cs.caseTitle)
        try:
            Operator.__init__(self, cs)
        except:
            # kill the workers too so everything dies.
            runLog.important("Master node failed on init. Quitting.")
            if armi.MPI_COMM:  # else it's a single cpu case.
                armi.MPI_COMM.bcast("quit", root=0)
            raise
```

Also, I noticed in `startLog` that the name of the `stdout` logger is hard-coded to `ARMI` and the name of the `stderr` logger is set to `ARMI_ERROR`. Should this be set based on the user-supplied name or do we want to keep this consistent across all ARMI-based applications?

image

john-science:
> The `setVerbosity` call is in the entry point code:
> https://github.com/terrapower/armi/blob/master/armi/cli/entryPoint.py#L137

Thanks! As I said, I have unit tests that show `setVerbosity()` works, so I imagine this is related to a workflow that I haven't tested against (because I didn't know it existed).  I will work on it this weekend.

> I feel strongly that being able to see the `runLog` statements within an interactive session is a requirement of at least an application, but probably the framework as well. Could this not be configured when calling `armi.configure` from `ipython`?

Sorry, let me be more clear. Sure, logging from an interactive session is great. But in the past, this wasn't done via our `runLog.py` code unless you called `runLog.startLog('my_run_name')`. And I can prove that, because without calling `startLog()` the system doesn't know several things, among them the name of any output log files is defined by that string: `"my_run_name"`. So any logging that was done was handled outside all of our custom logging logic, not piped to our runLog stdout/stderr locations, but handled with whatever the Python system defaults were.

So I am agreeing with you. We need to make a change to support your workflow. But there are no system defaults for the `logging` module, which we are using now, so we need to build our own. Options:

1. You have to call `startLog()` in iPython if you want custom logging, or we have to add `startLog()` somewhere in your workflow to automate this.
2. OR we can set some default logging before `startLog()` is called: but I need some opinions on what that default logging should look like (that don't break parallel runs).

> Do you have the set of requirements documented somewhere that we could review?

Sorry, there was no design doc. I asked people how they wanted the new single-module custom logging to work and the opinion was it should be a minimal layer on the old logging: no custom files or changes in logging powers. That's a changable decision.

jakehader:
@john-science - Note that it's OK to have multiple commits. We end up squashing the commits when merging them into master so no need to force push. Testing on the updates now.

armi/runLog.py
```diff
@@ -528,6 +543,19 @@ def setVerbosity(self, intLevel):
         self.setLevel(intLevel)
 
 
+class NullLogger(RunLogger):
+    def __init__(self, name, isStderr=False):
+        super().__init__(name)
```
jakehader, line 548: Typically we call the parent `__init__` directly in ARMI, so `RunLogger.__init__(name)`

armi/runLog.py
```diff
@@ -375,7 +388,7 @@ class DeduplicationFilter(logging.Filter):
     """
 
     def __init__(self, *args, **kwargs):
-        super(DeduplicationFilter, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
```
jakehader, line 391: `logging.Filter.__init__(*args, **kwargs)`

armi/tests/test_runLog.py
```diff
@@ -48,8 +41,10 @@ def test_setVerbosityFromString(self):
 
     def test_invalidSetVerbosityByRank(self):
         """Test that the log verbosity setting fails if the integer is invalid."""
```
jakehader, line 43: Test name and docstring should be updated now that the contents have changed.

armi/tests/test_runLog.py
```diff
@@ -106,13 +102,73 @@ def test_warningReport(self):
...
+        # close() and test that we have correctly nullified our loggers
+        runLog.close(1)
+        validate_loggers(log)
+
+        # in a real run, the parent process would close() after all the children
+        runLog.close(0)
+
+    def test_setVerbosity(self):
+        """Let's test the setVerbosity() method carefully"""
+        with mockRunLogs.BufferLog() as mock:
```
jakehader, line 143: Can you add tests for the lower verbosities like `extra`, and `debug`?

jakehader (changes requested):
Hi @john-science, it looks like it's working on my end in both interactive mode and submitting to our HPC cluster and it is respecting the value of `cs['verbosity']` now in the logger.

Just a couple of extra comments to consider.

PR #368, opened 2021-08-10
Remove Charge Fuel Pool (CFP) from core object

jakehader:
* Removing the ChargeFuelPool from the core since it was essentially unused and lead to duplication of assembly objects between the core's children and the CFP children. This resulted in erroneous sorting of assemblies within the `__lt__` implementation on the reactor. 

* Removing unused reactor/core timing parameters that are specific to plugins/third party codes like `DIF3D`, `MC2`, `Subchan`, etc.

* Removing the `numAssembliesFabricated` parameter as it was implied incorrectly within the fuel handler and unused throughout the code base.

* Removing the `numAssembliesInSFP` parameter since this is not being stored in the database and can be calculated directly by calling `len(core.sfp)`.

### DIFF
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -195,7 +195,6 @@ class Core(composites.Composite):
         # build a spent fuel pool for this reactor
         runLog.debug("Building spent fuel pools")
         self.sfp = assemblyLists.SpentFuelPool("Spent Fuel Pool", self)
-        self.cfp = assemblyLists.ChargedFuelPool("Charged Fuel Pool", self)
         self._lib = None  # placeholder for ISOTXS object
         self.locParams = {}  # location-based parameters
         # overridden in case.py to include pre-reactor time.
@@ -232,7 +231,6 @@ class Core(composites.Composite):
 
     def __setstate__(self, state):
         composites.Composite.__setstate__(self, state)
-        self.cfp.parent = self
         self.sfp.parent = self
         self.regenAssemblyLists()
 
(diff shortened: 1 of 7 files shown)


armi/reactor/reactors.py
```diff
@@ -1724,13 +1713,6 @@ def createAssemblyOfType(self, assemType=None, enrichList=None, cs=None):
                     # therefore breaks the burnup metric.
                     b.adjustUEnrich(enrich)
 
-        # see if there are tracked assemblies and add this to the list if so
-        if self._trackAssems:
-            self.cfp.add(a)
```
jakehader, line 1729: This is where the usage of the `cfp` was causing issues. Outside of this, the `cfp` was not being utilized in the framework or TerraPower plugin code.

ntouran: I don't think this can cause duplication since it will just hold pointers to the assemblies. But in any case, we do want to get rid of the (confusing) CFP so we should accept this as progress in that direction.

ntouran (approved):

youngmit (approved):

armi/reactor/reactors.py
youngmit: It's not really duplication that we are worried about, but rather lack of it. If `a` goes into `self.cfp`, then the same `a` gets returned to the caller (almost certainly to be `add()`ed to something else), unless the caller is explicitly `deepcopy`ing the returned assembly, we end up with two references to the same Assembly.

PR #370, opened 2021-08-11
Making concatLogs work with new context.LOG_DIR

john-science:
The `armi.runLog.concatenateLogs()` function should not be aware of the `context.LOG_DIR` path.

While it does not hurt us currently that `concatenateLogs()` is not fully in compliance, we really want this change to be made.


john-science:
@youngmit  This should have been part of PR #369 .

PR #372, opened 2021-08-13
Allow for loading from databases with extra parameters

youngmit:
This allows loading from old databases when parameters have since been
removed from the application. In such cases, a warning will be issued.

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -1263,12 +1263,14 @@ class Database3(database.Database):
             try:
                 pDef = pDefs[paramName]
             except KeyError:
-                if re.match(r"^n[A-Z][a-z]?\d*", paramName):
-                    # This is a temporary viz param made by _addHomogenizedNumberDensities
-                    # ignore it safely
-                    continue
-                else:
-                    raise
+                # If a parameter exists in the database but not in the application
+                # reading it, we can technically keep going. Since this may lead to
+                # potential correctness issues, raise a warning
+                runLog.warning(
+                    "Found `{}` parameter `{}` in the database, which is not defined. "
+                    "Ignoring it.".format(compTypeName, paramName)
+                )
+                continue
 
             data = dataSet[:]
             attrs = _resolveAttrs(dataSet.attrs, h5group)
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -1263,14 +1263,19 @@ class Database3(database.Database):
             try:
                 pDef = pDefs[paramName]
             except KeyError:
-                # If a parameter exists in the database but not in the application
-                # reading it, we can technically keep going. Since this may lead to
-                # potential correctness issues, raise a warning
-                runLog.warning(
-                    "Found `{}` parameter `{}` in the database, which is not defined. "
-                    "Ignoring it.".format(compTypeName, paramName)
-                )
-                continue
+                if re.match(r"^n[A-Z][a-z]?\d*", paramName):
+                    # This is a temporary viz param (number density) made by
+                    # _addHomogenizedNumberDensities ignore it safely
+                    continue
+                else:
+                    # If a parameter exists in the database but not in the application
+                    # reading it, we can technically keep going. Since this may lead to
+                    # potential correctness issues, raise a warning
+                    runLog.warning(
+                        "Found `{}` parameter `{}` in the database, which is not defined. "
+                        "Ignoring it.".format(compTypeName, paramName)
+                    )
+                    continue
 
             data = dataSet[:]
             attrs = _resolveAttrs(dataSet.attrs, h5group)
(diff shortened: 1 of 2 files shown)


john-science:
What is this about "correctness issues"? If the program doesn't need the data, the data won't affect how "correct" the program is, right?

jakehader:
> What is this about "correctness issues"? If the program doesn't need the data, the data won't affect how "correct" the program is, right?

@john-science, I think the warning is appropriate because it can tell a user that they may be using a database that is no longer "compatible" with the latest version of the framework/application. In this case we removed a core parameter that was implemented incorrectly and not very useful which was leading to failures when trying to load from that database.

armi/bookkeeping/db/database3.py
```diff
@@ -1264,11 +1264,18 @@ def _readParams(h5group, compTypeName, comps):
                 pDef = pDefs[paramName]
             except KeyError:
                 if re.match(r"^n[A-Z][a-z]?\d*", paramName):
-                    # This is a temporary viz param made by _addHomogenizedNumberDensities
-                    # ignore it safely
+                    # This is a temporary viz param (number density) made by
+                    # _addHomogenizedNumberDensities ignore it safely
```
jakehader, line 1268: This should be `_addHomogenizedNumberDensityParams`

armi/bookkeeping/db/database3.py
```diff
@@ -1264,11 +1264,18 @@ def _readParams(h5group, compTypeName, comps):
                 pDef = pDefs[paramName]
             except KeyError:
                 if re.match(r"^n[A-Z][a-z]?\d*", paramName):
```
jakehader, line 1266: It seems that because we have the comps passed in we should also verify that the first one is a subclass of `Block` just to be safe. I'm not aware of other instances that this may come up in the future but this seems like a good trap for us to miss loading in number densities of some type later on without realizing where the error is occurring?

armi/bookkeeping/db/database3.py
```diff
@@ -1264,11 +1264,18 @@ def _readParams(h5group, compTypeName, comps):
...
                 else:
-                    raise
+                    # If a parameter exists in the database but not in the application
+                    # reading it, we can technically keep going. Since this may lead to
+                    # potential correctness issues, raise a warning
+                    runLog.warning(
+                        "Found `{}` parameter `{}` in the database, which is not defined. "
+                        "Ignoring it.".format(compTypeName, paramName)
+                    )
+                    continue
```
jakehader, line 1278: This seems like we should add a unit test in `test_database3.py` to capture that this no longer raises an error for missing parameters.

jakehader (changes requested):

armi/bookkeeping/db/database3.py
youngmit: Number densities are not ever read from these into the reactor state; they are only written to the database as a crutch for data visualization. Since we are only checking this condition if we have already found that there is no corresponding parameter definition, this seems perfectly fine to me.

armi/bookkeeping/db/database3.py
youngmit: >What is this about "correctness issues"? If the program doesn't need the data, the data won't affect how "correct" the program is, right?

Correctness issues as in loading a data serialization file from at best a different version of the same program (where mechanisms to handle added/removed/renamed parameters are not explicitly specified), or perhaps a different application entirely, in which there's no guaranteed correspondence between the parameters that were written to the database and those that are defined in the application reading it. The last seems contrived, but most requests that I've gotten to relax this check have been to allow reading databases that were produced from completely different apps.

youngmit: Sidenote, this was kind of the inspiration for #191

armi/bookkeeping/db/tests/test_database3.py
```diff
@@ -194,6 +197,14 @@ def test_computeParents(self):
             database3.Layout.computeAncestors(serialNums, numChildren, 3), expected_3
         )
 
+    def test_load(self) -> None:
+        self.makeShuffleHistory()
+        with self.assertRaises(KeyError):
+            r = self.db.load(0, 0)
+
+        del self.db.h5db["c00n00/Reactor/missingParam"]
+        r = self.db.load(0, 0)
```
jakehader, line 206: Why delete and not just test passing the `allowMissing` argument as True?

armi/bookkeeping/db/database3.py
jakehader: Still needs to be updated.

armi/bookkeeping/db/database3.py
jakehader: Oh I see. That makes sense, thanks!

jakehader (changes requested):

armi/bookkeeping/db/tests/test_database3.py
youngmit: To make sure that the extra param is _why_ the KeyError is getting raised

jakehader: Right, but what about a test that allows the loading _not_ to fail when the parameter doesn't exist?

jakehader (changes requested):

jakehader (approved):

Issue #371, opened 2021-08-12
Explore immutable settings


john-science:
Currently, there are many places in the codebase where we call `getMasterCs()` and read in all the settings.  Slightly more troubling, there are many places in ARMI and in the plugins where we _change_ settings mid-run.

**Purpose**: explore how much havoc would be caused by making settings immutable.  

**Method**: change the settings API a bit to raise a `DeprecationWarning` every time a setting is altered.  Then test various real-world use cases to see how often (and where) this warning pops up. 

**Results**: report back on the prevalence and where these warnings pop up.

john-science:
@youngmit  Could you assign this to me, please?  Thanks!

youngmit:
Done!

john-science:
Before I go about the business of refactoring our settings code to make the settings truly mutable, I have outlined below all the places in the ARMI framework we mutate settings mid-run.  I would like your opinion on which of these cases below is **not** a valid use-case for mutable settings.

From this list, I will generate all the sections of the codebase and workflows that need to be refactored, aside from just the settings code itself.

| Num | File | Method | Is Valid Mutation? |
| :-- | :--- | :----- | :-------------: |
| 0 | [armi/bookkeeping/snapshotInterface.py](https://github.com/terrapower/armi/blob/73584062a24c135f323651f6bf2f95419ffbd168/armi/bookkeeping/snapshotInterface.py#L72) | SnapshotInterface.activateDefaultSnapshots() | Y |
| 1 | [armi/cases/case.py](https://github.com/terrapower/armi/blob/73584062a24c135f323651f6bf2f95419ffbd168/armi/cases/case.py#L687) | `Case.writeInputs()` | ???  "This allows input objects that have been modified in memory to be written" |
| 2 | [armi/cases/inputModifiers/neutronicsModifiers.py](https://github.com/terrapower/armi/blob/73584062a24c135f323651f6bf2f95419ffbd168/armi/cases/inputModifiers/neutronicsModifiers.py#L32) | `NeutronicConvergenceModifier.__call__()` | Y |
| 3 | [armi/cases/suiteBuilder.py](https://github.com/terrapower/armi/blob/73584062a24c135f323651f6bf2f95419ffbd168/armi/cases/suiteBuilder.py#L170) | `SuiteBuilder.buildSuite()` | Y |
[...]

Issue #246, opened 2021-02-04
Remove jagged numpy arrays


youngmit:
The [latest release of NumPy]( https://github.com/numpy/numpy/releases/tag/v1.24.0) has deprecated the jagged array treatment that we are using in the database when writing parameter data.

https://github.com/terrapower/armi/blob/36fe1bced3b817d6d76e66cf99dd355219993e63/armi/bookkeeping/db/database3.py#L841-L887

Until now, we have detected jagged data (and some other strange cases) by converting to a numpy array first, then checking its data type to see if it is `object` and figure out what to do from there (handle `None` values, pack jagged arrays, etc).

Notes
-----
Also, ARMI cannot support Python 3.12 with the older versions of NumPy we are forced to use for Jagged arrays. So this ticket is what is keeping ARMI from supporting Python 3.12.

theJollySin:
This seems like something I could handle.

And you're totally right, performance is my only real concern here.

john-science:
I implemented a pure-Python solution for this.  It's fine, but slow.  I think the real solution might be we wait until NumPy makes the change and then just catch the error they throw. It only needs to be caught in a few places. 

I can give it another whack, but NumPy has a lot of built-in logic for how to convert lists-of-sets-of-dicts to arrays. And a lot of special-case handling that, while it makes good sense, is slow to replicate in pure Python, for instance: handling of `None`s.

scottyak:
Perhaps [Awkward Array](https://github.com/scikit-hep/awkward-1.0) can help? From their README:

> Awkward Array is a library for nested, variable-sized data, including arbitrary-length lists, records, mixed types, and missing data, using NumPy-like idioms.
> 
> Arrays are dynamically typed, but operations on them are compiled and fast. Their behavior coincides with NumPy when array dimensions are regular and generalizes when they're not.

...

> Not only is the expression using Awkward Arrays more concise, using idioms familiar from NumPy, but it's much faster and uses less memory.
> 
> For a similar problem 10 million times larger than the one above (on a single-threaded 2.2 GHz processor),
> 
> the Awkward Array one-liner takes 4.6 seconds to run and uses 2.1 GB of memory,
> the equivalent using Python lists and dicts takes 138 seconds to run and uses 22 GB of memory.

`awkward` seems to be well-maintained with a strong community around it. If it's ok to include `awkward` as a dependency, I think the next step would be for someone to try porting the pure-Python logic over to `awkward`.

How does that sound?

[...]

PR #374, opened 2021-08-17
Adding a DeprecationWarning that settings will be immutable soon.

john-science:
Here is my little experiment in response to terrapower#371.

There are a lot of decisions that went into this experiment that people might have a different opinion on. For instance, I fixed a TON of unit tests so they wouldn't throw the deprecation warning. Also, the `lock` I put in the `Settings` class is very simple. I can change these if people like, but I was focused on the experiment.

This experiment also shows us that `reactors.py` and `database3.py` will need to be slightly re-designed if we want settings to be immutable during a run.

This PR would serve as a good warning to people working on plugins that a big change is on it's way.

### DIFF
--- a/armi/bookkeeping/snapshotInterface.py
+++ b/armi/bookkeeping/snapshotInterface.py
@@ -69,7 +69,9 @@ class SnapshotInterface(interfaces.Interface):
                 runLog.info(
                     "Adding default snapshot {0} to snapshot queue.".format(snapT)
                 )
+                self.cs.lock = False
                 self.cs["dumpSnapshot"] = self.cs["dumpSnapshot"] + [snapT]
+                self.cs.lock = True
 
     def _getSnapTimesEquilibrium(self):
         """Set BOEC, MOEC, EOEC snapshots."""
--- a/armi/bookkeeping/snapshotInterface.py
+++ b/armi/bookkeeping/snapshotInterface.py
@@ -69,9 +69,8 @@ class SnapshotInterface(interfaces.Interface):
                 runLog.info(
                     "Adding default snapshot {0} to snapshot queue.".format(snapT)
                 )
-                self.cs.lock = False
-                self.cs["dumpSnapshot"] = self.cs["dumpSnapshot"] + [snapT]
-                self.cs.lock = True
+                with self.cs.unlock():
+                    self.cs["dumpSnapshot"] = self.cs["dumpSnapshot"] + [snapT]
 
     def _getSnapTimesEquilibrium(self):
         """Set BOEC, MOEC, EOEC snapshots."""
--- a/armi/bookkeeping/tests/test_historyTracker.py
+++ b/armi/bookkeeping/tests/test_historyTracker.py
@@ -72,14 +72,13 @@ class TestHistoryTracker(ArmiTestHelper):
         runTutorialNotebook()
 
         reloadCs = settings.Settings(f"{CASE_TITLE}.yaml")
-        reloadCs.lock = False
-        reloadCs.caseTitle = "armiRun"
-        reloadCs["db"] = True
-        reloadCs["reloadDBName"] = pathlib.Path(f"{CASE_TITLE}.h5").absolute()
-        reloadCs["runType"] = "Snapshots"
-        reloadCs["loadStyle"] = "fromDB"
-        reloadCs["detailAssemLocationsBOL"] = ["001-001"]
-        reloadCs.lock = True
+        with reloadCs.unlock():
+            reloadCs.caseTitle = "armiRun"
+            reloadCs["db"] = True
+            reloadCs["reloadDBName"] = pathlib.Path(f"{CASE_TITLE}.h5").absolute()
+            reloadCs["runType"] = "Snapshots"
+            reloadCs["loadStyle"] = "fromDB"
+            reloadCs["detailAssemLocationsBOL"] = ["001-001"]
         o = armi_init(cs=reloadCs)
         cls.o = o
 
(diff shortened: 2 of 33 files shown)


john-science:
Well, I learned some things about how our settings are handled.  I also learned that our settings are modified during a run in a couple common places (`database3.py` and `reactors.py`).

Our unit tests modify settings left and right, but I'm okay with that. 

Also, in the `Settings` class we provide a number of ways to change settings: loading them from files, or strings, or replacing them in bulk with a dictionary.  I think aside from just placing a `lock` in this file, the long-term solution should be to reduce the size of the API that allows users to change settings.

john-science:
Well, I learned some things about how our settings are handled.  I also learned that our settings are modified during a run in a couple common places (`database3.py` and `reactors.py`).

Our unit tests modify settings left and right, but I'm okay with that. 

Also, in the `Settings` class we provide a number of ways to change settings: loading them from files, or strings, or replacing them in bulk with a dictionary.  I think aside from just placing a `lock` in this file, the long-term solution should be to reduce the size of the API surface that allows users to change settings.

youngmit:
>  I think aside from just placing a lock in this file, the long-term solution should be to reduce the size of the API surface that allows users to change settings.

If for each method that we remove, we could reasonably refactor away from the mutation, then this might be a good way to gradually do away with settings modifications; perhaps more gradually than with a hard lock. However, I do still see getting to a world where we could actually lock settings as valuable

youngmit:
There will still be places where certain modifications will be necessary (and to a careful programmer, acceptible), and in these circumstances we might expect the user to unlock the settings to make these changes. Better yet, if settings can be unlocked at creation, fiddled around with some, then locked forever, this might cover many, if not all, of the cases where we would want to modify something anyways (e.g., setting validation, loading settings from db, things like that).

armi/bookkeeping/snapshotInterface.py
```diff
@@ -69,7 +69,9 @@ def activateDefaultSnapshots(self):
                 runLog.info(
                     "Adding default snapshot {0} to snapshot queue.".format(snapT)
                 )
+                self.cs.lock = False
```
youngmit, line 72: Might be nice to expose the unlock capability through a context manager so that it's more difficult to forget to lock the settings again:
```
with cs.unlock():
    # change stuff
#sally forth
```

armi/bookkeeping/tests/test_historyTracker.py
```diff
@@ -72,12 +72,14 @@ def setUpClass(cls):
         runTutorialNotebook()
 
         reloadCs = settings.Settings(f"{CASE_TITLE}.yaml")
+        reloadCs.lock = False
         reloadCs.caseTitle = "armiRun"
         reloadCs["db"] = True
         reloadCs["reloadDBName"] = pathlib.Path(f"{CASE_TITLE}.h5").absolute()
         reloadCs["runType"] = "Snapshots"
         reloadCs["loadStyle"] = "fromDB"
         reloadCs["detailAssemLocationsBOL"] = ["001-001"]
```
youngmit, line 81: perhaps this is a bit idealistic and more reasonable in a world where a compiler would eliminate the runtime cost of doing so, but another option to avoid an unlock here could be to express the pattern of "these settings except for [something different]" functionally, like:

```
reloadCs = settings.Settings("something.yaml")
newSettings = reloadCs.withModifications({some: dict})
```

As alluded to above, the downside is the cost of copying the settings from reloadCs to make a brand new Settings object with just a handful of different settings. But at least we would be able to walk away knowing that reloadCs (and by extension anything that holds a reference to it) is not being mutated

john-science: Sorry, momentary paranoia here; that's out of scope for this ticket, right? This ticket was just to add a `DeprecationWarning`?

Okay, but, I like that idea. I think we also need to add some form of:

    reloadCs.save(filePath)

Because we are currently doing a `yamlize.dump()` to write YAML files in 12 different modules. And at some point we will want to add tooling to our YAML outputs, and that'll be easier to do in one place than 12.

armi/bookkeeping/snapshotInterface.py
john-science: Yeah, I am not entirely convinced by this `lock` process i have created.  I will do your `unlock()` context manager idea: it's better. 

I guess using the word "lock" here is throwing me for a loop. Because making this a truly thread safe lock would actually be a lot more work than just a little `DeprecationWarning`. Still, since our parallelization is multi-process, not multi-thread, I should probably not worry about it so much.

I'll go do that.

youngmit: Oh yeah, thread safety is way outside the scope.

armi/bookkeeping/tests/test_historyTracker.py
youngmit: I'd say this is about as in/out of scope as the lock()/unlock() methods. If we just want to do a deprecation warning (assuming it doesn't produce such a firehose of horror that it's not useful), then we can just implement that and not try to make any progress on the other side to fix it. Once we mark settings modification as bad behavior, the next steps are:
 - Decide when it _is_ okay to change settings (e.g. "Any time, as long as you are honest and `unlock()` them?", "Only for some period before we `lock()` them, after which they're locked forever?", or "Never; if you want to change settings, make a new Settings object?")
 - Refactor code that is currently modifying settings to play nice with the new approach/restrictions.

To not anticipate where we will end up with the above feels a little like that `numpy` deprecation warning discussed in #246, where they say effectively "This is going to go away, but we aren't telling you what it will be replaced with." So i think it's at least worth spitballing. Perhaps this is better discussion to have in the issue (#371)...

>doing a yamlize.dump() to write YAML files in 12 different modules

Do settings use yamlize, or just Blueprints?

> `reloadCs.save(filePath)`

See https://github.com/terrapower/armi/blob/0289b7573b2ac39c792b41e1ebfe300dfd918f7b/armi/settings/caseSettings.py#L272

youngmit (approved):
Alright, I think this is looking pretty good. Ran some cases with this and we don't get _too_ many warnings, and I think between those and what we have silenced with `_unlock()` we can start talking about what the new world will look like.

john-science:
> Alright, I think this is looking pretty good. Ran some cases with this and we don't get _too_ many warnings, and I think between those and what we have silenced with `_unlock()` we can start talking about what the new world will look like.

Awesome!  Yeah, my plan is something like this:

1. Give devs 1-2 weeks time before making settings immutable, so they have warning.
2. Throw together a list of places we are currently modifying settings and we can decide when are valid and which aren't.
3. Build the new settings API with immutable settings and deploy.

There will be time for group discussion/headed debate in (2) and (3) there.  :)

youngmit:
nvm, we already have #371

Issue #373, opened 2021-08-17
Somehow reduce the number of log files created by third party libraries


youngmit:
@john-science Did some cool work to make our logging stuff implicitly rub off on TPLs, but an unfortunate side effect of this is that we get lots of _empty_ log files, which I assume are being created when the library instantiates its `logging` object. Would it be possible to either automagically clean these up, or perhaps even lazily create the file on first write? something like that. Not super high priority, but would be really nice.

john-science:
@youngmit  Sure thing!

The obvious solution is:

> Delete the empty log file on the moment of creation. Easy.

If you prefer we act in a more lazy manner, I will look into it. As it happens, `logging.FileHandler` is exceptionally _not_ lazy. It pins the file itself to the I/O stream, so if the file doesn't exist at any point after the `FileHandler` is created, all logging will silently be lost. (I hate that it's silent.)  I think the solution would have to be creating a FileHandler the first time someone 'log()` is called. Which is maybe awkward, but probably also easy.

Any preference?

jakehader:
@john-science - the main ones I am seeing are from `matplotlib`:

![image](https://user-images.githubusercontent.com/26268658/129934362-2b265a04-03c9-4120-b0a9-bca2ed278395.png)

john-science:
@jakehader  That is super annoying.  Sorry, I wasn't seeing that in my usual in-use-test-case testing.  Though I was perhaps just not looking in the right folder?

In any case, there is a PR open to fix this: https://github.com/terrapower/armi/pull/376

jakehader:
@john-science, I'll check it out and let you know if it works on the case I'm running! Thanks for the quick fix.

john-science:
I believe PR #384 solved this issue. Can we close it?  @jakehader ?

jakehader:
I believe so! I haven't seen as many log files around with the changes, but I can reopen if I see them again.

PR #376, opened 2021-08-18
Cleanup empty logs from third party libraries

john-science:
This is in response to: https://github.com/terrapower/armi/issues/373

I decided to fix the problem (of lots of empty third-party logs) in two directions:

1. Added the `delay=True` to all of our custom `logging.FileHandler`s, to make our log file creation more lazy.
2. Removed all the empty log files that still exist at the end of a run.

Doing (1) meant doing some re-org of our `concatLogs()` function, but that's fine.

### DIFF
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -16,22 +16,34 @@ r"""
 This module handles logging of console output (e.g. warnings, information, errors)
 during an armi run.
 
-The default way of using the ARMI runLog is:
+The default way of calling and the global armi logger is to just import it:
 
 .. code::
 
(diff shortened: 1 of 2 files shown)


youngmit (approved):
I'm happy to merge this as-is, but did you want to play around with just the delay=True?

john-science:
> I'm happy to merge this as-is, but did you want to play around with just the delay=True?

I can play with that again, sure.  We have just hit so many edge cases with this logging, and the `delay=True` is so poorly documented I didn't want to risk another edge case. 

But, sure, I'll play with it again.

PR #377, opened 2021-08-18
Remove unused SINCE_LAST_DB_TRANSMISSION on the parameter definitions 

jakehader:
* Remove the SINCE_LAST_DB_TRANSMISSION since it was never unset when writing to the database.

* Update `toWriteToDB` to allow an implicit `None` to include writing all parameters that have been set `SINCE_ANYTHING`.

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -1145,7 +1145,7 @@ class Database3(database.Database):
             g = h5group[groupName]
 
         for paramDef in c.p.paramDefs.toWriteToDB(
-            parameters.SINCE_LAST_DB_TRANSMISSION
+            parameters.SINCE_ANYTHING | parameters.NEVER
         ):
             attrs = {}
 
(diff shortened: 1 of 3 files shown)


ntouran:
My read of this is that if the guarantee is not being respected before this change, then that represents a different bug in the way the DB system works. This fix seems to just bypass that underlying issue without fixing it, while also causing potentially large increases in database size for large runs (we should measure the impact). We should think about trying to get to the root of the issue.

youngmit:
In order to solve "the issue with the database system," we would need to:

- track specifically _which_ database a parameter has been previously been written to in order to justify eliding future writes to the _same_ database. This is doable and probably wouldn't be too onerous.
- actually go hunting for the previously-written parameter data whose write we elided, then make sense of whatever data is there. Even if we did this, _there would be not guarantee that the data for that previous time node is consistent with the current state of the Reactor_. For instance, shuffling fuel around without touching some parameter would require a full tree traversal to map old param data. And if new objects came into the reactor, its not possible anyways.

Ultimately the optimization is probably not worth it. *And* we aren't even ever clearing the SINCE_LAST_DATABASE_TRANSMISSION flag anyways, so we aren't using it in the first place. Combination 🤦‍♀️😅

ntouran (approved):

PR #378, opened 2021-08-19
Add code to distribute parameter assigned state

youngmit:
We were not explicitly communicating the state of parameter definitions' `.assigned` state, making behavior dependent on when parameters are set function strangely on higher-rank MPI nodes. This adds code to the DistributeStateAction that explicitly broadcasts this state to workers, so they have a consistent view of when parameters were set.

Specifically, this adds support for allowing worker nodes to create database files without skipping parameters because they think they have never been set.


ntouran (approved):
looks good to me.

PR #379, opened 2021-08-19
Document new logging features

john-science:
We have some new logging features that were already "documented" in the docstring of `runLog.py`, but that's not very visible. So this change also documents our logging abilities in the `/docs/` area as well.

(Also, I had a couple of extra unit tests lying around that I've been meaning to add, purely for coverage reasons. These aren't important or relevant, but every little bit helps.)


youngmit:
Bonus coverage, thanks!

PR #380, opened 2021-08-23
Cleanup - Unused Imports and settingsValidation tests

john-science:
This PR is just the leftover bits of cleanup I've done recently working on more interesting things:

* removing unused imports and variables
* removing code coverage from nearly unreachable code

I also added a couple of small unit tests to `settingsValidation.py`.

### DIFF
--- a/armi/tests/test_mpiActions.py
+++ b/armi/tests/test_mpiActions.py
@@ -111,7 +111,8 @@ if armi.MPI_SIZE > 1:
                         del original[key]
                     if key in current:
                         del current[key]
-                for key, val in original.items():
+
+                for key in original.keys():
                     self.assertEqual(original[key], current[key])
 
         def test_distributeReactor(self):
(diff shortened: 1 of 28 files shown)


armi/tests/test_mpiActions.py
```diff
@@ -114,7 +111,8 @@ def test_distributeSettings(self):
                         del original[key]
                     if key in current:
                         del current[key]
-                for key, val in original.items():
+
+                for key in original.keys():
```
youngmit, line 115: `pylint` might even complain that you could just iterate the dict without `.keys()`, but I find `.keys()` more readable. If it is complaining, should we consider silencing the suggestion?

youngmit (approved):
Not all heroes wear capes

armi/tests/test_mpiActions.py
john-science: Yeah, me putting `keys()` in there was definitely a style choice. And not one I'll fight over for a second, just a choice.

The `pylint` error here was actually that `val` was unnecessary.

PR #383, opened 2021-08-24
Various updates to runLog outputs to the standard output

jakehader:
* Add start of runLog to the case.py so that the verbosity can be set when a run begins versus waiting until an operator is created.

* Update to the `specifyInputs` labels in fuel handlers so that the stdout has the correct labels for shuffle logic and explicit shuffles files.

* Add more runLog messages around checking if the lattice physics interface will create XS for the current cycle. This helps a user to debug what is going on.

* Add check for `SameFileError` exception during the `copyOrWarn` path tools function. This skips providing a user warning if a file is being copied to the same path and just lets it happen silently.

### DIFF
--- a/armi/operators/operator.py
+++ b/armi/operators/operator.py
@@ -96,6 +96,7 @@ class Operator:  # pylint: disable=too-many-public-methods
         """
         self.r = None
         self.cs = cs
+        runLog.LOG.startLog(self.cs.caseTitle)
         self.timer = codeTiming.getMasterTimer()
         self.interfaces = []
         self.restartData = []
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -67,11 +67,11 @@ class FuelHandlerInterface(interfaces.Interface):
     @staticmethod
     def specifyInputs(cs):
         files = {
-            cs.settings[label]: [
-                cs[label],
+            cs.settings[setting].label: [
+                cs[setting],
             ]
-            for label in ["shuffleLogic", "explicitRepeatShuffles"]
-            if cs[label]
+            for setting in ["shuffleLogic", "explicitRepeatShuffles"]
+            if cs[setting]
         }
         return files
 
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -67,11 +67,11 @@ class FuelHandlerInterface(interfaces.Interface):
     @staticmethod
     def specifyInputs(cs):
         files = {
-            cs.settings[setting].label: [
-                cs[setting],
+            cs.settings[settingName].label: [
+                cs[settingName],
             ]
-            for setting in ["shuffleLogic", "explicitRepeatShuffles"]
-            if cs[setting]
+            for settingName in ["shuffleLogic", "explicitRepeatShuffles"]
+            if cs[settingName]
         }
         return files
 
--- a/armi/utils/pathTools.py
+++ b/armi/utils/pathTools.py
@@ -62,6 +62,8 @@ def copyOrWarn(fileDescription, sourcePath, destinationPath):
         runLog.debug(
             "Copied {}: {} -> {}".format(fileDescription, sourcePath, destinationPath)
         )
+    except shutil.SameFileError:
+        pass
     except Exception as e:
         runLog.warning(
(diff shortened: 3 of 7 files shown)


armi/operators/operator.py
```diff
@@ -96,6 +96,7 @@ def __init__(self, cs):
         """
         self.r = None
         self.cs = cs
+        runLog.LOG.startLog(self.cs.caseTitle)
```
youngmit, line 99: Does the case that _creates_ an operator not already call `startLog()`? If it does, are there side effects of calling it twice?

Starting to feel more and more that as we drift from the world of "everything interesting that ARMI can do is part of 'running a case'," the setup for logging and verbosity and stuff needs to get closer to `context`. Sure, maybe when a case starts running, or an operator gets built, or a `Settings` object gets created or whatever, we can use `[branch]Verbosity` (and future per-`Logger` verbosity dictionary setting) to tweak the `runLog` verbosities from their defaults.

@john-science I wonder what your thoughts are on this.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -67,11 +67,11 @@ def __init__(self, r, cs):
...
         files = {
-            cs.settings[label]: [
-                cs[label],
+            cs.settings[setting].label: [
+                cs[setting],
             ]
-            for label in ["shuffleLogic", "explicitRepeatShuffles"]
-            if cs[label]
+            for setting in ["shuffleLogic", "explicitRepeatShuffles"]
+            if cs[setting]
```
youngmit, line 74: Hecka minor, but while "setting" is way better than "label", it still seems like it might refer to a `Setting` instance. "settingName"?

armi/utils/pathTools.py
```diff
@@ -62,6 +62,8 @@ def copyOrWarn(fileDescription, sourcePath, destinationPath):
         runLog.debug(
             "Copied {}: {} -> {}".format(fileDescription, sourcePath, destinationPath)
         )
+    except shutil.SameFileError:
+        pass
```
youngmit, line 66: Huh... how has this not been blowing things up before? I suppose this only happens with absolute paths, but I would think we would avoid trying to copy them altogether

youngmit (commented):
I don't think we should be initializing the log components in the case code any more than I think we should be doing it in OperatorMPI. They both, in my opinion, are happening too late, and too... optionally? See my inline comment about this.

Since @john-science  is the last to look closely at the way logging is setup and what's going on in OperatorMPI/context/etc., I think it'd be good to have him as proper reviewer on this.

armi/operators/operator.py
john-science: On the Framework side, I'd like to say it is a minor change to move `startLog()` from `OperatorMPI` to `Operator`.

But, honestly, I can't at this exact second. What does this change downstream?  I've been looking around, and I can't be really confident just yet.  Still looking.

john-science: Okay, @jakehader  I have a question about your change.  In `armi.operators.factor()`, we call `getOperatorClassFromSettings`, which has this block, which is called directly from `case.py`:

```python
    if runType == RunTypes.STANDARD:                                                                                                                                                                               
        if armi.MPI_SIZE == 1:                                                                                                                                                                                     
            return Operator                                                                                                                                                                                        
        else:                                                                                                                                                                                                      
            return OperatorMPI 
```

Was this an intentional change?  It looks like we either will-or-won't get a `startLog()` in our `case` workflow, depending on if we are in an MPI run or not.

jakehader: Yeah seems like we should have a runLog if it's serial too so thats why I moved it

jakehader:
> I don't think we should be initializing the log components in the case code any more than I think we should be doing it in OperatorMPI. They both, in my opinion, are happening too late, and too... optionally? See my inline comment about this.
> 
> Since @john-science is the last to look closely at the way logging is setup and what's going on in OperatorMPI/context/etc., I think it'd be good to have him as proper reviewer on this.

This may be true, but right now the verbosity wasn't being set for the case. If there is a better place for the initialization we can move it there, but I believe a case needs to have a logger when started, otherwise we end up losing information until it is started (and when the verbosity is set which it isnt being updated without this change)

armi/operators/operator.py
jakehader: Also I haven't seen a problem with calling startLog multiple times, but maybe this isn't intended?

PR #384, opened 2021-08-24
 Cleanup empty logs from third party libraries

john-science:
This is in response to: #373

I decided to fix the problem (of lots of empty third-party logs) in two directions:

    Added the delay=True to all of our custom logging.FileHandlers, to make our log file creation more lazy.
    Removed all the empty log files that still exist at the end of a run.

Doing (1) meant doing some re-org of our concatLogs() function, but that's fine.


john-science:
Typo PR.  Ignore this.

PR #385, opened 2021-08-25
Updates to spontaneous fission data

jakehader:
Updates to the spontaneous fission data within `burn-chain.yaml` based on data from Table 5.1 of "Application Guide to Neutron Multiplicity Counting", Report Number: LA-13422-M

### DIFF
--- a/armi/resources/burn-chain.yaml
+++ b/armi/resources/burn-chain.yaml
@@ -1,4 +1,7 @@
+# Spontaneous fission data is added from Table 5.1 of
+# "Application Guide to Neutron Multiplicity Counting" - LA-13422-M
 AM241:
+- nuSF: 3.22
 - transmutation:
     branch: 1.0
     products:
@@ -39,6 +42,12 @@ AM241:
     products:
     - NP237
     type: ad
+- decay:
+    branch: 4.120055e-12
+    halfLifeInSeconds: 13652000000.0
+    products:
+    - LFP41
+    type: sf
 AM242G:
 - transmutation:
     branch: 1.0
@@ -225,7 +234,7 @@ BK249:
     - CF249
     type: bmd
 - decay:
-    branch: 4.70000e-10
+    branch: 4.755215e-10
     halfLifeInSeconds: 2.8512000000e+07
     products:
     - LFP41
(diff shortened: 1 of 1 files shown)


jakehader:
After reviewing the MCNP6.2 data on spontaneous fission, they provided a report for the spontaneous fission emission spectra data as well as data for the spontaneous fission `nu` and branching ratios.

I went through this report and updated some of our numbers to align with the values. To calculate the branching ratios, I took the total half-life and divided by the half-life reported for the spontaneous fission event. For this calculation I used the total half-lives that already existed in the `burn-chain.yaml` file. Additionally, based on reviewing the ENDF data, it appears that `nu` for Pu-239 should be 2.32 rather than 2.16 reported on this table (see: https://www.nndc.bnl.gov/sigma/getInterpreted.jsp?evalid=19424&mf=1&mt=452 from the Sigma database on https://www.nndc.bnl.gov/sigma/). 

https://permalink.lanl.gov/object/tr?what=info:lanl-repo/lareport/LA-13422-M

image

armi/resources/burn-chain.yaml
```diff
@@ -1,4 +1,7 @@
+# Spontaneous fission data is added from Table 5.1 of
+# "Application Guide to Neutron Multiplicity Counting" - LA-13422-M
```
john-science, line 2: I appreciate the reference.

It would be good to have a central place for references in the repo; just a MD/RST file with references that we can point to from code.

ntouran (approved):
Would be nice to have a more rigorous reference chain in the future so we can know for sure that we're updating from outdated or invalidated data to something better. As is, that's not clear. But that's at least partially because the existing data doesn't have a clear reference trail. So this is an improvement in that regard. 

Also I do like @john-science's suggestion of putting a rst reference in the docs somewhere so we can collect a full bibliography.

armi/resources/burn-chain.yaml
jakehader: Thanks @john-science - References for nuclear data would be good. Sometimes having the references as close to the source is helpful. Maybe we should modify the `burn-chain.yaml` structure to support references?

jakehader:
@ntouran  - Please see the comparison.

image

Issue #387, opened 2021-08-26
Docstrings in `basicShapes` classes erroneously include units


keckler:
Many of the method docstrings of various classes within the `basicShapes` and `complexShapes` module include units when the methods are actually unit-agnostic. For instance, the [`getComponentArea`](https://terrapower.github.io/armi/_modules/armi/reactor/components/basicShapes.html#Circle.getComponentArea) method of the `Circle` class says "Computes the area for the circle component in cm^2". Similar statements exist for other `ShapedComponent`  area and perimeter methods. 

The units in these docstrings should be removed because they are potentially confusing and do not describe anything about the method itself. Yes, when used in the way expected, wherein the user provides the underlying dimensions to a `ShapedComponent`  in units of cm, the method returns area in units of cm^2. But the method itself has no idea what the units of its arguments are, and therefore the computed area is returned in whatever units are passed to the method (i.e. no checks or conversions are performed to ensure that the value returned is in units of cm^2). It also does not matter that the dimensions can be thermally expanded, because `getThermalExpansionFactor` returns a unitless expansion factor and so unit compatibility is not an issue.

Because ARMI has an API, methods like `getComponentArea` are likely to be used in unexpected ways wherein it cannot be assumed that the dimensions passed to it are in a given set of units without additional checks being implemented.

PR #389, opened 2021-08-26
Fixing comments that shouldn't include fixed units.

john-science:
This is in response to: https://github.com/terrapower/armi/issues/387


john-science:
We've determined that these comments are accurate, and should remain.

PR #391, opened 2021-08-30
Writing a YAML file now correctly decides to write 

john-science:
In the past, if you read in a YAML file using a "lattice map" it
would get written out as both a "lattice map" and "grid contents".
Now we only write out the format we read in.


john-science:
@youngmit I ran in-use-test cases and did GUI testing manually, it all appears to be working. 

The only issue in making this happen is trying to pass information around in a `yamlize` object. The solution I came up with was to add a `readFromLatticeMap` boolean variable around, then explicitly bar it from being written to the final YAML file. This feels inelegant, but I couldn't find any elegant ways to interact with the logic inside `yamlize', so I feel pretty limited with making even small changes to the read/write YAML process in that library. 

Just FYI.

armi/reactor/blueprints/__init__.py
```diff
@@ -402,7 +402,7 @@ def _resolveNuclides(self, cs):
                 newNuclides = expansions[elemental]
                 # overlay code details onto nuclideFlags for other parts of the code
                 # that will use them.
-                # CRAP: would be better if nuclideFlags did this upon reading s.t.
+                # TODO: would be better if nuclideFlags did this upon reading s.t.
```
youngmit, line 405: :D

john-science: heh. Admittedly, that's not an important change, and I will happily revert it.

youngmit (changes requested):
This is great! Only request is to change the `save_to_stream` functions to `saveToStream`

🦆

john-science:
> This is great! Only request is to change the `save_to_stream` functions to `saveToStream`

ACK.  I am still not used to camel casing.  You're totally right, of course.

youngmit:
If we could wave a magic wand, I think we'd change over to snake in a heartbeat, but alas..

youngmit:
Going to assume that the coverage issue (504) is spurious, since the previous state passed just fine

PR #404, opened 2021-09-07
Using GitHub Actions to Auto-Deploy Docs

john-science:
The point of this PR is to ensure that we are always auto-deploying the latest docs to our public website.  And GitHub Pages makes this doable. Please see the new `docs.yaml` to see how I did this.

I also added a new `tox -e docs` that people can use to build the docs, and ensure that they have the right version of Sphinx and everything else to make that happen.  (Because Sphinx has some versioning issues.)

The bottom line is the docs now _build_ and _work_. Though they still throw a bunch of warnings and even a couple errors, but I figure we make it work first, then clean it up.

Please note that I tested if this auto-deploy works a bunch of times, and it seems to. In particular, I added the word "installed" and "TESTING" to `doc/user/user_install.rst`, and it auto-deployed a bunch of times correctly for my repos. See the example [here](https://antineutrino.net/terrapower.github.io/armi/installation.html).

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,0 +1,37 @@
+name: Documentation
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Update package index
+        run: sudo apt-get update
+      - name: Install mpi libs
+        run: sudo apt-get -y install libopenmpi-dev
+      - name: Install Pandoc
+        run: sudo apt-get -y install pandoc
+      - name: Install Tox and any other packages
+        run: pip install tox
+      - name: Make HTML Docs
+        run: tox -e doc
+        continue-on-error: true
+      - name: deploy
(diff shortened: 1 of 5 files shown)


john-science:
UPDATE: I have added the `ACCESS_TOKEN` to the TerraPower GitHub organization.  I believe this is sufficient to allow this PR to work.

Full Disclosure, I did a lot of testing to make sure it would work. But I now realize that I wasn't testing inside an organization. I don't see any reason that should matter.

youngmit:
Can we not dry run the action from a branch? You should have permissions to push to a feature branch on the honest-to-goodness terrapower/armi repo

john-science:
> Can we not dry run the action from a branch? You should have permissions to push to a feature branch on the honest-to-goodness terrapower/armi repo

Ah, I have permissions now! Yeah, I can give that a try.

I did a dozen dry runs on my own copies of these repos, so it's not untested. But I can test and see if the tokens work differently in an org, sure.

john-science:
@youngmit  I have confirmation that the deploy system works.  I made a copy of this PR's branch into the OG TerraPower ARMI repo and GitHub Actions auto-deployed the docs.  

Look here:  https://terrapower.github.io/armi/user/user_install.html#prerequisites

And notice that it says:

> You must have the following installed before proceeding:

But it used to say:

> You must have the following before proceeding:

The word "installed" is added in this PR, as a proof-of-concept.

Issue #403, opened 2021-09-03
Running CaseSuite.clone() deletes the repo and all adjacent directories


john-science:
When running the `CaseSuite.clone()` method, it can delete the entire repository, and all directories adjacent the repository.

I have reproduced this bug off master, and here are the exact steps you need to reproduce it yourself.

> Please be careful reproducing this bug on your own machine!

First, set up the env from scratch:

```bash
# make the temp directory and add a couple of files to it
mkdir -p ~/temp/safety/backup/be_careful/terrapower/
cd ~/temp/safety/backup/be_careful/terrapower/
touch things.txt
touch stuff.txt

# clone the repo and make the env
git clone https://github.com/terrapower/armi
cd armi/
python3 -m venv venv
source venv/bin/activate

# install the code
pip install -r requirements.txt 
pip install -r requirements-testing.txt 
pip install -e .
```

Then, just to prove everything is working, let's run some unit tests:

```bash
python -m pytest -n 4 armi/cases/tests/test_cases.py
```
That worked fine. Now add this text to `test_cases.py` unit test file:

https://github.com/terrapower/armi/blob/master/armi/cases/tests/test_cases.py#L131

```python
    def test_clone(self):
        clone = self.suite.test_clone("test_clone")
        print(clone)
        assert False
```

And run that test file again:

```bash
python -m pytest -n 4 armi/cases/tests/test_cases.py
```

That's it. The entire repo is now deleted, along with all the stuff adjacent to it:

```bash
cd ~/temp/safety/backup/be_careful/terrapower/
ls
# not only is `/armi/` gone, but so are `things.txt` and `stuff.txt`
```

youngmit:
Quickest way to find this might be to make some file read-only and try to reproduce and see where it stubs it's toe. Looking through the code, it's not immediately obvious where this might be happening.

@john-science for completeness sake, what platform are you running on? Shouldn't matter too much, but lots of the os/file system stuff in Python is surprisingly different between linux and windows.

john-science:
Update.

So, the problem is here:

https://github.com/terrapower/armi/blob/master/armi/cases/suite.py#L182

If the users specifies a non-existent directory as `oldRoot`, the variable `newDir` takes on the value `..`, and then everything gets wiped by the call to:

    directoryChangers.ForcedCreationDirectoryChanger(newDir, clean=True, dumpOnException=False)

So, I haven't decided the best way to fix it yet, but I think there are a couple things that need addressing:

[...]

PR #405, opened 2021-09-07
Protecting against calls to wiping entire directories

john-science:
First, the `clean=True` is removed from the `ForcedDirectoryChanger`. That is really what causes the issue, while this may break some people's workflows, having code around that recursively deletes directories is pretty scary, and those workflows should change.

Second, the pre-existing `cleanPath` method was moved to a better location and is now used in place of all bare calls to `shutil.rmtree`. It is scary that a bare 'shutil.rmtree` was called in 8 different places in the repo, with no safety checks.

Lastly, there is a whitelisted set of directories in `cleanPath` that can be wiped. In an ideal world, the codebase and our workflows would change to allow for only one: `/.armi/`.

NOTE: this PR relates to bug described in issue https://github.com/terrapower/armi/issues/403

### DIFF
--- a/armi/cases/suiteBuilder.py
+++ b/armi/cases/suiteBuilder.py
@@ -167,11 +167,11 @@ class SuiteBuilder:
                     )
 
                 previousMods.append(type(mod))
-                mod(case.cs, case.bp, case.geom)
+                with case.cs._unlock():
+                    mod(case.cs, case.bp, case.geom)
                 case.independentVariables.update(mod.independentVariable)
 
             case.cs.path = namingFunc(index, case, modList)
-
             caseSuite.add(case)
 
         return caseSuite
--- a/armi/utils/directoryChangers.py
+++ b/armi/utils/directoryChangers.py
@@ -193,9 +193,22 @@ class TemporaryDirectoryChanger(DirectoryChanger):
         DirectoryChanger.__init__(
             self, root, filesToMove, filesToRetrieve, dumpOnException
         )
-        root = root or armi.context.getFastPath()
+
+        # if no root dir is given, enforce that this tmp dir is in /.armi/
+        if not root:
+            root = armi.context.getFastPath()
+            if ".armi" not in os.path.normpath(root).split(os.path.sep):
+                raise ValueError("Temporary directory not found.")
+
+        # make the tmp dir, if necessary
         if not os.path.exists(root):
-            os.makedirs(root)
+            try:
+                os.makedirs(root)
+            except FileExistsError:
+                # ignore the obvious race condition
+                pass
+
+        # init the important path attributes
         self.initial = os.path.abspath(os.getcwd())
         self.destination = TemporaryDirectoryChanger.GetRandomDirectory(root)
         while os.path.exists(self.destination):
@@ -217,17 +230,12 @@ class TemporaryDirectoryChanger(DirectoryChanger):
 
     def __exit__(self, exc_type, exc_value, traceback):
         DirectoryChanger.__exit__(self, exc_type, exc_value, traceback)
-        shutil.rmtree(self.destination)
+        pathTools.cleanPath(self.destination)
 
 
 class ForcedCreationDirectoryChanger(DirectoryChanger):
     """
     Creates the directory tree necessary to reach your desired destination
-
-    Attributes
-    ----------
-    clean : bool
-        if True and the directory exists, clear all contents on entry.
     """
 
     def __init__(
@@ -236,14 +244,12 @@ class ForcedCreationDirectoryChanger(DirectoryChanger):
         filesToMove=None,
         filesToRetrieve=None,
         dumpOnException=True,
-        clean=False,
     ):
         if not destination:
             raise ValueError("A destination directory must be provided.")
         DirectoryChanger.__init__(
             self, destination, filesToMove, filesToRetrieve, dumpOnException
         )
-        self.clean = clean
 
     def __enter__(self):
         if not os.path.exists(self.destination):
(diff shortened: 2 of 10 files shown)


john-science:
1. @ntouran @jakehader  I expect this PR to be somewhat controversial because it might change someone's workflow. I hemmed and hawed about that for a long time, as I am generally reticent to do that. BUT my concern is that we have workflows around that involve automated tools recursively deleting directories with high permissions. (I may be convinced to support these workflows of course, but for obvious security reasons, I decided to start from a more conservative position.)
2. @ntouran Please note, I have removed the `clean=True` entirely from the change dirs methods. This is, indeed, not what you suggested this morning, it is more conservative. Again, of course, I can easily reverse that decision.
3. In a more ideal world, I would very much like to change the new `cleanPath` method to refuse to delete any data unless it is inside a `/.armi/` folder. How crazy does that sound to everyone?

jakehader:
Hi @john-science, would it be reasonable to have a setting that configures the allowable directory patterns to clean up that can be configured on an application or plugin basis? This would probably look like a kwarg on the implementation of DirectoryChangers but when used in an application we could pass this setting in. Maybe this is still dangerous though. 

The main question in this is whether or not the framework should support cleaning up a work space when it is done or not and if so, how should it be done "safely"?

It sounds like the current implementation is dangerous because you can just delete anything on accident. With this change we do lose the capability of cleaning up remote folders and files that are created and that could lead to excessive hard disk space consumption. Would it be appropriate to log the directories that get created in a run in a separate logger so that someone may choose to delete these directories manually? Or could we add that list of created directories to some global variable that gives access for removal? Globals are probably not a great solution, but just brainstorming...

john-science:
> ...would it be reasonable to have a setting that configures the allowable directory patterns to clean up...

Hey @jakehader , @ntouran suggested something similar. Since this code is called from `context.py`, it can be called BEFORE settings files are read, so that will be tricky.

I would be in favor of finding some way to keep the cluster from filling up that doesn't involve having code in the ARMI framework. This is an open-source project that should be usable by people who don't have access to our computing cluster, after all. For instance, the cluster could have a cleanup cron job.

If people really feel like the `rm -rf` HAS to be in ARMI, we will have to find some way of identifying a single folder that is safe to delete from for cluster purposes. For instance, all temp directories created for unit testing end up inside a directory called `/.armi/`.  If we could ensure that all temp directories were place inside a folder called `/.temp_armi_studies/` or something, that would keep laptop/desktop users safe.

Alternatively, we can try to make this new folder name configurable in settings, and find some way for `context` to default to not using it.

john-science:
@jakehader Could we move all "clean up the cluster" logic to the cluster plugin?

armi/cases/suiteBuilder.py
```diff
@@ -167,11 +167,11 @@ def namingFunc(index, _case, _mods):  # pylint: disable=function-redefined
                     )
 
                 previousMods.append(type(mod))
-                mod(case.cs, case.bp, case.geom)
+                with case.cs._unlock():
+                    mod(case.cs, case.bp, case.geom)
```
youngmit, line 171: Just as a heads-up for when we take next steps on immutable settings, this is a good example of where modifying a CaseSettings object is perfectly reasonable. The `case.cs` here are in the process of being created, so either they haven't been locked in the first place, or `case.cs = case.cs.modified(blah)` would be reasonable here

youngmit: ...I might even suggest leaving it as-is even, eating the warning until we figure out our next move

armi/utils/directoryChangers.py
```diff
@@ -193,9 +193,22 @@ def __init__(
         DirectoryChanger.__init__(
             self, root, filesToMove, filesToRetrieve, dumpOnException
         )
-        root = root or armi.context.getFastPath()
+
+        # if no root dir is given, enforce that this tmp dir is in /.armi/
```
youngmit, line 197: I'd say this could use a little more explanation of the issues that could be caused if this check weren't performed. _Why_ is it important to check this?

youngmit (changes requested):
This is great. Only thing i'd ask for is a bit more explanation about why we are doing those directory checks.

armi/cases/suiteBuilder.py
john-science: Actually, you originally asked me to suppress those warnings. And this was the reason I encountered the `clone()` bug.

@jakehader pointed out that when he ran a case study it started out with dozens of `DeprecationWarning`s about immutable settings before the run even started. And you guys both agreed that was a valid place to change settings, as it wasn't mid-run at all.  It was something I had missed in my first pass with `DeprecationWarning`s, so I went off to fix it.

Then things exploded, and here we are.

We certainly need one more PR to make the settings immutable. But since it might be a big, breaking change for some people, I decided to give people a couple weeks of watching the `DeprecationWarning`s roll by before we make the change go live.

youngmit: yeah, i did 🤦 dont change it back, but just bear in mind when we actually tackle immutable settings

john-science: Yeah, I am itching to work on making the settings fully immutable.  But trying to hold to my (completely arbitrary) two weeks notice situation.

Though... maybe I could get started on it this week, it will take a while for it to make it through review, I bet. Yeah, maybe I'll push that up the queue.

john-science:
@youngmit I added some comments to our directory checking. See what you think.

youngmit (approved):
so good, appropriately "here be dragons". Thanks!

PR #407, opened 2021-09-09
Make cleanPath work with Windows fast paths

youngmit:


john-science (approved):
I like using `APP_DATA` rather than `/.armi/`.  Mitch has convinced me.

Issue #400, opened 2021-09-02
Update `Case` to allow manually adding dependencies


jakehader:
When using the `Case` in `cases.py` the dependencies are auto-derived based on the plugins and some other settings like `explicitRepeatShuffles` in https://github.com/terrapower/armi/blob/b0c76656ec97c984272c1e9df0a7bb26db8045c9/armi/cases/case.py#L169-L180. There are a couple of challenges/problems with this that I think can be cleaned up:

1. The dependencies are auto-derived each time and there is no way to explicitly add your own dependency manually to the `Case` object. It would be great if there could be an `addDependency` method that a user could set explicitly. The use case for this is when I use the `copyFilesFrom` setting, I want to explicitly tell the case to wait until another case is completed in the suite before firing off.
2. The implementation of the `dependencies` method on `Case` relies on regular expressions to determine what the dependent cases are. This seems to work well for the scenarios that are implemented, but I had difficulties trying to extent this behavior within a neutronics plugin to using `copyFilesFrom` and an ISOTXS file. It _could_ be nice if this could use something like glob rather than regular expressions, but maybe this is just user/developer error because I am not good with regular expressions.
3. Finally, because of the above issues I ended up subclassing the `Case` to add dependencies manually. When doing this I noticed that the case objects stored on the `Suite` are private (i.e., `suite._cases`). Ideally there would be a way to access these through the `Suite` API with some `__getitem__` or other another method.

Just for completeness, here is the subclass I made, but I don't recommend this if we can fix/update this.

```
class CaseWithManualDependency(cases.Case):
    """
    Manually overwrite the dependency treatment in a general Case
    so that the dependencies can be set explicitly rather than implicitly
    based on the case settings.
    """
    
    def __init__(self, cs, caseSuite=None, bp=None, geom=None):
        cases.Case.__init__(self, cs, caseSuite, bp, geom)
        self._dependencies = []
        
    def setDependency(self, depCase):
        runLog.info(f"Setting dependency of {depCase} for {self}")
        self._dependencies = [depCase]
        
    @property
    def dependencies(self):
        return self._dependencies
 ```

This is outside the scope of this issue, but another thing to consider is updating the `Suite` to hold the dependencies for the cases rather than the `Case` objects themselves.

jakehader:
@youngmit, could you make another PR that provides a mechanism to get a case from a suite? Right now the Suite's cases are in `_cases` without a clear way to get them without just manually printing them and figuring out what the index is and then using it.

[...]

PR #408, opened 2021-09-09
Add support for explicit Case dependencies

youngmit:
This is in response to #400. We probably want to ultimately get to a place where Case dependencies are managed from the Suite side, but this should compose well with that (case.addDependency() could still function by interacting under the hood with it's parent Suite), and solves the immediate issue of supporting simple, explicit Case dependencies.

@jakehader, could you shake this down, see if it fits your needs?

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -213,7 +213,7 @@ class Case:
         with the other, implicitly discovered, dependencies.
         """
         if case in self._dependencies:
-            runLog.warn(
+            runLog.warning(
                 "The case {} is already explicity specified as a dependency of "
                 "{}".format(case, self)
             )
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -33,7 +33,7 @@ import trace
 import time
 import textwrap
 import ast
-from typing import Dict, Optional, Sequence
+from typing import Dict, Optional, Sequence, Set
 import glob
 import tabulate
 import six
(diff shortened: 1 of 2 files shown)


armi/cases/case.py
```diff
@@ -196,10 +197,28 @@ def dependencies(self):
...
+        using the current case settings and plugin hooks. Sometimes, it is necessary to
+        impose dependencies that are not expressed through settings and hooks. This
+        method stores another case as an explicit dependency, which will be included
+        with the other, implicitly discovered, dependencies.
+        """
+        if case in self._dependencies:
+            runLog.warn(
+                "The case {} is already explicity specified as a dependency of "
+                "{}".format(case, self)
+            )
```
jakehader, line 219: Is a warning necessary here if we are using a set for `self._dependencies`? Also, is this supposed to be `runLog.warn` or `runLog.warning`? I don't have the code in front of me so just checking.

jakehader (commented):
I didn't test it out, but this would appear to fix the subclassing issue that I put in #400. I'll approve once you respond to my comment. Thanks!

armi/cases/case.py
youngmit: I think it's better than not having it. Re-adding a dependency that's already there is a pretty good sign that the caller is not 100% sure what they're doing. Notice that this does not produce a warning if there is overlap between what's stored in `_dependencies` and what cases get discovered through evaluation of the `dependencies` property.

Good catch on `warn` fixed it.

john-science:
I like this. It seems like a small change with a direct purpose.

jakehader (approved):

PR #409, opened 2021-09-10
Remove calls to getMasterCs in plotting

youngmit:


john-science (approved):
I think it looks good.  A straight-forward way to ditch the `getMasterCs()` calls that is easy to understand and read.

I mean... "problome" in a couple places, but still.

youngmit:
Lol, I'll fix that before I merge. Thanks for the reminder

PR #412, opened 2021-09-13
Fix to block diagram plotting in new reports with custom materials

jakehader:
Update to handle custom material names in the block diagram creation and to fix the memory issue with creating a large subplot (i.e., 50 x 50).


youngmit (approved):
There is some opportunity for cleaning up some things here, but I can tackle those as I integrate more of the new reports code. Thanks, @jakehader !

PR #413, opened 2021-09-14
Add debug messages to uniform mesh converter

jakehader:
* Replace the existing messages with more information for a user to verify that the
axial homogenization of blocks within the `UniformMeshGeometryConverter` is
being performed correctly. 

* Add verification check that the block heights calculated in `getBlocksBetweenElevations` have a cumulative height equal to the height from the user-supplied `zUpper` and `zLower` values.

### DIFF
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -149,6 +149,7 @@ class UniformMeshGeometryConverter(GeometryConverter):
         applyStateToOriginal : basically the reverse on the way out.
         """
         newAssem = UniformMeshGeometryConverter._createNewAssembly(sourceAssem)
+        runLog.debug(f"Creating a uniform mesh of {newAssem}")
         bottom = 0.0
         for topMeshPoint in newMesh:
             overlappingBlockInfo = sourceAssem.getBlocksBetweenElevations(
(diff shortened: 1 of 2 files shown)


armi/reactor/converters/uniformMesh.py
```diff
@@ -162,37 +163,55 @@ def makeAssemWithUniformMesh(sourceAssem, newMesh):
...
+                    runLog.debug(
+                        f"  - {sourceBlock} accounts for {sourceBlockHeightOverlap/totalHeight * 100.0:<5.2f}% of the homogenized region"
+                    )
+                else:
+                    totalHeight = topMeshPoint - bottom
+                    runLog.debug(
+                        f"  - {potentialBlock} accounts for {heightOverlap/totalHeight * 100.0:<5.2f}% of the homogenized region"
+                    )
+
+            newXSType = sourceBlock.p.xsType if specialXSType is None else specialXSType
```
youngmit, line 208: At the very least, the title of this PR/commit will need to change. This is more than just debug statements

youngmit (changes requested):
Should this module get its own runLog object to allow granular verbosity in the future?

armi/reactor/converters/uniformMesh.py
jakehader: My understanding is that this should behave the same as before but it's more explicit rather than relying on the assignment of the XS type just from the deepcopy. Maybe this logic is too complex as it is though and hard to understand.

youngmit: Okay. if it's "supposed" to come from the copied assem, I'm happy

PR #415, opened 2021-09-15
Fix recovery of temporary files when except

youngmit:
DirectoryChanger was using shutil.copy to move everything back in the
event of failure. shutil.copy() does not work with directories, leading
to exceptions when there were any directories under the temp dir. this
switches to using plain-old copytree() instead.


john-science (approved):
Looks great!

PR #416, opened 2021-09-15
Clean up report code

youngmit:
This includes a handful of cleanups of the newReports code, mostly
related to things like:
 - general code organization
 - non-standard import locations
 - documentation

In addition, the Case.summarizeDesign() method is changed to use new
reports functionality, rather than needing to create an operator and
interact with the report interface.

### DIFF
--- a/armi/cases/tests/test_cases.py
+++ b/armi/cases/tests/test_cases.py
@@ -89,7 +89,11 @@ class TestArmiCase(unittest.TestCase):
             case = cases.Case(cs)
             c2 = case.clone()
             c2.summarizeDesign(True, True)
-            self.assertTrue(os.path.exists("Core Design Report.html"))
+            self.assertTrue(
+                os.path.exists(
+                    os.path.join("{}-reports".format(c2.cs.caseTitle), "index.html")
+                )
+            )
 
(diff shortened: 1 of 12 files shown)


armi/cases/tests/test_cases.py
```diff
@@ -89,7 +89,11 @@ def test_summarizeDesign(self):
             case = cases.Case(cs)
             c2 = case.clone()
             c2.summarizeDesign(True, True)
-            self.assertTrue(os.path.exists("Core Design Report.html"))
+            self.assertTrue(
+                os.path.exists(
```
john-science, line 93: We don't want to test the file contents at all? Look for a set string we added, or test the length of the file contents, to show it isn't empty?

john-science (approved):
I like the changes, they reduce clutter and complexity a lot.  And I like the reduction in `caseSettings.py`. There is a bit of iffy whitespace in a docstring in `plugins.py`, but oh well.

Maybe we should create a rule that the code coverage can never go down (that _actually_ fails a PR), that way every PR people would have to add a single unit test to make sure the coverage doesn't go down by two lines.  Just a thought.

Issue #417, opened 2021-09-20
Some gallery examples are broken


mgjarrett:
Six of the gallery examples show a "broken" thumbnail.

https://terrapower.github.io/armi/gallery/index.html

Related to #394; some of these examples are calling a `sefl._log()` with arguments that are consistent with the ARMI logger where a standard python logger is expected.

```
Traceback (most recent call last):
  File "/home/runner/work/armi/armi/doc/gallery-src/framework/run_fuelManagement.py", line 31, in <module>
    o, reactor = test_reactors.loadTestReactor(inputFileName="refTestCartesian.yaml")
  File "/home/runner/work/armi/armi/armi/reactor/tests/test_reactors.py", line 176, in loadTestReactor
    o = operators.factory(cs)
  File "/home/runner/work/armi/armi/armi/operators/__init__.py", line 52, in factory
    return getOperatorClassFromSettings(cs)(cs)
  File "/home/runner/work/armi/armi/armi/operators/operator.py", line 110, in __init__
    reportingUtils.writeWelcomeHeaders(self, cs)
  File "/home/runner/work/armi/armi/armi/bookkeeping/report/reportingUtils.py", line 201, in writeWelcomeHeaders
    _writeCaseInformation(o, cs)
  File "/home/runner/work/armi/armi/armi/bookkeeping/report/reportingUtils.py", line 85, in _writeCaseInformation
    runLog.header("=========== Case Information ===========")
  File "/home/runner/work/armi/armi/armi/runLog.py", line 384, in header
    LOG.log("header", msg, single=single, label=label)
  File "/home/runner/work/armi/armi/armi/runLog.py", line 163, in log
    self.logger.log(msgLevel, msg, single=single, label=label)
  File "/opt/hostedtoolcache/Python/3.8.11/x64/lib/python3.8/logging/__init__.py", line 1512, in log
    self._log(level, msg, args, **kwargs)
TypeError: _log() got an unexpected keyword argument 'single'
```

john-science:
@mgjarrett  I opened a PR for this:  https://github.com/terrapower/armi/pull/418

john-science:
PR #418 resolved this ticket.

PR #418, opened 2021-09-20
Fixing the example gallery in our docs.

john-science:
This is in response to: https://github.com/terrapower/armi/issues/417

It was actually a combination of a few things:

1. Our docs had gotten a bit out of date and had some issues, that we started addressing recently
2. The logging change broke a few of the API examples run by ARMI, but that didn't break the build
3. One of the examples wasn't actually broken, but it didn't generate an image, so a thumbnail popped up that said "BROKEN".

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -26,7 +26,6 @@ jobs:
         run: pip install tox
       - name: Make HTML Docs
         run: tox -e doc
-        continue-on-error: true
       - name: deploy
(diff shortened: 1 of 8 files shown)


youngmit (approved):
Not sure what's up with the Coveralls reporting; hopefully it was spurious.

youngmit:
> But you said you found a fix for it on our end?

Nope, was just hoping it was their fault. Sounds like it was :)

PR #419, opened 2021-09-21
Added extension to profiling instructions

ntouran:
The profiling page wasn't getting added to the doc build because it
didn't have the right extension. This fixes it so now it will be
included in the doc build.


youngmit:
Sorry swooped you, @john-science. 😁

Issue #424, opened 2021-09-27
Bug in `applyInputParams` of multiple material classes


keckler:
The `applyInputParams` methods of multiple material classes contain a common bug when applying enrichments to particular isotopes/elements. Take, for example, adjusting the enrichment of B10 within the [B4C class](https://terrapower.github.io/armi/.apidocs/armi.materials.b4c.html):

```
 def applyInputParams(
        self, B10_wt_frac=None, theoretical_density=None, *args, **kwargs
    ):
        if B10_wt_frac:
            # we can't just use the generic enrichment adjustment here because the
            # carbon has to change with enrich.
            self.adjustMassEnrichment(B10_wt_frac)
```

If the user intentionally specifies `B10_wt_frac=0`, as in they want their absorber to have only B11, the enrichment will not be adjusted because the conditional evaluates to `False`. 

This same bug applies to other material classes, including [UO2](https://terrapower.github.io/armi/_modules/armi/materials/uraniumOxide.html#UraniumOxide.applyInputParams), and possibly others.

youngmit:
Looks like this just needs an `is not None` to disambiguate the falsiness of zero (implicit truth checks of numbers is a pretty big footgun, as this demonstrates).

@keckler would you mind filing a PR to fix?

john-science:
Does this bug only apply in this one file?  I see several material have the `applyInputParams` method.

jakehader:
It appears that it's only in one material but the issue may be because B4C has a different constructor. It would be nice if the API for the materials with these applyInputParams was consistent. Maybe in the past we had kwargs but broke them out into specific args? I don't remember the history off hand.

keckler:
@jakehader I think it is more than just the one material. I just scanned through the material classes in the framework, and in addition to B4C, it looks like the following material classes suffer from the same bug. Let me know if you disagree, maybe you see something that I don't.

[UO2](https://terrapower.github.io/armi/_modules/armi/materials/uraniumOxide.html#UraniumOxide.applyInputParams)
[MOX](https://terrapower.github.io/armi/_modules/armi/materials/mox.html#MOX.applyInputParams)
[Sulfur](https://terrapower.github.io/armi/_modules/armi/materials/sulfur.html#Sulfur.applyInputParams)
[ThU](https://terrapower.github.io/armi/_modules/armi/materials/thU.html#ThU.applyInputParams)

But I do agree that all the materials seem to apply the input parameters in different ways. 

[...]

PR #425, opened 2021-09-28
Fix applyInputParams bug when values are zero

keckler:
See #424


john-science (approved):
This is a clear and easy-to-understand solution to a defined problem, and I like the light-weight touch of the solution.

john-science:
Let's start adding unit tests when we do PRs. 

I know they aren't exciting work. But if the problem a PR fixes was important enough to have a discussion and a bug fix PR, it's probably important enough for a single unit test in the repo, right?

(Side Note: I'm just pushing unit tests because our code coverage is still a bit low.)

PR #427, opened 2021-09-30
Latin hypercube suitbuilder

LMikeH:
This pull request adds a Latin Hypercube suite builder for both continuous and discrete variables. This method is used to provide a more efficient sampling of the design space. LHS more efficiently samples the space evenly across dimensions compared to random sampling. It requires fewer points than a full factorial since it samples quasi-randomly into nonoverlapping partitions. It is recommended to use a surrogate model with the sampled data to get the full benefit. 

This was implemented by inheriting the SuiteBuilder class. It differs from the FullFactorialSuiteBuilder in that it requires SamplingInputModifier objects that include parameter type (continuous or discrete) and bounds. Also, each instance of an InputModifier must have a unique name that relates to an input design parameter. An InputModifier is not a single case, but relates to a certain input variable.

### DIFF
--- a/armi/cases/tests/test_suiteBuilder.py
+++ b/armi/cases/tests/test_suiteBuilder.py
@@ -54,6 +54,13 @@ class TestLatinHyperCubeSuiteBuilder(unittest.TestCase):
             assert 0 < mod[0].value < 1e6
             assert mod[1].value in [0.0, 0.2, 0.4, 0.6, 0.8]
 
+    def test_addDegreeOfFreedom(self):
+        builder = LatinHyperCubeSuiteBuilder(case, size=20)
+        powerMod = LatinHyperCubeModifier("power", "continuous", [0, 1e6])
+        morePowerMod = LatinHyperCubeModifier("power", "continuous", [1e3, 1e5])
+
+        with self.assertRaises(ValueError):
+            builder.addDegreeOfFreedom([powerMod, morePowerMod])
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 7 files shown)


armi/cases/tests/test_suiteBuilder.py
```diff
@@ -16,6 +16,51 @@
 Unit tests for the SuiteBuilder
 """
 import unittest
+from armi.cases.inputModifiers.inputModifiers import SamplingInputModifier
+from armi import cases, settings
+from armi.cases.suiteBuilder import LatinHyperCubeSuiteBuilder
+
+
+cs = settings.Settings("armi/cases/tests/anl-afci-177.yaml")
+case = cases.Case(cs)
```
youngmit, line 25: loading the settings like this probably why you had to copy the blueprints and other files into the test directory. it would be nice if we could avoid that altogether (at least from the repository's point of view). could you try copying the inputs as part of the test setup, from within a TemporaryDirectoryChanger (https://github.com/terrapower/armi/blob/d1fc198dc422faa53151791b13485c24f497ad81/armi/utils/directoryChangers.py#L195)?

See https://github.com/terrapower/armi/blob/07a06e6e35d85d6be6bab4d95887007ad6a09173/armi/bookkeeping/tests/test_databaseInterface.py#L88 and https://github.com/terrapower/armi/blob/eeb28b46de305cf2477ca92fd2aec811704fdd13/armi/cases/tests/test_cases.py#L107 as examples

youngmit (changes requested):
This looks great! my only beef is the duplication of the input files for the tests. see in-line comments for explanation.

PR #428, opened 2021-10-01
Code Coverage - LatticePhysicsInterface & reportingUtils

john-science:
Adding code coverage to: `LatticePhysicsInterface` and `reportingUtils`, which were both at or near 0%.

And it turns out `HistoryProcessor` had 0% code coverage because it's not used in ARMI, TPAMI, or NALA, so it was removed.


youngmit (approved):
Thanks!

PR #429, opened 2021-10-02
Improving Code Coverage and Pylinting

john-science:
Increasing code coverage for:

* MacroXSGenerationInterface
* latticePhysics
* isotopicDepletionInterface
* neutronics.crossSectionTable
* plotting.py

Reducing pylint errors:

* unit tests
* fixing strange imports
* correctly using dict.items()

### DIFF
--- a/armi/bookkeeping/historyTracker.py
+++ b/armi/bookkeeping/historyTracker.py
@@ -78,7 +78,7 @@ from matplotlib import pyplot
 import tabulate
 
 from armi import interfaces
-import armi.runLog as runLog
+from armi import runLog
 from armi import utils
 from armi import operators
 from armi.utils import textProcessors
--- a/armi/utils/tests/test_gridGui.py
+++ b/armi/utils/tests/test_gridGui.py
@@ -1,4 +1,4 @@
-# Copyright 2021 TerraPower, LLC
+# Copyright 2021 Google, LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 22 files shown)


john-science:
@youngmit I had a slow morning. Here's a cleanup PR. Don't judge me.

armi/bookkeeping/historyTracker.py
```diff
@@ -78,7 +78,7 @@ def getHistoryParams(self):
 import tabulate
 
 from armi import interfaces
-import armi.runLog as runLog
```
youngmit, line 81: 😶

armi/utils/tests/test_gridGui.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2021 Google, LLC
+# Copyright 2021 TerraPower, LLC
```
youngmit, line 1: not sure if we can actually do this. original test was from @scottyak at Google. I think if you've added something you can put TerraPower in there, too, but dont think we can remove Google. Might do some research to be sure.

youngmit (changes requested):
Looks good, but my main concern is changing the copyright notice. I'll get back to you on that.

armi/utils/tests/test_gridGui.py
john-science: OH, is that what happened? 

I just assumed this was a typo.

Google has partially ownership of the copyright of ARMI?  That doesn't seem right.

So, any developer on ARMI can copyright a portion of the codebase for themselves?  That's problematic.  What do we do if a developer (or in this case Google) decides to flex their copyright?

This is honestly much stranger than what I originally assumed was just a typo. huh.

john-science: As I said, I thought this was just a typo. I can change it back.

PR #430, opened 2021-10-04
Update material modifications to use consistent keyword for theoretical density

keckler:


armi/materials/sulfur.py
```diff
@@ -24,9 +25,23 @@
 class Sulfur(material.Fluid):
     name = "Sulfur"
 
-    def applyInputParams(self, sulfur_density_frac=None):
+    def applyInputParams(self, sulfur_density_frac=None, TD_frac=None):
```
john-science, line 28: ARMI applies "camelCase" variable names everywhere through the repo. So these variable names should really be `sulfurDensityFrace` and something like `tdFrac`, to match there rest of the repo.

keckler: Hey John, thanks for the comment. I agree with you in general, but for whatever reason that is not the case for the material modifications at this point in time, to my understanding. I encourage you to check out the `applyInputParams` method of any of the [material classes](https://terrapower.github.io/armi/.apidocs/armi.materials.html).

Whether or not that should be changed in the future is a bigger question, but for now it seems to me that the material modifications do not use camelCase.

Let me know if you still disagree after looking into it.

john-science (approved):
Looks good to me.

Thanks for the unit tests!

Issue #420, opened 2021-09-23
Exporting a core with third-core symmetry assembly locations exports lattices in third-core symmetry as well


drewejohnson:
Using the input files from #251, I've stumbled across a strange bug/feature. Say I want to model a hexagonal reactor where fuel assemblies have some hexagonal pin lattice. The model is solid and plays well with ARMI. Now I want to model the core using 1/3 symmetry, and make a core map like
![image](https://user-images.githubusercontent.com/19477741/134568435-2645e45f-1b1b-4781-bc7d-c1db2936f545.png)

following

1. Launch the grid gui using the blueprints file that contains the core map from #251, including fuel lattices
2. Create new blueprints in the gui using 1/3rd symmetry with periodic boundary conditions
3. Make the core layout like above
4. Save new blueprints to produce

```yaml
core:
  geom: hex
  lattice map: |
    F
     F
    F F
     F
    F F
  grid bounds:
  symmetry: third periodic
  grid contents:
pins:
  geom: hex
  lattice map: |
    -   FP
      FP  FP
        CL  CL
      FP  FP  FP
        FP  FP  FP
          CL  CL
        FP  FP  FP
          FP  FP
        CL  CL  CL
  symmetry: full
  grid contents:
```
The core lattice has symmetry `third periodic` like expected, but the fuel lattice is now printed as if it used `third periodic` too. This may be acceptable if my fuel lattice has 1/3rd symmetry, but it's also a silent change. And the pin lattice has `symmetry: full`, which could lead to some very strange behavior.

The workaround is to manually copy back the original pin lattice so it's not a pain. Could be problematic if you don't inspect the generated blueprint file though

drewejohnson:
I'm using 5b7d215cc023098fd49e51d8c51336c315e9daad (0.1.6) for this replication. I found that later versions like 5d210605c8ef376fdf13cd1fb9ffd29ce63edff9 (current master HEAD) produced issues exporting the blueprints
```
Cannot write geometry with asciimap. Defaulting to dict. Issue: Cannot write asciimaps with blank rows from pure data yet.
```
produced
```yaml
core:
  geom: hex
  lattice map:
  grid bounds:
  symmetry: third periodic
  grid contents:
    ? - 0
      - 0
    : F
    ? - 0
      - 1
    : F
    ? - 1
      - 0
    : F
    ? - 2
      - -1
    : F
    ? - 0
      - 2
    : F
    ? - 1
      - 1
    : F
    ? - 2
      - 0
[...]

PR #439, opened 2021-10-07
Clean up grid blueprint serialization

youngmit:
- fix logic in which form to output
 - remove grid argument to saveToStream; it was causing all output grid
   blueprints to use the same geom settings, and wasnt really needed in
   the first place
 - add tryMap optional arg to attempt lattice map output regardless of
   how grid layout was read in.
 - add some more testing
 - change behavior of grid editor to prefer lattice map


youngmit:
This should resolve #420

PR #440, opened 2021-10-09
Fix asciimap ability to write full tips-up hex maps

### LINKED ISSUE #437: Asciimap implementation for full hex, flats up is broken for ij to ascii
The tests pass, because they dont actually create a new asciimap object for testing the round trip. Instead it re-uses the indices that come from reading the ascii form in the first place. Changing the test to look more like this exposes the issue:
```
    def test_hexFull(self):
        """Test sample full hex map against known answers."""
        # hex map is 19 rows tall, so it should go from -9 to 9
        asciimap = asciimaps.AsciiMapHexFullTipsUp()
        with io.StringIO() as stream:
            stream.write(HEX_FULL_MAP)
            stream.seek(0)
            asciimap.readAscii(stream.read())

        asciimap2 = asciimaps.AsciiMapHexFullTipsUp()

        for ij, spec in asciimap.items():
            asciimap2[ij] = spec

        with io.StringIO() as stream:
            asciimap2.gridContentsToAscii()
            asciimap2.writeAscii(stream)
            stream.seek(0)
            output = stream.read()
            self.assertEqual(output, HEX_FULL_MAP)
```

ntouran:
Fixes #437

### DIFF
--- a/armi/reactor/blueprints/gridBlueprint.py
+++ b/armi/reactor/blueprints/gridBlueprint.py
@@ -597,15 +597,6 @@ def saveToStream(stream, bluep, full=False, tryMap=False):
 
             if aMap is not None:
                 gridDesign.gridContents = None
-                if not isinstance(aMap, asciimaps.AsciiMapHexFullTipsUp):
-                    # this specific implementation of asciimap is broken, so we only
-                    # attempt to use the other implementations. If hex, full, tips-up is
-                    # being used we just preserve whatever was in the blueprints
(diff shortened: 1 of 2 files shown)


youngmit:
Sorry looks like you already have the test, my bad. Could you remove the special-case back that one did over here: https://github.com/terrapower/armi/commit/24708bc648671bf7b0fa793032b86ebf6cb44341#diff-4d90b47b8192a6689888e6cf841c450ac64f9e4c35c5e6579ba9b9147a2abf90R600

PR #441, opened 2021-10-11
Small edits to tutorial documentation.

mgjarrett:
I noticed a couple of small things that could/should be changed when scanning through the tutorial documentation recently. I think these are too small to warrant a formal issue, but it is still worth making the changes.

### DIFF
--- a/armi/tests/tutorials/data_model.ipynb
+++ b/armi/tests/tutorials/data_model.ipynb
@@ -12,7 +12,7 @@
     "\n",
     "## Initializing and Exploring the ARMI Model\n",
     "First we need to get some inputs. We built some from scratch in\n",
-    "[Building input files for a fast reactor](https://terrapower.github.io/armi/tutorials/walkthrough_inputs.html)\n",
+    "[Building input files for a fast reactor](walkthrough_inputs.html)\n",
     "and we pick those up and use them here as well:\n"
    ]
   },


john-science:
First off, don't worry about an issue for something like this. Our PRs represent a good traceable record, as do our commits.

And thanks so much for the cleanup!

I haven't tried to build your change _yet_, but one thing pops out at me.  I'm not certain, but could we change this:

```markdown
[Building input files for a fast reactor](https://terrapower.github.io/armi/tutorials/walkthrough_inputs.html)
```

to this?

```markdown
[Building input files for a fast reactor](walkthrough_inputs.html)
```

I am just thinking the relative path would be safer in case someone builds and deploys the documentation locally for testing.

john-science:
> Looks good to me. I initially attempted to link to the page using a reference in Sphinx but it wasn't working, so I just went with a URL. I didn't know this type of relative path would work, but I'm all for it.

Sorry, just to be clear: I haven't actually tried it, I just assume it works.  Have you tried making the docs yourself?

    tox -e doc

mgjarrett:
I did try myself to confirm it would work before pushing. I didn't try `tox -e doc`, but it did work with `make html`. I'll try `tox` now, but it should work.

john-science:
Merged!

Thanks @mgjarrett !

PR #446, opened 2021-10-15
Checking that each .py file has a license at the top, with GH Actions

john-science:
In the process I had to add a license to a bunch of files.

(Also, I am adding a couple of unrelated unit tests for code coverage purposes.)


john-science:
Just realized I still had the GH Action YAML file set to my test branch.

The branch should be good to go now.

PR #449, opened 2021-10-20
BlockNumberModifier class seems defunct

mgjarrett:
In `geometryConverters.py`, the `BlockNumberModifier` class doesn't seem to work. There is no unit test for it currently. I tried adding a unit test for the `.convert()` method, but it ends up calling a `block.duplicate()` which throws a `NotImplementedError`.

https://github.com/terrapower/armi/blob/190c0caccfd36a0442118a1ae666c2a20c52a4d1/armi/reactor/converters/geometryConverters.py#L229-L231

Unit test output:
```
  File "C:\Users\mjarrett\codes\armi\armi\reactor\converters\tests\test_geometryConverters.py", line 145, in test_BlockNumberModifier
    converter.convert(self.r)
  File "C:\Users\mjarrett\codes\armi\armi\reactor\converters\geometryConverters.py", line 161, in convert
    self.setNumberOfBlocks(a, blockFlags=a[fuelI].p.flags)
  File "C:\Users\mjarrett\codes\armi\armi\reactor\converters\geometryConverters.py", line 230, in setNumberOfBlocks
    newB = newBlockStack[-1].duplicate()  # copy the last fuel block.
  File "C:\Users\mjarrett\codes\armi\armi\reactor\composites.py", line 461, in duplicate
    raise NotImplementedError
NotImplementedError
```

Back in March, I had made some changes to make this method to get work correctly. But my changes altered the fundamental assumptions of the `.convert()` method and how it refines blocks, so I held off because I didn't know whether it would affect other code that depends on this. I was just revisiting it, and since the method doesn't seem to work anyway, I figure I may as well offer this pull request.

Let me know what your thoughts are on the new refinement method. The big difference is that Instead of adding `numToAdd` blocks to a fuel region, I changed it to divide each fuel block into `refinement` blocks.

### DIFF
--- a/armi/reactor/converters/tests/test_geometryConverters.py
+++ b/armi/reactor/converters/tests/test_geometryConverters.py
@@ -131,24 +131,6 @@ class TestGeometryConverters(unittest.TestCase):
         self.assertAlmostEqual(5, len(oneLine))  # same here
 
 
-class TestBlockNumberModifier(unittest.TestCase):
-    def setUp(self):
-        self.o, self.r = loadTestReactor(TEST_ROOT)
-        self.cs = settings.getMasterCs()
-
-    def test_BlockNumberModifier(self):
-        r"""
-        Tests that the addRing method adds the correct number of fuel assemblies to the test reactor
-        """
-        converter = geometryConverters.BlockNumberModifier(self.cs)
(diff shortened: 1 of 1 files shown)


john-science:
I love deleting code!

PR #458, opened 2021-10-28
Warning user when a Material Modification is unused.

### LINKED ISSUE #402: No warning when material modification is unused
Right now if you define an assembly with a material modification that is not recognized, ARMI does not tell you that it isn't recognized or that it is unused. For instance, I can define an assembly:

```yaml
assemblies:
  Testing123456:
    specifier: IC
    blocks:
    - name: DriverFuelBlock
      flags: FUEL
      fuel:
        shape: Circle
        material: UZr
        Tinput: 25.0
        Thot: 1500.0
        id: 0.0
        od: 0.5
        mult: 217
    height:
    - 666.0
    axial mesh points:
    - 666
    xs types:
    - B
    material modifications:
      this_is_a_fake_setting: # <--------------------------
        - 0.5
   ```

No mention of the modification `this_is_a_fake_setting` is made in the output. It would be useful for the user to know that their modifications are not recognized, at the very least as an input verification mechanism.

john-science:
When the user puts a material modification in a YAML file (or in code), they expect the mods to be used. In the past, a fictitious (or unused) mat mod would be passed silently. It seems greatly preferable that users be informed that their run probably wasn't what they expected.

This PR will close #402

### DIFF
--- a/armi/materials/b4c.py
+++ b/armi/materials/b4c.py
@@ -29,10 +29,13 @@ class B4C(material.Material):
     def applyInputParams(
         self, B10_wt_frac=None, theoretical_density=None, TD_frac=None, *args, **kwargs
     ):
+        self._warnIfUnusedModifications(kwargs)
+
         if B10_wt_frac is not None:
             # we can't just use the generic enrichment adjustment here because the
             # carbon has to change with enrich.
             self.adjustMassEnrichment(B10_wt_frac)
+
         if theoretical_density is not None:
             runLog.warning(
                 "The 'threoretical_density' material modification for B4C will be "
@@ -46,6 +49,7 @@ class B4C(material.Material):
                 )
             else:
                 self.updateTD(theoretical_density)
+
         if TD_frac is not None:
             self.updateTD(TD_frac)
 
--- a/armi/materials/b4c.py
+++ b/armi/materials/b4c.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 """Boron carbide; a very typical reactor control material."""
-from armi.materials import material
+from armi.materials.material import Material
 from armi.nucDirectory import nuclideBases
 from armi import runLog
 from armi.utils.units import getTc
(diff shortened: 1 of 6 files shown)


ntouran (approved):
sorry for the delay..

john-science:
@youngmit  I don't know. I really like the idea of adding the "did we use this material modification" check in the assembly blueprints:

https://github.com/terrapower/armi/blob/master/armi/reactor/blueprints/assemblyBlueprint.py#L97-L112

But I'm not sure if I see how that's possible. Each material would need to actually register a list of 'mat mods' they use for that to work. And they currently don't.  Maybe they should?

Or am I missing something?  Right now we just pass the material modifications to the `applyInputParams()` method on a material, and it will use it or not at it's leisure.  Is there currently some way to tell what mat mods are used that I'm not seeing?

Cheers!

ntouran:
You're not missing anything. The materials don't communicate with one another and just apply mods as they fly by if they see one they know about. The easiest thing I think we could do (but still not trivial) would be to somehow communicate up when a modification is used. You could imagine modifications being objects that have a `.used` boolean that gets flipped from False to True, and then a thing at the end checking for any `False`s. 

Other than that I don't think we should do too many big structural changes just for this.

youngmit:
> @youngmit I don't know. I really like the idea of adding the "did we use this material modification" check in the assembly blueprints:
> 
> https://github.com/terrapower/armi/blob/master/armi/reactor/blueprints/assemblyBlueprint.py#L97-L112
> 
> But I'm not sure if I see how that's possible. Each material would need to actually register a list of 'mat mods' they use for that to work. And they currently don't. Maybe they should?
> 
> Or am I missing something? Right now we just pass the material modifications to the `applyInputParams()` method on a material, and it will use it or not at it's leisure. Is there currently some way to tell what mat mods are used that I'm not seeing?
> 
> Cheers!

You could do this with a custom dict-like container to pass as kwargs to the materials. It could implement a getitem that logs whenever a key is accessed. When all possible materials have beet hit, the blueprint can then check to see if all of the entries have been accessed and raise a warning of not.

youngmit:
BTW @davis8dd has another PR open with #463 now. Someone should probably claim #402 so that we don't step on each others' toes :D

john-science:
I've decided to take a different approach for my solution, I will close this and attempt a different solution soon.

PR #462, opened 2021-11-05
Remove file name checks when reading in RIPL data.

jakehader:
Remove check for number of files and file names when reading the RIPL data into the nuclide bases. This check was found to be incorrect for the latest version of RIPL (from 7/12/2020) where the file names were no longer labeled as z000.dat to z117.dat. Instead they were changed to z001.dat to z118.dat.


john-science:
On balance, it seems like a clear, easy-to-understand change.  The obvious questions I have are:

1. Now that we have removed this check, what if one or more files are missing?  Will the user know why their run failed?
2. Why not add in a simpler check to make sure that there are >= 118 files named *.dat?  (That does mostly the same thing, but is a bit more future proof.)

jakehader:
Hi @john-science -

These are good questions and I will add the responses here on the PR so we can track them:

1. The original implementation was to make sure that this was checked, but the way that this functions now is that the `nuclideBases` will be populated based on the `zXXX.dat` files. There are missing `zXXX.dat` files then some of the `nuclideBases` may not be initialized and therefore the if some client code needs that nuclide then a `KeyError` exception will be thrown.
2. We _could_ check the number of files, but that may not be future proof. I could add at least 118 files possibly, if that helps?

john-science (approved):
A single, clear idea in a PR. And if an error is raised, it is clean and understandable to the user. I like it!

PR #463, opened 2021-11-08
Warn user on invalid material modification parameters in blueprints.

davis8dd:


armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -225,10 +225,10 @@ def _constructMaterial(self, blueprint, matMods):
             try:
                 # update material with updated input params from blueprints file.
                 mat.applyInputParams(**matMods)
-            except TypeError:
+            except TypeError as te:
                 # This component does not accept material modification inputs of the names passed in
                 # Keep going since the modification could work for another component
-                pass
+                runLog.warning("Error constructing {}: {}".format(mat, te))
```
youngmit, line 231: So this is a good start, but there are a couple of peculiarities with how we handle material modifications that make this function a bit differently than you might expect. For a given Assembly blueprint, I think all of the material modifications for that assembly get passed to all of the materials that make up that assembly. So if you have a material modification for, say Uranium enrichment, it would get passed to the other materials that dont know anything about U235 weight fraction, and they would all produce warnings. This might become more clear if you add some other components to your block design in the test.

We were talking about this some the other day, and I think ive got an interesting approach for this. I'll comment it in #402

armi/reactor/blueprints/tests/test_componentBlueprint.py
```diff
@@ -338,6 +341,48 @@ def test_componentInitializationThoriumNoBurnCustomIsotopics(self):
...
             self.assertIn(nuc, a[0][0].getNuclides())
 
+    def test_componentBlueprintConstructorWithInvalidMatMods(self):
+        materialBlueprint = r"""
+        nuclide flags:
+            U: {burn: false, xs: true}
+            ZR: {burn: false, xs: true}
+        blocks:
+            fuel: &block_fuel
+                fuel: &component_fuel_fuel
```
youngmit, line 351: This test will need to have at least 2 components with different materials to fully test the behavior, i think.

youngmit (changes requested):
See comments (to come) in #402. Pretty sure the coverage check isn't your fault, seems like its breaking for everything.

armi/reactor/blueprints/componentBlueprint.py
davis8dd: Thank you, I did actually debug this with more material modifications in the blueprint but hadn't investigated the myriad of warnings enough to appreciate this peculiarity you mentioned... I did notice that the way a material's "applyInputParams" method (such as uZr.py) directly calls `material.FuelMaterial.applyInputParams(self, *args, **kwargs)` instead of calling up the class hierarchy.  I think refactoring this, and doing input validation on these material modification parameters,, might make for a clearer design.  I'm glad that you propose a more robust solution and will continue the discussion on #402.

john-science:
If you were wondering why this PR was taking so long, it's because when you opened this PR, there was already another PR open for the same ticket:  https://github.com/terrapower/armi/pull/458

Also, I have been trying to weigh the pros and cons on this one. I feel like to do this _right_, the base case will have to change.  But of all the automated ways I've tried to make this happen, none are really sound, logically.

I think there are two possible ways forward that would solve this issue:

1. Each subclass of `Material` need to define a set maximum list of material modification it might take. And every time a material is USED (we apply material modifications), we do a check to see if there are any extra mods and fail hard if there are.
2. https://github.com/terrapower/armi/pull/458#issuecomment-961633138  Put the registration list of modifications that have been used in the blueprint class.

I think I might cancel both of our PRs, because I think (2) above is probably the best option.  I can take a whack at it, but would you like it?

davis8dd:
> I think I might cancel both of our PRs, because I think (2) above is probably the best option. I can take a whack at it, but would you like it?

Closing this PR sounds good to me.
I appreciate the offer but I should pass.  It would likely take me weeks to complete, due to my general unfamiliarity with the codebase as well as current demands from my day job.

PR #464, opened 2021-11-09
Add some more docstrings to Layout class

youngmit:


youngmit:
Not sure whats up with the coverage job... just added comments

PR #465, opened 2021-11-09
Reverting changes to snapshot interface

john-science:
I made some aggressive changes to the snapshot interface when I made settings immutable. These changes weren't wrong, but create more work in plugins to agree with than I think they're worth.


john-science:
@jakehader I have determined this will make updating a few of the plugins faster.

ntouran (approved):

PR #466, opened 2021-11-09
Update description of `burnSteps`

keckler:
The description of `burnSteps` was not clear in its definition of 'n'. It also mistakenly stated that the burnup step would be calculated as `cycle length / n+1`. This was (1) unclear because it was missing parentheses and (2) incorrect because it should be instead `cycle length / n`.


keckler:
Not sure why the coverage test is failing, but #464 is seeing the same issue, so this is probably not my fault...

youngmit (approved):

PR #467, opened 2021-11-09
Clarify the various convergence criteria on the eigenvalue

keckler:
There are two keff convergence criteria settings that have ambiguous descriptions. Clarifications have been added to the descriptions.

### DIFF
--- a/armi/physics/neutronics/settings.py
+++ b/armi/physics/neutronics/settings.py
@@ -185,7 +185,7 @@ def defineSettings():
             CONF_EPS_EIG,
             default=1e-07,
             label="Eigenvalue Epsilon",
-            description="convergence criterion for calculating the eigenvalue",
+            description="Convergence criteria for calculating the eigenvalue in the global flux solver",
         ),
         setting.Setting(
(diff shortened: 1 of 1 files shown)


armi/physics/neutronics/settings.py
```diff
@@ -185,7 +185,7 @@ def defineSettings():
             CONF_EPS_EIG,
             default=1e-07,
             label="Eigenvalue Epsilon",
-            description="convergence criterion for calculating the eigenvalue",
+            description="Convergence criteria for calculating the eigenvalue in the global flux solver",
```
youngmit, line 188: Technically it is singular criterion :P

youngmit (approved):

armi/physics/neutronics/settings.py
keckler: I agree with that! But all the other criterion are listed as criteria as well. Plus I looked it up:
https://www.merriam-webster.com/words-at-play/criteria-vs-criterion-singular-plural-grammar

So I just changed the word for consistency.

PR #469, opened 2021-11-10
Replace is with ==

youngmit:


john-science (approved):
Assuming this passes unit tests, it looks good to me.

Issue #471, opened 2021-11-11
Check on uniform mesh generator uses stale data


keckler:
[One of the checks to confirm that the uniform mesh generation has been done correctly](https://github.com/terrapower/armi/blob/master/armi/reactor/converters/uniformMesh.py#L351) uses stale data, and thus improperly fails.

In particular, when a reactor changes power between cycles using the `powerFractions` case setting, the check will improperly fail on the first node of the second cycle because it calculates `sourcePow` using stale block power data from the previous cycle, but `expectedPow` uses data from case settings of the current cycle.

Example case to come in a later comment...

keckler:
Working on a fix.

keckler:
[Here](https://github.com/terrapower/armi/files/7522734/example-case.zip) is an example input suite that will trigger this failure with the current code, but will run with the fix implemented in #472 .

john-science:
> [Here](https://github.com/terrapower/armi/files/7522734/example-case.zip) is an example input suite that will trigger this failure with the current code, but will run with the fix implemented in #472 .

Well, it's not a lot of code for a test. But it _does_ require a lot of input data, which is a bit much.  Shucks.

PR #472, opened 2021-11-11
Uniform mesh check modification

keckler:
This is in relation to #471 .


john-science (approved):

john-science:
I really appreciate the fix!

And the fast turn-around.

What I would like to push for, in the future, is if we are fixing a bug we add a new unit test to prove the bug is fixed.

While that may not always be easy to do, if it is, it will help us in lots of ways. In particular, the unit test will call out a feature that we think is important to work in a certain way, but has been tricky to do right in the past.

@keckler   Do you think it would be doable to unit test this issue?

keckler:
I did give some thought to a unit test for this PR. My conclusion was that a small unit test would be a little bit difficult for this particular piece of code, because it requires a lot of scaffolding just to have the required inputs. Since I am not really versed in how ARMI even gets to this point, I didn't make a test for it at this time. 

I am going to upload an example case to the original issue sometime today, and we can probably make an integration test from that.

jakehader:
@john-science, this is a tricky unit test because it requires some reactor model to be initialized with detailed axial expansion enabled, with power assigned to the blocks. This was uncovered through our benchmark work, so an in-use test might be more appropriate.

What are your thoughts? A test could probably we whipped up but it might be a lot of initialization code.

keckler:
Check [this comment on the original issue](https://github.com/terrapower/armi/issues/471#issuecomment-966557299) for an example case that causes the code to fail before the current PR is merged in.

Note that I am using an internal TerraPower application to run the neutronics calcs with ARMI. It may not fail when running just the framework, but the issue stems from the framework.

john-science:
@keckler That's a bit of a shame. It would be nice if there was an easy way to test this. But I can't think of a similar test to copy from off the top of my head.

Well, going forward we will try to find ways to test things we fix.  It's a good goal to have. Thanks!

Issue #228, opened 2020-12-18
Make standard way to represent TRISO fuel


ntouran:
[TRISO fuel ](https://en.wikipedia.org/wiki/Nuclear_fuel#TRISO_fuel) is a common need and it'd be nice if there were an obvious straightforward way to define it and use it in reactor models (e.g. #224). Let's try to draft requirements and then talk about implementing. 

####  Faithfully represent standard five-layer [TRISO fuel ](https://en.wikipedia.org/wiki/Nuclear_fuel#TRISO_fuel)(fuel, buffer, inner pyrolytic carbon, SiC, outer pyrolytic carbon

* User shall be able to define the dimensions and materials of each layer
* ? Should we allow non-standard layerings as well? Maybe as a follow-up.

### Follow-on development needs

* Add basic material properties to the material library for the layers
* Make it possible to define more arbitrarily nested collections of components in the input. 
* Add TRISO fuel example

ntouran:
@drewejohnson can you help us get more specific about the requirements of this? What specific parameters do you need to be able to specify about the fuel blocks? What information will you need as you do the modeling? How detailed will you want temperature distributions (or other state) within the fuel, etc?

drewejohnson:
Thanks for the fast response! Some thoughts off the top of my head

Specifying the particle loading as a packing fraction

Some option to indicate to the neutronic solver that the particles should be explicitly modeled or homogenized. Not saying ARMI should handle explicitly modeling the particles, but would be a very useful bit of data for post-processing and modeling comparisons

For state data, I'm not sure how much would be useful without being a overwhelming amount of data. I've pinged some other team members for their thoughts too.

A generalized particle structure would be useful in the same way LWR fuel pins aren't always a fuel / gap / clad pin. I wonder if there could be cross-over in the structures to model particle fuel and pin-type fuel with some number of materials in concentric rings (https://terrapower.github.io/armi/tutorials/walkthrough_lwr_inputs.html#the-uo2-block)

ntouran:
Ok great, thanks. One starting point as @onufer was reminding me is to use volumetric components to make a unit cell that defines the spherical shell volumes and a cubic graphite unit cell with a hole in it. Mocking it up (you'd need to bring in the materials and add the `CubeWithSphericalHole` component in a plugin, and add temperatures and whatnot to make this a fully valid input), I am thinking along the lines of:

```yaml
blocks:
   triso unit cell: 
       fuel:
         shape: Sphere
         material: UraniumOxide
[...]

Issue #224, opened 2020-12-09
Create ARMI inputs for Benchmark


ntouran:
In #223 we discussed how to model something similar to the MHTGR-350 described in [this benchmark document](https://art.inl.gov/NGNP/Water%20Ingress%20Assessment%20Review%20Material/INL%20Published%20Material/Preliminary%20Prismatic%20Coupled%20Neutronics%20Thermal%20Fluids%20Transient%20Benchmark%20of%20the%20MHTGR-350%20MW%20Core%20Design%2012-06-10.pdf). It would be nice to start building inputs representing this core as a good way to:

* Find and fix Framework issues related to modeling Prismatic TRISO-fueled HTGRs in general
* Add to the collection of useful ARMI benchmark inputs

drewejohnson:
Related to the TRISO fuel, is the current / planned support for a fixed structure reflecting the standard five-layer TRISO (fuel, buffer, inner pyrolytic carbon, SiC, outer pyrolytic carbon- https://en.wikipedia.org/wiki/Nuclear_fuel#TRISO_fuel) or a generalized structure?

onufer:
I am not certain what the plan is, but thought I would chime in and mention there is a Sphere  component class defined here:
https://github.com/terrapower/armi/blob/master/armi/reactor/components/volumetricShapes.py

It has ID/OD/multiplicity, so it might meet some of you needs. It could be a good starting place to begin discussion

ntouran:
Well it's a known need at least but we don't have it quite scheduled. Let me make another issue to track that explicitly because it's so fundamental. I'll tag you and try to get you to help define requirements.  (see #228)

PR #477, opened 2021-11-17
Draft: Particle fuel modeling

drewj-usnctech:
Provides a yaml-interface for
- modeling particle fuel with arbitrary layers
- multiple particle fuel types in the reactor
- assigning particle fuel as children to some parent component

We have been decently successful with these changes internally in that
downstream plugins can see `Component.particleFuel` and perform actions
based on their content. What follows is an overview of the interface,
implementation, and a discussion of where to go next.

Related to https://github.com/terrapower/armi/issues/228 and
would support modeling the MHTGR-350 benchmark
https://github.com/terrapower/armi/issues/224

This patch is submitted to kick-start a discussion on better ways to add
this feature into ARMI, leveraging the domain knowledge of the ARMI
developers and the "particle fuel aware" plugins internally developed at
USNC Tech.

Input interface
---------------

```yaml
particle fuel:
    demo:
        kernel:
            material: UO2
            id: 0
            od: 0.6
            Tinput: 900
            Thot: 900
            flags: DEPLETABLE
        buffer:
            material: SiC
            id: 0.6
            od: 0.61
            Tinput: 900
            Thot: 900
```

```yaml
matrix:
    shape: Circle
    material: Graphite
    Tinput: 1200
    Thot: 1200
    id: 0.0
    od: 2.2
    latticeIDs: [F]
    flags: DEPLETABLE
    particleFuelSpec: demo
    particleFuelPackingFraction: 0.4

```

With this interface it's possible to define several specifications
in the model and assign them to different cylindrical components.

Implementation
--------------

The particle fuel is stored as a child of the parent component, such
that `<Circle: Matrix>.children` is used to dynamically find the particle
fuel spec. We can't store the specification as an attribute because we
have to support potentially dynamic addition and removal of children to this
component. Something like `self.particleFuel = spec` that makes spec a
child would also have to understand what happens if we remove the spec
from the parent, e.g., `self.remove(self.particleFuel)`. What is then the
outcome of `self.particleFuel` unless we always check the children of the
matrix?

By making the particle fuel spec a `Composite` and placing it in the
`.children` of the parent, the spec is able to be written to and read
from the database.

Adds `Component.setParticleMultiplicity`. The method is called during
block construction when the block's height is available to the matrix
component (particle's parent). The multiplicity is determined from the
matrix volume and target packing fraction.

Note: the particle mult is, by design, for a single component.

Unresolved issues
-----------------

- Volume of the parent matrix is not reduced by the volume occupied
by the particles.
- No support for homogenizing regions that contain particle fuel
- Various homogenization properties don't account for particle fuel
  (e.g., `Core.getHM*`)
- Particle fuel is not included in some text-reporting, leading to
statements like

```
[info] Nuclide categorization for cross section temperature assignments:
       ------------------  ------------------------------------------------------
       Nuclide Category    Nuclides
       ------------------  ------------------------------------------------------
       Fuel
```
and
```
[warn] The system has no heavy metal and therefore is not a nuclear reactor. Please make sure that this is intended and not a input error.
```

- No provided routines for packing the particles into their parent. This
  could be facilitated with a dedicated packing plugin and an unstructured
  3-D `SpatialGrid` class. It's burdensome to expect the user to define
  the exact location of _every_ particle every time. But, if some `Plugin`
  performs the packing and creates this spatial grid, the multiplicity is
  tackled, and you potentially avoid adding or removing particles as their
  parent expands or contracts.
- Unsure if material modifications make their way down to the materials
  in the particle fuel spec. The implementation suggests it as the `matMods`
  argument is passed into the particle fuel YAML object constructor. But
  we have yet to stress test that

Next steps
----------

This patch is submitted because we continue to find places where this
approach does not play well with the rest of the ARMI stack. While Blocks
that contain particle fuel are correctly able to compute their heavy
metal mass by iterating over their children, which in turn finds the
particle fuel. However, higher-level actions like `Core.getHM*` do not
go down to the sub-block level, instead asking to homogenize Blocks.
The homogenization methods are not yet aware of the particle fuel because
they rely on `Component.getNuclides` which reports the nuclides for it's
`Material`, and does not include the children.

This approach is sensible because if I'm writing a neutronic input file
and I can exactly model the matrix and it's particle fuel, I would expect
`matrix.getNuclides` to return the nuclides for _just the matrix_. Then,
being informed of the particle fuel, I can write those materials and geometry
uniquely. However, codes that cannot handle particle fuel and/or work
with homogenized regions (e.g., nodal diffusion) would need this
homogenized data. Allowing `Component.getNuclides` to return the nuclides
on the child particle fuel would support this case, but not the previous case.

My speculation is that the optimal strategy lies somewhere in making the
matrix object not a `Component` but a `Block` and having the ARMI Composite
tree accept blocks that potentially contain blocks. I think this is valid
using the API but the user interface would need some work. Having the matrix
be a block would provide a better interface for homogenization and exact
representation of the particle fuel. I think...

Other related changes
---------------------

Added `Sphere.getBoundingCircleOuterDiameter` so that the particle fuel
composites can be properly added to the database. They need to be
"sortable" other wise `Database3._createLayout` breaks trying to sort
components. This enables the particle fuel spec to be added to and read
from the HDF data file

The material constructor is a more public function as it is needed both
in creating `Components` but also in creating the particle fuel spec.

Added `MATRIX` flag

Signed-off-by: Andrew Johnson <a.johnson@usnc-tech.com>

### DIFF
--- a/armi/bookkeeping/report/reportingUtils.py
+++ b/armi/bookkeeping/report/reportingUtils.py
@@ -25,6 +25,7 @@ import sys
 import time
 import tabulate
 from copy import copy
+from math import nan
 
 import numpy
 
@@ -56,6 +57,8 @@ Operator_ArmiCodebase = "ARMI Location:"
 Operator_MasterMachine = "Master Machine:"
 Operator_Date = "Date and Time:"
 Operator_CaseDescription = "Case Description:"
+# Convert a value in centimeters to micrometers
+CM_TO_MICRO_METER = 10000
 
 
 def writeWelcomeHeaders(o, cs):
@@ -1002,3 +1005,85 @@ def makeCoreAndAssemblyMaps(r, cs, generateFullCoreMap=False, showBlockAxMesh=Tr
 
 
 COMPONENT_INFO = "Component Information"
+
+
+def makeParticleFuelDesignReport(r):
+    """Add reports for particle fuel specification and usage
+
+    Parameters
+    ----------
+    r : Reactor
+        Reactor containing particle fuel designs. If none are found, no
+        work is done.
+
+    """
+    if not r.blueprints.particleFuelDesigns:
+        return
+
+    descriptions = {}
+
+    for name, design in r.blueprints.particleFuelDesigns.items():
+        rows = _particleFuelSpecToTable(design)
+        descriptions[name] = {
+            "spec": rows,
+            "usage": [],
+        }
+
+    # Find all usages of each specification
+    for blockDesign in r.blueprints.blockDesigns.values():
+        for componentDesign in blockDesign.values():
+            thisParticleFuel = componentDesign.particleFuelSpec
+            if thisParticleFuel is None:
+                continue
+            descriptions[thisParticleFuel]["usage"].append(
+                (
+                    f"{blockDesign.name} {componentDesign.name}",
+                    f"{componentDesign.particleFuelPackingFraction}",
+                )
+            )
+
+    for name, subdata in descriptions.items():
+        grp = report.data.Table(f"{name} Specification")
+        report.setData(
+            "Layer",
+            "Material, Outer diameter (μm), Thickness (μm)",
+            group=grp,
+            reports=report.DESIGN,
+        )
+        for layerName, layerData in subdata["spec"]:
+            report.setData(layerName, layerData, grp, report.DESIGN)
+
+        grp = report.data.Table(f"{name} Usage")
+        if subdata["usage"]:
+            report.setData("Block and Component name", "Packing fraction")
+            for desc, pf in subdata["usage"]:
+                report.setData(desc, pf, grp, report.DESIGN)
+        else:
+            msg = f"Particle fuel specification {name} not used"
+            report.setData("WARNING", msg, grp, report.DESIGN)
+            runLog.warning(msg)
+
+    return
+
+
+def _particleFuelSpecToTable(spec) -> list:
+    rows = []
+    names = []
+    prevOD = None
+    for layer in sorted(spec.values(), key=lambda ring: ring.od):
+        od = layer.od
+        if prevOD is not None:
+            thickness = round(CM_TO_MICRO_METER * 0.5 * (od - prevOD))
+        else:
+            thickness = "-"
+        names.append(layer.name)
+        rows.append(
+            "{}, {}, {}".format(
+                layer.material,
+                round(od * CM_TO_MICRO_METER, 6),
+                thickness,
+            )
+        )
+        prevOD = od
+
+    return list(zip(names, rows))
--- a/armi/reactor/blueprints/__init__.py
+++ b/armi/reactor/blueprints/__init__.py
@@ -104,6 +104,7 @@ from armi.reactor.blueprints.assemblyBlueprint import AssemblyKeyedList
 from armi.reactor.blueprints.blockBlueprint import BlockKeyedList
 from armi.reactor.blueprints import isotopicOptions
 from armi.reactor.blueprints.gridBlueprint import Grids, Triplet
+from armi.reactor.blueprints.componentBlueprint import ParticleFuelKeyedList
 
 context.BLUEPRINTS_IMPORTED = True
 context.BLUEPRINTS_IMPORT_CONTEXT = "".join(traceback.format_stack())
@@ -185,6 +186,9 @@ class Blueprints(yamlize.Object, metaclass=_BlueprintsPluginCollector):
     )
     systemDesigns = yamlize.Attribute(key="systems", type=Systems, default=None)
     gridDesigns = yamlize.Attribute(key="grids", type=Grids, default=None)
+    particleFuelDesigns = yamlize.Attribute(
+        key="particle fuel", type=ParticleFuelKeyedList, default=None
+    )
 
     # These are used to set up new attributes that come from plugins. Defining its
     # initial state here to make pylint happy
--- a/armi/reactor/blueprints/blockBlueprint.py
+++ b/armi/reactor/blueprints/blockBlueprint.py
@@ -168,6 +168,14 @@ class BlockBlueprint(yamlize.KeyedList):
                 b.autoCreateSpatialGrids()
             except (ValueError, NotImplementedError) as e:
                 runLog.warning(str(e), single=True)
+        # check if particle fuel exists and set particle mult
+        # Note: these changes occur here during block construction instead of during
+        # component contruction because component parent parameters (i.e., height)
+        # are needed for volume calculations
+        for component in b.getChildren():
+            if component.particleFuel:
+                component.setParticleMultiplicity()
+
         return b
 
     def _getGridDesign(self, blueprint):
--- a/armi/reactor/blueprints/componentBlueprint.py
+++ b/armi/reactor/blueprints/componentBlueprint.py
@@ -17,6 +17,8 @@ This module defines the ARMI input for a component definition, and code for cons
 
 Special logic is required for handling component links.
 """
+from math import isclose, inf
+
 import six
 import yamlize
 
@@ -24,8 +26,22 @@ from armi import runLog
 from armi import materials
 from armi.reactor import components
 from armi.reactor.flags import Flags
+from armi.reactor.particleFuel import ParticleFuel
 from armi.utils import densityTools
-from armi.nucDirectory import nuclideBases
+from armi.nucDirectory import nuclideBases, nucDir
+
+
+class ComponentParticleFuel(yamlize.Object):
+    specifier = yamlize.Attribute(type=str)
+    packingFraction = yamlize.Attribute(type=float)
+
+    @packingFraction.validator
+    def packingFraction(self, packingFraction):
+        if not 0 < packingFraction < 1:
+            raise ValueError(
+                "Packing fraction must be between 0 and 1, exclusive. Got "
+                "{}".format(packingFraction)
+            )
 
 
 class ComponentDimension(yamlize.Object):
@@ -144,6 +160,8 @@ class ComponentBlueprint(yamlize.Object):
     orientation = yamlize.Attribute(type=str, default=None)
     mergeWith = yamlize.Attribute(type=str, default=None)
     area = yamlize.Attribute(type=float, default=None)
+    particleFuelSpec = yamlize.Attribute(type=str, default=None)
+    particleFuelPackingFraction = yamlize.Attribute(type=float, default=None)
 
     def construct(self, blueprint, matMods):
         """Construct a component"""
@@ -167,6 +185,7 @@ class ComponentBlueprint(yamlize.Object):
         if component.hasFlags(Flags.DEPLETABLE):
             # depletable components, whether auto-derived or explicitly flagged need expanded nucs
             _insertDepletableNuclideKeys(component, blueprint)
+
         return component
 
     def _conformKwargs(self, blueprint, matMods):
@@ -189,6 +208,16 @@ class ComponentBlueprint(yamlize.Object):
                 # Don't pass these to the component constructor. These are used to
                 # override the flags derived from the type, if present.
                 continue
+            elif attr.name == "particleFuelSpec":
+                design = blueprint.particleFuelDesigns[val]
+                value = design.construct(blueprint, matMods)
+            elif attr.name == "particleFuelPackingFraction":
+                value = attr.get_value(self)
+                if value <= 0 or value >= 1:
+                    raise ValueError(
+                        f"Packing fraction {value} not allowed: must be between "
+                        "0 and 1 exclusive",
+                    )
             else:
                 value = attr.get_value(self)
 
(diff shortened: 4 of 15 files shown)


john-science:
@drewj-usnctech  Thanks so much!

Sorry, we had some staff turn-over recently, so between that and Thanksgiving I haven't been at my usual response rate.  This is really interesting though.

Thanks for all the work you put into this. Nick and I were looking at it over lunch.  Hopefully we'll be able to think about this "next steps" and "unresolved issues" a bit more.

armi/reactor/blueprints/__init__.py
```diff
@@ -185,6 +186,9 @@ class Blueprints(yamlize.Object, metaclass=_BlueprintsPluginCollector):
     )
     systemDesigns = yamlize.Attribute(key="systems", type=Systems, default=None)
     gridDesigns = yamlize.Attribute(key="grids", type=Grids, default=None)
+    particleFuelDesigns = yamlize.Attribute(
+        key="particle fuel", type=ParticleFuelKeyedList, default=None
+    )
```
ntouran, line 191: I'm hoping we can come up with a slightly more generic name for this that would still meet the need that you have. 

We have been having some related internal discussions that I have now put in #503 and #504 that I think are very related to this. 

In this case, one could imagine having a section at this level called something like `Mixtures`.

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -168,6 +168,14 @@ def construct(
                 b.autoCreateSpatialGrids()
             except (ValueError, NotImplementedError) as e:
                 runLog.warning(str(e), single=True)
+        # check if particle fuel exists and set particle mult
+        # Note: these changes occur here during block construction instead of during
+        # component contruction because component parent parameters (i.e., height)
```
ntouran, line 173: excellent explanatory comment, thanks!

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -144,6 +160,8 @@ def shape(self, shape):  # pylint: disable=no-self-use; reason=yamlize requireme
     orientation = yamlize.Attribute(type=str, default=None)
     mergeWith = yamlize.Attribute(type=str, default=None)
     area = yamlize.Attribute(type=float, default=None)
+    particleFuelSpec = yamlize.Attribute(type=str, default=None)
+    particleFuelPackingFraction = yamlize.Attribute(type=float, default=None)
```
ntouran, line 164: I'm trying to think of a way we could make these attributes on a subclass that is specialized for particle fuel. Still thinking about it though.

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -189,6 +208,16 @@ def _conformKwargs(self, blueprint, matMods):
...
+            elif attr.name == "particleFuelSpec":
+                design = blueprint.particleFuelDesigns[val]
+                value = design.construct(blueprint, matMods)
+            elif attr.name == "particleFuelPackingFraction":
+                value = attr.get_value(self)
+                if value <= 0 or value >= 1:
+                    raise ValueError(
+                        f"Packing fraction {value} not allowed: must be between "
+                        "0 and 1 exclusive",
+                    )
```
ntouran, line 220: probably redundant with the validator code above, yeah?

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -202,48 +231,77 @@ def _conformKwargs(self, blueprint, matMods):
...
-                mat.applyInputParams(**matMods)
-            except TypeError:
-                # This component does not accept material modification inputs of the names passed in
-                # Keep going since the modification could work for another component
-                pass
-
-        expandElementals(mat, blueprint)
-
-        missing = set(mat.p.massFrac.keys()).difference(nucsInProblem)
+        return constructMaterial(self.material, blueprint, matMods, self.isotopics)
```
ntouran, line 234: I'm sure you did this for a reason (you needed to call constructMaterial from elsewhere, but I'm wondering if it'd be best to just delete this method and call the new `constructMaterial` function directly from whoever calls this private method?

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -310,3 +368,179 @@ def _insertDepletableNuclideKeys(c, blueprint):
...
+    """Component-like specification for a single layer in particle fuel"""
+
+    name = yamlize.Attribute(key="name", type=str)
+    material = yamlize.Attribute(type=str)
+    innerDiam = yamlize.Attribute(key="id", type=float, default=0)
+    od = yamlize.Attribute(key="od", type=float)
+    Tinput = yamlize.Attribute(type=float)
+    Thot = yamlize.Attribute(type=float)
+    # Need this to pick up flags like depletable
+    flags = yamlize.Attribute(type=str, default=None)
```
ntouran, line 383: this is so tantalizingly close to a [sphere](https://github.com/terrapower/armi/blob/9eaa6dc95a1f3d564140b6d632dad482aed0e35b/armi/reactor/components/volumetricShapes.py#L23) that I really wonder if we could use it, probably as a subclass (`ParticleFuelSphere`) that has the extra code you have below. 

Advantages might include needing less code to set it up in general (since shared w/ component and it's blueprints), and the fact that you could either use blueprints or not more directly (e.g. for programmatically building models with fewer blueprints)

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -310,3 +368,179 @@ def _insertDepletableNuclideKeys(c, blueprint):
...
+    """Specification for a single particle fuel type"""
+
+    item_type = _ParticleFuelLayer
+    key_attr = _ParticleFuelLayer.name
+    name = yamlize.Attribute(type=str)
+
+    def construct(self, blueprint, matMods):
+        """Produce a particle fuel instance that can be attached to a component"""
+        bounds = set()
+
```
ntouran, line 443: As mentioned above my dream here is that we can make a slightly more generic thing (`mixture`?) that would give you the capability needed while also potentially being applicable to highly related needs (e.g. the kinds of models mentioned in #503). 

We should discuss further.

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -310,3 +368,179 @@ def _insertDepletableNuclideKeys(c, blueprint):
...
+        particle fuel:
+            <specifier>:
+                <layer>:  # not strictly in increasing order
+                    material: <string>
+                    id: <float>
+                    od: <float>
+                    Tinput: <float>
+                    Thot: <float>
+                    flags: <optional list of strings>
```
ntouran, line 505: fantastic docstring here. very clear what you're doing.

armi/reactor/blueprints/tests/test_blockBlueprints.py
```diff
@@ -318,6 +319,97 @@ def test_explicitFlags(self):
...
 
 
+FULL_BP_PARTICLES = """
+blocks:
+    block: &block
+        flags: fuel test
+        duct:
+            shape: Hexagon
+            material: Graphite
+            Tinput: 600
```
ntouran, line 329: glorious unit tests. THANKS!

armi/reactor/particleFuel.py
```diff
@@ -0,0 +1,90 @@
...
+    ----------
+    name : str
+        Name of this specification, typically something that is easily
+        recognizable by an engineering team, e.g., ``"AGR TRISO"``
+
+    Attributes
+    ----------
+    layers : tuple of :class:`~armi.reactor.components.Sphere`
+        Each layer of the specification in order of increasing outer
+        diameter
```
ntouran, line 24: tantalizingly similar to `children` on a `Composite` (as mentioned in #503)

ntouran (commented):
Thank you _so much_ for this amazing contribution! We're really excited with this level of engagement. Being able to represent TRISO fuel is a dream of ours and here you are making it happen. So awesome. 

There are a few things in here that I commented on that I think warrant a bit of discussion and exploration. We have a few related discussions that I've just posted today publicly that I think tie into this really well and can be combined. 

I'm adding these comments here for now to help facilitate our discussion. we should set up a call again soon to discuss the particular of how to move forward.

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -310,3 +368,179 @@ def _insertDepletableNuclideKeys(c, blueprint):
...
+                    Tinput: 900
+                    Thot: 900
+                    flags: DEPLETABLE
+                buffer:
+                    material: SiC
+                    id: 0.6
+                    od: 0.62
+                    Tinput: 900
+                    Thot: 900
+                ...
```
ntouran, line 527: Just a high-level question here. What if we allowed "free" component definitions (e.g. not in blocks), along the lines of:

```yaml
components:
      kernel:
          material: UO2
          id: 0.6
          od: 0.61
          Tinput: 900
          Thot: 900
          flags: DEPLETABLE
      buffer:
          material: SiC
          id: 0.6
          od: 0.62
          Tinput: 900
          Thot: 900
      matrix:
          material: Graphite
          Tinput: 900
          Thot: 900
```
(ignoring the shape/dimension issue for the moment)

and then upgraded the blocks input to allow mixtures, along the lines of:

```yaml
        <block>:
            fuel:
                name:  fuelmatrix
                shape: Mixture
                mix: 
                  - comp: matrix
                    frac: 0.6
                  - comp: triso
                    frac: 0.4

```

And then have the input processing code take that and build a block that has mixes of those. Would that potentially be workable here? I'm thinking yes as long as you could mix matrix with TRISO, and let TRISO itself also be a composite (e.g. your ParticleFuel object below)

armi/bookkeeping/report/reportingUtils.py
```diff
@@ -1002,3 +1005,85 @@ def makeCoreAndAssemblyMaps(r, cs, generateFullCoreMap=False, showBlockAxMesh=Tr
...
+
+    Parameters
+    ----------
+    r : Reactor
+        Reactor containing particle fuel designs. If none are found, no
+        work is done.
+
+    """
+    if not r.blueprints.particleFuelDesigns:
+        return
```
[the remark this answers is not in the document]
drewj-usnctech: If we add it to the `Composite` class, that would allow clients to check if a specific block or assembly contains particle fuel as well, right? Some possible pathways
- `Component.containsParticleFuel` always returns False as they would be leafs on the composite tree and thus don't contain anything
- Reactor and assembly iterate over children, calling `child.containsParticleFuel` or `any(c.containsParticleFuel() for c in self)`
- Then most of the work happens on the `Block` as a thing that contains components or potentially other blocks?

armi/reactor/blueprints/componentBlueprint.py
drewj-usnctech: Excellent suggestion

armi/reactor/blueprints/componentBlueprint.py
drewj-usnctech: Excellent discussion post! I'll contribute some thoughts / musings

armi/reactor/blueprints/componentBlueprint.py
drewj-usnctech: Having free components (#505) is an awesome move. 

One potential issue with the mixture snippet above would be the `triso` component would need to be something more like a block as it would contain all the child layers. Or including multiple elements in the `mix` section
```yaml
mix:
  - comp: matrix
     frac: 0.6
  - comp: kernel
  - comp: buffer
```
Providing the fraction of kernel and buffer is not difficult, just using 40% times the volume percent of each layer in a sphere. But that's some additional steps required if whipping up an input by hand

drewj-usnctech:
Thank you for the fantastic comments. I'm going to direct most of the higher level comments to https://github.com/terrapower/armi/discussions/503 and maybe sometime next week we could arrange a longer conversation

drewj-usnctech:
(comments originally for the discussion got too focused on this implementation so moving them here)

I think the larger goal of this #503  (blocks containing composites not just components) would help greatly here. You could then model the particle fuel compact as a `Block` containing 
- matrix cylinder to give fixed spatial boundaries (id > 0)
- matrix derived shape to fill in the space inside the bounding cylinder
- individual components making up the spherical layers

Then the parent block(for something like MHTGR-350 // #224) could have children like
- bounding hexagonal duct
- derived shape for the graphite block
- circles for the coolant holes
- sub-block representing the compact

The YAML builders could be updated to expect either a `shape` argument to build a component or another key (sub-block?) to refer to another block that is already defined in the blueprint, e.g., compact.

And then, because the packing fraction is so specific to this application, it might need it's own section just to say "that block that uses this specific sub-block, use this packing / volume fraction." Other plugin changes would likely be needed in order to convert the packing fraction to spatial locations. But if there's some 3-D grid structure that acts like `armi.reactor.grids.Grid`, then you could get multiplicity and packed locations inside the compact from that.

jakehader:
@ntouran - want to make sure that this is still in progress as it's been open for some time now. I see that you have a separate PR open, but are there any other changes here that are needed once that PR lands?

drewj-usnctech:
Closing in favor of #702

PR #478, opened 2021-11-19
Update a variety of doc strings.

ntouran:
These were causing errors in the documentation build.


ntouran:
Oops forgot formatter. And actually, there are only a 'handful' of errors left beyond the repeated ones in #479, so maybe I can get those in here as well.

john-science:
Nice!  It looks like the only thing `black` is complaining about is some trailing white space on a comment line.

So... awesome. I'll fix that and merge this PR this morning. 

For a second there I was worried that `black` and `sphinx` would disagree about formatting, which would lead us to some tedious places.  But I think we're good.

john-science (approved):
Possibly, that trailing white space `black` is complaining about should be fixed.

PR #483, opened 2021-11-20
Adding Instructions on How To Release ARMI

### LINKED ISSUE #479: Eliminate the 429 "duplicate object description" warnings in doc build
If you build the documentation:

```
cd doc
make clean
make html >docbuild.log 2>docbuild-err.log
grep "WARNING: duplicate object description" docbuild-err.log  | wc -l
```
You'll see that this message shows up 429 times. (Once for each object in armi?)

This is likely related to https://github.com/sphinx-doc/sphinx/issues/8664 and https://github.com/sphinx-contrib/napoleon/issues/30 since we do use napolean and apidoc

If we get rid of these, the number of our doc warnings will be *dramatically* reduced from 542 to 113!

john-science:
1. The primary goal of this PR _was_ to add instructions to our docs that give a clear, concise procedure to release a new version of ARMI. While I was at it, I added some some information about what our version numbers mean, and when to rev them.

2. This PR will also close #479.

3. Also, in `conf.py` I turned off documenting all `_private` and `__special__` code in the repo.  This *greatly* reduces the number of warnings and errors we get building the docs, but that's not why I did it. Our "API" is defined as the public methods and attributes in our code. And if those, with their docstrings, are not enough for users, that is a problem we need to fix in our code (by making a method public) or in our documentation (by adding more information). Adding thousands of private methods and attributes to our docs "just to make sure we cover everything" is bad. We should design our API better, and improve our docs, not just throw thousands of lines of _stuff_ into docs and tell people to figure it out. `#endrant`

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -112,14 +112,12 @@ extensions = [
     "sphinx_gallery.gen_gallery",
 ]
 
-# private-member docs are generally not great to link to in high-level implementation
-# documentation because the implementation may change rapidly. Prefer putting info in
-# public entities. We  render docs with private-members shown, however, because there
-# are important implementation details in them in many cases.
+# Our API should make sense without documenting private/special members.
 autodoc_default_options = {
     "members": True,
     "undoc-members": True,
-    "private-members": True,
+    "private-members": False,
+    "special-members": False,
 }
 autodoc_member_order = "bysource"
 autoclass_content = "both"
@@ -135,7 +133,7 @@ napoleon_google_docstring = False
 napoleon_numpy_docstring = True
 napoleon_include_init_with_doc = False
 napoleon_include_private_with_doc = False
-napoleon_include_special_with_doc = True
+napoleon_include_special_with_doc = False
 napoleon_use_admonition_for_examples = False
 napoleon_use_admonition_for_notes = True
 napoleon_use_admonition_for_references = False
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -137,7 +137,7 @@ napoleon_include_special_with_doc = False
 napoleon_use_admonition_for_examples = False
 napoleon_use_admonition_for_notes = True
 napoleon_use_admonition_for_references = False
-napoleon_use_ivar = False
+napoleon_use_ivar = True
 napoleon_use_param = True
 napoleon_use_rtype = True
 
--- a/doc/developer/tooling.rst
+++ b/doc/developer/tooling.rst
@@ -28,7 +28,29 @@ testing.
 
 Releasing a New Version of ARMI
 -------------------------------
-TBD
+In ARMI, we use the common ``major.minor.bump`` version scheme. Where a version string
+might look like ``0.1.7``, ``1.0.0``, or ``1.2.123``, and each number has a meaning:
+
+* ``major`` - Revved for NRC-sanctioned release or at the end of a long design cycle.
+* ``minor`` - Revved when we decide the code or our API has reached a stable point,
+  this might happen once a year.
+* ``bump`` - To be revved every time we modify the API, or at will, any time we want.
+
+**Any change to a major or minor version is considered a release.**
+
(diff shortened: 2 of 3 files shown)


john-science:
@ntouran  Please note, as well as documenting how we release ARMI, this PR also solves the issue you opened here; https://github.com/terrapower/armi/issues/479

The goal is to remove all these: `WARNING: duplicate object description`.

In the `doc/conf.py` file, I changed this line to `True`:

> napoleon_use_ivar = True

I am not 100% sure I understand the issue, but the `napolean_use_param = True` below it allows us to use `Parameters` in Sphinx/RST docstrings, but this `ivar` one allows us to use `Attributes`. For instance, about 100 of our "WARNING: duplicate object description" come from a single class in `yamlize` that has the word "Attribute" in a docstring. This `ivar` change may cause other differences in our docs, I don't know, but it looks all right.

**This one line reduces the number of Sphinx warnings we generate from 501 to 78.**  Which is suddenly much more manageable.

doc/conf.py
```diff
@@ -112,14 +112,12 @@ def setup(app):
     "sphinx_gallery.gen_gallery",
 ]
 
-# private-member docs are generally not great to link to in high-level implementation
-# documentation because the implementation may change rapidly. Prefer putting info in
-# public entities. We  render docs with private-members shown, however, because there
-# are important implementation details in them in many cases.
```
ntouran, line 118: I do love this PR. This one line is the one thing I think might warrant more discussion. In the plugins in particular there are in some cases vast docstrings of math that describe an implementation that I think are really valuable, and this will hide them from the rendered docs. Should we maybe just move all those out of private docstrings to handle this?

john-science: Yeah, this is worth a discussion.

1. This PR doesn't apply to plugins, right? If a plugin has important data hidden inside a private method docstring, then that plugin could decide to document it's private members.
2. You essentially hit the nail on the head. If something is _so_ important, we need to document it, than it shouldn't be in a private method. That method should be public. OR, of course, we could just document that one thing elsewhere, say in the class or module docstring. 

My main goal here is to start thinking about the "API" of our code. Sure, people can always "read all the source code" to understand what a program does. But, typically, more than 50% of a codebase is just implementation details. And what makes life easier for a user is to document the minimal interface they need to understand to interact with the code.

Maybe in our code base, we should add important notes on scientific data in class/module docstrings, because we want to document the SCIENCE.  I love that.  That can help a user way more than documenting 40 private methods on a class that the user doesn't ever need to know about.

doc/developer/tooling.rst
```diff
@@ -2,39 +2,60 @@ Tooling and infrastructure
...
+repos on GitHub.
+
+Every release should follow this process:
+
+1. Ensure all unit tests pass and the documentation is building correctly.
+2. Add release notes to the documentation:
+   `here <https://github.com/terrapower/armi/tree/master/doc/release>`_.
+3. Tag the commit after it goes into the repo: ``git tag -a 1.0.0 -m "Release v1.0.0"``
+4. Also add the release notes on `the GitHub UI <https://github.com/terrapower/armi/releases>`_.
+5. Follow the instructions `here <https://github.com/terrapower/terrapower.github.io>`_ to
```
ntouran, line 53: these two `here` labels will be a sphinx warning :S so you should use anonymous links by making the trailing `_` actually be `__` (!)

doc/conf.py
ntouran: Good points all around. Plugins can indeed still choose to show private docstrings. And yeah maybe class docstrings are generally a better place for that level of implementation detail. And if the plugin author wants private, they can keep them turned on. 

We may want to scan through for particularly useful private docstrings in the framework after landing this just in case.

ntouran (changes requested):
just asking for the `__` in the hyperlinks here.

ntouran (approved):

PR #491, opened 2021-11-29
Realistically, we have not been supporting Python 3.6

john-science:
To the best of my knowledge, everyone who uses ARMI only uses Python 3.7 or 3.9.

I am not sure we really support 3.6, and we certainly don't test for it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
     url="https://github.com/terrapower/armi/",
     license="Apache 2.0",
     long_description=README,
-    python_requres=">=3.7",
+    python_requires=">=3.7",
     packages=find_packages(),
     package_data={"armi": ["resources/*", "resources/**/*"] + EXTRA_FILES},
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -46,7 +46,7 @@ def collectExtraFiles():
     url="https://github.com/terrapower/armi/",
     license="Apache 2.0",
     long_description=README,
-    python_requres=">=3.6",
+    python_requres=">=3.7",
```
jakehader, line 49: Typo in "requires"

jakehader (changes requested):

setup.py
john-science: OMG. I just checked 'git blame' and that typo has been there for at least two years.

Amazing. Yeah, fixing it now.

john-science: Fixed

jakehader (approved):

PR #494, opened 2021-11-30
Fix to saving lattice map using the Grids GUI

jakehader:
Fix to writing the lattice map as a string when saving the core map from the Grids GUI. This fixes a bug introduced in #440.

### DIFF
--- a/armi/reactor/blueprints/gridBlueprint.py
+++ b/armi/reactor/blueprints/gridBlueprint.py
@@ -580,27 +580,32 @@ def saveToStream(stream, bluep, full=False, tryMap=False):
         if gridDesign.readFromLatticeMap or tryMap:
             geomType = geometry.GeomType.fromStr(gridDesign.geom)
             symmetry = geometry.SymmetryType.fromStr(gridDesign.symmetry)
+
+            aMap = asciimaps.asciiMapFromGeomAndDomain(
+                gridDesign.geom, symmetry.domain
+            )()
+            aMap.asciiLabelByIndices = {
+                (key[0], key[1]): val for key, val in gridDesign.gridContents.items()
+            }
             try:
-                aMap = asciimaps.asciiMapFromGeomAndDomain(
-                    gridDesign.geom, symmetry.domain
-                )()
-                aMap.asciiLabelByIndices = {
-                    (key[0], key[1]): val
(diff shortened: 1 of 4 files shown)


armi/reactor/blueprints/gridBlueprint.py
```diff
@@ -580,27 +580,32 @@ def saveToStream(stream, bluep, full=False, tryMap=False):
...
+                # If there is an ascii map available then use it to fill out
+                # the contents of the lattice map section of the grid design.
+                # This also clears out the grid contents so there is not duplicate
+                # data.
                 gridDesign.gridContents = None
+                mapString = StringIO()
+                aMap.writeAscii(mapString)
                 gridDesign.latticeMap = scalarstring.LiteralScalarString(
-                    gridDesign.latticeMap
+                    mapString.getvalue()
```
jakehader, line 608: This code was removed in #440 and is needed to save the lattice map data to the blueprints file.

armi/reactor/blueprints/gridBlueprint.py
```diff
@@ -580,27 +580,32 @@ def saveToStream(stream, bluep, full=False, tryMap=False):
         if gridDesign.readFromLatticeMap or tryMap:
             geomType = geometry.GeomType.fromStr(gridDesign.geom)
             symmetry = geometry.SymmetryType.fromStr(gridDesign.symmetry)
+
```
john-science, line 583: Why was this could move out of the try/except? 

What happens if there is an error in this code now?

armi/reactor/blueprints/gridBlueprint.py
ntouran: oh yeah, only the conditional should have been removed in #440. my bad. Yeah there should have been a test covering that.

armi/reactor/blueprints/gridBlueprint.py
jakehader: I think that there is an actual error that should be expressed and not caught with the same exception. I don't like having multiple actions in one exception because it masks the problem and it was making it harder to debug the issue.

jakehader:
@ntouran is going to work on a unit test for this and I will add to it.

john-science (approved):

Issue #461, opened 2021-11-04
No warning on unused/bogus case settings


keckler:
A user is able to to put a setting into the case settings that is not used anywhere during a run without being warned. There are (at least) two situations where this could be important:
- Similar to #402 , this could allow typos to go unnoticed if the user simply made a mistake in typing the name of the setting, but the setting has a default behavior when it is actually called
- The user typed a valid setting but the setting is not actually used anywhere, indicating that the user has a misunderstanding of what the setting is supposed to do

It would be beneficial if the user was warned in such situations. I imagine this could become quite difficult to actually do, considering plugins and whatnot, unless plugins are required to register their settings with the framework? Regardless, I think it would be very valuable.

jakehader:
Hi @keckler, are you seeing the issue where you put an invalid setting in the YAML input and it just allows this to exist without any warning or are you particularly interested in the plugins/framework reporting if the setting that is input never triggers any lines of code to execute?

If you are interested in the first part, that should be relatively straight-forward to check. I would assume that the case settings should defend against invalid settings, so a failing (or expected-to-fail) example would be really useful. 

The other case seems very challenging to implement generically. If there are specific settings that you are finding trouble with then it might be useful to implement a settings validator check on the framework or the plugin that implements the setting(s). Right now, all settings are registered to the global settings when plugins are registered, so if you get a new settings object after importing and configuring the application, then you should be able to access `cs['pluginSetting']`

keckler:
I am definitely most interested in the first case that you list, where an invalid setting is unnoticed. 

Sorry, I should have put an example case up. An example using a standard case settings file is [attached here](https://github.com/terrapower/armi/files/7506017/tutorial.zip), where I added the setting `THIS_IS_A_FAKE_SETTING: True` to the case settings. I don't see any warning or mention of this invalid setting in the output of the run. I also confirmed that the same invalid setting will pass through a programmatic input unnoticed.

The second type of situation that you describe is almost definitely beyond what a typical user might expect a code to guard against, and so I wouldn't worry about trying to check for it. What I meant in my second bullet point is different than what you describe in a subtle way, I think, but likely just as hard to guard against. I wouldn't worry about that second point as much. Implementing a check for invalid settings in the first place would be a big improvement already.

john-science:
[...]

PR #495, opened 2021-11-30
Validate YAML Files

john-science:
At some point, I believe we turned off validation of YAML files, but kept the validation on XML files. I'm not sure when, how, or why. But this PR applies the XML validation we have been doing the past ~two years to the YAML files.

This will probably raise some warnings for people that they don't expect. It might make people take a closer look at their materials, for instance. But it will also straight up throw warnings, and prompt users to acknowledge if there are bogus settings going into a file (see my unit test for an example).

This PR closes this ticket: https://github.com/terrapower/armi/issues/461

### DIFF
--- a/armi/settings/settingsIO.py
+++ b/armi/settings/settingsIO.py
@@ -275,6 +275,9 @@ class SettingsReader:
         for settingName, settingVal in caseSettings.items():
             self._applySettings(settingName, settingVal)
 
+        if handleInvalids:
+            self._checkInvalidSettings()
+
     def _checkInvalidSettings(self):
         if not self.invalidSettings:
             return
--- a/armi/settings/settingsIO.py
+++ b/armi/settings/settingsIO.py
@@ -205,7 +205,7 @@ class SettingsReader:
         self.format = fmt
         if self.format == self.SettingsInputFormat.YAML:
             try:
-                self._readYaml(stream, handleInvalids=handleInvalids)
+                self._readYaml(stream)
             except ruamel.yaml.scanner.ScannerError:
                 # mediocre way to detect xml vs. yaml at the stream level
                 runLog.info(
@@ -213,10 +213,14 @@ class SettingsReader:
                 )
                 self.format = self.SettingsInputFormat.XML
                 stream.seek(0)
-        if self.format == self.SettingsInputFormat.XML:
-            self._readXml(stream, handleInvalids=handleInvalids)
 
-    def _readXml(self, stream, handleInvalids=True):
+        if self.format == self.SettingsInputFormat.XML:
+            self._readXml(stream)
+
+        if handleInvalids:
+            self._checkInvalidSettings()
+
+    def _readXml(self, stream):
         """
         Read user settings from XML stream.
         """
@@ -246,10 +250,7 @@ class SettingsReader:
         for settingElement in list(settingRoot):
             self._interpretXmlSetting(settingElement)
 
-        if handleInvalids:
-            self._checkInvalidSettings()
-
-    def _readYaml(self, stream, handleInvalids=True):
+    def _readYaml(self, stream):
         """
         Read settings from a YAML stream.
 
@@ -274,9 +275,6 @@ class SettingsReader:
         self.inputVersion = tree["metadata"][Roots.VERSION]
         for settingName, settingVal in caseSettings.items():
             self._applySettings(settingName, settingVal)
-
-        if handleInvalids:
-            self._checkInvalidSettings()
 
     def _checkInvalidSettings(self):
         if not self.invalidSettings:
--- a/armi/tests/test_fuelHandlers.py
+++ b/armi/tests/test_fuelHandlers.py
@@ -506,9 +506,9 @@ class TestFuelHandler(ArmiTestHelper):
         self.assertEqual(widths, [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
 
         # test 2
-        schedule, widths = fh.buildRingSchedule(1, 17, jumpRingFrom=0)
+        schedule, widths = fh.buildRingSchedule(1, 17, jumpRingFrom=5)
         self.assertEqual(
-            schedule, [17, 16, 15, 14, 13, 12, 11, 10, 2, 3, 4, 5, 6, 7, 8, 9, 1]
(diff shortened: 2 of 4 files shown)


armi/tests/test_fuelHandlers.py
```diff
@@ -499,11 +498,20 @@ def test_buReducingAssemblyRotation(self):
...
+        # test 1
         self.assertEqual(
             schedule, [13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 14, 15, 16, 17]
         )
         self.assertEqual(widths, [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
 
+        # test 2
+        schedule, widths = fh.buildRingSchedule(1, 17, jumpRingFrom=0)
+        self.assertEqual(
+            schedule, [17, 16, 15, 14, 13, 12, 11, 10, 2, 3, 4, 5, 6, 7, 8, 9, 1]
```
jakehader, line 511: This pattern confuses me. I'm not sure why the ring scheduler would produce this output from that input in the buildRingSchedule method. I think this might be a bug with the jumpRingFrom equal to 0?

armi/tests/test_lwrInputs.py
```diff
@@ -42,15 +47,35 @@ def tearDownClass(cls):
...
+            b = o.r.core.getFirstBlock(Flags.MOX)
+
+            # test warnings are being logged for malformed isotopics info in the settings file
+            streamVal = mock._outputStream
+            self.assertGreater(streamVal.count("[warn]"), 32, msg=streamVal)
+            self.assertGreater(streamVal.count("custom isotopics"), 32, msg=streamVal)
+            self.assertIn("Uranium Oxide", streamVal, msg=streamVal)
+            self.assertIn("SaturatedWater", streamVal, msg=streamVal)
+            self.assertIn("fakeBad", streamVal, msg=streamVal)
+            self.assertIn("invalid setting", streamVal, msg=streamVal)
```
jakehader, line 70: Is there a way to combine this check with the fakeBad setting check?

armi/tests/tutorials/c5g7-settings.yaml
```diff
@@ -2,6 +2,7 @@ metadata:
   version: uncontrolled
 settings:
   availabilityFactor: 0.9
+  fakeBad: 123456
```
jakehader, line 5: Could you post the output from the stdout when this is added? Just want to confirm that I understand the warning that is made.

jakehader (changes requested):
See comments.

armi/tests/test_fuelHandlers.py
ntouran: might be a bug. weird that it jumps at ring 9. Probably worth digging into. 
Or making a ticket about that and changing jumpRingFrom here to be like `5`.

armi/tests/test_lwrInputs.py
```diff
@@ -42,15 +47,35 @@ def tearDownClass(cls):
...
+            o = armi_init(fName=TEST_INPUT_TITLE + ".yaml")
+            b = o.r.core.getFirstBlock(Flags.MOX)
+
+            # test warnings are being logged for malformed isotopics info in the settings file
+            streamVal = mock._outputStream
+            self.assertGreater(streamVal.count("[warn]"), 32, msg=streamVal)
+            self.assertGreater(streamVal.count("custom isotopics"), 32, msg=streamVal)
+            self.assertIn("Uranium Oxide", streamVal, msg=streamVal)
+            self.assertIn("SaturatedWater", streamVal, msg=streamVal)
+            self.assertIn("fakeBad", streamVal, msg=streamVal)
```
ntouran, line 69: might make sense to check more than the name, including the fact that it's being ignored?

armi/tests/test_fuelHandlers.py
john-science: That's interesting.  I just picked some arbitrary numbers for the test, of course. Well, it looks like that bore fruit really fast. Tests work!

The issue definitely arises there though: 

https://github.com/terrapower/armi/blob/b94744fdf12f247828dd07bf99e251651b494a13/armi/physics/fuelCycle/fuelHandlers.py#L1186-L1190

I assumed this was meant to catch "jumpFromRing = 0", but is this perhaps a mistake and this was meant to test for "jumpFromRing is None"?

armi/tests/test_lwrInputs.py
john-science: Done.

armi/tests/tutorials/c5g7-settings.yaml
john-science: Sure. The total output is, of course, hundreds of lines. But luckily, this part is right at the top:

```
[warn] Ignoring invalid settings: fakeBad
[warn] Burn chain already imposed. Skipping reimposition.
=========== Case Information ===========
[info] ---------------------  ---------------------------------------------------
       Case Title:            c5g7-settings
       Case Description:      C5G7 LWR Benchmark inputs
       Run Type:              Standard - Operator
       Current User:          john
       ARMI Location:         /home/john/git/terrapower/armi/armi
       Working Directory:     /home/john/git/terrapower/armi/armi/tests/tutorials
       Python Interpreter:    3.8.10 (default, Sep 28 2021, 16:10:42)
                              [GCC 9.3.0]
       Master Machine:        ?
       Number of Processors:  1
       Date and Time:         Wed Dec  1 06:20:31 2021
...
```

armi/tests/test_fuelHandlers.py
john-science: Ticket opened:  https://github.com/terrapower/armi/issues/498

Writing unit tests works!

armi/settings/settingsIO.py
```diff
@@ -275,6 +275,9 @@ def _readYaml(self, stream, handleInvalids=True):
         for settingName, settingVal in caseSettings.items():
             self._applySettings(settingName, settingVal)
 
+        if handleInvalids:
+            self._checkInvalidSettings()
```
ntouran, line 279: I recommend moving this to `readFromStream` and pulling out the handleInvalids arg to that level. Then you can remove the identical code from `_readFromXml` while you're at it.

john-science: I'll try it!

john-science: Yeah, that seems like a clean refactor. It also allows us to clean up the method declarations for both of those.

Done!

ntouran (approved):

PR #500, opened 2021-12-01
Fixed bug where SettingsModifier was using old settings API

john-science:
The `Settings` class no longer has a `.get()` method, we use `cs['thing']` or `cs.getSetting('thing')` now. This class was not tested, so it passed the recent settings changes without being modified.


ntouran (approved):
good find and test!

PR #501, opened 2021-12-01
Add warning when ring scheduler given questionable input

### LINKED ISSUE #498: Bug Found in buildRingScheduler
While trying to build some code coverage tests, I wrote this test of `FuelHandler.buildRingScheduler()`:

```python
schedule, widths = fh.buildRingSchedule(1, 17, jumpRingFrom=0)
self.assertEqual(schedule, [17, 16, 15, 14, 13, 12, 11, 10, 2, 3, 4, 5, 6, 7, 8, 9, 1])
```
Notice the strange jump from 10 to 2.

At a guess, perhaps the problem is here:

https://github.com/terrapower/armi/blob/b94744fdf12f247828dd07bf99e251651b494a13/armi/physics/fuelCycle/fuelHandlers.py#L1186-L1190

Perhaps this is meant to be: `elif jumpRingTo is None:`.

ntouran:
## Description

This just adds a warning to help users identify somewhat strange-seeming behavior in fuel handler ring scheduling.

Closes #498 

---

## Checklist


armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1174,18 +1175,24 @@ def buildRingSchedule(
         findAssembly
 
         """
+        maxRingInCore = self.r.core.getNumRings()
+        if dischargeRing > maxRingInCore:
+            runLog.warning(f"Discharge ring {dischargeRing} is outside the core (max {maxRingInCore}).")
```
jakehader, line 1180: Can you include some further description in the warning for how the use may fix this? Same for the warning below.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1174,18 +1175,24 @@ def buildRingSchedule(
...
             dischargeRing = 1
         if chargeRing is None:
-            chargeRing = self.r.core.getNumRings()
+            chargeRing = maxRingInCore
 
-        if chargeRing > dischargeRing and not jumpRingTo:
+        if chargeRing > dischargeRing and jumpRingTo is None:
             jumpRingTo = 1
-        elif not jumpRingTo:
+        elif jumpRingTo is None:
```
jakehader, line 1193: All this conditional logic is very hard to follow. Is there a way to clean this up so it's easier to understand and test?

jakehader (changes requested):

armi/physics/fuelCycle/fuelHandlers.py
john-science: I think this line is fine.  But, for my preference, at least some comments on the if-statements, or something. Just to make them easier to understand.

john-science (approved):

john-science:
@ntouran  This is still complaining of `black` formatting.

armi/physics/fuelCycle/fuelHandlers.py
ntouran: I tried. this method can be called by a variety of paths so I can 't be too specific.

armi/physics/fuelCycle/fuelHandlers.py
ntouran: I added some comments

ntouran:
Hey @jakehader  and @john-science  I changed this in a different way now that might end up crashing some existing shuffle logic. But I think it makes more sense now.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1174,18 +1176,44 @@ def buildRingSchedule(
...
+            # divergent case. Disable jumping by putting jumpring
+            # at periphery.
             if self.r:
-                jumpRingTo = self.r.core.getNumRings()
+                jumpRingTo = maxRingInCore
             else:
                 jumpRingTo = 18
```
jakehader, line 1218: Sort of an arbitrary choice, right? Can we do better than this now or at least document why the number 18 is selected?

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -497,20 +497,28 @@ def test_buReducingAssemblyRotation(self):
...
-        self.assertEqual(widths, [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
+        self.assertEqual(schedule, [6, 7, 8, 9, 5, 4, 3, 2, 1])
+        self.assertEqual(widths, [0, 0, 0, 0, 0, 0, 0, 0, 0])
+
+        # crash on invalid jumpring
+        with self.assertRaises(ValueError):
+            schedule, widths = fh.buildRingSchedule(1, 17, jumpRingFrom=0)
+
+        # test 4: Mid way jumping
+        schedule, widths = fh.buildRingSchedule(1, 9, jumpRingTo=6, jumpRingFrom=3)
```
jakehader, line 520: Adding documentation in the fuel handler docs would be great to explain more of what is going on when before jumpRingTo and jumpRingFrom are used. I am able to follow the cases with jumpRingFrom but I don't quite understand why is this case (from right to left) we go 1, 2, 3 and then 6, 5, 4 and then 7, 8, 9. I get a bit confused as to why there is a reordering of 4 and 5 as well.

jakehader (changes requested):

armi/physics/fuelCycle/fuelHandlers.py
john-science: Huh. I never even noticed that 18 before.

At the very least, we should probably add a comment for why that magic number was chosen.

But it scares me a little, yeah.

jakehader: @ntouran, can you look into this an explain why 18 is still a good number or think of maybe a better way to handle this. I think this code is pretty old so maybe the framework changes have made it easier to set a better max value?

jakehader (approved):
This is going to be relooked at shortly so approving now and my other comment can be addressed later.

armi/physics/fuelCycle/fuelHandlers.py
john-science: I think this should still be addressed. I'm not sure where `18` comes from, and that makes me nervous.

Issue #504, opened 2021-12-01
Allow components to be input outside of a block in blueprints


ntouran:
For lots of reasons (e.g. [this discussion](https://github.com/terrapower/armi/discussions/503), we'd like to be able to define components in a components: root level in the blueprints rather than just inside blocks.

:warning: This is a hard task

Complexities and questions:

* This will require updating the is3D logic and having explicit height included on "2D" components.
* We will need new syntax and block construction logic to bring components into blocks. Will we use strings instead of anchors as mentioned in Inputs?
* Will we still support the current style of inputs? I don't see why not. This could start as a way to just add more componentDesigns to a blueprints.
* There is the question of groups of components as well. In a pin lattice, you often want to place clad, fuel, spacers, gap/bond, liner together in a bunch of positions. How will we group these? Just sharing a grid specifier may work.

ntouran:
I'll take a stab at this and get a branch going to be collaborated on with @john-science

john-science:
@ntouran  This ticket is still assigned to you.  That's cool with me, but if you aren't planning on working on it, please un-assign it.

(You are not being singled out, I am just trying to go through all the ARMI tickets.)

PR #505, opened 2021-12-02
Add new top-level components input for blueprints.

ntouran:
## Description

This adds a new `components` section to the top-level
blueprints input file. Components defined there are loaded
into the new `blueprints.componentDesigns` data structure,
where they can be found and used to construct components.

This is a new and optional way to define components, separate
from the traditional way of defining them within parent blocks.

Note that with this commit, there is no way to automatically
load components into blocks, etc. That would have to be done
programmatically, for now.

This is related to a desire to have more flexible ways to define
models, including e.g. TRISO fuel, MSR models, etc.

This is part of #504, but doesn't fully complete it since we need to also
make ways to mix/match components and then load them into
systems. 

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/componentBlueprint.py
+++ b/armi/reactor/blueprints/componentBlueprint.py
@@ -295,7 +295,7 @@ def _insertDepletableNuclideKeys(c, blueprint):
 
 class ComponentKeyedList(yamlize.KeyedList):
     """
-    An OrderedDict of CompnentBlueprints keyed on the name.
+    An OrderedDict of ComponentBlueprints keyed on the name.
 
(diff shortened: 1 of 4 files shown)


armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -293,6 +293,20 @@ def _insertDepletableNuclideKeys(c, blueprint):
     )
 
 
+class ComponentKeyedList(yamlize.KeyedList):
+    """
+    An OrderedDict of CompnentBlueprints keyed on the name.
```
ntouran, line 298: oops spelled wrong

Issue #515, opened 2021-12-23
`_checkParamConsistency` not being used in `AssemblyBlueprint.construct()`


keckler:
There is a method inside of the `AssemblyBlueprint` class called [`_checkParamConsistency`](https://terrapower.github.io/armi/_modules/armi/reactor/blueprints/assemblyBlueprint.html#AssemblyBlueprint._checkParamConsistency), which is supposed to check that an assembly blueprint has the same number of mesh points, heights, xs types, and material modification entries as there are blocks. This is called within [`AssemblyBlueprint.construct`](https://terrapower.github.io/armi/_modules/armi/reactor/blueprints/assemblyBlueprint.html#AssemblyBlueprint.construct).

I think the issue is that this checking method is called _after_ the assembly is constructed, and if there is an inconsistency in the assembly blueprint stemming from different numbers of blocks/xs-types/etc. then `_constructAssembly` will expose that inconsistency by bombing: 
```
[CRIT 000 ] Stack trace: Traceback (most recent call last):
  File "c:\users\ckeckler\codes\armi\armi\__main__.py", line 34, in main
    code = ArmiCLI().run()
  File "c:\users\ckeckler\codes\armi\armi\cli\__init__.py", line 187, in run
    return self.executeCommand(args.command, args.args)
  File "c:\users\ckeckler\codes\armi\armi\cli\__init__.py", line 216, in executeCommand
    return cmd.invoke()
  File "c:\users\ckeckler\codes\armi\armi\cli\run.py", line 33, in invoke
    inputCase.run()
  File "c:\users\ckeckler\codes\armi\armi\cases\case.py", line 370, in run
    o = self.initializeOperator()
  File "c:\users\ckeckler\codes\armi\armi\cases\case.py", line 430, in initializeOperator
    r = reactors.factory(self.cs, self.bp)
  File "c:\users\ckeckler\codes\armi\armi\reactor\reactors.py", line 141, in factory
    coreDesign.construct(cs, bp, r, geom=geom)
  File "c:\users\ckeckler\codes\armi\armi\reactor\blueprints\reactorBlueprint.py", line 142, in construct
    self._loadAssemblies(cs, system, gridDesign, gridDesign.gridContents, bp)
  File "c:\users\ckeckler\codes\armi\armi\reactor\blueprints\reactorBlueprint.py", line 165, in _loadAssemblies
    newAssembly = bp.constructAssem(cs, specifier=aTypeID)
[...]

Issue #513, opened 2021-12-15
material modifications cannot handle different values for a key within a given block


keckler:
As described in the [docs on blueprints](https://terrapower.github.io/armi/user/inputs/blueprints.html?highlight=material%20modifications), the ability to specify material properties in a blueprint is typically through the use of material modifications, which are applied on the block level during assembly construction. 

Essentially what is done is that every component in a given block of the assembly tries to use all of the `matMods` keys, and the `matMods` that apply to the material class in each particular component end up being used while those that are irrelevant to the material class are ignored by the component. So if you had a configuration that looked like this (yes, I know it is weird to have Lithium as a coolant, just doing it to demonstrate):
```
blocks:
    fuel: &block_fuel
        fuel: &component_fuel_fuel
            shape: Circle
            material: UZr
            Tinput: 25.0
            Thot: 600.0
            id: 0.0
            mult: 169.0
            od: 0.86
        coolant: &component_fuel_coolant
            shape: DerivedShape
            material: Lithium
            Tinput: 450.0
            Thot: 450.0

assemblies:
    heights: &standard_heights [25.0]
    axial mesh points: &standard_axial_mesh_points [ 1]
    igniter fuel:
        specifier: IC
        blocks: &igniter_fuel_blocks [ *block_fuel]
        height: *standard_heights
        axial mesh points: *standard_axial_mesh_points
        material modifications:
            U235_wt_frac: [ 0.11]
            ZR_wt_frac: [0.06]
            LI_wt_frac: [0.5]
        xs types: [A]
```
then the UZr component would pick up the `U235_wt_frac` and `ZR_wt_frac` material modifications, while the Lithium coolant component would pick up the `LI_wt_frac` modification.

If you instead replaced the coolant component's material with UZr, then both components in the block would pick up the `U235_wt_frac` and `ZR_wt_frac` material modifications and neither would get the `LI_wt_frac`. And if the assembly were composed of multiple stacked blocks, one could specify the material modifications as a list to apply different values to the different blocks as in `ZR_wt_frac: [0.5, 0.6, 0.7]`.

However, right now there is no ability to specify unique enrichments for different materials that are _in the same block_. As an example, consider the following scenario where one has constructed a block with two different fuel components:
```
blocks:
    fuel: &block_fuel
        fuel1: &component_fuel_fuel1
            shape: Circle
            material: UZr
            Tinput: 25.0
            Thot: 600.0
            id: 0.0
            mult: 100
            od: 0.86
[...]

PR #517, opened 2021-12-27
Componentwise mat mods

keckler:
## Description
This PR addresses #513  by adding the ability to specify material modifications on a component level. Using the `componentwise` attribute in the `material modifictions` section of the blueprints, a given material modification may be specified to apply to only a selection of components in the block.

Backwards compatibility is maintained in terms of the blueprints input file. But an additional attribute `blockwise` has been added for the user to highlight that a given material modification applies to all applicable components in the block. This `blockwise` keyword may be used optionally and material modifications without either the `componentwise` or `blockwise` keyword will be treated blockwise as before.

The user docs have been updated to reflect this change.

---

## Checklist


keckler:
Any comments on the changes to the input parsing would be particularly welcomed. Also the attribute names `blockwise` and `componentwise` could potentially be improved upon.

keckler:
Reviewer needed. @john-science @ntouran @jakehader

john-science:
@keckler Thanks!  I get back from vacation today and I'll take a look!

@ntouran I will want your opinion on this, as you have a better view of other material/geometry work coming up in the future.

keckler:
Of course, no rush over the holidays!

armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -153,17 +192,67 @@ def _constructAssembly(self, cs, blueprint):
...
+        assemblyMaterialInput = {}
+
+        if self.materialModifications != None:
+            assemblyMaterialInput[""] = {}
+            for modName, modList in self.materialModifications.items():
+                assemblyMaterialInput[""][modName] = modList
+
+            if self.materialModifications.blockwise:
+                for modName, modList in self.materialModifications.blockwise.items():
+                    if modName in assemblyMaterialInput[""].keys():
```
ntouran, line 226: no need to say `.keys()` here. The `in` statement checks keys already.

armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -153,17 +192,67 @@ def _constructAssembly(self, cs, blueprint):
...
+        single value to apply to all blocks without needing to write out a whole list.
+
+        To maintain backwards compatibility, we first assumed that there might be
+        blockwise matMods on the base object, then we add in any blockwise matMods
+        that are explicitly denoted using the `blockwise` attribute.
+        """
+        assemblyMaterialInput = {}
+
+        if self.materialModifications != None:
+            assemblyMaterialInput[""] = {}
```
ntouran, line 220: I recommend defining a special module-level constant like `BLOCKWISE = "_BLOCKWISE"` and using that instead of `""` as the blockwise key. That will add some additional useful context.

armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -153,17 +192,67 @@ def _constructAssembly(self, cs, blueprint):
...
+                        )
+                    assemblyMaterialInput[""][modName] = modList
+
+            if self.materialModifications.componentwise:
+                for component, mods in self.materialModifications.componentwise.items():
+                    assemblyMaterialInput[component] = {}
+                    for modName, modList in mods.items():
+                        assemblyMaterialInput[component][modName] = modList
+
+        return assemblyMaterialInput
```
ntouran, line 239: Why not just modify the object's `self.materialModifications` data structure directly here rather than making a modified copy of the data and passing it around in function calls?

armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -174,15 +263,16 @@ def _createBlock(self, cs, blueprint, bDesign, axialIndex):
 
     def _checkParamConsistency(self, a):
         """Check that the number of block params specified is equal to the number of blocks specified."""
-        materialInputs = self.materialModifications or {}
+        materialInputs = self._compileAssemblyMaterialInput()
```
ntouran, line 266: It's surprising that a checkParamConsistency call would do data compilation step. I suggest reverting this line and calling `self._compileAssemblyMaterialInput()` at the higher level, right before `_checkParamConsistency` is called.

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -170,6 +179,46 @@ def construct(
...
+        Get the blockwise material modifications and those specifically for this
+        component.
+
+        If a material modification is specified both blockwise and componentwise
+        for a given component, the componentwise value will be used.
+        """
+        filteredMaterialInput = {}
+
+        # first add the blockwise modifications without question
+        if "" in materialInput.keys():
```
ntouran, line 204: can omit `.keys()`

doc/user/inputs/blueprints.rst
```diff
@@ -404,9 +404,98 @@ material modifications
...
+                xs types: [A]
+                material modifications:
+                    componentwise:
+                        fuel1:
+                            U235_wt_frac: [0.20]
+                        fuel2:
+                            Zr_wt_frac: [0.02]
+                    U235_wt_frac: [0.30]
+
+  This example would apply an enrichment of 20% to the `fuel1` component and an
```
ntouran, line 446: `fuel1` and other literals should have double-backticks around it for proper rst rendering

doc/user/inputs/blueprints.rst
```diff
@@ -404,9 +404,98 @@ material modifications
...
+  enrichment of 30% to all other components in the block that accept the `U235_wt_frac`
+  material modification. The user may alternatively use the `blockwise` attribute
+  to more clearly indicate that the material modification type is being applied to
+  all applicable components in the block::
+
+                material modifications:
+                    componentwise:
```
ntouran, line 453: does `bycomponent`  and `byblock` sound better to you?  Or just `component` and `block`? The `wise` part in these names feels slightly funny to me. Not a big deal though.

ntouran (changes requested):
Looks like a pretty reasonable solution to a legit inadequacy of the input. I have a few minor comments in here, and would like to get @john-science's take on it as well since it's a input change. 

I am a little concerned about how this will work when we have better treatment of blocks containing composites (collections of collections + components), but the overall concept will stand as long as we iterate over the final components in such scenarios. 

Thanks for the unit test and doc updates too!

armi/reactor/blueprints/assemblyBlueprint.py
keckler: I struggled with coming up with a good key name for the blockwise material modifications, because I didn't want to use a name that might possibly conflict with a name that the user might have for a component. 

Like, if the user were to name a component `_BLOCKWISE` for whatever odd reason, then any material modifications to that component would be overwritten internally. I know it is unlikely that a user might use that exact name, but still... Trying to avoid conflicts here.

I figured the safest way would be to not use a real word at all, hence the use of `""`. I figured that yamlize wouldn't allow an empty string to be used as an attribute name, but now that I actually checked, I find out I was wrong. So I'll change it to something more clear.

armi/reactor/blueprints/tests/test_materialModifications.py
```diff
@@ -76,6 +84,93 @@ def test_u235_wt_frac_modification(self):
...
+        assert_allclose(0.20, u235 / u)
+
+    def test_u235_wt_frac_componentwise_modification1(self):
+        a = self.loadUZrAssembly(
+            """
+        material modifications:
+            componentwise:
+                fuel1:
+                    U235_wt_frac: [0.20]
+            U235_wt_frac: [0.30]
```
jakehader, line 117: Can you describe how this would work if both a general wt frac and a component wt frac are provided together? Does one invalidate the other or is there a check that we should have to fail or warn the user?

Also, maybe if we are going this route then does it even make sense to maintain block material modifications? Maybe from a backwards compatibility perspective, but I think that what we really want are just mods on the component level anyways. 

@ntouran / @john-science?

doc/user/inputs/blueprints.rst
keckler: Yeah, I like `bycomponent` better. I'll probably change the key to `by component`, since yamlize is capable of that and it is more clear.

Still don't quite like `by block`, though. To me, it is not clear how the `by block` concept is different from how the material modifications are required to be entered as a list with one value per block. I'm thinking a name more like `block default` would work better. That strongly implies the concept that the material modification will fall back to the `block default` value if a `by component` value does not supersede it.

What do you think?

armi/reactor/blueprints/tests/test_materialModifications.py
keckler: Good question @jakehader . I added a section in the docs that describes how such a situation would work. Check [here](https://github.com/terrapower/armi/pull/517/files#diff-457ae5e548d65f136e8364a5191df93f944ca0b0734e8338f8dc0a6bbd2256fb).

The componentwise values supplant the blockwise values, if both are provided. Any value not provided componentwise will fall back to the blockwise value. And if neither are provided, then it will fall back to whatever the material class does (i.e. either use some default in the material class or fail).

doc/user/inputs/blueprints.rst
```diff
@@ -404,9 +404,98 @@ material modifications
...
+                height: [0.5, 0.5]
+                axial mesh points: [1, 1]
+                xs types: [A, A]
+                material modifications:
+                    componentwise:
+                        fuel1:
+                            U235_wt_frac: [0.20, ''] # <-- the U235_wt_frac for the second block will go to the blockwise value
+                        fuel2: # the U235_wt_frac for fuel2 component in both axial blocks will go to the blockwise value
+                            Zr_wt_frac: [0.02, ''] # <-- the Zr_wt_frac for the second block will default because there is no blockwise value
+                    blockwise:
```
jakehader, line 497: This answers my question above. This is nice that it keeps backwards compatibility. 

My main question here is - Is the length of the component and block mat mod lists required to be the same length as the number of blocks in the assembly? I believe they should be, but if not then we should make sure so there aren't any mistakes on the mapping. 

Also, it would be nice if there was a material modification summary table that could be produced for the user in the `stdout`. This is probably a separate diff, but would definitely help me verify that material mods are being applied as expected.

armi/reactor/blueprints/tests/test_materialModifications.py
keckler: My opinion is that the blockwise values are still useful, just to provide that default behavior if a componentwise value isn't provided.

doc/user/inputs/blueprints.rst
keckler: Yeah, the lengths of the mat mods lists all have to be the same. If they are not, the code will crash (or once #515 is fixed, a good error will be printed).

keckler: A summary table would definitely be useful. But I agree that it would be most appropriate in a separate PR.

Another thing that I think would be useful is a straight default mat mod value (not a list) on a given assembly so that the full list does not have to be written out in the case that all axial blocks take the same value. Obviously not needed, but could be convenient and reduce errors.

jakehader: That's true. It really does beg the question of what the mat mods are really supposed to do though. 

I understand that they interact with the material classes and there are some allowable kwargs for modifications, but it seems that with a full assembly modification we might get into the territory of considering setting number densities of nuclides directly on the blocks after the assembly had been modified. For example, reducing coolant density.

Maybe for now we just stick to what it is doing with applying mods to the material classes and also consider the full assembly mods and other uses for mat mods with @ntouran separately.

doc/user/inputs/blueprints.rst
jakehader: I like the idea of just adding `by component` and then leaving the block material modifications at the top level. This will keep backwards compatibility with inputs.

keckler: Yeah, that might be the best way to go. I always tend to like symmetry, so if there is a `by block` attribute, my gut wants there to be an alternative attribute as well. But just leaving the blockwise values on the `material modifications` level might be best.

keckler: I changed the latest commit to use `by component` and got rid of the `blockwise` option entirely. Now all block default values are entered as they were before on the `material modifications` level.

armi/reactor/blueprints/assemblyBlueprint.py
keckler: Fixed.

armi/reactor/blueprints/assemblyBlueprint.py
keckler: I changed the key.

armi/reactor/blueprints/assemblyBlueprint.py
keckler: I changed the flow to get rid of the `_compileAssemblyMaterialInput` method entirely. So no duplicates are made and nothing is passed around anymore.

armi/reactor/blueprints/assemblyBlueprint.py
keckler: I got rid of the `_compileAssemblyMaterialInput` method and changed this part of the code to just pull directly from `self.materialModifications` as needed.

armi/reactor/blueprints/blockBlueprint.py
keckler: Fixed.

doc/user/inputs/blueprints.rst
keckler: Fixed.

keckler:
I think that all of @ntouran and @jakehader 's comments have been addressed by the latest commit. If you have any other comments on the latest version, I'd be happy to hear them. And then also waiting to hear if @john-science has anything he wants to add.

armi/reactor/blueprints/assemblyBlueprint.py
john-science: Since the changes were made (possibly to address the comments directly), I can't really see or understand this conversation any more.  Keckler's comment makes it sound like a good change was made here.  One of you two should either continue this conversation or "resolve" it.

armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -174,15 +220,19 @@ def _createBlock(self, cs, blueprint, bDesign, axialIndex):
...
             "xs types": self.xsTypes,
         }
-        for modName, modList in materialInputs.items():
-            paramName = "material modifications for {}".format(modName)
-            paramsToCheck[paramName] = modList
+
+        for mod in [self.materialModifications] + [
+            self.materialModifications.byComponent[key]
+            for key in self.materialModifications.byComponent.keys()
+        ]:
```
john-science, line 232: Unless I'm missing something, I believe this might be more clearly written as:

```python
        for mod in [self.materialModifications] + list(self.materialModifications.byComponent.values()):
```

doc/user/inputs/blueprints.rst
john-science: It looks to me like this conversation can be "resolve"d.  @keckler ?

doc/user/inputs/blueprints.rst
john-science: > Yeah, the lengths of the mat mods lists all have to be the same. If they are not, the code will crash (or once #515 is fixed, a good error will be printed).

Do we have a unit test for that?  If not, can we get one?

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -170,6 +179,46 @@ def construct(
...
+            if component != "byBlock":
+                if component not in [componentDesign.name for componentDesign in self]:
+                    if materialInput[component]:  # ensure it is not empty
+                        raise ValueError(
+                            f"The component '{component}' used to specify a by-component"
+                            f" material modification is not in block '{self.name}'."
+                        )
+
+    @staticmethod
+    def _filterMaterialInput(materialInput, componentDesign):
```
john-science, line 193: Two thoughts:

1. If this is a `@staticmethod` then we want people to use this method outside this class, so it shouldn't start with an `_`.
2. This method seems cool, and I like it, but it also does a lot of work. Can we have a unit test or three on this method alone?

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -170,6 +179,46 @@ def construct(
                 runLog.warning(str(e), single=True)
         return b
 
+    def _checkByComponentMaterialInput(self, materialInput):
```
john-science, line 182: Can we have a unit test for this method?

armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -174,15 +220,19 @@ def _createBlock(self, cs, blueprint, bDesign, axialIndex):
 
     def _checkParamConsistency(self, a):
```
john-science, line 221: It is somewhat unfair to you, @keckler , this method didn't have unit tests to begin with. But now I'm wondering if we need one to validate your change works.

armi/reactor/blueprints/assemblyBlueprint.py
keckler: Of course, will fix and put back up.

john-science:
I would like some more unit tests for this PR.

Having said that, I think the feature is sorely needed.  And I love that documentation was added.

Part of the reason I like the idea of unit tests on these new methods here is I like the idea of this PR a lot, and can imagine there are edge cases here that might cause problems that I can't think of.

It's solid work.

doc/user/inputs/blueprints.rst
keckler: In my view it is resolved, just didn't mark it as such in case anyone else felt differently.

armi/reactor/blueprints/assemblyBlueprint.py
keckler: Resolving now.

armi/reactor/blueprints/blockBlueprint.py
keckler: Added.

armi/reactor/blueprints/blockBlueprint.py
keckler: The reason for the `@staticmethod` is just to avoid passing the instance to the method, since the instance is not needed. But I don't expect this method to be used from outside the class, hence the underscore. Per our offline discussion, I am going to leave that unchanged.

Unit test added.

doc/user/inputs/blueprints.rst
keckler: Unit tests are added for this function. I also went ahead and fixed #515 myself in addressing this, since the two were so tied up with each other.

armi/reactor/blueprints/assemblyBlueprint.py
keckler: Unit tests are added. Also went ahead and fixed #515 in the process, since it was so related.

keckler:
Okay @john-science , I think all comments are addressed. Added a bunch of unit tests. Let me know if anything else looks amiss.

armi/reactor/blueprints/tests/test_assemblyBlueprints.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def loadCustomAssembly(self, assemblyInput):
+        yamlString = assemblyInput
+        design = blueprints.Blueprints.load(yamlString)
+        design._prepConstruction(settings.getMasterCs())
+        return design.assemblies["fuel a"]
+
+    def test_checkParamConsistency(self):
+        # make sure a good example doesn't error
+        a = self.loadCustomAssembly(self.twoBlockInput_correct)
```
ntouran, line 188: this should technically have an assert statement afterwards to show not only that it runs but also that it loads values as expected

ntouran (approved):
Looks good to me! Could use one or two more asserts in the positive tests but NBD.

ntouran:
@john-science  I'm good on this. You have any other requests? I'll let you squash/merge it if not.

armi/reactor/blueprints/tests/test_assemblyBlueprints.py
john-science: Ah, good catch!

I agree.  Pull the most important things out of that assembly and make sure they are correct.

john-science:
> @john-science I'm good on this. You have any other requests? I'll let you squash/merge it if not.

I agree. Other than your comment about asserts in that test, this is ready. 

Nice work, @keckler !

keckler:
@john-science I just added the assert statements into the test in the latest push.

keckler:
Not rushing anybody, just wanted to make sure there is nothing more that needs to be done from me on this?
@john-science @ntouran

Issue #518, opened 2021-12-28
ARMI Application Sphinx Build Error with ARMI Logger


jakehader:
The ARMI framework documents can be built using the `make html` command within the `doc` directory. When building the documentation using Sphinx, it imports the ARMI `RunLogger` during the process. Once ARMI's logger is imported and configured then the logging is performed from then on with this logger rather than with Python's standard logger.

A problem arises if any errors are generated from Sphinx, as it expects the logger that is configured to have certain keywords defined in the logger's `log` method. Currently, we can see that we implement `single` and `label` kwargs that we use internally to catalog the message types as well as to perform some duplication filtering. 

https://github.com/terrapower/armi/blob/04526a658dcf59ba1e29c4613374fe23624d33f2/armi/runLog.py#L478-L519

The following table provides the kwargs that are implemented in Python's standard `Logger` (see: https://github.com/python/cpython/blob/576e38f9db61ca09ca6dc57ad13b02a7bf9d370a/Lib/logging/__init__.py#L1591-L1615):

| **Python Version** | `Logger` kwargs |
| -------------------- | ------------------ |
| 3.4 | `exc_info`, `extra`, `stack_info` |
| 3.5 | `exc_info`, `extra`, `stack_info` |
| 3.6 | `exc_info`, `extra`, `stack_info` |
| 3.7 | `exc_info`, `extra`, `stack_info` |
| 3.8 | `exc_info`, `extra`, `stack_info`, `stacklevel` |
| 3.9 | `exc_info`, `extra`, `stack_info`, `stacklevel` |
| 3.10 | `exc_info`, `extra`, `stack_info`, `stacklevel` |

Within Python's `Logger` implementation, the `log` method calls the `_log` method where it uses the `exc_info`, `extra`, `stack_info`, and `stacklevel` in Python 3.8+. The ARMI framework `log` method mirrors this implementation by implementing a `_log` method and calling it, but the main difference is that the Python's standard `log` method accepts `*args` and `**kwargs` to handle future changes.

To fix this issue with the Sphinx logging, I recommend that we at a minimum implement the `extra` kwarg to the ARMI framework's logger. Without this being implemented, I get the following error message when building my ARMI applications documentation:

```
Exception occurred:
  File "C:\tools\python379\lib\logging\__init__.py", line 1783, in log
    self.logger.log(level, msg, *args, **kwargs)
TypeError: log() got an unexpected keyword argument 'extra'
[...]

PR #519, opened 2021-12-28
Add the `extra` keyword argument to ARMI's logging.

jakehader:
## Description

See #518. This updates the ARMI logger to handle the `extra` keyword argument that is implemented in Python's standard `Logger` implementation and used by the Sphinx package.

## Checklist

- Tested by building the ARMI framework documents manually.

### DIFF
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -145,7 +145,7 @@ class _RunLog:
             except AttributeError:
                 exec(_ADD_LOG_METHOD_STR.format(longLogString, logValue))
 
-    def log(self, msgType, msg, single=False, label=None):
+    def log(self, msgType, msg, single=False, label=None, extra=None):
         """
         This is a wrapper around logger.log() that does most of the work and is
         used by all message passers (e.g. info, warning, etc.).
@@ -160,7 +160,7 @@ class _RunLog:
         msg = str(msg)
 
         # Do the actual logging
-        self.logger.log(msgLevel, msg, single=single, label=label)
+        self.logger.log(msgLevel, msg, single=single, label=label, extra=extra)
 
     def getDuplicatesFilter(self):
         """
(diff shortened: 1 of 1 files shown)


ntouran (commented):
Is it possible to just use kwargs for this? I was looking through the python logging docs for a official description of extra that we could point to but they just show the logging sigs with args and kwargs. We could unpack those in the low-level log function if we needed to and then maybe stay future proof with logging. 

Also, I'm interested in whether or not we should use https://www.sphinx-doc.org/en/master/extdev/logging.html to fix sphinx logging conflicts. 

PS you should uncheck the "added new tests" check in the checklist since you only did manual testing. Feel free to still say that you did test manually, but that doesn't tick that checkbox.

jakehader:
> Is it possible to just use kwargs for this? I was looking through the python logging docs for a official description of extra that we could point to but they just show the logging sigs with args and kwargs. We could unpack those in the low-level log function if we needed to and then maybe stay future proof with logging.
> 
> Also, I'm interested in whether or not we should use https://www.sphinx-doc.org/en/master/extdev/logging.html to fix sphinx logging conflicts.
> 
> PS you should uncheck the "added new tests" check in the checklist since you only did manual testing. Feel free to still say that you did test manually, but that doesn't tick that checkbox.

I added a comment on the ticket and unchecked the box for tests.

I think we could just use args and kwargs in our implementation, but then our own implementation of `single` and `label` would be abstracted, right? Maybe these should just be reconfigured so that they are implemented within the `extra` dictionary and then unpacked as needed. This seems to be the way that the `extra` dict is meant to be used.

john-science:
It looks like ~55 unit tests failed with (roughly) the same error:

```python
_____________________ C5G7ReactorTests.test_runAndLoadC5G7 _____________________
  File "/home/runner/work/armi/armi/armi/settings/settingsIO.py", line 199, in readFromFile
    self.readFromStream(f, handleInvalids, self.format)
  File "/home/runner/work/armi/armi/armi/settings/settingsIO.py", line 221, in readFromStream
    self._checkInvalidSettings()
  File "/home/runner/work/armi/armi/armi/settings/settingsIO.py", line 297, in _checkInvalidSettings
    runLog.warning("Ignoring invalid settings: {}".format(invalidNames))
  File "/home/runner/work/armi/armi/armi/runLog.py", line 376, in warning
    LOG.log("warning", msg, single=single, label=label, extra=extra)
TypeError: log() got an unexpected keyword argument 'extra'

During handling of the above exception, another exception occurred:

  File "/home/runner/work/armi/armi/armi/tests/test_lwrInputs.py", line 91, in test_runAndLoadC5G7
    o = armi_init(fName=TEST_INPUT_TITLE + ".yaml")
  File "/home/runner/work/armi/armi/armi/__init__.py", line 160, in init
    cs = settings.Settings(fName)
  File "/home/runner/work/armi/armi/armi/settings/caseSettings.py", line 92, in __init__
    self.loadFromInputFile(fName)
  File "/home/runner/work/armi/armi/armi/settings/caseSettings.py", line 232, in loadFromInputFile
    reader.readFromFile(fName, handleInvalids)
  File "/home/runner/work/armi/armi/armi/settings/settingsIO.py", line 201, in readFromFile
    raise InvalidSettingsFileError(path, str(ee))
armi.utils.customExceptions.InvalidSettingsFileError: Attempted to load an invalid settings file from: c5g7-settings.yaml. log() got an unexpected keyword argument 'extra'
```

I believe it is possible you added the `extra` keyword to the `_RunLog.log()` method, but not to the `RunLogger.log()` method.  And I wonder if that would be enough to fix your problem.

jakehader:
Closing this PR - See comments on #518.

PR #521, opened 2021-12-31
MPI unit tests finally working again

### LINKED ISSUE #516: MPI Unit Tests not working
We have several unit tests that are supposed to use our parallelization library (MPI) but have been deactivated for a year now, e.g.:

https://github.com/terrapower/armi/blob/3bfaa328299e7c672ab852c1c2bab314149976d5/armi/operators/tests/test_operators.py#L128-L130

Since this library is an optional requirement, we don't want to run them all by default (for developers who don't want to do the big, system-wide MPI install). So, we might want to add lines like these to all MPI-based unit tests, so they only run on machines with MPI installed:

```python
from distutils.spawn import find_executable

# determine if MPI is installed
MPI_EXE = None
if find_executable('mpiexec.exe') is not None:
    MPI_EXE = "mpiexec.exe"  # Windows
elif find_executable('mpiexec') is not None:
    MPI_EXE = "mpiexec"  # Linux
 
    @unittest.skipIf(MPI_EXE is None, "mpiexec is not in path.")
     def testMpiOperator(self):
```

Something like that solves the problem of how/when to run these tests.  But the problem still exists that all of our MPI tests appear to be broken.  My theory is that the tests are so old the code has changed around them and now we just have code rot.  But we need the code coverage on all this MPI-based code.

Here are some rotten MPI-based tests that need fixing:

* [ ] `armi/operators/tests/test_operators.py::MpiOperatorTests`
* [ ] `armi/tests/test_mpiActions.py::MpiDistributeStateTests`

john-science:
## Description

For the past two years, the MPI-based unit tests have been broken. This PR gets them running again. (And I added some more unit tests just to fill out our code coverage.)

This involved lots of trial an error, and was kind of a big lift. I can see why these haven't been running in a while. As MPI-based unit tests require installing the optional MPI library system-wide, I chose to set these MPI unit tests up in a special file that will be where we put our MPI-based tests. And the GitHub Actions connection I make to these unit test only test in a Linux environment, not a Windows one, as I didn't think testing in multiple operating systems would provide much benefit.

This PR will close #516 

---

## Checklist

- [X] GitHub Actions has been updated for the new tests.


john-science:
I'm not sure why the coverage build above failed.  It passed like 6 times on my fork of the repo.

I'll look into it.

```
INTERNALERROR>     raise DataError(f"Couldn't use data file {self.filename!r}: {msg}") from exc
INTERNALERROR> coverage.exceptions.DataError: Couldn't use data file '/home/runner/work/armi/armi/.coverage.mpi.fv-az249-374.4686.747253': no such table: file
```

ntouran (approved):
Amazing. thanks. we needed those tests back.

john-science:
I have made a fork of this branch in my repo, and I'm testing why the Coverage might have failed here.  It might be slow to test, as the failure is clearly only rarely, not 100% of the time.

Some sort of race condition with the log file location, it looks like.

john-science:
Okay, after spending FAR too much time on this, I am going to make sure we run the MPI unit tests in GitHub Actions. BUT we are not going to report code coverage on them.

This makes me a little sad.  But this appears to be a long-standing issue. Most teams have done the same and dropped support for code coverage on their MPI tests.  It seems like you CAN get it working if you use and old version of `coverage<4.0`, but then you lose support for Python 3.8+, which we can't stomach.

I'm happy the MPI tests will be working again, and maybe there will be an update to the `coverage` library that will allow us to report code coverage on our MPI code. But today is not that day.

Issue #520, opened 2021-12-29
Need some checks on the class1/class2 material modifications


keckler:
For materials derived from the [`FuelMaterial` class](https://terrapower.github.io/armi/_modules/armi/materials/material.html#FuelMaterial), a user can blend two materials together using the class1/class2 material modifications (see [here](https://terrapower.github.io/armi/user/inputs/blueprints.html#assemblies)). The `applyInputParams` method used to accomplish this could use some checks on the material modifications that are supplied by the user.

The following checks would be very useful:
1. Make sure that `class1_wt_frac` is between 0 and 1 (inclusive), otherwise negative isotopic densities can occur.
2. Make sure that `class1_custom_isotopics` and `class2_custom_isotopics` point to existing custom material definitions. This is very important because if the user points to a non-existing custom material, the code errors but catches an exception in the [_constructMaterial](https://terrapower.github.io/armi/_modules/armi/reactor/blueprints/componentBlueprint.html#ComponentBlueprint._constructMaterial) call, so `applyInputParams` fails silently and the code continues as if nothing went wrong.

I'd be happy to implement the checks myself, if others agree that they are needed.

keckler:
Also pointing out that this is different from #402. In that case it was the material modification key that leads to the problem, whereas in this case it is the value. In this case, the material modification names `class1_custom_isotopics` and `class2_custom_isotopics` are valid and will trigger ARMI to go off and do something, but the problem occurs because ARMI can't properly handle the supplied value. So the material modification is technically "used", it's just that nothing results of it.

keckler:
It seems like both checks could be put straight into the `applyInputParams` method of the `FuelMaterial` class. If that were the case, then the checks would run at the beginning of a case when the block blueprints are constructed. One set of checks for every `FuelMaterial` component in each block blueprint. So it's not like the check would be called at every burn step, or even at every burn cycle. For a typical case that has only a handful of unique block blueprints, I don't think it should be very onerous. 

I will try to implement this and see how much overhead it requires. I have a rather big case with a lot of blueprints that I'll test it out on.

jakehader:
Is this closed now or still more work to do?

keckler:
Closed! Thanks for the reminder!

PR #524, opened 2022-01-05
Checks on class1/class2 material modifications

keckler:
## Description
To address #520 , checks have been implemented within `applyInputParams` of the `FuelMaterial` class. The checks are quite lightweight and do not add much at all to run time. The method was re-tooled a little bit so that the bulk of the method is only executed if the material modification is actually input by the user.

Tests on the method have also been added.

---

## Checklist


john-science (approved):
I like it.  It seems clear and concise.

I might have pulled all the new conditional logic out into it's own method.  I just like shorter methods more than longer ones.

john-science:
The only thing I can't tell is if there is some way this could break existing (valid) runs.

I guess I can't think of anything, and all the unit tests pass. So that should be enough.

keckler:
All I can say wrt breaking existing runs is that I tested it against my own cases with wide variety of (SFR) assembly designs, and it worked fine. And like you said, the tests pass. So if it breaks something, I did all I could!

PR #525, opened 2022-01-06
Add component group capabilities

ntouran:
## Description

This just allows the user input of component groups intended to mix
different free components in different fractions. As is, this only
allows the input but does not implement any actual construction of
objects. It is a follow-up to #505 as part of the larger #504 task
related to improving input flexibility.

This needs to have a test that shows it reading/writing to/from
the database.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -834,7 +834,7 @@ class Block(composites.Composite):
         self.p.enrichmentBOL = self.getFissileMassEnrich()
         massHmBOL = 0.0
         sf = self.getSymmetryFactor()
-        for child in self:
+        for child in self.iterComponents():
             child.p.massHmBOL = child.getHMMass() * sf  # scale to full block
             massHmBOL += child.p.massHmBOL
         self.p.massHmBOL = massHmBOL
@@ -910,10 +910,14 @@ class Block(composites.Composite):
 
         self.derivedMustUpdate = True
         self.clearCache()
-        mult = int(c.getDimension("mult"))
-        if self.p.percentBuByPin is None or len(self.p.percentBuByPin) < mult:
-            # this may be a little wasteful, but we can fix it later...
-            self.p.percentBuByPin = [0.0] * mult
+        try:
+            mult = int(c.getDimension("mult"))
+            if self.p.percentBuByPin is None or len(self.p.percentBuByPin) < mult:
+                # this may be a little wasteful, but we can fix it later...
+                self.p.percentBuByPin = [0.0] * mult
+        except AttributeError:
+            # maybe adding a Composite of components rather than a single
+            pass
         self._updatePitchComponent(c)
 
     def removeAll(self, recomputeAreaFractions=True):
@@ -1972,7 +1976,7 @@ class HexBlock(Block):
         """
 
         # Check multiplicities...
-        mults = {c.getDimension("mult") for c in self}
+        mults = {c.getDimension("mult") for c in self.iterComponents()}
 
         if len(mults) != 2 or 1 not in mults:
             raise ValueError(
--- a/armi/reactor/blueprints/componentBlueprint.py
+++ b/armi/reactor/blueprints/componentBlueprint.py
@@ -17,15 +17,17 @@ This module defines the ARMI input for a component definition, and code for cons
 
 Special logic is required for handling component links.
 """
-import six
 import yamlize
 
 from armi import runLog
 from armi import materials
 from armi.reactor import components
+from armi.reactor import composites
 from armi.reactor.flags import Flags
 from armi.utils import densityTools
 from armi.nucDirectory import nuclideBases
+
+COMPONENT_GROUP_SHAPE = "group"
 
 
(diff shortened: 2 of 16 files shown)


john-science (approved):
I'm okay approving this as long as we keep working on the new components-with-components feature set. It's a good stepping stone.

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -307,6 +320,46 @@ class ComponentKeyedList(yamlize.KeyedList):
...
+    """
+    A single component group containing multiple GroupedComponents
+
+    Example
+    -------
+    triso:
+      kernel:
+        mult: 0.7
+      buffer:
+        mult: 0.3
```
drewj-usnctech, line 344: This feels a little strange and clunky. The name `mult` in this case is more like a multiplier than a multiplicity (number of things). I guess having 0.7 `kernel` is kind of right... 

Secondly, this in theory would mean a user would need to convert the dimension properties to volume fractions. The math is easy but it's another possible step that could go wrong

armi/reactor/blocks.py
```diff
@@ -910,10 +910,14 @@ def add(self, c):
 
         self.derivedMustUpdate = True
         self.clearCache()
-        mult = int(c.getDimension("mult"))
-        if self.p.percentBuByPin is None or len(self.p.percentBuByPin) < mult:
-            # this may be a little wasteful, but we can fix it later...
-            self.p.percentBuByPin = [0.0] * mult
+        try:
+            mult = int(c.getDimension("mult"))
```
drewj-usnctech, line 914: If one wanted to place a fuel compact full of TRISO on a fixed lattice (e.g., #224), it would be useful to leverage the `latticeIDs` feature to model the lattice. In that case, the compact should have a multiplicity reflecting the number of locations in this `Block`

armi/reactor/tests/test_components.py
```diff
@@ -204,10 +204,6 @@ def test_fromComponent(self):
 class TestShapedComponent(TestGeneralComponents):
     """Abstract class for all shaped components"""
 
-    def test_getBoundingCircleOuterDiameter(self):
```
john-science, line 207: Why are we removing this test?

armi/reactor/blueprints/tests/test_blueprints.py
```diff
@@ -128,7 +144,12 @@ class TestBlueprintsSchema(unittest.TestCase):
         Thot: 600.0
         id: 0.0
         mult: 1.0
-        od: 11.0
+        od: 4.0
+
+component groups:
+    group1:
+      freefuel:
+        mult: 1.0
```
ntouran, line 152: probably should do more than one here

armi/reactor/composites.py
```diff
@@ -3142,6 +3151,19 @@ def _genChildByLocationLookupTable(self):
         for child in self:
             self.childrenByLocator[child.spatialLocator] = child
 
+    def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
```
john-science, line 3154: Can we get a unit test for this? 

Thanks!

armi/reactor/components/__init__.py
```diff
@@ -343,28 +350,40 @@ def _deriveVolumeAndArea(self):
...
 
-        remainingArea = self.parent.getMaxArea() - siblingArea
         remainingVolume = self.parent.getMaxVolume() - siblingVolume
+        if siblingArea:
+            remainingArea = self.parent.getMaxArea() - siblingArea
 
-        # Check for negative area
-        if remainingArea < 0:
+        # Check for negative
+        if remainingVolume < 0:
```
john-science, line 365: This negative volume check isn't covered by unit tests.

Thanks!

armi/reactor/blueprints/componentBlueprint.py
ntouran: agreed. we need to make it more in line with your particlefuel class. I think we'll push this through as a stepping stone and then do #540 and then implement something that's basically identical to what you did, but with more first class support/usage of the builtin classes.

armi/reactor/blocks.py
ntouran: totally. I think if we just have a block that has a child called `triso compact` or something that happens to have matrix + kernels in it then it will look just like a 'pin' in non-triso fuel and should work with pin lattices all the same.

armi/reactor/tests/test_components.py
ntouran: because we implemented it! but you're right, we should just update the test to actually test it, eh?

armi/reactor/composites.py
ntouran: done.

armi/reactor/components/__init__.py
ntouran: Added in https://github.com/terrapower/armi/pull/525/commits/2ef113d8cf93b5c7f03704048d70bef94fce0f31

PR #526, opened 2022-01-07
Removing duplicate versioning strings

john-science:
## Description

In the past, every time I have wanted to bump our `armi.__version__`, I've had to do it in two places. This has always made me sad.

We have been duplicating our version string between `setup.py` and `meta.py`. Essentially, we had a lot of comments and documents explaining that putting the `__version__` string in `armi/__init__.py` file meant calling a bunch of code in `setup.py`, and that was deemed inefficient.

While this solution includes one kind of fugly line of Python code, it means we get to remove a lot of confusing comments, and streamline our release process.

---

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,9 +13,14 @@
 # limitations under the License.
 
 """Setup.py script for the Advanced Reactor Modeling Interface (ARMI)"""
-import re
 from setuptools import setup, find_packages
+import os
+import pathlib
+import re
 
+# grab __version__ from meta.py, without calling __init__.py
+this_file = pathlib.Path(__file__).parent.absolute()
+exec(open(os.path.join(this_file, "armi", "meta.py"), "r").read())
 
 with open("README.rst") as f:
     README = f.read()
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -13,9 +13,14 @@
...
 """Setup.py script for the Advanced Reactor Modeling Interface (ARMI)"""
-import re
 from setuptools import setup, find_packages
+import os
+import pathlib
+import re
 
+# grab __version__ from meta.py, without calling __init__.py
+this_file = pathlib.Path(__file__).parent.absolute()
+exec(open(os.path.join(this_file, "armi", "meta.py"), "r").read())
```
ntouran, line 23: maybe worth putting a comment in meta.py saying that each line will be executed line by line during setup.py? probably worth pointing to this at least since it may be hard to find/surprising.

Better to just read the version out of the code with a RE?

ntouran (approved):
approved with comments

setup.py
john-science: 1) Sure, I like the idea of that comment. Though, to be fair, not more code will be "executed" in `meta.py` than if we just did "from armi import meta". That's the only level of execution abound.  I'm just trying to save us from executing everything in `armi/__init__.py` by doing "from armi import __version__", which would also work.
2) We could do regex for this. I have done things like that in the past. But it's always problematic. I prefer to use Python to read Python than to write our own parser to read Python, you know?

PR #528, opened 2022-01-10
Create a separate *.log file for the worker node *.stdouts.

mgjarrett:
## Description
At the end of a run, all of the logs from the individual MPI processes get concatenated together and appended to the end of the `*.stdout`. If there are a lot of MPI processes, this can make the `*.stdout` obnoxiously bloated, with a bunch of information that the user would rarely ever want to look at.

Instead, this PR puts all of those worker node logs into a separate file, `.\logs\armi-mpi-workers.log`
It would probably be best if the file had the name of the case (i.e., the same as the `*.stdout` and `*.stderr`), but I don't know if that name is available to the logger at this level. It's just writing those to the streams `sys.stdout` and `sys.stderr`.  If somebody has a better idea for naming I'm all ears; this was just kind of a temporary name to make sure the code worked.

---

## Checklist

- [ ] The code is understandable and maintainable to people beyond the author

### DIFF
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -314,10 +314,11 @@ def concatenateLogs(logDir=None):
 
     # default worker log name if none is found
     caseTitle = "armi-workers"
-    for stdoutFile in stdoutFiles:
+    for stdoutPath in stdoutFiles:
+        stdoutFile = os.path.normpath(stdoutPath).split(os.sep)[-1]
         prefix = STDOUT_LOGGER_NAME + "."
         if stdoutFile[0 : len(prefix)] == prefix:
-            caseTitle = stdoutFile.split(".")[1]
+            caseTitle = stdoutFile.split(".")[-3]
             break
 
     combinedLogName = os.path.join(logDir, "{}-mpi.log".format(caseTitle))
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -47,12 +47,12 @@ Or change the log level the same way:
 
 """
 from __future__ import print_function
+from glob import glob
 import collections
 import logging
 import operator
 import os
 import sys
-from glob import glob
 
 from armi import context
 
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -249,7 +249,7 @@ class _RunLog:
     def startLog(self, name):
         """Initialize the streams when parallel processing"""
         # open the main logger
-        self.logger = logging.getLogger(STDOUT_LOGGER_NAME + SEP + str(self._mpiRank))
+        self.logger = logging.getLogger(name + SEP + str(self._mpiRank))
 
         # if there was a pre-existing _verbosity, use it now
         if self._verbosity != logging.INFO:
@@ -313,6 +313,8 @@ def concatenateLogs(logDir=None):
 
     caseTitle = settings.getMasterCs().caseTitle
     combinedLogName = os.path.join(logDir, "{}-workers.log".format(caseTitle))
+    print(caseTitle)
+    print(combinedLogName)
     with open(combinedLogName, "w") as workerLog:
         workerLog.write(
             "\n{0} CONCATENATED WORKER LOG FILES {1}\n".format("-" * 10, "-" * 10)
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -312,7 +312,14 @@ def concatenateLogs(logDir=None):
 
     info("Concatenating {0} log files".format(len(stdoutFiles)))
 
-    caseTitle = stdoutFiles[0].split(".")[-3]
+    # default worker log name if none is found
+    caseTitle = "armi-workers"
+    for stdoutFile in stdoutFiles:
+        prefix = STDOUT_LOGGER_NAME + "."
+        if stdoutFile[0 : len(prefix)] == prefix:
+            caseTitle = stdoutFile.split(".")[1]
+            break
+
     combinedLogName = os.path.join(logDir, "{}-mpi.log".format(caseTitle))
     with open(combinedLogName, "w") as workerLog:
         workerLog.write(
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -55,7 +55,6 @@ import os
 import sys
 
 from armi import context
-from armi import settings
 
 
 # global constants
@@ -249,7 +248,9 @@ class _RunLog:
     def startLog(self, name):
         """Initialize the streams when parallel processing"""
         # open the main logger
-        self.logger = logging.getLogger(name + SEP + str(self._mpiRank))
+        self.logger = logging.getLogger(
+            STDOUT_LOGGER_NAME + SEP + name + SEP + str(self._mpiRank)
+        )
 
         # if there was a pre-existing _verbosity, use it now
         if self._verbosity != logging.INFO:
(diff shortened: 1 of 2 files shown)


armi/runLog.py
```diff
@@ -310,20 +310,30 @@ def concatenateLogs(logDir=None):
 
     info("Concatenating {0} log files".format(len(stdoutFiles)))
 
+    dirName = os.path.dirname(stdoutFiles[0])
```
john-science, line 313: Can you add some sort of unit testing for this new code you added?

(There are already a ton of `runLog.py` tests in `armi/tests/test_runLog.py`, you could probably snag what you need there.)

Thanks!

mgjarrett: I added a little to `test_concatenateLogs` to improve the coverage.

john-science:
@ntouran Michael has explained his reasoning, but I am curious what you think of this change to our logging.

ntouran:
This is reasonable in concept for sure. I think it is worth trying to get the same to be synced up with the caseTitle though, because otherwise it will be easy to miss in archiving files, and multiple runs of in the same dir (e.g. param sweeping) will overwrite one another. (though honestly running concurrently in the same dir probably is a bad idea regardless).

john-science:
> I think it is worth trying to get the same to be synced up with the `caseTitle`...

@mgjarrett What do you think of Nick's idea?

mgjarrett:
Yeah I think it would be best to have the `caseTitle` as the name, but I'm not sure exactly how it would be done and I haven't had time to look at it.

armi/runLog.py
```diff
@@ -310,20 +310,29 @@ def concatenateLogs(logDir=None):
...
+        workerLog.write(
+            "\n{0} CONCATENATED WORKER LOG FILES {1}\n".format("-" * 10, "-" * 10)
+        )
+
     for stdoutName in stdoutFiles:
         # NOTE: If the log file name format changes, this will need to change.
         rank = int(stdoutName.split(".")[-2])
-
-        # first, print the log messages for a child process
         with open(stdoutName, "r") as logFile:
```
onufer, line 322: stdoutName is just here, so i think you can get the name from that. you will need to move the above

{0} CONCATENATED WORKER LOG FILES 

into the loop since the overall code allows for multiplied groups of stdouts, but shouldn't be a big deal.

mgjarrett:
`stdoutName` is not the `caseTitle` here. It's something like `ARMI.0001.stdout`, `ARMI.0002.stdout`, etc.

armi/runLog.py
mgjarrett: `stdoutName` is not the `caseTitle` here. It's something like `ARMI.0001.stdout`, `ARMI.0002.stdout`, etc.

john-science:
>  I think it is worth trying to get the same to be synced up with the caseTitle though....

I like this idea.  In retrospect, it seems like the `caseTitle` should be in the output file name for all log files. That would have been a good design.

However... the `caseTitle` is not readily assessible in `runLog.py` right now.  We could call `getMasterCS()`, but I'm fairly sure that will break a bunch of plugins and our docs. Because people want to be able to use our runLogger in all sorts of very general situations.

Maybe that could be called inside `concatLogs()` since that is probably only ever called during a parallel/cluster-type run. That might not break things.  But certainly, let's not call `getMasterCS()` anywhere else.

One question I have is WHY isn't this already set to `caseTitle`?  I notice these two lines exist:

```python
armi/cases/case.py:        runLog.LOG.startLog(self.cs.caseTitle)
armi/operators/operator.py:        runLog.LOG.startLog(self.cs.caseTitle)
```

onufer:
I think startLog is the log in the execution directly. The logs that Mike is reading are in the fast path.

armi/tests/test_runLog.py
```diff
@@ -285,8 +285,10 @@ def test_concatenateLogs(self):
             runLog.concatenateLogs(logDir=logDir)
 
             # verify output
-            combinedLogFile = os.path.join(logDir, "armi-mpi-workers.log")
-            self.assertTrue(os.path.exists(combinedLogFile))
+            #logFile = glob(os.path.join(logDir, "*-mpi.log"))
```
john-science: Can we remove all unused (commented out code) from this PR?

(Though I LOVE that the unit tests were touched!)

ntouran:
Sorry for the late questions on this, but on the discussion level, can we talk a little bit more about the motivation of this? Ideally we'd answer these questions in implementation docstrings in the code itself so it lives on. (technically, this would have been a good issue where we could discuss and then move on to implementation later, but we're not super strict on that workflow :))

* Having ~hundreds of files will be harder to manage and archive for qualified runs. Concatenating them into one provides one record of the run's output and its pedigree (e.g. its hash), which is fairly convenient from a QA point of view.
* Can you explain in more detail how having lots of info in one big stdout is inconvenient? I typically just use find/filtering to filter do the node of interest using my text editor.

mgjarrett:
The goal is not to have ~hundreds of files, but to separate the current  `stdout` into:
!. The master node `stdout` that has everything somebody might usually want to read from `stdout`, like CR worth, k-eff, assembly designs, etc.
2. A separate file that has all of the `stdout`s from the workers, which are usually very repetitive and don't have much new useful information in most cases.

mgjarrett:
The `branchVerbosity` setting might already be a satisfactory solution to the root problem, which is that the `stdout` tends to be too big when `branchVerbosity` is `extra`. `warning` or `error` are probably better options when the user doesn't care about these worker logs.

onufer:
There are times when the workers are useful... so i don't want to get rid of the data entirely. Its just appending all the workers to the main standard does not enable the use of the data though since a lot of time you are just looking at the stdout to see how your run is and then it updates, but the file gets like 50x larger.

Setting workers to "error" sounds like an okay solution, but a lot of times you may want the data, but just don't want to view it at the same time you are viewing the normal stdout.  There should be an option to store the worker data without having to view it with the main stdout data.

ntouran:
I say we go ahead with this PR. Since it's just 2 files, that's not a huge QA burden, and even with branchVerbosity options it can be a bit verbose. If stdout gets too big and you're trying to refresh it over a slow VPN there a good case to be made to have them be separated. 

I would like to see a follow-up ticket, maybe for our internal plugins, that says "get rid of all these totally useless xtra messages in the worker nodes" :)

mgjarrett:
Just a heads up, we're still working on getting the `caseTitle` out of the logger filenames, so it's not fully baked yet.

ntouran:
Ok so is it fully baked now after those last two commits? Ready to be merged?

mgjarrett:
Needed one more commit because the `__init__` for `RunLogger` was throwing away the `caseTitle` in MPI cases and just initializing a logger with the first piece of the name, which is just "ARMI". Running a case now to verify that everything works correctly when the files are concatenated at the end of the run.

john-science:
> Is it something I can fix? It looks like a bunch of files unrelated to this PR.

Can you run this command from within your repo/venv?

```python
black armi
```

That will reformat your changes to make our style guide.  Then you can commit the changes `black` makes to your files:

*  armi/runLog.py
*  armi/tests/test_runLog.py

And push that commit to this branch/PR.

That's the process I use before every commit these days.

Thanks!

mgjarrett:
I have already run `black`. 

```
(nala-dev) C:\Users\mjarrett\codes\nala\framework>black armi
All done! ✨ � ✨
406 files left unchanged.
```

If you look at the output from this `black` run, the changes are all to files that are not in this PR. That's why I don't know what I'm supposed to do here.

john-science:
I have fixed the `black` issue in the ARMI `master` branch. 

It was unrelated to this PR.

PR #534, opened 2022-01-13
Fixing the Setting copy to also copy non-default values.

john-science:
## Description

It appears that a past PR removed a single line of code from `setting.py`, this has lead to a invisibly broken system that works for nearly all unit tests.

I have added a unit test to explicitly check for this important feature in the future.

---

## Checklist

---

## Questions

I am trying to decide if this bug fix should come with a minor version bump (to `0.2.1`).

### DIFF
--- a/armi/meta.py
+++ b/armi/meta.py
@@ -16,4 +16,4 @@
 Metadata describing an ARMI distribution.
 """
 
-__version__ = "0.2.0"
+__version__ = "0.2.1"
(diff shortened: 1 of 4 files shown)


jakehader (approved):
Looks good to me!

john-science:
> Looks good to me! Yeah let's bump the version to 0.2.1 and note that this fixes a immutable settings copy error.

Yeah, the more we talked today, the more I realized that would be helpful.

Thanks!

PR #538, opened 2022-01-14
Add allowMissing kwarg to loadOperator

onufer:
## Description

Adds `kwarg` to `loadOperator` to support more easily loading databases that have parameters not defined within the current ARMI version.  This is helpful for backward compatibility when post-processing a database that had a param that was removed from ARMI .

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/__init__.py
+++ b/armi/bookkeeping/db/__init__.py
@@ -75,7 +75,7 @@ __all__ = [
 ]
 
 
-def loadOperator(pathToDb, loadCycle, loadNode):
+def loadOperator(pathToDb, loadCycle, loadNode, allowMissing=False):
     """
     Return an operator given the path to a database.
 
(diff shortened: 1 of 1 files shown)


john-science:
I'll wait for the unit tests to finish running before approving.

But this looks like a very reasonable change.

It might be nice to have a unit test that flexes this for `allowMissing" = `True` and `False`.  There is already one unit test that uses `loadOperator()`: 

https://github.com/terrapower/armi/blob/b1c18bd2ccfc4898634c5bc771ccc9233fa4ecfc/armi/tests/test_lwrInputs.py#L91-L95

onufer:
@john-science, Yes, the main reason i didn't add a test is that to trully test the feature I would need a database that had params that don't exist in the framework. I dont really want to commit another database to the framework, and dont want to write a test that doesn't actually test the feature. Open to feedback though.

john-science:
@onufer Is it true that we need a whole need database to test this?

Couldn't we just create a new test YAML file that lists the parameters we need, to exclude on of the parameters in the default test database?

Sorry, maybe I'm missing something.

onufer:
@john-science, you can test the method signature but not the feature.

PR #543, opened 2022-01-19
Fix add option

john-science:
## Description

As it turns out, the last fix for copying the value of a setting was problematic for a subset of settings that had a list of custom (non-default) options.  This amounts to something that would only happening when reading certain kinds of YAML files. And the fix is only one character. Never-the-less, I added a unit test to cover this YAML-reading, custom-options issue in the future.

---

## Checklist

### DIFF
--- a/doc/release/0.2.rst
+++ b/doc/release/0.2.rst
@@ -3,17 +3,26 @@ ARMI v0.2 Release Notes
 =======================
 
 
-ARMI v0.2.2
+ARMI v0.2.3
 ===========
 Release Date: TBD
 
-What's new in ARMI v0.2.2
+What's new in ARMI v0.2.3
 -------------------------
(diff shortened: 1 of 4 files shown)


doc/release/0.2.rst
```diff
@@ -20,6 +20,13 @@ ARMI v0.2.2
 ===========
 Release Date: 2022-01-19
 
+What's new in ARMI v0.2.2
+-------------------------
+#. Improved type hinting
+#. Flushed out the ability to build the docs as PDF
+#. Material modifications can now be made per-component
+#. The ``loadOperator`` method now has the optional ``allowMissing`` argument
```
jakehader, line 28: Make a note that this allows for handling backwards compatibility with database loading if parameters are removed from the data model.

PR #545, opened 2022-01-24
Lithium enrichment material modification key

keckler:
## Description
The material modification keyword for enrichment in the `Lithium` material class is ambiguous to the user as to which isotope is being enriched. This is particularly confusing for the case of lithium, because 'enriched lithium' means different things to different people. People who want to use lithium as an absorber typically think 'enriched' refers to Li6, while those who use lithium in coolant materials think 'enriched' refers to Li7.

In addition, the keyword is just inconsistent with all the other material classes which employ enrichment material modifications. In all other existing cases, the isotope is specified in the key.

This PR adds the new `LI6_wt_frac` key and deprecates the old `LI_wt_frac` one. In addition, it adds unit tests on the application of the material modification for this material.

---

## Checklist


keckler:
Wait, something went wrong here. Killing this PR.

PR #546, opened 2022-01-24
Update key for enrichment material modification in `Lithium` material class

keckler:
## Description
The material modification keyword for enrichment in the `Lithium` material class is ambiguous to the user as to which isotope is being enriched. This is particularly confusing for the case of lithium, because 'enriched lithium' means different things to different people. People who want to use lithium as an absorber typically think 'enriched' refers to Li6, while those who use lithium in coolant materials think 'enriched' refers to Li7.

In addition, the keyword is just inconsistent with all the other material classes which employ enrichment material modifications. In all other existing cases, the isotope is specified in the key.

This PR updates the key for the enrichment material modification in the `Lithium` class to avoid the ambiguity and be consistent with the other material classes.

Tests have been added.

---

## Checklist


armi/materials/tests/test_lithium.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2019 TerraPower, LLC
```
john-science, line 1: I believe this should say "Copyright 2022".

john-science:
I have the same feeling about this as I did about the Thorium Oxide PR: it seems good, I wonder why this wasn't necessary before.

I suppose the main concern is: do we think this will break anyone's downstream models?

Perhaps we can talk about that over some sort of a call today?

john-science (approved):
Nice.

Also, thanks again for the unit test coverage of the new code.

armi/materials/lithium.py
```diff
@@ -43,8 +60,8 @@ def density(self, Tk=None, Tc=None):
         return 0.512  # g/cc
 
     def setDefaultMassFracs(self):
-        self.setMassFrac("LI6", 0.075)
-        self.setMassFrac("LI7", 0.925)
+        self.setMassFrac("LI6", nb.byName["LI6"].abundance)
```
john-science, line 63: How do we know this "byName" dict won't be empty?

Sorry if that's an obvious question, it's the first time I've seen this dictionary, and by default, it starts empty:

https://github.com/terrapower/armi/blob/master/armi/nucDirectory/nuclideBases.py#L118

keckler: When you import the `nuclideBases` module, the dict is populated. 

To be honest, the logic inside of `nuclideBases` that makes this happen is a little confusing to me, but it definitely happens when the module is initialized, based on my testing.

john-science: Ack. I see, it's during the top-level project bootstrap:

https://github.com/terrapower/armi/blob/276c48fc11b582869dd1f7894c9fd989a0702e1a/armi/_bootstrap.py#L69

Well, that's not my favorite thing. But that's not your fault.

Thanks!

PR #547, opened 2022-01-24
Fixing strange parallelism in folder deletion

davis8dd:
## Description

---

## Checklist

### DIFF
--- a/armi/utils/pathTools.py
+++ b/armi/utils/pathTools.py
@@ -263,23 +263,30 @@ def cleanPath(path):
             "You tried to delete {0}, but it does not seem safe to do so.".format(path)
         )
 
-    for i in range(3):
-        try:
-            if os.path.exists(path) and os.path.isdir(path):
-                shutil.rmtree(path)
-            elif not os.path.isdir(path):
-                # it's just a file. Delete it.
-                os.remove(path)
-        except:
-            if i == 2:
-                pass
-            # in case the OS is behind or something
+    try:
+        if armi.MPI_RANK == 0:
+            clearPath(path)
+        while os.path.exists(path):
             sleep(0.1)
+        armi.MPI_COMM.barrier()
+    except:
+        for i in range(3):
+            try:
+                if os.path.exists(path) and os.path.isdir(path):
+                    shutil.rmtree(path)
+                elif not os.path.isdir(path):
+                    # it's just a file. Delete it.
+                    os.remove(path)
+            except:
+                if i == 2:
+                    pass
+                # in case the OS is behind or something
+                sleep(0.1)
 
-        sleep(0.3)
-        if not os.path.exists(path):
-            break
-        sleep(0.3)
+            sleep(0.3)
+            if not os.path.exists(path):
+                break
+            sleep(0.3)
 
     if os.path.exists(path):
         return False
--- a/armi/utils/tests/test_pathTools.py
+++ b/armi/utils/tests/test_pathTools.py
@@ -72,7 +72,10 @@ class PathToolsTests(unittest.TestCase):
         self.assertTrue(pathTools.moduleAndAttributeExist(thisFile + ":PathToolsTests"))
 
     def test_cleanPathOnValidPaths(self):
-        """Test cleanPaths by cleaning a path for a file, an empty folder, and a file containing a folder."""
+        """Test cleanPaths by cleaning a path for a file, an empty folder, and a file containing a folder.
+        Test in an initialized armi context to include MPI."""
+        import armi
+
         testParentFolderPath = tempfile.mkdtemp(
             dir=str(os.path.join(THIS_DIR)), prefix="testCleanPath"
         )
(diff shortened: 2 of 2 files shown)


john-science:
@davis8dd  Thanks, David!   I was a bit busy yesterday, but I will review it today!

armi/utils/pathTools.py
```diff
@@ -263,23 +263,30 @@ def cleanPath(path):
...
-                shutil.rmtree(path)
-            elif not os.path.isdir(path):
-                # it's just a file. Delete it.
-                os.remove(path)
-        except:
-            if i == 2:
-                pass
-            # in case the OS is behind or something
+    try:
+        if armi.MPI_RANK == 0:
```
john-science, line 267: Does this function work?

If `MPI_RANK` is `0`, then this function calls itself in an infinite loop, right?

So if the program is only single-threaded (meaning it doesn't have any `MPI_RANK > 0`, then wouldn't this function (a) spin forever, and (b) fail to delete the folder?

I could be missing something.

armi/utils/tests/test_pathTools.py
```diff
@@ -70,6 +71,57 @@ def test_moduleAndAttributeExist(self):
         self.assertTrue(pathTools.moduleAndAttributeExist(thisFile + ":THIS_DIR"))
         self.assertTrue(pathTools.moduleAndAttributeExist(thisFile + ":PathToolsTests"))
 
+    def test_cleanPathOnValidPaths(self):
+        """Test cleanPaths by cleaning a path for a file, an empty folder, and a file containing a folder.
+        Test in an initialized armi context to include MPI."""
+        import armi
```
john-science, line 77: This import should got at the top of the file.

armi/utils/tests/test_pathTools.py
```diff
@@ -70,6 +71,57 @@ def test_moduleAndAttributeExist(self):
...
+        self.assertFalse(os.path.exists(folderWithFileInside))
+        self.assertFalse(os.path.exists(testFolderPath))
+
+        os.close(fileFd)
+        os.rmdir(testParentFolderPath)
+
+    def test_cleanPathOnInvalidPath(self):
+        """Test cleanPath when passed an 'invalid' path not present in the list of valid paths inside cleanPath.
+        Test in an initialized armi context to include MPI."""
+        import armi
```
john-science, line 120: This import should go at the top of the file.

armi/utils/tests/test_pathTools.py
john-science, line 77: This test doesn't run without this import?

Did you try?

armi/utils/pathTools.py
```diff
@@ -263,23 +263,30 @@ def cleanPath(path):
...
+            clearPath(path)
+        while os.path.exists(path):
             sleep(0.1)
```
john-science, line 270: This bare `sleep(0.1)` gives me pause.

I'm not sure these bare `sleep()` calls are 100% sufficient.  What if this loop spins for an hour? Surely that should be a FAIL, right?

Should we perhaps have a max amount of time the loop can spin, waiting for the deletion, before it fails? 10 minutes? What do you think?

armi/utils/tests/test_pathTools.py
davis8dd: It would run, but it could not test MPI execution the way it is currently written.  It seems to me like there tend to be fewer integration-level tests rather than many unit tests, so I have written the test this way.  I'm happy to add unit tests that mock the global armi object and separately test the MPI execution path and non-MPI path.

john-science: Hello @davis8dd !

1. GitHub Actions are not a good place for multi-threaded or multi-process code. And it's also not well-suited to integration tests.  So every project I've ever seen uses GitHub Actions more for unit tests.  This goes doubly so for big, scientific models whose integration tests might take days spread across many nodes in a huge computing cluster.
2. I definitely like the idea of having unit tests that test this method in the MPI and no-MPI pathways directly, even if we have to fake that a bit in a no-MPI unit test. Have you seen this class of unit tests, that does such testing?

https://github.com/terrapower/armi/blob/fd5d2dd31051206ce62696027f05a9304b499677/armi/tests/test_mpiActions.py#L30-L40

armi/utils/pathTools.py
davis8dd: The method does need to be updated to use error handlers and explicitly catch exceptions to cover cases such as a file locked by another process... I'll add these.
Any sleep is undesirable, but might be good to have (along with logging) to account for situations such as a  "lag" between the OS response and actual deletion on the disk, which appears to be an issue on Windows.  My guess is at most a delay of 0.1 to 1 seconds would be acceptable to a user, before returning an error.

john-science: > My guess is at most a delay of 0.1 to 1 seconds would be acceptable to a user, before returning an error.

1 second won't be enough. If the code is run on a large enough cluster there can be big IO delays if, says, 1000 cores are writing to the same disk at the same time. Perhaps 1 minute.

armi/utils/tests/test_pathTools.py
davis8dd: Thank you for the feedback! I'll take another look at the test_mpiActions.py file and begin writing new tests.

john-science:
I'm afraid this PR was kind of left in stasis for too long, and the problem was fixed elsewhere.

Thanks for volunteering!

PR #548, opened 2022-01-25
`ThoriumOxide` upgrades

keckler:
## Description
The `ThoriumOxide` material class is very rudimentary right now. This PR upgrades some of the methods in the `ThoriumOxide` material class to make it more usable with typical ARMI workflows.

Make the following changes:
- theoretical density changes are enabled through the material modifications using `TD_frac`
- change the material to a `FuelMaterial`
- allow for type1/type2 material modifications
- add `linearExpansionPercent` and enable thermal expansion to be used
- subclass the class to allow users to also use `ThO2` as a valid name

None of the material property correlation were upgraded.

Also a small, unrelated typo in the `UraniumOxide` class was fixed as part of this, just because I happened to find it.

---

## Checklist


john-science:
1. These seem like reasonable and decent changes, and I'm happy with them.
2. I can't help but why this wasn't already done.

armi/materials/tests/test_thoriumOxide.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2019 TerraPower, LLC
```
john-science, line 1: I believe this Copyright should be 2022.

john-science (approved):
Looks good!

And thanks so much for the extra unit tests!

Issue #554, opened 2022-02-02
`plotFaceMap` arguments `minScale` and `maxScale` don't do anything


keckler:
In the plotting module, there is the [`plotFaceMap`](https://github.com/terrapower/armi/blob/master/armi/utils/plotting.py#L180) method. The method accepts two arguments that are supposed to allow a user to change the color scaling of the printed plots, `minScale` and `maxScale`. 

I have found that using these arguments don't do anything, and the printed plots are always auto-scaled to the minimum and maximum values in the provided `data` argument. I think the root of the problem is these lines:

https://github.com/terrapower/armi/blob/3f0e63769abb2a57b3aac7a2f0975dcd629b015a/armi/utils/plotting.py#L322-L325

First of all, I think the last line there should be within an `else` clause of the conditional. Right now the autoscaling is just clobbering any bounds that were set above in the if-statement. But second, I think some additional work needs to be done in order to allow for only one of `minScale` and `maxScale` to be provided, as is implied by the `or` within the conditional.

Here's an example using the [`zpprTest.yaml` in the tests directory](https://github.com/terrapower/armi/blob/master/armi/tests/zpprTest.yaml):

```
>> import armi; armi.configure(); o = armi.init(fName='zpprTest.yaml')
>> data = [a.p.assemNum for a in o.r.core]
>> labels = [a.p.assemNum for a in o.r.core]
>> armi.utils.plotting.plotFaceMap(o.r.core, data=data, labels=labels)
```
![image](https://user-images.githubusercontent.com/21959600/152191221-d353e45d-4ef9-4e8a-b240-ad868a847a09.png)
```
>> armi.utils.plotting.plotFaceMap(o.r.core, data=data, labels=labels, maxScale=1)
```
![image](https://user-images.githubusercontent.com/21959600/152191380-1f4d1577-9325-4503-8e63-9ec190b9f117.png)
```
>> armi.utils.plotting.plotFaceMap(o.r.core, data=data, labels=labels, maxScale=666)
```
![image](https://user-images.githubusercontent.com/21959600/152191516-e60aaef5-a17d-454a-95a3-479eda8f4881.png)
```
>> armi.utils.plotting.plotFaceMap(o.r.core, data=data, labels=labels, minScale=666, maxScale=666)
```
![image](https://user-images.githubusercontent.com/21959600/152191646-e0d4546b-278a-474e-b20d-016090a4d173.png)

Nothing changes between plots!

PR #555, opened 2022-02-02
Fixing plotFaceMap issue

john-science:
## Description

There is a missing `else` in the `plotFaceMap()` method.  Please see this Issue #554 

---

## Checklist


john-science:
@keckler This is in response to your issue #554

keckler:
Question -- Do we not now need to handle the case when the user specifies one of `minScale`/`maxScale` but not the other? Just looking at the code, it looks like the scaling will break. But maybe I'm wrong, I didn't test it.

PR #558, opened 2022-02-04
Improving Helix constructor to be more correct

john-science:
## Description

The constructor of the `Helix` class is a bit off. The default value for `mult` should be 1.0. And it does not make physical sense to try and fully define the geometry of a helix without defining non-`None` values for: `od`, `helixDiameter`, and `axialPitch`. So all three of those are now mandatory, with no defaults.

---

## Checklist


john-science:
This PR will close https://github.com/terrapower/armi/discussions/533

PR #559, opened 2022-02-04
`ISOXX` parsing

keckler:
## Description
ARMI uses `glob` to parse the current directory for any `ISOXX` files by trying to match against `ISO*`.  Then ARMI tries to extract the last two characters of the matches, under the assumption that `glob` only returned files that look like `ISOXX`. The problem with this is that `glob` might return other things, in which case the regular expression search returns `None`, the call to `group` fails, and ARMI crashes, even if there were legitimate `ISOXX` files in the directory to operate on. 

This was found due to my specific use-case, but I imagine it is not so uncommon. In some Windows environments, [glob is apparently not case-sensitive](https://jdhao.github.io/2019/06/24/python_glob_case_sensitivity/). So if someone has a file that starts with `iso*` (for instance, the words 'isotope' or 'isotopics'), then `glob` will return it in addition to any `ISOXX` files, and then ARMI can't handle it and crashes.

So I added an exception handler to skip any failures in the regular expression parsing, which prints a debug message if the exception is reached.

---

## Checklist


john-science:
That seems like a fine change.

The code coverage is failing because you added that `debug` line, but it doesn't get hit in a unit test.

We should probably cover it in unit tests, or remove it. I noticed that a few lines above this, we have a very similar case-sensitive check for the same file names:

https://github.com/terrapower/armi/blob/22ebd1a2b6bdd0ebfd14b2e56273f8351db57f82/armi/nuclearDataIO/xsLibraries.py#L134

So, maybe we can get away with your `try/except`. Or maybe we should pull the above line and your new line out into a special helper `staticmethod` to test for that file name?

Your call, whichever seems best.

Thanks for the fix!

jakehader:
I wonder if there is a more intelligent way to detect if this is an ISOTXS file. As in, I wonder if we can perform a quick file open/read without reading in the entire file. If the file looks like an ISOTXS file based on it's first card / header then we can close and proceed, otherwise we can just skip.

This method/function has been a pain to deal with since we are trying to use a naming convention to detect which files are the valid ones to include.

keckler:
Yeah, I think we could probably do something like that. The only thing I would say, though, is we will still likely want to have some sort of `try`/`except` statement in there when opening and reading the file, because if we are reading any random file, who knows what might be inside of it.

So I guess I don't see it as being so much different than what we're currently doing. 

That's just my opinion, though, and if others have different opinions than I am open to going that route.

jakehader:
I think what you have now is good to cover your case so no need to change again now. I was sort of thinking aloud w.r.t. to quick opening CCCC type files in general.

jakehader (approved):

jakehader:
Could be good to add a unit test to this. There should be tests for this function already so just consider extending to it. I see no reason that this won't work though. I'll wait to merge until a simple test is written.

keckler:
Yeah, I'm working on a unit test. Hopefully tomorrow. It ended up being a bit more complicated than I expected to test it, but mostly just because I'm a noob.

john-science:
Thanks for working on the test!

As far as the `try`/`except` goes, maybe we should at least be consistent, and have them in all the places in this file we detect `ISOXX` files, or none of them. You know?

Anyway, thanks for working on the ticket, you rock!

keckler:
@john-science @jakehader 

Test is up.

keckler:
Grrr, failing. I think I know why, let me see...

keckler:
Okay, good to go now!

armi/nuclearDataIO/tests/test_xsLibraries.py
```diff
@@ -245,6 +245,24 @@ def test_mergeFailsWithNonIsotxsFiles(self):
...
+        with open(dummyFileName, "w") as file:
+            file.write(
+                "This is a file that starts with the letters 'ISO' but will"
+                " break the regular expression search."
+            )
+
+        try:
+            with mockRunLogs.BufferLog() as log:
+                lib = xsLibraries.IsotxsLibrary()
+                xsLibraries.mergeXSLibrariesInWorkingDirectory(lib, "ISOTXS", "")
```
jakehader, line 258: I think the inputs into this function should be (lib, "", False) because there shouldn't be any library suffix required for this to fail. Additionally, the final kwargs is a boolean even though an empty string does resolve to False. Can you try this to see if the test still fails?

jakehader (changes requested):

armi/nuclearDataIO/tests/test_xsLibraries.py
keckler: Yeah, you're right about the inputs. I just copied them from the test above this one, but you're right that they are weird, so I changed them to use the defaults and pushed a new version of the test just now.

keckler: Also yes, the test still passes with the updated inputs.

jakehader (approved):

PR #572, opened 2022-02-11
Provide ArmiPlugin.defineSystemGrid builders

drewj-usnctech:
## Description

---
Provide ArmiPlugin.defineSystemGrid builders

Goal is to let a plugin define additional sections that
can reside in the ``systems`` section, e.g.,
```yaml
systems:
    core:
        grid name: core
        origin:
            x: 0
            y: 0
            z: 0
    sfp:
        grid name: spent fuel pool
        type: sfp
        ...
```

The end goal is to have a new system for control drums, pointing to a designated
grid, and allowing assemblies to be added to that grid rather than the core grid.

The plugins basically define a map from the system grid type (defaults to "core"
in the blueprint) and the builder function. The builder function will most likely
be the class of the thing to be built (e.g., ``"core"`` returns a ``Core`` **type**
and the blueprint construct method creates the instance given the name of the grid.
In the core example it's not super helpful. But for the spent fuel pool, the
``type`` is ``sfp`` but the name is defined by the key one level up and does not
have to match any specific criteria.

The `ReactorPlugin` defines the existing ``"core"`` and ``"sfp"`` builders, but is
tagged to go last (or near last) in the YAML construction sequence. This allows other
plugins to define their own core builders and let those have preference over the
default.

Signed-off-by: Andrew Johnson <a.johnson@usnc-tech.com>

## Checklist

- [ ] The code is understandable and maintainable to people beyond the author.
- [ ] There is no commented out code in this PR.
- [ ] The commit message follows [good practices](https://terrapower.github.io/armi/developer/tooling.html).

### DIFF
--- a/armi/plugins.py
+++ b/armi/plugins.py
@@ -120,12 +120,15 @@ the Plugin-based architecture, and as the need arise may be migrated to here.
    user settings. This also predated the plugin infrastructure, and may one day be
    replaced with plugin-based fuel handler logic.
 """
-from typing import Dict, Union
+from typing import Dict, Union, Callable, Optional, TYPE_CHECKING
 
 import pluggy
 
 from armi import pluginManager
 from armi.utils import flags
+
+if TYPE_CHECKING:
+    from armi.reactor.composites import Composite
 
 
 HOOKSPEC = pluggy.HookspecMarker("armi")
@@ -560,6 +563,37 @@ class ArmiPlugin:
 
         blueprint : Blueprint, optional
             for a reactor (if None, only partial contents created)
+        """
+
+    @staticmethod
+    @HOOKSPEC
+    def defineSystemGridBuilders() -> Dict[str, Callable[[str], "Composite"]]:
+        """
+        Convert a user-string from the systems section into a valid composite builder
+
+        Parameters
+        ----------
+        name : str
+            Name of the system type defined by the user, e.g., ``"core"``
+
+        Returns
+        -------
+        dict
+            Dictionary that maps a grid type from the input file (e.g., ``"core"``)
+            to a function responsible for building a grid of that type, e.g.,
+
+            .. code::
+
+                {
+                    "core": armi.reactor.reactors.Core,
+                    "sfp": armi.reactor.assemblyLists.SpentFuelPool,
+                }
+
+        Notes
+        -----
+        The default :class:`~armi.reactor.ReactorPlugin` defines a ``"core"`` lookup
+        and a ``"sfp"`` lookup, triggered to run after all other hooks have been run.
+
         """
 
 
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -188,7 +188,7 @@ class Block(composites.Composite):
 
         core = self.core
         if core is None:
-            return None
+            return self.getAncestor(lambda o: isinstance(o, Reactor))
 
         if not isinstance(core.parent, Reactor):
             raise TypeError(
--- a/armi/reactor/blueprints/reactorBlueprint.py
+++ b/armi/reactor/blueprints/reactorBlueprint.py
@@ -74,25 +74,29 @@ class SystemBlueprint(yamlize.Object):
 
     @staticmethod
     def _resolveSystemType(typ: str):
-        # avoid circular import, though ideally reactors shouldnt depend on blueprints!
-        from armi.reactor import reactors
-        from armi.reactor import assemblyLists
+        # Loop over all plugins that could be attached and determine if any
+        # tell us how to build a specific systems attribute. Sub-optimial
+        # as this check is called for each system (e.g., core, spent fuel pool).
+        # It is assumed that the number of systems is currently low enough to justify
+        # this structure.
 
-        # TODO: This is far from the the perfect place for this; most likely plugins
-        # will need to be able to extend it. Also, the concept of "system" will need to
-        # be well-defined, and most systems will probably need their own, rather
-        # dissimilar Blueprints. We may need to break with yamlize unfortunately.
-        _map = {"core": reactors.Core, "sfp": assemblyLists.SpentFuelPool}
+        manager = armi.getPluginManagerOrFail()
 
-        try:
-            cls = _map[typ]
-        except KeyError:
-            raise ValueError(
-                "Could not determine an appropriate class for handling a "
-                "system of type `{}`. Supported types are {}.".format(typ, _map.keys())
-            )
+        # Only need this to handle the case we don't find the grid we expect
+        seen = set()
+        for options in manager.hook.defineSystemGridBuilders():
+            for key, builder in options.items():
+                # Take the first match we find. This would allow other plugins to
+                # define a new core builder before finding those defined by the
+                # ReactorPlugin
+                if key == typ:
+                    return builder
+                seen.add(key)
 
-        return cls
+        raise ValueError(
+            "Could not determine an appropriate class for handling a "
+            "system of type `{}`. Supported types are {}.".format(typ, sorted(seen))
+        )
 
     def construct(self, cs, bp, reactor, geom=None):
         """Build a core/IVS/EVST/whatever and fill it with children."""
--- a/armi/reactor/__init__.py
+++ b/armi/reactor/__init__.py
@@ -30,7 +30,14 @@ The key classes of the reactor package are shown below:
 See :doc:`/developer/index`.
 """
 
+from typing import Dict, Callable, Union, TYPE_CHECKING
+
 from armi import plugins
+
+# Provide type checking but avoid circular imports
+if TYPE_CHECKING:
+    from armi.reactor.reactors import Core
+    from armi.reactor.assemblyLists import SpentFuelPool
 
 
 class ReactorPlugin(plugins.ArmiPlugin):
@@ -66,3 +73,16 @@ class ReactorPlugin(plugins.ArmiPlugin):
             (CartesianBlock, CartesianAssembly),
             (ThRZBlock, ThRZAssembly),
         ]
+
+    @staticmethod
+    @plugins.HOOKIMPL(trylast=True)
+    def defineSystemGridBuilders() -> Dict[
+        str, Callable[[str], Union["Core", "SpentFuelPool"]]
+    ]:
+        from armi.reactor.reactors import Core
+        from armi.reactor.assemblyLists import SpentFuelPool
+
+        return {
+            "core": Core,
+            "sfp": SpentFuelPool,
+        }


armi/reactor/blocks.py
```diff
@@ -188,7 +188,7 @@ def r(self):
 
         core = self.core
         if core is None:
-            return None
+            return self.getAncestor(lambda o: isinstance(o, Reactor))
```
drewj-usnctech, line 191: This change only appeared for `Composites` that were not directly attached to the core, but
more acted as siblings to the `Core`. For example, if the spent fuel pool composite was
a child of `Reactor`, and the spent fuel pool contained assemblies which in turn contained blocks,
various HDF serialization methods would error out because Block.core -> None.

One example:
_addHomogenizedNumberDensityParams -> collectBlockNumberDensities ->  getSymmetryFactor
calls self.r.core.spatialGrid.overlapsWhichSymmetryLine

The docstring for Block.r indicates that most calls to self.r are to obtain the core
and may be better off just calling Block.core. But, here exists a case where a Block
may not have a Core as the parent (e.g., ex-core structures that physically reside
in the reactor) but still needs to access, for some reason, the spatial grid.

A deeper issue is that this call stack is trying to find the spatial grid for a block
assuming the block is on the Core.spatialGrid. Not sure how to square that. Maybe
the call for symmetry line tries to find a parent with a spatialGrid?

armi/plugins.py
```diff
@@ -120,13 +120,16 @@
...
+from typing import Dict, Union, Callable, Optional, TYPE_CHECKING
 
 import pluggy
 
 from armi import pluginManager
 from armi.utils import flags
 
+if TYPE_CHECKING:
+    from armi.reactor.composites import Composite
+
```
drewj-usnctech, line 132: Not sure the stance on type hinting but seems positive (#531  / #493). This block is only to help with an IDE showing a type that exists in the function signature, e.g.,
```python
def func(a: "Composite"):
```
does not error if `Composite` is imported in this way. But does if you remove the quotes

drewj-usnctech:
@john-science @jakehader @ntouran any interest in this feature?

armi/plugins.py
john-science: We are definitely pro-type hinting. For sure.

I'm not thrilled about extra import and `TYPE_CHECKING`.  But perhaps this is the way of the future.  I haven't decided about these details yet.

Still, yes, generally, we like type hinting.

john-science:
Sorry for not looking at this a long time ago.

I've been busy, and just ignore PRs with the red "X" next to them, but I see that's a minor code coverage thing here.

john-science:
@ntouran Do you have an opinion?

armi/reactor/blocks.py
john-science: Perhaps this could use a one-line comment explaining why?

john-science: @ntouran  I can't tell, will this be confusing to users?  They could ask for `b.r` and get something that is not the `Reactor`.

armi/reactor/blueprints/reactorBlueprint.py
```diff
@@ -74,25 +74,29 @@ def __init__(self, name=None, gridName=None, origin=None):
...
+                # define a new core builder before finding those defined by the
+                # ReactorPlugin
+                if key == typ:
+                    return builder
+                seen.add(key)
+
+        raise ValueError(
+            "Could not determine an appropriate class for handling a "
+            "system of type `{}`. Supported types are {}.".format(typ, sorted(seen))
+        )
```
john-science, line 99: @drewj-usnctech This is a non-trivial change.  And while it _looks_ fine to me, it is a bit hard for me to tell if there are edge cases where this would fail. What would help me understand/grok the changes if if we had a unit test to cover this new/re-written method. Does that make sense?

The obvious place would seem to be `armi/reactor/blueprints/tests/test_reactorBlueprints.py`.

Side Note: Again, I'm sorry it took so long for me to get to this PR.  I wasn't ignoring you, I've just been very busy and when I saw the little red "X" next to your PR I thought it wasn't ready yet.  So my triage procedure was faulty. My  bad.

drewj-usnctech: @john-science you're totally fine on the delay. and yes, this is a non-trivial change where a unit test would help. I'm not sure how to handle a unit test where we would add something to the plugin manager returned by `armi.getPluginMnagerOrFail`. The testable way might be to make this rely on another method that, given any plugin manager, performs this lookup. Then we don't have to touch `armi.getPluginManagerOrFail`

Possible path: refactor the call signature to accept a string `typ` and the plugin manager, then test that "private" method? How do you feel about that?

armi/plugins.py
drewj-usnctech: Excellent. And by

> I'm not thrilled about extra import

Do you mean the import here of `from armi.reactor.composites import Composite`? If it's a performance issue you're considering, that import will not trigger in a production run (maybe the cause for a coverage drop) because of how python and type checkers like IDEs sets the `TYPE_CHECKING` variable

> A special constant that is assumed to be True by 3rd party static type checkers. It is False at runtime.

So then this import isn't tripped in e.g., `armi run`

https://docs.python.org/3.7/library/typing.html#typing.TYPE_CHECKING

armi/reactor/blocks.py
drewj-usnctech: > They could ask for b.r and get something that is not the Reactor.

That actually would have been the previous case if you had spun up a Reactor that contained a `Core` and some sibling `Composite` that also contained `Block` s and `Assembly` s. Because this sibling isn't the `Core`, blocks assigned have a non-None `self.core` attribute from
https://github.com/terrapower/armi/blob/a3f23ed01538aaaeb1dc58b5bf158def918901ef/armi/reactor/blocks.py#L165-L170

So here, for the blocks in the `Core`-adjacent `Composite`, their parent tree looks like `Block` -> `Assembly` -> `Sibling` -> `Reactor`. Then, if we did `b.core` we get `None,` which makes `b.r` return `None` without this change.

armi/plugins.py
john-science: Yeah, I was just thinking about performance. 

To be clear, your typing code here is fine. I am just tenuously unhappy with the general idea of having more imports in Python modules, just for the purposes of typing. I think the Python team's design decision seems likely to generally cause more imports and performance hits, to provide typing info. Which... is just insanely backwards.

Again, your code here is fine. I'm just not sure the languages typing solution is a good one.

drewj-usnctech:
@john-science @ntouran thank you for the feedback. I added some docs in a place that I think makes sense, but let me know if there's a better way to present this information.

In adding the docs, I found some possible shortcomings related to the `sfp` system documentation. The [Systems blueprint documentation](https://github.com/terrapower/armi/blob/master/doc/user/inputs/blueprints.rst#systems) has content like
```yaml
systems:
    core:
        grid name: core
        origin:
            x: 0.0
            y: 10.1
            z: 1.1
    sfp:
        grid name: sfp
        origin:
            x: 1000.0
            y: 12.1
            z: 1.1
```
which should communicate to the user that a `SpentFuelPool` object should be created and attached to the `sfp` grid. But, because the `type` field is missing, another `Core` is built instead.

Demoing this with
```diff
diff --git a/armi/reactor/blueprints/tests/test_reactorBlueprints.py b/armi/reactor/blueprints/tests/test_reactorBlueprints.py
index 6d47f5a..39f002f 100644
--- a/armi/reactor/blueprints/tests/test_reactorBlueprints.py
+++ b/armi/reactor/blueprints/tests/test_reactorBlueprints.py
@@ -19,6 +19,7 @@ import os
 from armi.reactor import blueprints
 from armi import settings
 from armi.reactor import reactors
+from armi.reactor.assemblyLists import SpentFuelPool
 from armi.reactor.blueprints import reactorBlueprint
 from armi.reactor.blueprints import gridBlueprint
 from armi.reactor.blueprints.tests import test_customIsotopics
@@ -104,6 +105,7 @@ class TestReactorBlueprints(unittest.TestCase):
         core, sfp = self._setupReactor()
         self.assertEqual(len(core), 2)
         self.assertEqual(len(sfp), 4)
+        self.assertIsInstance(sfp, SpentFuelPool)
 
     def test_materialDataSummary(self):
         """Test that the material data summary for the core is valid as a printout to the stdout."""
```
yields an assertion error
```
  File "/Users/ajohnson/codes/armi/armi/reactor/blueprints/tests/test_reactorBlueprints.py", line 108, in test_construct
    self.assertIsInstance(sfp, SpentFuelPool)
  File "/opt/anaconda3/envs/raptor/lib/python3.7/unittest/case.py", line 1276, in assertIsInstance
    self.fail(self._formatMessage(msg, standardMsg))
  File "/opt/anaconda3/envs/raptor/lib/python3.7/unittest/case.py", line 693, in fail
    raise self.failureException(msg)
AssertionError: <Core: sfp id:140658844990032> is not an instance of <class 'armi.reactor.assemblyLists.SpentFuelPool'>
```

drewj-usnctech:
@john-science ah that's on me. I thought, because the upstream main branch had been updated with the click changes, I wouldn't need anything on my end. But I've merged in main and it _should_ work now 🤞

drewj-usnctech:
@john-science potentially very minor coverage drop. How do you want to proceed?

john-science:
> @john-science potentially very minor coverage drop. How do you want to proceed?

No worries. It's a 4th decimal-place drop.  I'm not bothered.

john-science:
> @john-science @ntouran thank you for the feedback. I added some docs in a place that I think makes sense, but let me know if there's a better way to present this information.
> 
> In adding the docs, I found some possible shortcomings related to the `sfp` system documentation. The [Systems blueprint documentation](https://github.com/terrapower/armi/blob/master/doc/user/inputs/blueprints.rst#systems) has content like
> 
> ```yaml
> systems:
>     core:
>         grid name: core
>         origin:
>             x: 0.0
>             y: 10.1
>             z: 1.1
>     sfp:
>         grid name: sfp
>         origin:
>             x: 1000.0
>             y: 12.1
>             z: 1.1
> ```
> 
> which should communicate to the user that a `SpentFuelPool` object should be created and attached to the `sfp` grid. But, because the `type` field is missing, another `Core` is built instead.
> 
> Demoing this with
> 
> ```diff
> diff --git a/armi/reactor/blueprints/tests/test_reactorBlueprints.py b/armi/reactor/blueprints/tests/test_reactorBlueprints.py
> index 6d47f5a..39f002f 100644
> +++ b/armi/reactor/blueprints/tests/test_reactorBlueprints.py
> @@ -104,6 +105,7 @@ class TestReactorBlueprints(unittest.TestCase):
>          core, sfp = self._setupReactor()
>          self.assertEqual(len(core), 2)
>          self.assertEqual(len(sfp), 4)
> +        self.assertIsInstance(sfp, SpentFuelPool)
>  
> ```
> 
> yields an assertion error
> 
> ```
> Traceback (most recent call last):
>   File "/Users/ajohnson/codes/armi/armi/reactor/blueprints/tests/test_reactorBlueprints.py", line 108, in test_construct
>     self.assertIsInstance(sfp, SpentFuelPool)
>   File "/opt/anaconda3/envs/raptor/lib/python3.7/unittest/case.py", line 1276, in assertIsInstance
>     self.fail(self._formatMessage(msg, standardMsg))
>   File "/opt/anaconda3/envs/raptor/lib/python3.7/unittest/case.py", line 693, in fail
>     raise self.failureException(msg)
> AssertionError: <Core: sfp id:140658844990032> is not an instance of <class 'armi.reactor.assemblyLists.SpentFuelPool'>
> ```

So, this is the last question in this PR?  Thoughts? @ntouran ?

PR #573, opened 2022-02-14
Bug fix for setting nuclide categories on the `HexToRZThetaConverter`

jakehader:
## Description

Add mapping from the original core's nuclide categorization to 
the convert core within the `HexToRZThetaConverter`. This is
important for lattice physics temperatures and was a bug.

Update the `HexToRZThetaConverter` type strings to leverage
the Flags system.

---

## Checklist

**The following are not applicable for this change**


armi/reactor/converters/geometryConverters.py
```diff
@@ -560,6 +564,15 @@ def _setupConvertedReactor(self, grid):
...
         self.convReactor.core.name += " - {0}".format(self._GEOMETRY_TYPE)
 
+        # It is important to map the nuclide categories from the source
+        # reactor so that the fuel, coolant, and structures are mapped to
+        # the same nuclides. This comes into play in areas of the framework
+        # (like in lattice physics), where the categorization makes an impact
+        # for physics modeling.
+        self.convReactor.core._nuclideCategories = (
+            self._sourceReactor.core._nuclideCategories
+        )
```
ntouran, line 574: The fact that it's listed as a `_private` variable is a bit broken with this external usage. Given this, I guess it probably shouldn't be private, or it should be accessed via a setter.

armi/reactor/converters/geometryConverters.py
```diff
@@ -744,7 +757,7 @@ def _createRadialThetaZone(
 
             # Set new homogenized block parameters
             material = materials.material.Material()
-            material.name = "mixture"
+            material.name = "mixture "
```
ntouran, line 760: Add comment explaining why this has a space.

armi/reactor/converters/geometryConverters.py
```diff
@@ -760,7 +773,7 @@ def _createRadialThetaZone(
             for nuc in self._sourceReactor.blueprints.allNuclidesInProblem:
                 material.setMassFrac(nuc, 0.0)
             newComponent = components.DifferentialRadialSegment(
-                "mixture", material, **dims
+                f"component {newBlockType}", material, **dims
```
ntouran, line 776: did you mean `material.name` here?

armi/reactor/converters/geometryConverters.py
```diff
@@ -892,14 +905,15 @@ def _getHomogenizedBlockType(self, numHexBlockByType):
         # Find the most common block type out of the types in the block mixture type exclusions list
         excludedBlockTypesInBlock = set(
             [
-                x
+                Flags.toString(x)
```
ntouran, line 908: explain in a comment why we're trafficking in strings rather than flags in this section

ntouran (changes requested):
looks good. thanks for adding the test. I have a few requests for adding new comments for clarity.

armi/reactor/converters/geometryConverters.py
jakehader: No, I didn't mean to update the material name. I'll revert. I want the component to inherit the same type as the block so that the flags and type are set based on the block.

armi/reactor/tests/test_flags.py
```diff
@@ -104,6 +104,35 @@ def test_isPickleable(self):
...
+                   "plenum", "duct"]
+        sortedStrings = ["axial shield", "control", "duct", "fuel", "plenum", "radial shield",
+                         "structure"]
+        self.assertEqual(sorted(strings), sortedStrings)
+        
+        print(sorted([flags.Flags.fromString(x) for x in strings]))
+        print([flags.Flags.fromString(x) for x in sortedStrings])
+        self.assertEqual(sorted([flags.Flags.fromString(x) for x in strings]), 
+                         [flags.Flags.fromString(x) for x in sortedStrings])
+
```
jakehader, line 135: WIP - don't comment. Will fix.

armi/reactor/converters/geometryConverters.py
```diff
@@ -789,6 +780,105 @@ def _createRadialThetaZone(
...
+                
+                
+                #### THIS IS WRONG -- I am stumped how to get the right volume fractions for the
+                # new components within the homogenized block so that mass is preserved....
+                if c.getName() == "coolant":
+                    volumeFracsByNuclideCategory["coolant"] += volFracInNewBlock
+                elif "fuel" in c.getName():
+                    volumeFracsByNuclideCategory["fuel"] += volFracInNewBlock 
+                else:
+                    volumeFracsByNuclideCategory["structure"] += volFracInNewBlock
```
jakehader, line 850: @ntouran - I am stumped here. Trying to create components that are based on the categorization of the nuclides so that the (1) temperatures of the nuclides can be preserved from XS generation and (2) so that masses are all conserved when converting between the reactors.

I reused the `getBlockNuclideTemperatureAvgTerms` from the XS group manager above to get the correct temperatures, but I am not sure what to do with calculating the volume fractions.

Basically we have:

newBlock represents the homogenization of bunch of blocks in the core. These other blocks aren't necessarily full blocks because they can get cut off based on the user-defined mesh during the conversion. I believe I am calculating the scaling number densities correctly, but I don't know how to determine which "fraction" of a component belongs to which category of nuclides.

First I tried to scale by number densities rather than using the names of the components, but that didn't work. Some peer programming/insight would be super helpful here.

armi/reactor/tests/test_flags.py
john-science: Any notes on progress here? 

Just curious, I don't mean to rush.

jakehader: I think that Nick just recommended that I implement `__eq__` and get rid of the others. Sorry my comment should have been much better... 

I'm still held up in the core below but haven't had time to work on it. This is an important but fix though to the converter.

jakehader: Also I think I implemented hash on the strings from the flag but I think that flags should be hashable without that.

john-science:
Is this PR still active?

jakehader:
> Is this PR still active?

It's a draft because I haven't had time to get back to it. Still an important bug to fix but I got stuck

armi/reactor/converters/geometryConverters.py
```diff
@@ -789,6 +780,105 @@ def _createRadialThetaZone(
         self.convReactor.core.add(newAssembly)
 
         return outerDiameter
+    
+    def _addNewComponents(self, newBlock, innerDiameter, outerDiameter, 
+                          lowerAxialZ, upperAxialZ, lowerTheta, upperTheta):
+        """
+        Add a set of components to the new homogenized block that was created.
+
```
john-science, line 788: @ntouran I think this method will have to be re-written soon, as Components can now have sub-components.

Thoughts?

jakehader: This class just wouldn't be able to support that. Maybe we need a namespace for classes that will not be that general. Lots of code in the framework assumes a reactor -> core -> assembly -> block -> component -> material relationship and it's going to difficult to generalize or take lots of work to generalize. Not sure how we would like to handle generic composite trees going forward

Issue #575, opened 2022-02-15
Do we have too many YAML libraries?


john-science:
I noticed that in our dependencies we have three different YAML libraries:

* `PyYAML`
* `ruamel`
* `yamlize`

https://github.com/terrapower/armi/blob/db90bc8e27398f840c9e345f0b0e8dd5c7eb69c4/setup.py#L69-L74

Surely we could reduce the number of these, right?

I should note that `ruamel` is a dependency of `yamlize`, so that's where we get that.  In that case, it would seem that `PyYAML` is the odd-one-out, and the easiest to ditch to reduce our dependencies. 

> Can we replace PyYAML with `ruamel.yaml`? Is that convenient?

john-science:
Yes, we can and should remove the `PyYAML` dependency from ARMI.  We are only using it in 3 places (in 2 files) to quickly load up a YAML file into a dictionary:

https://github.com/terrapower/armi/blob/8f52db65ef29d34a7432f339b6a69d05a282d330/armi/nucDirectory/nuclideBases.py#L310-L311

https://github.com/terrapower/armi/blob/8f52db65ef29d34a7432f339b6a69d05a282d330/armi/nucDirectory/nuclideBases.py#L402

https://github.com/terrapower/armi/blob/8f52db65ef29d34a7432f339b6a69d05a282d330/armi/cases/inputModifiers/tests/test_inputModifiers.py#L204

There is nothing complicated, special, or unique about our usage of `PyYAML` that we could 100% drop-in replace with `ruamel.yaml`, which we need anyway as part of `Yamlize`.

> TLDR; we can _easily_ remove one of our three YAML dependencies: PyYAML. And I will l do that.

PR #586, opened 2022-02-28
Remove PyYaml Dependency

john-science:
## Description

PyYaml is a fine library, I have nothing against it. But per #575 , we have THREE different YAML libraries as dependencies in ARMI.  And that seems like too many.

By far, the easiest one of the three to lose is `PyYaml`, as that is only used in two files, for super basic YAML-to-dictionary reading. And one of those files is a unit test.

So, in the interest of reducing our memory foot print, our [dependency overhead](https://xkcd.com/1987/), and making it easier to understand how we interact with our YAML files, I am removing `PyYaml` as an ARMI dependency.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
+++ b/armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
@@ -284,30 +284,20 @@ class LumpedFissionProductCollection(dict):
         return self.keys()
 
     def getAllFissionProductNames(self):
-        """
-        Gets names of all fission products in this collection
-
-        TODO: This can use a set, continually updated from the keys() for each lfp, then converted
-        to a list. This will change the order of the fission products, though, so should be done
-        with care
-        """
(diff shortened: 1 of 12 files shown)


armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -284,30 +284,20 @@ def getLumpedFissionProductNames(self):
...
-        fpNames = []
+        """Gets names of all fission products in this collection"""
+        fpNames = set()
         for lfp in self.values():
             for fp in lfp.keys():
-                if fp not in fpNames:
-                    fpNames.append(fp.name)
-        return fpNames
+                fpNames.add(fp.name)
+        return sorted(fpNames)
```
ntouran, line 292: nice call

ntouran (approved):
very nice. love to see the optimizations along the way.

PR #590, opened 2022-03-04
Tracking ARMI Requirements

john-science:
## Description

This PR allows us to track ARMI's requirements. It leverages the `sphinx-needs` extension to our Sphinx documents, to allow us to track requirements. In particular, we can sprinkle links to requirements in unit tests and code docstrings. And `sphinx-needs` will collate that data and build us a unified Requirements document.

A couple things to note:

* I think this `sphinx-needs` solution is probably a good way to go. But when using `.. impl::` or `.. test::` commands in our code and unit tests, we need to try to reserve those (wherever possible) for class and module-level docstrings. This will mean developers run into them less while working in the code base to make small changes.
* As a start, I have chosen 2 high-level Requirements as examples of the tool. One of which has child requirements. If this PR is merged, we will have the technology in place and will spend time over the coming weeks/months formalizing a full list of requirements for ARMI. But I consider that to be a separate work effort than this.

I am attaching an example build of the updated docs:  [html.tar.gz](https://github.com/terrapower/armi/files/8189124/html.tar.gz)

---

## Checklist


armi/reactor/components/basicShapes.py
```diff
@@ -17,8 +17,14 @@
 
 Many reactor components can be described in 2D by circles, hexagons, rectangles, etc. These
 are defined in this subpackage.
-"""
 
+.. impl:: ARMI shall support a reasonable set of basic shapes.
```
ntouran, line 21: This implementation reads like a requirement, which may be kinda confusing. Maybe say "Support a reasonable set of basic shapes"?

armi/reactor/components/basicShapes.py
```diff
@@ -17,8 +17,14 @@
...
+   :id: IMPL_0_2_0
+   :links: REQ_0_2
```
ntouran, line 23: do we want to do numbering like this, or maybe use slightly more meaningful short names? I think names might be easier.

armi/reactor/components/complexShapes.py
```diff
@@ -14,6 +14,13 @@
 
 """
 Components represented by complex shapes, and typically less widely used.
+
+.. impl:: ARMI shall support a reasonable set of basic shapes.
```
ntouran, line 18: Same name as previous implementation. Maybe better to title this one "Add support for Holed hexes, rectangles, squares, and helices" and remove the follow-up comment

armi/reactor/components/component.py
```diff
@@ -180,6 +180,10 @@ class Component(composites.Composite, metaclass=ComponentType):
         Temperature in C to which dimensions were thermally-expanded upon input.
     material : str or material.Material
         The material object that makes up this component and give it its thermo-mechanical properties.
+
+    .. impl:: All components in ARMI can be thermally-expanded using custom curves.
```
ntouran, line 184: not to be too nit picky with these sample tags, but here again I feel like it should say "here's where the thermal expansion is applied" (not actually here though)

armi/reactor/converters/tests/test_uniformMesh.py
```diff
@@ -37,7 +37,8 @@ class TestUniformMeshComponents(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        random.seed(987324987234)  # so it's always the same
+        # TODO: WHY would this be necessary? Are our uniform meshes random???
```
ntouran, line 40: I believe a unit test does (or used to) set up a random mesh and run this for it. This seed allows the random mesh to always be the same random mesh. could also just hard-code the random mesh.

armi/reactor/grids.py
```diff
@@ -621,6 +621,10 @@ class Grid:
 
     * Unit step calculations use dot products and must not be polluted by the bound
       indices. Thus we reduce the size of the unitSteps tuple accordingly.
+
+    .. impl:: ARMI supports a number of structured mesh options.
+       :id: IMPL_0_3_0
+       :links: REQ_0_3
```
ntouran, line 627: I just feel like for the initial sample ones of these we should put them right on the method that does a very specific implementation of a requirement

armi/reactor/grids.py
```diff
@@ -1145,6 +1149,9 @@ class CartesianGrid(Grid):
         Note that ring 1 has four locations, and that the center of the (0, 0)-index
         location is offset from the origin.
 
+    .. impl:: ARMI supports a Cartesian mesh.
+       :id: IMPL_0_3_1
+       :links: REQ_0_3
```
ntouran, line 1154: this one feels perfect!

ntouran (approved):
Lookin' good! I just have a variety of basically off-topic comments about a few implementation tags. I do like how sphinx-needs ties these all together.

armi/reactor/components/basicShapes.py
john-science: Ooooh, yeah. I bet I did that in a few places.  Good call!

armi/reactor/components/basicShapes.py
john-science: We can pick some other name if you liked.

For "REQ_4" I thought "IMPL_4_X" and "TEST_4_X" would be really quick and easy to associate back to their requirements, for humans reading.

But you want the requirements to be short strings, rather than numbers?  We can do that.

armi/reactor/converters/tests/test_uniformMesh.py
john-science: I hadn't actually tried to track this down. I see now this is used in `applyNonUniformHeightDistribution()`.  So I can document that in a comment, yeah. Good call.

armi/reactor/components/basicShapes.py
john-science: Fixed

armi/reactor/components/basicShapes.py
john-science: Okay, take a look. I have changed the requirements to be short strings, instead of numbers.

(Sorry, I'm just used to requirements being numbered. But I don't really have a preference.)

PR #593, opened 2022-03-09
Application permissive configure fix 

jakehader:
## Description

Fix to a bug where an ARMI application could not call `armi.configure(app, permissive=True)`. This error is due to fact that the type check for the application that is passed in would never be equal to the type of the base ARMI app class. This now checks if the application passed in is instead an instance of the ARMI app.

### DIFF
--- a/armi/__init__.py
+++ b/armi/__init__.py
@@ -306,7 +306,7 @@ def configure(app: Optional[apps.App] = None, permissive=False):
     app = app or apps.App()
 
     if _app is not None:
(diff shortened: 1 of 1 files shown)


john-science:
Obviously, I'm okay with this PR.   What is there to complain about, it's only one line.

I just now realized we can't really unit test the `armi.configure()` method. I did not know that. 

Huh. Well, I know that now.

john-science (approved):

Issue #583, opened 2022-02-18
Clarify deprecation of select `HistoryTrackerInterface` methods


keckler:
Within the `HistoryTrackerInterface` class, the `getTimeIndices` method's docstring says that the method is no longer functional:
https://github.com/terrapower/armi/blob/a3d78a9330ec549bf2fe586230dd3a0044aaf66d/armi/bookkeeping/historyTracker.py#L307-L312

I noticed that some other methods in the class call to that method, however, and do not have similar warnings. For instance [`getBOCEOCTimeIndices`](https://github.com/terrapower/armi/blob/a3d78a9330ec549bf2fe586230dd3a0044aaf66d/armi/bookkeeping/historyTracker.py#L350) and [`getAssemParamHistory`](https://github.com/terrapower/armi/blob/a3d78a9330ec549bf2fe586230dd3a0044aaf66d/armi/bookkeeping/historyTracker.py#L370).

The status of these methods should probably be clarified. Assuming they are not working as a result of `getTimeIndices` being deprecated, we should make that clear. 

Or alternatively, is there any reason not to get rid of them entirely? I grep'd through the codebase and can't find anywhere that they are used.

jakehader:
Hi @keckler, if I recall I believe the goal was to remove the history tracker going forward in favor of the getHistories implementation on database3, which is pretty quick doesn't blow up the RAM for a case (trading hard drive space in the hdf5 database for storage of assembly objects on the interface instance).  

@john-science / @ntouran, I am curious as well where we are at with removing the history tracker and if there are still fundamental things that need to be implemented in database3 so that it covers the use cases.

john-science:
> This is no longer functional

Sigh.  No, I hadn't seen that.  The best policy is certainly to remove all broken code from the repo ASAP.

I note that it is only used in two places in ARMI:

In `getBOCEOCTimeIndices`:
https://github.com/terrapower/armi/blob/a3f23ed01538aaaeb1dc58b5bf158def918901ef/armi/bookkeeping/historyTracker.py#L350-L352

And `getAssemParamHistory`:
https://github.com/terrapower/armi/blob/a3f23ed01538aaaeb1dc58b5bf158def918901ef/armi/bookkeeping/historyTracker.py#L384

Neither of those two methods are used elsewhere in ARMI. They are not covered in unit tests. And I can not find anyone in the ARMI ecosystem who is using them.

So. I wonder if any of the three of these methods would even be missed if they were removed entirely:

[...]

PR #594, opened 2022-03-10
Removing defunct code from historyTracker Interface

john-science:
## Description

This is in response to #583 

It appears there are three entirely unused methods in the `HistoryTrackerInterface`. Notably, the `getTimeIndices()` method states in its docstring that it is defunct and/or broken. So, hopefully it is removable. Also, it is only used in two other places in ARMI, both in the same class: `getBOCEOCTimeIndices()`, and `getAssemParamHistory()`.  I note also that none of these three methods have any unit tests either.  

They seem ripe for removal, so I am testing that idea here.

---

## Checklist


keckler (approved):
Brute force, I like it!

PR #595, opened 2022-03-11
Add validation to settings using schema

keckler:
## Description

Add validation to many of the framework settings using the schema parameter and voluptuous. Mainly validating ranges and sizes of user inputs.

---

## Checklist


john-science (approved):
I love that you filled out the ranges.

I'm not sure that using "DB" over "database" was really a problem before. But I don't have strong opinions.

I liked that you filled out some missing labels and descriptions.

keckler:
This is in response to #595 .

keckler:
Hm... The [test_kineticsParameterAssignment](https://github.com/terrapower/armi/blob/b5ddc61fb74f4cd97d3e9de4d087402b1eaf57e8/armi/physics/neutronics/tests/test_neutronicsPlugin.py#L91) test is failing. I see why. Lemme change.

john-science:
Well, that's a funny little hiccup. But I think we can just do the type casting. For instance, consider this:

```python
import voluptuous as vol
x = vol.All(lambda v: int(v), vol.Range(min=3, max=7))

x(5)  # returns 5
x(-5)  # fails
x("5")  # returns 5
x("-5")  # fails
```

Alternately, if you really want to be careful:

```python
import voluptuous as vol

def str2int(v):
    try:
        return int(v)
    except:
        return vol.Invalid("Cannot cast value to int: {0}".format(v))

x = vol.All(str2int, vol.Range(min=3, max=7))

x(5)  # returns 5
x(-5)  # fails in a very "voluptuous" way
x("5")  # returns 5
x("-5")  # fails in a very "voluptuous" way
```

Does that help?

jakehader:
> Well, that's a funny little hiccup. But I think we can just do the type casting. For instance, consider this:
> 
> ```python
> import voluptuous as vol
> x = vol.All(lambda v: int(v), vol.Range(min=3, max=7))
> 
> x(5)  # returns 5
> x(-5)  # fails
> x("5")  # returns 5
> x("-5")  # fails
> ```
> 
> Alternately, if you really want to be careful:
> 
> ```python
> import voluptuous as vol
> 
> def str2int(v):
>     try:
>         return int(v)
>     except:
>         return vol.Invalid("Cannot cast value to int: {0}".format(v))
> 
> x = vol.All(str2int, vol.Range(min=3, max=7))
> 
> x(5)  # returns 5
> x(-5)  # fails in a very "voluptuous" way
> x("5")  # returns 5
> x("-5")  # fails in a very "voluptuous" way
> ```
> 
> Does that help?

Hey @john-science, you could also use the Coerce option for voluptuous rather than using a lambda function. That's what I've done in the past and it should work the same.

john-science:
@jakehader  I haven't seen the [vol.Coerce](https://www.programcreek.com/python/example/104567/voluptuous.Coerce) before, that's another good pathway.

I haven't actually seen people use `voluptuous` before, it's all new to me.

But it sounds like we have a couple of pathways forward. Awesome.

armi/settings/fwSettings/globalSettings.py
```diff
@@ -267,14 +272,18 @@ def defineSettings() -> List[setting.Setting]:
...
             "will be n+1 time nodes and the burnup step time will be computed as cycle "
             "length/n.",
+            schema=vol.Range(min=0),
         ),
         setting.Setting(
             CONF_BETA,
             default=None,
-            label="Delayed neutron fraction",
-            description="Set individual precursor group delayed neutron fractions.",
+            label="Delayed Neutron Fraction",
```
jakehader, line 280: We use setting the beta and lambda values to zero to determine if we should use the values directly or expect them to be calculated. I would recommend allowing zero values still.

armi/settings/fwSettings/globalSettings.py
```diff
@@ -284,9 +293,12 @@ def defineSettings() -> List[setting.Setting]:
             CONF_DECAY_CONSTANTS,
             default=None,
             label="Decay Constants",
-            description="Set individual precursor group delayed neutron decay constants.",
+            description="Individual precursor group delayed neutron decay constants",
             schema=vol.Any(
-                [float], None, float, msg="Expected NoneType, float, or list of floats."
+                [vol.Range(min=0, min_included=False)],
```
jakehader, line 298: Same comment above.

jakehader (changes requested):

jakehader (approved):

jakehader:
@keckler, the changes look good to me, but there are still some failing unit tests that should be addressed before merging in.

keckler:
Yes, I know, a few things still need to be addressed here. I was waiting before for the XML stuff to be removed from the framework, and then I got distracted. This is still on my radar though!

armi/settings/fwSettings/globalSettings.py
keckler: Done.

armi/settings/fwSettings/globalSettings.py
keckler: Done.

keckler:
Alright, well even though I wanted to figure a way around #669, I didn't. So the schemas look kinda gross sometimes, but whatever.

keckler:
@jakehader @john-science 
_Finally_, I think this PR is ready. Sorry for taking so long on it!

john-science:
I will merge this tonight: we had a couple of other merges today, and I just want to give some time for all of the downstream repos to run their CI separately.

Thanks!

keckler:
Yeah, thanks for the suggestions both of you. I haven't given up on this,
just gotten a little busy with other things. Will come back to it.

Issue #591, opened 2022-03-08
Request: Allow `UnshapedComponent`s to thermally-expand


keckler:
Right now we have the `UnshapedComponent` which allows a user to just enter the volume/area fraction of a component without having to worry explicitly what shape the component is:
https://github.com/terrapower/armi/blob/6dcedacdea505a508cf85119b4eb22953a1763e0/armi/reactor/components/__init__.py#L116-L121

This is nice, but right now these components are not allowed to thermally-expand:
https://github.com/terrapower/armi/blob/6dcedacdea505a508cf85119b4eb22953a1763e0/armi/reactor/components/__init__.py#L150-L159

(Note that the method signature and docstring imply that thermal expansion is possible, but the actual implementation contradicts this)

It would be nice to allow such components to thermally-expand. Though we don't know specific parameters of the shape like radius, pitch, or side-length, we could still get the thermally-expanded area of the component just by using the thermal expansion coefficient directly. Since the areal thermal expansion is just twice the linear thermal expansion, and all material classes will have the linear expansion coefficient, it seems like we should have all that is needed to calculate the thermally-expanded area of `UnshapedComponents`.

This would allow for modeling thermal expansion effects in models when only the volume fractions are known (common in fast reactor benchmark specifications).

john-science:
> Note that the method signature and docstring imply that thermal expansion is possible

Where does it say that?

I see that it says "Compute the area with as-input dimensions instead of thermally-expanded".

Logically, it doesn't seem possible to thermally expand something you don't know the geometry of.  For instance: if you don't know the height, how are you supposed to know how much to expand the height?

If something doesn't have known dimensions, how would you expand those dimensions?

You know?

keckler:
> Where does it say that?

The method has an optional parameter `cold` that is said to allow you to compute the area with "as-input dimensions instead of thermally-expanded". But `cold` is never used. So it implies that you can either get the area for the cold or hot component, but that's not the case. It will always be cold.

keckler:
> Logically, it doesn't seem possible to thermally expand something you don't know the geometry of. For instance: if you don't know the height, how are you supposed to know how much to expand the height?

> If something doesn't have known dimensions, how would you expand those dimensions?

But we have the area. You can expand the area itself without knowing the dimensions that lead to that area.

https://courses.lumenlearning.com/boundless-physics/chapter/thermal-expansion/#:~:text=area%20thermal%20expansion%20coefficient

john-science:
[...]

PR #600, opened 2022-03-15
Unshaped component

keckler:
## Description

This addresses #591 by allowing for `UnshapedComponent`s to thermally expand. Tests have been added to ensure that the thermal expansion is consistent with the changes in density of the material with temperature. Other existing tests have been modified to prove that the new functionality does not break any existing features.

Also, a bug in the `getBoundingCircleOuterDiameter` method was fixed.

---

## Checklist


armi/reactor/tests/test_components.py
```diff
@@ -189,14 +199,59 @@ def test_getDimension(self):
...
+                material=self.componentMaterial,
+                Tinput=self.component.inputTemperatureInC,
+                Thot=self.component.inputTemperatureInC,
+                area=math.pi,
+            )
+        )
+        coldDensity = coldComponent.density()
+        coldArea = coldComponent.getArea()
+
+        self.assertNotEqual(thermalExpansionFactor, 1)
```
john-science, line 236: Would this perhaps be more clear if it said `self.assertGreater(thermalExpansionFactor, 1)`?

john-science (approved):
I appreciate the lift.

See my one comment on the unit test.  And change it if you like.

Otherwise, let me wait until after work hours today to merge this?  Thanks.

armi/reactor/tests/test_components.py
keckler: Sure, I could see that. I'll change.

keckler:
No problem on any delay. Thanks for taking a look!

armi/reactor/components/__init__.py
```diff
@@ -118,9 +118,13 @@ class UnshapedComponent(Component):
     A component with undefined dimensions.
 
     Useful for situations where you just want to enter the area directly.
-    For instance, in filler situations where the exact shape of this component is
-    is unknown but you have some left-over space between other components filled
-    with a known material you might need to model.
+
+    For instance, when you want to model neutronic behavior of an assembly based
+    on only knowing the area fractions of each material in the assembly.
+
```
albeanth, line 124: I'd consider even taking this a step further:
1. the sum of the areal fractions of unshaped components need to be equal with the total area of the block. (but maybe there are other places this could/should go)
2. expanding on the neutronics argument that this type is useful for homogenization-like studies when particular shapes/locations of components is not important (since they end up getting smeared together anyway) but rather knowing the area (and ultimately the volume) of a material within a block and how it can change with temperature.

Just a thought.

john-science: Sorry, I've already merged the PR.  But @keckler , do you have any thoughts here?

PR #612, opened 2022-04-01
Removing support for XML Settings files

john-science:
## Description

This PR removes support for the XML CaseSettings files.  From here on out we will only support the YAML format.

This is meant as a simplification step. You will notice that much code can be removed when dropping this support; and much of the code that is left is simpler.

As part of this "clean out the old stuff we don't use any more" I hope to soon drop support for XML blueprint files, and the old geomFiles.

I think ARMI could be simpler and easier to understand. Dropping complicated support for old, defunct features is an easy win toward this end. For instance, our support for XML input/data file is holding back a PR at this very moment:

https://github.com/terrapower/armi/pull/595#issuecomment-1065575012

Also, I think I have seen a lot of activity lately where XML input files and the old geomFiles have made our current work much slower than it needs to be.

---

## Checklist


keckler:
Very happy about this. I will take a detailed look later today, hopefully.

jakehader (approved):
Great cleanup! Looks like the depreciation warning was there for a long time so that should've given folks time to migrate as needed.

armi/settings/tests/test_settingsIO.py
```diff
@@ -45,15 +45,15 @@ def test_loadFromXmlFailsOnBadNames(self):
         with self.assertRaises(TypeError):
             ss.loadFromInputFile(None)
         with self.assertRaises(IOError):
-            ss.loadFromInputFile("this-settings-file-does-not-exist.xml")
+            ss.loadFromInputFile("this-settings-file-does-not-exist.yaml")
```
keckler, line 48: I notice that even though you changed the file extension here, the name of the test that this falls within still refers to XML. I think either that name should be changed, or maybe this entire test can be removed?

keckler (approved):

PR #613, opened 2022-04-03
Update control rod parameters and generalize the neutronics mesh converter to copy all multi-group results.

jakehader:
## Description

Adding a few control rod parameters and renaming others to support more general control rod plugin(s) for the framework. This also removes the `getBlockLengthAboveAndBelowHeight` because it _was_ used by
the internal reactivity control plugin and now provides minimal value on the Assembly class.

Finally, the uniform mesh converter was improved on to copy all neutronics parameters back to the original core state compared to only copying back flux and mgFlux. This now uses the parameter system's `multiGroupQuantities` and `detailedAxialExpansion` categories on the block-level and the `neutronics` category on the reactor core level.

---

## Checklist


jakehader:
Not sure why black is failing. Maybe there is a new release and something is broken?

jakehader:
> > Not sure why black is failing. Maybe there is a new release and something is broken?
> 
> Are you using the right version of black?
> 
> https://github.com/terrapower/armi/blob/4fcea827c16b471e067fff809f775caec54e2de8/setup.py#L91
> 
> It sounds to me like you might just be using the latest version of `black`. Possibly.

I can check but this venv I'm on should have consistent requirements with the framework.

armi/reactor/reactorParameters.py
```diff
@@ -168,27 +168,46 @@ def detailedNucKeys(self, value):
...
+            description=(
+                "Core assembly location for the most valuable secondary control rod."
+            ),
         )
-
         pb.defParam(
             "crWorthRequiredPrimary",
-            units="$",
+            units="pcm",
             description="The total worth in $ required for primary control rods to shutdown reactor accounting for uncertainties and margins",
```
ntouran: description says $ which is now inconsistent with units. recommend just removing the `in __` clause to avoid possiblity.

armi/reactor/reactorParameters.py
```diff
@@ -168,27 +168,46 @@ def detailedNucKeys(self, value):
...
         )
-
         pb.defParam(
             "crWorthRequiredSecondary",
-            units="$",
+            units="pcm",
             description="The total worth in $ required for secondary control rods to shutdown reactor accounting for uncertainties and margins",
```
ntouran: remove units clause from description

ntouran (commented):
looks good to me. just minor comments

john-science:
> I have the correct version of `black` and `click` installed. Is the `click` package in the setup.py being installed?

Okay, there is a problem in the `black` / `clink` dependency we're using. They have some bug. I fixed this in the `master` branch by doing:

https://github.com/terrapower/armi/blob/907f2b49d4063cfd07068431698d22dab4c7d267/.github/workflows/black.yaml#L16

That one line would fix your branch too.

armi/reactor/converters/uniformMesh.py
```diff
@@ -244,13 +236,33 @@ def _buildAllUniformAssemblies(self):
...
+            assemsToPlot.append(coreAssems[0])
+
+        # Obtain the plot numbering based on the existing files so that existing plots
+        # are not overwritten.
+        start = 0
+        existingFiles = glob.glob(
+            f"{self.convReactor.core.name}AssemblyTypes" + "*" + ".png"
+        )
+        for f in existingFiles:
+            newStart = int(re.search(r"\d+", f).group())
```
john-science, line 257: Sorry, I can't tell...  What is the regex here for?

At the very least, we need a good comment describing what this regex is doing, because if I can't tell, neither can others.

armi/reactor/converters/uniformMesh.py
```diff
@@ -244,13 +236,33 @@ def _buildAllUniformAssemblies(self):
             src = sourceAssem.spatialLocator
             newLoc = self.convReactor.core.spatialGrid[src.i, src.j, 0]
             self.convReactor.core.add(newAssem, newLoc)
+        self.plotConvertedReactor()
 
     def plotConvertedReactor(self):
```
john-science, line 241: @ntouran  What do you think of having a plotting method in a place like this?

I think it seems strangely out of place, and means we need to import matplotlib when we are just importing a reactor or converter.  I don't know... should we move all plotting tools to armi/utils/plotting.py, or similar?  Or am I crazy?

john-science:
If you can add a comment explaining that regex, I think we're good here.

jakehader:
Still working on some stuff but will update the regex soon.

armi/reactor/converters/uniformMesh.py
jakehader: This is just for debugging now and testing.

jakehader:
https://github.com/terrapower/armi/pull/645 should be merged in first.

jakehader:
@john-science / @ntouran - This is ready for last reviews now!

armi/reactor/converters/uniformMesh.py
john-science: Hmmm... I don't like things that are used for testing cluttering up our code though. This may have to get moved out of the class.

armi/reactor/converters/uniformMesh.py
```diff
@@ -52,11 +52,17 @@
...
 import copy
-import logging
+import collections
+import multiprocessing
+from timeit import default_timer as timer
 
 import numpy
 
+import armi
+from armi.mpiActions import Multiprocessor
```
john-science, line 65: This appears to be imported, but not used.

Or am I missing something?

john-science (approved):

armi/reactor/converters/uniformMesh.py
john-science: I think this still needs to be removed, right?

PR #614, opened 2022-04-05
Add rotational updates to growToFullCore()

alexhjames:
- spatial variables are rotated when the assemblies are copied

## Description

---

## Checklist

### DIFF
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -1317,6 +1317,16 @@ class Assembly(composites.Composite):
         """Return the symmetry factor of this assembly."""
         return self[0].getSymmetryFactor()
 
+    def rotate(self, deg):
+        # note this can only rotate in 60 degree increments.
+        # as this is only called by growToFullCore which is strictly hex, the 60 degree restriction is acceptable 
+        rotNum = round((deg%(2*math.pi))/math.radians(60))
+        for b in self.getBlocks():
+            b.rotatePins(rotNum)
+            b.p["THcornTemp"] = b.p["THcornTemp"][-rotNum:] + b.p["THcornTemp"][:-rotNum]
+            b.p["THedgeTemp"] = b.p["THedgeTemp"][-rotNum:] + b.p["THedgeTemp"][:-rotNum]
+
+
 
 class HexAssembly(Assembly):
     pass
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -1323,12 +1323,38 @@ class Assembly(composites.Composite):
         rotNum = round((deg % (2 * math.pi)) / math.radians(60))
         for b in self.getBlocks():
             b.rotatePins(rotNum)
-            b.p["THcornTemp"] = (
-                b.p["THcornTemp"][-rotNum:] + b.p["THcornTemp"][:-rotNum]
-            )
-            b.p["THedgeTemp"] = (
-                b.p["THedgeTemp"][-rotNum:] + b.p["THedgeTemp"][:-rotNum]
-            )
+            for param in b.p.paramDefs.atLocation(ParamLocation.CORNERS).names:
+                if isinstance(b.p[param], list):
+                    if len(b.p[param]) == 6:
+                        b.p[param] = b.p[param][-rotNum:] + b.p[param][:-rotNum]
+                    else:
+                        runLog.warning(
+                            "No rotation method defined for spatial parameters that aren't defined once per hex edge/corner. No rotation performed on {}".format(
+                                param
+                            )
+                        )
+                else:
+                    # this is a scalar and there shouldn't be any rotation.
+                    pass
+            for param in b.p.paramDefs.atLocation(ParamLocation.EDGES).names:
+                if isinstance(b.p[param], list):
+                    if len(b.p[param]) == 6:
+                        b.p[param] = b.p[param][-rotNum:] + b.p[param][:-rotNum]
+                    else:
+                        runLog.warning(
+                            "No rotation method defined for spatial parameters that aren't defined once per hex edge/corner. No rotation performed on {}".format(
+                                param
+                            )
+                        )
+                else:
+                    # this is a scalar and there shouldn't be any rotation.
+                    pass
+
+            dispx = b.p.get("displacementX")
+            dispy = b.p.get("displacementY")
+            if (dispx is not None) and (dispy is not None):
+                b.p["displacementX"] = dispx * math.cos(deg) - dispy * math.sin(deg)
+                b.p["displacementY"] = dispx * math.sin(deg) + dispy * math.cos(deg)
 
 
 class HexAssembly(Assembly):
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -1318,35 +1318,16 @@ class Assembly(composites.Composite):
         return self[0].getSymmetryFactor()
 
     def rotate(self, deg):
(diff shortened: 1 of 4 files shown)


jakehader (changes requested):
Hi @alexhjames, thank you for the contribution! We have a notion of object orientation within the framework and I am curious if there is some overlap. Also, can you provide maybe a problem statement of what is wrong, why it's wrong, and in what conditions? For example, it seems that the impression here is that the hex side numbering would be different for the assemblies symmetric cases, but I'm having trouble visualizing it. Also, if there are other parameters that have this condition we can leverage the parameter system to detect params that need rotations so we don't miss any.

alexhjames:
@jakehader, I poked around in the assembly object and the only orientation related thing I could find was the rotate pins. Could you point me towards what you were thinking of? 

This fix is making it so that an spatial parameters get rotated when the core is grown to a full core. I noticed this error in the oxbow/armi interface. Oxbow took a third symmetric core, applied corner temperatures, grew it to full core and then filtered it back down to Oxbow's third symmetric core model. The only difference, is that for assemblies that lie on the symmetry plane, oxbow uses the top left assemblies while armi uses the bottom right assemblies. The corner temperatures on the top left assemblies were the same orientation as the bottom right inputs to growToFullCore(). They should have been rotated 120 degrees in the growing process. 

I included the two things I knew of in corner temperature and edge temperature. It also looked like the pins should be rotated as well. If there is other information that should be rotated as well, we should add it.

john-science:
@alexhjames   Thanks for the new unit test!

I love that we are unit testing the function directly.  Thanks for that.  The unit test seems clear and concise.  Sometimes I say "don't forget to write tests for your edge cases", but I'm not sure what your edge cases would be here.

(It looks like you forgot to run `black` on your unit test. But, again, `black` is just a small thing.)

alexhjames:
@jakehader I modified the rotate function to use the parameter framework. Anything that is array defined at the corners or edges will be rotated. Also the displacement X and displacement Y are handled separately as they need to interact with each other.  A new ticket #619 requests updates to how scalar vs vector quantities are handled. That ticket would result in some modifications to this function when implemented.

john-science:
The new changes to the `rotate()` method introduce some duplicate code:

```python
if isinstance(b.p[param], list):
      if len(b.p[param]) == 6:
          b.p[param] = b.p[param][-rotNum:] + b.p[param][:-rotNum]
      else:
          runLog.warning(
              "No rotation method... No rotation performed on {}".format(param)
        )
else:
    # this is a scalar and there shouldn't be any rotation.
    pass
```

This is repeated twice.  Perhaps instead of this:

```python
for param in b.p.paramDefs.atLocation(ParamLocation.CORNERS).names:
    # block
for param in b.p.paramDefs.atLocation(ParamLocation.EDGES).names:
    # block
```

Perhaps that block can be a helper method?  Or maybe something like this:

```python
params = b.p.paramDefs.atLocation(ParamLocation.CORNERS).names
params += list(b.p.paramDefs.atLocation(ParamLocation.EDGES).names)
for param in params:
    # block
```

I'm just spit-balling here. But there's a bug chunk of duplicate code, and I'm always a fan of reducing repetition.

jasonbmeng:
Hi @jakehader @ntouran , are there any further changes requested for this review or anything left to discuss?

This PR and its eventual inclusion in the Nala framework is important to the mechanical team's current modeling efforts for spatially varying field data (temperature, dose, etc.)

armi/reactor/assemblies.py
```diff
@@ -1317,6 +1317,37 @@ def getSymmetryFactor(self):
...
+                            "No rotation method defined for spatial parameters that aren't defined once per hex edge/corner. No rotation performed on {}".format(
+                                param
+                            )
+                        )
+                else:
+                    # this is a scalar and there shouldn't be any rotation.
+                    pass
+
+            dispx = b.p.get("displacementX")
+            dispy = b.p.get("displacementY")
```
ntouran, line 1346: should just use the normal `b.p.displacementX` syntax so IDEs can track the variables.

armi/reactor/assemblies.py
```diff
@@ -1317,6 +1317,37 @@ def getSymmetryFactor(self):
         """Return the symmetry factor of this assembly."""
         return self[0].getSymmetryFactor()
 
+    def rotate(self, deg):
+        # note this can only rotate in 60 degree increments.
```
ntouran, line 1321: this public method needs a valid docstring with a 1 line summary and then more details describing what it does and how.

ntouran (changes requested):
So I think this is an enhancement on top of an existing capability which is in blocks.py as the method `rotatePins`. Can we move the crux of this method into blocks.py in that area and have this function at this level just loop over the blocks of an assembly and call b.rotatePins?

In any case, the docstrings of this method and the rotatePins method should refer to each other.

armi/reactor/assemblies.py
alexhjames: I have updated the other calls to use b.p.displacementX. In this particular case, I left it as is because I specifically used the .get() function in case the parameter does not exist. The get function will return None which will be caught by the next if statement. I think if I were to access by the .displacementX and it didn't exist, I would get an error?

armi/reactor/assemblies.py
alexhjames: I have added docstrings for all the functions added.

alexhjames:
@ntouran @jasonbmeng I agree with Jason that this does not belong in the `rotatePins` method as it has nothing to do with pins. It is rotating duct corner and edge parameters. I have refactored it so that the `rotate` method in assemblies loops over all the blocks and rotates each one. In blocks, there is a abstract function not implemented for rotation. The hex block class then has it's specific rotate functionality which is the code copied over from assemblies. If other block types in the future have a need for rotation, they can gain a rotate method rather easily.

ntouran:
Thanks for that update. The one concern I didn't express clearly enough is that we need to either combine this method with `rotatePins` so that there is only one, **or** we need to clearly define in the docstrings of both `rotate` and `rotatePins` how the method differs from the other and when to use one vs. the other. 

As is they seem to be doing roughly the same thing. Can we easily combine them both into a single method?

ntouran (changes requested):
Thanks for that update. The one concern I didn't express clearly enough is that we need to either combine this method with `rotatePins` so that there is only one, **or** we need to clearly define in the docstrings of both `rotate` and `rotatePins` how the method differs from the other and when to use one vs. the other. 

As is they seem to be doing roughly the same thing. Can we easily combine them both into a single method?

alexhjames:
@ntouran @john-science how is `rotatePins` currently used? As it exists prior to this PR, someone is using it for something. I'm hesitant to touch that function without knowing more about it. I think the `rotate` docstring is pretty clear that it rotates spatially varying quantities on a block. The `rotatePins` docstring could use some additional clarification but that goes back to the question of what that function is used for currently.

ntouran (approved):
beautiful. thank you!

john-science (approved):

Nebbychadnezzar:
Is there any larger discussion on how to deal with rotations that don't align with EDGES or CORNERS? For example, if we wanted to assign values to a generic set of points (e.g., pins, cladding, etc.), what is the best way to do this?

alexhjames:
@Nebbychadnezzar I have not seen any further discussion on that topic. I would think that any of those rotation types could be added to the current rotate functionality. For example, the rotate function added in this PR rotates the edges/corner information but also has a call to the rotatePins() method.

PR #615, opened 2022-04-05
Fixing assembly number issue in DB Load

john-science:
## Description

It appears there has been a bug in our DB `load` for several years now. Essentially, when reading from certain (though not all) H5 databases, the assembly number can start off in a bad state. This is not immediately obvious, but if you try to do something serious with the resultant `Reactor`/`Core` objects, you immediately find the problem. In this case, trying to use `growToFullCore()` immediately after the `load()` raises the problem.

**NOTE**: Relatedly, it turns out `Block.makeUnique()` was unused. So I removed it. Thanks @ntouran !

---

## Checklist


john-science:
@[jasonbmeng](https://github.com/jasonbmeng) @[jeffbaylor](https://github.com/jeffbaylor)

Just FYI, the PR to fix `growToFullCore()` is up. It turns out to be a problem in the database `load()`.

ntouran (approved):

PR #618, opened 2022-04-05
Split fuelHandlers.py into multiple scripts

HunterPSmith:
## Description

Reconfigured fuelHandlers.py into multiple scripts. Updated references to objects and methods as appropriate.

---

## Checklist

- [ ] The code is understandable and maintainable to people beyond the author.
- [ ] There is no commented out code in this PR.
- [ ] The commit message follows [good practices](https://terrapower.github.io/armi/developer/tooling.html).


john-science:
Heya!

Can I ask why we're splitting this stuff up?

1. You will need to sign the CLA before GitHub will let your PR spin up.
2. It looks like you have a Merge conflict.
3. We are trying to not use "Object" as part of the names for things in Python: `armi/physics/fuelCycle/fuelHandlerObject.py`. Essentially, everything in Python is an "Object", so you sometimes end up seeing codebases where every other variable or file has the word "object" in the name. So that doesn't help.

Thanks for contributing!

john-science:
This PR changes the ARMI API.  And on balance, the change seems fine.

But the question here is the same as for any open-source project: 

> When changing the API of an open-source project, you will break a lot of downstream projects. Is the API change useful enough to warrant all the effort put upon people who are supporting downstream projects?

For instance, if the only reason to change the API here is that it seems nicer, than people who make ARMI plugins may not want to go through the effort to support this change.  But if the API change here gives us some new utility, and improves the functionality of ARMI in some way, people who support ARMI plugins and applications probably won't mind altering their code to meet the new API.

john-science:
Let's say you move two classes (`FuelHandlerInterface` and `FuelHandlerFactory`) out of `fuelHandlers.py`.

But, instead of stopping there, at the top of `fuelHandlers.py` you put this:

```python
from armi.whatever.fuelHandlerInterface import fuelHandlerInterface
from armi.whatever.fuelHandlerFactory import fuelHandlerFactory
```

If you did THAT then you would get what you want (cleaner code), but it would be backwards compatible.  And there would be a 0% chance of breaking someone's plugin.

HunterPSmith:
For additional context, this work is being done to support implementation of the requirements for the fuel handling capabilities of ARMI drafted in the Fuel Module Management SRSD.

jakehader:
Due to CLA issues, this is being closed now and the code is now in https://github.com/terrapower/armi/pull/620. There were issues with the wrong email address on the `git config --global user.email`.

PR #625, opened 2022-04-15
Adding MPI tests to code coverage

john-science:
## Description

This PR adds the MPI tests into the code coverage.  I'll be honest, I thought this would net us like 5% code coverage, but it was more like 0.5%.  BUT it's done now, so fine.

This is only a change to the build system, not to the code.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,6 +28,7 @@ commands =
     git submodule update
     make html
 
+# First, run code coverage over the unit tests that run MPI library code.
 [testenv:cov1]
 deps=
     pip>=20.2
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -28,13 +28,28 @@ commands =
     git submodule update
     make html
 
-[testenv:cov]
+[testenv:cov1]
```
ntouran, line 31: Can you add comments here describing the separation between cov1 and cov2?

ntouran (changes requested):

tox.ini
john-science: Yeah, that's a fair point.  Those names aren't that great either.  Good call.

john-science: Done.

ntouran (approved):

PR #626, opened 2022-04-15
Removing bare "import armi" statements

john-science:
## Description

While using namespace is great, having hundreds of "import project" statements at the top of every module in a Python "project" is just bad design.

In this PR I went though and removed nearly every bare `import armi` statement in our repo. In the process I found several circular import loops. This PR is just the necessary first step in a top-level clean up. This is just cleanup, but I found it quite fruitful.

---

## Checklist


ntouran (approved):
beautiful! Thanks for doing that.

PR #627, opened 2022-04-19
Update docstring formatting in `getPotentialParentFromSettingValue`

keckler:
## Description

The docstring was being rendered all crazy:
image

Turns out that the ordering of different sections in the docstrings matters, so moved the remainder of the description out of the "Parameters" block.

---

## Checklist


keckler:
I have no idea why `black` is failing here, but I don't think it was anything I did.

armi/cases/case.py
```diff
@@ -226,6 +226,10 @@ def getPotentialParentFromSettingValue(self, settingValue, filePattern):
         """
         Get a parent case based on a setting value and a pattern.
 
+        This is a convenient way for a plugin to express a dependency. It uses the
+        ``match.groupdict`` functionality to pull the directory and case name out of a
+        specific setting value an regular expression.
+
```
john-science, line 232: I am like 95% sure `black` is failing because there is extra whitespace on this line.

keckler: Ah, you're right. Fixed!

john-science (approved):

PR #628, opened 2022-04-19
Removing inherited docstrings from Docs

john-science:
## Description

It has been noted that our auto-docs use super class docstrings if the class doesn't have a docstring:

https://github.com/terrapower/armi/discussions/423

That's no good, so this PR fixes that.  

**Caveat**: I was not able to fix this problem for `yamlize` classes for some reason. For instance, [radialMeshPoints](https://terrapower.github.io/releases/armi/v0.2.3/.apidocs/armi.reactor.blueprints.assemblyBlueprint.html#armi.reactor.blueprints.assemblyBlueprint.AssemblyBlueprint.radialMeshPoints) are a subclass of `yamlize.Attributes`.  And the only solution I have found for this is I can completely remove the `radialMeshPoints` member from the `AssemblyBlueprint` doc entirely. But I have not found a way to remove the inherited docs from `yamlize` subclasses.

---

## Checklist

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -135,6 +135,7 @@ autodoc_default_options = {
     "private-members": False,
 }
 autodoc_member_order = "bysource"
+# this line removes huge numbers of false and misleading, inherited docstrings
 autodoc_inherit_docstrings = False
 autoclass_content = "both"
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -135,6 +135,7 @@ autodoc_default_options = {
     "private-members": False,
 }
 autodoc_member_order = "bysource"
+autodoc_inherit_docstrings = False
 autoclass_content = "both"
 
 apidoc_module_dir = SOURCE_DIR


john-science:
Question Time:

Which should I do?

1. Just remove all subclasses of `yamlize.Attribute` from the autodocs?
2. Open a ticket for solving this inheritance problem for `yamlize` subclases?

ntouran (approved):
I think you should add a short comment to this explaining why we turned that off. Git blame may be useful but it's probably worth a comment so someone else doesn't just turn it back on. 

Also, I think opening a ticket on yamlize would be useful, as I expect other uses of the lib will run into a similar issue and they're more likely to find it there than here.

john-science:
Okay, the problem is generally fixed in the docs except for subclasses of `yamlize.Attribute`, which I have opened a ticket for here:

https://github.com/terrapower/armi/issues/629

PR #632, opened 2022-04-20
Update DoseResultsMapper.apply() to accept blockList

Nebbychadnezzar:
`apply()` is a convenience method that makes a call to `updateFluenceAndDpa()` in the class `DoseResultsMapper`.
However, it does not allow for passing in a `blockList` parameter, which is a parameter that `updateFluenceAndDpa()` accepts. This PR updates `apply()` to accept a `blockList` parameter that is then passed to `updateFluenceAndDpa()`.

This is a trivial change, but it's my first pull request ever, so I am partly doing this to get some exercise on the process. The motivation for the PR is for the use of this method in the flux reconstruction plugin. The only other plugin that makes use of `apply()` is depletion, and I have confirmed that this change does not alter the behavior of that plugin.

Please be gentle.

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -632,7 +632,17 @@ class DoseResultsMapper(GlobalFluxResultMapper):
     """
     Updates fluence and dpa when time shifts.
 
-    Often called after a depletion step.
+    Often called after a depletion step. It is invoked using :py:meth:`apply() <.DoseResultsMapper.apply>`.
+
+    Parameters
+    ----------
+    depletionSeconds: float, required
+        Length of depletion step in units of seconds
+
+    options: GlobalFluxOptions, required
+        Object describing options used by the global flux solver. A few attributes from
+        this object are used to run the methods in DoseResultsMapper. An example
+        attribute is aclpDoseLimit.
 
     Notes
     -----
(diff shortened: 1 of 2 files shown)


armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -632,7 +632,17 @@ class DoseResultsMapper(GlobalFluxResultMapper):
...
+
+    Parameters
+    ----------
+    depletionSeconds: float, required
+        Length of depletion step in units of seconds
+
+    options: GlobalFluxOptions, required
+        Object describing options used by the global flux solver. A few attributes from
+        this object are used to run the methods in DoseResultsMapper. An example
+        attribute is aclpDoseLimit.
```
john-science, line 645: While I LOVE improving our docs, I'm not sure this is right.

I think this "Parameters" thing only applies to the input arguments of a function/method, right?  And classes don't have parameters.

So, where are these Params coming from?

john-science:
I think the change looks solid. Though I put up a question about your comment.

I love that you used RST-style comments and changed a unit test!

armi/physics/neutronics/globalFlux/globalFluxInterface.py
Nebbychadnezzar: Yay, I'm a big boy now!

I have not seen a consistent approach on this in ARMI, so I should have asked first, sorry. My understanding is that there are different conventions for where to put the arguments to the constructor (`__init__()`) for a class. My previous experience involved putting them in the docstring for the class. Two examples of this would be in `armi.nuclearDataIO.cccc.cccc.IORecord` and `armi.reactor.grids.Grid`. However, I don't think there is a universally-acknowledged Pythonic place to put them, nor probably in ARMI?

john-science: Hmmm... perhaps that's a fine place to put the `__init__()` constructor documentation.

It just threw me for a loop a bit because I couldn't actually _see_ the constructor for that class.

john-science (approved):

PR #636, opened 2022-04-22
Directory changers skip retrieval of missing files with a user warning.

jakehader:
## Description

This updates the directory changers behavior to allow skipping retrieval of files that do not exist rather than failing with an operating system error. This now presents itself as a warning and allows the other files that _do_ exist to be copied. This is useful in instances when running a physics code may not return all the "expected" files based on other user settings. 

One specific example of this is that the `ASUMMAR` file may not exist in the running directory when the summary section is not populated due to other user settings. Rather than introducing more complex logic into the file retrieval system within the the physics plugin, this makes the directory changer less rigid. In most cases the `ASUMMAR` in this example is not required. There are likely other cases where this will pop up.

---

## Checklist


jakehader:
I have the correct black requirements installed locally, so I don't understand why `black` is still failing when I create a new PR.

john-science:
> I have the correct black requirements installed locally, so I don't understand why `black` is still failing when I create a new PR.

The branch you are using is too old, and does not match the master branch.

You need to do the following to make this PR work:

```
git checkout master
git pull origin master
git checkout warnCopyFiles
git merge master
# do merge
git push origin warnCopyFiles
```

john-science:
> How do you see this? Also the change shouldn't need to be at the latest master of there aren't merge conflicts, right? Not sure I understand how that affects black running...

I checked our your branch on my laptop. I created a new venv and ran `black` and it failed. That's all.

Having said that, the only part of this that's strange to me is that GITHUB appears to be running unit tests/black/coverage from your branch BEFORE merging with master, and not AFTER.  And that's odd.  I mean, I looked it up, and that's apparently "working as intended", it's just not what I would have expected, you know?

jakehader:
> > How do you see this? Also the change shouldn't need to be at the latest master of there aren't merge conflicts, right? Not sure I understand how that affects black running...
> 
> I checked our your branch on my laptop. I created a new venv and ran `black` and it failed. That's all.
> 
> Having said that, the only part of this that's strange to me is that GITHUB appears to be running unit tests/black/coverage from your branch BEFORE merging with master, and not AFTER. And that's odd. I mean, I looked it up, and that's apparently "working as intended", it's just not what I would have expected, you know?

Sure... But what is that click error in getting? How is that related?

john-science (approved):

Issue #576, opened 2022-02-15
Extending burnup and cycle inputs


keckler:
Building off of the discussion in #561, a number of upgrades/extensions to the current cycle input are planned. This ticket will keep track of the various pieces.

## Purpose
Right now the structure of a burnup case is defined by a few case settings: `nCycles`, `power`, `powerFraction(s)`, `burnSteps`, `availabilityFactor(s)`, and `cycleLength(s)`. These settings allows one to define a case with multiple cycles, where each cycle can have variable length, power, and availability. However it has a number of drawbacks that are not ideal for modeling some realistic systems:
1. Power within the cycle cannot vary, so if a user wants to change the power level they must define an entirely new cycle
2. The length of each burnup step is fixed to be uniform within each cycle, so capturing some important physics like Xenon buildup can be difficult

This ticket will primarily address these two shortcomings, in addition to a handful of associated enhancements.
## Needs
- [x] Allow for variable numbers of burn steps within each cycle
- [x] Allow for non-uniform burn step lengths within each cycle
- [x] Allow for power to change at each burn step within each cycle

## Would-be-nice-to-have's
- [x] Allow burn step lengths to be input as either cumulatives or deltas
- [ ] Allow powers to be input as either fractions or absolutes
- [x] Allow for cycles to have names
- [ ] Allow for the decay period currently defined by `availabilityFactor(s)` to contain multiple steps
- [x] Continue to allow the use of repeated lists as is currently allowed within vectors like `availabilityFactors` and `cycleLengths`
- [ ] Only trigger a specific subset of interfaces on zero-power cycles (i.e. decay only)

## Method
These capabilities will be enabled for the user through the use of new yaml-style input options to be implemented in the case settings file. This input should notionally look something like the following:

```
cycles:
  - name: "Cycle 1 (initial power tests)"
    cumulative days: [1, 5, 10, 100]
    powerFractions: [0.1, 0.5, 1, 1]
  - name: "Cycle 2.a" 
    cumulative days: [1, 5, 6, 15, 50]
    powerFractions: [0.1, 0.5, 0.7, 1.0, 0.9]
  - name: "Cycle 2.b" 
    cumulative days: [50]
    powerFractions: [1.0]
```

These changes should maintain backwards compatibility by retaining the current cycle input in addition to the new method. The user will be able to use either format, and the code will resolve whichever input format that the user decides to use into the same data internal structure. 

## Other considerations
[...]

PR #637, opened 2022-04-22
Upgrade the cycles inputs

keckler:
## Description

Upgrading the cycles case settings in response to #576 .

This PR makes the following major changes:
* Adds the `cycles` case setting, which allows a user to define the cycles history in more granular detail using nested YAML entries
* Adds documentation on the different ways to define the cycles history
* Adds methods for resolving and getting all aspects of the cycles history on the operator
* Propagates the use of the new getter methods
* Maintains backwards compatibility with the use of the "old" cycle history settings (i.e. cycleLengths, burnSteps, etc.)
* Enforces that the "old" cycle history settings cannot be directly grabbed out of the case settings if the `cycles` setting is also entered, which will prevent errors in accidentally grabbing conflicting information during a run
* Updates all tests to be compatible with the changes
* Adds tests for the new functionalities
* Adds some minor helper methods to the `mathematics` module

---

## Checklist


keckler:
Note, the ability to give names to individual cycles has been added, though nothing is being done with these names as of yet.

keckler:
@ntouran @jakehader @john-science 

I think this PR is ready for a first-pass review. A few things to note:

The changes in this PR have the possibility to break downstream plugins and existing user inputs in a few cases:
1. The API on a number of utility functions has changed to make them more clear as to what they are actually doing. See [getTimeStepNum](https://github.com/terrapower/armi/pull/637/commits/233c85f59cb9bc4f3cf3700a3d1a5c0d2fa208c4), [getCycleNode](https://github.com/terrapower/armi/pull/637/commits/384f26a4b7059ae5fca35bb05d0e0f53fa4625bf), [getPreviousTimeNode](https://github.com/terrapower/armi/pull/637/files#diff-42e67339a650cb2bd1978ea248284c1441d9328deb7af26587886cb486c12e15L92). These functions were essentially unused, and the limited places that they were used have been updated.
2. The case settings now require that, if the simple cycles inputs are to be used, all of them must be entered. i.e. cases which don't use the new cycles input format and which don't have all of `burnSteps`, `cycleLength(s)`, `availabilityFactor(s)` will be prompted by the settings validator to fix it. This means that the default values previously assumed are no longer blindly applied, but the settings validator will prompt the user to use those same defaults if any of them are missing. (sidenote, this is why I also submitted #651)
3. `stepLength` is a new parameter now stored on the reactor parameters. This means that database comparisons using `compareDB3` will complain that a new parameter is present between new and old database files.
4. [`prepRestartRun` no longer needs the last two arguments](https://github.com/terrapower/armi/pull/637/files#diff-00e5ad260418536d73065a8e941a3961784e1a1cae7925ee2b2ff0a170846bbdL307) because they were redundant and never being used anyways.

But finally, the biggest change is that downstream plugins should no longer directly reach into the case settings to get things like `cs["burnSteps"]` or `cs["availabilityFactors"]`, because it can longer be assumed that these values will exist! If these values _do_ exist in the case settings, they can be grabbed, but it is advised that, going forward, all cycle information be pulled from any of the following places: 
1. Directly from the operator via `o.cycleLengths`, `o.availabilityFactors`, etc.
2. Directly from the current reactor parameters like `r.p.cycleLength`, `r.p.stepLength`, etc.
3. Or by passing the case setting object to the utility functions `utils.getCycleLengths(cs)`, `utils.getBurnSteps(cs)`, etc.

Where necessary, the framework has been updated to reflect these changes. Same for all internal downstream plugins.

New tests have been added to confirm the new capabilities, and breaking tests that already existed have been fixed. 

To test that everything is still working in harmony, I ran some full multi-cycle cases using our (updated) internal applications with the following differences:
1. Cycle history defined using the simple cycles option (i.e. the way everyone has been specifying cycle histories up until now)
2. Cycle history defined using the detailed cycles option, entered in such a way as to have the same exact cycle history
3. Cycle history defined using the detailed cycles option and having non-uniform step lengths for various cycles

In all cases, the results match as would be reasonably expected, for time nodes which overlap (obviously results for time nodes which don't overlap do not match, but they are confirmed to be as expected).

So, I think that this PR is reasonably complete. Please let me know if you have any questions at all.

jakehader:
Thanks @keckler! We will make it a priority to get this in. Appreciate the improvement!

keckler:
Also, I might recommend first reading the updated documentation to frame the rest of your review:
https://github.com/terrapower/armi/pull/637/files#diff-ba684d52873b132e511eec8df7c08255c91c99aaa7bd50fbd7bae8da42826b23

john-science:
Thanks so much, Chris!  I'm looking at it, but this is a big PR, so it'll be a minute.

This was surprising:

```python
    setting.Setting(
            CONF_CYCLE_LENGTH,
            default=365.242199,   # This becomes None
```

lt's not immediately apparent why that change was made.  BUT, it's a big PR.  Let me look at it.

keckler:
@john-science I agree that it is not immediately apparent why I removed those defaults. And that is probably the most controversial thing that I did in the PR. The reason for that is the following:

I want downstream applications to avoid reaching into the case settings to get things like `cycleLength`. Going forward, cycle history information should be obtained from the operator. But ultimately I cannot _prevent_ them from doing so, since they will have access to the case settings throughout the run. So if an application reaches into the case settings and pulls `cycleLength` and doesn't realize that the user is actually using the new detailed `cycles` case setting to set that information, the application would blindly grab the default `cycleLength`, which could be totally wrong. 

And because the default setting is set right away when the case settings are resolved, it is impossible for the framework to know if the user set that value or if it was just the default (because obviously it is possible that the user set their case setting to the same value as the default). 

So if I want to guard against this type of mistake, then I need a way to check that the user has actually entered the value that is being grabbed. And to do that, I thought the best way was to change the relevant defaults to `None`. Then, if an application reaches for `cycleLength` and gets `None`, it is clear that they are trying to get that value from the wrong place.

Understanding that this change is likely somewhat undesirable, I did go ahead and implement a new set of settings validations that sort through the relevant case settings and detect if some settings that would have otherwise had a default value previously are not entered (namely `cycleLength`, `burnSteps`, and `availabilityFactor`). If this setting validator is triggered, it will offer to the user the possibility to automatically implement the default values. So in effect, the default values are still there, just kinda stored in a different way.

Does that make sense? It's a little convoluted, and I struggled with a way to handle the problem, but ultimately this was the best option that I came up with. Open to suggestions, though.

john-science:
@keckler 

Quick Question:

```
The case settings now require that, if the simple cycles inputs are to be used, all of them must be entered. i.e. cases which don't use the new cycles input format and which don't have all of burnSteps, cycleLength(s), availabilityFactor(s) will be prompted by the settings validator to fix it.
```

Wait, are you saying people _can't_ run it the old way, exactly as-is? People who have old input files can't run them exactly as they used to, with the new code?

keckler:
@john-science 
I am saying that if they had an old input that didn't include all of `cycleLength(s)`, `burnSteps`, and `availabilityFactor(s)`, and they were relying on the defaults for those values, then that user will now have to respond "Yes" when the settings validator asks them if they want to use the defaults for those values.

The run will not break, it will just verify from the user if they actually want to use the defaults or not.

To me that feels no different than if we were to migrate any other setting. But if it is a problem, then we'll need to figure out another way around the problem that I outlined above.

john-science:
> The run will not break, it will just verify from the user if they actually want to use the defaults or not.

And for cluster runs, the runs will proceed as normal if no one is there to press a button, right?  Is that the default functionality for the SettingsValidator during automated runs?

armi/physics/neutronics/fissionProductModel/tests/test_fissionProductModel.py
```diff
@@ -29,7 +29,7 @@ class TestFissionProductModel(unittest.TestCase):
     """
 
     def setUp(self):
-        o = buildOperatorOfEmptyHexBlocks()
+        o = buildOperatorOfEmptyHexBlocks({"cycleLength": 365, "burnSteps": 4})
```
john-science, line 32: Is there a reason we are passing this helper function inputs, rather than just changing the helper function?

armi/utils/mathematics.py
```diff
@@ -290,6 +291,39 @@ def getStepsFromValues(values, prevValue=0.0):
...
+    relation : {'increaseOrEqual', 'increaseExclusive', 'decreaseOrEqual', 'decreaseExclusive'}
+        The relation between the elements to check.
+
+    Returns
+    -------
+    bool
+    """
+    try:
+        op = {
+            "increaseOrEqual": operator.le,
```
john-science, line 313: I _think_ you only need two of these.

For instance, if you are testing "less than or equal to", you don't need to also test "strictly less then".

I _think_ you can replace these for with `>=` and `<=`.

armi/utils/mathematics.py
```diff
@@ -290,6 +291,39 @@ def getStepsFromValues(values, prevValue=0.0):
     return steps
 
 
+def isMonotonic(inputList, relation):
```
john-science, line 294: Minor things:

1. I believe `inputList` is a bit mis-named, since this method works for any iterable. I might do `inputIter` or something.
2. I am not super thrilled with the long names for the inputs: "increaseOrEqual", etc.  Why not just ">=", as that is a lot faster to write, and more standard in Python. Or, Hell, just let people pass in the actual function `>`?
3. Very Minor: But I think I would pull out your dictionary mapping strings to operators, before the `try` loop. You are mixing data (your dictionary) with logic (the `try`/`accept`).

Sorry!

armi/tests/armiRun.yaml
```diff
@@ -51,6 +53,7 @@ settings:
   percentNaReduction: 10.0
   power: 100000000.0
   refType: inner igniter fuel
+  reloadDBName: reloadingDB.h5
```
john-science, line 56: Interesting. Why did this get added here?

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -107,6 +107,7 @@ def test_interaction(self):
         Check that a 1000 pcm rx swing is observed due to the mock.
         """
         cs = settings.Settings()
+        cs = cs.modified(newSettings={"burnSteps": 4})
```
john-science, line 110: This makes it seem like maybe we need better defaults.  Or am I missing something?

doc/user/inputs/settings.rst
```diff
@@ -132,3 +132,97 @@ These settings come with a few caveats:
...
+
+    3. Constant full-power operation for 30 days split into six even increments
+
+    4. Constant full-power operation for 90 days, split into two equal-length 45 day segments, followed by 10 days of downtime
+
+As can be seen, the detailed cycle history option provides much greated flexibility for simulating realistic operations, particularly power ramps or scenarios that call for unevenly spaced burnup nodes, such as xenon buildup in the early period of thermal reactor operations.
+
+.. note:: Although the detailed cycle history option allows for powers to change within each cycle, it should be noted that the power over each step is still considered to be constant.
+
+.. note:: The detailed cycle history may not be used for equilibrium calculations at this time.
```
john-science, line 224: Can you show me where this breaks down?

Thanks!

armi/bookkeeping/db/tests/test_database3.py
```diff
@@ -128,6 +128,71 @@ def _compareRoundTrip(self, data):
         roundTrip = database3.unpackSpecialData(packed, attrs, "testing")
         self._compareArrays(data, roundTrip)
 
+    def test_prepRestartRun(self):
+        # This test is based on the armiRun.yaml case that is loaded during the `setUp`
+        # above. In that cs, `reloadDBName` is set to 'reloadingDB.h5', `startCycle` = 1,
+        # and `startNode` = 2. The nonexistent 'reloadingDB.h5' must first be
+        # created here for this test.
```
john-science, line 135: Can we just make this a multi-line docstring using `"""`?

Thanks!

armi/operators/tests/test_operators.py
```diff
@@ -76,5 +78,79 @@ def test_addInterfaceSubclassCollision(self):
...
+    def test_getCycleNames(self):
+        self.assertEqual(self.detailedOperator.cycleNames, self.cycleNamesSolution)
+
+    def test_getAvailabilityFactors(self):
+        self.assertEqual(
+            self.detailedOperator.availabilityFactors,
+            self.availabilityFactorsSolution,
+        )
+
+    def test_getStepLengths(self):
```
john-science, line 142: I love these tests!

I was wondering if we could add a line to all of these "getter" tests.  For instance, here, if you set:

```python
self.detailedOperator.stepLengths = None
```

And then you ran the same `self.assertEqual()` line again, you would expect to get the same result, right?

But, if you add these two extra lines we make sure to cover more of the code under the "getter" method you wrote by unit tests.

That's all, I'm just looking at the code coverage of all these getters you wrote, and since you start with a fully complete operator, your tests don't flex the `if not self._stepLengths:` parts of these methods.

armi/settings/fwSettings/globalSettings.py
```diff
@@ -209,64 +211,67 @@ def defineSettings() -> List[setting.Setting]:
...
             "setting.",
             oldNames=[
                 ("capacityFactor", None),
             ],
+            schema=(vol.Any(float, int, None)),
         ),
         setting.Setting(
             CONF_AVAILABILITY_FACTORS,
-            default=[],
+            default=None,
```
john-science, line 248: This default has the wrong type. It _looks_ invalid.

Perhaps you meant to just remove the default value?

keckler:
Alright, I actually found that removing those defaults is breaking something downstream which cannot be fixed without the defaults being there.

I'm gonna give something else a go, so hold off on any further reviews until further notice.

armi/bookkeeping/db/tests/test_database3.py
keckler: Done.

armi/utils/mathematics.py
keckler: Done.

doc/user/inputs/settings.rst
keckler: It doesn't actually break down explicitly in the code. But the issue is more a theoretical one. The equilibrium cycle has only one cycle, so using the detailed `cycles` input where each cycle can have a unique history doesn't make sense. So instead of checking their `cycles` input to make sure that it has only 1 cycle defined, we just say they can't use the `cycles` input in equilibrium cases. 

I had a short discussion with @ntouran about this, and this seemed like the best, least confusing path.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
keckler: Outdated. These types of changes were able to be unwound after the settings defaults were reinstated.

armi/settings/fwSettings/globalSettings.py
keckler: Outdated. This was unwound when the settings defaults were reinstated.

armi/physics/neutronics/fissionProductModel/tests/test_fissionProductModel.py
keckler: Outdated. This was no longer necessary once the setting defaults were reinstated.

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -60,7 +60,7 @@ def setUp(self):
         but none of these have any number densities.
         """
         self.o, self.r = test_reactors.loadTestReactor(
-            self.directoryChanger.destination
+            self.directoryChanger.destination, customSettings={"nCycles": 3}
```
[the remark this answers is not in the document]
keckler: This one was an anomaly. For whatever reason, the rest of this test was set up under the assumption that there would be 3 cycles. But the input that is loaded says 6 cycles.

That input is used for a bunch of other tests, so I don't want to change that.

And the rest of this test is already set up to expect 3 cycles. So I don't want to change the rest of the tests.

I can't remember exactly how it was working before, but I do remember that there were a few places where this type of thing was getting tangled up, that I had to sort out. For whatever reason, it just landed this way that this test needed to be fixed.

But all the other places that previously had similar `customSettings` being passed in have since been removed, so this is the only one left.

armi/tests/armiRun.yaml
keckler: This needed to be added so that a reload file would actually be loaded up in the [`test_prepRestartRun`](https://github.com/terrapower/armi/pull/637/files#diff-aca297e341a545c450119034f3293962547c6d11999bd09426d5504c9b241d4fR131). Before, the test that was in place didn't require actually loading anything because it only checked against the case settings. But that was no longer possible, so an actual restart file needed to be loaded, and thus the case setting pointing to that restart file needed to be added.

armi/bookkeeping/db/database3.py
```diff
@@ -304,28 +305,65 @@ def interactDistributeState(self) -> None:
...
+                            currentCaseStepLengths[cycleIdx][:dbNode]
+                        )
+                    else:
+                        dbStepHistory.append(dbStepLengths[cycleIdx])
+                        currentCaseStepHistory.append(currentCaseStepLengths[cycleIdx])
+            except IndexError:
+                runLog.error(
+                    f"DB cannot be loaded to this time: cycle={dbCycle}, node={dbNode}"
+                )
+                raise
```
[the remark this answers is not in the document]
keckler: Done.

armi/operators/tests/test_operators.py
keckler: Done.

keckler:
Okay, I changed a number of things up so that the defaults are reinstated back to the various simple cycles settings. 

So no longer is the setting validator used to prevent the user from entering both simple and detailed cycles inputs, since that had a number of issues.

What I have done now is to introduce a small check in the `Settings.__getitem__` method that doesn't allow for a user to pull any of the simple cycles settings from directly from the case settings if the detailed `cycles` setting is in use. And I blended this check in with the existing checks that were in the `__getitem__` so hopefully this doesn't look too out of place. And it potentially allows for other checks to be added in the future for similar scenarios.

It also allowed me to roll back a number of the changes that I had in this PR before, so some of the @john-science 's comments are no longer applicable.


Hopefully this resolves any lingering doubts!

doc/user/inputs/settings.rst
john-science: That makes sense, thanks for taking the time to explain it.

Hmm... You know, in `SettingsValidation` we can check the `RunType`. So, we could explicitly check if the `cycles` are valid in the uncommon case where `cs["runType"] == operators.RunTypes.EQUILIBRIUM`.

But just because we _can_ do a thing, doesn't mean we _should_ do a thing.

I can't decide if that's a good idea or not.  Do you have an opinion?

keckler: To be honest, I _do_ think that it would be good to allow for the detailed `cycles` input to be used for equilibrium cases, in particular because it would probably be necessary for thermal-spectrum reactors to converge to an accurate equilibrium cycle.

I didn't allow for it, though, because my initial survey made it appear kinda difficult to make it work in our downstream projects. I can reevaluate that, though.

If we _did_ allow the use of the detailed `cycles` with equilibrium runs, I guess the only requirement that would need to be checked is that the `cycles` entry has only 1 cycle's information in it.

So I'll look into this a little more tomorrow. And if it looks like it is going to appreciably stall getting this change into the framework, then I'll just punt it into a new ticket for future work.

john-science: >  I guess the only requirement that would need to be checked is that the cycles entry has only 1 cycle's information in it.

Yeah, man. That's all I meant.

But it's at your discretion either way.

armi/operators/operator.py
```diff
@@ -65,16 +73,28 @@ class Operator:  # pylint: disable=too-many-public-methods
     cs : CaseSettings object
             Global settings that define the run.
 
+    cycleNames : list
```
ntouran, line 76: `list of str` may be more helpful

armi/operators/operator.py
```diff
@@ -65,16 +73,28 @@ class Operator:  # pylint: disable=too-many-public-methods
...
+        The name of each cycle. Cycles without a name are `None`.
+
+    stepLengths : list
```
ntouran, line 79: `list of list of float` (see https://numpydoc.readthedocs.io/en/latest/format.html#parameters)

armi/operators/operator.py
```diff
@@ -65,16 +73,28 @@ class Operator:  # pylint: disable=too-many-public-methods
...
+        A two-tiered list, where primary indices correspond to cycle and
+        secondary indices correspond to the length of each intra-cycle step (in days).
+
     cycleLengths : list
```
ntouran, line 83: `list of float`

armi/operators/operator.py
```diff
@@ -65,16 +73,28 @@ class Operator:  # pylint: disable=too-many-public-methods
...
         The duration of each individual cycle in a run (in days). This is the entire cycle,
         from startup to startup and includes outage time.
 
+    burnSteps : list
```
ntouran, line 87: `list of int`

armi/operators/operator.py
```diff
@@ -65,16 +73,28 @@ class Operator:  # pylint: disable=too-many-public-methods
...
+        The number of sub-cycles in each cycle.
+
     availabilityFactors : list
```
ntouran, line 90: shouldn't this be a two-tiered list as well? or are we by definition saying that a cycle has a pre-determined availability that doesn't change? Might be worth adding to the docstring the reason this is not two-tiered

armi/operators/operator.py
```diff
@@ -65,16 +73,28 @@ class Operator:  # pylint: disable=too-many-public-methods
...
         The fraction of time in a cycle that the plant is producing power. Note that capacity factor
         is always less than or equal to this, depending on the power fraction achieved during each cycle.
 
     powerFractions : list
```
ntouran, line 94: `list of list of float`

armi/operators/tests/test_operators.py
```diff
@@ -76,5 +78,105 @@ def test_addInterfaceSubclassCollision(self):
...
+    """
+
+    detailedCyclesSettings = """
+metadata:
+  version: uncontrolled
+settings:
+  power: 1000000000.0
+  nCycles: 3
+  cycles:
+    - name: dog
```
ntouran, line 93: nbd but I kind of prefer like "physics testing" and "initial startup" or something slightly more analogous to what they'll really be called

armi/operators/tests/test_operators.py
```diff
@@ -76,5 +78,105 @@ def test_addInterfaceSubclassCollision(self):
...
+      availability factor: 0.5
+    - name: ferret
+      step days: [3, R4]
+      power fractions: [0.3, R4]
+  runType: Standard
+"""
+
+    powerFractionsSolution = [
+        [0.1, 0.2, 0.3],
+        [0.2, 0.2, 0.2, 0.2, 0],
```
ntouran, line 109: recommend `0.0` for the last one so they're all floats

armi/settings/caseSettings.py
```diff
@@ -135,11 +144,33 @@ def __repr__(self):
             self.__class__.__name__, self.caseTitle, total, altered
         )
 
+    def _settingIsOkayToGrab(self, key):
+        """
+        A way to check if specific settings can be grabbed out of the case settings.
+
+        Could be updated with other specific instances as necessary.
```
ntouran, line 151: "ok to grab" is a bit ambiguous in this context. Maybe refer to some of the utils using :py:func:`armi.utils.blah` linkages or similar?

Maybe `_directAccessOfSettingAllowed` or something?

in general I'm not super happy about this special casing that has to be applied to all settings. It seems like it might be nicer if the settings system was more flexible in being able to allow special derived settings more universally. 

That might be nice to add here in the `Notes` section of the docstring for future work.

armi/settings/caseSettings.py
```diff
@@ -39,6 +39,15 @@
...
 DEP_WARNING = "Deprecation Warning: Settings will not be mutable mid-run: {}"
 
+SIMPLE_CYCLES_INPUTS = [
+    "availabilityFactor",
+    "availabilityFactors",
+    "powerFractions",
+    "burnSteps",
+    "cycleLength",
+    "cycleLengths",
+]
```
ntouran, line 49: change to set for O(1) member testing

ntouran (changes requested):
Great addition of a useful feature. Thanks!. I love the thorough documentation and testing. Just a few minor requests and comments from me.

armi/operators/operator.py
keckler: Done

armi/operators/operator.py
keckler: Done

armi/operators/operator.py
keckler: Done

armi/operators/operator.py
keckler: done

armi/operators/operator.py
keckler: Sure, I see that confusion. 

The assumption is that the availability factor is on a cycle-basis, not on a step-basis. So every cycle has its own availability factor, but that availability factor is applied to the cycle as a whole.

Added some clarification.

armi/operators/operator.py
keckler: done

armi/operators/tests/test_operators.py
keckler: Done

armi/operators/tests/test_operators.py
keckler: Well I actually wanted to test having an int in there, just to make sure that it actually works, since I anticipate users might very likely put an int in their own inputs.

Was there some other reason why you make this recommendation that maybe I'm overlooking?

armi/settings/caseSettings.py
keckler: Good call, fixed!

armi/settings/caseSettings.py
keckler: I very much agree that I don't like this. I tried to avoid doing it for a long time, but ultimately this was the only way I could see to completely prevent other code from grabbing the defaults in things like `cs["burnSteps"]` or `cs["cycleLengths"]` when the cycle information is actually entered by the user in `cycles`. 

On the other hand though, it isn't all that often that things are pulled from the case settings, at least from the perspective of slowing down the overall calculation.

I tried to make it a bit more universal by pulling the key-check into the function, which is being done regardless, but to me it still feels like this doesn't really belong here.

I will change the name of the method and also add the note that you suggested for now.

keckler:
@ntouran 
Thanks for the review! I made the requested changes, except for one. Let me know if any of those responses don't address your concerns.

john-science (approved):

john-science:
Approved.

I am not sure what the consequences will be for the `def _directAccessOfSettingAllowed(self, key):` change. 

That seems somewhat indirect and a little nebulous to me.  So I'm not _certain_ I can say it is safe.

But all the tests pass, so I am convinced for now.

keckler:
Though, in that case, having coerce(float) as a schema wouldn't be
necessary, because that is what the setting will default to if no custom
schema is provided.

armi/settings/fwSettings/globalSettings.py
```diff
@@ -211,62 +213,65 @@ def defineSettings() -> List[setting.Setting]:
...
             default=1.0,
             label="Plant Availability Factor",
             description="Availability factor of the plant. This is the fraction of the "
-            "time that the plant is operating. If variable, use availabilityFactors "
+            "time that the plant is operating. If variable, use `availabilityFactors` "
             "setting.",
             oldNames=[
                 ("capacityFactor", None),
             ],
+            schema=(vol.Any(float, int, None)),
```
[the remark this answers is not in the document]
john-science: That seems like a fine change to me, if you want to open a PR for it.

PR #649, opened 2022-04-30
Fix filepath update during case cloning

opotowsky:
## Description

A file included in the original settings input file was being cloned properly , but the file path was not being updated in the input file its new cloned location. This PR should fix that issue by enforcing the setting label came from a valid setting object and making the new/destination filepath an absolute path to the new file location. 

---

## Checklist


jakehader:
Thanks @opotowsky! Please make sure to sign the CLA agreement.

jakehader (changes requested):
It's probably a good idea to add tests to the copyInterfaceInputs so that we can demonstrate that the relative paths are transformed into an absolute path for a simple case that has the shuffle logic setting provided. This was a tricky thing to debug.

armi/cases/case.py
```diff
@@ -769,10 +769,9 @@ def copyInterfaceInputs(
         # guess. In future, it might be nice to have interfaces specify which
         # explicitly.
         for key, files in interfaceFileNames.items():
-            if isinstance(key, settings.Setting):
-                label = key.name
-            else:
-                label = key
+            if not isinstance(key, settings.Setting):
+                key = cs.getSetting(key)
```
jakehader, line 773: It might make sense to capture this in a try/except and provide a better exception that will point us to the specifyInputs method(s) providing an incorrect setting name.

opotowsky:
@jakehader : Let me know if the exception and the test are OK. I did not add a test that asserts an exception, but could do that as well.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -69,7 +69,7 @@ def __init__(self, r, cs):
     @staticmethod
     def specifyInputs(cs):
         files = {
-            cs.getSetting(settingName).label: [
+            cs.getSetting(settingName): [
```
john-science, line 72: I believe with this change, you have to go and change the docstring for `Interface.specifyInputs()`:

https://github.com/terrapower/armi/blob/fdc6bab730b9bb06ee78810f8f8d1693bb3ec9ea/armi/interfaces.py#L422-L423

armi/cases/case.py
```diff
@@ -769,10 +769,16 @@ def copyInterfaceInputs(
         # guess. In future, it might be nice to have interfaces specify which
         # explicitly.
         for key, files in interfaceFileNames.items():
-            if isinstance(key, settings.Setting):
-                label = key.name
-            else:
-                label = key
+            try:
+                if not isinstance(key, settings.Setting):
+                    key = cs.getSetting(key)
```
jakehader, line 774: I think that the exception check should be `NonexistentSetting` based on the behavior of `getSetting`. Also, I'd recommend moving the try/except just around the `getSetting` call.

armi/physics/fuelCycle/fuelHandlers.py
jakehader: I agree!

jakehader (changes requested):

armi/physics/fuelCycle/fuelHandlers.py
opotowsky: Noted on the docstring. 

I'm not super familiar with the annotation syntax, but should the return now exclude the string since the key in the dictionary is now only a setting object?

armi/cases/case.py
```diff
@@ -769,10 +770,13 @@ def copyInterfaceInputs(
...
         for key, files in interfaceFileNames.items():
-            if isinstance(key, settings.Setting):
-                label = key.name
-            else:
-                label = key
+
+            if not isinstance(key, settings.Setting):
+                try:
+                    key = cs.getSetting(key)
+                except NonexistentSetting(key):
```
jakehader, line 777: Lets not do a pass here but include that failure that we had before with the Value error instead. This would just be a `raise ValueError("...blah")` instead of pass so that this does actually fail if the specify inputs is incorrect. Once you have this I'm good to merge in.

jakehader (changes requested):

armi/cases/case.py
opotowsky: @jakehader that makes sense! 

Would I need to also include a `runLog` error message or is the failing with `ValueError` enough?

armi/interfaces.py
```diff
@@ -431,11 +431,10 @@ def specifyInputs(cs) -> Dict[Union[str, settings.Setting], List[str]]:
         The files returned by an implementation should be those that one would want
         copied to a target location when cloning a Case or CaseSuite. These can be
         absolute paths, relative paths, or glob patterns that will be interpolated
-        relative to the input directory. Aboslute paths will not be copied anywhere.
+        relative to the input directory. Absolute paths will not be copied anywhere.
 
-        The returned dictionary should be keyed off of a descriptive string, or an
-        actual Setting object. If a Setting is used, then the source CaseSettings object
-        will be updated to the new file location.
+        The returned dictionary should be keyed off of a Setting object, so the source
```
opotowsky, line 436: @jakehader can you confirm this looks good and that I don't need to update the annotation in the definition?

armi/physics/fuelCycle/fuelHandlers.py
jakehader: Yes I agree with you @opotowsky. I also don't know where the list comes from. I think should only be a dict?

armi/cases/case.py
jakehader: Failing is fine because that will emit the error message clearly. RunLog errors can be used in instances where say you are in a loop and want to emit many messages before failing. In this case because the failure is at the same place where you'd add a runLog too it's optional and not really needed in my opinion.

armi/interfaces.py
jakehader: I would add one more sentence that says something along the lines of - While the dictionary keys are recommended to. E setting objects, the name of the setting as a string (e.g., "shuffleLogic") is still interpreted. If the string name does not point to a valid setting then this will lead to a failure.

jakehader (changes requested):
Getting close! Two quick comments left.

armi/physics/fuelCycle/fuelHandlers.py
opotowsky: The dictionary that a given `specifyInputs` method returns can have a key that is either a string or setting object, still (so my question is probably not relevant, because I was thinking it should _only_ be a setting object when I asked the question). The value is enforced as a list because there can be a list of files for some settings, I think.

jakehader (approved):

jakehader:
Running tests one last time after approval and I'll merge in! Thanks @opotowsky!

opotowsky:
Oops I think my update means you need to re-approve the checks again @jakehader (sorry!)

armi/physics/fuelCycle/fuelHandlers.py
john-science: Darn, so we didn't get to this in this PR?

Okay, I will open another PR to fix it.

armi/cases/case.py
```diff
@@ -769,10 +770,17 @@ def copyInterfaceInputs(
         # guess. In future, it might be nice to have interfaces specify which
         # explicitly.
         for key, files in interfaceFileNames.items():
-            if isinstance(key, settings.Setting):
-                label = key.name
-            else:
-                label = key
+
```
keckler, line 773: This change seems to have taken away the ability to copy over files which don't correspond to a specific setting. For instance, one may have auxiliary files that support a calculation but are not called out as a specific case setting. 

It seems that we used to be able to copy such files with `copyInterfaceFiles()` in conjunction with a `specifyInputs()` method on the interface in question. But from the changes here, we can no longer do that I guess.

Is this intentional? Because it broke some of our downstream projects and we'll need a way around, if so.

@john-science @jakehader @opotowsky

keckler: (Also, FYI, this wasn't even my downstream project that was broken, for once! 😆 )

john-science: @opotowsky Do you have bandwidth to fix this one? 

It seems like there's a use-case that we didn't consider in reviewing this PR.  It happens.

Perhaps we could build tests for both use-cases, so this will be clearer in the future?  (You just Know I love tests.)

keckler: I can speak to (2).

Often in our models we have some files that are just a big data dump of things like control rod positions at different times, or fuel shuffling operations for each cycle. This data doesn't readily live anywhere in ARMI, and thus needs to be stored in some "auxiliary" file that is parsed and utilized during the run.

I suppose, in theory, both of the examples I listed just now could be stored in either the custom `controlLogic` file or the `shuffleLogic` file, but it typically doesn't make sense to do that based on the size/format of the data.

PR #653, opened 2022-05-03
Axial expansion update

albeanth:
## Description
A few commits based on updated feedback from @jakehader . Thanks, Jake! 

The only non-docstring related change is in regard to the constructor of the expansion changer object. Instead of passing in a dictionary with relevant settings (which only looked for "detailedAxialExpansion"), they are now just read from the core ARMI settings. This simplifies the use of the changer and expects "detailedAxialExpansion" to be set in the settings input. By default it's set to False. 
 

---

## Checklist

### DIFF
--- a/armi/reactor/converters/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger.py
@@ -33,17 +33,16 @@ class AxialExpansionChanger:
         establishes object to store and access relevant expansion data
     """
 
-    def __init__(self, converterSettings: dict):
+    def __init__(self, detailedAxialExpansion: bool = False):
         """
         Build an axial expansion converter.
 
         Parameters
         ----------
-        converterSettings : dict
-            A set of str, value settings used in mesh conversion. Required
-            settings are implementation specific.
+        detailedAxialExpansion : bool, optional
+            A boolean to indicate whether or not detailedAxialExpansion is to be utilized.
         """
-        self._converterSettings = converterSettings
+        self._detailedAxialExpansion = detailedAxialExpansion
         self.linked = None
         self.expansionData = None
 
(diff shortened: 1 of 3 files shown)


albeanth:
@jakehader @john-science

jakehader (approved):
Thanks @albeanth!

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -94,13 +125,12 @@ def _isTopDummyBlockPresent(self):
...
-                        "Cannot run detailedAxialExpansion without a dummy block"
-                        "at the top of the assembly!"
-                    )
-                    raise RuntimeError
+            if self._detailedAxialExpansion:
+                runLog.error(
+                    "Cannot run detailedAxialExpansion without a dummy block"
+                    "at the top of the assembly!"
+                )
+                raise RuntimeError
```
jakehader, line 133: Add the same message you have as the error into the exception so that the error is clearer when it crashes.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -73,7 +98,13 @@ def setAssembly(self, a, setFuel=True):
         ----------
          a : :py:class:`Assembly <armi.reactor.assemblies.Assembly>` object.
             ARMI assembly to be changed
+        setFuel : boolean, optional
+            Boolean to determine whether or not fuel blocks should have their target components set
+            This is useful when target components within a fuel block need to be determined on-the-fly.
         """
+        self._detailedAxialExpansion = Settings.instance.__getitem__(
```
jakehader, line 105: This won't work because you need to pass in a cs object into the method to get the actual cases settings to use. Otherwise this will also resolve to the default I believe. We want to move away from having global case settings.

jakehader (changes requested):

armi/reactor/converters/axialExpansionChanger.py
jakehader: I looked over this more closely and it does work because we have a global CS available in ARMI, but I believe that the design goal for ARMI is to remove the master CS. Is this still true @ntouran? Nonetheless I think that after we spoke offline it's better to just pass in `detailedAxialExpansion` as a kwarg on the init for this changer.

john-science: We _do_ want to stop using `getMasterCS()`, as it is somewhat problematic. I would _love_ to remove the global CS state, but it will not happen today.

albeanth: addressed. 

https://github.com/terrapower/armi/commit/286c294a6bb6f0f91f5b0ea7c230912f62fbcd29

armi/reactor/converters/axialExpansionChanger.py
albeanth: addressed. 

https://github.com/terrapower/armi/commit/286c294a6bb6f0f91f5b0ea7c230912f62fbcd29

PR #657, opened 2022-05-04
add `inputHeightsConsideredHot` setting

albeanth:
## Description
Enable thermal expansion of assemblies when cores are built from blueprints + settings within `armi.reactor.reactors.py::loadFromCs`. The thermal expansion of each assembly is determined by the DeltaT of each component (DeltaT = Thot - Tinput).  

To support backward compatibility (where no thermal expansion occurs) the setting, `inputHeightsConsideredHot`, was added. The default value for this is `True`. That way, current ARMI inputs retain their functionality. By setting to `False`, the armi.reactor.converters.axialExpansionChanger is spun up to expand each assembly as it is built. 

To add a unit test that leverages this new capability (and that has the required dummy block for detailedAxialExpansion and the axialExpansionChanger), a new test reactor was created. The details:

1. is based off of `armi.tests.armiRun.yaml` but sets:
```yaml
axialExpansion: true
detailedAxialExpansion: true
```
2. defines the required dummy block (`dummy : &block_dummy`) for `armi.reactor.converters.axialExpansionChanger.py`
3. adds said dummy block to top of each assembly 
    - dummy block height = 10% of original assembly height (i.e., total assembly height w/out dummy block)

---

## Checklist


albeanth:
@john-science @jakehader

john-science:
I see that you've added new test files.  But shouldn't I also see a new test that uses these files?

jakehader:
> I see that you've added new test files. But shouldn't I also see a new test that uses these files?

Agreed. @onufer talked about adding a setting to control the thermal expansion on load and setting it false by default. Can we add that setting in this PR and then add a test that shows that the expansion occurs with this setting is enabled with these test inputs and then doesn't when it is disabled?

armi/tests/detailedAxialExpansion/armiRun.yaml
```diff
@@ -0,0 +1,47 @@
+metadata:
```
john-science, line 1: Please add a unit test or two that use these files, before merging them into the repo.

albeanth: Addressed.

https://github.com/terrapower/armi/pull/657/commits/cdf1aed4b09b0d57a39c7e0aee68a975fccb43ec

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -555,6 +556,46 @@ def test_isFuelLocked(self):
...
+
+    def setUp(self):
+        """provide the base case"""
+        _o, r = loadTestReactor(
+            os.path.join(TEST_ROOT, "detailedAxialExpansion"),
+            {"inputHeightsConsideredHot": True},
+        )
+        self.stdAssems = [a for a in r.core.getAssemblies()]
+
+    def test_coldHeight(self):
```
john-science, line 570: So, the total height of the assemblies doesn't change if we enter cold inputs. But the height of every block inside it will?

I would have expect only certain blocks to grow in height, and ditto the assemblies.

I'm not _trying_ to be obtuse, I swear. But maybe a one-liner comment before each `for`-loop, that says what the `assert` statements are trying to prove?

john-science: Love the PR.

I only have this one question about what I'm supposed to understand from this test.  Then we can merge it, sure.

albeanth: > So, the total height of the assemblies doesn't change if we enter cold inputs. But the height of every block inside it will?

Yes, exactly. In the blueprints defined in `armi.tests.detailedAxialExpansion`, the overall height of the assembly is preserved through the use of the "dummy" block.
```yaml
    dummy : &block_dummy
        coolant:
            shape: Hexagon
            material: Sodium
            Tinput: 25.0
            Thot: 450.0
            ip: 0.0
            mult: 1.0
            op: 16.75
```
So in the standard case, 
```python
self.stdAssems = [a for a in r.core.getAssemblies()]
```
_most_ of the blocks are an even 25.0 cm (the control assemblies have some differences to eventually stress test uniformMesh.py). And in the expanded case,
```python
testAssems = [a for a in rCold.core.getAssemblies()]
```
all of the blocks (except the dummy) are expanded based on the DeltaT between `Tinput` and `Thot`. The dummy block is then "chopped" to account for the accumulated change in height in the blocks below (this happens in armi.reactor.converters.axialExpansionChanger). So assuming that `Thot` > `Tinput` everywhere (which it is in `armi.tests.detailedAxialExpansion`), all of the blocks will be different heights (the specific height will vary on the temperature difference and expansion coefficients of the materials) but the overall assembly height will be preserved.

albeanth: @john-science , a few comments added to the docstring.

john-science (approved):

john-science:
@albeanth  Is this PR ready to be merged?

albeanth:
> @albeanth Is this PR ready to be merged?

@john-science yes, unless @ntouran, @jakehader, or anyone else wants to review.

Issue #406, opened 2021-09-08
Folder deletion in pathTools appears to be accounting for potential parallelism oddly


youngmit:
In the implementation of `cleanPath()` there is this fun chunk of code:

https://github.com/terrapower/armi/blob/54664d800db9eec1f6dff584d5eda65329048d49/armi/utils/pathTools.py#L266-L277

The only reason I can imagine for attempting the deletion in a loop with delays and a broad try/except if for scenarios in which the folder deletion is being attempted in parallel, and/or targeting a directory structure on a shared network drive. If this is the case, then having every processor try to delete the folder and hoping for the best is a pretty sketchy way to go about it. If this is supposed to be possible in parallel, we should actually address the complexities of air-traffic control explicitly in MPI.

This would look something like having only one rank responsible for the deletion, while all others wait until the directory is apparently removed, and a barrier at the end to synchronize. The main concern here are questions like:
 - do we expect all call sites to be collective? if not, any communication that we do may lead to deadlocks if not all processors in a communicator are trying to `cleanPath()` at the same time
 - are all processors in a communicator attempting to clear the same path? if not, one rank per desired path deletion will need to be responsible

These aren't simple questions to answer from within the function, so it is likely that decisions like this should be made from the call site. Something like this is easy enough to do:

```python
if armi.MPI_RANK == 0:
    clearPath(path)
while os.path.exists(path):
    sleep(0.1)
armi.MPI_COMM.barrier()
```

jakehader:
This may be separate from this, but I have experienced issues when writing a file to a drive location and then immediately trying to access it where I can an OSError. I wonder if we were to delete these sleep timers and run some cases if this would uncover other issues.

davis8dd:
Would it be overkill to refactor **pathTools** to rely completely on mpi4pi for file I/O?

john-science:
> Would it be overkill to refactor **pathTools** to rely completely on mpi4pi for file I/O?

That might be slightly problematic, as `mpi4py` is not a mandatory requirement:

https://github.com/terrapower/armi/blob/8cc5bce4e6c4f8c8a34df0761eac426c10666603/setup.py#L75-L76

The system is designed so someone can run little cases single-threaded on their laptop, without needing to install the Fortran MPI libraries.

That being said, `pathTools` could behave in more cleaver ways if MPI is present.

davis8dd:
[...]

PR #664, opened 2022-05-09
Let's not delete a file inside a loop

john-science:
## Description

The old `pathTools()` would repeatedly delete the same file/directory inside a loop.  That seems suspect.

It also did not handle the situation well for when we want to delete a file/directory in an MPI or non-MPI scenario well.

This change to the method linearizes our file deletion, to only one delete call. It also gives the user control of whether or not this deletion should be MPI-safe or not.

This is in response to: https://github.com/terrapower/armi/issues/406

**NOTE**:
I also fixed a bug in the `HexBlock.hasPinPitch()`, which sometimes returned the Block itself, rather than just a boolean. That is, it's return type was internally inconsistent, and did not match `Block.hasPinPitch()`.

---

## Checklist

### DIFF
--- a/armi/tests/test_mpiFeatures.py
+++ b/armi/tests/test_mpiFeatures.py
@@ -252,7 +252,7 @@ class MpiDistributeStateTests(unittest.TestCase):
 
 class MpiPathToolsTests(unittest.TestCase):
     @unittest.skipIf(context.MPI_SIZE <= 1 or MPI_EXE is None, "Parallel test only")
-    def xxxx_cleanPathMpi(self):
+    def test_cleanPathMpi(self):
         # """Simple tests of cleanPath(), in the MPI scenario"""
         # Test 0: File is not safe to delete, due to name pathing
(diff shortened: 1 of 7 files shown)


armi/tests/test_mpiFeatures.py
```diff
@@ -248,6 +250,52 @@ def test_compileResults(self):
         self.assertEqual(results1, results2)
 
 
+class MpiPathToolsTests(unittest.TestCase):
+    @unittest.skipIf(context.MPI_SIZE <= 1 or MPI_EXE is None, "Parallel test only")
+    def xxxx_cleanPathMpi(self):
```
ntouran, line 255: what's with the xxxx naming?

ntouran (approved):
good update, good tests. One question about test name but otherwise fine.

armi/tests/test_mpiFeatures.py
john-science: Fixed

PR #665, opened 2022-05-09
Update stationary block methods

HunterPSmith:
## Description

Updated the way that ARMI handles stationary blocks. ARMI now has a global setting called `stationaryBlockFlags` which can be set by the user to identify flags for stationary blocks. By default, `GRID_PLATE` has a stationary flag, unless set by user in the input file. When a reactor object is initialized, a reactor attribute called `stationaryBlockFlagsList` is generated which contains a list of the flags in the reactor that are considered stationary. ARMI now operates on `stationaryBlockFlagsList` in order to control stationary blocks.

Added testing to satisfy requirement MDN-FMM-004 of the [FMM SRSD](https://sfr.terrapower.com/display/NATRIUMD/MDN+Report+-+Fuel+Management+Module+SRSD). 

This PR will require the MCNP plugin to be updated.

---

## Checklist


armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from _operator import index
+from _ast import Or
```
john-science, line 15: Sorry, I might be blind. But I don't see these imports being used.  

Are these unused imports, or am I missing something obvious?

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -736,6 +736,127 @@ def test_swapFluxParamDifferentLengths(self):
         for b in a2.getBlocks():
             self.assertEqual(b.p.flux, 50000000000.0)
 
+    def test_transferStationaryBlocks(self):
```
john-science, line 739: I love these tests!

Can we also get a test or two for `dischargeSwap()`, that shows this feature working.

Thanks!

john-science:
I love it!

I have made two comments/requests. But they are pretty minor.

ntouran (changes requested):
Seems like a good change. 

* This is a breaking change and will break all previous inputs that used the `stationaryBlocks` setting. This needs to be documented in the documentation and in the release notes. 
* Would you consider adding a migration so that users with the deprecated setting can proceed easily? Or at least provide a meaningful error message telling them how to proceed.

armi/settings/fwSettings/globalSettings.py
```diff
@@ -590,10 +590,10 @@ def defineSettings() -> List[setting.Setting]:
...
         ),
         setting.Setting(
-            CONF_STATIONARY_BLOCKS,
-            default=[],
-            label="stationary Blocks",
-            description="blocks with these indices (int values) will not move in "
+            CONF_STATIONARY_BLOCK_FLAGS,
+            default=["GRID_PLATE"],
+            label="stationary Block Flags",
+            description="blocks with these flags (int values) will not move in "
```
jakehader, line 596: Add to the description that this is used in the fuel management so that it is not confused with load pads for vertical movements.

jakehader (changes requested):

HunterPSmith:
> Seems like a good change.
> 
> * This is a breaking change and will break all previous inputs that used the `stationaryBlocks` setting. This needs to be documented in the documentation and in the release notes.
> * Would you consider adding a migration so that users with the deprecated setting can proceed easily? Or at least provide a meaningful error message telling them how to proceed.

I am not familiar which specific documentation this would go in. I assume that the release notes are here: `/nala/framework/doc/release/0.2.rst` ?

john-science:
> I am not familiar which specific documentation this would go in. I assume that the release notes are here: `/nala/framework/doc/release/0.2.rst` ?

That's it!  A quick one-sentence explanation of the breaking change.

armi/settings/fwSettings/globalSettings.py
```diff
@@ -604,12 +605,20 @@ def defineSettings() -> List[setting.Setting]:
...
+        ),
         setting.Setting(
             CONF_STATIONARY_BLOCKS,
             default=[],
             label="stationary Blocks",
-            description="blocks with these indices (int values) will not move in "
-            "moves.",
+            description="Blocks with these indices (int values) will not move in "
+            "moves. Used for fuel management. "
+            "Deprecated setting, use CONF_STATIONARY_BLOCK_FLAGS",
```
jakehader, line 621: I think we may as well remove this now that it is being replaced.

armi/reactor/reactors.py
```diff
@@ -2251,19 +2251,19 @@ def processLoading(self, cs):
...
-                # TODO: remove hard-coded assumption of grid plates (T3019)
-                runLog.extra(
-                    "Detected a grid plate {}.  Adding to stationary blocks".format(b)
-                )
+        # Generate list of flags that are to be stationary during assembly shuffling
+
+        if len(cs["stationaryBlocks"]) >= 1:
+            self.ValueError(
+                "stationaryBlocks setting is deprecated. Use stationaryBlocksFlags."
+            )
```
jakehader, line 2259: Maybe just remove logic that uses the `stationaryBlocks` setting since it will fail anyways.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1172,17 +1166,43 @@ def _transferStationaryBlocks(self, assembly1, assembly2):
...
-            tempBlock2 = assembly2[index]
-            assembly2.remove(tempBlock2)
+        # check for any inconsistencies in stationary blocks and ensure alignment
+        if [block[1] for block in a1StationaryBlocks] != [
+            block[1] for block in a2StationaryBlocks
+        ]:
+            raise ValueError(
+                """Different number and/or locations of stationary blocks 
+                 between {} and {}.""".format(
+                    assembly1, assembly2
```
jakehader, line 1192: It would be a better error to present the user with which specific stationary blocks are mismatched by also including the the print out of `a1StationaryBlocks` and `a2StationaryBlocks` so that the user isn't left wondering what to do with this failure.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1172,17 +1166,43 @@ def _transferStationaryBlocks(self, assembly1, assembly2):
...
+                )
+            )
 
-            assembly1.insert(index, tempBlock2)
-            assembly2.insert(index, tempBlock1)
+        # swap stationary blocks
+        for blocksAssembly1, blocksAssembly2 in zip(
```
jakehader, line 1197: Might be more readable if we did something like:

```
for (assem1Block, assem1BlockIndex),  (assem2Block, assem2BlockIndex) in zip(a1StationaryBlocks, a2StationaryBlocks):
...
```

This way it's clearer in the code that the remove and insert statements below are using the correct indices and block replacements.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1243,6 +1263,10 @@ def _swapFluxParam(self, incoming, outgoing):
         incoming, outgoing : Assembly
```
jakehader, line 1263: This flux and parameter swapping seems very strange to me still and is likely error prone. I really don't see why we _need_ to migrate the data from the assembly being discharged to the assembly being charged. Basically I foresee that this can introduce many bugs.

For example, consider a replacement of an assembly that is not of the same type (i.e., fuel with reflector or control). In this case, the power is flux and power are going to be immediately out dated and could be drastically incorrect. I think that removing this `_swapFluxParam` method is going to cause other things in the framework and code to break, but this entire process just seems unsafe.

jakehader (changes requested):

armi/settings/fwSettings/globalSettings.py
HunterPSmith: Removed the setting.

armi/reactor/reactors.py
HunterPSmith: Removed the logic here.

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: Now prints out the stationary blocks

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: Implemented changes for clarity

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: Removed ``_swapFluxParam`` entirely.

armi/settings/fwSettings/globalSettings.py
john-science: Awesome! This will affect some downstream repos.

armi/physics/fuelCycle/fuelHandlers.py
john-science: @keckler  @ntouran  Can you guys weigh in on the solution used here, to completely remove `_swapFluxParam()`?

Thanks!

keckler: I have never liked `_swapFluxParam()`. I think it could lead to some very confusing situations, in particular for everything that sits in the SFP. And I agree with all the issues that Jake stated.

That being said, I do think things will break if we just get rid of it completely.  But you probably would only see that in an integration test that (1) runs deterministic transport and (2) has shuffling. Might be other necessary features as well to see issues, but I'm not sure.

My _wish_ is to delete it, and barring that, to replace it with something smarter and more targeted.

jakehader (approved):
@HunterPSmith, very sorry for the long review time on this. Can you work with @opotowsky to make sure we understand the full scope of the impact to internal plugins offline and testing before we land this?

opotowsky (changes requested):
Just a quick cleanup request: 

This line in `armi/reactor/tests/test_reactors.py` should be deleted:

image

opotowsky:
@jakehader @HunterPSmith Internal plugins and testing are a go. This PR is good to merge from that perspective.

jakehader:
Great, thanks @opotowsky! @HunterPSmith, can you address the last comments here and we will merge?

HunterPSmith:
Thank you both! Should be good to go now

opotowsky (approved):

john-science (approved):

PR #670, opened 2022-05-13
Update/upgrade for component axial linkage

albeanth:
## Description
The current methodology for determining component axial linkage was found to be error prone. The current approach requires that Components be the same class type and have an area with an absolute difference of 10^-3. For some cases, this can provide false positive linkage. To improve this, an overhaul of the methodology was undertaken. In this PR, this methodology is updated/upgraded to to account for the following: 1) Component classes must be solids, 2) the same class type, 3) have the same multiplicity, 4) and meet overlap criteria. A `runLog.warning` has also been added if a Component is found to be linked to multiple Components in a neighboring block.

The methodology can be found in `axialExpansionChanger.py::AssemblyAxialLinkage::_determineAxialLinkage`. Several relevant unit tests have been added to cover various expected cases - see `test_AxialExpansionChanger.py::TestLinkage`. Other unit tests have also been added to maintain coverage in `basicShapes.py` and `complexShapes.py`.  

---

## Checklist

### DIFF
--- a/armi/reactor/components/basicShapes.py
+++ b/armi/reactor/components/basicShapes.py
@@ -71,6 +71,9 @@ class Circle(ShapedComponent):
     def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
         return max(self.getDimension("id", Tc, cold), self.getDimension("od", Tc, cold))
 
+    def getBoundingCircleInnerDiameter(self, Tc=None, cold=False):
+        return min(self.getDimension("id", Tc, cold), self.getDimension("od", Tc, cold))
+
     def getComponentArea(self, cold=False):
         """Computes the area for the circle component in cm^2."""
         idiam = self.getDimension("id", cold=cold)
@@ -126,6 +129,10 @@ class Hexagon(ShapedComponent):
 
     def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
         sideLength = self.getDimension("op", Tc, cold) / math.sqrt(3)
+        return 2.0 * sideLength
+
+    def getBoundingCircleInnerDiameter(self, Tc=None, cold=False):
+        sideLength = self.getDimension("ip", Tc, cold) / math.sqrt(3)
         return 2.0 * sideLength
 
     def getComponentArea(self, cold=False):
@@ -212,6 +219,11 @@ class Rectangle(ShapedComponent):
         lengthO = self.getDimension("lengthOuter", Tc, cold=cold)
         widthO = self.getDimension("widthOuter", Tc, cold=cold)
         return math.sqrt(widthO ** 2 + lengthO ** 2)
+
+    def getBoundingCircleInnerDiameter(self, Tc=None, cold=False):
+        lengthI = self.getDimension("lengthInner", Tc, cold=cold)
+        widthI = self.getDimension("widthInner", Tc, cold=cold)
+        return math.sqrt(widthI ** 2 + lengthI ** 2)
 
     def getComponentArea(self, cold=False):
         """Computes the area of the rectangle in cm^2."""
@@ -348,6 +360,10 @@ class Square(Rectangle):
     def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
         widthO = self.getDimension("widthOuter", Tc, cold=cold)
         return math.sqrt(widthO ** 2 + widthO ** 2)
+
+    def getBoundingCircleInnerDiameter(self, Tc=None, cold=False):
+        widthI = self.getDimension("widthInner", Tc, cold=cold)
+        return math.sqrt(widthI ** 2 + widthI ** 2)
 
     def getPitchData(self):
         """
--- a/armi/reactor/components/basicShapes.py
+++ b/armi/reactor/components/basicShapes.py
@@ -71,7 +71,7 @@ class Circle(ShapedComponent):
     def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
         return max(self.getDimension("id", Tc, cold), self.getDimension("od", Tc, cold))
 
-    def getBoundingCircleInnerDiameter(self, Tc=None, cold=False):
+    def getCircleInnerDiameter(self, Tc=None, cold=False):
         return min(self.getDimension("id", Tc, cold), self.getDimension("od", Tc, cold))
 
     def getComponentArea(self, cold=False):
@@ -131,7 +131,7 @@ class Hexagon(ShapedComponent):
         sideLength = self.getDimension("op", Tc, cold) / math.sqrt(3)
         return 2.0 * sideLength
 
-    def getBoundingCircleInnerDiameter(self, Tc=None, cold=False):
+    def getCircleInnerDiameter(self, Tc=None, cold=False):
         sideLength = self.getDimension("ip", Tc, cold) / math.sqrt(3)
         return 2.0 * sideLength
 
@@ -220,7 +220,7 @@ class Rectangle(ShapedComponent):
         widthO = self.getDimension("widthOuter", Tc, cold=cold)
         return math.sqrt(widthO ** 2 + lengthO ** 2)
 
-    def getBoundingCircleInnerDiameter(self, Tc=None, cold=False):
+    def getCircleInnerDiameter(self, Tc=None, cold=False):
         lengthI = self.getDimension("lengthInner", Tc, cold=cold)
         widthI = self.getDimension("widthInner", Tc, cold=cold)
         return math.sqrt(widthI ** 2 + lengthI ** 2)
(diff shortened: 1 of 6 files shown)


albeanth:
@john-science @onufer

armi/reactor/components/basicShapes.py
```diff
@@ -71,6 +71,9 @@ def __init__(
     def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
         return max(self.getDimension("id", Tc, cold), self.getDimension("od", Tc, cold))
 
+    def getBoundingCircleInnerDiameter(self, Tc=None, cold=False):
+        return min(self.getDimension("id", Tc, cold), self.getDimension("od", Tc, cold))
```
john-science, line 75: Sorry, I don't follow.   Why does this method and the "outer" version use min/max?

Should "id" be min and "od" be max?

Just wondering what I don't understand.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -153,6 +153,7 @@ def axiallyExpandAssembly(self, thermal: bool = False):
         mesh = [0.0]
         numOfBlocks = self.linked.a.countBlocksWithFlags()
         for ib, b in enumerate(self.linked.a):
+            runLog.debug(msg="Block {0}".format(b))
```
john-science, line 156: I leave it to you to decide.  But do we want these `debug` statements here for all time?  

Or are these only useful to you during your dev process?

armi/reactor/components/basicShapes.py
albeanth: This is great question. I figured there was a reason for using `max()` in the "outer" method so I just followed suit for the `min()` case. Happy to change both if there is support for it.

armi/reactor/tests/test_components.py
```diff
@@ -339,13 +339,14 @@ def test_getBoundingCircleOuterDiameter(self):
...
     _od = 10
     _coldTemp = 25.0
     componentDims = {
         "Tinput": _coldTemp,
         "Thot": 25.0,
         "od": _od,
-        "id": 5.0,
+        "id": _id,
         "mult": 1.5,
     }
```
john-science, line 351: I love the tests!

Can we move this stuff into a `def setUp()` method?  Take a look around ARMI, we use them in many/most test files.

Just to make things a bit more consistent across the repo.

Thanks!

armi/reactor/components/basicShapes.py
john-science: hmm...

I'm not thrilled with not know why things are the way they are.  hmmm

armi/reactor/converters/axialExpansionChanger.py
albeanth: They certainly originated from the dev process. I found them to be particularly useful for debugging and suspect that users would too if their results aren't producing any obvious errors and expected results. Happy to take them out/modify them if users don't like them but I think they have utility.

armi/reactor/components/basicShapes.py
albeanth: @ntouran @jakehader @onufer  do you have any idea why `getBoundingCircleOuterDiameter()` uses `max()`?

armi/reactor/tests/test_components.py
albeanth: Happy to. This was just how the tests were set up when I got there. Many of the other test classes in this file do not use the `def setUp()` approach and just define class variables. Do you want me to take a stab at that too? If so, maybe it would be better to have it in a separate PR? Just to keep changes separated a little.

john-science: Nevermind, if it's in lots of places, I will fix them all in one swoop later.

armi/reactor/components/basicShapes.py
onufer: Sometimes when you define components between solid interfaces that have vacuums, they have negative area, and their ID becomes greater than OD because ID is linked to one component and OD to another, and those components overlap.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -427,6 +449,54 @@ def _getLinkedComponents(self, b, c):
...
+    linked : bool
+        status is componentA and componentB are axially linked to one another
+    """
+    if (
+        (componentA.containsSolidMaterial() and componentB.containsSolidMaterial())
+        and isinstance(componentA, type(componentB))
+        and (componentA.getDimension("mult") == componentB.getDimension("mult"))
+    ):
+        idA, odA = (
+            componentA.getBoundingCircleInnerDiameter(),
```
onufer, line 479: looking at the way this is implemented it needs to be how you have structured it so we should probably rename to 
getCircleInnerDiameter

armi/reactor/components/basicShapes.py
```diff
@@ -128,6 +131,10 @@ def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
         sideLength = self.getDimension("op", Tc, cold) / math.sqrt(3)
         return 2.0 * sideLength
 
+    def getBoundingCircleInnerDiameter(self, Tc=None, cold=False):
+        sideLength = self.getDimension("ip", Tc, cold) / math.sqrt(3)
```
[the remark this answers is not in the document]
onufer: On second examination, it needs to work like this for your functions to work. so lets remove the word bounding since the circle will pass through he component. or call it 

getInteriorCornerCircleDiameter

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -427,6 +449,54 @@ def _getLinkedComponents(self, b, c):
...
+            componentA.getBoundingCircleOuterDiameter(),
+        )
+        idB, odB = (
+            componentB.getBoundingCircleInnerDiameter(),
+            componentB.getBoundingCircleOuterDiameter(),
+        )
+
+        biggerID = max(idA, idB)
+        smallerOD = min(odA, odB)
+        if biggerID >= smallerOD:
```
onufer, line 489: Elegant, nice! maybe ad a comment in the next line saying
#one object fits inside the other.

onufer (changes requested):

armi/reactor/components/basicShapes.py
john-science: Thanks @onufer !

armi/reactor/converters/axialExpansionChanger.py
albeanth: Addressed. https://github.com/terrapower/armi/pull/670/commits/b5b468c5b7e1ea8838fbc9d7e933fa957c6701f9

armi/reactor/components/basicShapes.py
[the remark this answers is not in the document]
albeanth: Addressed. https://github.com/terrapower/armi/pull/670/commits/81318e16eddf5fa360597ee3cf1c148fcc72a560

armi/reactor/components/component.py
```diff
@@ -783,6 +783,10 @@ def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
         """Abstract bounding circle method that should be overwritten by each shape subclass."""
         raise NotImplementedError
 
+    def getCircleInnerDiameter(self, Tc=None, cold=False):
+        """Abstract inner bounding circle method that should be overwritten by each shape subclass."""
```
onufer, line 787: This needs a better doc string.  Mentioning intersecting corners would add value.

onufer (changes requested):

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -396,12 +418,12 @@ def _getLinkedComponents(self, b, c):
...
                 for otherC in linkdBlk.getChildren():
-                    if isinstance(
-                        otherC, type(c)
-                    ):  # equivalent to type(otherC) == type(c)
-                        area_diff = abs(otherC.getArea() - c.getArea())
-                        if area_diff < self._TOLERANCE:
-                            lstLinkedC[ib] = otherC
+                    if _determineLinked(c, otherC):
+                        if lstLinkedC[ib] is not None:
+                            runLog.warning(
```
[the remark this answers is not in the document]
onufer: if we think this indicates a input error, lets change to a value error instead of warning

armi/reactor/converters/axialExpansionChanger.py
albeanth: Addressed https://github.com/terrapower/armi/pull/670/commits/138bef7694cfa3b360b67775e5fcec140f1515c1

armi/reactor/converters/axialExpansionChanger.py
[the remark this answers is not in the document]
albeanth: Addressed https://github.com/terrapower/armi/pull/670/commits/bc4be11ff2c2805f870478c1ebd6ea22d97a3817

armi/reactor/components/component.py
albeanth: Addressed https://github.com/terrapower/armi/pull/670/commits/7eb3cccc97bc6d135c73c9ab372f5a5627fc0585

onufer (approved):

PR #671, opened 2022-05-13
Fix `invokeAsMaster` references in the docs

keckler:
For whatever reason, all the docs were using the phrase `invokeAsprimary`, but the real method name is `invokeAsMaster`: https://terrapower.github.io/armi/.apidocs/armi.mpiActions.html?highlight=invokeasmaster#armi.mpiActions.MpiAction.invokeAsMaster

## Description

---

## Checklist


john-science (approved):
I actually like the word "Primary" better than "Master".

But the docs were wrong, yeah.

PR #677, opened 2022-05-20
Fix setting the target component in axialExpansion

opotowsky:
## Description

If more than one flag is in a block definition, the `specifyTargetComponent` method would fail with a Runtime error. The proposed fix is to loop the through a list of flags in order of importance break the loop once one of those flags matches. Still, this list of flags is incomplete so it is possible that `componentWFlag` would remain empty. In this case the original behavior is retained. 

The unit test is a little redundant (testing that a RuntimeError is not raised, which is implicit in testing?) but I felt it important to be explicit there. Definitely willing to do something different. 

---

## Checklist

### DIFF
--- a/armi/reactor/converters/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger.py
@@ -670,7 +670,7 @@ class ExpansionData:
         Notes
         -----
         - if flagOfInterest is None, finds the component within b that contains flags that
-          are defined in a preferred order of flags, or barring that, in b.p.flags 
+          are defined in a preferred order of flags, or barring that, in b.p.flags
         - if flagOfInterest is not None, finds the component that contains the flagOfInterest.
 
         Raises
@@ -691,13 +691,13 @@ class ExpansionData:
             ]
             for targetFlag in targetFlagsInPreferedOrder:
                 componentWFlag = [c for c in b.getChildren() if c.hasFlags(targetFlag)]
-                if componentWFlag != []:  # or just `if componentWFlag`
+                if componentWFlag != []:
                     break
+            # some blocks/components are not included in the above list but should still be found
+            if not componentWFlag:
+                componentWFlag = [c for c in b.getChildren() if c.p.flags in b.p.flags]
         else:
             componentWFlag = [c for c in b.getChildren() if c.hasFlags(flagOfInterest)]
-        # some blocks/components are not included in the above list but should still be found
-        if not componentWFlag:
-            componentWFlag = [c for c in b.getChildren() if c.p.flags in b.p.flags]
         if len(componentWFlag) == 0:
             raise RuntimeError("No target component found!\n   Block {0}".format(b))
         if len(componentWFlag) > 1:
--- a/armi/reactor/converters/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger.py
@@ -682,10 +682,6 @@ class ExpansionData:
         """
 
         if flagOfInterest is None:
-            componentWFlag = [c for c in b.getChildren() if c.p.flags in b.p.flags]
-        else:
-            componentWFlag = [c for c in b.getChildren() if c.hasFlags(flagOfInterest)]
-        if len(componentWFlag) > 1:
             # Follow expansion of most neutronically important component, fuel first then control/poison
             targetFlagsInPreferedOrder = [
                 Flags.FUEL,
@@ -694,11 +690,14 @@ class ExpansionData:
                 Flags.SHIELD,
             ]
             for targetFlag in targetFlagsInPreferedOrder:
-                for c in componentWFlag:
-                    if targetFlag in c.p.flags:
-                        componentWFlag = [c]
-                    if len(componentWFlag) == 1:
-                        break
+                componentWFlag = [c for c in b.getChildren() if c.hasFlags(targetFlag)]
+                if componentWFlag != []:  # or just `if componentWFlag`
+                    break
+        else:
+            componentWFlag = [c for c in b.getChildren() if c.hasFlags(flagOfInterest)]
+        # some blocks/components are not included in the above list but should still be found
+        if not componentWFlag:
+            componentWFlag = [c for c in b.getChildren() if c.p.flags in b.p.flags]
         if len(componentWFlag) == 0:
             raise RuntimeError("No target component found!\n   Block {0}".format(b))
         if len(componentWFlag) > 1:
--- a/armi/reactor/converters/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger.py
@@ -680,10 +680,25 @@ class ExpansionData:
         RuntimeError
             multiple target components found
         """
+
         if flagOfInterest is None:
             componentWFlag = [c for c in b.getChildren() if c.p.flags in b.p.flags]
         else:
             componentWFlag = [c for c in b.getChildren() if c.hasFlags(flagOfInterest)]
+        if len(componentWFlag) > 1:
+            # Follow expansion of most neutronically important component, fuel first then control/poison
+            targetFlagsInPreferedOrder = [
+                Flags.FUEL,
+                Flags.CONTROL,
+                Flags.POISON,
+                Flags.SHIELD,
+            ]
+            for targetFlag in targetFlagsInPreferedOrder:
+                for c in componentWFlag:
+                    if targetFlag in c.p.flags:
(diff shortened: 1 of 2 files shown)


opotowsky:
@albeanth I'd appreciate your feedback on the change since this affects your recent work. 

And @onufer, note that I had to change this quite a bit after our offline discussion on this change (which I can outline why if it isn't clear to you).

john-science:
I appreciate the fix!

Is there some unit test way we could see the fix in action?

opotowsky:
> Is there some unit test way we could see the fix in action?

In progress, for certain :-)

albeanth:
@opotowsky I think this is a solid idea! This section may easily be the least robust in the axialExpansionChanger. Just a few points:

> If more than one flag is in a block definition, the specifyTargetComponent method would fail with a Runtime error.

I wouldn't expect this to be the case, but that's a major flaw if it is! When I wrote up the `RuntimeError`, the intent was that if more than one component has flags that match the block flags, the `RuntimeError` is thrown. Having multiple flags on a block should be totally fine. 

This test provides context for multiple components with the same flag as the block flag. You could play with this and change the flags on both the block and components to see if/how things break and how you'd like to improve them.
https://github.com/terrapower/armi/blob/bb3db53b4274d2d61d8fc6a9ad585c97a4529ea7/armi/reactor/converters/tests/test_axialExpansionChanger.py#L514-L530.

I agree with @john-science, a unit test or several would be great. If there is a context (contexts) that you and @onufer have thought of, a new test class may be a good idea.

opotowsky:
@jstilley a test has been added to show the behavior I was trying to fix. I'm not confident what I did was a best practice (see PR description), so hopefully you and @albeanth can comment on that.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -680,10 +680,24 @@ def specifyTargetComponent(self, b, flagOfInterest=None):
...
+            # Follow expansion of most neutronically important component, fuel first then control/poison
+            targetFlagsInPreferedOrder = [
+                Flags.FUEL,
+                Flags.CONTROL,
+                Flags.POISON,
+                Flags.SHIELD,
+            ]
+            for targetFlag in targetFlagsInPreferedOrder:
+                componentWFlag = [c for c in b.getChildren() if c.hasFlags(targetFlag)]
+                if componentWFlag != []:  # or just `if componentWFlag`
```
onufer, line 694: probably delete comment here

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -680,10 +680,24 @@ def specifyTargetComponent(self, b, flagOfInterest=None):
...
+                    break
         else:
             componentWFlag = [c for c in b.getChildren() if c.hasFlags(flagOfInterest)]
+        # some blocks/components are not included in the above list but should still be found
+        if not componentWFlag:
```
onufer, line 699: this block should probably be after line 695. If the flag of interest isn't present this should still be an error.

onufer (changes requested):

armi/reactor/converters/axialExpansionChanger.py
opotowsky: agreed, done

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -529,6 +529,20 @@ def test_specifyTargetComponentRuntimeErrorSecond(self):
             the_exception = cm.exception
             self.assertEqual(the_exception.error_code, 3)
 
+    def test_specifyTargetComponentBlockWithMultipleFlags(self):
+        # build block for testing
```
onufer, line 533: you could update this comment to say why this block is special ( block has both fuel and poison flags, and components)

onufer (approved):

john-science:
@opotowsky   I like it!

Two thoughts:

1. Should we move `targetFlagsInPreferedOrder` to be an ALL_CAPS class or module-level constant?  I'm not sure we should be defining constants in such an easy to overlook place.
2. Why is it we only have unit tests that show `specifyTargetComponent` failing?  Should we have a test that shows it working to reset the block height?

I could be way off base here, if so shout me down.  Just asking.

opotowsky:
@john-science both very good points. 
1. I do agree that `targetFlagsInPreferredOrder` would be easy to overlook. Happy to follow the norm of all caps + module level. 
2. Yes, good observation! And my test only checks that the method doesn't give the Runtime error I was seeing, not that `_componentDeterminesBlockHeight` is `True`.  Can @albeanth give some insight into this? We probably should add a test with an assembly that gets a component key in the dict.

john-science:
@opotowsky As far as number (2) goes, I was thinking a unit tests would like something like:

1. Build a block
2. test `self._componentDeterminesBlockHeight` looks like it should
3. Call `specifyTargetComponent()` in some way that it doesn't raise an error
4. test that `self._componentDeterminesBlockHeight` has changed in some way

That's all. I just got surprised when I only saw tests where the method `specifyTargetComponent()` failed.  That's strange. Like, normally, people start by proving the method _works_ in some way.

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -563,6 +563,50 @@ def test_isFuelLocked(self):
...
+        clad = Circle("clad", "FakeMat", **cladDims)
+        b.add(fuel)
+        b.add(clad)
+        # call method, and check that _componentDeterminesBlockHeight has been updated
+        self.obj.expansionData.specifyTargetComponent(b)
+        newEntry = dict(
+            set(self.obj.expansionData._componentDeterminesBlockHeight.items())
+            - set(initialComponentDeterminesBlockHeight.items())
+        )
+        self.assertEqual(newEntry, {fuel: True})
```
opotowsky, line 592: @john-science can you check that this is an OK way to test for the update to the dictionary? Also note line 575 above, where I made a copy to snapshot the dict prior to updating.

john-science (approved):
Thanks!

opotowsky:
@john-science That took a minute but I have added a test that mostly accomplishes what you said. I also moved the test for multiple flags out of the test-excptions class (since I'm not actually testing an exception).

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -563,6 +563,50 @@ def test_isFuelLocked(self):
...
+        b = HexBlock("fuel poison", height=10.0)
+        fuelDims = {"Tinput": 25.0, "Thot": 600.0, "od": 0.9, "id": 0.5, "mult": 200.0}
+        poisonDims = {"Tinput": 25.0, "Thot": 400.0, "od": 0.5, "id": 0.0, "mult": 10.0}
+        fuel = Circle("fuel", "FakeMat", **fuelDims)
+        poison = Circle("poison", "FakeMat", **poisonDims)
+        b.add(fuel)
+        b.add(poison)
+        try:
+            self.obj.expansionData.specifyTargetComponent(b)
+        except RuntimeError:
```
albeanth, line 606: @opotowsky , one comment here - we could make this assertion more precise/restrictive by requiring that the target component be equal to `fuel`, rather than just checking that it `specifyTargetComponent` ran through - i.e., check that the method set the right/expected target component. So you could do something along the lines of:
```python
self.obj.reset() # just to clear everything out, or you could generate a new AxialExpansionChanger instance
self.obj.specifyTargetComponent(b)
self.assertTrue(self.obj.isTargetComponent(fuel), msg="<something to say that it failed>"
```

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -18,6 +18,13 @@
 from armi import runLog
 from armi.reactor.flags import Flags
 
+TARGET_FLAGS_IN_PREFERRED_ORDER = [
+    Flags.FUEL,
+    Flags.CONTROL,
+    Flags.POISON,
+    Flags.SHIELD,
```
albeanth, line 25: @opotowsky @onufer Does it make sense to include `Flags.SLUG` in this list as well? Possibly the last value in the list.

armi/reactor/converters/tests/test_axialExpansionChanger.py
john-science: Oh, this comment comes after I hit Merge.  hmm

You guys can decide if these things need to be thrown in as a mini-PR.

I'm all for little PRs that improve things!

Issue #680, opened 2022-05-23
NHFLUX reader is broken for DIF3D-VARIANT


Nebbychadnezzar:
The `nhflux` module is advertised to be able to read/write the DIF3D-VARIANT NHFLUX output file. However, both testing (using the `variant` flag) and inspection of the file specification demonstrate this this is not true. I do not know the full history of VARIANT's NHFLUX file format, so the cause of this issue is not clear to me. However, what is definitely true is that DIF3D-Nodal and DIF3D-VARIANT currently have different NHFLUX file formats, and the `nhflux` module can only process Nodal.

There are several issues that are causing this to occur. In no particular order:

1. The 2D block for VARIANT contains additional datasets for what it calls "symmetric" and "sector" boundaries. Nodal does not have these boundaries.
2. VARIANT stores both even and odd flux moments to the 3D block. Nodal has no concept of even and odd flux moments. The number of flux moments that Nodal thinks are present is controlled by the "nMom" parameter. The number of even and odd flux moments that VARIANT thinks are present is controlled by the "nMom" and "nMoms" parameters; both parameters need to be processed.
3. Somewhat related, when running VARIANT and using the VARPOW postprocessor (which stores power moments instead of flux moments), the NHFLUX file produced by VARPOW does not contain any currents (since the idea of a power current doesn't make any sense). The lack of currents is noted in the file using the "iwnhfl" parameter, but the `nhflux` is not prepared for addressing this parameter; it assumes that "iwnhfl"=0 (both fluxes and currents are present), but in the case of using VARPOW, "iwnhfl"=1 (only fluxes are present). In this case, the 4D and 5D blocks need to be ignored.

The best way to address this issue is not clear to me. I have developed short-term fixes that address the problem, but it means that some VARIANT data is ignored by the module. This is fine for my purposes, and probably fine for other people as well. The module currently ignores some Nodal data too, so clearly reading and not storing the entire NHFLUX file was previously acceptable...so, that makes me believe that my short-term fixes may be acceptable too.

ntouran:
There's some long and relevant discussion and conversations with Mike Smith of ANL on this topic somewhere. I'll see if I can find it.

Nebbychadnezzar:
[...]

PR #681, opened 2022-05-23
Nhflux variant bug

Nebbychadnezzar:
## Description

This pull request fixes issue [680](https://github.com/terrapower/armi/issues/680). Achieving complete reading of the VARIANT NHFLUX file is not intended by this pull request.

---

## Checklist


Nebbychadnezzar:
I know there aren't added tests here yet. I'd like some input on that before I start writing a test.

I have tested this updated code on a representation model using both Nodal and VARIANT (P1/P1 and P3/P3 transport options), though.

john-science:
> I know there aren't added tests here yet. I'd like some input on that before I start writing a test.

In the _ideal_ scenario, when you fix a bug you add a unit test that used to fail, but doesn't any more.

If that scenario requires just TOO MUCH input data to run, it would be nice to have unit tests that test each little/tiny method you changed to show that they do what you want them to do now.

As always, feel free to contact me about testing.

Nebbychadnezzar:
Testing?! What's that?

I wanted some input before I started writing tests, since this at first appeared to be a bit of an odd update. However, after looking a bit through the `test_nhflux.py` file, I think I can whip something up that would be serviceable. Hang tight while I do that and add some commits.

ntouran (approved):
Looks good to me. Would it be possible to add a small dummy `NHFLUX_VARIANT` file that we could use in a new unit test to ensure this works as expected? I think the `NHFLUX_VARIANT` file format must be changing a big vs. version because we've *definitely* used this NHFLUX data with variant in the past.

Nebbychadnezzar:
Sounds good, @ntouran. I will include this information in the tests.

Nebbychadnezzar:
A new test class has been added in`test_nhflux` to cover at least some of the aspects of the new code that's been added. Adding something that has nMoms > 0 would require executing P3 with anisotropic cross sections, which is not currently part of the test case used with `test_nhflux`. Running something with iwnhfl == 1 would require running VARPOW, which in turn would require running a coupled n/g job.

john-science (approved):

PR #685, opened 2022-05-25
Assorted cleanup and bug fixes

john-science:
## Description

This PR addresses a few sundry issues;

* The MANIFEST.in was not updated for the next detailed axial expansion test data files.
* The change to `cleanPath()` has been reverted to the old defaults of "remove file from all processes" for the core ARMI use cases.
* Fixed a bug in runLog where `time` wasn't imported.
* Removed a bunch of useless `pytlint` warnings.
* Adding code coverage to FuelHandlers

---

## Checklist


Nebbychadnezzar:
Solid use of sundry.

armi/utils/directoryChangers.py
```diff
@@ -260,7 +260,7 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_value, traceback):
         DirectoryChanger.__exit__(self, exc_type, exc_value, traceback)
         try:
-            pathTools.cleanPath(self.destination)
+            pathTools.cleanPath(self.destination, context.MPI_RANK)
```
keckler, line 263: @john-science 
This change seems to be causing me issues. In particular, during _some_ multi group XS generation runs, my code starts to hang after exiting the temporary directory. 

My code was working before this change, and if I instead use the regular `DirectoryChanger` in place of this, then my code continues as intended.

This appears to be very closely related to @mgjarrett 's fixes in #689. But I honestly have no idea what is the reason why the code hangs...

PR #688, opened 2022-05-27
Address some potential MPI bugs

mgjarrett:
## Description

These two functions that either made MPI blocking communication calls or updated the MPI state had `return` statements that could potentially cause a process to exit the function with MPI in an unintended state. This code update attempts to avoid any race conditions / hanging on an MPI `barrier()`.

---

## Checklist

If user exposed functionality was added/changed: N/A

### DIFF
--- a/armi/mpiActions.py
+++ b/armi/mpiActions.py
@@ -300,6 +300,7 @@ def runActions(o, r, cs, actions, numPerNode=None, serial=False):
     call MPI operations without being blocked by tasks which are not doing the same thing.
     """
     if not context.MPI_DISTRIBUTABLE or serial:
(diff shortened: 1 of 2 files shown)


armi/mpiActions.py
```diff
@@ -439,6 +440,8 @@ def invokeHook(self):
             context.MPI_SIZE = mpiSize
             context.MPI_NODENAMES = mpiNodeNames
 
+        return actionResult
```
john-science, line 443: I'm not super pumped about adding code and not testing it.

But both of the changes here _are_ in the MPI/parallel space, and I know it's harder to write useful tests here. Hmm

john-science (approved):

PR #689, opened 2022-05-28
Fix MPI hanging bug.

mgjarrett:
## Description

Making the fix from #688 more robust. There are some situations where `MPI_SIZE` can be greater than 1, but `MPI_DISTRIBUTABLE` is `False` and only the rank 0 process has called `cleanPath()`. This code prevents the code from getting to an infinite MPI barrier in that situation.

---

## Checklist

If user exposed functionality was added/changed: N/A

### DIFF
--- a/armi/utils/pathTools.py
+++ b/armi/utils/pathTools.py
@@ -240,7 +240,7 @@ def cleanPath(path, mpiRank=0):
     """
     valid = False
     if not os.path.exists(path):
-        if context.MPI_SIZE > 1:
+        if context.MPI_SIZE > 1 and context.MPI_DISTRIBUTABLE:
             context.MPI_COMM.barrier()
         return True
 
@@ -283,7 +283,7 @@ def cleanPath(path, mpiRank=0):
         sleep(waitTime)
 
     # Potentially, wait for all the processes to catch up.
-    if context.MPI_SIZE > 1:
+    if context.MPI_SIZE > 1 and context.MPI_DISTRIBUTABLE:
         context.MPI_COMM.barrier()
 
     if os.path.exists(path):
--- a/armi/utils/pathTools.py
+++ b/armi/utils/pathTools.py
@@ -219,7 +219,7 @@ def moduleAndAttributeExist(pathAttr):
     return moduleAttributeName in userSpecifiedModule.__dict__
 
 
-def cleanPath(path, mpiRank=0):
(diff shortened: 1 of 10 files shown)


armi/utils/pathTools.py
```diff
@@ -240,7 +240,7 @@ def cleanPath(path, mpiRank=0):
     """
     valid = False
     if not os.path.exists(path):
-        if context.MPI_SIZE > 1:
+        if context.MPI_SIZE > 1 and context.MPI_DISTRIBUTABLE and barrier:
```
john-science, line 243: For computational efficiency, can we write this as:

> if barrier and context.MPI_SIZE > 1 and context.MPI_DISTRIBUTABLE:

The idea here is there is no real cost to checking the value of `barrier`, but there is an incurred cost to pulling up the `context` object and checking the namespace inside it.

This is not super important, but it's also easy to fix.

armi/utils/pathTools.py
```diff
@@ -283,7 +283,7 @@ def cleanPath(path, mpiRank=0):
         sleep(waitTime)
 
     # Potentially, wait for all the processes to catch up.
-    if context.MPI_SIZE > 1:
+    if context.MPI_SIZE > 1 and context.MPI_DISTRIBUTABLE and barrier:
```
john-science, line 286: Same as my other comment on the ordering of this multi-boolean check.

mgjarrett:
I moved the MPI barrier outside of `cleanPath`, and I adjusted the logic for calling the barrier slightly. I think this is the right logic to call MPI barrier only when processes are running in parallel.

mgjarrett:
It's difficult to meaningfully cover many of the places where `cleanPath()` and `context.waitAll()` are called because they happen when cleaning up folders from ARMI runs, but those folders don't get generated when running the unit tests.

mgjarrett:
Still having trouble with downstream applications... don't merge this yet. I'm stuck in a position where the code that passes unit test fails downstream in-use tests, and vice versa.

john-science:
> Still having trouble with downstream applications... don't merge this yet. I'm stuck in a position where the code that passes unit test fails downstream in-use tests, and vice versa.

Tell me when you're good, man.

opotowsky:
@jstilley I just chatted with @mgjarrett and he says it's ready!

Issue #691, opened 2022-05-31
with default nCycles (nCycles=1) database writes at cycle 0 are suppressed in snapshots


onufer:
The setting `nCycles` has very little meaning for a snapshots run, so many users delete the line which reverts to default value of 1. this suppress the database write for snapshots operator because it thinks it is at EOL during this if statement when cycle = 0
https://github.com/terrapower/armi/blob/f6bf598525efaac72f0391296af832b556d415ad/armi/bookkeeping/db/database3.py#L264-L268

Also, the snapshots operator does some big backflips to get ARMI to write to the database at just the right times. See exclusions/logic in code below
https://github.com/terrapower/armi/blob/f6bf598525efaac72f0391296af832b556d415ad/armi/operators/snapshots.py#L64-L80

It might be more clear/obvious to just disable the database interface during snapshots and explicitly call a to-be-created "writeToDb" public method on databaseInterface.

john-science:
@ntouran 

Tell me if I'm wrong.

But since `class OperatorSnapshots(operatorMPI.OperatorMPI):` and `class OperatorMPI(Operator):`, could we just override the `def atEOL` method to fix this problem for him?

```python

class OperatorSnapshots(operatorMPI.OperatorMPI):
    # ...
    @property
    def atEOL(self):
        return self.r.p.cycle == self.cs["nCycles"]     # ditch the "- 1"
```

It looks like we would want to delete the first 2 of these 3 lines as well:

https://github.com/terrapower/armi/blob/f6bf598525efaac72f0391296af832b556d415ad/armi/operators/snapshots.py#L78-L80

Though... I'm sure those were there for a reason, I'm not sure I see it.

ntouran:
We should definitely override `atEOL` for the snapshot operator. The proposal from John is fine, but is till dependent on the setting always being in the default time. Maybe it'd be better to just always return `False` for snapshots?

PR #692, opened 2022-06-01
Fixing Snapshot DB issue

john-science:
## Description

This PR is in reference to #691 

The bug should be addressed, though building a proper unit test for the feature has proved tricky.

---

## Checklist


armi/cases/case.py
```diff
@@ -428,6 +428,7 @@ def initializeOperator(self, r=None):
         with DirectoryChanger(self.cs.inputDirectory, dumpOnException=False):
             self._initBurnChain()
             o = operators.factory(self.cs)
+            print("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", o)
```
onufer, line 431: probably delete.

armi/operators/snapshots.py
```diff
@@ -90,3 +87,8 @@ def setStateToDefault(cs):
         from armi.operators.runTypes import RunTypes
 
         return cs.modified(newSettings={"runType": RunTypes.STANDARD})
+
+    @property
+    def atEOL(self):
+        """NOTE: This operator's atEOL method behaves very differently than other operators."""
```
onufer, line 93: you could add some context that snapshots don't really have an EOL since they can go in any order and may or may not contain the last time node form the load database. It kind of sets the context.

onufer (approved):

PR #697, opened 2022-06-06
Allow jumpRingTo to be equal to 1

opotowsky:
## Description

Fix `ValueError` check for `jumpRingTo` to allow it to be equal to 1 (since the default for convergent shuffle is in fact 1). 

---

## Checklist


opotowsky:
@john-science the convergent scenario this fix applies to wasn't previously explicitly tested. I'm not sure if it's important to do that?

PR #699, opened 2022-06-06
Remove pinPowers parameter from block module

Nebbychadnezzar:
## Description

This pull request removes the `pinPowers` parameter from the framework. The history of this parameter is not clear, but after some investigating, I have concluded that this parameter is either not used anymore and has been replaced by `linPowByPin`.

`setPinPowers()` now only makes reference to `linPowByPin`.

I also judged that the `gamma` parameter for the method is redundant, since it can be derived from the `powerKeySuffix` parameter, so I removed that.

I did not see any testing for `linPowByPin`, so I also modified the `test_blocks.py` module to replace `pinPowers` testing with `linPowByPin` testing.

Finally, I have added a TODO in the method because there is a section of its behavior that is unclear to me. I am unable to make a decision of what to do about it on my own, so I decided to leave it alone and just add a note. A future pull request could address the TODO.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -1951,7 +1951,6 @@ class HexBlock(Block):
         return self._getPinCoordinatesHex()
 
     def _getPinCoordinatesHex(self):
-        """"""
         pinPitch = self.getPinPitch()
         if pinPitch is None:
             return []
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -1574,11 +1574,7 @@ class HexBlock(Block):
         nPins = self.getNumPins()
         self.p.pinLocation = list(range(1, nPins + 1))
 
-    def setPinPowers(
-        self,
-        powers,
-        powerKeySuffix="",
-    ):
+    def setPinPowers(self, powers, powerKeySuffix=""):
         """
         Updates the pin linear power densities of this block for the current rotation.
         The linear densities are represented by the *linPowByPin* parameter.
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -1585,56 +1585,67 @@ class HexBlock(Block):
         powerKeySuffix="",
     ):
         """
-        Updates the pin powers of this block for the current rotation.
+        Updates the pin linear power densities of this block for the current rotation.
+        The linear densities are represented by the *linPowByPin* parameter.
 
         Parameters
         ----------
-        powers : list of floats
-            The block-level pin linear power densities. pinPowers[i] represents the average linear
-            power density of pin i.
-            Power units are Watts/cm (Watts produced per cm of pin length).
-            The "ARMI pin ordering" is used, which is counter-clockwise from 3 o'clock.
+        powers : list of floats, required
+            The block-level pin linear power densities. powers[i] represents the average
+            linear power density of pin i. The units of linear power density is watts/cm
+            (i.e., watts produced per cm of pin length). The "ARMI pin ordering" must be
+            be used, which is counter-clockwise from 3 o'clock.
+
+        numPins : int, required
+            Number of pins in the block. This parameter should probably be removed because it
+            can be derived from the block.
+
+        imax: int, required
+            Number of pin rings. This parameter should probably be removed because it
+            can be derived from the block.
+
+        jmax: list of ints, required
+            Number of pins per ring. Element 0 is the number of pins in ring 1, element 1
+            is the number of pins in ring 2, and so on. This parameter should probably be
+            removed because it can be derived from the block.
+
+        gamma: bool, optional
+            Currently unused
+
+        removeSixCornerPins: bool, optional.
+            Defaults to False
+
+        powerKeySuffix: str, optional
+            Must be either an empty string, :py:const:`NEUTRON <armi.physics.neutronics.const.NEUTRON>`,
+            or :py:const:`GAMMA <armi.physics.neutronics.const.GAMMA>`. Defaults to empty
+            string.
 
         Notes
         -----
-        This handles rotations using the pinLocation parameters.
-
-        This sets:
-
-        self.p.pinPowers : 1-D numpy array
-            The block-level pin linear power densities. pinPowers[i] represents the average linear
-            power density of pin i.
-            Power units are Watts/cm (Watts produced per cm of pin length).
-            The "ARMI pin ordering" is used, which is counter-clockwise from 3 o'clock.
+        This method can handle assembly rotations by using the *pinLocation* parameter.
         """
-        self.p.pinPowers = numpy.zeros(numPins)
         self.p["linPowByPin" + powerKeySuffix] = numpy.zeros(numPins)
+
+        # Loop through rings
         j0 = jmax[imax - 1] / 6
         pinNum = 0
-        for i in range(imax):  # loop through rings
-            for j in range(jmax[i]):  # loop through positions in ring i
+        for i in range(imax):
+            # Loop through positions in ring i
+            for j in range(jmax[i]):
                 if removeSixCornerPins and i == imax - 1 and math.fmod(j, j0) == 0.0:
                     linPow = 0.0
                 else:
+                    # TODO: This appears to need fixing to account for blocks in fueled
+                    # TODO: assemblies that contain elevations with pins but no fuel,
+                    # TODO: such as for an axial shield.
                     if self.hasFlags(Flags.FUEL):
-                        # -1 to map from pinLocations to list index
+                        # -1 is needed in order to map from pinLocations to list index
                         pinLoc = self.p.pinLocation[pinNum] - 1
                     else:
                         pinLoc = pinNum
                     linPow = powers[pinLoc]
-                self.p.pinPowers[pinNum] = linPow
                 self.p["linPowByPin" + powerKeySuffix][pinNum] = linPow
                 pinNum += 1
-
-        if powerKeySuffix == GAMMA:
-            self.p.pinPowersGamma = self.p.pinPowers
-        elif powerKeySuffix == NEUTRON:
-            self.p.pinPowersNeutron = self.p.pinPowers
-
-        if gamma:
-            self.p.pinPowers = self.p.pinPowersNeutron + self.p.pinPowersGamma
-        else:
-            self.p.pinPowers = self.p.pinPowersNeutron
 
     def rotate(self, deg):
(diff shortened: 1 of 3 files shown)


Nebbychadnezzar:
I still need to look into why tests appear to be failing...

armi/reactor/blocks.py
```diff
@@ -1580,61 +1580,80 @@ def setPinPowers(
...
+            # Loop through positions in ring i
+            for j in range(jmax[i]):
                 if removeSixCornerPins and i == imax - 1 and math.fmod(j, j0) == 0.0:
                     linPow = 0.0
                 else:
+                    # TODO: This appears to need fixing to account for blocks in fueled
+                    # TODO: assemblies that contain elevations with pins but no fuel,
+                    # TODO: such as for an axial shield. These blocks should still have
+                    # TODO: powers that are eligible for rotation.
                     if self.hasFlags(Flags.FUEL):
```
john-science, line 1639: Generally, if a PR comes with a `TODO`, that means it needs a little more work until it's ready for prime-time.

"fueled assembles that contain elevations with pins but no fuel"

@ntouran  Do you understand why this if-statement exists?

Nebbychadnezzar: I added it so that it could be discussed as part of the PR. The if/else doesn't make sense to me.

Nebbychadnezzar: In doing a line blame, it looks like this if/else has been here for at least 3 years...I don't know how to go beyond that time period because it looks like the file was refactored from somewhere else, so that commit shows one gigantic add operation to the file.

Again, my confusion here is that the line says "if you are fuel, then do a rotation lookup before getting your pin number; otherwise, don't do a lookup." I don't understand why you would want to do that ;why does it matter whether you are fuel or not?

Nebbychadnezzar:
> I still need to look into why tests appear to be failing...

Found the error, so now the tests are passing.

Nebbychadnezzar:
@mgjarrett , I am submitting the pinPowers removal for a PR and saying it should be just be linPowByPin. Wanted to make sure you didn't have objections or concerns about this.

armi/reactor/blocks.py
```diff
@@ -1580,61 +1580,80 @@ def setPinPowers(
...
-            self.p.pinPowersGamma = self.p.pinPowers
-        elif powerKeySuffix == NEUTRON:
-            self.p.pinPowersNeutron = self.p.pinPowers
-
-        if gamma:
-            self.p.pinPowers = self.p.pinPowersNeutron + self.p.pinPowersGamma
-        else:
-            self.p.pinPowers = self.p.pinPowersNeutron
+        # If using the *powerKeySuffix* parameter, we also need to set total power. If
+        # you were setting gamma power, the logic below assumes that the neutron powers
```
mgjarrett, line 1649: This logic assumes that `setPinPowers()` will always be called in order: first with `powerKeySuffix == NEUTRON`, then with `powerKeySuffix == GAMMA`. I think it would be more robust to implement in a way that will give you the correct total power regardless of the order in which `setPinPowers()` is called. Maybe we could just check for the existence of each type of power and add it to the total if it exists?

Nebbychadnezzar: Agreed; the existing code assumes that you'll do the operations sequentially. I guess the real problem is with with gamma. The code will abort at least, since it's trying to add None + np.ndarray().

if powersuffix == "":
   # assume pure neutron
   total = neutron values
   neutron = neutron values
   gamma = None (should we set to 0.0?...probably not, since we'd may break others depending on seeing None.
elif powersuffix == NEUTRON:
   # assume pure neutron:
   same as powersuffix = ""
else:
   # coupled
   gamma = gamma values
   assert neutron values are present
   total = neutron + gamma

mgjarrett: I think that makes sense

armi/reactor/blocks.py
mgjarrett: I'm also confused here; I can't think of a reason why only the fuel blocks should be rotated. I think this if/else is unnecessary and/or wrong.

armi/reactor/blocks.py
Nebbychadnezzar: Both the module and test has been updated to address Mike's suggestion.

mgjarrett:
Everything looks good, thanks @Nebbychadnezzar

armi/reactor/tests/test_blocks.py
```diff
@@ -16,30 +16,21 @@
 # pylint: disable=missing-function-docstring,missing-class-docstring,abstract-method,protected-access,no-member,invalid-name,consider-using-f-string
 import copy
 import math
-import os
-import unittest
 import numpy
 from numpy.testing import assert_allclose
+import os
```
john-science, line 21: I believe these imports were moved in error.  Have you seen our docs on the ordering of `import` statements?

https://terrapower.github.io/armi/developer/standards_and_practices.html#import-ordering

armi/reactor/blocks.py
```diff
@@ -1580,61 +1580,85 @@ def setPinPowers(
...
+        # is sum of neutron and gamma powers. We assume that a solo gamma calculation
+        # to set total power does not make sense.
+        if powerKeySuffix:
+            if powerKeySuffix == GAMMA:
+                if self.p[f"linPowByPin{NEUTRON}"] is None:
+                    msg = (
+                        "Neutron power has not been set yet. Cannot set total power for "
+                        f"{self}."
+                    )
+                    raise UnboundLocalError(msg)
```
john-science, line 1658: I believe this is why your code coverage went down for this PR.

You added a new special exception, but you aren't testing for it in a unit test.

armi/reactor/tests/test_blocks.py
Nebbychadnezzar: Completed.

armi/reactor/blocks.py
```diff
@@ -1580,61 +1580,85 @@ def setPinPowers(
...
+            linear power density of pin i. The units of linear power density is watts/cm
+            (i.e., watts produced per cm of pin length). The "ARMI pin ordering" must be
+            be used, which is counter-clockwise from 3 o'clock.
 
-        Notes
-        -----
-        This handles rotations using the pinLocation parameters.
+        numPins : int, required
+            Number of pins in the block. This parameter should probably be removed because it
+            can be derived from the block.
```
jakehader, line 1600: Let's just remove this now and replace it with `self.getNumPins()` within the method.

armi/reactor/blocks.py
```diff
@@ -1580,61 +1580,85 @@ def setPinPowers(
...
-            power density of pin i.
-            Power units are Watts/cm (Watts produced per cm of pin length).
-            The "ARMI pin ordering" is used, which is counter-clockwise from 3 o'clock.
+        jmax: list of ints, required
+            Number of pins per ring. Element 0 is the number of pins in ring 1, element 1
+            is the number of pins in ring 2, and so on. This parameter should probably be
+            removed because it can be derived from the block.
+
+        removeSixCornerPins: bool, optional.
+            Defaults to False
```
jakehader, line 1612: This is silly. We should just have the block design describe where the pins are. This should be removed.

armi/reactor/blocks.py
```diff
@@ -1580,61 +1580,85 @@ def setPinPowers(
...
-        -----
-        This handles rotations using the pinLocation parameters.
+        numPins : int, required
+            Number of pins in the block. This parameter should probably be removed because it
+            can be derived from the block.
 
-        This sets:
+        imax: int, required
+            Number of pin rings. This parameter should probably be removed because it
+            can be derived from the block.
```
jakehader, line 1604: imax and jmax can be completely removed and derived using grids.py now.

See: `_getPinCoordinatesHex` for an example of how the hexagon locations for i, j are determined based on the pin pitch and the number of pinned components within the block. This would be better to use the grids system since it is newer than this code now.

armi/reactor/blocks.py
jakehader: I think that you're right that this is probably wrong. The `blocks` module has `PIN_COMPONENTS` defined as a module-level variable. You might be able to see any children of the block have flags that match these and if so then the block contains "pins". 

Honestly, I think that the concept of "pins" needs to be standardized because I can see that there are places in the code that assume if "mult" > 1 then it's a pin and then there's other places that now use this `PIN_COMPONENTS` list. This seems sketchy and fragile.

@ntouran - See question above by @john-science please. I don't understand this either.

jakehader (changes requested):

Nebbychadnezzar:
The signature of this method has been substantially changed as of today. We will have to make updates to downstream plugins if this PR is approved.

john-science:
@Nebbychadnezzar  We definitely look at the conversations to see if the PR is ready.  For instance, if I look at the PR and there are conversations left open/unresolved, I assume people are still talking about the best way to fix some line of code, and I don't Approve yet.

Just a thought, if you were looking to move this PR along.

Looks promising!

Nebbychadnezzar:
Updated documentation for `setPinPowers` to address Nick's comments.

armi/reactor/blocks.py
```diff
@@ -1574,67 +1574,61 @@ def initializePinLocations(self):
...
+        for pinNum in range(numPins):
+            # TODO: This appears to need fixing to account for blocks in fueled
+            # TODO: assemblies that contain elevations with pins but no fuel,
+            # TODO: such as for an axial shield. These blocks should still have
+            # TODO: powers that are eligible for rotation.
+            if self.hasFlags(Flags.FUEL):
+                # -1 is needed in order to map from pinLocations to list index
+                pinLoc = self.p.pinLocation[pinNum] - 1
+            else:
+                pinLoc = pinNum
```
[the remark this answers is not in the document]
john-science: Important info to have in the PR, but not the _fault_ of this PR.

armi/reactor/blocks.py
```diff
@@ -1952,12 +1951,15 @@ def getPinCoordinates(self):
         return self._getPinCoordinatesHex()
 
     def _getPinCoordinatesHex(self):
-        coordinates = []
-        numPins = self.getNumPins()
-        numPinRings = hexagon.numRingsToHoldNumCells(numPins)
+        """"""
```
john-science, line 1954: Can we remove this empty docstring thing?

Thanks!

john-science (approved):
I have one little thing about remove an empty docstring (or filling it in), but other than that I'm happy.

PR #702, opened 2022-06-07
Allow components to have children in support of particle fuel modeling

ntouran:
## Description

This changes the ARMI data model to allow components to have children. It also adds a input
capability to allow users to input particle type fuel (e.g. TRISO).

This is a proof-of-concept at the moment, intended for getting feedback from @drewj-usnctech 
wrt #477 . 

Mostly, take a look at the input file and let us know if it satisfies the need. We will need
to add more testing and documentation to verify that everything is working 
and explained. 

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/componentBlueprint.py
+++ b/armi/reactor/blueprints/componentBlueprint.py
@@ -180,13 +180,13 @@ class ComponentBlueprint(yamlize.Object):
             if self.blends:
                 # build a blended object
                 for groupName, blendFrac in zip(self.blends, self.blendFracs):
-                    # blendFrac is a volume fraction, and so we need to adjust the multiplicities 
+                    # blendFrac is a volume fraction, and so we need to adjust the multiplicities
                     # so that the background component and the child group match up
                     # The area of the background component will be fully determined by its dims and mult.
                     # but internally, its number densities and volumes need to be set to match
                     # the blendFrac
                     group = blueprint.componentGroups[groupName]
-                    # strip off the blend/blendFrac args since they aren't valid on any 
+                    # strip off the blend/blendFrac args since they aren't valid on any
                     # specific shape's constructor
                     del kwargs["blends"]
                     del kwargs["blendFracs"]
@@ -194,10 +194,14 @@ class ComponentBlueprint(yamlize.Object):
                     constructedObject = components.factory(shape, [], kwargs)
                     # build the child objects
                     for groupedComponent in group:
-                        componentDesign = blueprint.componentDesigns[groupedComponent.name]
+                        componentDesign = blueprint.componentDesigns[
+                            groupedComponent.name
+                        ]
                         component = componentDesign.construct(blueprint, matMods=dict())
                         # override free component multiplicity if it's set based on the group definition
-                        component.setDimension("mult", groupedComponent.mult*blendFrac)
+                        component.setDimension(
+                            "mult", groupedComponent.mult * blendFrac
+                        )
                         _setComponentFlags(component, self.flags, blueprint)
                         insertDepletableNuclideKeys(component, blueprint)
                         constructedObject.add(component)
--- a/armi/reactor/blueprints/componentBlueprint.py
+++ b/armi/reactor/blueprints/componentBlueprint.py
@@ -198,10 +198,10 @@ class ComponentBlueprint(yamlize.Object):
                             groupedComponent.name
                         ]
                         component = componentDesign.construct(blueprint, matMods=dict())
-                        # override free component multiplicity if it's set based on the group definition
-                        component.setDimension(
-                            "mult", groupedComponent.mult * blendFrac
-                        )
+                        # temporarily set grouped component mults to the blend fraction
+                        # these will be updated based on parent component volume
+                        # during block construction
+                        component.setDimension("mult", blendFrac)
                         _setComponentFlags(component, self.flags, blueprint)
                         insertDepletableNuclideKeys(component, blueprint)
                         constructedObject.add(component)
--- a/armi/reactor/blueprints/componentBlueprint.py
+++ b/armi/reactor/blueprints/componentBlueprint.py
@@ -199,7 +199,7 @@ class ComponentBlueprint(yamlize.Object):
                         # override free component multiplicity if it's set based on the group definition
                         component.setDimension("mult", groupedComponent.mult*blendFrac)
                         _setComponentFlags(component, self.flags, blueprint)
-                        _insertDepletableNuclideKeys(component, blueprint)
+                        insertDepletableNuclideKeys(component, blueprint)
                         constructedObject.add(component)
                     children = {c.name: c for c in constructedObject.getChildren()}
                     for child in children.values():
--- a/armi/reactor/blueprints/componentBlueprint.py
+++ b/armi/reactor/blueprints/componentBlueprint.py
@@ -193,10 +193,18 @@ class ComponentBlueprint(yamlize.Object):
                     # build the background object
                     constructedObject = components.factory(shape, [], kwargs)
                     # build the child objects
+                    # all grouped components have to be input with the same mult for now
+                    inputMult = None
                     for groupedComponent in group:
                         componentDesign = blueprint.componentDesigns[
                             groupedComponent.name
                         ]
+                        if inputMult is None:
+                            inputMult = componentDesign.mult 
+                        if componentDesign.mult != inputMult:
+                            raise ValueError(
+                                f"Grouped components must all have the same input mult of {inputMult}"
+                            )
                         component = componentDesign.construct(blueprint, matMods=dict())
                         # temporarily set grouped component mults to the blend fraction
                         # these will be updated based on parent component volume
--- a/armi/reactor/blueprints/componentBlueprint.py
+++ b/armi/reactor/blueprints/componentBlueprint.py
@@ -153,6 +153,8 @@ class ComponentBlueprint(yamlize.Object):
     isotopics = yamlize.Attribute(type=str, default=None)
     latticeIDs = yamlize.Attribute(type=list, default=None)
     origin = yamlize.Attribute(type=list, default=None)
+    blends = yamlize.Attribute(type=list, default=None)
+    blendFracs = yamlize.Attribute(type=list, default=None)
     orientation = yamlize.Attribute(type=str, default=None)
     mergeWith = yamlize.Attribute(type=str, default=None)
     area = yamlize.Attribute(type=float, default=None)
@@ -175,9 +177,37 @@ class ComponentBlueprint(yamlize.Object):
                 constructedObject.add(component)
 
         else:
-            constructedObject = components.factory(shape, [], kwargs)
-            _setComponentFlags(constructedObject, self.flags, blueprint)
-            insertDepletableNuclideKeys(constructedObject, blueprint)
+            if self.blends:
+                # build a blended object
+                for groupName, blendFrac in zip(self.blends, self.blendFracs):
+                    # blendFrac is a volume fraction, and so we need to adjust the multiplicities 
+                    # so that the background component and the child group match up
+                    # The area of the background component will be fully determined by its dims and mult.
+                    # but internally, its number densities and volumes need to be set to match
+                    # the blendFrac
+                    group = blueprint.componentGroups[groupName]
+                    # strip off the blend/blendFrac args since they aren't valid on any 
+                    # specific shape's constructor
+                    del kwargs["blends"]
+                    del kwargs["blendFracs"]
+                    # build the background object
+                    constructedObject = components.factory(shape, [], kwargs)
+                    # build the child objects
+                    for groupedComponent in group:
+                        componentDesign = blueprint.componentDesigns[groupedComponent.name]
+                        component = componentDesign.construct(blueprint, matMods=dict())
+                        # override free component multiplicity if it's set based on the group definition
+                        component.setDimension("mult", groupedComponent.mult*blendFrac)
+                        _setComponentFlags(component, self.flags, blueprint)
+                        _insertDepletableNuclideKeys(component, blueprint)
+                        constructedObject.add(component)
+                    children = {c.name: c for c in constructedObject.getChildren()}
+                    for child in children.values():
+                        child.resolveLinkedDims(children)
+            else:
+                constructedObject = components.factory(shape, [], kwargs)
+                _setComponentFlags(constructedObject, self.flags, blueprint)
+                insertDepletableNuclideKeys(constructedObject, blueprint)
         return constructedObject
 
     def _conformKwargs(self, blueprint, matMods):
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -626,14 +626,16 @@ class Component(composites.Composite, metaclass=ComponentType):
         """
         
         if self._children:
-            childDens = composites.Composite.getNuclideNumberDensities(self, nucNames)
+            childDens = dict(zip(nucNames, composites.Composite.getNuclideNumberDensities(self, nucNames)))
+            # get volume of children by using the composite method which loops over all children
             childVol = composites.Composite.getVolume(self)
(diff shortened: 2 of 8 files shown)


keckler:
@ntouran 

I don't have a good enough grasp of these changes to know the answer, but does this change account for this comment you had made about the material modifications?

https://github.com/terrapower/armi/pull/517#pullrequestreview-841033091

And a further question -- Do we want to allow different material modifications at this additional layer as well? Sounds messy...

armi/tests/refTriso-bp.yaml
```diff
@@ -0,0 +1,145 @@
...
+      kernel:
+        mult: 1000
+      buffer:
+        mult: 1000
+      ipyc:
+        mult: 1000
+      sic:
+        mult: 1000
+      opyc:
+        mult: 1000
```
drewj-usnctech, line 93: Should the multiplicity here be 1 for each component? In that there is one `kernel` in a `triso particle` but maybe thousands of `triso particle` components in the `triso compact` component with the blend.

armi/reactor/components/component.py
```diff
@@ -582,30 +582,42 @@ def getNuclides(self):
...
+            totalVol = self.getVolume()
+            childFrac = childVol / totalVol
+            selfFrac = 1.0 - childFrac
+            return [
+                self.p.numberDensities.get(nucName, 0.0) * selfFrac
+                + childDens.get(nucName, 0.0) * childFrac
+                for nucName in nucNames
+            ]
+        else:
+            return [self.p.numberDensities.get(nucName, 0.0) for nucName in nucNames]
```
drewj-usnctech, line 620: How would a plugin be able to get the number densities of just the matrix component with this change?

On one hand this is consistent with how the design of the composite tree: `Composite.getNuclideNumberDensities` gives the densities for all children on a node. So this would match well with `Block.getNuclideNumberDensities` for some nodal homogenization methods. But modeling this `Component` exactly in a monte carlo code would be problematic w/o the number densities of just the `Component` and not its children

armi/reactor/components/component.py
```diff
@@ -582,30 +582,42 @@ def getNuclides(self):
 
         This includes anything that has been specified in here, including trace nuclides.
         """
-        return list(self.p.numberDensities.keys())
+        nucs = set(self.p.numberDensities.keys())
+        if self._children:
+            childNuclides = set(composites.Composite.getNuclides(self))
+            nucs |= childNuclides
+        return nucs
```
drewj-usnctech, line 589: Same concern on getting nuclides just in the component vs. component and all the children

drewj-usnctech (commented):
@ntouran thanks for putting this together. The input file changes look good. 

Maybe a hyperbolic situation, but if I had a `Component` that had two blended `component groups`, how could I separate out the children? Or maybe a better way to put it: how can plugins know that all (or some) of the children on a component belong together in the same particle fuel `component group`?

I don't have great answers for these but hopefully the questions are helpful.

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -175,9 +177,37 @@ def construct(self, blueprint, matMods):
...
+                        componentDesign = blueprint.componentDesigns[groupedComponent.name]
+                        component = componentDesign.construct(blueprint, matMods=dict())
+                        # override free component multiplicity if it's set based on the group definition
+                        component.setDimension("mult", groupedComponent.mult*blendFrac)
+                        _setComponentFlags(component, self.flags, blueprint)
+                        insertDepletableNuclideKeys(component, blueprint)
+                        constructedObject.add(component)
+                    children = {c.name: c for c in constructedObject.getChildren()}
+                    for child in children.values():
+                        child.resolveLinkedDims(children)
```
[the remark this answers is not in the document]
ntouran: yeah it should. though we should make more unit tests proving it.

armi/reactor/components/component.py
ntouran: Great point. You would have to access the matrix's `.p.numberDensities` parameter directly in order to differentiate between it and its children with this change as written. Willing to consider more elegant options.

armi/reactor/components/component.py
ntouran: yeah that's a good concern. I'd hate to make you basically re-write these methods for accessing the details of the matrix itself. We could add in a flag or something as a method argument? `getNuclides(includeChildren=False)` or something?

armi/reactor/components/component.py
```diff
@@ -975,8 +987,20 @@ def mergeNuclidesInto(self, compToMergeWith):
...
+        -----
+        This could probably be made recursive to allow arbitrary depth, but it's more complex
+        """
+        if self._children:
+            for child in self._children:
+                if self.hasFlags(typeSpec, exact):
+                    yield child
+        else:
+            if self.hasFlags(typeSpec, exact):
+                yield self
```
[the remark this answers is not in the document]
ntouran: very nice, that does seem better. I'll try your suggestion.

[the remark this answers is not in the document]
ntouran: ok so I tried With that suggested iterComponents change, and my `test_trisoCase.py` unit test fails with max recursion errors. If you iterate over yourself and your children you will just keep iterating over yourself... 

I tried this kind of thing for a while and just couldn't quite get it working. I'm sure it can be done... it's just hard. 

```  File "/home/nick/repos/armi/armi/reactor/composites.py", line 1256, in <listcomp>
    [c.getNumberDensity(nuc) for nuc in nucNames]
  File "/home/nick/repos/armi/armi/reactor/composites.py", line 1256, in <listcomp>
    [c.getNumberDensity(nuc) for nuc in nucNames]
  File "/home/nick/repos/armi/armi/reactor/composites.py", line 1233, in getNumberDensity
    return self.getNuclideNumberDensities([nucName])[0]
  File "/home/nick/repos/armi/armi/reactor/components/component.py", line 605, in getNuclideNumberDensities
    composites.Composite.getNuclideNumberDensities(self, nucNames),
  File "/home/nick/repos/armi/armi/reactor/composites.py", line 1255, in getNuclideNumberDensities
    [
  File "/home/nick/repos/armi/armi/reactor/composites.py", line 1256, in <listcomp>
    [c.getNumberDensity(nuc) for nuc in nucNames]
  File "/home/nick/repos/armi/armi/reactor/composites.py", line 1256, in <listcomp>
    [c.getNumberDensity(nuc) for nuc in nucNames]
  File "/home/nick/repos/armi/armi/reactor/composites.py", line 1233, in getNumberDensity
    return self.getNuclideNumberDensities([nucName])[0]
  File "/home/nick/repos/armi/armi/reactor/components/component.py", line 605, in getNuclideNumberDensities
    composites.Composite.getNuclideNumberDensities(self, nucNames),
  File "/home/nick/repos/armi/armi/reactor/composites.py", line 1244, in getNuclideNumberDensities
    [
  File "/home/nick/repos/armi/armi/reactor/composites.py", line 1245, in <listcomp>
    c.getVolume() / (c.parent.getSymmetryFactor() if c.parent else 1.0)
  File "/home/nick/repos/armi/armi/reactor/components/component.py", line 430, in getVolume
    if self.p.volume is None:
  File "/home/nick/repos/armi/armi/reactor/parameters/parameterDefinitions.py", line 293, in __get__
    return self._getter(obj)
RecursionError: maximum recursion depth exceeded
```

john-science:
Bump.

Needs `black` formatting.  (I'd do it for your, but it's in your repo fork.)

armi/reactor/components/component.py
drewj-usnctech: `includeChildren=False` feels like a good solution. Then whatever application or plugin is doing the writing could check if a `Component` has children prior to writing the material definitions

drewj-usnctech:
@ntouran I feel decently good about the internals. My biggest request comes from not requiring the user to input number of particles per compact in the blueprints file. There are some occasions where you might know the exact number of particles in a compact, especially as a design matures. But for more preliminary studies, or where you have compacts of multiple sizes but the same volume fraction, requiring the user to know the number of particles at the start can be burdensome

ntouran:
@drewj-usnctech ok I have made some small changes that now auto-derive the mult completely based on the blendFrac. So you put in dummy mults of 1.0 and the system updates the kernel mults during init to match the expected blend frac. 

One weird thing here is that the blendfrac is applied to the hot dimensions at init, (but will stay constant after init even if temperatures change)

drewj-usnctech:
@ntouran thanks for the update. I've been out for a bit and am finally catching up. 

> One weird thing here is that the blendfrac is applied to the hot dimensions at init

This was my suspicion as well. We've been having some internal discussion about how thermal expansion impacts our particle fuel modeling and guess this might be a problem. I imagine this component expansion is done very early in the blueprint loading and there isn't a way to delay it? Or wholesale disable it?

armi/reactor/components/component.py
drewj-usnctech: @ntouran is

> matrix's .p.numberDensities

still the recommended approach here? I've been testing this out over the last few days and had a hard time remembering this thread.

Would something like `Component.getNuclideNumberDensities(children=False)` be acceptable? I tend to be hesitant pulling data from these parameter containers as the feel (maybe incorrectly) more private than a public method

jakehader:
> @ntouran thanks for the update. I've been out for a bit and am finally catching up.
> 
> > One weird thing here is that the blendfrac is applied to the hot dimensions at init
> 
> This was my suspicion as well. We've been having some internal discussion about how thermal expansion impacts our particle fuel modeling and guess this might be a problem. I imagine this component expansion is done very early in the blueprint loading and there isn't a way to delay it? Or wholesale disable it?

You can disable thermal expansion by setting `Thot` to be the same as `Tinput` in the blueprints!

drewj-usnctech:
@jakehader 

> You can disable thermal expansion by setting `Thot` to be the same as `Tinput` in the blueprints!

🤦 yeah, I forgot about that when writing that comment. Didn't mean to bring a larger discussion (#766) into this feature.

But there is still hiccup *if* you decide to enable thermal expansion and model particle fuel here. As @ntouran [mentioned](https://github.com/terrapower/armi/pull/702#issuecomment-1175402641)

> One weird thing here is that the blendfrac is applied to the hot dimensions at init

This means the number of particles packed into the compact would reflect the post-expansion dimension rather than the as built dimension. There would be an opaque change in fuel loading due to this thermal expansion.

FWIW, as the main advocate for this feature, I'm okay saying this pre-expansion should not be a blocker for this feature yet. We're still trying to sort out thermal expansion and the various properties and knock-on effects internally, and are probably a ways from fully diving in. Hence my posting of questions like #766 and #694

drewj-usnctech:
@john-science This has been up a while and we've started some internal discussions about adjacent modifications to the composite tree. I don't have anything I can provide at this point, but I'll hopefully be reaching out to the dev team in a week or so about how a larger enhancement. This could capture both the particle fuel modeling and provide some additional modeling benefits. We think. It's still in a conceptual phase and we'd like to give it some more time before sending it off

drewj-usnctech:
(fully aware that I am somewhat responsible for the delay)

drewj-usnctech:
This is on our task list for things to do this week and report back if we find any thing that seriously blocks / breaks our internal plugins.

One thing that I'm thinking about, while I'm thinking about composite tree stuff (#503 // #960), does this break the notion that a `Component` is a leaf in the composite tree now? It was kind of the implied consequence of `Component.iter_children` needing to be updated to yield itself but also any potential particle fuel children. Could this be the motivating case to introduce a non-block group of components?

armi/reactor/components/component.py
```diff
@@ -1019,8 +1031,20 @@ def mergeNuclidesInto(self, compToMergeWith):
...
```
drewj-usnctech, line 1047: @ntouran we've found that the mass of the matrix material is not accounted for when doing `getMass` calls. I think due to this section here

We modified the triso mass test blueprint file such that each element only appears in one `Component`. Then, `block.getMass("C12")` (in theory) picks up mass from the `triso compact` component and should be non-zero. Unfortunately this is not the case 😬 

```yaml
# A full blueprint file that contains assems made with ComponentGroups
blocks:
    fuel2: &block_fuel2
        triso compact: 
            shape: Circle
            blends: 
              - triso particle
            blendFracs: 
              - 0.70
            material: Graphite
            Tinput: 600.0
            Thot: 600.0
            mult: 1
            id: 0.0
            od: 1.2446
        duct: 
            shape: Hexagon
            material: Cu
            Tinput: 600.0
            Thot: 600.0
            ip: 9.0
            mult: 1.0
            op: 10.0
        coolant: 
            shape: DerivedShape
            material: NaCl
            Tinput: 25.0
            Thot: 600.0

components:
    # triso dims from Table 4 in TRISO benchmark
    kernel:
        shape: Sphere
        material: Lead
        Tinput: 600.0
        Thot: 600.0
        id: 0.0
        mult: 1.0
        od: 0.010625
    buffer:
        shape: Sphere
        material: Lead
        Tinput: 600.0
        Thot: 600.0
        id: kernel.od
        mult: 1.0
        od: 0.015625
    ipyc:
        shape: Sphere
        material: Lead
        Tinput: 600.0
        Thot: 600.0
        id: buffer.od
        mult: 1.0
        od: 0.017375
    sic:
        shape: Sphere
        material: Lead
        Tinput: 600.0
        Thot: 600.0
        id: ipyc.od
        mult: 1.0
        od: 0.019125
    opyc:
        shape: Sphere
        material: Lead
        Tinput: 600.0
        Thot: 600.0
        id: sic.od
        mult: 1.0
        od: 0.021125

component groups:
    triso particle:
      kernel:
        mult: 1
      buffer:
        mult: 1
      ipyc:
        mult: 1
      sic:
        mult: 1
      opyc:
        mult: 1


assemblies:
    fuel c: &assembly_c
        specifier: OC
        blocks: [*block_fuel2]
        height: [1.0]
        axial mesh points: [1]
        xs types: [A]
systems:
    core:
        grid name: core
        origin:
            x: 0.0
            y: 0.0
            z: 0.0
grids:
    core:
        geom: hex
        symmetry: full
        lattice map: |
          OC


nuclide flags:
    B: &nuc_flags {burn: false, xs: true}
    C: *nuc_flags
    PB: *nuc_flags
    CU63: *nuc_flags
    CU65: *nuc_flags
    CL35: *nuc_flags
    CL37: *nuc_flags
    CL37: *nuc_flags
    NA23: *nuc_flags
```

ntouran: Thanks @drewj-usnctech I will bring in your modified blueprints file and update the unit test so it fails as you've found and then try to fix.

ntouran:
Yeah so I think I need to re-implement this to follow a more robust version of the [Composite design pattern](https://en.wikipedia.org/wiki/Composite_pattern). This will take a bit more work, sorry, but should be more robust. 

I have been adding to the unit tests and splitting them out to cover a few expected behaviors. We probably will want to add even more. They currently fail, as they should. 

Current plan to get them running is to:

* Go back to having Components be the **Leaf** of the composite pattern and not have children (sorry I we have the names goofed compared to the offical pattern)
* Have `Composite`s only be made of children. So composite pattern participants will either be completely made of children or be a leaf. Having things that have their own component and then also have child components is not a good idea, turns out.
* Re-implement this change to make triso compact `Composite` have children be:
  * background `Component` (matrix) (with mult reduced to make room for the particles based on the blend frac)
  * particle `Composite` with children:
    * Various spherical particle components with mults set based on blend frac

I think the net result will be better support of arbitrary nesting.

drewj-usnctech:
@ntouran sounds like a good plan. Let me know how we can help, either in developing or providing tests / requirements.

One question

> with mult reduced to make room for the particles based on the blend frac

Are you thinking about having `Matrix.mult` be less than one to account for the space taken up by particles? Would ARMI allow a non-integer `mult`?

keckler:
> Are you thinking about having `Matrix.mult` be less than one to account for the space taken up by particles? Would ARMI allow a non-integer `mult`?

ARMI handles non-integer mults. See https://github.com/terrapower/armi/discussions/542

jakehader:
I guess we should close this pull request? @john-science, @ntouran?

PR #707, opened 2022-06-11
update getPinToDuctGap to ignore holedHex

onufer:
## Description

The current code cannot handle it since it has no `ip`.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -1849,7 +1849,7 @@ class HexBlock(Block):
                 # getPinCenterFlatToFlat only works for hexes
                 # inner most duct might be circle or some other shape
                 duct = None
-            elif not isinstance(duct, components.HoledHexagon):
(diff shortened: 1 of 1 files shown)


armi/reactor/blocks.py
```diff
@@ -1849,6 +1849,10 @@ def getPinToDuctGap(self, cold=False):
                 # getPinCenterFlatToFlat only works for hexes
                 # inner most duct might be circle or some other shape
                 duct = None
+            elif not isinstance(duct, components.HoledHexagon):
```
jakehader, line 1852: This looks like it won't ignore the holed hexagon still. Maybe this should be `elif isinstance(...)`?

jakehader (approved):

john-science:
A recent PR to the docs broke the build a little.  So I will either:

a. Wait until the morning to merge this PR, so that can be fixed first.

or 

b. Just merge this one, and as long as everything BUT the docs builds, we're fine.

PR #713, opened 2022-06-14
Uniform mesh converter update

mgjarrett:
## Description

The existing uniform mesh converter will fail if it needed to convert multiple `FUEL` or `CONTROL` blocks with different `xsTypes` into a single block.

It would be an improvement to handle this situation gracefully, since it is not an unreasonable condition. Here are a couple of examples:

    A reactor with multiple axial enrichments with different XS in the different enrichment zones.
    A perturbation calculation where the perturbed state has different XS in a specific axial region of the core.

This PR handles the situation by using the XS type from the source block that represents the largest fraction of the destination block in the uniform mesh. I think this is a very reasonable assumption but I'm open to discussion here.

---

## Checklist


mgjarrett:
The existing uniform mesh converter will fail if it needed to convert multiple `FUEL` or `CONTROL` blocks with different `xsType`s into a single block.

It would be an improvement to handle this situation gracefully, since it is not an unreasonable condition. Here are a couple of examples:
1. A reactor with multiple axial enrichments with different XS in the different enrichment zones. 
2. A perturbation calculation where the perturbed state has different XS in a specific axial region of the core.

This PR handles the situation by using the XS type from the source block that represents the largest fraction of the destination block in the uniform mesh. I think this is a very reasonable assumption but I'm open to discussion here.

armi/reactor/converters/uniformMesh.py
```diff
@@ -186,33 +186,45 @@ def makeAssemWithUniformMesh(sourceAssem, newMesh):
...
+                for b in blocks:
+                    if b.hasFlags([Flags.FUEL, Flags.CONTROL]):
+                        if b.p.xsType == xsType:
+                            sourceBlock = b
+                            break
+
+                if sourceBlock:
+                    totalHeight = sum(typeHeight.values())
+                    runLog.extra(
+                        "Multiple XS types exist between {} and {}".format(
```
jakehader, line 215: Just as a note, it might be good to start using f-strings throughout the code base because it can be easier to read and know which parameters are going where. Example: `print(f"This value is: {someVar}")`

No change needed, just a comment.

jakehader, line 215: Just as a note, it might be good to start using f-strings throughout the code base because it can be easier to read and know which parameters are going where. Example: `print(f"This value is: {someVar}")`

No change needed, just a comment.

armi/reactor/converters/uniformMesh.py
```diff
@@ -244,6 +256,11 @@ def _buildAllUniformAssemblies(self):
             newAssem = self.makeAssemWithUniformMesh(sourceAssem, self._uniformMesh)
             src = sourceAssem.spatialLocator
             newLoc = self.convReactor.core.spatialGrid[src.i, src.j, 0]
+            runLog.debug(
```
jakehader, line 259: Is this debug statement needed/useful for any particular reason?

jakehader, line 259: Is this debug statement needed/useful for any particular reason?

albeanth:
@mgjarrett this is a solid improvement to the uniform mesh converter. Nice man!

armi/reactor/converters/uniformMesh.py
mgjarrett: No, I was using it but we probably shouldn't keep it around. There were some cases where it was taking ~20 minutes to convert a third core reactor to full core, and this was the only log message that would let me know the code was still running and not frozen during that period. But it does generate a lot of output and is probably not necessary.

armi/reactor/converters/uniformMesh.py
mgjarrett: I changed it when I was addressing the other comments

armi/reactor/converters/uniformMesh.py
jakehader: That's a long time. Is that still happening? Maybe the loop is inefficient if so

armi/reactor/converters/tests/test_uniformMesh.py
```diff
@@ -79,6 +79,26 @@ def test_blueprintCopy(self):
...
+
+        for newB, sourceB in zip(newAssem.getBlocks(), sourceAssem.getBlocks()):
+            self.assertEqual(
+                newB.p["xsType"],
+                sourceB.p["xsType"],
+            )
+            self.assertEqual(
+                newB.p["xsType"],
+                sourceB.p["xsType"],
+            )
```
albeanth, line 100: These assertions are the same?

albeanth: Also, since the test reactor in `armi/tests/refSmallReactorBase.yaml` is already axially uniform, I'd be curious to know how this test works for the test reactor in `armi/tests/detailedAxialExpansion`.

armi/reactor/converters/uniformMesh.py
mgjarrett: In a recent run, some were taking 4 minutes and others about 17 min:
```
> grep "Reactor core conversion" *.stdout
[xtra] Reactor core conversion time: 248.86255460200005 seconds
[xtra] Reactor core conversion time: 244.63209826000002 seconds
[xtra] Reactor core conversion time: 247.14147353800013 seconds
[xtra] Reactor core conversion time: 246.28612149000037 seconds
[xtra] Reactor core conversion time: 818.0751615429999 seconds
[xtra] Reactor core conversion time: 1006.6690055039999 seconds
[xtra] Reactor core conversion time: 1036.225465257 seconds
[xtra] Reactor core conversion time: 1056.5681029470006 seconds
[xtra] Reactor core conversion time: 1057.0175631559996 seconds
[xtra] Reactor core conversion time: 1062.2484797609995 seconds
[xtra] Reactor core conversion time: 1067.229568883 seconds
[xtra] Reactor core conversion time: 1043.9269293489997 seconds
[xtra] Reactor core conversion time: 1061.6862967769994 seconds
[xtra] Reactor core conversion time: 1000.8848847240006 seconds
[xtra] Reactor core conversion time: 1016.5796172129994 seconds
[xtra] Reactor core conversion time: 1052.6184727269992 seconds
```

armi/reactor/converters/tests/test_uniformMesh.py
mgjarrett: @albeanth the `setUpClass` method for `TestUniformMeshComponents` perturbs the mesh of one of the assemblies.

mgjarrett:
@albeanth I added a unit test with the `detailedAxialExpansion` reactor. It doesn't directly test the new capability, because that reactor doesn't generate any situation where two fuel blocks with different `xsType` overlap the same uniform mesh block. But it's helpful to test that the code handles a non-uniform mesh without any trouble.

mgjarrett:
I think this is good to merge, assuming everybody is satisfied. I believe I've addressed @albeanth's comments and others, but somebody speak up if not.

albeanth:
> It doesn't directly test the new capability, because that reactor doesn't generate any situation where two fuel blocks with different xsType overlap the same uniform mesh block.

Doesn't this line accomplish this? Also, you could totally change the detailedAxialExpansion test reactor blueprints to include that situation if you wanted to (assuming it doesn't break other tests that use it). https://github.com/mgjarrett/armi/blob/3f5961f504771974a74f84e397bce47959c6f49d/armi/reactor/converters/tests/test_uniformMesh.py#L56

Also, we can remove the commented out code if we aren't going to use it. https://github.com/mgjarrett/armi/blob/3f5961f504771974a74f84e397bce47959c6f49d/armi/reactor/converters/tests/test_uniformMesh.py#L57

mgjarrett:
Okay, I updated the detailedAxialExpansion test reactor to have overlapping blocks. This test covers the new code now, so I removed the test that used the regular test reactor.

armi/reactor/converters/uniformMesh.py
john-science: So, this change to the code makes it takes 20 minutes to convert a 1/3 core reactor to a full core?

But it normally only takes seconds, right?

If I understand the situation correctly, we can't let this PR be merged until we have fixed this issue.  People would be very upset with us.

mgjarrett: The reactor core conversion was taking 20 minutes before I started this PR. I have no idea when or why that started happening, but it's not from this code change.

jakehader: Hm. When I profiled the core it the major slow downs were due to setting the number densities on the blocks because we reach down the composite chain to derive volumes each time and these aren't being cached. We used to have caching but that was removed because it caused a ton of issues. I think that if we could speed up the set number densities call then this would speed up significantly @john-science.

albeanth: IMO, since this slowdown predates this PR, I vote we make a new issue to address the number density calculation and get this PR merged. Just my two cents.

jakehader (approved):

armi/reactor/converters/uniformMesh.py
jakehader: I agree - Just reviewed and approved. Ill make a general ticket on number density speed ups

jakehader: https://github.com/terrapower/armi/issues/721

john-science: Thanks all. I didn't realize quite the extent of how long this was taking (before this PR).

I can't help but think that is a solvable problem.

albeanth: > Thanks all. I didn't realize quite the extent of how long this was taking (before this PR).
> 
> I can't help but think that is a solvable problem.

100%.

mgjarrett:
@john-science this PR is ready for review

john-science (approved):

PR #714, opened 2022-06-14
Add a warning about excessive plotting

keckler:
## Description

Self-explanatory.

---

## Checklist

### DIFF
--- a/armi/utils/plotting.py
+++ b/armi/utils/plotting.py
@@ -12,6 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
+This module makes heavy use of matplotlib. Beware that plots generated with matplotlib
+may not free their memory, even after the plot is closed, and excessive use of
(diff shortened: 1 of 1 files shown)


john-science (approved):

armi/utils/plotting.py
```diff
@@ -12,6 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
+This module makes heavy use of matplotlib. Beware that plots generated with matplotlib
```
jakehader, line 16: You can do something like this, which will render as a big warning box on the docs:

```
 .. warning:: 
    This is a warning message
```

See: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Admonitions.html#warning

jakehader: Maybe `.. caution::` could be good instead

Issue #690, opened 2022-05-30
`updateGlobalAssemblyNum` called twice during `loadState`


keckler:
`loadState` has an input parameter that is supposed to allow you to choose whether to call `updateGlobalAssemblyNum` or not. That parameter is called `updateGlobalAssemNum`:
https://github.com/terrapower/armi/blob/f6bf598525efaac72f0391296af832b556d415ad/armi/bookkeeping/db/database3.py#L398-L445

Towards the end of that method there is a check as to whether `updateGlobalAssemNum` is `True` or `False`, and `updateGlobalAssemblyNum` is called depending on the check.

However, further up in the method, `loadDB.load` is called, which itself calls to `updateGlobalAssemblyNum` without regards to the parameter that was passed into `loadState`:
https://github.com/terrapower/armi/blob/f6bf598525efaac72f0391296af832b556d415ad/armi/bookkeeping/db/database3.py#L1082-L1147

So I'm pretty sure it is being called duplicate.

I believe this was introduced in #615 @john-science .

keckler:
I'll take this.

PR #717, opened 2022-06-15
`updateGlobalAssemblyNum` called twice during `loadState`

keckler:
## Description

In response to #690 .

The duplicate call has been removed and another input option `updateGlobalAssemNum` has been added to `loadState` so that it can be passed along to `load`. This way both calls are consistent and `updateGlobalAssemblyNum` doesn't need to be called within `loadState` itself.

The default value of `updateGlobalAssemNum` is `True` so that the previous behavior is maintained by default.

Tests have been added as well.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -423,7 +423,9 @@ class DatabaseInterface(interfaces.Interface):
                         cs=self.cs,
                         bp=self.r.blueprints,
                         allowMissing=True,
+                        updateGlobalAssemNum=updateGlobalAssemNum,
                     )
+                    self.o.reattach(newR, self.cs)
                     break
         else:
             # reactor was never set so fail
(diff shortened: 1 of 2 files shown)


armi/bookkeeping/db/database3.py
```diff
@@ -439,11 +441,6 @@ def loadState(
                 )
             )
 
-        if updateGlobalAssemNum:
-            updateGlobalAssemblyNum(newR)
-
-        self.o.reattach(newR, self.cs)
```
john-science, line 445: Why did you remove the `self.o.reattach()`?

Just curious.

keckler: I didn't remove it, I just moved it closer to where it is actually relevant.

The `else` clause of that conditional only raises errors. So having the `reattach` wait until after the `else` didn't seem necessary, and I just thought it helped with code clarity to move the call into the `if` clause.

john-science (approved):

PR #719, opened 2022-06-16
Plotting should really be in the utils folder

john-science:
## Description

As we discussed a couple months ago, all our plotting tools should probably be in the "utils" folder. This PR has just been sitting idol on my laptop for ages, but since we've been talking about plotting so much this week it occurred to me to break it out.

---

## Checklist


john-science:
To the best of my knowledge, only one downstream project is affected by this move.

That repo will only need a one-line change, so I will make that change myself.

armi/utils/plotting.py
```diff
@@ -1423,3 +1417,147 @@ def plotBlockDiagram(block, fName, cold, cmapName="RdYlBu", materialList=None):
...
+        Won't finalize plot. Useful for using this to make custom plots.
+
+    Examples
+    --------
+    >>> l = ISOTXS()
+    >>> l.plotNucXs('U238NA','fission')
+
+    >>> # Plot n,g for all xenon and krypton isotopes
+    >>> f = lambda name: 'XE' in name or 'KR' in name
+    >>> l.plotNucXs(sorted(filter(f,l.nuclides.keys())),itertools.repeat('nGamma'))
```
keckler, line 1532: I think you need to update these examples to reflect that this is no longer a method, but is not a function. i.e. you need to pass in ISOTXS explicitly.

armi/bookkeeping/report/reportInterface.py
```diff
@@ -120,11 +120,11 @@ def interactEOL(self):
         b = self.o.r.core.getFirstBlock(Flags.FUEL)
         b.setAreaFractionsReport()
 
-        from armi.bookkeeping import plotting
+        from armi.utils import reportPlotting
```
keckler, line 123: Does this import need to be inside the method due to circular import issues? I can't tell right away.

keckler (approved):
Small comments.

armi/nuclearDataIO/cccc/rtflux.py
```diff
@@ -207,94 +202,3 @@ def getFDFluxReader(adjointFlag):
...
-    plt.ylabel("cm")
-    plt.xlabel("cm")
-    ax.autoscale_view()
-    plt.savefig("RTFLUX-z" + str(axialZ + 1) + "-g" + str(energyGroup + 1) + imgFileExt)
-    plt.close()
-
-
-if __name__ == "__main__":
-    rtflux = RtfluxStream.readBinary("RTFLUX")
-    plotTriangleFlux(rtflux, axialZ=10, energyGroup=4)
```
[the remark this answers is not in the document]
john-science: Yeah, I meant to get rid of it.

This type of thing doesn't really fit well into the `import armi` idea, and I'm trying to politely remove it from the repo (long term).

armi/utils/plotting.py
john-science: Ah! You are totally right. I will look and see if I did this elsewhere to.  Good catch.

armi/bookkeeping/report/reportInterface.py
john-science: Yeah, that WAS to prevent a circular import. But I can fix it now.

Issue #569, opened 2022-02-09
Allow users to provide user/case-specific user plugins as input


ntouran:
There are situations where users may want to provide case-specific material properties, component shapes, etc. that are typically provided by plugins. Thus, it'd be nice if users could point to a plugin that resides in an input file using a new setting (e.g. `userPluginPath`). 

Since it's related to the overall repeatability of the run, the new file should be hashed in the stdout and copied into the output h5 file for safe keeping. 

This will also close #535

Sidenote: our pre-existing custom Python input related to the `shuffleLogic` and `controlLogic`  settings could likely be merged into this capability at a later time to make this an overall cleanup.

I'll take a stab at implementing this.

ntouran:
Hmm, well maybe. See [comment here](https://github.com/terrapower/armi/issues/535#issuecomment-1033305726)

jakehader:
I like the thoughts on the other ticket for the materials, but do we still want this for the other plugins too (fuel management/control rod logic)?

jakehader:
After talking it sounds like we just need a one line change in the Case to add the working directory of the case it `sys.path` and the custom materials will work. Maybe we can add an example case of this somewhere in the gallery too?

jakehader:
@ntouran are you able to make a PR for this?

ntouran:
Yeah so after more discussion it's clear that this kind of thing would still be useful for some purposes, including doing rapid prototyping of test code that does things at special hooks. Effectively, this would allow people to use the plugin system as a convenient pathway for monkey patching during development/testing. I think as long as we put a strong indication near the ARMI version information that user plugins are in use, then we can maintain knowledge in the stdout about what code was actually run. Obviously, app-endorsed plugins explicitly imported and registered by the application should generally be used for actual qualified/traceable runs and analysis.

ntouran:
K so in the PR linked above I fixed this with environmental variables. Using user settings would be perhaps nicer from a UI perspective, but is challenging from an architectural point of view because plugins must be loaded and locked down before settings can be parsed (since settings are defined by plugins...)

sombrereau:
@keckler -- FYI this might help you configure a reactor without shuffling logic or control logic

keckler:
Yeah, I agree, thanks @sombrereau . _If_ it ever happens 😆

john-science:
> Yeah, I agree, thanks @sombrereau . _If_ it ever happens laughing

I mean, it happened like a month ago.

PR #723, opened 2022-06-17
Adding new User Plugin feature

john-science:
## Description

In response to #569 , this PR attempts to implement a quick-and-easy variation on our current `Plugin` functionality. These new `UserPlugin`s can be added ad-hoc, for more one-off study analysis. For instance, you might add this to your settings YAML file:

```yaml
  userPlugins::
    - armi.tests.test_user_plugins.UserPluginOnProcessCoreLoading
```

And this would mean that during a run, your custom `UserPluginOnProcessCoreLoading` would be added to the plugin list (from the test file `armi/test/test_user_plugins.py`):

```python
class UserPluginOnProcessCoreLoading(plugins.UserPlugin):
    """
    This plugin flex-tests the onProcessCoreLoading() hook,
    and arbitrarily adds "1" to the height of every block,
    after the DB is loaded.
    """

    @staticmethod
    @HOOKSPEC
    def onProcessCoreLoading(core, cs):
        blocks = core.getBlocks(Flags.FUEL)
        for b in blocks:
            b.p.height += 1.0
```

This PR is currently a draft because the full range and power of all `Plugin` and `Interface` tooling is technically supported by this new `UserPlugin`. But I feel like I haven't done nearly enough testing to prove that ALL of that works. (More testing is needed.)

**UPDATE**: This PR also now support plugins that are outside ARMI, and can be anywhere defined by a full file path:

```yaml
  userPlugins::
    - //path/to/my/pluginz.py:AwesomeUserPlugin
```

---

## Checklist


armi/settings/caseSettings.py
```diff
@@ -272,9 +272,20 @@ def loadFromInputFile(self, fName, handleInvalids=True, setPath=True):
...
         reader.readFromFile(fName, handleInvalids)
         self._applyReadSettings(path if setPath else None)
+        self.registerUserPlugins()
 
         return reader
 
+    def registerUserPlugins(self):
+        """TODO"""
+        userPlugins = self.__settings.get("userPlugins", [])
+        userPlugins = self["userPlugins"]
```
onufer, line 282: can you add a comment here describing what is going on. Its surprising to set `userPlugins` 2x but in different ways

armi/settings/fwSettings/globalSettings.py
```diff
@@ -832,6 +833,14 @@ def defineSettings() -> List[setting.Setting]:
                 ]
             ),
         ),
+        setting.Setting(
+            CONF_USER_PLUGINS,
+            default=[],
+            label="User Plugins",
+            description="An optional place for users to define on-the-fly plugins "
+            "for things like one-off analysis.",
```
onufer, line 841: add example of how this should be defined. like if my plugin lived in 
c:\mystuff\here\myPlugin.py
how does it expect it to be defined

armi/plugins.py
```diff
@@ -600,10 +600,72 @@ def defineSystemBuilders() -> Dict[str, Callable[[str], "Composite"]]:
...
+        a run, and as such, we want to limit how flexible they are, so we can correctly
+        corral their side effects during a run.
+        """
+        if issubclass(self.__class__, UserPlugin):
+            assert len(self.__class__.defineSettings()) == 0, "TODO"
+            # NOTE: These are the class methods that we are staunchly _not_ allowing people
+            # to change in this class. If you need these, please use a regular ArmiPlugin.
+            self.defineParameterRenames = lambda: None
+            self.defineSettings = lambda: []
+            self.defineSettingsValidators = lambda: []
```
onufer, line 634: can we raise errors if set rather than suppress silently?

onufer (changes requested):

armi/settings/caseSettings.py
onufer: also there is still todo in doc string, but i imagine you planned to update that.

armi/plugins.py
john-science: Well, a few lines up we DO raise an error if `defineSettings()` has anything in it.

But, from a strictly technological perspective, I'm not sure _how_ to detect code in the other two functions. You see, function content is mutable in Python (as you see here), so there is no guarantee that you can correctly determine if a method is empty.  (This is particularly annoying because "docstrings" are actually Python string objects, so technically any method with a docstring is not empty.

I can look at this again, but I don't know how to test these other two and be 100% sure I was correct.

john-science:
There was a side-request made here:

> Can we add a place in `Plugin` or `Interface` that executes code after a DB is read?

Right now `Plugin.onProcessCoreLoading()` does something like this, but after this is executed, the Reactor load state is called, which will stomp all over any changes you make to the Reactor. Ditto for `afterConstructionOfAssemblies()`.

armi/settings/fwSettings/globalSettings.py
john-science: There are examples in the unit tests and at the top of this PR. 

I could put an example in the tutorial, sure.

armi/settings/caseSettings.py
john-science: > can you add a comment here describing what is going on. Its surprising to set `userPlugins` 2x but in different ways

Ah, that's a typo.

john-science:
> TODO: John.  Is the setting "UserPlugins", "User Plugins", or "userPlugins".

I believe I support two of these three, which is ridiculous.

armi/plugins.py
onufer: They all tend to evaluate to lists or dictionaries so I was thinking you could do it very simialry to how you dothe same as you do settings

assert  not self.__class__.defineSettings(), "defined settings are not supported for user plugins"
assert  not self.__class__.defineParameterRenames(), "param renames are not supported for user plugins"
ssert  not self.__class__.defineSettingsValidators(),  "settings validators are not supported for user plugins"

john-science:
Feature Request:  

> Be able to read a Python file based on any absolute file path, not just inside the repo.

There is [precedent](https://github.com/terrapower/armi/blob/93d1946010726c88c0bbe01e7a9b6c4cd85da7f4/armi/physics/fuelCycle/fuelHandlerFactory.py#L31-L85) for this.

john-science:
> Feature Request:
> 
> > Be able to read a Python file based on any absolute file path, not just inside the repo.

@onufer  Hurray!  We now support `UserPlugin`s that are outside of ARMI, somewhere you need to provide an absolute path for.

Sorry it took me so long to get back to your request, I got sick and had a couple of surprise days off of work.

armi/settings/fwSettings/globalSettings.py
john-science: I can put some description of how to set the file paths in their, sure.

john-science: Okay, I did both:

1. I improved this description, and
2. I added a basic guide to the docs.

armi/apps.py
```diff
@@ -221,6 +229,76 @@ def getParamRenames(self) -> Dict[str, str]:
...
+                # The path is of the form: armi.thing.what.MyPlugin
+                self.__registerUserPluginsInternalImport(pluginPath)
+
+    def __registerUserPluginsAbsPath(self, pluginPath):
+        """Helper method to register a single UserPlugin where
+        the given path is of the form: /path/to/why.py:MyPlugin
+        """
+        # determine if we have that Windows file path (C:\\path\to\whatever.py)
+        isWindows = False
+        if ":\\" in pluginPath:
```
john-science, line 270: **TODO**: JOHN!

I just found out that most Windows machines use the format: `C:\whatever`, but it's only the OLD versions of Windows that use `C:\\whatever`.

So, apparently, to support Windows, you need to support both. So, joy.

john-science: Fixed

armi/settings/fwSettings/globalSettings.py
```diff
@@ -109,7 +109,7 @@
 CONF_BLOCK_AUTO_GRID = "autoGenerateBlockGrids"
 CONF_INPUT_HEIGHTS_HOT = "inputHeightsConsideredHot"
 CONF_CYCLES = "cycles"
-CONF_USER_PLUGINS = "userPlugins"
+CONF_USER_PLUGINS = "UserPlugins"
```
onufer, line 112: other settings are lower camel case, why is user plugin different?

armi/plugins.py
onufer: that makes sense and defineSettings is the big bang for buck. if you cant define settings people will easily notice they cant define validators.

armi/settings/fwSettings/globalSettings.py
john-science: Fixed!

armi/apps.py
```diff
@@ -223,6 +231,76 @@ def getParamRenames(self) -> Dict[str, str]:
             self._paramRenames = renames, self._pm.counter
         return renames
 
+    def registerUserPlugins(self, pluginPaths):
+        """
+        Register additional plugins passed in by importable paths.
```
ntouran, line 236: docstring should have blank line after short summary line https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

armi/apps.py
```diff
@@ -223,6 +231,76 @@ def getParamRenames(self) -> Dict[str, str]:
...
+        These plugins may be provided e.g. by an application during startup
+        based on user input.
+        Format expected to be a list of full namespaces to plugin classes.
+        There should be a comma between individual plugins and dots representing
+        the importable python namespace. e.g.
+
+        ``armi.stuff.plugindir.pluginMod.pluginCls,armi.whatever.plugMod2.plugCls2``.
```
ntouran, line 243: could use Example clause here

https://numpydoc.readthedocs.io/en/latest/format.html#examples

armi/apps.py
```diff
@@ -223,6 +231,76 @@ def getParamRenames(self) -> Dict[str, str]:
...
+        or on Linux/Unix:
+        ``/path/to/pluginMod.py:pluginCls,/path/to/plugMod2.py:plugCls2```,
+        or on Windows:
+        ``C:\\path\\to\\pluginMod.py:pluginCls,C:\\\\path\\to\\plugMod2.py:plugCls2```,
```
ntouran, line 247: are there too many `\\\\`'s in the 2nd path?

armi/apps.py
```diff
@@ -223,6 +231,76 @@ def getParamRenames(self) -> Dict[str, str]:
...
+                self.__registerUserPluginsInternalImport(pluginPath)
+
+    def __registerUserPluginsAbsPath(self, pluginPath):
+        """Helper method to register a single UserPlugin where
+        the given path is of the form: /path/to/why.py:MyPlugin
+        """
+        # determine if we have that Windows file path (C:\path\to\whatever.py)
+        isWindows = False
+        if ":\\" in pluginPath:
+            isWindows = True
```
ntouran, line 273: would Pathlib be useful here? https://docs.python.org/3.10/library/pathlib.html

armi/apps.py
```diff
@@ -223,6 +231,76 @@ def getParamRenames(self) -> Dict[str, str]:
...
+            else:
+                # The path is of the form: armi.thing.what.MyPlugin
+                self.__registerUserPluginsInternalImport(pluginPath)
+
+    def __registerUserPluginsAbsPath(self, pluginPath):
+        """Helper method to register a single UserPlugin where
+        the given path is of the form: /path/to/why.py:MyPlugin
+        """
+        # determine if we have that Windows file path (C:\path\to\whatever.py)
+        isWindows = False
```
ntouran, line 271: could check `sys.platform`? or I guess you're checking input for windows format rather than the run platform

armi/plugins.py
```diff
@@ -600,10 +600,77 @@ def defineSystemBuilders() -> Dict[str, Callable[[str], "Composite"]]:
         """
 
 
-def getNewPluginManager() -> pluginManager.ArmiPluginManager:
+class UserPlugin(ArmiPlugin):
     """
-    Return a new plugin manager with all of the hookspecs pre-registered.
+    This is a variation on the usual ArmiPlugin that is meant to be read in during run time,
```
ntouran, line 605: change to be short summary 1-liner, blank line, more deets

armi/plugins.py
```diff
@@ -600,10 +600,77 @@ def defineSystemBuilders() -> Dict[str, Callable[[str], "Composite"]]:
...
+    from a hard-coded path in a settings file, or the like.
+
+    This is obviously a more limited use-case than the usual ArmiPlugin, as those are meant
+    to be defined at import time, instead of run time. As such, this class has some built-in
+    tooling to limit how these run-time plugins are used. They are meant to be more limited.
+
+    NOTE: The usual ArmiPlugin is much more flexible, if the UserPlugin does not support what
+    you want to do, just use an ArmiPlugin.
```
ntouran, line 613: prefer `Notes` section or `.. note::` directive for these kinds of callouts

armi/plugins.py
```diff
@@ -600,10 +600,77 @@ def defineSystemBuilders() -> Dict[str, Callable[[str], "Composite"]]:
...
     """
+
+    def __init__(self, *args, **kwargs):
+        ArmiPlugin.__init__(self, *args, **kwargs)
+        self.__enforceLimitations()
+
+    def __enforceLimitations(self):
+        """
+        This method is designed to enforce that UserPlugins are more limited
```
ntouran, line 622: make docstring 1 liner. 

(we should consider turning on more docstring linting)

armi/plugins.py
```diff
@@ -600,10 +600,77 @@ def defineSystemBuilders() -> Dict[str, Callable[[str], "Composite"]]:
...
+        a run, and as such, we want to limit how flexible they are, so we can correctly
+        corral their side effects during a run.
+        """
+        if issubclass(self.__class__, UserPlugin):
+            assert (
+                len(self.__class__.defineParameterRenames()) == 0
+            ), "UserPlugins cannot define parameter renames"
+            assert (
+                len(self.__class__.defineSettings()) == 0
+            ), "UserPlugins cannot define new Settings"
```
ntouran, line 634: consider adding an actionable suggestion in these message, like "...consider using an ArmiPlugin instead"

armi/plugins.py
```diff
@@ -600,10 +600,77 @@ def defineSystemBuilders() -> Dict[str, Callable[[str], "Composite"]]:
...
+            self.defineSettings = lambda: []
+            self.defineSettingsValidators = lambda: []
+
+    @staticmethod
+    @HOOKSPEC
+    def defineParameterRenames():
+        """
+        NOTE: This is not overridable.
+        It is a design limitation of user plugins that they not generate parameter renames,
+        so that they are able to be added to the plugin stack during run time.
```
ntouran, line 647: update docstring format. similar comments below

```
Prevent parameter renames.

.. warning:: This is not overridable.

Notes
-----
It is a design limitation of user plugins that they not generate parameter renames,
so that they are able to be added to the plugin stack during run time.
```

ntouran (changes requested):
very schnazzy. thanks  for the good docs and release notes updates. 

Most comments are just about docstrings

onufer (approved):
this looks good to me. nick had some comments that should probably be addressed

armi/apps.py
john-science: Python uses `\` as an escape character for strings.

So, in order to build a Python string for a Windows file path, you need to handle the escape character situation.

armi/apps.py
john-science: Done!

armi/apps.py
john-science: Done!

armi/apps.py
john-science: Good call.  I switch to `os.name`.

Done.

armi/apps.py
john-science: Done

armi/plugins.py
john-science: Done

armi/plugins.py
john-science: Done

armi/plugins.py
john-science: Done

armi/plugins.py
john-science: Done

ntouran (approved):
cool thanks. 🚢

PR #727, opened 2022-06-20
Fix centering in the Cartesian test case.

ntouran:
The center for refSmallCartesian should be at 0,0,0.  As it was, the
Fuel management in a LWR gallery example had negative burnups, which,
though manufactured as a demo, were confusing. This this change, the
gallery example is again a typical full core case.

Sidenote: first commit from an airplane?

[Before fix](https://terrapower.github.io/armi/gallery/framework/run_fuelManagement.html#sphx-glr-gallery-framework-run-fuelmanagement-py): 

image

After fix: 
image

### DIFF
--- a/armi/tests/refSmallCartesian.yaml
+++ b/armi/tests/refSmallCartesian.yaml
@@ -282,8 +282,8 @@ systems:
     core:
         grid name: core
         origin:
-            x: -65.0
-            y: -65.0
+            x: 0.0
+            y: 0.0
             z: 0.0
 
 grids:


john-science (approved):

john-science:
> Did something else change that resulted in the origin no longer being defined at the bottom left of the core? Let's make sure that 0, 0 is correct before just changing the blueprints.

@ntouran  I'm digging through the history and not seeing any changes.  Does this sound familiar to you?

PR #730, opened 2022-06-22
Fix the hex orientation of pin locations.

ntouran:
The pin locations in a hex block were rotated 30 degrees from where they
should be. This was discovered when plotting pin-level assembly plots
and they were all sticking out the edges. The major confusion comes in
from the fact that even though pointed end up  doesn't seem right when
you look at the pins, if you look at the little outline hexes
surrounding each pin then you can see that indeed the pointed end is
supposed to be up in this case.

This could impact results for anyone using auto-generated pin-level
grids and using the pin-wise pin coordinates.

Before: 
image

After: 
Screenshot from 2022-06-22 23-20-36

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -1935,6 +1935,8 @@ class HexBlock(Block):
         """
         Compute the local centroid coordinates of any pins in this block.
 
+        The pins must have a CLAD-flagged component for this to work.
+
         Returns
         -------
         localCoordinates : list
@@ -1952,7 +1954,7 @@ class HexBlock(Block):
                     [locator.getLocalCoordinates() for locator in clad.spatialLocator]
                 )
             else:
-                coords.append(locator.getLocalCoordinates())
+                coords.append(clad.spatialLocator.getLocalCoordinates())
         return coords
 
     def autoCreateSpatialGrids(self):
@@ -1969,6 +1971,8 @@ class HexBlock(Block):
         -----
         If the block meets all the conditions, we gather all components to either be a multiIndexLocation containing all
         of the pin positions, otherwise, locator is the center (0,0).
+
+        Also, this only works on blocks that have 'flat side up'.
 
         Raises
         ------
--- a/armi/reactor/tests/test_blocks.py
+++ b/armi/reactor/tests/test_blocks.py
@@ -1693,6 +1693,10 @@ class HexBlock_TestCase(unittest.TestCase):
         self.assertTrue(self.HexBlock.hasFlags(Flags.INTERCOOLANT))
 
     def test_getPinCoords(self):
+        blockPitch = self.HexBlock.getPitch()
+        pinPitch = self.HexBlock.getPinPitch()
+        nPins = self.HexBlock.getNumPins()
+        side = hexagon.side(blockPitch)
         xyz = self.HexBlock.getPinCoordinates()
         x, y, _z = zip(*xyz)
         self.assertAlmostEqual(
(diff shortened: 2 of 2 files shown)


armi/reactor/blocks.py
```diff
@@ -2002,10 +1985,14 @@ def autoCreateSpatialGrids(self):
...
             )
 
-        spatialLocators = grids.MultiIndexLocation(grid=self.spatialGrid)
         ringNumber = hexagon.numRingsToHoldNumCells(self.getNumPins())
         # For the below to work, there must not be multiple wire or multiple clad types.
-        grid = grids.HexGrid.fromPitch(self.getPinPitch(cold=True), numRings=0)
+        # note that it's the pointed end of the cell hexes that are up (but the
+        # macro shape of the pins forms a hex with a flat top fitting in the assembly)
+        grid = grids.HexGrid.fromPitch(
+            self.getPinPitch(cold=True), numRings=0, pointedEndUp=True
```
keckler, line 1993: Is this still working if a user puts into their layout blueprint the `geom: hex_corners_up` option?

I can't quite tell myself, but it feels like maybe in that case you should set `pointedEndUp=False`?

armi/reactor/blocks.py
```diff
@@ -1944,40 +1940,27 @@ def getPinCoordinates(self):
...
         -----
         This assumes hexagonal pin lattice and needs to be upgraded once more generic geometry
-        options are needed.
-
-        A block with fully-defined pins could just use their individual spatialLocators in a
-        block-level 2-D grid. However most cases do not have this to minimize overhead and maximize
-        speed. Thus we want to just come up with a uniform mesh of pins if they're not explicitly
-        placed in the grid.
-
+        options are needed. Only works if pins have clad.
```
keckler, line 1943: I guess this should be clarified that it only works if there is a component with the `CLAD` flag, right? A user could potentially have a clad component with other flags, and this would fail.

armi/reactor/blocks.py
```diff
@@ -1944,40 +1940,27 @@ def getPinCoordinates(self):
...
-                coordinates.append(xyz)
-        return coordinates
+        coords = []
+        for clad in self.getChildrenWithFlags(Flags.CLAD):
+            if isinstance(clad.spatialLocator, grids.MultiIndexLocation):
+                coords.extend(
+                    [locator.getLocalCoordinates() for locator in clad.spatialLocator]
+                )
+            else:
+                coords.append(locator.getLocalCoordinates())
```
keckler, line 1952: Does this line work? I don't think the local variable `locator` is ever defined?

keckler (changes requested):
I think also that your `black` tests are failing.

john-science:
@ntouran  `black` is broken. I'd fix it for you, but it's in your private repo.

armi/reactor/blocks.py
ntouran: yeah, I'll clarify that in the docstring

armi/reactor/blocks.py
ntouran: no this line does not work.

armi/reactor/blocks.py
ntouran: it only works on flat side up cases, you're right. I added a note in the docstring to the effect. It's be nice to detect that situation and throw an error, but we don't currently have a good way to look at a grid and determine which 'orientation' it is at this level.

armi/reactor/tests/test_blocks.py
```diff
@@ -1615,6 +1615,7 @@ def setUp(self):
         self.HexBlock.add(
             components.DerivedShape("coolant", "Sodium", Tinput=273.0, Thot=273.0)
         )
+        self.HexBlock.autoCreateSpatialGrids()
```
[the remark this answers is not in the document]
ntouran: I've added more checks to the test that would fail in the bugged condition and pass in the new one.

armi/reactor/blocks.py
john-science: > we don't currently have a good way to look at a grid and determine which 'orientation'

Sad trombone noises.

drewj-usnctech: Found this researching #1284

Issue #511, opened 2021-12-09
trackAssems should be False by default


john-science:
The default behavior for `trackAssems` should be turned to `False`:

https://github.com/terrapower/armi/blob/08e276833069478e8dd6c40c0dcbedf1e5de3456/armi/settings/fwSettings/globalSettings.py#L606-L612

The code change is obviously minor, but some work will need to be done looking around the ARMI ecosystem to make sure this doesn't hurt anyone's workflow.

ntouran:
This setting effectively turns on indefinite appending of all discharged assemblies to the spent fuel pool container. Obviously in infinitely long or large runs this inevitably fills up RAM. Long ago, we used assemblies there to track history. Today we can just read the output database from disk when history is needed within a run (e.g. to build an irradiation history for a fuel performance calc). 

If we turn this off completely, then nothing will ever get added to the SFP. We still can allow fuel handler subclasses to make that kind of decision and move themselves, which is honestly probably a better approach.

keckler:
If this change is implemented, I think some work will also need to be done on [`removeAssembly`](https://terrapower.github.io/armi/_modules/armi/reactor/reactors.html#Core.removeAssembly) to make its intention clear. Right now, because `trackAssems` and the `discharge` option in `removeAssembly` both default to `True`, the behavior when turning one of them to `False` is clear. But if `trackAssems` defaults to `False` while `discharge` defaults to `True`, I think it becomes potentially confusing to the user when trying to use either of them.

john-science:
> ...I think it becomes potentially confusing to the user when trying to use either of them.

This is an interesting point. I hadn't seen the `discharge` option there before. So that might be a bit of a wrinkle.

PR #733, opened 2022-06-24
Setting the trackAssems default to False

john-science:
## Description

As we have recently been having much discussion about memory, it is time to change the default value of `trackAssems` to False.

Really, this seems like the best option, though we should watch for how this affects downstream projects.

(ALSO: Sorry, I can't help myself. I also did two other cleanup activities as part of this PR: I removed a completely unused `countAssemblies()` method, and I fixed some RST formatting in our docs.)

---

## Checklist

### DIFF
--- a/armi/settings/fwSettings/globalSettings.py
+++ b/armi/settings/fwSettings/globalSettings.py
@@ -658,8 +658,10 @@ def defineSettings() -> List[setting.Setting]:
             CONF_TRACK_ASSEMS,
             default=False,
             label="Save Discharged Assemblies",
-            description="Track assemblies for detailed fuel histories. Disable in case "
-            "you get memory errors.",
+            description="Track assemblies for detailed fuel histories. For instance, "
+            "assemblies are tracked after they come out of a reactor by putting them "
+            "in a Spent Fuel Pool. This might be necessary for your work, but it "
+            "certainly increases the memory usage of the program.",
         ),
         setting.Setting(
             CONF_VERBOSITY,
--- a/armi/settings/fwSettings/globalSettings.py
+++ b/armi/settings/fwSettings/globalSettings.py
@@ -656,7 +656,7 @@ def defineSettings() -> List[setting.Setting]:
         ),
         setting.Setting(
             CONF_TRACK_ASSEMS,
-            default=True,
+            default=False,
             label="Save Discharged Assemblies",
             description="Track assemblies for detailed fuel histories. Disable in case "
             "you get memory errors.",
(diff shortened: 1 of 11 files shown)


john-science:
This PR is in reference to:  https://github.com/terrapower/armi/issues/511

keckler:
> (ALSO: Sorry, I can't help myself. I also did two other cleanup activities as part of this PR: I removed a completely unused countAssemblies() method, and I fixed some RST formatting in our docs.)

Add to that list "added some unit tests on `getNumEnergyGroups` and `getMinimumPercentFluxInFuel`."

armi/settings/fwSettings/globalSettings.py
```diff
@@ -656,7 +656,7 @@ def defineSettings() -> List[setting.Setting]:
         ),
         setting.Setting(
             CONF_TRACK_ASSEMS,
-            default=True,
+            default=False,
             label="Save Discharged Assemblies",
             description="Track assemblies for detailed fuel histories. Disable in case "
```
keckler, line 661: Can we add a comment here about what it means to "track" assemblies? Similar to what you say elsewhere about the assemblies going into the SFP or not.

armi/utils/dynamicImporter.py
```diff
@@ -24,7 +24,9 @@
 def importEntirePackage(module):
     """Load every module in a package
 
-    NOTE: this method may only work for a flat directory?
+    Notes
+    -----
+    This method may only work for a flat directory?
```
keckler, line 29: Not a huge fan of having a comment visible in the API that has a question mark on it.

I guess whatever, and not at all related to the current PR intent, but for how long are we gonna kick this can down the road?

keckler (approved):
Looks good to me. Thanks for adding this change to the release notes, I'm sure it will catch someone (i.e. myself) off guard.

armi/utils/dynamicImporter.py
john-science: That's a fair point.  I didn't put the question mark there, so I just didn't see it.

I'll just test that statement to see if it's true, and adjust the note accordingly.

armi/settings/fwSettings/globalSettings.py
john-science: Fixed

armi/utils/dynamicImporter.py
john-science: Oh, yeah, there is no reason for that "?" at all, that little method is dead simple and only works on flat directories. I cleaned up the comment.

keckler (approved):

Issue #731, opened 2022-06-23
`fuelPerformance.plugin` can't handle multiple different pins with bond


keckler:
https://github.com/terrapower/armi/blob/4894a56c4c5498272d47b6bc307765eb207f4801/armi/physics/fuelPerformance/plugin.py#L66-L87

In that function, there is a call to `bond = b.getComponent(Flags.BOND, quiet=True)`. This call can only pick up a single component with the `BOND` flag. If there are multiple components with the `BOND` flag, `getComponent()` fails with a ValueError: https://terrapower.github.io/armi/_modules/armi/reactor/composites.html#ArmiObject.getComponent

```
ValueError: Multiple components match in <IFR1BlanketBlock B0082-002 at ExCore XS: A BU GP: A> match typeSpec Flags.BOND: [<Circle: gap_U19Pu10Zr>, <Circle: gap_U8Pu10Zr>, <Circle: gap_U10Zr>]
```

I guess _okay_, I understand why it is difficult to set a parameter named `bondBOL` for an entire block when there are multiple potential values for it. 

But also, this is a scenario that is reasonably likely to occur. So we should probably be able to handle it. Not sure what to do about it though...

On another note, this does not feel like something that should be in the framework. It feels like it belongs in a plugin, because even the concept of a bond inside the pins is pretty specific to a small subset of reactor designs.

 @sammiller11235 Do you have any thoughts on this?

sammiller11235:
Hmm, seems like we need to add some logic to deal with multiple bonds. If you add an "s" to getComponent:  `bondS = b.getComponents(Flags.BOND, quiet=True)` does that get you the 3 bonds from the components? If you so, you could add some logic like: `b.p.bondBOL = sum([sum(bond.getNuclideNumberDensities(bond.getNuclides())) for bond in bondS])`

keckler:
...and what if the different bonds are different materials? :-\

sammiller11235:
Hmm, good points. It looks like the parameter `b.p.bondBOL` is used in a couple of places to reduces the nuclide densities based on a bondRemovalFraction (see `crucible._applyBondRemovalFractions` and `fuelPerformance.utils.enforceBondRemovalFraction`), so those functions will have to get updated as well to handle this change. 

I think if the bonds are different materials that is reason to fail with a NotImplementedError. Otherwise, as long as the bonds are the same material and if the removal can be handled appropriately I think this doesn't need a warning . Let me know if you'd like some help with the logic to sum or remove the bonds

PR #734, opened 2022-06-24
Remove `_setBOLBond` and other fuel performance utils from the framework

keckler:
## Description

This PR is in response to #731 . 

Essentially, there was an issue where a call to `_setBOLBond()` in the `afterConstructionOfAssemblies` hook could not handle the situation in which a block has multiple components defined with the `BOND` flag. In reality, this was a restriction related to one of our downstream plugins and has nothing to do with any inherent limitations of ARMI. 

Further, the `_setBOLBond` function itself is pretty specific to SFR pin types, so it doesn't really belong in the framework in the first place.

To get around these issues, the `_setBOLBond` function and its call from within `afterConstructionOfAssemblies` have been migrated from the framework to our internal downstream fuel performance plugin.

Another change that was made was the removal of the `enforceBondRemovalFraction` function entirely. This function has multiple issues:
(1) It is not used anywhere either in the framework or in our internal downstream plugins
(2) It is, again, rather specific to SFR designs, and thus doesn't really belong in the framework at all
(3) The methodology that it implements (moving bond material from the bond into the coolant) does not appear to represent any actual physics that one is likely to encounter, so its intended use in the first place is not clear

So this function and the associated unit test are just removed entirely.

Finally, the `bondBOL` block parameter has been migrated from the framework to our internal fuel performance plugin, since it is now not used anywhere in the framework or in any other of our downstream plugins.

---

## Checklist


keckler:
Review requested from @sammiller11235  and @ntouran

armi/physics/fuelPerformance/plugin.py
```diff
@@ -59,46 +55,4 @@ def defineParameters():
     @staticmethod
     @plugins.HOOKIMPL
     def afterConstructionOfAssemblies(assemblies, cs):
```
sammiller11235, line 57: Should these three lines be deleted as well?

sammiller11235 (commented):
The changes to fuel performance plugin look good. I don't quite understand what you're doing with `database3.py` though

keckler:
Oh, whoops, thats from another PR. Crap.

I'll pull that out!

armi/physics/fuelPerformance/plugin.py
keckler: They could be. I just put a `pass` into the body of the function just to acknowledge the presence of the hook without actually doing anything when its called.

Not sure what happens if you remove the hook entirely...

keckler:
@sammiller11235 
Got rid of the changes you were seeing with `database3`. Sorry about that!

armi/physics/fuelPerformance/plugin.py
sammiller11235: I don't think this method is defined for all other plugins, so I think it can be removed without issue. @ntouran would probably know for certain

sammiller11235 (approved):
To summarize the approach we're taking: 
the b.p.bondBOL is used exclusively by Crucible to remove bond, so the _setBOLBond is moving to Crucible. To address the problem raised by #731, Crucible needs to implement generic capability to handle blocks with multiple fuel components (for FFTF and EBR-II assemblies, as well as LDAs). This is larger in scope than dealing with the bondBOL parameter.

armi/physics/fuelPerformance/plugin.py
keckler: I guess it is part of the base `plugin` class: https://terrapower.github.io/armi/.apidocs/armi.plugins.html?highlight=afterconstructionofassemblies

So I'll delete it here and let the fuel performance plugin just default to that of the parent class.

armi/physics/fuelPerformance/plugin.py
```diff
@@ -55,50 +51,3 @@ def defineParameters():
         from armi.physics.fuelPerformance import parameters
 
         return parameters.getFuelPerformanceParameterDefinitions()
-
-    @staticmethod
-    @plugins.HOOKIMPL
-    def afterConstructionOfAssemblies(assemblies, cs):
```
john-science, line 61: **Question**: Have you check other downstream ARMI plugins to make sure they don't depend on `afterConstructionOfAssemblies()`?

There are some common ones you could `git grep` through to see if this PR will break a bunch of plugins, you know?

keckler: But the method is empty after the removal of `_setBOLBond`. And yes, we are certain that `_setBOLBond` is not used in any downstream plugins, with the exception of our internal metal fuel performance plugin, to which the `_setBOLBond` call has been added. 

(Let me know if you want a link to the relevant PR in that plugin, and I can send it to you offline.)

And then once `afterConstructionOfAssemblies` is empty, it is the same as that in its `super` class, so removing it shouldn't be a problem.

But let me know if you think otherwise.

john-science (approved):

Issue #705, opened 2022-06-09
Allow users to always provide input for non-uniform assemblies in the blueprints


jakehader:
Currently, if the `detailedAxialExpansion` setting is set to `False`, the user input is for all the assemblies in the blueprints file are required to contain axial mesh points that are all aligned, whereas when the `detailedAxialExpansion` setting is set to `True` this restriction is lifted. It would be best if the ARMI framework allowed users to provide a mesh structure that they would like (based on the core design of each assembly) and then for the framework to handle this on reactor construction.

For instance, if we allowed any mesh to be provided when `detailedAxialExpansion` is set to `False` we could immediately call the `UniformMesh` converter on reactor construction before any work is performed and this would still allow a user to perform analyses using a uniform mesh. 

In this case, the logic for `processLoading` on the reactor needs to be modified to call the `UniformMesh` converter if `detailedAxialExpansion` is set to `False` within the `makeAxialSnapList` method.

`processLoading`:

https://github.com/terrapower/armi/blob/0c7ef201c9f1ce5256e925b127b4a49c40659456/armi/reactor/reactors.py#L2245-L2248

`makeAxialSnapList`:

https://github.com/terrapower/armi/blob/0fa09f9fb43d2d1dae062f7fa89d8b2084cb0ab7/armi/reactor/assemblies.py#L575

This update will allow more flexibility for users and will also allow users to create a single blueprints file that represents the design of their reactor system. This is important for @onufer in the near term.

jakehader:
@albeanth - Would you be able to take a stab at this?

onufer:
@jakehader  just to clarify, specifying a non-uniform mesh will result in the reactor state being stored on a uniform mesh in memory, correct (i think this is what direction @albeanth  is going)? Rather than converting to a uniform mesh prior to each global flux solve?

jakehader:
@onufer that is correct!

albeanth:
My working branch to address this issue.

https://github.com/albeanth/armi/tree/processLoadUpdate

john-science:
How far have our recent PR go to solving this Issue?   

#735  -  Process load update and new uniform mesher
#827  -  Support non-uniform assemblies with detailedAxialExpansion disabled

If we made some progress, what do we have left to do to complete this ticket?

albeanth:
[...]

Issue #706, opened 2022-06-09
Deepcopy of Blocks, Assemblies, and Blueprints also copy the whole Reactor


john-science:
It appears that doing a `copy.deepcopy()` of certain ARMI objects will also make a "deep copy" of the entire `Reactor` and `Core`, which is slightly terrifying:

* `Block`
* `Assembly`
* `Blueprints`

First things first, we should validate that this is true, and for what ARMI objects. For instance, it might be that `Blueprints` do not have this problem, but only `ArmiObject`s do.

Next, we should override the `__deepcopy__` method on any ARMI object with this problem, and fix it.  One good lead would be to utilize Python's [weakref](https://docs.python.org/3.9/library/weakref.html), for a little reference to the `Reactor` on every object that won't cause memory leaks during garbage collection.

I'm not thrill with all the backrefs in every little `ArmiObject` all the way up to the `Reactor` and the `Core`. But since we are here, we better protect ourselves.

john-science:
@keckler  @jakehader @ntouran @albeanth  FYI

john-science:
It appears `Block` has a first-draft version of `__deepcopy__`, that doesn't (yet) use weakref:

https://github.com/terrapower/armi/blob/8c46f8c4f625fcf7e72fdca45801afe46ec105aa/armi/reactor/blocks.py#L138

Ditto for `Core`:

https://github.com/terrapower/armi/blob/0c7ef201c9f1ce5256e925b127b4a49c40659456/armi/reactor/reactors.py#L247

`Reactor`:

https://github.com/terrapower/armi/blob/0c7ef201c9f1ce5256e925b127b4a49c40659456/armi/reactor/reactors.py#L95

And, for some reason, `ParameterCollection`:

https://github.com/terrapower/armi/blob/260f518fb9e54d14149fb5969ce83547e52bd6f6/armi/reactor/parameters/parameterCollections.py#L275

There is some coverage, but it is a bit eclectic. I would be curious if we could place a good version of this method on `ArmiObject`.

albeanth:
I recently found that the operator object on the reactor, `r.o`, is not copied when deepcopy'ing a reactor. Which is interesting. 

So the following results in a `None`:
```python
from copy import deepcopy

operator, reactor = test_reactors.loadTestReactor()
print(reactor.o)
# prints --> <Operator Standard <Settings name: ... >

rCopy = deepcopy(reactor)
print(rCopy.o)
# prints --> None
```

[...]

PR #735, opened 2022-06-24
Process load update and new uniform mesher (beta)

albeanth:
## Description
This addresses #705 and provides the beginnings of a new uniform mesher; though is *not* an ARMI _converter_ like the original/current but is rather an ARMI _changer_.

The primary outcomes of this PR are:
1. Blueprint assembly definitions are allowed to be axially disjoint in a core and have `detailedAxialExpansion: False`. 
2. The introduction of a new uniform mesher that aims to minimize (and in some cases remove) inter-block material smearing that is a result of the current uniform mesh converter https://github.com/terrapower/armi/blob/4894a56c4c5498272d47b6bc307765eb207f4801/armi/reactor/converters/uniformMesh.py#L35-L40
    - the new framework settings, `primaryAssemblyToConserve` and `secondaryAssemblyToConserve` provide options for users to provide strings that correspond to assembly `Flags` that group assemblies together and aim to best preserve their material interfaces. E.g., setting `primaryAssemblyToConserve: "control"` would give control assemblies highest priority in preserving their block bounds and would aim to minimize (or eliminate) inter-block material smearing (e.g., control into plenum). 
    - Note, only assemblies of a single flag type can be grouped together.
    - The "primary" and "secondary" settings were included to to enable priorities for multiple assembly types.
 
Two other points:
1. the new unit tests that come with this PR are particularly heavy... Future commits will be included to alleviate their cost. 
2. I included the new uniform mesher into the axial expansion changer here. https://github.com/terrapower/armi/blob/4894a56c4c5498272d47b6bc307765eb207f4801/armi/reactor/converters/axialExpansionChanger.py#L288-L295 To improve ease of use, I changed the constructor for `class AxialExpansionChanger` and had to make a slew of associated adjustments. 

---

## Checklist


albeanth:
@onufer, @jakehader, @keckler, @mgjarrett, @john-science

john-science:
@albeanth  The tests are all borked. Just FYI.

albeanth:
Yeah... I'm seeing that and don't know why. They pass on my machine....?

armi/tests/detailedAxialExpansion/refSmallReactorBase.yaml
```diff
@@ -346,8 +356,8 @@ blocks:
             Thot: 450.0
             ip: 15.277
             mult: 1.0
-            op: 16.28228
-        intercoolant: &component_control_intercoolant
+            op: 16.6
+        intercoolant:  &component_control_intercoolant
```
mgjarrett, line 360: Extra space here

john-science:
> Yeah... I'm seeing that and don't know why. They pass on my machine....

Two classic problems that can cause this:

1. You didn't push all your code to GitHub. (check `git status` to see if this is true)
2. Try running your tests _exactly_ as GitHub does (`python -m tox -e test`)

albeanth:
Adding some context for commit https://github.com/terrapower/armi/pull/735/commits/54832a6b283a0b7e4426858c9cdf7a13223bcacb. 

The changes in this PR to `reactor.py::Core::processLoading` remove the call to `makeAxialSnapList`, which in turn, avoids a call to b.p.topIndex. So those params are never set and therefore the number of differences in the bookkeeping unit test is reduced.

armi/settings/fwSettings/globalSettings.py
```diff
@@ -154,6 +156,28 @@ def defineSettings() -> List[setting.Setting]:
             "block. Used for axial mesh refinement.",
             schema=vol.All(vol.Coerce(int), vol.Range(min=0, min_included=False)),
         ),
+        setting.Setting(
+            CONF_PRIMARY_ASSEMBLY_CONSERVE,
+            default="fuel",
```
jakehader, line 161: I recommend adding a setter to this setting that transforms the inputs into valid flags so that we arent passing string inputs around.

armi/tests/detailedAxialExpansion/refSmallReactorBase.yaml
albeanth: Addressed. https://github.com/terrapower/armi/pull/735/commits/b8ba5f29b2ba56b5c1f60a6b1a68b7337e44867e

armi/reactor/converters/uniformMesh_V2_beta.py
```diff
@@ -0,0 +1,450 @@
...
+    heights = [h for _b, h in overlappingBlockInfo]
+    blocks = [b for b, _h in overlappingBlockInfo]
+    sourceBlk = blocks[heights.index(max(heights))]
+
+    # create a new block, b, that is the same type and contains the same params, grid, etc as sourceBlk
+    b = sourceBlk.__class__(sourceBlk.getType())
+    for param in sourceBlk.getParamNames():
+        b.p[param] = sourceBlk.p[param]
+    for c in sourceBlk.getChildren():
+        b.add(c.__copy__())
```
albeanth, line 199: @john-science Just wanted to bring this section to your attention since you have the open issue, https://github.com/terrapower/armi/issues/706, on deepcopies. 

I made a concerted effort to avoid using `deepcopy` when creating a copy of an assembly -> block -> component, but gave in with components. I decided to use the `Component::__copy__` method with the hope that when #706 is resolved, it too will be resolved. Anyway, just figured I'd get your eyes on this section in particular.

john-science:
This PR temporarily in a holding cycle while we fix the affects on downstream projects.

Soooooon!

armi/reactor/converters/uniformMesh_V2_beta.py
john-science: Thanks for keeping it in mind.

My concern is that "fixing" that will break a dozen major plugins, so it won't be possible for months.

Fingers crossed I'm just paranoid.  We'll find out!

PR #742, opened 2022-06-30
Properly handle volume integrated parameters in the conversion from 1/3 to full core

keckler:
## Description

In the `ThirdCoreHexToFullCoreChanger`, the central assembly needs to be handled specially due to it being only 1/3 of a full assembly in the 1/3 core representation. 

Right now this is neglected, so when the reactor is converted to a full core, block parameters possessing the `VOLUME_INTEGRATED` attribute (for instance, block power) end up having only 1/3 of their proper value in the full-core representation.

If using the `ThirdcoreHexToFullCoreChanger`, this issue will ultimately trigger a `ValueError` from the [`_checkConversion`](https://github.com/terrapower/armi/blob/93f908d845605ed2d38959fbe62445be883d23c5/armi/reactor/converters/uniformMesh.py#L371) method of the `NeutronicsUniformMeshConverter`, but the origin of the error isn't immediately apparent when that error is raised.

This issue has been fixed in this PR. Tests confirming correct behavior have also been added.

---

## Checklist


keckler:
@john-science FYI this is ready for merge.

In addition to the unit testing added, I did some tests on my own of a full run, and it appears to work as intended.

opotowsky (approved):
Really great work on this!

armi/reactor/converters/geometryConverters.py
```diff
@@ -1180,6 +1180,27 @@ class ThirdCoreHexToFullCoreChanger(GeometryChanger):
...
+
+    def _scaleBlockVolIntegratedParams(self, b, direction):
+        if direction == "up":
+            op = operator.mul
+        elif direction == "down":
+            op = operator.truediv
+
+        for param in self.listOfVolIntegratedParamsToScale:
+            print(param)
+            print(b.p[param])
```
mgjarrett, line 1195: Was this left intentionally? It's creating a ton of output.

mgjarrett: @keckler @john-science

keckler: Haha, whoops!

keckler: So much for code review...

I'll fix it.

keckler: https://github.com/terrapower/armi/pull/753

PR #746, opened 2022-07-05
Cleanup composite.getComponent

john-science:
## Description

This PR cleans up the `Composite.getComponent()` method, by removing an unused parameter.

I also threw in some code coverage for various things, just on general principles.

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_composites.py
+++ b/armi/reactor/tests/test_composites.py
@@ -387,6 +387,16 @@ class TestCompositeTree(unittest.TestCase):
 
         self.assertEqual(fuelMass, a.getFuelMass())
 
+    def test_getNeutronEnergyDepositionConstants(self):
+        """Until we improve test architecture, this test can not be more interesting"""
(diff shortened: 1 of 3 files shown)


armi/reactor/tests/test_composites.py
```diff
@@ -387,6 +387,16 @@ def test_getFuelMass(self):
 
         self.assertEqual(fuelMass, a.getFuelMass())
 
+    def test_getNeutronEnergyDepositionConstants(self):
```
keckler, line 390: Perhaps you could add a short comment to this one and the next about _why_ an error is raised from this call? Because at first glance, it doesn't look like something I would expect to fail.

keckler (approved):

armi/reactor/tests/test_composites.py
john-science: Done.

The problem is the little test reactor doesn't have a cross-section library loaded. But loading one into it would break a bunch of other tests.

PR #747, opened 2022-07-05
update helix docstrings + bug fix

albeanth:
- docstrings expanded upon to improve clarity
- bug fix in `getCircleInnerDiameter`
- `getCircleOuterDiameter` adjusted for consistency with `getCircleInnerDiameter`

---

## Checklist

### DIFF
--- a/armi/reactor/components/complexShapes.py
+++ b/armi/reactor/components/complexShapes.py
@@ -185,6 +185,13 @@ class Helix(ShapedComponent):
     http://mathworld.wolfram.com/Helix.html
     In a single rotation with an axial climb of P, the length of the helix will be a factor of
     2*pi*sqrt(r^2+c^2)/2*pi*c longer than vertical length L. P = 2*pi*c.
+
+    - od: outer diameter of the helix wire
+    - id: inner diameter of the helix wire (if non-zero, helix wire is annular.)
+    - axialPitch: vertical distance between wraps
+    - helixDiameter: can be visualized if the axial pitch is 0.0 (creates a circle). The helix diameter
+                     is the distance from the center of the wire-wrap on one side to the center of the wire-wrap
+                     on the opposite side.
     """
 
     is3D = False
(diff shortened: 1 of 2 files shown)


armi/reactor/components/complexShapes.py
```diff
@@ -230,13 +237,18 @@ def __init__(
...
 
     def getCircleInnerDiameter(self, Tc=None, cold=False):
-        return self.getDimension("id", Tc, cold) + self.getDimension(
-            "helixDiameter", Tc, cold=cold
+        """the diameter of a circle which is encompassed by the interior of the wire-wrap
+
+        - should be equal to the outer diameter of the pin in which the wire is wrapped around
+        """
+        return self.getDimension("helixDiameter", Tc, cold=cold) - self.getDimension(
+            "od", Tc, cold
```
john-science, line 251: Why would we be using the outer diameter here, if the method is called "getCircleInnerDiameter"?

This is not clear to me.

onufer: its because a helix is weird. getCircleInnerDiameter is supposed to be the circle the component would enclose. For a helix, to enclose means to wrap around rather than looking at the anulus of the helix wire.

armi/reactor/components/complexShapes.py
```diff
@@ -185,6 +185,13 @@ class Helix(ShapedComponent):
     http://mathworld.wolfram.com/Helix.html
     In a single rotation with an axial climb of P, the length of the helix will be a factor of
     2*pi*sqrt(r^2+c^2)/2*pi*c longer than vertical length L. P = 2*pi*c.
+
+    - od: outer diameter of the helix wire
+    - id: inner diameter of the helix wire (if non-zero, helix wire is annular.)
+    - axialPitch: vertical distance between wraps
```
onufer, line 191: call out "full revolution" , and mention 2pi or 360 degrees to be explicit

armi/reactor/components/complexShapes.py
```diff
@@ -185,6 +185,13 @@ class Helix(ShapedComponent):
...
+    - helixDiameter: can be visualized if the axial pitch is 0.0 (creates a circle). The helix diameter
```
onufer, line 192: probably move `can be visualized if the axial pitch is 0.0 (creates a circle).`  to the end. the first thing that should be written is the defintion.

onufer (changes requested):

armi/reactor/components/complexShapes.py
albeanth: In the context of a Helix, the outer diameter (`od`) is the diameter of the helical wire wrap. In order to get the diameter of the inner circle of the helix, which is bounded by the interior of the wire wrap, we start from the `helixDiameter` and increment inwards. The opposite is the case for `getBoundingCircleOuterDiameter`. 

For Helixes (Helixi?) the inner diameter, `id`, is the inner diameter of the wire wrap. So before we were returning the `getCircleInnerDiameter` incorrectly - for 99% of the time when the `id = 0.0`, the `getCircleInnerDiameter` was returning the `helixDiameter` which isn't quite right. 

The bullet points added to the Notes section of the class docstring try to address this. 

@onufer

armi/reactor/components/complexShapes.py
albeanth: Addressed. https://github.com/terrapower/armi/pull/747/commits/f03f374c117b2fece94b2b1beec752bd7c1a103d

armi/reactor/components/complexShapes.py
albeanth: Addressed. https://github.com/terrapower/armi/pull/747/commits/f03f374c117b2fece94b2b1beec752bd7c1a103d

onufer (approved):

armi/reactor/components/complexShapes.py
john-science: okay, then

john-science (approved):

PR #748, opened 2022-07-05
Making component naming conventions more clear

john-science:
## Description

As it happens, component names are parsed by Python regex, so you cannot use any regex identifiers in component names:

> . ^ $ * + ? { } [ ] \ | ( )

And component names are used in YAML, so it would be best to keep `:` out of the names as well.

This PR adds these component-naming convention rules to the docs, and improved the clarity of one of the error messages in `components.py`.

---

## Checklist

### DIFF
--- a/doc/developer/tooling.rst
+++ b/doc/developer/tooling.rst
@@ -3,7 +3,7 @@ Tooling and Infrastructure
 
 Good Commit Messages
 --------------------
-The ARMI project follows a few basic rools for "good" commit messages:
+The ARMI project follows a few basic rules for "good" commit messages:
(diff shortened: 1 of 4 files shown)


doc/developer/tooling.rst
```diff
@@ -3,7 +3,7 @@ Tooling and Infrastructure
 
 Good Commit Messages
 --------------------
-The ARMI project follows a few basic rools for "good" commit messages:
```
keckler, line 6: I prefer the old version LOL

keckler (approved):

PR #749, opened 2022-07-05
Remove duplicate, incorrect setting of core `power` parameter

keckler:
## Description

In [`processLoading`](https://github.com/terrapower/armi/blob/00f738f16cc5b6d028c19083b43cdacca44f7aa1/armi/reactor/reactors.py#L2198), the core's `power` parameter is set. This is not necessary, as core power is already set both within the operator while stepping through time and during restarts.

But the real problem here is that power is being set _incorrectly_ during `processLoading`. Specifically, the value being used does not account for the `powerFraction(s)` setting. So right now, it always sets the power to the _rated_ power. This becomes apparent during restart cases when `powerFraction(s)` is non-unity, and then the code fails during [`NeutronicsUniformMeshConverter._checkConversion`](https://github.com/terrapower/armi/blob/d04e1be8937c44e506dfcd5ff25459fca24af6cf/armi/reactor/converters/uniformMesh.py#L371) (thank god for that little function, it's caught a number of bugs recently!).

This PR:
1. Fixes this problem by removing the setting of `core.p.power` from within `processLoading`
2. Adds unit tests to confirm that the changes don't break the code
3. Repairs a unit test that was broken by the change

I also have been using this updated code in my personal runs (both restart and standard) and it seems to have fixed the problem without breaking anything else.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/tests/test_comparedb3.py
+++ b/armi/bookkeeping/db/tests/test_comparedb3.py
@@ -155,7 +155,7 @@ class TestCompareDB3(unittest.TestCase):
 
         # end-to-end validation that comparing a photocopy database works
         diffs = compareDatabases(dbs[0]._fullPath, dbs[1]._fullPath)
-        self.assertEqual(len(diffs.diffs), 459)
+        self.assertEqual(len(diffs.diffs), 456)
         self.assertEqual(diffs.nDiffs(), 3)
 
 
--- a/armi/bookkeeping/db/tests/test_database3.py
+++ b/armi/bookkeeping/db/tests/test_database3.py
@@ -24,6 +24,7 @@ from armi.bookkeeping.db import database3
 from armi.reactor import grids
 from armi.reactor.tests import test_reactors
 from armi.tests import TEST_ROOT
+from armi.utils import getPreviousTimeNode
 from armi.utils.directoryChangers import TemporaryDirectoryChanger
 
 
@@ -138,14 +139,20 @@ class TestDatabase3(unittest.TestCase):
         # first successfully call to prepRestartRun
         o, r = test_reactors.loadTestReactor(TEST_ROOT)
         cs = o.cs
+        ratedPower = cs["power"]
+        startCycle = cs["startCycle"]
+        startNode = cs["startNode"]
+        cyclesSetting = [
+            {"step days": [1000, 1000], "power fractions": [1, 1]},
+            {"step days": [1000, 1000], "power fractions": [1, 1]},
+            {"step days": [1000, 1000], "power fractions": [1, 1]},
+        ]
+        cycleP, nodeP = getPreviousTimeNode(startCycle, startNode, cs)
+        cyclesSetting[cycleP]["power fractions"][nodeP] = 0.5
         cs = cs.modified(
             newSettings={
                 "nCycles": 3,
-                "cycles": [
-                    {"step days": [1000, 1000], "power fractions": [1, 1]},
-                    {"step days": [1000, 1000], "power fractions": [1, 1]},
-                    {"step days": [1000, 1000], "power fractions": [1, 1]},
-                ],
+                "cycles": cyclesSetting,
                 "reloadDBName": "something_fake.h5",
             }
         )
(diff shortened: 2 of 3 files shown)


john-science:
`processLoading()` is also called when you delete a ring with `Reactor.removeAssembliesInRing()`.

Would your change affect that method?

keckler:
> `processLoading()` is also called when you delete a ring with `Reactor.removeAssembliesInRing()`.
> 
> Would your change affect that method?

I don't see any reason why the power would need to be reset when calling `removeAssembliesInRing()`. 

And that method is tested, and that test doesn't break with the new changes. https://github.com/terrapower/armi/blob/5a20a05deb99d939ba5b467903ada8ff63143207/armi/reactor/tests/test_reactors.py#L713-L727

So I'm pretty sure removing the power reset from `processLoading()` doesn't have a negative impact on `removeAssembliesInRing()`. In fact, I would say that the method was broken before precisely because the power was being reset incorrectly anytime `removeAssembliesInRing()` was being called.

opotowsky (approved):
Hey this looks good to me!

armi/bookkeeping/db/tests/test_comparedb3.py
```diff
@@ -155,7 +155,7 @@ def test_compareDatabaseSim(self):
 
         # end-to-end validation that comparing a photocopy database works
         diffs = compareDatabases(dbs[0]._fullPath, dbs[1]._fullPath)
-        self.assertEqual(len(diffs.diffs), 459)
+        self.assertEqual(len(diffs.diffs), 456)
```
john-science, line 158: I feel like I was a bit too glib when I wrote this unit test.

I should probably re-write it so the actual `assert`s are a little more meaningful and transparent.

john-science:
@keckler Sorry, just to make sure I understand the situation (`removeAssembliesInRing()`).

If you remove an entire ring from the reactor... wouldn't we expect the power of the Reactor to change dramatically?  Like, it would go down a lot, right?

Or am I thinking of this all wrong?

keckler:
That's a good question, but I think you're thinking about it wrong. 

First of all, `removeAssembliesInRing()` is meant to be used as part of a fuel handler. So it is likely, though maybe not required, that those assemblies will be replaced with some other assemblies via the `add()` method or something similar.

But also, the power at which one drives the reactor is independent of the number of assemblies in that reactor. So I could take out an assembly and still operate my reactor at the same power level (assuming it is still at least critical). It would just change the power level of each individual assembly. 

But the act of simply removing/adding assemblies to the reactor does not imply anything about reactor power. Surely it does not imply that `core.p.power` should be reset to the reactor's _rated_ power with no consideration to things like `powerFraction(s)`, as is currently being done.

Does that seem clear?

jakehader:
> That's a good question, but I think you're thinking about it wrong.
> 
> First of all, `removeAssembliesInRing()` is meant to be used as part of a fuel handler. So it is likely, though maybe not required, that those assemblies will be replaced with some other assemblies via the `add()` method or something similar.
> 
> But also, the power at which one drives the reactor is independent of the number of assemblies in that reactor. So I could take out an assembly and still operate my reactor at the same power level (assuming it is still at least critical). It would just change the power level of each individual assembly.
> 
> But the act of simply removing/adding assemblies to the reactor does not imply anything about reactor power. Surely it does not imply that `core.p.power` should be reset to the reactor's _rated_ power with no consideration to things like `powerFraction(s)`, as is currently being done.
> 
> Does that seem clear?

One thing I may add here is that we need to make sure that the reactor state data is being set / initialized even without an operator being loaded. In theory we should be able to load a reactor without an operator or interfaces. I do think that you're right that this power is wrong, but maybe there is something else we are missing or that this was masking.

keckler:
I guess I am somewhat confused by your comment @jakehader . Maybe some more clarification is needed, but let me elaborate my intention.

We can definitely load up a reactor without the operator or interfaces and without setting `core.p.power` during `processLoading`. An example of this would be in our `loadTestReactor` method.

I think one thing that is being potentially confused here is the difference between `core.p.power` and the case setting `power`. `core.p.power` is dependent on the time node, as the reactor can change power with time via things like the `powerFractions` case setting. So a parameter like this is really akin to something like a block's flux value or a component's temperature, which we would not expect the be set upon initialization outside of an operator.

But the case setting `power` refers to the reactor's _rated_ power, and is not time-dependent. This will be available via the reactor's case settings when it is loaded, either from an operator or another standalone method.

So in my view, a call like `core.p.power = 1234` doesn't belong in the initialization of a standalone reactor. In such a situation, I would only expect the case setting `power` and other related case settings like `powerFractions` to be defined.

Am I misunderstanding what you're trying to say?

jakehader:
I guess my thought here is, _is_ it the responsibility of the operator to initialize reactor parameters from case settings or is the initialization, or the digestion of the case settings supplied a reactor / core method. I could see it either way, but say I want to load a reactor right now (ignoring that the power set from case settings is not correct) then I think there should be some way to say "reactor.setParamsFromCs(cs)". If this existed then we could (1) refactor the operator to just call this after process loading is called, and (2) have a mechanism in place so that if I whip up a new reactor without an operator (like loadTestReactor), I have a formal way to initialize the parameters that would be consistent with how the operator would do it anyways.

I could be off base here, but I was just thinking about whose responsibility it is to init the reactor and if the reactor or core should be able to do this from case settings independent of the operator.

keckler:
This seems to have caused an issue for cases where `nCycles=1` and `burnSteps=0`. In that case, the power is never set because the operator never enters the main loop in `_cycleLoop`:
https://github.com/terrapower/armi/blob/1990fcd5c77d11797f41d29212b3d17a82ad1ed4/armi/operators/operator.py#L334-L365

So something needs to be added in there to set the power even if the main loop is not entered.

Thanks @mgjarrett for finding this!

Working on a fix.

Issue #732, opened 2022-06-23
`returnNull` parameter of `getComponent()` never used in method


keckler:
Self explanatory.

I don't think the `returnNull` parameter is being used in any calls, but needs to be double checked.

https://github.com/terrapower/armi/blob/4894a56c4c5498272d47b6bc307765eb207f4801/armi/reactor/composites.py#L2451-L2493

john-science:
Agreed.

I have looked around our ecosystem a bit and can't find anyone that is subclassing this and using `returnNull` either.

So, it seems like it should be quite safe to remove.

PR #754, opened 2022-07-06
Removing unused method arguments

john-science:
## Description

@keckler  recently opened [a ticket](https://github.com/terrapower/armi/issues/732) for an unused method argument.  Well, we fixed his method, but I thought I should do a quick cull through the rest of the repo to see if I could find other methods with unused arguments, and fix those as well.

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_components.py
+++ b/armi/reactor/tests/test_components.py
@@ -938,12 +938,12 @@ class TestHelix(TestShapedComponent):
         "id": 0.1,
     }
 
-    def test_getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
+    def test_getBoundingCircleOuterDiameter(self):
         ref = 2.0 + 0.25
         cur = self.component.getBoundingCircleOuterDiameter(cold=True)
         self.assertAlmostEqual(ref, cur)
 
-    def test_getCircleInnerDiameter(self, Tc=None, cold=False):
+    def test_getCircleInnerDiameter(self):
         ref = 2.0 - 0.25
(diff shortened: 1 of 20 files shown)


opotowsky (approved):
The work of a true hero -- good cleanup!

armi/reactor/tests/test_components.py
```diff
@@ -938,12 +938,12 @@ class TestHelix(TestShapedComponent):
         "id": 0.1,
     }
 
-    def test_getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
+    def test_getBoundingCircleOuterDiameter(self):
```
keckler, line 941: Good catch.

This went through the review of #747 yesterday without any notice!

armi/utils/pathTools.py
```diff
@@ -116,25 +116,10 @@ def isAccessible(path):
...
-    Raises
-    ------
-    ValueError:
-        If the path does not exist or
-
-
-    """
-    pass
-
-
```
keckler, line 132: This function is just egregious. How did this get in here??

keckler (approved):
Awesome!

armi/utils/pathTools.py
john-science: I can't tell. It had been there for three years.  

I know clean up PRs aren't sexy, but man, sometimes they are necessary.

ntouran:
Running internal in-use tests suggests that at least some of these are indeed used by internal plugins. Did yall run the internal tests?

PR #758, opened 2022-07-11
Fix setting of `core.p.power` in zero-burnup cases

keckler:
## Description

#749 recently fixed a bug with the setting of the `core.p.power` parameter. Apparently, though, this bug was a necessary part of the workflow for zero-power cases. This was discovered and documented here:
https://github.com/terrapower/armi/pull/749#issuecomment-1179253196

This PR handles the needs of zero-burnup cases by explicitly using the full rated power. This is consistent with what used to happen before #749 was "fixed".

I'm not quite sure how to add a unit test for this one, because we don't seem to have any infrastructure right now for testing out the operator as it steps through time. But our internal integration tests caught this issue in the first place, so I guess that's good enough for now.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/compareDB3.py
+++ b/armi/bookkeeping/db/compareDB3.py
@@ -243,7 +243,7 @@ def _compareTimeStep(
         )
 
     for aux in auxData:
-        _compareAuxData(out, refGroup[aux], srcGroup[aux], diffResults, exclusions)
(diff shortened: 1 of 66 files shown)


john-science (approved):
This would be very hard to test currently, I agree. We can move-on without testing.

My take on it is that the `Operator` class was not designed to be testable. Which is a design/architecture issue which we shouldn't blame you for, or make you fix.

Issue #750, opened 2022-07-05
removeAssembliesInRing() broken


alexhjames:
When running `removeAssembliesInRing()` on a 12 ring reactor to remove all rings that are outside ring 6, I get a varying levels of rings removed. Most of the reactor has 8 rings, the 30/90/etc planes have 7 rings and the 60/120/etc planes have 9 rings left. The order of ring removal does not affect the final result (ie removing ring 12, 11, 10... vs 7, 8, 9...). 

For example, neither of these end in the correct output: 

`for ring in range(7, reactorData.r.core.getNumRings() + 1):

           reactorData.r.core.removeAssembliesInRing(ring)`

`nRings = reactorData.r.core.getNumRings()
for ring in range(nRings, 7 - 1, -1):
    reactorData.r.core.removeAssembliesInRing(ring)`

The following code does what is intended without the function: 
`for a in reactorData.r.core.getAssemblies():
            ring, _ = a.parent.spatialGrid.getRingPos(a.spatialLocator)
            if ring > 6:
                reactorData.r.core.removeAssembly(a, discharge=False)`

john-science:
Okay, I just tried this in ARMI in `test_reactors.py`, and it worked exactly as expected:

```python
    def test_removeAssembliesInRing(self):
        self.assertEqual(self.r.core.getNumRings(), 9)
        for ring in range(7, self.r.core.getNumRings() + 1):
            self.r.core.removeAssembliesInRing(ring)

        self.assertEqual(self.r.core.getNumRings(), 6)
```

It started off with 9 rings. I remove 7, 8, and 9, as per the solution you said failed. And there were 6 left. 

Hmm.

ntouran:
After talking to @john-science, I think this is related to circular vs. hex rings as controlled by the `circularRingMode` setting. It seems that the code is removing "circular rings" (rings of hexes determined by looking underneath a circle placed over the hex core) rather than hex rings. You want to remove hex rings, not circular rings, but circular rings are still active for your case to get nice circular shuffling. Maybe having a new optional parameter to force hex ring mode on `removeAssembliesInRing` would be enough to help you. Then you could run e.g. 

```python
for ring in range(7, reactorData.r.core.getNumRings() + 1):
    reactorData.r.core.removeAssembliesInRing(ring, forceHexRing=True)
```

and get the behavior you expect?

PR #765, opened 2022-07-13
Fixing removeAssembliesInRing for circular hex reactors

john-science:
## Description

There is some confusing language in ARMI that you can remove a "circular" ring of assemblies from a "hexagonal" reactor.

While the method `removeAssembliesinSquareOrHexRing()` does exist, it seems likely that people will keep trying to use `removeAssembliesInRing()` and running into the problem in [this ticket](https://github.com/terrapower/armi/issues/750).

As such, this PR fixes the issue.

---

## Checklist

### DIFF
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -435,12 +435,26 @@ class Core(composites.Composite):
         else:
             self._removeListFromAuxiliaries(a1)
 
-    def removeAssembliesInRing(self, ringNum):
+    def removeAssembliesInRing(self, ringNum, forceHexRing=False):
         """
         Removes all of the assemblies in a given ring
+
+        Parameters
+        ----------
+        ringNum : int
+            The ring to remove
+
+        forceHexRing : bool, optional
+            If you want to remove a circular ring from a square or hex reactor.
         """
-        for a in self.getAssembliesInRing(ringNum):
+        if forceHexRing:
+            assems = self.getAssembliesInSquareOrHexRing(ringNum)
+        else:
+            assems = self.getAssembliesInRing(ringNum)
+
+        for a in assems:
             self.removeAssembly(a)
+
         self.processLoading(settings.getMasterCs())
 
     def _removeListFromAuxiliaries(self, assembly):
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -445,7 +445,8 @@ class Core(composites.Composite):
             The ring to remove
 
         forceHexRing : bool, optional
-            If you want to remove a circular ring from a square or hex reactor.
+            True ~ if you want to remove assemblies in square or hex ring from a squre or hex reactor.
+            False ~ if you want to remove assemblies in a circular ring.
         """
         if forceHexRing:
             assems = self.getAssembliesInSquareOrHexRing(ringNum)
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -435,7 +435,7 @@ class Core(composites.Composite):
         else:
             self._removeListFromAuxiliaries(a1)
 
-    def removeAssembliesInRing(self, ringNum, forceHexRing=False):
+    def removeAssembliesInRing(self, ringNum, overrideCircularRingMode=False):
         """
         Removes all of the assemblies in a given ring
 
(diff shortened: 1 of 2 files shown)


john-science:
@alexhjames  Just FYI.

armi/reactor/tests/test_reactors.py
```diff
@@ -725,6 +725,21 @@ def test_removeAssembliesInRing(self):
             self.assertNotEqual(aLoc[i], a.spatialLocator)
             self.assertEqual(a.spatialLocator.grid, self.r.core.sfp.spatialGrid)
 
+    def test_removeAssembliesInRingByCount(self):
```
alexhjames, line 728: This test appears to use the same test reactor as `test_removeAssembliesinRingHex`. Since it's a hex reactor, won't the call to `removeAssembliesInRing` run into the same error as the original ticket since the flag isn't being used?

armi/reactor/reactors.py
```diff
@@ -435,12 +435,26 @@ def removeAssembly(self, a1, discharge=True):
...
         """
         Removes all of the assemblies in a given ring
+
+        Parameters
+        ----------
+        ringNum : int
+            The ring to remove
+
+        forceHexRing : bool, optional
+            If you want to remove a circular ring from a square or hex reactor.
```
alexhjames, line 448: As a user, I find this sentence confusing. It sounds like based on this sentence that "True" would result in a circular ring being removed from a square or hex reactor which conflicts with the actual function and the variable name. Maybe use instead, "If you want to remove assemblies in square or hex ring from a squre or hex reactor. False removes assemblies in a circular ring"

armi/reactor/tests/test_reactors.py
john-science: I tested this fix outside of ARMI using our proprietary reactor.  But I can't copy that proprietary reactor here, because this is public.  So, without extensive work to produce another test reactor for ARMI, I'm stuck only building unit tests with the test reactor we have.

armi/reactor/reactors.py
john-science: Fixed

armi/reactor/tests/test_reactors.py
alexhjames: I get this unit test does pass, but I guess I'm not sure why it does. You have a hex reactor object that you are removing a single ring from without the flag. This means that it should be removing assemblies in a circular shape from the test reactor and therefore that the number of rings after removing may not match the number of initial rings minus one. I'm guessing this test passes only because the approximately circular shape of this particular number of rings does actually result in the entire outer ring being removed. 

It seems this test should be testing the resulting shape is actually circular. Something along the lines of removing some number of rings and then checking that given assemblies (that we know are in the circle or not) are removed/not removed. The number of rings chosen to remove for the test should make the resulting reactor not have a uniform number of hex rings to differentiate this behavior from the remove hex rings behavior.

armi/reactor/reactors.py
```diff
@@ -435,12 +435,27 @@ def removeAssembly(self, a1, discharge=True):
...
         Removes all of the assemblies in a given ring
+
+        Parameters
+        ----------
+        ringNum : int
+            The ring to remove
+
+        forceHexRing : bool, optional
+            True ~ if you want to remove assemblies in square or hex ring from a squre or hex reactor.
+            False ~ if you want to remove assemblies in a circular ring.
```
ntouran, line 449: Actually on second thought, I think a better arg would be called `overrideCircularRingMode=None` which would just get passed directly down to `getAssembliesInRing`. Then you would also add that same arg to `getAssembliesInRing`, which would do a `is not None` check and if it's not `None` override the `_circularRingMode` param in that method. 

Otherwise we're mixing levels of abstraction in this method. Furthermore, this fix will enable this override to any other client of getAssembliesInring.

armi/reactor/reactors.py
```diff
@@ -435,12 +435,27 @@ def removeAssembly(self, a1, discharge=True):
         else:
             self._removeListFromAuxiliaries(a1)
 
-    def removeAssembliesInRing(self, ringNum):
+    def removeAssembliesInRing(self, ringNum, forceHexRing=False):
         """
         Removes all of the assemblies in a given ring
```
ntouran, line 440: add comment similar to the one in `getAssembliesInRing` or a `See Also` clause to that method where the definition of "ring" is given.

ntouran (changes requested):

armi/reactor/reactors.py
john-science: Done

armi/reactor/reactors.py
john-science: Done

ntouran (approved):
ok that seems fine. I was somehow imagining it'd be possible to override in either direction with one boolean but I guess that's not super doable. I was imagining a three-state boolean of `None`, `True`, `False` but that may be overly complicated.

PR #773, opened 2022-07-15
Clarifying the inputs of plotAssemblyTypes

john-science:
## Description

The function `plotAssemblyTypes()` is now set up to take one of two inputs and still function. But that doesn't really jive with the Python "default" arguments functionality. So I clarify the function here.

NOTE: I also ran `pylint` on the two files I edited for this touched for this PR, and fixed some linting errors.

---

## Checklist

### DIFF
--- a/armi/utils/plotting.py
+++ b/armi/utils/plotting.py
@@ -68,7 +68,6 @@ def colorGenerator(skippedColors=10):
     -----
     Will cycle indefinitely to accommodate large cores. Colors will repeat.
     """
-
     colors = list(mcolors.CSS4_COLORS)
 
     for start in itertools.cycle(range(20, 20 + skippedColors)):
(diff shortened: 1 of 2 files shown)


armi/utils/plotting.py
```diff
@@ -956,6 +958,8 @@ def __init__(
             self.peak = peak
             self.avgHistogram = None
             self.eHistogram = None
+            self.peakHistogram = None
+            self.E = None
```
keckler, line 962: These changes seem random, not sure how you caught that these instance variables were required here.

Obviously nobody has used this plotter in a while...

keckler (approved):
Changes look good.

I would recommend that you update the PR description with the changes you made to `plotBlockFlux` as well, so that it is clear that these were intentional.

armi/utils/plotting.py
john-science: Oh, I just ran `pylint` on the two files I was editing in this PR.  Based on how the code runs, these are not bugs, but it is extremely bad programming practice to define `self.` variables on-the-fly in the code, and not in the `__init__()` method.

PR #776, opened 2022-07-18
Again move axial exp process load

albeanth:
Kudos to @john-science for the solid suggestion on a Friday afternoon!

It's not ideal to require the axial expansion changer to _always_ initiate when blueprints are read. It's a very `Core` specific task and should live closer to where it is appropriately used. E.g., a user may not always define/want a `Core` in their blueprints (e.g., spent fuel pool analyses). So this commit moves it off of `armi/reactor/blueprints/__init__.py` and onto `armi/reactor/reactors.py::Core::processLoading` - and even more specifically, only if `inputHeightsConsideredHot = False`. 

## Checklist

### DIFF
--- a/armi/reactor/blueprints/__init__.py
+++ b/armi/reactor/blueprints/__init__.py
@@ -92,9 +92,6 @@ from armi.reactor import systemLayoutInput
 from armi.scripts import migration
 from armi.utils import textProcessors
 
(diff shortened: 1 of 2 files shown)


albeanth:
@john-science @keckler

john-science (approved):
We can just _move_ it wholesale, without a major redesign? What a win!

PR #777, opened 2022-07-18
Enable Users To Explicitly Specify Target Component for Axial Expansion

albeanth:
## Description
In some cases, the standard [methodology to infer a target component](https://github.com/terrapower/armi/blob/d0123831319c746da4150fa6bedcf540d8010a4f/armi/reactor/converters/axialExpansionChanger.py#L671-L715) for the axial expansion changer may not easily enable specifically what is needed. This update adds a new setting to enable users to explicitly specify the target component for axial expansion on a block.

- Class method name change from `specifyTargetComponent` to `determineTargetComponent`
- Modifies `runLog.debug` statement to specify whether the target component was set or inferred.
- Has a check during block construction to determine if the specified axial expansion target component is valid. 
- Unit test added to support new additions.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -1532,12 +1532,12 @@ class Block(composites.Composite):
 
     def setTargetComponent(self, targetComponent):
         """sets the targetComponent for the axial expansion changer
-        
+
         Parameter
         ---------
         targetComponent: :py:class:`Component <armi.reactor.components.component.Component>` object
             component specified to be target component for axial expansion changer
-        
+
         See Also
         --------
         armi.reactor.converters.axialExpansionChanger.py::ExpansionData::_setTargetComponents
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -97,7 +97,7 @@ class Block(composites.Composite):
 
         self.points = []
         self.macros = None
-        self.targetComponent = None
+        self.axialExpTargetComponent = None
 
         # flag to indicated when DerivedShape children must be updated.
         self.derivedMustUpdate = False
@@ -1530,7 +1530,7 @@ class Block(composites.Composite):
         """
         raise NotImplementedError
 
-    def setTargetComponent(self, targetComponent):
+    def setAxialExpTargetComp(self, targetComponent):
         """sets the targetComponent for the axial expansion changer
 
         Parameter
@@ -1542,7 +1542,7 @@ class Block(composites.Composite):
         --------
         armi.reactor.converters.axialExpansionChanger.py::ExpansionData::_setTargetComponents
         """
-        self.targetComponent = targetComponent
+        self.axialExpTargetComponent = targetComponent
 
     def getPinCoordinates(self):
         """
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -97,6 +97,7 @@ class Block(composites.Composite):
 
         self.points = []
         self.macros = None
+        self.targetComponent = None
 
         # flag to indicated when DerivedShape children must be updated.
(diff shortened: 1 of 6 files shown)


albeanth:
@keckler @john-science @onufer

armi/reactor/blocks.py
```diff
@@ -1529,6 +1530,20 @@ def rotate(self, deg):
         """
         raise NotImplementedError
 
+    def setTargetComponent(self, targetComponent):
```
onufer, line 1533: suggest rename. target component could be different for different types of physics. for depletion target component would be fuel, for assembly distortion in might be some structure. Thermal expansion should not reserve the total namespace of  `setTargetComponent`

Something like `setThermalExpComp` would be better

albeanth: How about `setThermalExpTargetComp`? In the interest of being consistent with the "target component" in the axial expansion changer.

keckler: In addition to changing the name of this method, it might also be useful to change the name of the `self.targetComponent` attribute here and in `blockBlueprint.py`.

keckler (approved):
Looks good. I can also add that I have been unofficially using this feature for a few weeks now, and haven't had any issues.

albeanth:
> Looks good. I can also add that I have been unofficially using this feature for a few weeks now, and haven't had any issues.

Awesome! That's good to hear.

armi/reactor/blocks.py
albeanth: Addressed https://github.com/terrapower/armi/pull/777/commits/293a0ef98b64bb85055814dbbe3923d955123c42

onufer (dismissed):
didn't really review in detail, just minor comment

onufer (approved):

armi/reactor/blocks.py
```diff
@@ -97,6 +97,7 @@ def __init__(self, name: str, height: float = 1.0):
 
         self.points = []
         self.macros = None
+        self.axialExpTargetComponent = None
```
jakehader, line 100: It might be better if this was a parameter that was a string. This will (1) keep attributes stored a block to a minimum and (2) allow target components can be determined on DB loading, (3) allow us to make sure that we can track changes to the parameters as time goes on (code changes) and also will allow us to visually verify the target expansion component with a DB viewer.

armi/reactor/blocks.py
```diff
@@ -1529,6 +1530,20 @@ def rotate(self, deg):
...
+        Parameter
+        ---------
+        targetComponent: :py:class:`Component <armi.reactor.components.component.Component>` object
+            component specified to be target component for axial expansion changer
+
+        See Also
+        --------
+        armi.reactor.converters.axialExpansionChanger.py::ExpansionData::_setTargetComponents
+        """
+        self.axialExpTargetComponent = targetComponent
```
jakehader, line 1545: See comment about the parameter above. I also recommend that we check that on assignment of the value that this against the available components. Additionally, this method can be embedded into the parameters and it can be set of b.p.whatever.

Also, I'd add this check as part of the parameter's `setter`.

jakehader (changes requested):

armi/reactor/blocks.py
albeanth: > I also recommend that we check that on assignment of the value that this against the available components

This occurs in `blockBlueprint.py`. If the component requested isn't available, a Runtime Error is thrown. This could be moved to a parameter's setter.

armi/reactor/blocks.py
jakehader: We are going open a separate issue/PR on this!

jakehader (approved):

PR #778, opened 2022-07-18
Adding axial expansion "support" for unshaped components 

albeanth:
## Description
In some cases, users may specify components within a block that meet the three criteria to check for axial linkage but do not have the class methods `getBoundingCircleOuterDiameter` and/or `getCircleInnerDiameter` -- and it doesn't make physical sense to add them. An example of this is for unshaped components. 

~This change adds a `try/except` clause to account for these types of component classes by addressing a `runLog.warning` and just not linking the components in question, rather than crashing.~

~Note: This could be perceived risky(ish) as some classes (new or existing) that don't have `getBoundingCircleOuterDiameter` and/or `getCircleInnerDiameter` but maybe should. This addition would gloss over this and just issue a `runLog.warning` that could be missed by a user, generating potentially unexpected results.~

Edit (8/4/22):
- After a discussion (see below), it was decided to forgo the try/except route and simply add a conditional to appropriately handle unshaped components only. 

---

## Checklist

    - Will update the "implicitly" included list in https://github.com/terrapower/armi/pull/790 once this gets merged.


armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -506,22 +506,30 @@ def _determineLinked(componentA, componentB):
...
+            smallerOD = min(odA, odB)
+            if biggerID >= smallerOD:
+                # one object fits inside the other
+                linked = False
+            else:
+                linked = True
+        except NotImplementedError:
+            runLog.warning(
+                "Components {0} and {1} do not have a 'getCircleInnerDiameter' and/or getBoundingCircleOuterDiameter method. "
+                "Instead of crashing and raising an error, I'm just going to assume they aren't linked.",
```
keckler, line 529: Not a big deal at all, but my personal preference would be to avoid using first person in the warning.

keckler (approved):
Looks great, and some well-written, concise code.

armi/reactor/converters/axialExpansionChanger.py
albeanth: That's fair! And I don't think it's anywhere else in ARMI so I'll change it to be consistent with everything else. Though I think it's kind of fun when codes spit out first person error/warning messages 😄

albeanth: Addressed https://github.com/terrapower/armi/pull/778/commits/f7321919ca4c7f1b4dd47541d581407d51df6dcf

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -506,22 +506,31 @@ def _determineLinked(componentA, componentB):
...
     ):
-        idA, odA = (
-            componentA.getCircleInnerDiameter(cold=True),
-            componentA.getBoundingCircleOuterDiameter(cold=True),
-        )
-        idB, odB = (
-            componentB.getCircleInnerDiameter(cold=True),
-            componentB.getBoundingCircleOuterDiameter(cold=True),
-        )
+        try:
```
jakehader, line 509: I'd rather that we just explicitly check the instances / types of the components are `Unshaped` and then handle this directly. I don't mind the try / except but that should result in a failure because we should enforce that for other components that the developers do implement `getCircleInnerDiameter` and `getBoundingCircleOuterDiameter`.

Just my thought here so we don't hide future problems leading to some complex / time consuming debugging sessions. Also, does this work if we specify the target components from in #777. Just trying to make sure I understand when this method is called.

jakehader (changes requested):

armi/reactor/converters/axialExpansionChanger.py
albeanth: @jakehader I tend to agree with your sentiment - see the "Note:" in my PR description. 

The blanket try/except was the result of a conversation between @keckler and I. Chris - what are your thoughts on changing the try/except to just handle the unshaped component explicitly? Off the top of my head we could do something like this:

```python
if (
    (componentA.containsSolidMaterial() and componentB.containsSolidMaterial())
    and isinstance(componentA, type(componentB))
    and (componentA.getDimension("mult") == componentB.getDimension("mult"))
):
    if isinstance(componentA, type(UnshapedComponent):
        runLog.warning(
            f"Components {componentA} and {componentB} are UnshapedComponents "
            "and do not have 'getCircleInnerDiameter' or getBoundingCircleOuterDiameter methods; "
            "nor is it physical to do so. Instead of crashing and raising an error, they are going to be assumed to not be linked.",
            single=True,
        )
        linked = False
    else:
        idA, odA = (
            componentA.getCircleInnerDiameter(cold=True),
            componentA.getBoundingCircleOuterDiameter(cold=True),
        )
       ...
```

keckler: I personally prefer it the way that it is currently implemented. My reasoning for this is that it will not always make sense for every component type to have the `getCircleInnerDiameter` and/or `getBoundingCircleOuterDiameter` methods. Right now we are thinking only about `UnshapedComponent`s, but I think there are other components that also wouldn't have those methods, like any 3D shapes (which we probably wouldn't want to be linked ever anyways).

So, in my opinion, I think the current `try` / `except` makes more sense because it is more general. We give a warning, so what more can we really do without being overly restrictive?

albeanth: Is it too bulky/unmaintainable to do something like this for future component types?
```python 
if (
      isinstance(componentA, type(UnshapedComponent)
      or isinstance(componentA, type(NewFutureComponent1)
      or isinstance(componentA, type(NewFutureComponent2)
    ):
        runLog.warning( ...
``` 
This could definitely be cleaner, but is a potential jumping off point for discussions. 

But I see your point @keckler . I guess it comes down to two options:
1. we leave the try/except with a runLog.warning and rely on users to recognize it/check the warnings at the end of an ARMI stdout.
2. we do the above and have ARMI throw a Runtime Error if the new component isn't supported. Then either they can modify the source as needed and create a PR or create an issue for one of us to fix it. 

@john-science do you have a preference? Or maybe a secret third option?

jakehader: I would like the framework to fail fast and fail early, so I am afraid of having generic exception handling. We have done this in the past and in many ways results in a lot of time to debug issues or it masks bugs that may already exist today! In the spirit of keeping this PR to one solution for one issue I am still recommending this. I am interested to hear your take on this though @keckler and of course @john-science.

keckler: I definitely understand what you're saying. 

The main reasoning behind my opinion is that I feel the current implementation is more in line with ARMI being a "framework" that anybody could use for any use-case, as opposed to a tool that is tailored specifically to our needs at TP. In this context, I think adding a clause to an if-statement for the specific case of an `UnshapedComponent` is just very overly-specific and not in the spirit of a "framework", because it `UnshapedComponent`s are just one use-case out of many potential use-cases.

That is my opinion, but I do not need to die on this sword. So I won't argue if we go with @jakehader 's suggestion, which it sounds like we will be doing. It still fixes my current problem, so I can't complain.

jakehader: I understand. I think we have been bitten in the past with over-generalizing too soon and that is my main concern here. I am not sure which issues that will pop-up and masking them with unbounded `try/excepts` has caused issues in the past. @albeanth, if you can implement my suggestion then I will approve.

albeanth: Addressed https://github.com/terrapower/armi/pull/778/commits/73469189bb1feec05fd54a73caad769abc55de0a

jakehader (approved):

jakehader:
@john-science you're up when you have some time!

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -634,6 +634,12 @@ def test_isFuelLocked(self):
             the_exception = cm.exception
             self.assertEqual(the_exception.error_code, 3)
 
+    def test_determineLinked(self):
```
john-science, line 637: Can we also add one more `assert` here that tests what happens if one `Component` is unshaped but one _has_ a shape?

It's just nice to test all the permutations.

And if you're thinking "but that's not a concern with _my_ code", consider it future-proofing for 4 years from now when someone _else_ makes a change to this feature.  :)

john-science:
> @john-science you're up when you have some time!

While the _need_ behind the feature is quite interesting, the code is very simple.  I requested a minor addition to the testing, but then it looks good to me.

armi/reactor/converters/tests/test_axialExpansionChanger.py
albeanth: Addressed https://github.com/terrapower/armi/pull/778/commits/6438d7540afd2e7ccb839d2c7b7735a91946f813

john-science (approved):

PR #786, opened 2022-07-20
fix database compare for numpy arrays of different length

onufer:
## Description

---

## Checklist


opotowsky (commented):
The test case with diffs in `test_comparedb3.py` doesn't cover the possibility of two different length attributes, which I'm looking into

opotowsky:
Adding some context to this PR: 

The code didn't anticipate a  specialData situation with two different length numpy arrays. This code handles that situation rather than causing a failure.

PR #787, opened 2022-07-20
Thermal expansion update

albeanth:
## Description
An update to thermal expansion within the axial expansion changer:
- demonstrates conservation with thermal expansion/contraction at both hot and cold temperatures
    - the thermal expansion factors (and block heights they were relative to) previously were incorrect. This PR resolves this. 
- removes unused class methods that are _very_ heavy to unit test. They just aren't worth keeping around and lead to more confusion then they're worth. 
- make `manageCoreMesh` public, so that users can unify the axial mesh that results from expanding a core (if desired and if `detailedAxialExpansion: False`. 

---

## Checklist

    - Will update the "implicitly" included list in #790  once this gets merged.


armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -623,6 +574,7 @@ def mapHotTempToComponents(self, tempGrid, tempField):
```
jakehader, line 574: This is outside the changes here, but I would appreciate a rename of this method from `mapHotTempToComponents` to `updateComponentTempsBy1DTempField` or something like that. The reason for this is that it makes it a bit more clear to me that we are supplying a 1D temperature field to the method and then that is being used to apply some average temperature to all the components in a block.

This would also allow us to expand this later to add `updateComponentTempsByComponentTempField` or something of that nature that could directly accept temperatures that we would assign to each component. The use case here is if we have a detailed thermal-hydraulic solver (or some mechanism) to get component-level temperatures it would be better to allow direct temperature assignments to the components rather than enforcing that all the components in a block have the same average temperature.

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -326,7 +284,91 @@ def setUp(self):
...
+                        "Current temp is {1:.5f} and block height is {2:.5f}".format(
+                            old, temp, new
+                        ),
+                    )
+
+    def test_HotThermalExpansionContractionConservation(self):
+        """thermally expand and then contract to ensure original state is recovered
+
+        Notes:
+        - temperature field is isothermal and initially at 25 C
```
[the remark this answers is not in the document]
albeanth: Addressed https://github.com/terrapower/armi/pull/787/commits/5b59ae9e7042f55cfd2ba0ad92d3004d2b1b80b3

jakehader (changes requested):
Thanks @albeanth - I will pull this code down and test it out now.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -636,7 +588,12 @@ def computeThermalExpansionFactors(self):
 
         for b in self._a:
             for c in b:
-                self._expansionFactors[c] = c.getThermalExpansionFactor() - 1.0
+                if self._oldHotTemp:
```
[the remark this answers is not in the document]
albeanth: Addressed https://github.com/terrapower/armi/pull/787/commits/74b87b4bd5fc2fa7b9a09f70cd09fe9b52e1cf69

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -534,6 +484,7 @@ class ExpansionData:
 
     def __init__(self, a, setFuel):
         self._a = a
+        self.oldHotHeight = {}
```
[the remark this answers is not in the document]
albeanth: Addressed https://github.com/terrapower/armi/pull/787/commits/74b87b4bd5fc2fa7b9a09f70cd09fe9b52e1cf69

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -183,6 +183,8 @@ def axiallyExpandAssembly(self, thermal: bool = False):
                             c, growFrac
                         )
                     )
+                    if thermal and self.expansionData.oldHotHeight:
+                        blockHeight = self.expansionData.oldHotHeight[c]
```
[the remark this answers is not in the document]
albeanth: Addressed https://github.com/terrapower/armi/pull/787/commits/74b87b4bd5fc2fa7b9a09f70cd09fe9b52e1cf69

armi/reactor/converters/axialExpansionChanger.py
jakehader: Sounds good

jakehader (approved):

jakehader:
Going to wait for @john-science to review before merging. This seems to work for thermal contract and expansion now.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -534,7 +485,8 @@ class ExpansionData:
 
     def __init__(self, a, setFuel):
         self._a = a
-        self._oldHotTemp = {}
+        self.componentReferenceHeight = {}
+        self.componentReferenceTemperature = {}
```
john-science, line 489: 1. Are these two dictionaries filled out during every run, or just during Axial Expansion runs?
2. How safe are they?  Could the height of these blocks get changed due to a non-axial-thermal-expansion method, during a run, and invalidate these dicts?

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -623,20 +575,32 @@ def mapHotTempToComponents(self, tempGrid, tempField):
...
-                # and ruins mass conservation via number densities. Instead,
-                # set manually.
+                # and ruins mass conservation via number densities. Instead, set manually.
                 c.temperatureInC = blockAveTemp
 
     def computeThermalExpansionFactors(self):
         """computes expansion factors for all components via thermal expansion"""
 
         for b in self._a:
             for c in b:
```
john-science, line 594: So, what happens if a component/block get removed from the Reactor?

For instance, if an entire assembly is moved to the Spent Fuel Pool (SFP).

Will further calls to this code still work?  That is, will we get wrong data returned?  If someone adds a NEW assembly to the Reactor to replace the one in the SFP, will we see that new Assembly here?

john-science:
The change itself seems solid enough.

I'm a little wary of keeping a BIG dictionary around with every Component in the Reactor in it.  Maybe it's necessary, but we certainly don't want to be duplicating that much memory.

More importantly, since this is essentially a shadow Reactor object (an entire copy of all of the Reactor components, just stored differently), I am a little concerned that this will lead to wild complications. In that any change elsewhere in the code will have to be mimicked here.

For instance, I mention moving an assembly to the Spent Fuel Pool (SFP).  When someone does that (and they do), what happens to these dictionaries?  Do we need to go find all places in ARMI and the downstream repos that deal with SFPs and make changes to support this shadow Reactor?

I'm just talking here. This might be a smaller feature than I think it is.

armi/reactor/converters/axialExpansionChanger.py
albeanth: > Are these two dictionaries filled out during every run, or just during Axial Expansion runs?

They are initialized and filled out when an axial expansion instance is created. So only during axial expansion runs. 

> Could the height of these blocks get changed due to a non-axial-thermal-expansion method, during a run, and invalidate these dicts?

The heights and temps that get stored in these dicts can certainly change. And that's expected. 

The class `ExpansionData` is only used within the `AxialExpansionChanger` class; so when an instance of the axial expansion changer goes out of scope, so does the expansion data instance (at least that's my intent). 

These dicts are designed to store the "previous" heights and temps (usually from a previous time/burn step or iteration from a coupled solve). So if you were to iterate on thermal hydraulics, fuel performance (e.g., burnup-based expansion), and neutronics for a coupled calculation, every time the axial expansion changer would fire up the "old" temps and heights would get stored here.

armi/reactor/converters/axialExpansionChanger.py
albeanth: I wouldn't expect it to be a problem. The axial expansion changer is designed to live for a short period of time - only as long as it's needed. For instance, in `Core::processLoading` it goes out of scope at the conclusion of the method and is released from memory. 

So at BOL during `Core::processLoading` we fire up an instance of the `AxialExpansionChanger`, loop over each assembly in the core, assign an assembly to the instance of the changer, expand it, and then move on to the next. Each time an assembly is set to the changer, all of the data on the changer is reset for that specific assembly.

The `AxialExpansionChanger` is designed to be blind to swapping of assemblies (e.g., shuffling). So if were to do some shuffle at some arbitrary point during a run, we could follow the same process to loop over the core and expand each assembly in the Core.

albeanth:
> I'm a little wary of keeping a BIG dictionary around with every Component in the Reactor in it. Maybe it's necessary, but this is essentially a shadow Reactor object (an entire copy of all of the Reactor components, just stored differently).

This is a valid concern, I think. The purpose of these dictionaries is to just store heights and temps, but by using the `Component`s as the keys, this probably is way heavier than it needs to be. I wonder if there is a way that we can maintain the same behavior but with a lighter key...? 🤔 

Once consolation is that this method isn't called very often. The dictionaries are initialized each time the `AxialExpansionChanger` is fired up, but they aren't filled unless `updateComponentTempsBy1DTempField` is called... But that certainly doesn't excuse it...

armi/reactor/converters/axialExpansionChanger.py
john-science: So, is it true to say that after Axial Expansion is done, these dicts get deleted?

The situation we don't want is if these dicts were to persist between cycles in a long simulation.

We've seen this before, if you have a reference to every Block/Component/Whatever in the Reactor that persists after a Cycle is done, then you have a memory leak, and memory will just slowly go up over time. Does that make sense?

armi/reactor/converters/axialExpansionChanger.py
john-science: Okay, I see that now:

https://github.com/terrapower/armi/blob/c2d6c767ef0f5e95c76d9d3b21b03eab1d7d4636/armi/reactor/reactors.py#L2243-L2248

So, I don't have to worry about persistence.

Thanks!

john-science (approved):

jakehader:
Go for the merge if you are good @john-science! Thanks for opening that issue @albeanth

Issue #788, opened 2022-07-20
test_comparedb3.py has very few unit tests


opotowsky:
It has a few small unit tests, and two integration tests that don't touch all possible cases. Unit tests for all the other functions are needed. 

This was discovered when the need for PR #786 was discovered.

john-science:
We have raised the code coverage from 60% to 83%:

https://coveralls.io/builds/51061594/source?filename=armi%2Fbookkeeping%2Fdb%2FcompareDB3.py

I think most of what is left are exceptions and errors.  Which we have not done a lot of testing for yet in ARMI.  Sadly.

john-science:
This file is at 85% now, and I think the tests look pretty good. (Obviously, I will want them above 90% one day soon.)

But I am just going to close this ticket.

PR #789, opened 2022-07-21
Adding unit tests for compareDB utils

john-science:
## Description

That is in response to #788 .

This is a cleanup PR, which does two things:

* Adding code coverage to compareDB utils.
* Python cleanup for recent PRs.

---

## Checklist

- [X] This PR has only one purpose or idea.  (Cleanup!)

- [X] The [release notes](https://terrapower.github.io/armi/release/index.html) are up-to-date with any bug fixes or new features.

### DIFF
--- a/armi/bookkeeping/db/tests/test_comparedb3.py
+++ b/armi/bookkeeping/db/tests/test_comparedb3.py
@@ -198,7 +198,7 @@ class TestCompareDB3(unittest.TestCase):
             srcData3.attrs["2"] = 22
             srcData3.attrs["numDens"] = a2
 
-            # there should be no difference
+            # there should be no difference, and a logged error
             with mockRunLogs.BufferLog() as mock:
                 _diffSpecialData(refData, srcData3, out, dr)
                 self.assertEqual(dr.nDiffs(), 0)
@@ -234,7 +234,7 @@ class TestCompareDB3(unittest.TestCase):
         srcData3.attrs["2"] = 22
         srcData3.attrs["numDens"] = a2
 
-        # there should be no difference
+        # there should be a small difference
         _diffSimpleData(refData, srcData3, dr)
         self.assertEqual(dr.nDiffs(), 3)
 
--- a/armi/bookkeeping/db/tests/test_comparedb3.py
+++ b/armi/bookkeeping/db/tests/test_comparedb3.py
@@ -198,7 +198,7 @@ class TestCompareDB3(unittest.TestCase):
             srcData3.attrs["2"] = 22
             srcData3.attrs["numDens"] = a2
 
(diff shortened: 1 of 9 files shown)


john-science:
@opotowsky  Not sure if I should just open this PR, or add more tests first.

armi/bookkeeping/db/tests/test_comparedb3.py
```diff
@@ -158,6 +165,78 @@ def test_compareDatabaseSim(self):
...
+
+        # spin up a different size example H5 Dataset
+        f3 = h5py.File("test_diffSimpleData3.hdf5", "w")
+        a2 = np.arange(90, dtype="<f8")
+        srcData3 = f3.create_dataset("numberDensities", data=a2)
+        srcData3.attrs["1"] = 1
+        srcData3.attrs["2"] = 22
+        srcData3.attrs["numDens"] = a2
+
+        # there should be no difference
```
opotowsky, line 236: Update comment, since nDiffs gives 3

armi/bookkeeping/db/tests/test_comparedb3.py
```diff
@@ -158,6 +165,78 @@ def test_compareDatabaseSim(self):
...
+            a2 = np.arange(90, dtype="<f8")
+            srcData3 = f3.create_dataset("numberDensities", data=a2)
+            srcData3.attrs["1"] = 1
+            srcData3.attrs["2"] = 22
+            srcData3.attrs["numDens"] = a2
+
+            # there should be no difference
+            with mockRunLogs.BufferLog() as mock:
+                _diffSpecialData(refData, srcData3, out, dr)
+                self.assertEqual(dr.nDiffs(), 0)
```
opotowsky, line 203: Now that I'm thinking about it, it's weird that this doesn't first add a diff to the count. Not proposing that the code change, just mentioning it is counterintuitive (in hindsight)

opotowsky (approved):
I'm fine re-reviewing if you add more tests, but these are just an absolutely lovely start to improving the testing!

armi/bookkeeping/db/tests/test_comparedb3.py
john-science: Fixed

armi/bookkeeping/db/tests/test_comparedb3.py
john-science: Yeah, what we get here is a logged error, and the method executes.

Essentially, the logged error says it can't _do_ any comparison. Which means the comparison count it lower than it should be. But... you're not wrong.

opotowsky: true true. The error should be enough

PR #783, opened 2022-07-19
fix snap list for blueprints assemblies

onufer:
## Description

Fixing snap list for blueprints assemblies

---

## Checklist

- [x] The code is understandable and maintainable to people beyond the author.
- [x] Tests have been added/updated to verify that the new or changed code works.
- [x] There is no commented out code in this PR.
- [x] The commit message follows [good practices](https://terrapower.github.io/armi/developer/tooling.html).
- [x] All docstrings are still up-to-date with these changes.

PR #797, opened 2022-07-22
Clean up axial expansion in core.processLoading for DB load

albeanth:
## Description
This addresses two bugs in `armi/reactor/reactors.py::Core:processLoading`. 

1. When loading from a database the axial expansion changer would be instantiated every time `Core::processLoading` is called. It should only be called when processing blueprints at BOL of an ARMI run. 
      - This was fixed by adding an optional parameter to `Core::processLoading` to recognize if it is being called from a database load or from blueprints processing. 
3. Axial expansion of the assemblies at BOL was occurring too early. Axial expansion was occurring before `makeAxialSnapList` was called on each assembly so there was no way to unify the disjoint axial mesh after the expansion. This meant that `inputHeightsConsideredHot: False` could only be used with `detailedAxialExpansion: True`. 
      - This was fixed by simply moving the call to the axial expansion changer to after `makeAxialSnapList` is called. 

---

## Checklist

### DIFF
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -2236,17 +2236,6 @@ class Core(composites.Composite):
         updateAxialMesh : Perturbs the axial mesh originally set up here.
 
         """
-        if not cs["inputHeightsConsideredHot"]:
-            runLog.header(
-                "=========== Axially expanding all (except control) assemblies from Tinput to Thot ==========="
-            )
-            axialExpChngr = AxialExpansionChanger(cs["detailedAxialExpansion"])
-            for a in self.getAssemblies(includeAll=True):
-                if not a.hasFlags(Flags.CONTROL):
-                    axialExpChngr.setAssembly(a)
-                    axialExpChngr.expansionData.computeThermalExpansionFactors()
-                    axialExpChngr.axiallyExpandAssembly(thermal=True)
-
         runLog.header(
             "=========== Initializing Mesh, Assembly Zones, and Nuclide Categories =========== "
         )
@@ -2299,6 +2288,18 @@ class Core(composites.Composite):
             for a in self.parent.blueprints.assemblies.values():
                 a.makeAxialSnapList(refAssem=finestMeshA)
 
+        if not cs["inputHeightsConsideredHot"]:
+            runLog.header(
+                "=========== Axially expanding all (except control) assemblies from Tinput to Thot ==========="
+            )
+            axialExpChngr = AxialExpansionChanger(cs["detailedAxialExpansion"])
+            for a in self.getAssemblies():
+                if not a.hasFlags(Flags.CONTROL):
+                    axialExpChngr.setAssembly(a)
+                    axialExpChngr.expansionData.computeThermalExpansionFactors()
+                    axialExpChngr.axiallyExpandAssembly(thermal=True)
+            axialExpChngr._manageCoreMesh(self.parent)
+        
         self.numRings = self.getNumRings()  # TODO: why needed?
 
         self.getNuclideCategories()
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -2219,7 +2219,7 @@ class Core(composites.Composite):
             fuelBottoms.append(fuelHeightInCm)
         return lowestFuelHeightInCm
 
-    def processLoading(self, cs):
+    def processLoading(self, cs, dbLoad: bool = False):
         """
         After nuclide densities are loaded, this goes through and prepares the reactor.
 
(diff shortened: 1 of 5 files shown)


john-science:
I note with some interest that a different PR very recently touched processLoading:

https://github.com/terrapower/armi/pull/783/files

Is this PR related to that one in some way?

albeanth:
@john-science yeah, they are related. This PR has two actions:
1. A slight refactor and cleanup of the code introduced in `reactors.py::processLoading` #783. The optional parameter `dbLoad: bool` was included to split off the tasks specific to loading database files and creating Cores from blueprints.
2. A bug fix in axial expansion during process loading where 1) it was occurring at all during database loading, and 2) too early for regular ARMI runs. This was not addressed in #783, but included here for completeness.

onufer:
@john-science, yeah after writing #783, i realized there would be bugs when we had a database load and asked @albeanth  to look at it

onufer (changes requested):
I need to read through this a little more too 3x check it makes sense, but its looking good.

armi/reactor/reactors.py
```diff
@@ -2262,42 +2251,54 @@ def processLoading(self, cs):
...
-        if not self.p.referenceBlockAxialMesh:
+        # TODO: There is an inconsistency with makeAxialSnapList when detailedAxialExpansion: False
+        # In a standard run from the blueprints:
+        #     - All assemblies (blueprints included) have their b.p.topIndex
+        #       set in reference to self.refAssem.getAxialMesh().
+        # In a database load run:
+        #     - All assemblies in the core have their b.p.topIndex as set in the standard run.
+        #     - The blueprints assemblies have the b.p.topIndex set in reference to the assembly with the finest mesh.
+        #       This guarantees functionality of makeAxialSnapList, but, if the assembly with the finest mesh is
+        #       not a fuel assembly, assembly.setBlockMesh may produce unexpected results.
```
onufer, line 2262: but if finest mesh is not fuel, detailedAxialExpansion wont work anyway since all assemblies must have submesh of ref assembly. I think this inconsistency is outside the scope of this PR and more to do with determining the refAssembly by the centermost assembly. Lets make an issue describing that we need a better way to determine ref assembly and delete this todo, or at least reduce it.

armi/reactor/reactors.py
```diff
@@ -2262,42 +2251,54 @@ def processLoading(self, cs):
...
+        if dbLoad:
+            # reactor.blueprints.assemblies need to be populated
+            # this normally happens during armi/reactor/blueprints/__init__.py::_prepConstruction
+            # but for DB load, this is not called so it must be here.
+            # pylint: disable=protected-access
+            self.parent.blueprints._prepConstruction(cs)
+            if not cs["detailedAxialExpansion"]:
+                # Apply mesh snapping for self.parent.blueprints.assemblies
+                # To guarantee mesh snapping will function, makeAxialSnapList should be in reference
+                # to the assembly with the finest mesh as defined in the blueprints.
```
onufer, line 2272: you could further expand in comments that core assemblies have this param stored from the database load.

armi/reactor/reactors.py
```diff
@@ -2262,42 +2251,54 @@ def processLoading(self, cs):
...
+                runLog.header(
+                    "=========== Axially expanding all (except control) assemblies from Tinput to Thot ==========="
+                )
+                axialExpChngr = AxialExpansionChanger(cs["detailedAxialExpansion"])
+                for a in self.getAssemblies():
+                    if not a.hasFlags(Flags.CONTROL):
+                        axialExpChngr.setAssembly(a)
+                        axialExpChngr.expansionData.computeThermalExpansionFactors()
+                        axialExpChngr.axiallyExpandAssembly(thermal=True)
+                axialExpChngr._manageCoreMesh(self.parent)
```
onufer, line 2301: this seems like it needs to be a public method if called in this way.

Do you think it would be reasonable to move 2295-2301 into a new method AxialExpansionChanger  called `applyThermalExpansion` or something like that so that there is an easy way to apply it and you don't have to worry about calling the manageMesh method?

onufer (changes requested):

armi/reactor/reactors.py
```diff
@@ -2262,42 +2251,54 @@ def processLoading(self, cs):
...
+        #     - All assemblies (blueprints included) have their b.p.topIndex
+        #       set in reference to self.refAssem.getAxialMesh().
+        # In a database load run:
+        #     - All assemblies in the core have their b.p.topIndex as set in the standard run.
+        #     - The blueprints assemblies have the b.p.topIndex set in reference to the assembly with the finest mesh.
+        #       This guarantees functionality of makeAxialSnapList, but, if the assembly with the finest mesh is
+        #       not a fuel assembly, assembly.setBlockMesh may produce unexpected results.
+        if dbLoad:
+            # reactor.blueprints.assemblies need to be populated
+            # this normally happens during armi/reactor/blueprints/__init__.py::_prepConstruction
```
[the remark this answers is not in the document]
albeanth: Addressed https://github.com/terrapower/armi/pull/797/commits/7df5b13b56fe5ec2afb5ffa07487001a816833d2

armi/reactor/reactors.py
albeanth: Removed TODO here https://github.com/terrapower/armi/pull/797/commits/7df5b13b56fe5ec2afb5ffa07487001a816833d2

Will create issue to document the inconsistency and if/how we want to go about resolving it.

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/797/commits/7df5b13b56fe5ec2afb5ffa07487001a816833d2

armi/reactor/reactors.py
albeanth: > this seems like it needs to be a public method if called in this way.

Good catch. Changing `_manageCoreMesh` from private to public was coming in #787 , but will be included here as well. In https://github.com/terrapower/armi/pull/797/commits/7df5b13b56fe5ec2afb5ffa07487001a816833d2 I change it to be a public facing method.

> Do you think it would be reasonable to move 2295-2301 into a new method AxialExpansionChanger called applyThermalExpansion or something like that so that there is an easy way to apply it and you don't have to worry about calling the manageMesh method?

So this is certainly possible - and actually sort of exists in `axiallyExpandCoreThermal`, but is getting removed in #787. Having a core level axial expansion functionality is useful, but I found that it doesn't support general applications very well. E.g., here we want to do thermal expansion based on a Delta T between `Tinput` and `Thot`, but elsewhere someone may want to do thermal expansion based on a different Delta T (e.g., a new `Thot` and previous `Thot`). So then the method we would write for one case would require another method to support it - and core level methods are particularly heavy to get unit testing coverage for. We might(?) be able to configure a core level function that better supports generality, but it would still be really heavy to test.

onufer (approved):

jakehader (approved):

PR #799, opened 2022-07-28
Parameter mapping fix in the uniform mesh converter.

jakehader:
## Description

This adds a fix to the uniform mesh converter where data was being removed from the source reactor if it was not set on the converted reactor. This allows us to preserve the existing data on the blocks and reactor core if it is unchanged from any work that is performed on the converted reactor core.

Additionally, this moves the `axialMesh` parameter on the core outside of a "neutronics" parameter category, as we want to avoid changing this during the transfer of data from the uniform mesh reactor core back to the non-uniform mesh reactor core. 

This is a significant bug that would lead the reactor core state on non-uniform mesh to be inconsistent with the core assembly meshes.

---

## Checklist

Note to @john-science: This mainly has one idea of fixing the bug, but other smaller cleanups were found, in addition to some cross section library reporting, and changes to variable names.


armi/reactor/converters/uniformMesh.py
```diff
@@ -83,17 +85,29 @@ def __init__(self, cs=None):
         self.blockDetailedAxialExpansionParamNames = []
         self.reactorParamNames = []
 
+        # These dictionaries represent back-up data from the source reactor
+        # that can be recovered if the data is not being brought back from
+        # the uniform mesh reactor when ``applyStateToOriginal`` to called.
+        # This prevents clearing out data on the original reactor that should
+        # be preserved since no changes were applied.
```
john-science, line 92: Is there any way to shorten these comments?

People who are stuck trying to fix/write code have to read a lot of stuff really fast, and this will slow people down significantly if they need look at the `uniformMesh` code at all.

armi/reactor/converters/uniformMesh.py
```diff
@@ -83,17 +85,29 @@ def __init__(self, cs=None):
...
         self.convReactor = self.initNewReactor(r)
         self._setParamsToUpdate()
         self._computeAverageAxialMesh()
         self._buildAllUniformAssemblies()
-        self._clearStateOnReactor(self.convReactor)
+
+        # Clear the state of the converted reactor that was just built
+        # to ensure that there is no stale data from creation a new
+        # reactor. The caching is disabled in this case because we don't
+        # care about storing this data since it might be overwritten later.
```
john-science, line 109: Could this be shortened some?

"Clear the state of the converted reactor to ensure there is no data from the creation of a new Reactor"

armi/reactor/converters/uniformMesh.py
```diff
@@ -121,9 +135,14 @@ def initNewReactor(sourceReactor):
         # attributes are set when assemblies are added in coreDesign.construct(), however
         # since we skip that here, they never get set; therefore the need for the deepcopy.
         bp = copy.deepcopy(sourceReactor.blueprints)
-        newReactor = Reactor(sourceReactor.o.cs.caseTitle, bp)
+        newReactor = Reactor(sourceReactor.name, bp)
         coreDesign = bp.systemDesigns["core"]
-        coreDesign.construct(sourceReactor.o.cs, bp, newReactor, loadAssems=False)
+        if sourceReactor.o is None:
+            cs = settings.getMasterCs()
```
john-science, line 141: What changed here?  We didn't used to have to support the case where `sourceReactor` was incomplete.

(If possible, I'm just trying to reduce the dependency on `setMasterCs()` and `getMasterCs()`, that's all.

armi/reactor/converters/uniformMesh.py
```diff
@@ -321,7 +344,15 @@ def applyStateToOriginal(self):
...
         )
         completeStartTime = timer()
-        self._clearStateOnReactor(self._sourceReactor)
+
+        # Clear the state of the original source reactor to ensure that
+        # a clean mapping between the converted reactor for data that has been
+        # changed. In this case, we cache the original reactor's data so that
+        # after the mapping has been applied, we can recover data from any
+        # parameters that did not change.
+        self._cachedReactorCoreParamData = {}
```
john-science, line 353: I'm a little worried that after this uniform mesh work is done, this dictionary will sit around forever with a reference to every Block in the Reactor.

This will amount to a memory leak, during long simulations.

After the mesh conversion is done, can we wipe out all the cache in this class?

armi/reactor/converters/uniformMesh.py
jakehader: Added a comment in the code, but basically we didn't have a test case (integration or otherwise) that tested the use of the uniform mesh converter being chained together with the use of another geometry converter. For example, in most cases we have tested with `detailedAxialExpansion` set to False, but in the case where this is set to True (i.e., some benchmark problems we have) this results in the incoming reactor not having an operator attached. I don't think that it makes a ton of sense for a temporary reactor to _require_ an operator, so this is unfortunately needed at the moment until we can do a large refactor and decoupling of the reactor and the operator. I can make an issue to track this more explicitly, because it seems that the issue isn't present on Github.

armi/reactor/converters/uniformMesh.py
jakehader: Yep! This is now updated. Please see the `reset()` method that is implemented to be called.

john-science:
Do you know if these two changes will break any of our `Plugin`s?

1. Removed `quiet` kwarg from geometry and block converters as this was generally unused in the code base and was only implemented for developmental debugging.
2. Migration of the `axialMesh` core parameter from the `neutronics` category to be general to prevent inconsistencies in the axial core mesh and the assembly meshes.

jakehader:
> Do you know if these two changes will break any of our `Plugin`s?
> 
> 1. Removed `quiet` kwarg from geometry and block converters as this was generally unused in the code base and was only implemented for developmental debugging.
> 2. Migration of the `axialMesh` core parameter from the `neutronics` category to be general to prevent inconsistencies in the axial core mesh and the assembly meshes.

I do not know _for sure_ but we can verify.

PR #796, opened 2022-07-22
Adding post-DB-load hook

john-science:
## Description

There was [a request](https://github.com/terrapower/armi/pull/723#issuecomment-1159285744) that we add a `Plugin` hook for analysis, that would allow users to do something juuuuuuuust after a `Reactor` was loaded from a DB.

---

## Checklist

- [x] This PR has only one purpose or idea.
- [x] Tests have been added/updated to verify that the new/changed code works.

- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html).

- [x] The [release notes](https://terrapower.github.io/armi/release/index.html) are up-to-date with any bug fixes or new features.
- [x] The documentation is still up-to-date in the `doc` folder.
- [x] The dependencies are still up-to-date in `setup.py`.

Issue #772, opened 2022-07-14
copyInterfaceInputs should not require a valid settings object


opotowsky:
The change, made in PR #649, breaks downstream use cases. See the comment below:

> This change seems to have taken away the ability to copy over files which don't correspond to a specific setting. For instance, one may have auxiliary files that support a calculation but are not called out as a specific case setting. 
> 
> It seems that we used to be able to copy such files with `copyInterfaceFiles()` in conjunction with a `specifyInputs()` method on the interface in question. But from the changes here, we can no longer do that I guess.
> 
> Is this intentional? Because it broke some of our downstream projects and we'll need a way around, if so.
> 
> @john-science @jakehader @opotowsky
> 
> _Originally posted by @keckler in https://github.com/terrapower/armi/pull/649#discussion_r920381899_

A little more context:

> Often in our models we have some files that are just a big data dump of things like control rod positions at different times, or fuel shuffling operations for each cycle. This data doesn't readily live anywhere in ARMI, and thus needs to be stored in some "auxiliary" file that is parsed and utilized during the run.
> 
> I suppose, in theory, both of the examples I listed just now could be stored in either the custom controlLogic file or the shuffleLogic file, but it typically doesn't make sense to do that based on the size/format of the data.

PR #802, opened 2022-08-01
copyInterfaceInputs bug fix

opotowsky:
## Description

`copyInterfaceInputs` was not properly handling the use case of a list of filenames. It worked only when a single input file was associated with a setting. 

The primary logic change in this PR is changing the update being a single file: 
```python
newSettings[label] = str(destFilePath)
```
To the option for the update being a list: 
```python
newSettings[label].append(str(destFilePath))
```

Additionally, the handling of wildcards has moved around and a check for that has been implemented by setting `globFilePaths` to `None` and only processing a glob if that variable exists.

I should have code coverage COVERED. Several added tests. 

This function should now be able to handle the updating of a single file (relative or absolute), a list of files (relative or absolute), and a file entry that has a wildcard processing into multiple files. Although, that makes it a bit [smelly](https://en.wikipedia.org/wiki/Code_smell) (Thanks for the reference @sombrereau!).

---

## Checklist

**(Note to PR merger: the first commit message is bad so please grab the first line of the PR description when squash-and-merging instead!)**

### DIFF
--- a/armi/cases/tests/test_cases.py
+++ b/armi/cases/tests/test_cases.py
@@ -22,6 +22,7 @@ import platform
 import pathlib
 
 from armi import cases
+from armi import plugins
 from armi import settings
 from armi.utils import directoryChangers
 from armi.tests import ARMI_RUN_PATH, TEST_ROOT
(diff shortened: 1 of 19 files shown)


jakehader:
@opotowsky, this is a good fix! I went to review but noticed that it was still a draft. 

I know that this issue (https://github.com/terrapower/armi/issues/772) was opened previously, but I'm wondering if it's still an issue that users (i.e., @keckler) are still seeing. I tend to think that we should have `copyInterfaceInputs` be based on already defined case settings, but maybe that is overly restrictive as in some cases we need to bring more than a single file along (thinking custom fuel management or custom control rod management scripts that have other dependencies). 

I like one of the unit tests that @john-science is able to add that demonstrates a custom plugin, so maybe we can create a lightweight plugin with an interface that utilizes this `copyInterfaceInputs` for improved test coverage from end to end. Don't want to detail your progress, but it might be nice to close the loop on that other issue at the same time as this? See - #796

opotowsky:
@jakehader I actually am not addressing that issue in this PR, although it is certainly on my todo list (it became very time intensive). @keckler has a workaround but he convinced me of the need to support non-interface-related file clone support, so I self-assigned the issue since I broke downstream use cases. 

For this PR, I ran into a use case where an interface has a setting that was a list of filenames, which this function implied support for, but did not actually support. 

I thought it was an easy fix but I didn't understand the reasons for the complexity in this function, so it was not in fact simple. It is draft until I get the tests fully figured out, as well as fix some code-quality issues. Thank you for the suggestion, I think it will help with progress and not derail it!

opotowsky:
Marking as draft--again

armi/cases/tests/test_cases.py
```diff
@@ -272,20 +277,159 @@ def test_writeInput(self):
             self.assertTrue(os.path.exists(cs["shuffleLogic"]))
 
 
+class MultiFilesInterfaces(interfaces.Interface):
+    """
+    A mostly meaningless little test interface, just to prove that TODO
```
john-science, line 282: Damn. Sorry, I put a TODO in your code. My bad.

```
A little test interface that adds a setting that we need to test copyInterfaceInputs
with multiple files.
```

john-science: My fault!

opotowsky: Good catch! on it

john-science (approved):
Nice work!

Man, that function is just _too_ complicated.

opotowsky:
> Nice work!
> 
> Man, that function is just _too_ complicated.

My work here is not done. Until tomorrow!

PR #804, opened 2022-08-01
Number density bug fix

albeanth:
## Description
PR #657  introduced a bug into how number densities are initialized for a component.

The bug:
- when `detailedAxialExpansion: True`, number densities were being initialized with input temperatures.

This is incorrect behavior as number densities should be initialized with hot temperatures to account for radial expansion of components. Otherwise, number densities are never adjusted for radial expansion and masses are incorrect. 

This PR fixes this by initializing the number densities with hot temperatures. Accounting for axial expansion via `inputHeightsConsideredHot` is then accounted for in `adjustNDensForHotHeight`. 
- In addition to this fix, I've changed method names and improved docstrings in the interest clarity. 
- Also cleaned up the code within `adjustNDensForHotHeight` to be simpler and more efficient (instead of completely recalculating the number densities previously calculated in `setNDensFromMassFracsAtTempInC`, it now just scales them accordingly. 

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/blockBlueprint.py
+++ b/armi/reactor/blueprints/blockBlueprint.py
@@ -119,12 +119,12 @@ class BlockBlueprint(yamlize.KeyedList):
             if cs["inputHeightsConsideredHot"]:
                 if "group" in c.name:
                     for component in c:
-                        component.applyHotHeightDensityReduction()
+                        component.adjustNDensForHotHeight()
                         componentBlueprint.insertDepletableNuclideKeys(
                             component, blueprint
                         )
                 else:
-                    c.applyHotHeightDensityReduction()
+                    c.adjustNDensForHotHeight()
                     componentBlueprint.insertDepletableNuclideKeys(c, blueprint)
             components[c.name] = c
             if spatialGrid:
--- a/armi/reactor/tests/test_components.py
+++ b/armi/reactor/tests/test_components.py
@@ -493,6 +493,15 @@ class TestCircle(TestShapedComponent):
         self.component.changeNDensByFactor(3.0)
         self.assertEqual(self.component.getNumberDensity("NA23"), 3.0)
 
+    def test_MassConserved(self):
+        """Demonstrate that mass is conserved at different temperatures."""
+        circle1 = Circle("circle", "HT9", 20, 300, 1.0)
+        circle2 = Circle("circle", "HT9", 20, 600, 1.0)
+        self.assertAlmostEqual(
+            circle1.p.numberDensities["FE"] * circle1.getArea(),
+            circle2.p.numberDensities["FE"] * circle2.getArea(),
+        )
+
 
 class TestTriangle(TestShapedComponent):
     componentCls = Triangle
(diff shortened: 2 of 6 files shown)


albeanth:
@mgjarrett

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -119,12 +119,12 @@ def construct(
             if cs["inputHeightsConsideredHot"]:
                 if "group" in c.name:
```
jakehader, line 120: I'd like to see a comment on what "group" in c.name means in this context since I haven't seen a use case of this. If this is assigned by the user then this might be very fragile. If it is set in the code then there might an opportunity to make this into a "Flag" or at least make this a constant rather than a string.

armi/reactor/tests/test_components.py
```diff
@@ -493,6 +493,15 @@ def test_changeNumberDensities(self):
         self.component.changeNDensByFactor(3.0)
         self.assertEqual(self.component.getNumberDensity("NA23"), 3.0)
 
+    def test_MassConserved(self):
```
jakehader, line 496: Can you explain in a bit more detail in the comments of this PR how this test would behave without these changes. Can you also extend this test to more components with different material types, like `Void` or a non-solid like `He` or `Sodium`? Just want to make sure that this test covers what it should and any edge cases

jakehader (changes requested):

armi/reactor/tests/test_components.py
onufer: Before, when input temperature density was used, the nDense * area were not equal. I am updating the the test to show how ARMI is expected to behave for a variety of cases.

This was the initial failing unit test i gave Tony to bug fix based on.

jakehader: Thanks for the addition! Unit tests are failing now because of a name issue but this test is much more detailed. I assume that you modify the ndens of Iron by dividing by the thermal expansion factor because that's emulating what happens on setting a new height if the component had a block it was attached to? That's my read on it and if so then I think this is good.

armi/reactor/blueprints/blockBlueprint.py
albeanth: @jakehader I know we chatted about this offline but I wanted to loop in @ntouran to close the loop. Nick, can you provide context here to resolve Jake's comment?

armi/reactor/tests/test_components.py
onufer: yes, the comment was attempting to clarify how setBlock height modifies ndense of solid components

armi/reactor/blueprints/blockBlueprint.py
jakehader: I think that this is out of scope so we will leave as is

armi/reactor/tests/test_components.py
jakehader: Thanks!

jakehader (approved):

PR #809, opened 2022-08-03
Add try/except around unpackSpecialData

opotowsky:
## Description

`compareDatabases` should be able to continue to run even if `unpackSpecialData` fails. 

Some parameters are not able to be compared, which in a recent scenario led to a failure inside `unpackSpecialData`. This PR adds the try/except so that there is just an error and not a failure, and it also adds a the traceback to the log printout so parameters that cause this step to fail can be more easily addressed / manually compared. 

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/compareDB3.py
+++ b/armi/bookkeeping/db/compareDB3.py
@@ -47,6 +47,7 @@ from typing import Sequence, Optional, Pattern, Tuple
 import collections
 import os
 import re
+import traceback
 
 from tabulate import tabulate
 import h5py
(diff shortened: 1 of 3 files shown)


jakehader:
I don't think im qualified for this review, so i will leave it to @john-science.

john-science (approved):
The change seems good, sure.

But I don't know Why this exception would need to be here.  Can we put something in the PR description, or somewhere, that would explain (roughly) why this would ever fail?

Thanks!  Nice work!

jakehader:
> The change seems good, sure.
> 
> But I don't know Why this exception would need to be here. Can we put something in the PR description, or somewhere, that would explain (roughly) why this would ever fail?
> 
> Thanks! Nice work!

Would including more information in the description help? I think the main thing here is that some parameters are not and to be compared between two dbs and that leads to failures on the compareDB call and then this aborts the process. This aims to note which parameters have issues and then they can be manually compared by an analyst as needed.

john-science:
The code coverage in `comparedb.py` went down some with this PR. We lost three lines of coverage.

I'm trying to decide HOW that happened. Did we exit a function sooner than usual, and lose the coverage after some point?  Hmm...

Could be a fluke that won't happen again, of course.

opotowsky:
> The change seems good, sure.
> 
> But I don't know Why this exception would need to be here. Can we put something in the PR description, or somewhere, that would explain (roughly) why this would ever fail?
> 
> Thanks! Nice work!

Added a little extra detail based on Jake's comment!

Issue #808, opened 2022-08-03
`Block.getReactionRates()` doesn't work for certain isotopes 


keckler:
The `Block.getReactionRates()` method supposedly will allow one to specify an isotope and get the rate of a specified reaction on the given block:
https://github.com/terrapower/armi/blob/69c778934b9e0cfe3751f2efde836126889a1902/armi/reactor/blocks.py#L1489-L1520

If the user does not specify the `nDensity` argument, `Block.getReactionRates()` will attempt to look up the isotope's number density in the block using the `Block.getNumberDensity()` method, which under the hood iteratively calls the `Component.getNumberDensity()` method:
https://github.com/terrapower/armi/blob/69c778934b9e0cfe3751f2efde836126889a1902/armi/reactor/components/component.py#L596-L610

Of note is that this `Component.getNumberDensity()` method will return `0.0` if the `nucName` is not found in the dict.

This is all fine and dandy.

So if you had a block `blockInstance`, had a properly-loaded ISOTXS file, and called `blockInstance.getReactionRates("U235")`, you would get the result you're looking for:
```
In [13]: r.core.getAssemblyWithStringLocation('001-001')[6].getReactionRates('U238')['nF']
Out[13]: 5134734346273870.0
```

BUT, if you instead called the method for a certain subset of isotopes, you might have problems:
```
In [14]: r.core.getAssemblyWithStringLocation('001-001')[6].getReactionRates('PU239')['nF']
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-14-0590e0975867> in <module>
----> 1 r.core.getAssemblyWithStringLocation('001-001')[6].getReactionRates('PU239')['nF']

c:\users\ckeckler\codes\nala\framework\armi\reactor\blocks.py in getReactionRates(self, nucName, nDensity)
   1517                 self.p.xsType,
   1518                 self.getIntegratedMgFlux(),
-> 1519                 nDensity,
   1520             )
   1521         except AttributeError:

c:\users\ckeckler\codes\nala\framework\armi\reactor\components\component.py in getReactionRateDict(nucName, lib, xsType, mgFlux, nDens)
   1239     """
   1240     key = "{}{}A".format(nucName, xsType)
-> 1241     libNuc = lib[key]
   1242     rxnRates = {"n3n": 0}
   1243     for rxName, mgXSs in [

[...]

PR #810, opened 2022-08-03
Fix `Block.getReactionRates()`

keckler:
## Description

This PR fixes the `Block.getReactionRates()` method by switching to the correct key when looking up isotope cross-sections from the ISOTXS file. It also adds tests to prove that it is now working.

This is in response to #808 .

---

## Checklist

### DIFF
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -1217,7 +1217,8 @@ def getReactionRateDict(nucName, lib, xsType, mgFlux, nDens):
     Parameters
     ----------
     nucName - str
-        nuclide name -- e.g. 'U235'
+        nuclide name -- e.g. 'U235', 'PU239', etc. Not to be confused with the nuclide
+        _label_, see the nucDirectory module for a description of the difference.
     lib - isotxs
         cross section library
     xsType - str
@@ -1237,7 +1238,8 @@ def getReactionRateDict(nucName, lib, xsType, mgFlux, nDens):
     assume there is no n3n cross section in ISOTXS
 
     """
-    key = "{}{}A".format(nucName, xsType)
+    nucLabel = nuclideBases.byName[nucName].label
+    key = "{}{}A".format(nucLabel, xsType)
     libNuc = lib[key]
     rxnRates = {"n3n": 0}
     for rxName, mgXSs in [
--- a/armi/reactor/tests/test_blocks.py
+++ b/armi/reactor/tests/test_blocks.py
@@ -1555,6 +1555,29 @@ class Block_TestCase(unittest.TestCase):
         hasPitch = self.block.hasPinPitch()
         self.assertTrue(hasPitch)
 
+    def test_getReactionRates(self):
+        block = blocks.HexBlock("HexBlock")
+        block.setType("defaultType")
+        comp = components.basicShapes.Hexagon("hexagon", "MOX", 1, 1, 1)
+        block.add(comp)
+        block.setHeight(1)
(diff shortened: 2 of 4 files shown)


armi/reactor/components/component.py
```diff
@@ -1217,7 +1217,8 @@ def getReactionRateDict(nucName, lib, xsType, mgFlux, nDens):
     Parameters
     ----------
     nucName - str
-        nuclide name -- e.g. 'U235'
```
jakehader, line 1220: Thanks for this update @keckler! 

Here are a couple comments:

1. Let's change the `xsType` input to be `microSuffix`, so that this encapsulates both the XS Type (i.e., A, B, C) as well as the burnup group (i.e., A, B, C) so that we can actually get reaction rates based on burnup-dependent cross sections.

To do this, we can use `self.getMicroSuffix()`, where on the block level this is implemented as is, but on a component this is implemented as reaching up to its parent and getting it from the block-level.

2. It would be advantageous for us to standardize the dictionary of reaction rates that can be obtained from the nuclide data library - `{"nG": 0, "nF": 0, "n2n": 0, "nA": 0, "nP": 0, "n3n": 0}`. This is a bit of a refactor, but having this mapping of these strings to the following xsLibrary micros would be good:

```
        ("nG", libNuc.micros.nGamma),
        ("nF", libNuc.micros.fission),
        ("n2n", libNuc.micros.n2n),
        ("nA", libNuc.micros.nalph),
        ("nP", libNuc.micros.np),
```

Just grepping around the code base, I see this mapping occurring in `blocks.py/getReactionRates`, `composites.py/getReactionRates`, `component.py/getReactionRateDict`,  `isotopicsDepletionInterface.py/makeXsecTable`, and `crossSectionTable.py/CrossSectionTable`. 

It might make sense to have the strings for `nG`, `nF`, `n2n`, `nA`, `nP`, etc. mapped back to `xsCollections.py` module-level constants that already exist:

```
NGAMMA = "nGamma"      # radiative capture
NAPLHA = "nalph"       # (n, alpha)
NP = "np"              # (n, proton)
ND = "nd"              # (n, deuteron)
NT = "nt"              # (n, triton)
FISSION_XS = "fission" # (n, fission)
N2N_XS = "n2n"         # (n,2n)
NUSIGF = "nuSigF"      
NU = "neutronsPerFission"
```

armi/reactor/components/component.py
```diff
@@ -1237,7 +1238,8 @@ def getReactionRateDict(nucName, lib, xsType, mgFlux, nDens):
     assume there is no n3n cross section in ISOTXS
 
     """
-    key = "{}{}A".format(nucName, xsType)
+    nucLabel = nuclideBases.byName[nucName].label
+    key = "{}{}A".format(nucLabel, xsType)
```
jakehader, line 1242: Based on the comment above, this "key" would be replaced with `f"{nucLabel}{microSuffix}"`

jakehader (changes requested):

armi/reactor/components/component.py
keckler: I agree with all the things you suggest above, but can we possibly make that into a separate PR?

I think technically what I implemented "fixed" the bug, and the things you're suggesting are feature requests.

Just trying to limit the scope of this PR so that I can get it into the codebase more quickly.

jakehader: Sure thing! Ill make an issue.

jakehader (approved):

PR #816, opened 2022-08-04
rewrite of copyInterfaceInputs

opotowsky:
## Description

This is just a ~rewrite of an overly complicated function, `copyInterfaceInputs`. This, for the most part, does not change what it does, but hopefully clears up the use cases it is trying to handle. (Caveat: `copyInputsHelper` is a little different but the overall call to `copyInterfaceInputs` should have the same behavior. And a downstream use-case was behaving undesirably)

A unit test that passed on CI failed for someone locally because of the code's fragility, and so instead of fixing that one thing I burned most/half of it and rewrote it. It cannot ever be simpler because it needs to handle so many file path options (and single files and lists), but hopefully this makes it clearer. 

This just offers a modicum of sanity for future me. 

---

## Checklist

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -853,7 +853,7 @@ def copyInterfaceInputs(
                             f"with a glob of `{sourceFullPath}`. Will not update "
                             f"`{label}`."
                         )
-                        newFiles.append(f)  # pass?
+                        newFiles.append(f)
                     else:
                         for gFile in globFilePaths:
                             destFilePath = copyInputsHelper(
@@ -871,6 +871,7 @@ def copyInterfaceInputs(
                         f"the following path: `{sourceFullPath}`. Will not update "
                         f"`{label}`."
                     )
+
             # Some settings are a single filename. Others are lists of files. Make
             # sure we are returning what the setting expects
             if len(files) == 1 and not WILDCARD:
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -715,7 +715,7 @@ class Case:
             self.cs.writeToYamlFile(self.title + ".yaml")
 
 
-def copyInputsHelper(
+def _copyInputsHelper(
     fileDescription: str,
     sourcePath: str,
     destPath: str,
@@ -856,12 +856,12 @@ def copyInterfaceInputs(
                         newFiles.append(f)
                     else:
                         for gFile in globFilePaths:
-                            destFilePath = copyInputsHelper(
+                            destFilePath = _copyInputsHelper(
                                 label, gFile, destination, f
                             )
                             newFiles.append(str(destFilePath))
                 else:
-                    destFilePath = copyInputsHelper(
+                    destFilePath = _copyInputsHelper(
                         label, sourceFullPath, destination, f
                     )
                     newFiles.append(str(destFilePath))
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -723,7 +723,8 @@ def copyInputsHelper(
 ) -> str:
     """
     Helper function for copyInterfaceInputs: Creates an absolute file path, and
-    copies the file to that location.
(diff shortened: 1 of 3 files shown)


armi/cases/case.py
```diff
@@ -716,31 +716,44 @@ def writeInputs(self, sourceDir: Optional[str] = None):
...
-    return destFilePath
+    sourceName = os.path.basename(sourcePath)
+    destFilePath = os.path.join(destPath, sourceName)
+    try:
+        pathTools.copyOrWarn(fileDescription, sourcePath, destFilePath)
+        if pathlib.Path(destFilePath).exists():
+            return destFilePath
+        else:
+            return origFile
+    except Exception:
```
opotowsky, line 755: Not sure what this should be here. `copyOrWarn` throws its own exception, but I think it's still important to explicitly check `destFilePath` exists. But that can sometimes fail depending on the operating system (see `OSError` in new-line 841.)

armi/cases/case.py
```diff
@@ -716,31 +716,44 @@ def writeInputs(self, sourceDir: Optional[str] = None):
 
 
 def copyInputsHelper(
```
john-science, line 718: Should this be `_copyInputsHelper()`, so that no one tries to use it outside of this file?

Your call, just wondering if it was meant to be public or not.

opotowsky: I don't think it needs to be public, fixing

armi/cases/case.py
john-science: Sometimes on certain really old computing clusters, that will remain nameless, copying a file might take a really long time.  And so checking if the file exists gets hard, because Python might think the copy took a long time, or sometimes Python might think the copy is done a few seconds before it really is.

I think it's clean enough as it is.  At least, I can't think of any clever way to make it cleaner.

opotowsky:
I cannot figure out how I'm decreasing coverage in database3.py?!

opotowsky:
passing coveralls step: CHALLENGE ACCEPTED! (working on adding tests)

opotowsky:
So the coverage for `case.py` went down 0.12%, but the coverage check passed. 

The three lines missing test coverage are all exceptions. One is a check that will be moved elsewhere in the future (See Issue #825). The other two need to be there because while certain lines pass on many systems (in this case, for me locally as well as the CI build), they could cause a failure on other systems.

armi/cases/case.py
```diff
@@ -810,55 +823,55 @@ def copyInterfaceInputs(
...
-                        destFilePath = copyInputsHelper(label, sourceFullPath, destPath)
-                        # Some settings are a single filename. Others are lists of files.
-                        # Either overwrite the empty list at the top of the loop, or
-                        # append to it.
-                        if len(files) == 1:
-                            newSettings[label] = str(destFilePath)
-                        else:
-                            newSettings[label].append(str(destFilePath))
+                path = pathlib.Path(f)
+                if WILDCARD == False and RELATIVE == False:
```
john-science, line 836: I think `pylint` would prefer it if this read:

`if not WILDCARD and not RELATIVE:`

armi/cases/case.py
```diff
@@ -810,55 +823,55 @@ def copyInterfaceInputs(
...
+                    try:
+                        if path.is_absolute() and path.exists() and path.is_file():
+                            # Path is absolute, no settings modification or filecopy needed
+                            pass
+                    except OSError:
+                        pass
+                # Attempt to construct an absolute file path
+                sourceFullPath = os.path.join(sourceDirPath, f)
+                if WILDCARD == True:
```
john-science, line 845: I think `pylint` prefers:  `if WILDCARD:`

john-science (approved):
A couple of little `pylint` comments, but otherwise it looks a lot more clear than it used to!

PR #818, opened 2022-08-05
add component density Tests

onufer:
## Description

 - revert renames of `applyHotHeightDensityReduction`, and `applyMaterialMassFracsToNumberDensities` from [recent PR](https://github.com/terrapower/armi/pull/804)
   - Justification: the old names are correct enough and we prefer a stable API
 - add `getMassDensity()` function to components
 - adding component density/mass tests for expansion
 - add test demonstrating that construction of components programmatically isn't the same as through blueprints

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/tests/test_blockBlueprints.py
+++ b/armi/reactor/blueprints/tests/test_blockBlueprints.py
@@ -18,6 +18,7 @@ import io
 from armi.reactor import blueprints
 from armi import settings
 from armi.reactor.flags import Flags
+from armi.reactor.tests import test_blocks
 
 FULL_BP = """
 blocks:
@@ -29,15 +30,15 @@ blocks:
             Tinput: 25.0
             Thot: 600.0
             id: 0.0
-            od: 0.86602
+            od: 0.7
             latticeIDs: [1]
-        clad:
+        clad: # same args as test_blocks (except mult)
             shape: Circle
             material: HT9
             Tinput: 25.0
-            Thot: 470.0
-            id: 1.0
-            od: 1.09
+            Thot: 450.0
+            id: .77
+            od: .80
             latticeIDs: [1,2]
         coolant:
             shape: DerivedShape
@@ -69,15 +70,15 @@ blocks:
             Tinput: 25.0
             Thot: 600.0
             id: 0.0
-            od: 0.86602
+            od: 0.67
             latticeIDs: [1]
         clad:
             shape: Circle
             material: HT9
             Tinput: 25.0
-            Thot: 470.0
-            id: 1.0
-            od: 1.09
+            Thot: 450.0
+            id: .77
+            od: .80
             latticeIDs: [1,2]
         coolant:
             shape: DerivedShape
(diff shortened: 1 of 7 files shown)


armi/reactor/blueprints/tests/test_blockBlueprints.py
```diff
@@ -317,6 +318,40 @@ def test_explicitFlags(self):
...
+            delta=biggerDelta,
+        )
+        # This should be equal, but block construction calls applyHotHeightDensityReduction
+        # while programmatic construction allows components to exist in a state where
+        # their density is not consistent with material density.
+        self.assertNotAlmostEqual(
+            clad.getMassDensity(),
+            programaticClad.getMassDensity(),
+            delta=biggerDelta,
+        )
```
onufer, line 346: We probably want to fix this. it is documented in  https://github.com/terrapower/armi/issues/819

armi/reactor/tests/test_components.py
```diff
@@ -497,47 +497,158 @@ def test_changeNumberDensities(self):
...
+        #   # 1 2D Expansion
+        ###########
         # expansion only happens in 2D so only area is necessary
-        # since component expansion is only in 2D
-        tHotC = 20
-        circle1 = Circle("circle", "HT9", 20, tHotC, 1.0)
+
+        # when comparing to 3D density, the comparison is not quite correct.
+        # We need a bigger delta, this will be investigated/fixed in another PR
+        biggerDelta = 0.001  # g/cc
```
onufer, line 509: The fact that we need this is indicative of a bug. Thankfully the differences are very minor.
https://github.com/terrapower/armi/issues/820

armi/reactor/blueprints/tests/test_blockBlueprints.py
```diff
@@ -317,6 +318,40 @@ def test_explicitFlags(self):
...
+        programaticClad = programmaticBlock.getComponent(Flags.CLAD)
+        self.assertAlmostEqual(
+            clad.getMassDensity(),
+            clad.material.density3(Tc=clad.temperatureInC),
+            delta=biggerDelta,
+        )
+        # This should be equal, but block construction calls applyHotHeightDensityReduction
+        # while programmatic construction allows components to exist in a state where
+        # their density is not consistent with material density.
+        self.assertNotAlmostEqual(
```
onufer, line 342: We probably want to fix this. it is documented in https://github.com/terrapower/armi/issues/819

john-science:
I love it when people add tests!  Reviewing now.

armi/reactor/blueprints/tests/test_blockBlueprints.py
john-science: I have been waffling back-and-forth on this.

Adding a test to the repo that passes, but shouldn't seems strange.

But sure, it serves a purpose. I'm on board for that.

Can we add a line: `# TODO: This test passes, but shouldn't` or something, just to ease my nerves?

john-science:
I would be curious to know how this problem relates to this docstring:

https://github.com/terrapower/armi/blob/bc556fc5ce3f0445a4b1e841f00b8bdc3888f5a3/armi/materials/material.py#L341-L348

This has been in our codebase for at least 3 years.

armi/reactor/tests/test_components.py
john-science: To my understanding, if we expand in 2D, we use `Material.density()` and if we expand in 3D, we use `Material.density3()` which will then be the same as 2D density.

So, it appears to me that for the 2D example here, we could be seeing that.  It's the 3D expansion below that gives me pause.

onufer:
@john-science, that doc string indicates that until you fix it it will be off by axial expansion factor. but its not exactly off axial expansion factor, and that is the problem.

john-science (approved):
I love it when people add more tests.

These two kind of "prove a negative", which is unusual. But they are still really helpful to have around.

PR #823, opened 2022-08-05
Update `b.p.heightBOL` in `Core::processLoading` for thermal expansion at BOL 

albeanth:
## Description
If `inputHeightsConsideredHot: False`, assemblies are thermally expanded from `Tinput` to `Thot`. This feature branch modifies `Core::processLoading` to update the BOL block height, `b.p.heightBOL`, to be the thermally expanded height rather than the cold height (as inputted in the blueprints).

---

## Checklist


armi/reactor/reactors.py
```diff
@@ -2280,17 +2286,21 @@ def processLoading(self, cs, dbLoad: bool = False):
...
+                f"=========== Axially expanding {aExpStr} assemblies (except control) from Tinput to Thot ==========="
+            )
+            axialExpChngr = AxialExpansionChanger(cs["detailedAxialExpansion"])
+            for a in assemsToAxiallyExpand:
+                if not a.hasFlags(Flags.CONTROL):
+                    axialExpChngr.setAssembly(a)
+                    axialExpChngr.expansionData.computeThermalExpansionFactors()
+                    axialExpChngr.axiallyExpandAssembly(thermal=True)
+            axialExpChngr.manageCoreMesh(self.parent)
+            self.updateBlockBOLHeights()
```
onufer, line 2303: I think this should only be called on the core when you are not loading from DB. Loads from DB should just use what is stored in DB.

armi/reactor/reactors.py
```diff
@@ -2316,3 +2326,12 @@ def processLoading(self, cs, dbLoad: bool = False):
         self.p.maxAssemNum = self.getMaxParam("assemNum")
 
         getPluginManagerOrFail().hook.onProcessCoreLoading(core=self, cs=cs)
+
+    def updateBlockBOLHeights(self):
+        """post thermal expansion, update block BOL heights"""
+        self.p.referenceBlockAxialMesh = self.findAllAxialMeshPoints(applySubMesh=False)
+        self.p.axialMesh = self.findAllAxialMeshPoints()
+        for a in self.getAssemblies():
```
albeanth, line 2334: this should be `for a in assemsToAxiallyExpand`

armi/reactor/reactors.py
albeanth: Do we not want to expand the blueprints assemblies when loading from DB?

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/823/commits/5ad8f492ef706848ec14c6e7d8c3df676113bc3c

armi/reactor/reactors.py
onufer: you could move the update inside the for loop and it would work

armi/reactor/reactors.py
```diff
@@ -2316,3 +2326,12 @@ def processLoading(self, cs, dbLoad: bool = False):
         self.p.maxAssemNum = self.getMaxParam("assemNum")
 
         getPluginManagerOrFail().hook.onProcessCoreLoading(core=self, cs=cs)
+
+    def updateBlockBOLHeights(self):
+        """post thermal expansion, update block BOL heights"""
+        self.p.referenceBlockAxialMesh = self.findAllAxialMeshPoints(applySubMesh=False)
+        self.p.axialMesh = self.findAllAxialMeshPoints()
```
onufer, line 2340: why does the mesh setting code need to run again?

armi/reactor/reactors.py
```diff
@@ -2280,17 +2286,21 @@ def processLoading(self, cs, dbLoad: bool = False):
...
+            runLog.header(
+                f"=========== Axially expanding {aExpStr} assemblies (except control) from Tinput to Thot ==========="
+            )
+            axialExpChngr = AxialExpansionChanger(cs["detailedAxialExpansion"])
+            for a in assemsToAxiallyExpand:
+                if not a.hasFlags(Flags.CONTROL):
+                    axialExpChngr.setAssembly(a)
+                    axialExpChngr.expansionData.computeThermalExpansionFactors()
+                    axialExpChngr.axiallyExpandAssembly(thermal=True)
+            axialExpChngr.manageCoreMesh(self.parent)
```
onufer, line 2302: I dont think we want to call manager core mesh when its only doing blueprints, right? They should just stay as they were in the database?

What do you think about moving this code back inside the else statement of IsDB and doing a separate expansions of the blueprints after the if/else?

I also think the blueprints need setBlockMeshCalled on them, and this will never be called since manageCoreMesh only calls it on 
a in r.core.getAssemblies():

We dont want the blueprints to use the core mesh anyway though, since we dont want blueprints to have the burnup mesh the core has, just the thermal mesh.

albeanth: Addressed. https://github.com/terrapower/armi/pull/823/commits/15499a11f58af5a0803e4645e818c488da9b0f4b

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/823/commits/15499a11f58af5a0803e4645e818c488da9b0f4b

armi/reactor/reactors.py
albeanth: It is handled pre-expansion, yes. But I think it should be updated. `findAllAxialMeshPoints` ends up in `findAllMeshPoints` and this recalculates mesh positions based on `b.spatialLocator.getGlobalCellBase()` and `b.spatialLocator.getGlobalCellTop()` which change after expansion.

armi/reactor/reactors.py
```diff
@@ -2269,6 +2269,19 @@ def processLoading(self, cs, dbLoad: bool = False):
...
+                self._axiallyExpandAssems(
+                    self.parent.blueprints.assemblies.values(),
+                    dbLoad,
+                    "blueprints",
+                    cs["detailedAxialExpansion"],
+                )
+                if not cs["detailedAxialExpansion"]:
+                    refMesh = (
+                        self.refAssem.getAxialMesh()
+                    )  # calculate once for efficiency
```
onufer, line 2282: with the way black formats, its sometimes nice to  move the comment to the line above. no action NEEDED though.

armi/reactor/reactors.py
```diff
@@ -2269,6 +2269,19 @@ def processLoading(self, cs, dbLoad: bool = False):
...
+            if not cs["inputHeightsConsideredHot"]:
+                self._axiallyExpandAssems(
+                    self.parent.blueprints.assemblies.values(),
+                    dbLoad,
+                    "blueprints",
+                    cs["detailedAxialExpansion"],
+                )
+                if not cs["detailedAxialExpansion"]:
+                    refMesh = (
+                        self.refAssem.getAxialMesh()
```
onufer, line 2281: it mgiht be good to use finestAssemblyMesh

armi/reactor/reactors.py
```diff
@@ -2316,3 +2324,41 @@ def processLoading(self, cs, dbLoad: bool = False):
...
+            a list of :py:class:`Assembly <armi.reactor.assemblies.Assembly>` objects
+            that have been axially expanded
+        dbLoad: bool
+            used to determine if Core::processLoading is originating from database load or not
+        """
+        if not dbLoad:
+            self.p.referenceBlockAxialMesh = self.findAllAxialMeshPoints(
+                applySubMesh=False
+            )
+            self.p.axialMesh = self.findAllAxialMeshPoints()
```
onufer, line 2360: get rid of dbLoad argument and move to the else of the if Dbload statment

armi/reactor/reactors.py
onufer: great point

armi/reactor/reactors.py
```diff
@@ -2316,3 +2324,41 @@ def processLoading(self, cs, dbLoad: bool = False):
...
 
         getPluginManagerOrFail().hook.onProcessCoreLoading(core=self, cs=cs)
+
+    def _axiallyExpandAssems(
+        self, assems: list, dbLoad: bool, expTag: str, detAxExp: bool
+    ):
+        """expand assembliues"""
+        axialExpChngr = AxialExpansionChanger(detAxExp)
+        runLog.header(
+            f"=========== Axially expanding {expTag} assemblies (except control) from Tinput to Thot ==========="
```
onufer, line 2334: optional: consider moving this header outside the function call. i know its duplicated code, but sending a string in to be printed right is not great. it will also help the reader to understand whats happing as they read code.

armi/reactor/reactors.py
onufer: actually, we need to use finestAssemblyMesh. if we don't it will over expand the blueprints to the irradiate fuel core height.  We shoudl probably rename
finestAssemblyMesh
to 
finestMeshAssembly

armi/reactor/reactors.py
```diff
@@ -2316,3 +2324,41 @@ def processLoading(self, cs, dbLoad: bool = False):
...
+        """
+        if not dbLoad:
+            self.p.referenceBlockAxialMesh = self.findAllAxialMeshPoints(
+                applySubMesh=False
+            )
+            self.p.axialMesh = self.findAllAxialMeshPoints()
+        for a in assems:
+            if not a.hasFlags(Flags.CONTROL):
+                for b in a:
+                    b.p.heightBOL = b.getHeight()
```
onufer, line 2364: I dont think this is going to work because you haven't called setBlock mesh on the blueprints yet, so they are not yet the final heights

armi/reactor/reactors.py
albeanth: > we need to use finestAssemblyMesh

I'm not really sure it matters. When an assembly gets created from the blueprint assemblies, it gets created in `reactors.py::Core::createAssemblyOfType`. In this method, if `detailedAxialExpansion: False`, `setBlockMesh` is run. So whichever mesh `setBlockMesh` uses here (in Core::processLoading) just gets overwritten in `createAssemblyOfType` (which happens to be `core.p.referenceBlockAxialMesh()`.

armi/reactor/reactors.py
```diff
@@ -2316,3 +2324,41 @@ def processLoading(self, cs, dbLoad: bool = False):
         self.p.maxAssemNum = self.getMaxParam("assemNum")
 
         getPluginManagerOrFail().hook.onProcessCoreLoading(core=self, cs=cs)
+
+    def _axiallyExpandAssems(
+        self, assems: list, dbLoad: bool, expTag: str, detAxExp: bool
+    ):
+        """expand assembliues"""
```
john-science, line 2331: spelling error: "assembliues"

armi/reactor/reactors.py
john-science: I don't know, I'm not sure it will make a big difference to me, to save that one method argument.  #shrug

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/823/commits/de07353fcb87ce30d7af984de9d00e4199f1a408

armi/reactor/reactors.py
albeanth: Change made
https://github.com/terrapower/armi/pull/823/commits/de07353fcb87ce30d7af984de9d00e4199f1a408

armi/reactor/reactors.py
albeanth: Good catch! 
https://github.com/terrapower/armi/pull/823/commits/de07353fcb87ce30d7af984de9d00e4199f1a408

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/823/commits/de07353fcb87ce30d7af984de9d00e4199f1a408

armi/reactor/reactors.py
albeanth: Good catch. 
https://github.com/terrapower/armi/pull/823/commits/d0ebfe51170939f45c11f3024492cabaf2eebab0

john-science:
Forgive an ignorant question; I'm sure more people have talked about this change.

But isn't it confusing that we are changing the value of `heighBOL`?  The `BOL` part of that name sounds like this value should be set at the beginning, and then never change.

At least, that's what the name _sounds_ like to me.

albeanth:
@john-science that's a fair question! We changing it just once to ensure consistency between `inputHeightsConsideredHot: True/False`. 

When True:
- The block heights in the blueprints are supposed to be at hot temps. So when the core is built, `b.p.heightBOL` is stored at hot temps. This occurs during in the constructor of `Block`. 

When False:
- The block heights in the blueprints are supposed to be at cold temps. So when the core is built, `b.p.heightBOL` is initially stored using cold (i.e. room) temps (in the `Block` constructor). But in this case, during `Core::processLoading` the assemblies are thermally expanded to their hot temperatures. So to be consistent with the previous case, we need to update the `b.p.heightBOL` parameter in process loading after the assemblies have been expanded. This PR addresses this.

john-science (approved):

armi/reactor/reactors.py
```diff
@@ -2284,18 +2293,15 @@ def processLoading(self, cs, dbLoad: bool = False):
                     if a.hasFlags(nonUniformAssems, exact=True):
```
onufer, line 2293: It wont let me comment on  line 2280/2289, but you told me those need to be called after the expansion in a past comment and i think you are right

armi/reactor/reactors.py
```diff
@@ -2321,3 +2327,30 @@ def processLoading(self, cs, dbLoad: bool = False):
...
+        if not dbLoad:
+            axialExpChngr.manageCoreMesh(self.parent)
+            self.p.referenceBlockAxialMesh = self.findAllAxialMeshPoints(
+                applySubMesh=False
+            )
+            self.p.axialMesh = self.findAllAxialMeshPoints()
+        else:
+            if not detAxExp:
+                # calculate refMesh once for efficiency
+                refMesh = self.refAssem.getAxialMesh()
```
onufer, line 2349: This needs to be the finestMeshAssembly. We don't want to expand blueprints to the post burnup heights. Blueprints should be loaded the same with the same mesh regardless of the time node.

armi/reactor/reactors.py
```diff
@@ -2321,3 +2327,30 @@ def processLoading(self, cs, dbLoad: bool = False):
...
+                axialExpChngr.expansionData.computeThermalExpansionFactors()
+                axialExpChngr.axiallyExpandAssembly(thermal=True)
+        # resolve axially disjoint mesh (if needed)
+        if not dbLoad:
+            axialExpChngr.manageCoreMesh(self.parent)
+            self.p.referenceBlockAxialMesh = self.findAllAxialMeshPoints(
+                applySubMesh=False
+            )
+            self.p.axialMesh = self.findAllAxialMeshPoints()
+        else:
```
onufer, line 2346: can be reduced to `elif`

armi/reactor/reactors.py
onufer: Oh, i see now that its also set below. Is the initial set here needed? if not maybe only set once and set after 
_updateBlockBOLHeights since then  we are doing one less random thing in the _updateBlockBOLHeights

armi/reactor/reactors.py
```diff
@@ -2321,3 +2327,30 @@ def processLoading(self, cs, dbLoad: bool = False):
...
+            if not detAxExp:
```
onufer, line 2347: could use self._detailedAxialExpansion if you liked to eliminate arg

onufer (changes requested):

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/823/commits/cc59a8f0e474953d2c57a6f588e2208027c3d5f4

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/823/commits/cc59a8f0e474953d2c57a6f588e2208027c3d5f4

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/823/commits/76cfc4f4095a94773a16c5f0b808838db7f36d4c

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/823/commits/76cfc4f4095a94773a16c5f0b808838db7f36d4c

onufer (approved):

armi/reactor/reactors.py
```diff
@@ -2284,18 +2293,11 @@ def processLoading(self, cs, dbLoad: bool = False):
                     if a.hasFlags(nonUniformAssems, exact=True):
                         continue
                     a.makeAxialSnapList(self.refAssem)
-
             if not cs["inputHeightsConsideredHot"]:
                 runLog.header(
-                    "=========== Axially expanding all (except control) assemblies from Tinput to Thot ==========="
+                    "=========== Axially expanding all assemblies (except control) from Tinput to Thot ==========="
```
jakehader, line 2298: Maybe change this to "Thermal Axial Expansion of Core Assemblies" as a header and then when you get to control just create a runLog info that CONTROL is being skipped. I'm curious if there should be a setting to say which assemblies to skip and we can just configure our application to neglect control rods.

keckler: ^^^ Yes, this!

albeanth: One thing with changing from "all" to "Core". In this case, we are thermally expanding _all_ assemblies - both Core and blueprints assemblies. So it may be confusing if we say just Core when we are doing more than just the Core assems. 

@jakehader either way, your idea feels worthy of a separate PR to me. Maybe create an issue for it and we can reference it in a new PR?

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/823/commits/ded838291fa5683863138fec22c72a69339d494d

armi/reactor/reactors.py
jakehader: Sure, the configuration can be another PR. Just wanted to give that feedback

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -257,9 +259,10 @@ def manageCoreMesh(self, r):
 
         oldMesh = r.core.p.axialMesh
         r.core.updateAxialMesh()
-        runLog.extra("Updated r.core.p.axialMesh (old, new)")
-        for old, new in zip(oldMesh, r.core.p.axialMesh):
-            runLog.extra(f"{old:.6e}\t{new:.6e}")
+        if oldMesh:
+            runLog.extra("Updated r.core.p.axialMesh (old, new)")
```
jakehader, line 263: Let's format this better. The output is pretty ugly. You could use tabulate.

armi/reactor/reactors.py
albeanth: Issue made. https://github.com/terrapower/armi/issues/832

armi/reactor/converters/axialExpansionChanger.py
albeanth: This is a good idea. Just going to (temporarily) punt to an issue to keep the content of this PR as minimal as possible.

albeanth: Issue here https://github.com/terrapower/armi/issues/833

jakehader (approved):

PR #827, opened 2022-08-09
Support non-uniform assemblies with `detailedAxialExpansion` disabled

onufer:
## Description

---

## Checklist

### DIFF
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -713,8 +713,13 @@ class Assembly(composites.Composite):
         belowFuelColumn = True
 
         if self[-1].p.topIndex == 0:
-            # this appears to not have been initialized, so initialize it
-            self.makeAxialSnapList(refMesh=blockMesh)
+            runLog.warning(
+                "Reference uniform mesh not being applied to {}. It was likely "
+                "excluded through the setting `nonUniformAssemFlags`.".format(
+                    self.p.type
+                )
+            )
+            return
 
         for b in self:
             if b.isFuel():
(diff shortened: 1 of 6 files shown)


armi/reactor/assemblies.py
```diff
@@ -713,8 +713,13 @@ def setBlockMesh(self, blockMesh, conserveMassFlag=False, adjustList=None):
         belowFuelColumn = True
 
         if self[-1].p.topIndex == 0:
-            # this appears to not have been initialized, so initialize it
-            self.makeAxialSnapList(refMesh=blockMesh)
+            runLog.warning(
```
onufer, line 716: we initialize this in all assemblies, including blueprints assemblies in processLoading now, so every assembly should have this except for the ones we want to ignore expansion on.

jakehader (approved):
This is great! This will allow folks to implement certain assembly types (like control assemblies) in the blueprints without requiring them to strictly conform to the same mesh resolution as the reference assembly. This is useful in cases where assemblies can translate vertically throughout the core (like control rods). The main thing to look out for here is that physics solvers that require a uniform mesh must ensure that there is uniformity globally before running the analysis. This is implemented through specific physics plugins within said application.

john-science:
Okay, this new `CaseSetting` seems fine, and I like the unit test.

But... who is planning on using this?  My thought is that since no one has needed this the last ~8 years, I'm surprised this went into `armi` and not a `Plugin` or `App`.  I assume that this was added because you want to use it yourself right now, right?

I'm kind of waffling; do we think this is generally useful enough to be in `ARMI`?

onufer:
@john-science, we will likely be using within TP. please call me for a non-public chat i can expain.

PR #835, opened 2022-08-12
Remove point class

Nebbychadnezzar:
## Description

This PR proposes to remove the `Point` class from the framework. The idea behind the class makes sense, but its implementation should be explored some more before releasing it into the wild. Specifically, subclassing it from `Block` makes it a very heavy object, so perhaps its weight can be paired down with some additional work.

An inspection of internal code suggests that there is only one particular plugin that makes use of the class. To avoid breaking the functionality of that plugin, I propose moving this class (and its unit test) to that plugin so that it can continue to function.

---

## Checklist


armi/physics/neutronics/parameters.py
```diff
@@ -729,7 +729,7 @@ def linPowByPinGamma(self, value):
             units="n/cm^2/s",
             description="Neutron flux above 100keV at hexagon block corners",
             location=ParamLocation.CORNERS,
-            saveToDB=False,
+            saveToDB=True,
```
john-science, line 732: Do these changes have to do with the `Point` change?

Nebbychadnezzar: It does not. I forgot to add this to the description, sorry.

We had talked last week about this PR having this change included as well, since it was a small update. I am happy to separate it out into another PR though, if you'd prefer that.

jakehader:
I'm wondering if rather than removing Point we can just turn it into a normal composite with a different set of parameter definitions. Maybe we can pull it out as @Nebbychadnezzar is suggesting and then come up with something that is lighter weight and then go from there. I think we should figure out how a point would fit within a spatial grid system and if it can have children or ancestors. If not, then we should figure out how best to put the data into the database. 

Either way, not many things use point and they could use a big update!

Nebbychadnezzar:
@jakehader, The idea behind it makes sense and should be further explored, but the current implementation doesn't look sustainable to me, as its use is causing memory trouble in an internal plugin for me. So, I thought rather than advertising it as something that could/should be used, we could rework the implementation and then release it again for use.

jakehader:
> @jakehader, The idea behind it makes sense and should be further explored, but the current implementation doesn't look sustainable to me, as its use is causing memory trouble in an internal plugin for me. So, I thought rather than advertising it as something that could/should be used, we could rework the implementation and then release it again for use.

@Nebbychadnezzar agreed!

jakehader:
@drewj-usnctech, are you using this functionality right now?

armi/physics/neutronics/parameters.py
Nebbychadnezzar: Alternatively, these four parameters could also be moved into the internal plugin, since they are also "associated" with the `Point` class.

john-science: I'm just asking.  You don't have to change anything. Thanks for the explanation.

john-science (approved):

armi/physics/neutronics/parameters.py
```diff
@@ -739,7 +739,7 @@ def linPowByPinGamma(self, value):
             units=None,
             description="Fraction of flux above 100keV at points within the block",
             location=ParamLocation.CHILDREN,
-            saveToDB=False,
+            saveToDB=True,
```
jakehader, line 742: These are still going to be used even If you're removing the Points class?

drewj-usnctech:
@jakehader thanks for reaching out and checking in. No we are not using the `Point` class

armi/physics/neutronics/parameters.py
jakehader: Okay cool just asking!

jakehader (approved):

jakehader:
@john-science it looks like this is probably good to go since no internal or known external folks are using this at the moment. I agree with @Nebbychadnezzar assessment that these are too heavy since they are inheriting from blocks. Honestly, a point in space having data doesn't really fit into the standard ArmiObject because it doesn't have mass, volume, number densities, children, etc. It might be best suited as a different type of object that has a spatial location relative to some composite with some basic parameters, but not the full set that are defined by the existing code. Will still need a way to represent these in the database though. I'd love if a collection of points could represent field data, like temperatures, velocities, forces (?), etc that are generally useful for non-reactor physics folks

opotowsky:
From reviewing the file in question, this PR should not have reduced code coverage. I do not believe merging main into this feature branch is necessary since the change is simple and I've examined the uncovered lines of code (and they have nothing to do with this code deletion PR). Merging -- great work!

Issue #819, opened 2022-08-05
Component density differs if you create component through blueprints vs programatically


onufer:
Component density differs if you create component through blueprints vs programmatically .

This is due to having to remember to call `applyHotHeightDensityReduction()` if you do so programmatically. This is not great because if you forget to call it, your component wont be equal to the 3D density. We should strive to have components density always equal to 3D density. We used to do this on component construction, but #657 changed this.

 I don't think that it needed to, the math just might be a little different for axial expansion.

#818 writes a unit tests demonstrating this behavior

onufer:
Thoughts @john-science, @albeanth ?

john-science:
@albeanth  Okay, my read of the situation is this is a bug.  Unless you disagree, I am assigning the bug to you, because it has the phrase `applyHot` in it.  Holler at me if I'm wrong.

albeanth:
Looking into it.

albeanth:
> Component density differs if you create component through blueprints vs programmatically .

I believe the issue you've identified here boils down to an inconsistency between the default behavior of the Component constructor and case setting, `inputHeightsConsideredHot` (introduced with #657). Testing on my end shows that if the case settings have `inputHeightsConsideredHot: False`, then the number densities from a Component built via blueprints and programmatically are the same.

Some more details:

As of now (after #804), the constructor of Component expects that the "height" of the component is cold. So when a Component is built programmatically, the number densities are computed with the adjustment for hot radial dimensions only. However, the default behavior of the blueprints expects the "height" of the Component to be hot (via inputHeightConsideredHot: True) , and therefore the number densities are adjusted for both hot radial and hot axial conditions via a call to `applyHotHeightDensityReduction`. This inconsistency results in the discrepancy you've identified.

So an option for a fix: Have the default behavior of the Component constructor compute number densities that assume hot radial and hot axial dimensions to match the default behavior of Components built through the blueprints. 

Some impacts of this change:
  - If the case setting `inputHeightsConsideredHot: False`, then we will need to recompute/adjust Component number densities to account for a cold axial height when building from blueprints. 
  - If a user wants to build a Component programatically that has a cold "height", they will need to manually call the method to adjust the number densities.
      - note: this is the same scenario you are experiencing now, just reversed.

john-science:
I would be curious to know how this problem relates to this docstring:

[...]

PR #841, opened 2022-08-16
Component initial denisty

onufer:
## Description

fixed an Issue (#819) where components initialized through the constructor would have different density than those created through blueprints. 

---

## Checklist


armi/reactor/reactors.py
```diff
@@ -215,7 +215,6 @@ def __init__(self, name):
         self._circularRingPitch = 1.0
         self._automaticVariableMesh = False
         self._minMeshSizeRatio = 0.15
-        self._inputHeightsConsideredHot = True
```
onufer, line 218: unused

onufer:
this will resolve #819

albeanth (commented):
Here's an initial review for `component.py` and `test_components.py`. 
- There is a great deal of value in these new tests, they are great additions!
- In an effort to improve clarity, I would suggest splitting these tests off into their own test class and into individual tests. 
- I think doing so will not only help the review, but also make the tests later on (e.g. when switching between temps) clearer. Anything that is needed for other assertions (e.g., mass1) can be stored in the setUp method. 

```python 
class TestAxialExpansion(unittest.TestCase):
    """demonstrate conservation during radial and axial expansion"""
    def setUp(self):
        hotHeight = 1.0
        tWarm = 50
        coldOuterDiameter = 1.0
        circle1 = Circle("circle", "HT9", 20, tWarm, coldOuterDiameter)
        tHot = 500
        circle2 = Circle("circle", "HT9", 20, tHot, coldOuterDiameter)

    def test_massDensityProportional(self):
        """mass density is proportional to Fe number density and derived from all the number densities and atomic masses"""
        self.assertAlmostEqual(
            circle1.p.numberDensities["FE"] / circle2.p.numberDensities["FE"],
            circle1.getMassDensity() / circle2.getMassDensity(),
        )

    def test_massColdGreaterThanMassHot(self):
        """docstring"""
        mass1 = circle1.getMassDensity() * circle1.getArea() * hotHeight
        mass2 = circle2.getMassDensity() * circle2.getArea() * hotHeight
        self.assertGreater(mass1, mass2)
    
    etc...
```

opotowsky (approved):
I'd love a PR description (I'll edit to include the issue), but I think this work looks good, and solves a very important issue!

Note that the slight reduction in code coverage is because this branch is 2 commits behind main and is not indicative of lack of testing coverage in this PR.

opotowsky (changes requested):
@onufer Can you please update this bug fix in the release notes?

armi/reactor/components/component.py
```diff
@@ -348,10 +348,12 @@ def applyMaterialMassFracsToNumberDensities(self):
         self.p.numberDensities = densityTools.getNDensFromMasses(
             density, self.material.p.massFrac
         )
+        self.applyHotHeightDensityReduction()
```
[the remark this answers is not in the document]
onufer: good point, updated.

onufer:
The test is meant to be a tutorial for how to expand something. it will be a lot of duplicated code to show each step individually and be much less clear to read. The idea is that it describes a tutorial how to do this type of thing in a script and what you can expect to be conserved.

As a result of your comment, i did split it into its own test class, and 1 test for height taken as hot, and 1 for height taken as cold. I hope that helps with addressing your concerns. Let me know if its more clear. Its quite a bit fewer lines than the original testing.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -175,9 +192,10 @@ def axiallyExpandAssembly(self, thermal: bool = False):
             # if ib == 0, leave block bottom = 0.0
             if ib > 0:
                 b.p.zbottom = self.linked.linkedBlocks[b][0].p.ztop
-            # if not in the dummy block, get expansion factor, do alignment, and modify block
-            if ib < (numOfBlocks - 1):
+            isDummyBlock = ib == (numOfBlocks - 1)
```
jakehader, line 195: Can this be based on flags?

opotowsky (approved):

armi/reactor/converters/axialExpansionChanger.py
onufer: scope creep.... they are treated as flags in other places, but this is a potential separate issue.

there are warnings if they are not dummy
image

albeanth: @jakehader - I don't think it should be, actually. If we were to change it to be something like 

```python
if not b.hasFlags(Flags.DUMMY):
    # do the expansion/contraction on the block
``` 

then we'd be expanding/contracting the top-most block of an assembly if it is not have a dummy  block. This could be problematic for downstream plugins as some require the assemblies to all be the same height. 

By leaving it as is, we ensure that the top-most block, dummy or not, is chopped to preserve assembly height.

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -765,6 +766,14 @@ def test_coldAssemblyHeight(self):
...
                 else:
                     checkColdBlockHeight(bStd, bExp, self.assertNotEqual, "different")
+                if bStd.hasFlags(Flags.FUEL):
+                    # fuel mass should grow because heights are considered cold heights
+                    # and a cold 1 cm column has more mass than a hot 1 cm column
+                    if not isinstance(
+                        bStd.getComponent(Flags.FUEL).material, custom.Custom
+                    ):
+                        # custom materials don't expand
+                        self.assertGreater(bExp.getMass("U235"), bStd.getMass("U235"))
```
albeanth, line 776: Just a note to my future self. 

Since the heights in the blueprints are the same, the mass of `inputHeightsConsideredHot: False` (bExp) will be higher than `inputHeightsConsideredHot: True` (bStd). The former is cooler and therefore has a higher density than its hot counterpart - since the two cases have the same volume, the former has a higher mass. 

When bExp is expanded at BOL, mass is conserved. So post expansion, the mass is still higher than bStd.

albeanth (approved):
Coverage failure looks to me to be coverage spray. 

I think this is ready to go.

armi/reactor/converters/axialExpansionChanger.py
jakehader: Got it! Thank you for the clarification. I'm good then, but note I haven't done a complete review of the work or tests so I am leaving that to @onufer or other reviewers unless I'm needed.

PR #842, opened 2022-08-17
allow no blueprints to load

onufer:
## Description

---

## Checklist

### DIFF
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -2257,30 +2257,32 @@ class Core(composites.Composite):
             # this normally happens during armi/reactor/blueprints/__init__.py::constructAssem
             # but for DB load, this is not called so it must be here.
             # pylint: disable=protected-access
-            self.parent.blueprints._prepConstruction(cs)
-            if not cs["detailedAxialExpansion"]:
-                # Apply mesh snapping for self.parent.blueprints.assemblies
-                # This is stored as a param for assemblies in-core, so only blueprints assemblies are
-                # considered here. To guarantee mesh snapping will function, makeAxialSnapList
-                # should be in reference to the assembly with the finest mesh as defined in the blueprints.
-                finestMeshAssembly = sorted(
-                    self.parent.blueprints.assemblies.values(),
-                    key=lambda a: len(a),
-                    reverse=True,
-                )[0]
-                for a in self.parent.blueprints.assemblies.values():
-                    if a.hasFlags(nonUniformAssems, exact=True):
-                        continue
-                    a.makeAxialSnapList(refAssem=finestMeshAssembly)
-            if not cs["inputHeightsConsideredHot"]:
-                runLog.header(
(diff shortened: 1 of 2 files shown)


armi/reactor/reactors.py
```diff
@@ -2257,30 +2257,32 @@ def processLoading(self, cs, dbLoad: bool = False):
...
-                runLog.header(
-                    "=========== Axially expanding blueprints assemblies (except control) from Tinput to Thot ==========="
-                )
-                self._applyThermalExpansion(
-                    self.parent.blueprints.assemblies.values(),
-                    dbLoad,
-                    finestMeshAssembly,
-                )
+            if not self.parent.blueprints:
+                # we don't require reactors to have blueprints
```
onufer, line 2261: we dont require BP, so we should allow load without them
https://github.com/terrapower/armi/blob/cacbd3d30d58e713348328d3a745cbe4adfdcc65/armi/bookkeeping/db/database3.py#L832-L836

opotowsky (changes requested):
Simple change, and looks good to me! But we have a testing problem (that isn't your fault)

The reason for the code coverage reduction is that the code here wasn't originally covered by a test, and this PR adds lines to that uncovered `else`: 
``` python
            else:
                runLog.warning(
                    f"Blueprints file not found at {bpPath}"
                    "not writing blueprints to database."
                )
                bpString = ""
```

Would you mind adding a little test that loads `TEST_ROOT`, modifies `loadingfile` to an empty string, and tests that `initDB` or `writeInputsToDB` gives `self.h5db["inputs/blueprints"] = ""`?

armi/reactor/reactors.py
```diff
@@ -2253,34 +2253,37 @@ def processLoading(self, cs, dbLoad: bool = False):
...
-                runLog.header(
-                    "=========== Axially expanding blueprints assemblies (except control) from Tinput to Thot ==========="
-                )
-                self._applyThermalExpansion(
-                    self.parent.blueprints.assemblies.values(),
-                    dbLoad,
-                    finestMeshAssembly,
-                )
+            # we don't require reactors to have blueprints
+            if self.parent.blueprints:
```
jakehader, line 2257: I think we _should_ require this going forward though. Databases need to have blueprints on them for `createAssemblyOfType` to function when loading from a database. Otherwise, we cannot do things like load from a state point, swap a fresh assembly, run a calculation. Is the plan to add this elsewhere?

jakehader: Spoke offline and we agreed that BP in the database will be enforced later.

onufer:
Abandoning because there are other places blueprints are addressed. we really need them.

john-science:
> Abandoning because there are other places blueprints are addressed. we really need them.

Also, blueprints can be addressed (and then hypothetically changed) at almost any time in ARMI or any plugin that imports ARMI.

I have tried to make it harder to do that, but it is a _very_ common idea to write your code around changing the blueprints object. So, fighting blueprint modification is always a lot of work.

Issue #820, opened 2022-08-05
Component density does not exactly agree with 3D density


onufer:
#818 created a test comparing component density to material 3D density, and the differences are larger than expected (but still somewhat small, on the order of 0.001 g/cc).

john-science:
@onufer   Is this different than #819 ?

onufer:
@john-science  yes. See the 2 comments in the unit tests i created. i call them out sepeartly

PR #846, opened 2022-08-19
Conserve component density

mgjarrett:
## Description

The `axialExpansionChanger` was implemented to conserve the initial mass of each component in a block with axial expansion determined by the target component, using the funciton `_conserveComponentMass`. Since different components are made of different materials that thermally expand at different rates, this necessarily means that the densities have to be adjusted artificially for the mass to be conserved over the block for each individual component.

After testing and discussing the methodology, it has been decided that it is preferred to conserve the realistic density of the material rather than conserving mass within a block. Effectively, we are allowing mass to move between blocks.

The `Tinput` values in the test reactor for `detailedAxialExpansion` were updated because the unit test to ensure density conservation requires that the `Tinput` is equal to the material reference temperature for calculating thermal expansion factors.

---

## Checklist

### DIFF
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -249,7 +249,8 @@ class Component(composites.Composite, metaclass=ComponentType):
             runLog.warning(
                 f"Thermal expansion for {self.material} at Tinput = {Tinput} is non-zero "
                 f"({expansionFactor}). The modeled density for this material will be off "
-                f"by a factor of {(1 + expansionFactor) ** 2}."
+                f"by a factor of {(1 + expansionFactor) ** 2}.",
+                single=True,
             )
 
     @property
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -232,10 +232,25 @@ class Component(composites.Composite, metaclass=ComponentType):
         self.temperatureInC = Thot
         self.material = None
         self.setProperties(material)
+        self.tInputWarning(Tinput)
         self.applyMaterialMassFracsToNumberDensities()  # not necessary when duplicating...
         self.setType(name)
         self.p.mergeWith = mergeWith
         self.p.customIsotopicsName = isotopics
+
+    def tInputWarning(self, Tinput):
+        """
+        Check whether thermal expansion factor is 0.0% exactly at T=Tinput
+        """
+        expansionFactor = (
+            self.material.linearExpansionPercent(Tc=self.inputTemperatureInC) / 100.0
+        )
+        if not (abs(expansionFactor) < 1.0e-6):
+            runLog.warning(
+                f"Thermal expansion for {self.material} at Tinput = {Tinput} is non-zero "
+                f"({expansionFactor}). The modeled density for this material will be off "
+                f"by a factor of {(1 + expansionFactor) ** 2}."
+            )
 
     @property
     def temperatureInC(self):
--- a/armi/reactor/converters/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger.py
@@ -289,8 +289,8 @@ class AxialExpansionChanger:
     def _conserveComponentDensity(self, b, oldHeight, oldVolume):
         """Update block height dependent component parameters
 
-        1) update component volume (used to compute block volume)
-        2) update number density
+        1) update component volume for all materials (used to compute block volume)
+        2) update number density for solid materials only (no fluid)
 
         Parameters
         ----------
@@ -300,7 +300,7 @@ class AxialExpansionChanger:
             list containing component volumes pre-expansion
         """
 
-        solidMaterials = _getSolidMaterials(b)
+        solidMaterials = _getSolidComponents(b)
         for ic, c in enumerate(b):
             c.p.volume = oldVolume[ic] * b.p.height / oldHeight
             if c in solidMaterials:
@@ -313,7 +313,15 @@ class AxialExpansionChanger:
                     c.setNumberDensity(key, c.getNumberDensity(key) / growth)
 
 
-def _getSolidMaterials(b):
+def _getSolidComponents(b):
+    """
+    Return list of components in the block that have solid material.
+
+    Notes
+    -----
+    Axial expansion only needs to be applied to solid materials. We should not update
+    number densities on fluid materials to account for changes in block height.
+    """
     return [c for c in b if not isinstance(c.material, material.Fluid)]
 
 
--- a/armi/reactor/converters/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger.py
@@ -300,10 +300,10 @@ class AxialExpansionChanger:
             list containing component volumes pre-expansion
         """
 
-        solidMaterials = _getSolidComponents(b)
+        solidComponents = _getSolidComponents(b)
         for ic, c in enumerate(b):
             c.p.volume = oldVolume[ic] * b.p.height / oldHeight
-            if c in solidMaterials:
+            if c in solidComponents:
                 growFrac = self.expansionData.getExpansionFactor(c)
                 if growFrac >= 0.0:
                     growth = 1.0 + growFrac
--- a/armi/reactor/converters/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger.py
@@ -16,6 +16,7 @@
 from statistics import mean
 from numpy import array
 from armi import runLog
+from armi.materials import material
 from armi.reactor.flags import Flags
 from armi.reactor.components import UnshapedComponent
 from armi.utils import densityTools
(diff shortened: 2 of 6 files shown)


mgjarrett:
@albeanth I took a quick glance at the docs/release notes and didn't see anything that needed to be updated, but you would probably know better than me.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -284,24 +286,43 @@ def manageCoreMesh(self, r):
...
+            list containing component volumes pre-expansion
+        """
 
-    Parameters
-    ----------
-    oldHeight : list of floats
-        list containing block heights pre-expansion
-    oldVolume : list of floats
-        list containing component volumes pre-expansion
+        solidMaterials = _getSolidComponents(b)
```
jakehader, line 303: Should be `solidComponents`.

armi/tests/detailedAxialExpansion/refSmallReactorBase.yaml
```diff
@@ -160,15 +160,15 @@ blocks:
...
+            Tinput: 27.548
             Thot: 470.0
             id: 1.0
             mult: 169.0
             od: 1.09
         wire: &component_plenum_wire
             shape: Helix
             material: HT9
-            Tinput: 25.0
+            Tinput: 27.548
```
jakehader, line 171: Can you explain in the PR description why these are changing? I assume that this is because we have `inputHeightsConsideredHot` set to `True` for the tests? Not clear though.

jakehader (changes requested):

jakehader (changes requested):
See comments.

albeanth:
@mgjarrett Haven't had a chance to review this yet. But am noticing that the coverage dropped pretty drastically. Especially in axialExpansionChanger.py. Can you confirm that you are getting the same coverage drop on your local machine?

armi/tests/detailedAxialExpansion/refSmallReactorBase.yaml
mgjarrett: The reference temperature of HT9 for thermal expansion is 27.548C. The test to check tohat density is conserved only works if Tinput is equal to the reference temperature for thermal expanison.

armi/reactor/converters/axialExpansionChanger.py
mgjarrett: Updated

mgjarrett:
What's confusing is that the coverage rose in commit `ec29050`

The next 3 commits are just changing function and variable names, but no functional changes have been implemented at all. Now it says coverage decreased. One of those commits broke a unit test, which might have confused coveralls.

mgjarrett:
@jakehader I've addressed your comments. I'm trying to understand what happened to the coverage but I'm having trouble understanding what happened from the coveralls report. It does appear that all of the new lines are covered.

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -810,6 +824,19 @@ def test_coldAssemblyExpansion(self):
                     ):
                         # custom materials don't expand
                         self.assertGreater(bExp.getMass("U235"), bStd.getMass("U235"))
+                if not aStd.hasFlags(Flags.CONTROL) and not aStd.hasFlags(Flags.TEST):
+                    if not hasCustomMaterial:
+                        # skip blocks of custom material where liner is merged with clad
+                        for cExp in bExp:
+                            if not isinstance(cExp.material, custom.Custom):
+                                matDens = cExp.material.density3(Tc=cExp.temperatureInC)
+                                compDens = cExp.getMassDensity()
```
albeanth, line 833: `matDens` and `compDens` are unused variables.

albeanth (changes requested):

armi/tests/detailedAxialExpansion/refSmallReactorBase.yaml
albeanth: In line with @jakehader 's comment:

@mgjarrett - for our future selves I think it would be good to leave a comment in this PR on precisely why you/we are making the changes to these temperatures and to armi/materials/zr.py (especially the latter). 

I could easily see us coming back a few months later and being totally stumped as to why these temperatures were all changed.

albeanth (approved):
Approving, but would love to see a comment addressing this. Thanks @mgjarrett ! 
https://github.com/terrapower/armi/pull/846#discussion_r951981623

jakehader:
Okay, after talking with @mgjarrett I think I understand the scope of this change. Basically, we are demonstrating that _if_ the component's input temperature is equal to its material's `refTemp` then thermal expansion will be correctly treated. Otherwise, if these inputs are not equal then the `material.density3()` and `c.getMassDensity()` will not be equal and therefore be off by a factor of the thermal expansion coefficients between the material at the reference temperature and the input temperatures. Please correct my @mgjarrett if this is wrong.

If the above is correct, then I suggest adding a _warning_ to the user that the component input reference does not align with the material's `refTemp`, which can probably be added here: https://github.com/terrapower/armi/blob/main/armi/reactor/components/component.py#L318-L328

I am not sure, but this will probably break downstream unit tests that we will have to fix. @opotowsky, would you mind pulling this in and running all internal plugin unit tests and the integration tests?

mgjarrett:
@jakehader that summary is correct. I've added a warning when the absolute value of the thermal expansion factor at `Tinput` is greater than 1e-6.

I am firing off some downstream tests right now. I expect the impacts to be minimal but we will see.

armi/reactor/components/component.py
```diff
@@ -232,11 +232,26 @@ def __init__(
...
 
+    def tInputWarning(self, Tinput):
+        """
+        Check whether thermal expansion factor is 0.0% exactly at T=Tinput
+        """
+        expansionFactor = (
+            self.material.linearExpansionPercent(Tc=self.inputTemperatureInC) / 100.0
+        )
+        if not (abs(expansionFactor) < 1.0e-6):
+            runLog.warning(
```
jakehader, line 249: Let's make this a single warning so it doesnt show up a million times in the stdout

albeanth:
> @albeanth I took a quick glance at the docs/release notes and didn't see anything that needed to be updated, but you would probably know better than me.

To my knowledge, no need to add anything to the docs. 

I would add one line to the release notes commenting on the density conservation bug fix.

armi/reactor/components/component.py
mgjarrett: Done

armi/tests/detailedAxialExpansion/refSmallReactorBase.yaml
mgjarrett: @albeanth are you looking for a comment directly in the yaml file? Or just include it in the commit message?

albeanth: The commit message would be fine, but I was thinking about just a general comment in this PR. That way if someone in the future does a git blame and comes back to this PR it will be (somewhat) front and center.

I suppose a comment in the yaml would be ok too - maybe at the very top right off the bat? Not sure where it's best to include it. I'll leave it to your discretion.

jakehader (approved):

mgjarrett:
Downstream tests are currently failing; waiting to merge this until we get that resolved

opotowsky:
Downstream tests resolved -- merging!

onufer:
I would expect that after #820 is resolved the odd initial temps in blueprints would not longer be needed

armi/reactor/components/component.py
```diff
@@ -232,11 +232,27 @@ def __init__(
         self.temperatureInC = Thot
         self.material = None
         self.setProperties(material)
+        self.tInputWarning(Tinput)
         self.applyMaterialMassFracsToNumberDensities()  # not necessary when duplicating...
         self.setType(name)
         self.p.mergeWith = mergeWith
         self.p.customIsotopicsName = isotopics
 
+    def tInputWarning(self, Tinput):
```
john-science, line 241: I think we should move this to the unit tests.

Validating the input data during run time makes our code slower. And we can't do ALL of the validation in the world at run time.

For instance, if we add this as a unit test here, we will automatically run this validation against every material in the repo:

https://github.com/terrapower/armi/blob/main/armi/materials/tests/test_materials.py#L28-L34

A benefit of that is we can control where this validation gets done. 

For instance, what if this Component is made of a void / empty material. Do we expect that material to have the `linearExpansionPercent()` method?

I think I will open a ticket to move this to unit tests.

john-science: I'll handle it:  https://github.com/terrapower/armi/issues/861

mgjarrett: 1. This is checking the user input, so it's not possible to perform the check as a unit test.
2. #860 removes the code in question because it fixed the root of the problem.

PR #852, opened 2022-08-23
Set default value to None for listy neutronics parameters

Nebbychadnezzar:
## Description

In PR #849, changes were made to create new neutronics parameters by replacing existing parameters. However, I did not realize that for parameters that use lists (which these do), any non-assignment needs to have a default value of None instead of 0.0; one cannot write to a database if the default value is 0.0.

This commit fixes this issue and explicitly assigns default None values to those parameters.

---

## Checklist


jakehader (approved):
Good catch! Essentially this was failing on DB writing because there were inconsistent types (some assigned with `numpy.array` and others set to a float of 0.0).

PR #854, opened 2022-08-24
Uniform mesh refactor for single assembly number density and parameter data mapping

jakehader:
## Description

The changes in this PR support performing uniform mesh conversions on a single assembly, while also mapping number densities and parameter data from some starting assembly to another. During this mapping, the user can provide the reference, or `source` assembly as well as a new mesh structure. A new assembly is created based on the `source` assembly, the new axial mesh is applied, and then the number densities/parameters provided are mapped.

* Refactoring the `setStateFromOverlaps` to handle both the parameter mapping between assemblies as well as the number density mapping. This is now applied when calling `makeAssemWithUniformMesh`.
---

## Checklist

### DIFF
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -180,6 +180,9 @@ class UniformMeshGeometryConverter(GeometryConverter):
         self._cachedReactorCoreParamData = {}
         self._clearStateOnReactor(self._sourceReactor, cache=True)
         self._mapStateFromReactorToOther(self.convReactor, self._sourceReactor)
+
+        # We want to map the converted reactor core's library to the source reactor
+        # because in some instances this has changed (i.e., when generating cross sections).
         self._sourceReactor.core.lib = self.convReactor.core.lib
         completeEndTime = timer()
         runLog.extra(
@@ -375,7 +378,7 @@ class UniformMeshGeometryConverter(GeometryConverter):
                     raise ValueError(
                         f"The state of {destBlock} on {destinationAssembly} "
                         f"was not cleared prior to calling ``setAssemblyStateFromOverlaps``. "
-                        f"This is an implementation bug in the mesh converter that should "
+                        f"This indicates an implementation bug in the mesh converter that should "
                         f"be reported to the developers. The following parameters should be cleared:\n"
                         f"Parameters: {paramNames}\n"
                         f"Values: {existingDestBlockParamVals}"
@@ -458,36 +461,48 @@ class UniformMeshGeometryConverter(GeometryConverter):
 
                     setter(destBlock, updatedDestVals.values(), updatedDestVals.keys())
 
-                # For parameters that have not been set on the destination block, recover these
-                # back to their originals based on the cached values.
-                for paramName in paramNames:
+                UniformMeshGeometryConverter._applyCachedParamValues(
+                    destBlock, paramNames, cachedParams
+                )
 
-                    # Skip over any parameter names that were not previously cached.
-                    if paramName not in cachedParams[destBlock]:
-                        continue
+    @staticmethod
+    def _applyCachedParamValues(destBlock, paramNames, cachedParams):
+        """
+        Applies the cached parameter values back to the destination block, if there are any.
 
-                    if isinstance(destBlock.p[paramName], numpy.ndarray):
-                        # Using just all() on the list/array is not sufficient because if a zero value exists
-                        # in the data then this would then lead to overwritting the data. This is an edge case see
-                        # in the testing, so this excludes zero values on the check.
-                        if (
-                            not all(
-                                [val for val in destBlock.p[paramName] if val != 0.0]
-                            )
-                            or not destBlock.p[paramName].size > 0
-                        ):
-                            destBlock.p[paramName] = cachedParams[destBlock][paramName]
-                    elif isinstance(destBlock.p[paramName], list):
-                        if (
-                            not all(
-                                [val for val in destBlock.p[paramName] if val != 0.0]
-                            )
-                            or not destBlock.p[paramName]
-                        ):
-                            destBlock.p[paramName] = cachedParams[destBlock][paramName]
-                    else:
-                        if not destBlock.p[paramName]:
-                            destBlock.p[paramName] = cachedParams[destBlock][paramName]
+        Notes
+        -----
+        This is implemented to ensure that if certain parameters are not set on the original
+        block that the destination block has the parameter data recovered rather than zeroing
+        the data out. The destination block is cleared before the mapping occurs in ``clearStateOnAssemblies``.
+        """
+
+        # For parameters that have not been set on the destination block, recover these
+        # back to their originals based on the cached values.
+        for paramName in paramNames:
+
+            # Skip over any parameter names that were not previously cached.
+            if paramName not in cachedParams[destBlock]:
+                continue
+
+            if isinstance(destBlock.p[paramName], numpy.ndarray):
+                # Using just all() on the list/array is not sufficient because if a zero value exists
+                # in the data then this would then lead to overwritting the data. This is an edge case see
+                # in the testing, so this excludes zero values on the check.
+                if (
+                    not all([val for val in destBlock.p[paramName] if val != 0.0])
+                    or not destBlock.p[paramName].size > 0
+                ):
+                    destBlock.p[paramName] = cachedParams[destBlock][paramName]
+            elif isinstance(destBlock.p[paramName], list):
+                if (
+                    not all([val for val in destBlock.p[paramName] if val != 0.0])
+                    or not destBlock.p[paramName]
+                ):
+                    destBlock.p[paramName] = cachedParams[destBlock][paramName]
+            else:
+                if not destBlock.p[paramName]:
+                    destBlock.p[paramName] = cachedParams[destBlock][paramName]
 
     @staticmethod
     def clearStateOnAssemblies(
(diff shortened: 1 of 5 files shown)


jakehader:
@onufer, please take a look at this code. I am going to update the release notes, some docstrings, and add a couple more tests still. Need to also check the impacts (if any) to the other internal modules before landing.

armi/reactor/converters/uniformMesh.py
```diff
@@ -329,8 +568,62 @@ def plotConvertedReactor(self):
...
+        pass
+
+    def _computeAverageAxialMesh(self):
+        """Computes an average axial mesh based on the core's reference assembly."""
+        src = self._sourceReactor
+        refAssem = src.core.refAssem
+        refNumPoints = (
+            len(src.core.findAllAxialMeshPoints([refAssem], applySubMesh=True)) - 1
+        )  # pop off zero
+        # skip the first value of the mesh (0.0)
```
onufer, line 596: does this  comment belong a few lines down? if not explain a little more.

armi/reactor/converters/uniformMesh.py
```diff
@@ -329,8 +568,62 @@ def plotConvertedReactor(self):
...
+        allMeshes = []
+        for a in src.core:
+            aMesh = src.core.findAllAxialMeshPoints([a], applySubMesh=True)[1:]
+            if len(aMesh) == refNumPoints:
```
onufer, line 600: this is a  pretty big assumption and deserves a comment.  What if there are multiple assems that have same numb of points?

armi/reactor/converters/uniformMesh.py
```diff
@@ -269,32 +286,254 @@ def makeAssemWithUniformMesh(sourceAssem, newMesh):
...
+            blockArrayParamNames = []
+
+        cachedBlockParamData = collections.defaultdict(dict)
+
+        blocks = []
+        for a in assems:
+            blocks.extend(a.getBlocks())
+        for b in blocks:
+            for scalarParam in blockScalarParamNames:
+                if cache:
```
onufer, line 525: i think this code is same for scalor and array params so you could iterate the 2 lists together if you want to save lines of code.

armi/reactor/converters/uniformMesh.py
```diff
@@ -269,32 +286,254 @@ def makeAssemWithUniformMesh(sourceAssem, newMesh):
...
+                ]
+                if not all(clearedValues):
+                    raise ValueError(
+                        f"The state of {destBlock} on {destinationAssembly} "
+                        f"was not cleared prior to calling ``setAssemblyStateFromOverlaps``. "
+                        f"This is an implementation bug in the mesh converter that should "
+                        f"be reported to the developers. The following parameters should be cleared:\n"
+                        f"Parameters: {paramNames}\n"
+                        f"Values: {existingDestBlockParamVals}"
+                    )
```
jakehader, line 382: This is not new code.

armi/reactor/converters/uniformMesh.py
```diff
@@ -269,32 +286,254 @@ def makeAssemWithUniformMesh(sourceAssem, newMesh):
...
+                    f"This a major bug in the uniform mesh converter that should "
+                    f"be reported to the developers."
+                )
+
+            # Iterate over each of the blocks that were found in the uniform mesh
+            # source assembly within the lower and upper bounds of the destination
+            # block and perform the parameter mapping.
+            updatedDestVals = collections.defaultdict(float)
+
+            _setNumberDensitiesFromOverlaps(destBlock, sourceBlocksInfo)
```
jakehader, line 412: This is new now. Previously this was done outside of this mapping routine, but it's easier if this just does it all IMO.

armi/reactor/converters/uniformMesh.py
```diff
@@ -269,32 +286,254 @@ def makeAssemWithUniformMesh(sourceAssem, newMesh):
...
+
+                existingDestBlockParamVals = getter(destBlock, paramNames)
+                clearedValues = [
+                    True if not val else False for val in existingDestBlockParamVals
+                ]
+                if not all(clearedValues):
+                    raise ValueError(
+                        f"The state of {destBlock} on {destinationAssembly} "
+                        f"was not cleared prior to calling ``setAssemblyStateFromOverlaps``. "
+                        f"This is an implementation bug in the mesh converter that should "
```
onufer, line 378: "this indicates a bug"
"or this indicates an implementation bug"

armi/reactor/converters/uniformMesh.py
```diff
@@ -269,32 +286,254 @@ def makeAssemWithUniformMesh(sourceAssem, newMesh):
...
+                        if (
+                            not all(
+                                [val for val in destBlock.p[paramName] if val != 0.0]
+                            )
+                            or not destBlock.p[paramName]
+                        ):
+                            destBlock.p[paramName] = cachedParams[destBlock][paramName]
+                    else:
+                        if not destBlock.p[paramName]:
+                            destBlock.p[paramName] = cachedParams[destBlock][paramName]
```
jakehader, line 490: This is not new code. Showing up as new due to moving.

armi/reactor/converters/uniformMesh.py
```diff
@@ -529,42 +744,73 @@ def multiGroupParamGetter(block, paramNames):
...
+    def arrayParamGetter(block, paramNames):
+        """Returns a set of list/array values for a given set of parameters on a block."""
+        paramVals = []
+        for paramName in paramNames:
+            val = block.p[paramName]
+            if val is None or len(val) == 0:
+                paramVals.append(None)
+            else:
+                paramVals.append(numpy.array(val))
+        return numpy.array(paramVals, dtype=object)
```
jakehader, line 813: This is not new code. Showing up as new due to moving.

armi/reactor/converters/uniformMesh.py
```diff
@@ -329,8 +568,62 @@ def plotConvertedReactor(self):
...
+        """
+        runLog.debug(
+            f"Creating new assemblies from {self._sourceReactor.core} "
+            f"with a uniform mesh of {self._uniformMesh}"
+        )
+        for sourceAssem in self._sourceReactor.core:
+            newAssem = self.makeAssemWithUniformMesh(sourceAssem, self._uniformMesh)
+            src = sourceAssem.spatialLocator
+            newLoc = self.convReactor.core.spatialGrid[src.i, src.j, 0]
+            self.convReactor.core.add(newAssem, newLoc)
```
jakehader, line 626: This is not new code. Showing up as new due to moving.

armi/reactor/converters/uniformMesh.py
```diff
@@ -269,32 +286,254 @@ def makeAssemWithUniformMesh(sourceAssem, newMesh):
...
+                                sourceBlockOverlapHeight / destinationBlockHeight
+                            )
+
+                        updatedDestVals[paramName] += sourceBlockVal * integrationFactor
+
+                    setter(destBlock, updatedDestVals.values(), updatedDestVals.keys())
+
+                # For parameters that have not been set on the destination block, recover these
+                # back to their originals based on the cached values.
+                for paramName in paramNames:
```
onufer, line 463: suggest making this a private method called
_applyCachedValues 
or something like that

onufer (changes requested):

armi/reactor/converters/uniformMesh.py
onufer: i know its moved code, but maybe just add a comment saying 
# If lengths are the same, this assembly contributes to the uniform mesh

In the future it would be good to confirm that the meshes detected this way are all roughly similar.

jakehader: Right. I'll add a comment to this because this was just what has been there for a _long_ time. I think the point of this is that we want to make sure if there is some burn-up or different thermal expansion for different fuel assemblies that we average out those changes rather than just assuming the refAssem mesh.

jakehader: Added a detailed comment.

armi/reactor/converters/uniformMesh.py
```diff
@@ -561,10 +759,50 @@ def multiGroupParamGetter(block, paramNames):
...
+        for paramName, val in zip(paramNames, vals):
+            block.p[paramName] = val
+
+    @staticmethod
+    def scalarParamGetter(block, paramNames):
+        """Returns a set of float/integer/string values for a given set of parameters on a block."""
+        paramVals = []
+        for paramName in paramNames:
+            val = block.p[paramName]
+            if not val:
```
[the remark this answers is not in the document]
jakehader: I updated to check if its the default instead.

armi/reactor/converters/uniformMesh.py
```diff
@@ -458,36 +461,48 @@ def setAssemblyStateFromOverlaps(
 
                     setter(destBlock, updatedDestVals.values(), updatedDestVals.keys())
 
-                # For parameters that have not been set on the destination block, recover these
-                # back to their originals based on the cached values.
-                for paramName in paramNames:
+                UniformMeshGeometryConverter._applyCachedParamValues(
```
onufer, line 464: probably self.

armi/reactor/converters/uniformMesh.py
onufer: I like the note you added, but right above the if statement is where i was hoping to see a comment, because that if statement is saying 
"if the lenghs are equal, it must be part of the uniform mesh i am trying to create", but what if it just happens to be equal

onufer (approved):

john-science:
Have we verified that this won't break any downstream repos?

So far, I like the PR, but the uniform mesh converter is used wildly, and in many repos.

I just want make sure we have a good eye on if (or exactly what) will need to be fixed downstream.

armi/reactor/converters/uniformMesh.py
jakehader: This is a static method so no self

armi/reactor/converters/uniformMesh.py
jakehader: Okay I'll update again

jakehader:
> Have we verified that this won't break any downstream repos?
> 
> So far, I like the PR, but the uniform mesh converter is used wildly, and in many repos.
> 
> I just want make sure we have a good eye on if (or exactly what) will need to be fixed downstream.

Testing will be run before landing. This doesn't impact any reactor conversions and maintains the same behavior. This just allows for parameter mapping on assemblies rather than only on the reactor.

PR #855, opened 2022-08-24
Fix bug in rotatePins()

Nebbychadnezzar:
## Description

The assignment for `pinLocation` in the `rotatePins` method is broken. I am not sure why it was set up the way it was, but it was using `ParameterCollection.__setitem__()` to make the assignment. The syntax used ended up setting the `name` argument to the tuple `("pinLocation", pinNum)`. That throws a `TypeError` from `setattr()`, which gets caught within `__setitem__()` and then instead makes an assignment to `self._hist` (where `self` is the parameter collection) because `name` is a tuple. That is definitely not what is intended.

This PR fixes this bug. Doc string improvements have also been attempted.

---

## Checklist


Nebbychadnezzar:
Looping in fellow pin aficionado @mgjarrett

armi/reactor/blocks.py
```diff
@@ -1757,38 +1767,43 @@ def rotatePins(self, rotNum, justCompute=False):
...
 
-            if not justCompute:
-                self.p["pinLocation", rotateIndexLookup[pinNum]] = pinNum
-
+        # Because the above math creates indices based on the absolute rotation number,
+        # the old values of pinLocation (if they've been set in the past) can be overwritten
+        # with new numbers
         if not justCompute:
             self.setRotationNum(rotNum)
+            self.p["pinLocation"] = [*rotateIndexLookup.values()]
```
mgjarrett, line 1806: `roateIndexLookup.values()` will probably be sorted in ascending order of the keys... but should we code it explicitly to be sorted this way? For example:

`self.p["pinLocation"] = [v for k, v in sorted(rotateIndexLookup.items())]`

mgjarrett: Or
`self.p["pinLocation"] = [ rotateIndexLookup[pinNum] for pinNum in range(1, numPins + 1) ]`

Nebbychadnezzar: I think your way is more robust. I believe `values()` returns based on the order of how keys are added to the dictionary since Python maintains dictionary order based on how the keys are added to the dictionary. Since `rotateIndexLookup` is declared with the desired order, then `keys()` and `values()` will maintain that order. But if someone were to edit that declaration, then `values()` may no longer yield the desired result.

I will make this change.

mgjarrett (approved):
Great find and fix!

jakehader (approved):

john-science:
Can someone help me out here?

What happens when you call `rotatePins()` on a `Reactor` that doesn't have pins?  Two of three big reactors ARMI supports (off the top of my head) don't have pins.

For those users... does `rotatePins()` fail gracefully?

mgjarrett:
It's only on the `HexBlock`. If you have a `HexBlock` without pins, you probably don't have `Flags.CLAD`, so it will just produce a null lookup dictionary, `{}`.

If you do have a `Flags.CLAD` component with a multiplicity other than 1, it will produce a lookup table. The table will only be meaningful if your `Flags.CLAD` components are in a regular hexagonal lattice.

john-science:
> It's only on the `HexBlock`. 

AH!  I'm just blind.  (And probably sleep deprived.)  

Thanks!

PR #860, opened 2022-08-25
Fixing Component density to agree with 3D density

### LINKED ISSUE #820: Component density does not exactly agree with 3D density

onufer:
## Description

Resolves #820 so component agrees exactly with 3D density. Removed some testing work arounds.

## Checklist

### DIFF
--- a/armi/materials/zr.py
+++ b/armi/materials/zr.py
@@ -130,7 +130,7 @@ class Zr(Material):
         Tk = getTk(Tc, Tk)
         self.checkPropertyTempRange("linear expansion percent", Tk)
 
-        if Tk >= 291.62 and Tk < 1137:
+        if Tk >= 293 and Tk < 1137:
             return (
                 -0.111 + (2.325e-4 * Tk) + (5.595e-7 * Tk ** 2) - (1.768e-10 * Tk ** 3)
             )
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -232,26 +232,10 @@ class Component(composites.Composite, metaclass=ComponentType):
         self.temperatureInC = Thot
         self.material = None
         self.setProperties(material)
-        self.tInputWarning(Tinput)
         self.applyMaterialMassFracsToNumberDensities()  # not necessary when duplicating...
         self.setType(name)
         self.p.mergeWith = mergeWith
         self.p.customIsotopicsName = isotopics
-
-    def tInputWarning(self, Tinput):
-        """
-        Check whether thermal expansion factor is 0.0% exactly at T=Tinput
-        """
-        expansionFactor = (
-            self.material.linearExpansionPercent(Tc=self.inputTemperatureInC) / 100.0
-        )
-        if not (abs(expansionFactor) < 1.0e-6):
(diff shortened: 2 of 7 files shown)


onufer:
@mgjarrett, please feel free to add comments

armi/materials/zr.py
```diff
@@ -130,7 +130,7 @@ def linearExpansionPercent(self, Tk=None, Tc=None):
         Tk = getTk(Tc, Tk)
         self.checkPropertyTempRange("linear expansion percent", Tk)
 
-        if Tk >= 291.62 and Tk < 1137:
```
onufer, line 133: This reverted a change that was needed to change the blueprints which also were not needed.

armi/reactor/components/component.py
```diff
@@ -382,13 +367,24 @@ def applyHotHeightDensityReduction(self):
...
+            )
+            self.changeNDensByFactor(1.0 / coldMatAxialExpansionFactor)
+        else:
+            # this is the same as getThermalExpansionFactor but doesn't fail
+            # on non-fluid materials that have 0 or undefined thermal expansion
+            # (we don't want materials to fail on  __init__ which calls this)
+            # axialExpansionFactor = 1.0 + self.material.linearExpansionFactor(
+            #     self.temperatureInC, self.inputTemperatureInC
+            # )
+            self.changeNDensByFactor(1.0 / self.getThermalExpansionFactor())
```
mgjarrett, line 387: The `else` seems to only be invoked in tests, as far as I can tell. Could we just remove the `initialColdMaterialExpansion` argument, and then rewrite and/or remove those tests as appropriate?

onufer: It is invoked in a TP submodule. I would also expect it could also be useful for axial expansion, although it is not used now.

onufer: would moving the code from the if statement into `applyMaterialMassFracsToNumberDensities` be acceptable, and then the if statement would be eliminated but we would still have a useful hot height density reduction changer.

mgjarrett: If the code is used then I don't have a problem with it being here. I was just proposing removing the code if in fact it was not being used anymore.

onufer:
wait for @ntouran to have a chance to provide feedback to merge

armi/reactor/tests/test_components.py
```diff
@@ -586,19 +586,15 @@ def test_ExpansionConservationHotHeightDefined(self):
             circle1.temperatureInC, circle1.inputTemperatureInC
         )
         factorToUndoHotHeight = circle1.getThermalExpansionFactor()
```
mgjarrett, line 588: `factorToUndoHotHeight` is unused now

onufer: removed

mgjarrett (approved):

jakehader:
Taking myself off review @onufer. I think that @mgjarrett and @ntouran can have this covered.

PR #867, opened 2022-09-02
Moving newReports into report package

john-science:
## Description

About a year ago, some new reporting tools were added to ARMI: "newReports.py".

Unfortunately, the transition was not fully made form the old report format to the new.  This will have to be fixed.

But step 1 of making this transition from the old to the new reports is to move the new reports into the `/report/` package/folder/directory.  This PR accomplishes that.

(Step 2 will be to determine if the new reports are complete enough for us to transition. Step 3 will be to make some transition, even if we have to move some tools from the old to new reports first. But those are other PRs.)

---

## Checklist


john-science:
The code coverage problem is super minor, but seems to be due to some sort of random slop in our unit tests. Essentially, these two lines were (ostensibly) uncovered:

https://coveralls.io/builds/52181036/source?filename=armi%2Fphysics%2Fneutronics%2Freports.py#L37

But when I look back at our past code coverage reports, these are usually covered.

It's probably fine, but it might be nice to add a little test case to force these two lines to always be covered, so we don't have this problem in the future.

opotowsky (approved):
Sorry this took so long. This looks good and I'm ok with the weirdo coverage situation :-)

john-science:
> I'm ok with the weirdo coverage situation :-)

I have fixed the coverage situation, by adding another test.

PR #872, opened 2022-09-08
Removing user defined blueprints from snapshot runs

### LINKED ISSUE #871: Remove the ability for users to load separate blueprints in snapshots run
When doing a snapshots run, ARMI allows users to use both new settings and blueprints. 
https://github.com/terrapower/armi/blob/0716e80f3d03ce188a1859b73fe02e40a09747ea/armi/bookkeeping/db/database3.py#L1103-L1106
 
I can understand the utility in using new settings, but the option for different blueprints is a little strange to me. I would think that if one unintentionally loaded a blueprints file that didn't match the blueprints assemblies on the reactor database there would be unexpected issues. 
 
Looking at the docstring in `loadBlueprints` I can totally appreciate the danger of unexpected things happening if the database was written with a different ARMI version. But I think(?) it might be the lesser of two downsides compared to unintentionally loading a different blueprints file than what's on the database.
https://github.com/terrapower/armi/blob/0716e80f3d03ce188a1859b73fe02e40a09747ea/armi/bookkeeping/db/database3.py#L815-L823

To avoid having users unintentionally load blueprints that differ from what is on their reactor database, I propose that we require the reactor be loaded with the blueprints that live on the database. This would effectively remove the optional param, `bp`, from database3.py::Database3::load.

albeanth:
## Description
When doing a snapshots run, ARMI allows users to use both new settings and blueprints. This PR removes the ability to use new (potentially different) blueprints. 

Effects:
- Prevents users from unintentionally applying blueprints that are different than what is stored on the H5 database. 
- If the database was written with a different version of ARMI the db load may not work. This could occur if an overhaul is done to the database system (though doing such a task would likely break a slew of other things as well). 

More relevant information can be found in #871. This PR will close #871 as well. 

---

## Checklist


albeanth:
Note: Testing impacts on downstream projects. Reverting to draft.

armi/reactor/blueprints/__init__.py
```diff
@@ -304,10 +304,20 @@ def _prepConstruction(self, cs):
             self._assembliesBySpecifier.clear()
             self.assemblies.clear()
 
+            # retrieve current count of assemblies to restore after
+            # creating blueprints assemblies. This is particularly useful for
+            # doing snapshot based runs with database loads where the
+            # assembly counter is likely not equal to 0.
```
onufer, line 310: Add "To ensure that the assembly count is not a function of how many times load state/snapshots has been called"

onufer (approved):

PR #873, opened 2022-09-08
Removing the unused / unnecessary DB baseclass

john-science:
## Description

There must be some historical reason for this `Database3(Database)` class situation.  But, looking through the history, the `Database` class has been unnecessary for at least 4 years now.  So I removed it.

Per Discussion: https://github.com/terrapower/armi/discussions/633#discussioncomment-3600564

Will removing code, also lower the code coverage?  Probably. 

---

## Checklist


john-science:
Okay, the code coverage went down because I removed lines of code.  To be expected.

I'll see if I can add some more unit tests to prop it back up again.

opotowsky (approved):
I'm all about that cleanup! And kudos for worrying about the reduction in code coverage. This is a case where I'd not worry about it since you didn't change anything about the coverage in a qualitative sense. It's just a numbers effect from removing lines of code? I don't always understand the coverage situation...Does the coverage diff get generated by looking at the numerator and not the denominator? 

Anyway, merge whenever you're ready.

PR #875, opened 2022-09-09
Uniform mesh refactor to simplify block-level parameter mapping

jakehader:
Small refactor of the `UniformMeshConverter` to allow users to provide just the block-level parameter names as a single list rather than requiring the users/developers to differentiate in the code which parameters are scalar or array values. This will now auto-detect the type based on the parameter definition and then apply the correct setters/getters.

Add an optional argument the `makeAssemWithUniformMesh` and `setAssemblyStateFromOverlaps` staticmethods to disable the mapping of number densities on the creation of a new uniform assembly or when mapping block-level parameter data between two assemblies. This allows for more application-dependent flexibility for handling edge cases. An example of where this is useful is in the situation of moving meshes in vertical control rod movements.

## Description

---

## Checklist


opotowsky (approved):
This looks fantastic. 

And unit tests passing, coverage is good...merging!

PR #877, opened 2022-09-13
Allow for onProcessCoreLoading to know if database is being loaded

Nebbychadnezzar:
## Description

The docstring for `armi.plugins.ArmiPlugin.onProcessCoreLoading()` suggests that the method is called when either constructing a Core from a blueprints file or from loading a database. Wouldn't it be nice if the method knew if one or the other was occurring, in case that might be useful information for the implementing plugin?

This PR proposes such a nicety.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/__init__.py
+++ b/armi/physics/neutronics/__init__.py
@@ -137,7 +137,7 @@ class NeutronicsPlugin(plugins.ArmiPlugin):
 
     @staticmethod
     @plugins.HOOKIMPL
-    def onProcessCoreLoading(core, cs):
+    def onProcessCoreLoading(core, cs, dbLoad):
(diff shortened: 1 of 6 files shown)


jakehader:
@john-science - I worked with @Nebbychadnezzar on this and it seems reasonable from my perspective that `OnProcessCoreLoading` would have access to whether or not the core is being loaded from a database. The main thing here is that we do have some branching in `core::processLoading` for when `dbLoad` is set to `True` or `False` and it seems natural to propagate this downwards to the plugins. Also, from the docstring it looks like this should have been included during its inception:

```
        Function to call whenever a Core object is newly built.
        This is usually used to set initial parameter values from inputs, either after
        constructing a Core from Blueprints, **or after loading it from a database**.
````

Issue #876, opened 2022-09-13
Inconsistency with 'sfp' SystemBlueprint object and o.r.core.sfp


albeanth:
## The problem/observation
- in marching through `armi/reactor/tests/test_reactors.py::loadTestReactor`, there are two designs stored in `bp.systemDesigns`, `Core` and `sfp`. https://github.com/terrapower/armi/blob/55766354067d8b1cffed02a48d36cd5d59c10718/armi/reactor/reactors.py#L145-L147
- The `core` object is built just before this for loop here, https://github.com/terrapower/armi/blob/55766354067d8b1cffed02a48d36cd5d59c10718/armi/reactor/reactors.py#L143-L144. For `loadTestReactor` it gets 73 assemblies in core and 8 blueprints assemblies. 
- The `sfp` object is then built in `structure.construct`. It contains the 4 sfp assemblies defined in [the input](https://github.com/terrapower/armi/blob/main/armi/tests/refSmallSfpGrid.yaml) and 8 blueprints assemblies.

After both objects have been constructed, when `r` is returned the `sfp` object seems to be missing. 
- `r.sfp` is not a thing...
- `r.core.sfp` is an "AssemblyList object: Spent Fuel Pool", but contains no assemblies. 

Then in `loadTestReactor`, we put stuff into r.core.sfp https://github.com/terrapower/armi/blob/55766354067d8b1cffed02a48d36cd5d59c10718/armi/reactor/tests/test_reactors.py#L188-L191
- So, now we have these 10 assemblies in `r.core.sfp`, but the original 4 assemblies from the input are no where to be found (at least I can't find them). 

## Impacts
For the most part, I don't think this a huge deal (otherwise we would have found it by now). But, it does raise an issue if a user assigns assemblies in the SFP via the inputs or wants to see what assemblies are in the sfp on a database load.
- Since there isn't a registered `r.sfp` object and `r.core.sfp` doesn't get written to the database (at least in my tests it doesn't), then there is no way to check what is actually in the sfp on a database load (either put in via a shuffle or initialized there via an input).
  - I imagine this is certainly something we'll want to support. Knowing what is in the SFP feels important. Especially when we're shuffling assemblies out of the core. 
- There is a disconnect between where ARMI stores SFP assemblies from inputs (in Core: sfp object) vs where it stashes them as a result of shuffling (in r.core.sfp). 

[...]

PR #879, opened 2022-09-14
Allow treatment of non-uniform assemblies in the uniform mesh converter

mgjarrett:
## Description

The uniform mesh module was updated to treat the conversion of non-uniform assemblies in the core model (using the `nonUniformAssemFlags` setting). In this change, a new branch of code is introduced within the `UniformMeshConverter` that short cuts the entire new reactor creation and entire reactor remeshing/data mapping when the `nonUniformAssemFlags` is provided by the user. This new feature allows for significant acceleration of the reactor unification because only a subset of all the assemblies need to be converted.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -15,6 +15,7 @@
 """
 The Global flux interface provide a base class for all neutronics tools that compute the neutron and/or photon flux.
 """
+import copy
 import math
 from typing import Dict, Optional
 
@@ -328,7 +329,10 @@ class GlobalFluxOptions(executers.ExecutionOptions):
         self.adjoint = neutronics.adjointCalculationRequested(cs)
         self.real = neutronics.realCalculationRequested(cs)
         self.detailedAxialExpansion = cs["detailedAxialExpansion"]
-        self.hasNonUniformAssems = any(cs["nonUniformAssemFlags"])
+        self.nonUniformMeshFlags = [
+            Flags.fromStringIgnoreErrors(f) for f in cs["nonUniformAssemFlags"]
+        ]
+        self.hasNonUniformAssems = any(self.nonUniformMeshFlags)
         self.eigenvalueProblem = cs["eigenProb"]
 
         # dose/dpa specific (should be separate subclass?)
@@ -411,11 +415,49 @@ class GlobalFluxExecuter(executers.DefaultExecuter):
                     None,
                     calcReactionRates=self.options.calcReactionRatesOnMeshConversion,
                 )
-                converter.convert(self.r)
-                neutronicsReactor = converter.convReactor
-                if makePlots:
-                    converter.plotConvertedReactor()
-                self.geomConverters["axial"] = converter
+                if self.options.detailedAxialExpansion:
+                    converter.convert(self.r)
+                    neutronicsReactor = converter.convReactor
+
+                # Here we are taking a short cut to homogenizing the core by only focusing on the
+                # core assemblies that need to be homogenized. This will have a large speed up
+                # since we don't have to create an entirely new reactor perform the data mapping.
+                # We are using the Spent Fuel Pool (SFP) as a container just to temporarily store
+                # the detailed assembly so that it can be recovered later when the geometry
+                # conversions are undone.
+                elif (
+                    not self.options.detailedAxialExpansion
+                    and self.options.hasNonUniformAssems
+                ):
+                    converter = uniformMesh.NeutronicsUniformMeshConverter(
+                        None,
+                        calcReactionRates=self.options.calcReactionRatesOnMeshConversion,
+                    )
+                    uniformMesh = converter._computeAverageAxialMesh()
+                    b = self.r.core.getFirstBlock()
+                    blockParamNames = []
+                    self.blockParamNames.extend(
+                        b.p.paramDefs.inCategory(
+                            parameters.Category.detailedAxialExpansion
+                        ).names
+                    )
+                    self.blockParamNames.extend(
+                        b.p.paramDefs.inCategory(
+                            parameters.Category.multiGroupQuantities
+                        ).names
+                    )
+                    for assem in self.r.core.getAssemblies(
+                        self.options.nonUniformMeshFlags
+                    ):
+                        storedAssem = copy.deepcopy(assem)
+                        core.sfp.add(storedAssem)
+                        converter.makeAssemWithUniformMesh(
+                            assem, uniformMesh, blockParamNames
+                        )
+
+                    if makePlots:
+                        converter.plotConvertedReactor()
+                    self.geomConverters["axial"] = converter
 
         if self.edgeAssembliesAreNeeded():
             converter = self.geomConverters.get(
@@ -457,15 +499,49 @@ class GlobalFluxExecuter(executers.DefaultExecuter):
         meshConverter = self.geomConverters.get("axial")
         if meshConverter:
             if self.options.applyResultsToReactor:
-                meshConverter.applyStateToOriginal()
 
-            self.r = meshConverter._sourceReactor  # pylint: disable=protected-access;
+                # For the case of having non-uniform assemblies in the core,
+                # we are undoing the modifications using a short cut that doesn't
+                # require the entire reactor to be mapped back.
+                if self.options.hasNonUniformAssems:
+                    b = self.r.core.getFirstBlock()
+                    blockParamNames = []
+                    self.blockParamNames.extend(
+                        b.p.paramDefs.inCategory(
+                            parameters.Category.detailedAxialExpansion
+                        ).names
+                    )
+                    self.blockParamNames.extend(
+                        b.p.paramDefs.inCategory(
+                            parameters.Category.multiGroupQuantities
+                        ).names
+                    )
+                    for assem in self.r.core.getAssemblies(
+                        self.options.nonUniformMeshFlags
+                    ):
+                        for storedAssem in self.r.core.sfr.getChildren():
+                            if storedAssem.getName() == assem.getName():
+                                meshConverter.setAssemblyStateFromOverlaps(
+                                    assem,
+                                    storedAssem,
+                                    blockParamNames,
+                                    mapNumberDensities=False,
+                                )
+                                self.r.core.remove(assem)
+                                storedAssem.spatialLocator = assem.spatialLocator
+                                self.r.core.add(storedAssem)
+                                break
+                else:
+                    meshConverter.applyStateToOriginal()
+                    self.r = (
+                        meshConverter._sourceReactor
+                    )  # pylint: disable=protected-access;
 
-            # Resets the stored attributes on the converter to
-            # ensure that there is state data that is long-lived on the
-            # object in case the garbage collector does not remove it.
-            # Additionally, this will reset the global assembly counter.
-            meshConverter.reset()
+                    # Resets the stored attributes on the converter to
+                    # ensure that there is state data that is long-lived on the
+                    # object in case the garbage collector does not remove it.
+                    # Additionally, this will reset the global assembly counter.
+                    meshConverter.reset()
 
         # clear the converters in case this function gets called twice
         self.geomConverters = {}
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -438,9 +438,7 @@ class GlobalFluxExecuter(executers.DefaultExecuter):
                     b = self.r.core.getFirstBlock()
                     blockParamNames = []
                     for category in converter.BLOCK_PARAM_MAPPING_CATEGORIES:
-                        blockParamNames.extend(
-                            b.p.paramDefs.inCategory(category).names
-                        )
+                        blockParamNames.extend(b.p.paramDefs.inCategory(category).names)
                     for assem in self.r.core.getAssemblies(
                         self.options.nonUniformMeshFlags
                     ):
@@ -504,9 +502,7 @@ class GlobalFluxExecuter(executers.DefaultExecuter):
                     b = self.r.core.getFirstBlock()
                     blockParamNames = []
                     for category in meshConverter.BLOCK_PARAM_MAPPING_CATEGORIES:
-                        blockParamNames.extend(
-                            b.p.paramDefs.inCategory(category).names
-                        )
+                        blockParamNames.extend(b.p.paramDefs.inCategory(category).names)
                     for assem in self.r.core.getAssemblies(
                         self.options.nonUniformMeshFlags
                     ):
@@ -518,6 +514,7 @@ class GlobalFluxExecuter(executers.DefaultExecuter):
                                     blockParamNames,
                                     mapNumberDensities=False,
                                 )
+                                self.r.core.sfp.remove(storedAssem)
                                 storedAssem.spatialLocator = assem.spatialLocator
                                 self.r.core.removeAssembly(assem, discharge=False)
                                 self.r.core.add(storedAssem)
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -28,7 +28,7 @@ from armi.utils import units, codeTiming, getMaxBurnSteps
 from armi.reactor import parameters
 from armi.reactor import geometry
 from armi.reactor import reactors
-from armi.reactor.converters import uniformMesh
+from armi.reactor.converters.uniformMesh import NeutronicsUniformMeshConverter
 from armi.reactor.converters import geometryConverters
 from armi.reactor import assemblies
 from armi.reactor.flags import Flags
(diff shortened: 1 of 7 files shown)


armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -327,6 +328,7 @@ def fromUserSettings(self, cs):
         self.adjoint = neutronics.adjointCalculationRequested(cs)
         self.real = neutronics.realCalculationRequested(cs)
         self.detailedAxialExpansion = cs["detailedAxialExpansion"]
+        self.nonUniformAssems = len(cs["nonUniformAssemFlags"]) > 0
```
jakehader, line 331: Can this just be `self.nonUniformAssems = cs["nonUniformAssemFlags"]`?

armi/reactor/converters/uniformMesh.py
```diff
@@ -115,7 +115,9 @@ def convert(self, r=None):
         self._setParamsToUpdate()
         self._computeAverageAxialMesh()
         self._buildAllUniformAssemblies()
-        self._mapStateFromReactorToOther(self._sourceReactor, self.convReactor)
+        self._mapStateFromReactorToOther(
+            self._sourceReactor, self.convReactor, mapNumberDensities=False
```
mgjarrett, line 119: We need `mapNumberDensities=False` here in a `self.options.nonUniformAssems` case. But I'm not sure it will be right for `self.options.detailedAxialExpansion`. I will need to test this.

mgjarrett, line 119: Number densities are mapped during `_buildAllUniformAssemblies`; don't need to do it here.

armi/reactor/converters/uniformMesh.py
```diff
@@ -588,20 +592,12 @@ def _computeAverageAxialMesh(self):
...
+        self._uniformMesh = numpy.array(
+            src.core.findAllAxialMeshPoints([src.core.refAssem], applySubMesh=True)[1:]
         )
-        allMeshes = []
-        for a in src.core:
-            # Get the mesh points of the assembly, neglecting the first coordinate (typically zero).s
-            aMesh = src.core.findAllAxialMeshPoints([a], applySubMesh=True)[1:]
-            if len(aMesh) == refNumPoints:
-                allMeshes.append(aMesh)
-        self._uniformMesh = average1DWithinTolerance(numpy.array(allMeshes))
```
mgjarrett, line 604: The mesh averaging was giving me trouble during early stages of development, so I switched to just using the mesh directly from the `refAssem`. I will likely revert this change and test a few cases to see if it is working with the mesh averaging.

armi/reactor/reactors.py
```diff
@@ -1862,21 +1861,14 @@ def updateAxialMesh(self):
...
-                    for a in self
-                    if self.findAllAxialMeshPoints([a]) == refMesh
-                ]
-            )
+        meshHeights = numpy.array(
+            [
+                h
+                for b in refAssem
+                for h in [(b.p.ztop - b.p.zbottom) / b.p.axMesh] * b.p.axMesh
+            ]
```
mgjarrett, line 1869: Same story as above with respect to mesh averaging. I would like to test to see if htis change can be reverted before merging. If not, the viability of this approach could be questionable because changing the way `core.p.axialMesh` is computed could have lots of downstream effects.

armi/reactor/reactors.py
```diff
@@ -2287,8 +2276,6 @@ def processLoading(self, cs, dbLoad: bool = False):
             if not cs["detailedAxialExpansion"]:
                 # prepare core for mesh snapping during axial expansion
                 for a in self.getAssemblies(includeAll=True):
-                    if a.hasFlags(nonUniformAssems, exact=True):
-                        continue
```
mgjarrett, line 2291: Needed to make the `axialSnapList` for the non-uniform assemblies to snap them to the same mesh as the core when thermal expansion is on (i.e., `inputHeightsConsideredHot: false`). This is another piece of the PR that I am planning to do more testing on and see if it can be reverted or somehow handled differently.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
mgjarrett: `self.nonUniformAssems` is supposed to be a bool, but `cs["nonUniformAssemFlags"]` is a list.
Technically, it `self.nonUniformAssems` is a list it would probably evaluate to the correct `bool` value when we use it in logical statements, but it seems better for it to explicitly be a bool.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -402,7 +404,7 @@ def _performGeometryTransformations(self, makePlots=False):
                 + "This is a programming error and requires further investigation."
             )
         neutronicsReactor = self.r
-        if self.options.detailedAxialExpansion:
+        if self.options.detailedAxialExpansion or self.options.nonUniformAssems:
             converter = self.geomConverters.get("axial")
```
jakehader, line 408: To make this more efficient, I recommend that we have an option to handle remeshing _only_ the nonUniformAssems if possible. Remeshing the entire reactor will be time consuming and actually not needed here. I will look at the code and suggest a change.

armi/reactor/converters/uniformMesh.py
```diff
@@ -178,7 +180,9 @@ def applyStateToOriginal(self):
         # parameters that did not change.
         self._cachedReactorCoreParamData = {}
         self._clearStateOnReactor(self._sourceReactor, cache=True)
-        self._mapStateFromReactorToOther(self.convReactor, self._sourceReactor)
+        self._mapStateFromReactorToOther(
+            self.convReactor, self._sourceReactor, mapNumberDensities=False
```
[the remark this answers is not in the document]
jakehader: This seems reasonable.

armi/reactor/converters/uniformMesh.py
jakehader: I think we want the average mesh so that it accounts for the fuel burnup or other thermal expansion.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
mgjarrett: Yeah, I tried an approach like this but what I had ended up being just as time consuming as remeshing the whole reactor. I can show you what I had and get your thoughts.

armi/reactor/converters/uniformMesh.py
```diff
@@ -296,7 +300,7 @@ def makeAssemWithUniformMesh(
             block = copy.deepcopy(sourceBlock)
             block.p.xsType = xsType
             block.setHeight(topMeshPoint - bottom)
-            block.p.axMesh = 1
+            block.p.axMesh = sourceBlock.p.axMesh
```
Nebbychadnezzar, line 303: I am not sure which is "right" here...do we expect this to create any problems? e.g., a new small ARMI block borrowing from a larger ARMI block with axMesh = 4.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
Nebbychadnezzar: Minor comment, but consider making this `hasNonUniformAssems`.

jakehader: Let's do any([]) then

armi/physics/neutronics/globalFlux/globalFluxInterface.py
jakehader: Okay I'm good then

armi/reactor/converters/uniformMesh.py
```diff
@@ -588,17 +592,16 @@ def _computeAverageAxialMesh(self):
...
-        )
+
+        refNumPoints = len(src.core.findAllAxialMeshPoints([refAssem])) - 1
         allMeshes = []
         for a in src.core:
-            # Get the mesh points of the assembly, neglecting the first coordinate (typically zero).s
-            aMesh = src.core.findAllAxialMeshPoints([a], applySubMesh=True)[1:]
+            # Get the mesh points of the assembly, neglecting the first coordinate
+            # (typically zero).
+            aMesh = src.core.findAllAxialMeshPoints([a])[1:]
```
jakehader, line 604: Can you explain why we don't want the submeshes? I think DIF3D will still need these

mgjarrett: `applySubMesh=True` is the default. I was having trouble with the submeshes originally in downstream modules, but I believe it's resolved now. Trying to move forward with this current version, which includes the submesh.

armi/reactor/converters/uniformMesh.py
jakehader: I think that it might be reasonable to base the axial mesh size based on the ratio of the sourceBlock height and its defined axial mesh points so that the aspect ratios remain as consistent as possible.

`--> (block.getHeight() / block.p.axMesh) = (sourceBlock.getHeight() / source.p.axMesh)`
`-->  block.p.axMesh = block.getHeight() / (sourceBlock.getHeight() / source.p.axMesh)`
`--> block.p.axMesh = (block.getHeight() / sourceBlock.getHeight()) * source.p.axMesh)`

To prevent a zero value, we can do:

`--> block.p.axMesh = (block.getHeight() / sourceBlock.getHeight()) * source.p.axMesh) or 1`

or 

`--> block.p.axMesh = math.ceil(block.getHeight() / sourceBlock.getHeight()) * source.p.axMesh))`

jakehader: I think that it might be reasonable to base the axial mesh size based on the ratio of the sourceBlock height and its defined axial mesh points so that the aspect ratios remain as consistent as possible.

`--> (block.getHeight() / block.p.axMesh) = (sourceBlock.getHeight() / source.p.axMesh)`
`-->  block.p.axMesh = block.getHeight() / (sourceBlock.getHeight() / source.p.axMesh)`
`--> block.p.axMesh = (block.getHeight() / sourceBlock.getHeight()) * source.p.axMesh)`

To prevent a zero value, we can do:

`--> block.p.axMesh = (block.getHeight() / sourceBlock.getHeight()) * source.p.axMesh) or 1`

or 

`--> block.p.axMesh = math.ceil(block.getHeight() / sourceBlock.getHeight()) * source.p.axMesh))`

jakehader: What do you think @Nebbychadnezzar / @mgjarrett ?

Nebbychadnezzar: That seems reasonable to me.

jakehader: Looks like @mgjarrett went back to how it is currently implemented so no change there.

jakehader (approved):

armi/reactor/converters/uniformMesh.py
mgjarrett: yep, sorry about the confusion. I was having a lot of mesh issues early on and I ended up making a lot of wrong turns before I figured out the root cause.

jakehader:
Thanks @albeanth - I refactored the code changes to just be in the `UniformMeshConverter` and added a simple test that executes the conversion using the `nonUniformAssemFlags` on the test reactor for code coverage. We have more in-depth coverage internally though!

armi/reactor/converters/uniformMesh.py
```diff
@@ -102,25 +106,74 @@ def __init__(self, cs=None):
...
+        ]
+        self._hasNonUniformAssems = any(self._nonUniformMeshFlags)
 
     def convert(self, r=None):
         """Create a new reactor core with a uniform mesh."""
         if r is None:
             raise ValueError(f"No reactor provided in {self}")
 
-        runLog.extra("Building copy of {} with a uniform axial mesh".format(r))
+        runLog.extra(f"Building with a uniform axial mesh reactor from {r}")
```
albeanth, line 119: This new runLog statement seems less clear to me. What was the motivation for changing it?

armi/reactor/converters/uniformMesh.py
```diff
@@ -105,27 +106,74 @@ def __init__(self, cs=None):
...
+                self.convReactor.core.sfp.add(assem)
+                self.convReactor.core.add(homogAssem)
+
+            self.convReactor.core.updateAxialMesh()
+
+            self._checkConversion()
+            completeEndTime = timer()
+            runLog.extra(
+                f"Reactor core conversion time: {completeEndTime-completeStartTime} seconds"
+            )
```
mgjarrett, line 162: We should just use the runLog after the if/else

armi/reactor/converters/uniformMesh.py
jakehader: In the case of using non uniform assemblies we don't copy the reactor.

armi/reactor/converters/uniformMesh.py
jakehader: Yeah that's a good idea

armi/reactor/converters/uniformMesh.py
albeanth: Got it. What do you think about having two `runLog` statements then? One for the `self._hasNonUniformAssems` case and one for the other? Something like... 
```python
if self._hasNonUniformAssems:
    runLog.extra(f"Temporarily replacing non-uniform assemblies in reactor, {r}, with assemblies whose axial mesh is uniform with r.core.refAssem.getAxialMesh().")
    ....
else:
    runLog.extra(f"Building copy of {r} with a uniform axial mesh.")
```

armi/reactor/converters/uniformMesh.py
albeanth: Taking it one step further - it looks like this block can be safely moved out of the if/else. 

```python
self.convReactor.core.updateAxialMesh()
self._checkConversion()
completeEndTime = timer()
runLog.extra(
    f"Reactor core conversion time: {completeEndTime-completeStartTime} seconds"
)
```

armi/reactor/converters/uniformMesh.py
```diff
@@ -102,23 +106,66 @@ def __init__(self, cs=None):
...
-        self._computeAverageAxialMesh()
-        self._buildAllUniformAssemblies()
-        self._mapStateFromReactorToOther(self._sourceReactor, self.convReactor)
-        self.convReactor.core.updateAxialMesh()
 
-        self._newAssembliesAdded = self.convReactor.core.getAssemblies()
+        # Here we are taking a short cut to homogenizing the core by only focusing on the
+        # core assemblies that need to be homogenized. This will have a large speed up
+        # since we don't have to create an entirely new reactor perform the data mapping.
+        # We are using the Spent Fuel Pool (SFP) as a container just to temporarily store
```
albeanth, line 126: So I get that in this context the SFP serves as a temporary container and assemblies that get added to it here are soon removed down in `_mapStateFromReactorToOther`. But it feels like this could be error prone in the long term. Moving assemblies to a data structure that can persist outside of this class feels.... fishy. Especially when it's filling a short term/temporary local namespace storage need.

I wonder if it would be safer to create an instance variable (e.g., `self.tempAssemblyStorage = []` or something) that we know will go out of scope when the class instance is finished up. If/when we tackle #876 we may have to make this change anyway. 

Just something to consider.

armi/reactor/converters/uniformMesh.py
```diff
@@ -171,18 +222,66 @@ def applyStateToOriginal(self):
...
+        else:
+            # Clear the state of the original source reactor to ensure that
+            # a clean mapping between the converted reactor for data that has been
+            # changed. In this case, we cache the original reactor's data so that
+            # after the mapping has been applied, we can recover data from any
+            # parameters that did not change.
+            self._cachedReactorCoreParamData = {}
+            self._clearStateOnReactor(self._sourceReactor, cache=True)
+            self._mapStateFromReactorToOther(
+                self.convReactor, self._sourceReactor, mapNumberDensities=False
```
albeanth, line 278: To maintain the previous behavior here, shouldn't `mapNumberDensities = True`?

armi/reactor/converters/uniformMesh.py
jakehader: Got it! Updated.

albeanth: Awesome! Can you update the comment too? The big explanatory one right before `if self._hasNonUniformAssems`.

armi/reactor/converters/uniformMesh.py
jakehader: At least that is the assumption here. Maybe it's "safer" right now just to keep the same behavior so we don't introduce some weird bug...

jakehader: Okay updated back to how it was.

armi/reactor/converters/uniformMesh.py
albeanth: https://github.com/terrapower/armi/pull/879/commits/ac6bc4324de00c0029fb9dea4fdde121e4ad7585
https://github.com/terrapower/armi/pull/879/commits/839885e81835c9a9d2d39d6db393234ddedf00e9
https://github.com/terrapower/armi/pull/879/commits/f84a5b865fe5f7e6088853cc44964b07c60aeb88

armi/reactor/converters/uniformMesh.py
albeanth: https://github.com/terrapower/armi/pull/879/commits/522d885ddf78aa7ce9224cb9f2c0ce18212cab47

If you believe that it's not needed, let's make an issue for it and get it in a subsequent PR 👍

albeanth (approved):
One comment for our future selves:

Within `UniformMeshGeometryConverter::convert`, the optional param `mapNumberDensities=False` is changed since the number density mapping already occurs in `_buildAllUniformAssemblies` and was therefore a pre-existing redundancy.

PR #882, opened 2022-09-14
Density3 temp k

sombrereau:
## Description

An ARMI user was observing differences in density results due to the https://github.com/terrapower/armi/pull/868 caused in materials that are subclasses of Fuel Material. Fuel Material initially inherited from SimpleSolid, but in reality Fuel Materials are not really "Simple Solid Materials".

I refactored the lineage of Fuel Materials, back to the original lineage. This simplified the Fuel Materials because I didn't need to overload density functions back to the base Material class. 

I also noticed a cancellation of errors, that should be resolved where linearExpansionPercent was using Tc with a reference temperature in K -- this should be resolved.

---

## Checklist

### DIFF
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -717,7 +717,7 @@ class SimpleSolid(Material):
 
     def __init__(self):
         Material.__init__(self)
-        self.p.refDens = self.density3(Tc=self.refTempK)
+        self.p.refDens = self.density3(Tk=self.refTempK)
 
     def linearExpansionPercent(self, Tk: float = None, Tc: float = None) -> float:
         """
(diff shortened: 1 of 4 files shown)


jakehader:
Hi @sombrereau - Can you add any testing for this to cover the issue with the subclassing that you found?

sombrereau:
> 

@jakehader in principle this is tested by the unit testing here: https://github.com/terrapower/armi/blob/02f99a80548e3c1d0a1910a2ef09b9722fd40062/armi/materials/tests/test_materials.py#L464 

The precipitating error, was in application specific code.

sombrereau:
> Hi @sombrereau - Can you add any testing for this to cover the issue with the subclassing that you found?

I could do an issubclass test on FuelMaterial if you really want?

jakehader (approved):

PR #884, opened 2022-09-18
Adding Code Coverage to Utils

john-science:
## Description

Just adding a little code coverage.

In the process, my pylint tripped up on a bunch of unit tests, so I fixed that too.

---

## Checklist

- [X] This PR has only one purpose or idea.  (Well, one idea plus cleanup.)

### DIFF
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -572,42 +572,6 @@ def slantSplit(val, ratio, nodes, order="low first"):
     return X
 
 
-def runFunctionFromAllModules(funcName, *args, **kwargs):
-    r"""
-    Runs funcName on all modules of ARMI, if it exists.
-
-    Parameters
-    ----------
-    funcName : str
-        The function to run if it is found in a module.
-
-    \*args, \*\*kwargs : arguments to pass to func if it is found
-
-    Notes
-    -----
-    This imports all modules in ARMI, and if you have a script that isn't inside a
-    ``if __name__=='__main__'``, you will be in trouble.
-
-    This could also be useful for finding input consistency checkers for the GUI.
-
-    See Also
-    --------
-    armi.settings.addAllDefaultSettings : gets all the settings from all modules
-
-    """
-    for _modImporter, name, _ispkg in pkgutil.walk_packages(
-        path=armi_path, prefix=armi_name + "."
-    ):
-        try:
-            mod = importlib.import_module(name)
-            if funcName in dir(mod):  # there is a module.funcName. so call it.
-                func = getattr(mod, funcName)
-                func(*args, **kwargs)
-        except:
-            # just print traceback but don't throw an error.
-            traceback.print_exc()
-
-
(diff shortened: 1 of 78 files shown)


armi/utils/__init__.py
```diff
@@ -572,42 +572,6 @@ def slantSplit(val, ratio, nodes, order="low first"):
     return X
 
 
-def runFunctionFromAllModules(funcName, *args, **kwargs):
```
albeanth, line 575: This is just totally unused, yes? Thus the removal?

armi/utils/tests/test_units.py
```diff
@@ -73,3 +71,70 @@ def test_getTf(self):
 
         with self.assertRaisesRegex(ValueError, "Tc=0 and Tk=200"):
             units.getTf(Tc=0, Tk=200)
+
+    def test_convertPascalToPascal(self):
+        """This is a pass-through function"""
+        for val in [0.0, -99.141, 123, 3.14159, -2.51212e-12]:
+            self.assertEqual(val, units.convertPascalToPascal(val))
```
albeanth, line 79: So this test is fine, but I just peeked at `units.py::convertPascalToPascal`. What is this used for? It doesn't seem particularly useful... 

Doing a grep for `calculatePressure` also returns nothing...? I wonder if instead of adding a test, we can just remove `convertPascalToPascal` altogether with... 🤔

opotowsky (approved):
Other than addressing Tony's comment (which was an astute observation!) I think this PR looks good.

armi/utils/__init__.py
john-science: 1. Yes, totally unused, even downstream.
2. It is wildly dangerous and a "bad idea", so I want it out.
3. I had never noticed it before. I only saw it because I was looking for things in the `/utils/` folder to add code coverage to.

armi/utils/tests/test_units.py
john-science: Yeah, I definitely considered that. But the docstring convinced me there was a good use for this method.

I just did some digging, and I couldn't find a single downstream repo that uses this dubious method.

You're right; I like the idea of deleting it better.

john-science: Sadly, I _did_ find downstream use-cases of `PRESSURE_CONVERTERS`, which use `convertPascalToPascal`.

BUT, sure, I replaced `convertPascalToPascal` with an inline lambda, since all anyone needs is `PRESSURE_CONVERTERS` and not `convertPascalToPascal`.

Good call!

opotowsky: Wow!!! Nice work

albeanth (approved):

john-science:
@opotowsky @albeanth 
Thanks ya'll.  

I realize I do too much cleanup in my PRs.  At the time, my little pylint lines in this one seemed so minor.  But I will try to do better about splitting this PR into two in the future.

Thanks!

PR #893, opened 2022-09-21
Fix non uniform assembly treatment

mgjarrett:
## Description

When working on PR #879 , most of PR #827 was reverted because it was not working correctly with #879.

About halfway through development of #879, the implementation of non uniform assembly treatment was significantly overhauled, so that only the individual non-uniform assemblies are processed by the uniform mesh converter (instead of putting the full reactor through the uniform mesh conversion). With this updated implementation, we could have taken out the partial reversion of PR #827, but this wasn't immediately apparent.

It is now clear that we can bring back the parts of PR #827 that were reverted and still have PR #879 function as intended.

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_reactors.py
+++ b/armi/reactor/tests/test_reactors.py
@@ -832,6 +832,21 @@ class HexReactorTests(ReactorTests):
             coords = b.getPinCoordinates()
             self.assertGreater(len(coords), -1)
 
+    def test_nonUniformAssems(self):
+        o, r = loadTestReactor(
+            customSettings={"nonUniformAssemFlags": ["primary control"]}
+        )
+        a = o.r.core.getFirstAssembly(Flags.FUEL)
+        self.assertTrue(all(b.p.topIndex != 0 for b in a[1:]))
+        a = o.r.core.getFirstAssembly(Flags.PRIMARY)
+        self.assertTrue(all(b.p.topIndex == 0 for b in a))
+        originalHeights = [b.p.height for b in a]
+        differntMesh = [val + 2 for val in r.core.p.referenceBlockAxialMesh]
+        # wont change because nonUnfiform assem doesn't conform to reference mesh
+        a.setBlockMesh(differntMesh)
+        heights = [b.p.height for b in a]
+        self.assertEqual(originalHeights, heights)
+
     def test_applyThermalExpansion_CoreConstruct(self):
         """test Core::_applyThermalExpansion for core construction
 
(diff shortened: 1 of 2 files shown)


armi/reactor/tests/test_reactors.py
```diff
@@ -832,6 +832,21 @@ def test_pinCoordsAllBlocks(self):
             coords = b.getPinCoordinates()
             self.assertGreater(len(coords), -1)
 
+    def test_nonUniformAssems(self):
```
john-science, line 835: I love the test! Thanks so much.

But with all the activity on this feature being reverted then put in and then taken out and put in again, I am trying to figure out where the edge cases are in this logic. 

Can we add a more complex case, where there are two or three different items in the "nonUniformAssemsFlags" list?  It should be nearly copy/paste.  OR, if that's a hassle, I can throw in the extra test after this merges.  Thanks!

john-science:
@mgjarrett 

Is this non-uniform flag meant to ONLY affect the uniform mesh converter?

For instance, I see it also affects detailed axial expansion. So, is that _just_ a current limitation of detailed axial expansion?

albeanth:
> @mgjarrett
> 
> Is this non-uniform flag meant to ONLY affect the uniform mesh converter?
> 
> For instance, I see it also affects detailed axial expansion. So, is that _just_ a current limitation of detailed axial expansion?

Not sure if this has been resolved, so I'll try and provide some context as well. The non-uniform flag came into existence in #827 (though in that PR, it didn't do much and served more as a placeholder). The important

armi/reactor/tests/test_reactors.py
albeanth: A good place for this would be here
https://github.com/terrapower/armi/blob/908c92d86546d5c6e6e7df946cf1fdfa9515b964/armi/reactor/converters/tests/test_uniformMesh.py#L408

I think a good test would be to make sure that the assembly mesh gets changed to the `refAssem` mesh after `convert` and then after the call to `applyStateToOriginal`, the assembly mesh is back to its original mesh.

albeanth (approved):

mgjarrett:
Sorry for the late reply; I was out for a few days. @albeanth summed it up perfectly.

john-science:
Thanks @albeanth and @mgjarrett .

I guess my confusion arises from this explanation not being in this PR:

https://github.com/terrapower/armi/blob/f2516e29749e9b7bda588bb57533f7f57250b75b/armi/settings/fwSettings/globalSettings.py#L168-L178

I suppose the explanation is covered well enough in that description.

john-science (approved):

Issue #838, opened 2022-08-13
Settings validator related to detailed cycles input being raised incorrectly


keckler:
It appears that _anytime_ one uses the detailed cycles settings (see #637 ), a validator is raised:
![image](https://user-images.githubusercontent.com/21959600/184460794-fd60b581-0e93-4529-9a2c-83d365370199.png)

This is raised even if your input does not actually contain any of the mentioned offending settings.

This incorrect validator should be either fixed, or possible removed. I am inclined to believe that it might not be possible to fix... We'll see.

keckler:
Nah, I'll work on it today. Sounds like a good change of pace.

PR #894, opened 2022-09-22
Cycles validation

keckler:
## Description

This addresses #838 .

The `_checkForBothSimpleAndDetailedCyclesInputs()` method has been updated to work correctly. The behavior of the associated setting validator is now as expected, i.e. the validator will not be raised if the user does not enter any of the simple cycles settings.

The method is now explicitly tested as well. To do this, the method had to be moved from inside of the `_inspectSettings()` method (where it was technically not a method, but a nested function) to being an actual _method_ of the `Inspector` class. This is cleaner anyways, so a win-win.

---

## Checklist


john-science (approved):
Awesome. I even like that we're moving away from those untestable nested methods.  Thanks!

PR #898, opened 2022-09-24
Removing defunct settingsRules tools

john-science:
## Description

The file `settingsRules.py` has been pretty defunct for a long time. We have the settings validators now, which are much better. So this PR simply removes the last vestigial usage of `settingsRules.py` in the neutronics settings. I verified that these are not used downstream anywhere, so this is a pretty safe move.

---

## Checklist


armi/physics/neutronics/settings.py
```diff
@@ -342,251 +343,170 @@ def defineSettings():
...
-    else:
-        newValue = value
-
-    if value != newValue:
-        runLog.info("The `genXS` setting has been set to {}.".format(newValue))
-    return {name: newValue}
-
-
-@include_as_rule("existingFIXSRC")
-def newFixedSourceOption(cs, _name, value):
```
keckler, line 382: Did we miss this migration function when you revamped this file? Or did you mean to remove it without replacement?

armi/physics/neutronics/settings.py
```diff
@@ -342,251 +343,170 @@ def defineSettings():
...
-        newValue = NEUTRONGAMMA
-    elif value == "False":
-        newValue = NEUTRON
-
-    cs["globalFluxActive"] = newValue
-    runLog.info(
-        "The `globalFluxActive` setting has been set to {} based on deprecated "
-        "`gammaTransportActive`.".format(newValue)
+
+    def migrateDpa(name0):
```
keckler, line 457: My feeling is that this belongs in our internal neutronics plugin, since we're making reference to TWR and specific materials here.

keckler (approved):

armi/physics/neutronics/settings.py
john-science: As far as I can tell, it is not used anywhere downstream.  If I am mistaken, by all means, holler.  But I couldn't find any references to it anywhere.

armi/physics/neutronics/settings.py
john-science: I agree with that.

But we will have to copy/paste over the settingsValidator version, not the old settingsRules version.

Issue #897, opened 2022-09-23
Add third style option to SettingsWriter


opotowsky:
Right now, `SettingsWriter` has two style options: `short` and `full`. 

https://github.com/terrapower/armi/blob/908c92d86546d5c6e6e7df946cf1fdfa9515b964/armi/settings/settingsIO.py#L275-L283

This former strips all settings that have default values, and the latter adds every possible setting to the settings file. It would be nice to have a file written that doesn't strip the pre-defined settings even if they are in fact the default value. Default settings are not always backwards compatible, and it can add clarity to users to understand what a particular case is doing to have some settings listed.

john-science:
Sorry, can you clarify what you want that's different from `full`?

PR #904, opened 2022-09-27
Add writer style option to SettingsWriter

opotowsky:
~Opening a draft PR for a preliminary review before adding a fuller description/notes/tests since the content of this PR is subject to change a good bit. This is a first stab and I'd like some feedback first.~ Opening PR and adding tests. 

## Description

As described in Issue #897, we would like a third option that preserves the original settings file even if a setting defined in it is a default value. Currently, `short` removes all settings that have the default value, and `full` prints out every possible setting to the file. 

To accomplish that, I added a `medium` style option to `SettingsWriter` (which involved editing the handling of the strings of "acceptable" styles to constant strings as well as a helper function in the `caseSettings` module called `getSettingSetByUser` so the full list of user defined settings can be tracked.).

I propagated the `style` argument back from it's original location in `SettingsWriter` to the yaml-writing methods in `caseSettings` (`writeToYamlFile` & `writeToYamlFileStream`, and further back from there to the case/suite `clone`s and `writeInputs` (with the updated argument name of `writeStyle`). I also made the `settingsWriteStyle` flag available to the `modify` and `clone`-related CLI, so one can alter the writing style from the command line.

The testing was a little difficult. `writeToYamlFile` method and `SettingsWriter` class aren't explicitly tested, and neither is `clone`. I have the 3 writing style options tested, the new method (`getSettingsSetByUser`) in `caseSettings.py` is tested, and some attempts at testing the CLIs that were changed in this PR. 

Small PR-addendum:
There was a stale `coveragerc` file in `armi/resources`. I now have the coverage code in `case.py` pointing to the main/active version in the repo, using a helper function with an `shutil.copy` to accomplish the reason for having a copy of the original file in the first place. I have tested that this works with a case (and a case suite). 

---

## Checklist


armi/settings/caseSettings.py
```diff
@@ -368,13 +368,62 @@ def writeToYamlFile(self, fName, style="short"):
...
+            originalSettings = None
         with open(self.path, "w") as stream:
-            writer = self.writeToYamlStream(stream, style)
+            writer = self.writeToYamlStream(stream, style, originalSettings)
         return writer
 
-    def writeToYamlStream(self, stream, style="short"):
-        """Write settings in yaml format to an arbitrary stream."""
-        writer = settingsIO.SettingsWriter(self, style=style)
+    def getOriginalCaseSettings(self, fPath):
```
opotowsky, line 379: I decided to have this live here, and pass the resulting list of "old" settings names into the instantiation of the `SettingsWriter` object (this happens later on in `writeToYamlStream`). This is because the file path names are here and my choice was either to pass in this list or pass the file name and have this method there. The downside to having it here is there must be logic about the style choice in this class too, which might not be desirable. See line 371 above:

https://github.com/terrapower/armi/pull/904/files#diff-0dc0a268887c52884d0cc28261cb1c576277485a38433172ef5bfc4914549780R371-R374

Happy to change the name and name of the original settings names variable if someone has a better suggestion.

armi/settings/caseSettings.py
```diff
@@ -368,13 +368,62 @@ def writeToYamlFile(self, fName, style="short"):
...
+        can be tracked outside of a Settings Object
+
+        Parameters
+        ----------
+        fPath : str
+            The absolute file path of the settings file
+
+        Returns
+        -------
+        list(originalSettings.keys()) : list
```
opotowsky, line 391: I'm not sure how to handle documentation in this situation!

armi/settings/caseSettings.py
keckler: I might change the name of `originalSettings` and `getOriginalCaseSettings()` to something like `settingsSetByUser` and `getSettingsSetByUser()`.

I'm not super wild about that name either, so do whatever you think is best here. I just think `original` could be interpreted in a number of ways.

armi/settings/caseSettings.py
```diff
@@ -368,13 +368,62 @@ def writeToYamlFile(self, fName, style="short"):
...
+
+        # We do not want to load these as settings, but just grab the dictionary straight
+        # from the settings file to know which settings are user-defined
+        with open(fPath, "r") as stream:
+            yaml = YAML()
+            tree = yaml.load(stream)
+            originalSettings = tree[settingsIO.Roots.CUSTOM]
+        return list(originalSettings.keys())
+
+    def writeToYamlStream(self, stream, style="short", originalSettingsNames=None):
```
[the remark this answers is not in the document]
keckler: Or maybe not, if it is possible to run a case with a completely blank case settings file?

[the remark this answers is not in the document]
keckler: Well even if the case settings were blank, `getOriginalCaseSettings()` would still return an empty list.

armi/settings/settingsIO.py
```diff
@@ -268,13 +270,16 @@ class Styles:
         """Enumeration of valid output styles"""
 
         short = "short"
+        medium = "medium"
         full = "full"
```
keckler, line 274: I'm not really sure what the purpose of these three class variables is. My thought would be to delete them and just use the words `"short"`, `"medium"`, and `'long"` when they are needed. Less typing, no downside.

opotowsky: yea I was just coding within the existing architecture, but you're right that it does end up causing more typing. @john-science do you have an opinion on me deleting this class?

armi/settings/caseSettings.py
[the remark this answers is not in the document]
opotowsky: `short` never needs to know about `originalSettingsNames` since it is only writing non-defaults. I see what you mean though, which you're right, it could end up writing a blank file! I guess that was a risk before any of this. I can add a check to the writer to abort the writing if the `settingData` is empty, if you think that would be helpful.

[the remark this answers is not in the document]
keckler: Whoops, that was a typo before. I meant to say `style == "medium"`

opotowsky:
> This looks awesome so far. Thanks for doing this.
> 
> I did see that `writeToYamlFile()` is being used in `armi.cases.case` as well. I think it would be good to propagate the writing style to that usage instance as well. I think that should be possible with just a bit more scaffolding.

I'm happy to do that, but would want some buy-in from others beforehand. making changes and preserving the previous default behavior keeps me out of trouble ;-)

@Nebbychadnezzar @john-science @albeanth @jakehader @mgjarrett Anyone have a problem with me changing the default case writing and cloning in `case.py` to retain the user-defined settings even if they are default values?

My 2 pennies: I do not see an issue with it.

armi/settings/caseSettings.py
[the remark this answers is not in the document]
opotowsky: Oh! Yes we can't have the comparison happen if the list is None. Thanks for bringing that up.

If the (renamed) `settingsSetByUser` list defaults instead to an empty list then the comparison would happen without failure. Then it would give back an empty `settingsData` dictionary. Is that OK? Or should I add something to stop the writing?

keckler:
> I'm happy to do that, but would want some buy-in from others beforehand. making changes and preserving the previous default behavior keeps me out of trouble ;-)

Well I'm not saying to change the default behavior. I'm suggesting that, if the new command-line option `--style=medium` is specified, then also the `medium` style should be used when the YAML files are printed out from the `case` module.

Does that make sense?

armi/cli/modify.py
```diff
@@ -61,6 +61,13 @@ def addOptions(self):
             default=".",
             help="A root directory in which to search for settings files, e.g., armi/tests.",
         )
+        self.parser.add_argument(
+            "--style",
```
keckler, line 65: I'm wondering if we want a more descriptive name for this option? "style" could refer to a lot of things, and maybe we don't want to occupy such a generic name.

Maybe something like "--settingsDumpStyle"?

armi/settings/caseSettings.py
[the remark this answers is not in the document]
keckler: Yeah, sounds like a fine plan.

opotowsky:
> Well I'm not saying to change the default behavior. I'm suggesting that, if the new command-line option `--style=medium` is specified, then also the `medium` style should be used when the YAML files are printed out from the `case` module.
> 
> Does that make sense?

So you're just wanting the `settingsWriteStyle` argument to be available to the `clone` CLI, correct?

armi/settings/settingsIO.py
john-science: Perhaps, at least make them defined constant strings, so the spelling and capitalization is enforced:

https://github.com/terrapower/armi/blob/fe7f42f9ff838f5e3b8fb457ec5334d8ead05c1a/armi/physics/neutronics/settings.py#L29-L31

Passing around magic strings frequently causes trouble later. IMO

armi/cli/clone.py
```diff
@@ -28,24 +28,34 @@ class CloneArmiRunCommandBatch(EntryPoint):
     settingsArgument = "required"
 
     def addOptions(self):
-        for settingName in self.cs.keys():
-            # verbosity and branchVerbosity already have command line options in the default parser
-            # adding them again would result in an error from argparse.
-            if settingName not in ["verbosity", "branchVerbosity"]:
-                self.createOptionFromSetting(settingName)
```
opotowsky, line 35: For some reason I was getting a weird failure testing the CLI with this being first versus last.

john-science: I'm indifferent to the position.  Move it if you need.

Strange though.

opotowsky:
Closing this PR to split into 2-3 smaller ones

albeanth:
> Closing this PR to split into 2-3 smaller ones

🙌 Such a fan of this! Love it.

opotowsky:
> > Closing this PR to split into 2-3 smaller ones
> 
> 🙌 Such a fan of this! Love it. The recognition!

Thank you. I was a victim of scope creep and my own 'ooh this is tiny it can get snuck in'. Do not recommend!

PR #909, opened 2022-09-29
Use b.core instead of b.r.core.

mgjarrett:
## Description

Blocks have an `r` property and a core property. Both functions just look up the chain of ancestors of the block to find the `core` and `reactor`. It's unnecessary to go all the way up to the `reactor` just to then grab the `core` off of it.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -168,16 +168,23 @@ class Block(composites.Composite):
     @property
     def r(self):
         """
-        A block should only have a reactor through a parent assembly.
+        Look through the ancestors of the Block to find a Reactor, and return it.
 
+        Notes
+        -----
+        Typical hierarchy: Reactor <- Core <- Assembly <- Block
+        A block should only have a reactor through a parent assembly.
         It may make sense to try to factor out usage of ``b.r``.
 
-        For now, this is presumptive of the structure of the composite hierarchy; i.e.
-        the parent of a CORE must be the reactor. Fortunately, we probably don't
-        ultimately want to return the reactor in the first place. Rather, we probably want the core
-        anyways, since practically all `b.r` calls are historically `b.r.core`. It may be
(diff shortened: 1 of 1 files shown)


mgjarrett:
Basically acting on the note in this docstring:

https://github.com/terrapower/armi/blob/646f23da6ce77fb410dffd202dbddb0fba57da63/armi/reactor/blocks.py#L173-L181

Can't completely remove `b.r` yet because it's used to get active nuclides from reactor blueprints here:

https://github.com/terrapower/armi/blob/646f23da6ce77fb410dffd202dbddb0fba57da63/armi/reactor/blocks.py#L737

john-science:
I love it!

Can we remove the comment you link on line ~173 now?

mgjarrett:
> Can we remove the comment you link on line ~173 now?

I would like to remove that entire function, but we still have one case of `b.r.blueprints` so we can't remove this property altogether. I think we can trim some of the docstring and make it better. I'll push a commit with a new docstring.

keckler (approved):
I think these are good changes.

With regards to your comment:
> Can't completely remove b.r yet because it's used to get active nuclides from reactor blueprints here:

Is there a reason why we want to remove `b.r` completely? I agree that, in the cases you changed `b.r.core` -> `b.core`, it makes sense because it avoids unnecessary computation. But I think that `b.r` could still be useful in the API.

Plus, if we're trying to avoid changes to the API, I would view that as an unnecessary change to the API.

Just a thought.

mgjarrett:
I agree that removing `b.r` would be an unnecessary change to the API. The docstring had this comment:

```
It may make sense to try to factor out the usage of ``b.r``
```

So that's why I thought it might be desirable to remove it. But as long as it's being used, and it's not hurting anything, it makes more sense to leave it in for continuity.

john-science (approved):

Issue #813, opened 2022-08-04
Generalization reaction rate calcutions 


jakehader:
Based on the changes proposed in https://github.com/terrapower/armi/pull/810#discussion_r937171187_, the following suggestions were made to further improve the reaction rate calculation for blocks/components but we're not yet implemented.

1. Let's change the `xsType` input to be `microSuffix`, so that this encapsulates both the XS Type (i.e., A, B, C) as well as the burnup group (i.e., A, B, C) so that we can actually get reaction rates based on burnup-dependent cross sections.

To do this, we can use `self.getMicroSuffix()`, where on the block level this is implemented as is, but on a component this is implemented as reaching up to its parent and getting it from the block-level.

2. It would be advantageous for us to standardize the dictionary of reaction rates that can be obtained from the nuclide data library - `{"nG": 0, "nF": 0, "n2n": 0, "nA": 0, "nP": 0, "n3n": 0}`. This is a bit of a refactor, but having this mapping of these strings to the following xsLibrary micros would be good:

```
        ("nG", libNuc.micros.nGamma),
        ("nF", libNuc.micros.fission),
        ("n2n", libNuc.micros.n2n),
        ("nA", libNuc.micros.nalph),
        ("nP", libNuc.micros.np),
```

Just grepping around the code base, I see this mapping occurring in `blocks.py/getReactionRates`, `composites.py/getReactionRates`, `component.py/getReactionRateDict`,  `isotopicsDepletionInterface.py/makeXsecTable`, and `crossSectionTable.py/CrossSectionTable`. 

It might make sense to have the strings for `nG`, `nF`, `n2n`, `nA`, `nP`, etc. mapped back to `xsCollections.py` module-level constants that already exist:

```
NGAMMA = "nGamma"      # radiative capture
NAPLHA = "nalph"       # (n, alpha)
NP = "np"              # (n, proton)
ND = "nd"              # (n, deuteron)
NT = "nt"              # (n, triton)
FISSION_XS = "fission" # (n, fission)
N2N_XS = "n2n"         # (n,2n)
NUSIGF = "nuSigF"      
NU = "neutronsPerFission"
```

keckler:
I'm gonna work on this. Thinking about it more, this is really a bug that is leading to inaccuracies in my calculations against data.

keckler:
#911 addresses Point (1) above. 

Point (2) is for another PR. Not sure I'm gonna do that one at this point though.

john-science:
[...]

PR #911, opened 2022-09-29
Fix burnup-dependent reaction rates

keckler:
## Description

The primary purpose of this commit is to fix the `getReactionRates()` methods
on the `Block` and `Component` classes to account for burnup-dependent cross-sections, as was brought up in #813 .

This is completed primarily by changing the underlying `getReactionRateDict()`
method to accept the entire library suffix instead of just the first letter of
the suffix (i.e. "AA" or "AB" instead of just "A").

In the process, it was seen that a refactor could be achieved where the two classes could share the method of the parent instead of implementing their own, since they effectively do the same thing. This was accomplished by generalizing the calls in the method and moving that method into the `Composite` class.

Unit tests have also been updated.

I looked around, and do not believe that the change in the `getReactionRateDict()` signature should impact anybody. And even if it does, this is a necessary change because this was previously a bug.

In terms of impact of this bug-fix on physics results, I only know of two potential impacts, both on internal projects. I have already reached out to both of those projects individually.

---

## Checklist


john-science (approved):
As we discussed, I think `getReactionRateDict()` started in kind of a strange place before you got here. But that doesn't change the value of this PR.

Thanks!

keckler:
Alright @john-science , I moved `getReactionRateDict()` into `composites` where it belongs.

PR #912, opened 2022-09-30
Making Case.run more testable

john-science:
## Description

While looking at a possible duplicate `coveragerc` file, I noticed that the important `Case.run()` method had no code coverage because it was so tremendously long it was untestable. 

So I split the really long method out into something with helper methods, so it could be at least partially testable.

---

## Checklist

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -459,7 +459,7 @@ class Case:
         statsStream = six.StringIO()
         summary = pstats.Stats(profiler, stream=statsStream).sort_stats("cumulative")
         summary.print_stats()
-        if context.MPI_SIZE > 0:
+        if context.MPI_SIZE > 0 and context.MPI_COMM is not None:
             allStats = context.MPI_COMM.gather(statsStream.getvalue(), root=0)
             if context.MPI_RANK == 0:
                 for rank, statsString in enumerate(allStats):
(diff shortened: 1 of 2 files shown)


armi/cases/case.py
```diff
@@ -341,14 +342,41 @@ def run(self):
...
+            else:
+                o.operate()
+
+        # if in the settings, report the coverage and profiling
+        Case._endCoverage(cov)
+        Case._endProfiling(profiler)
+
+    def _startCoverage(self):
+        """Helper to the Case.run(): spin up the code coverage tooling,
+        iff the CaseSettings file says to.
```
opotowsky, line 373: typo: iff

opotowsky: Or do you mean if and only if? :-)

This applies to all the other new cov/profiling docstrings too

opotowsky (commented):
This looks straightforward. No changes, just breaking of chunks of code and testing them. Good work! 

I'm confused as to why the coverage is failing and the previous runs had a  drop of 5%....do you know what's going on there?

armi/cases/case.py
john-science: iff = "if and only if"

But if that's not clear, I can change it.

opotowsky: It's fine!

opotowsky (approved):
Oh! No testing coverage=True then!

PR #914, opened 2022-10-01
Fix issue where case was mistakenly making absolute paths relative.

### LINKED ISSUE #913: Absolute file paths being dumped as relative
Basically in a case suite the locations of input interface files are being dumped back out as relative file paths when they were originally specified as absolute. If the file path is written in absolute terms, it should stay that way. There is a bug making it so that sometimes the absolute path gets truncated to just the local, relative path.

The bug is in this line specifically:
https://github.com/terrapower/armi/blob/main/armi/cases/case.py#L892

The fix is just to change that line to:
```
                            newFiles.append(path)
                            continue
```

I can do this, I'm just too lazy at this point in the day to write the unit tests to prove it.

For another day.

ntouran:
## Description
Closes #913

Though should have a test.

---

## Checklist

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -889,7 +889,8 @@ def copyInterfaceInputs(
                     try:
                         if path.is_absolute() and path.exists() and path.is_file():
                             # Path is absolute, no settings modification or filecopy needed
-                            pass
+                            newFiles.append(path)
+                            continue
                     except OSError:
                         pass
                 # Attempt to construct an absolute file path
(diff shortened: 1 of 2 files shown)


john-science:
@ntouran   Could you add this test to the end of `armi/cases/tests/test_cases.py`?

```python
    def test_copyInterfaceInputs_absPath(self):
        testSetting = "shuffleLogic"
        cs = settings.Settings(ARMI_RUN_PATH)
        shuffleFile = cs[testSetting]
        absFile = os.path.dirname(os.path.abspath(ARMI_RUN_PATH))
        absFile = str(os.path.join(absFile, os.path.basename(shuffleFile)))
        cs = cs.modified(newSettings={testSetting: absFile})

        with directoryChangers.TemporaryDirectoryChanger() as newDir:
            newSettings = cases.case.copyInterfaceInputs(
                cs, destination=newDir.destination
            )
            self.assertEqual(str(newSettings[testSetting]), absFile)
```

I'd do it myself, but this branch is in your fork repo.

Thanks!

PR #919, opened 2022-10-03
Uniform mesh update for reaction rate calculations of non-uniform assemblies

jakehader:
## Description

Update the uniform mesh converter to handle the reaction rate calculations and to remove the overly restrictive power conversion check. This allows the following block-level parameters for non uniform assemblies: `["rateCap", "rateFis", "rateProdN2n", "rateProdFis", "rateAbs"]`

---

## Checklist

### DIFF
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from framework.armi.physics.neutronics.globalFlux.globalFluxInterface import calcReactionRates
 
 """
 Converts reactor with arbitrary axial meshing (e.g. multiple assemblies with different
@@ -63,11 +64,11 @@ import numpy
 
 import armi
 from armi import runLog
-from armi import settings
 from armi.utils.mathematics import average1DWithinTolerance
 from armi.utils import iterables
 from armi.utils import plotting
 from armi.reactor import grids
+from armi.reactor.reactors import Core
 from armi.reactor.flags import Flags
 from armi.reactor.converters.geometryConverters import GeometryConverter
 from armi.reactor import parameters
@@ -99,6 +100,7 @@ class UniformMeshGeometryConverter(GeometryConverter):
         self._uniformMesh = None
         self.reactorParamNames = []
         self.blockParamNames = []
+        self.calcReactionRates = False
 
         # These dictionaries represent back-up data from the source reactor
         # that can be recovered if the data is not being brought back from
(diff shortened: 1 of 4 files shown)


armi/reactor/converters/uniformMesh.py
```diff
@@ -732,6 +756,30 @@ def _mapStateFromReactorToOther(
...
+            # Checks if the block has a multi-group flux defined and if it
+            # does not then this will skip the reaction rate calculation. This
+            # is captured by the TypeError, due to a `NoneType` divide by float
+            # error.
+            try:
+                b.getMgFlux()
+            except TypeError:
+                continue
+
+            print(b.getMgFlux())
```
albeanth: Did you mean to keep these two `print` statements?

jakehader: Should be removed now.

albeanth: https://github.com/terrapower/armi/pull/919/commits/d30112ac4f8dc981e37d179e7fcb3f959764ce5b

armi/reactor/converters/uniformMesh.py
```diff
@@ -537,6 +540,20 @@ def setAssemblyStateFromOverlaps(
...
+            if calcReactionRates:
+                core = sourceAssembly.getAncestor(lambda c: isinstance(c, Core))
+                if core is not None:
+                    UniformMeshGeometryConverter._calculateReactionRates(
+                        lib=core.lib, keff=core.p.keff, assem=destinationAssembly
+                    )
+                else:
+                    runLog.warning(
+                        f"Reaction rates requested for {destinationAssembly}, but no core object exists. This calculation "
+                        "will be skipped."
```
albeanth, line 554: Will we get a zillion of these warnings? If so, do we want to add a `single = True` ?

albeanth:
Just have one more suggestion to consider. Let's make sure that this PR doesn't break any existing uniform mesh capabilities with `detailedAxialExpansion: True`. This may be covered by unit tests, but I'm not sure off the top of my head.

armi/reactor/converters/uniformMesh.py
jakehader: We can add single = True for sure!

jakehader:
> Just have one more suggestion to consider. Let's make sure that this PR doesn't break any existing uniform mesh capabilities with `detailedAxialExpansion: True`. This may be covered by unit tests, but I'm not sure off the top of my head.

Yes, this will work with an without detailedAxialExpansion set to True:

image

The only quirk here is that if `detailedAxialExpansion: True` _and_ we have `cs["nonUniformAssemFlags"]` then it will favor converting the reactor with `cs["nonUniformAssemFlags"]` in the following `def convert` method. I will add a settings validation that checks for this for the user.

image

albeanth:
Awesome. The concern I had was in `UniformMeshGeometryConverter::convert`. The settings validator should cover that concern nicely.

PR #922, opened 2022-10-05
Update the EntryPoint class to provide user feedback on required positional arguments

jakehader:
## Description

Update the entry point class to provide the settings file as an required position argument when a user requests the help menu. Additionally, add a try/except around adding a new argument from case settings to handle duplicate entries more robustly rather than requiring exclusion of settings like `verbosity` and `branchVerbosity`.

@opotowsky update:

In getting the unit tests to work, a few more changes were made.
1. the `reloadDBName` setting was removed from `armiRun.yaml`. This ended up not being necessary after other fixes were made, but it is only used for 3 tests (in `bookkeeping/db`). 
2. I altered the above-mentioned test setup there to have custom settings to handle the deletion. _Side note: ARMI should be able to handle settings that are files that do not yet exist, but it can't right now._ Seemed ideal to keep the base case for all unit tests simple. 
3. the CLI for `checkInputs` needed some love. There are minor code updates, deletions, and some small updates to the tests.

---

## Checklist


opotowsky:
For the life of me, I couldn't invoke any of the clone commands in `test_runEntryPoint.py`

This update allows me to. 

THANK YOU!!!!!!!!!!!!!!!!

opotowsky:
I'll push up some unit test fixes to this branch

armi/cli/tests/test_runEntryPoint.py
```diff
@@ -32,38 +33,69 @@
...
+                    ),
+                )
 
 
 class TestCheckInputEntryPoint(unittest.TestCase):
     def test_checkInputEntryPointBasics(self):
         ci = CheckInputEntryPoint()
         ci.addOptions()
-        ci.parse_args(["/path/to/fake.yaml"])
+        ci.parse_args([ARMI_RUN_PATH])
```
opotowsky, line 75: In some failing test cases, there just needed to be a real file path provided

armi/cli/tests/test_runEntryPoint.py
```diff
@@ -32,38 +33,69 @@
...
 
         with mockRunLogs.BufferLog() as mock:
             self.assertEqual("", mock._outputStream)
 
             ci.invoke()
 
-            self.assertIn("/path/to/fake.yaml", mock._outputStream)
+            # self.assertIn(ARMI_RUN_PATH, mock._outputStream)
```
opotowsky, line 90: But here is where I realized all of the optional patterns were being processed as the default `["*.yaml"]`

This is the output stream:
```
[info] Finding potential settings files matching **['*.yaml']**.
[info] Checking for valid settings files.
[impt] ------  -----------  --------------------------
case    can start    input is self consistent
------  -----------  --------------------------
------  -----------  --------------------------
```

armi/cli/tests/test_runEntryPoint.py
```diff
@@ -32,38 +33,69 @@
...
             self.assertIn("input is self consistent", mock._outputStream)
 
 
 class TestCloneArmiRunCommandBatch(unittest.TestCase):
     def test_cloneArmiRunCommandBatchBasics(self):
         ca = CloneArmiRunCommandBatch()
         ca.addOptions()
-        ca.parse_args(["--additional-files", "test"])
+        ca.parse_args([ARMI_RUN_PATH, "--additional-files", "test"])
```
opotowsky, line 98: Additionally, the `patterns` arg needed to be first and not last, which is not my experience on the command line

armi/cli/clone.py
```diff
@@ -29,10 +29,8 @@ class CloneArmiRunCommandBatch(EntryPoint):
 
     def addOptions(self):
         for settingName in self.cs.keys():
-            # verbosity and branchVerbosity already have command line options in the default parser
-            # adding them again would result in an error from argparse.
-            if settingName not in ["verbosity", "branchVerbosity"]:
```
john-science, line 34: I love that we're ditching this "special case".   

I expected to see this moved down into the `createOptionFromSetting()` method, but it's not there.  Where did that go?

Don't get me wrong, I think the code looks much cleaner this way.

jakehader: This is captured by the try/except in that method. Basically the reason we skipped this is because it was "already added" but catching the ArgumentError when trying to add_argument will just skip an optional argument (or a setting in this case) that already exists. The parser doesn't have any public attributes to see what is already an argument without iterating through it's ._actions attribute so I thought it was cleaner to just catch an exception versus trying to write some tooling to check if the action was there already. I am concerned that if they rename that attribute in the future then it could break our stuff. Let me know if want me to explicitly search down that actions list though.

armi/cli/checkInputs.py
```diff
@@ -140,4 +139,4 @@ def invoke(self):
         )
 
         if any(t[2] != "PASSED" or t[3] != "PASSED" for t in table):
-            sys.exit(-1)
+            runLog.warning(f"Check did not pass with flying colors")
```
opotowsky, line 142: @john-science what do you think about this? I really don't think the fail-fail is necessary. This also allows me to address the issues with `armiRun.yaml` separately.

@keckler @albeanth do you see any downstream issues with not exiting? This is only the CLI that does check-inputs, not the code itself

armi/cli/clone.py
```diff
@@ -28,17 +28,16 @@ class CloneArmiRunCommandBatch(EntryPoint):
...
         self.parser.add_argument(
             "--additional-files",
             nargs="*",
             default=[],
             help="Additional files from the source directory to copy into the target directory",
         )
+        # somehow running `armi clone-batch -h` on the command line requires this to
+        # not be first?
+        for settingName in self.cs.keys():
+            self.createOptionFromSetting(settingName, suppressHelp=True)
```
opotowsky, line 40: anyone have any clue why this works?

armi/cli/checkInputs.py
keckler: I don't have any issue. I've never used this feature 😄

john-science: 1. I guess the question is: do we want people to be able to use this "check inputs" as a safety check? Like, to stop a run?
2. What about making it a `runLog.error()`?  That won't exit the program, but it make it very visible.

armi/cli/clone.py
john-science: I can reproduce that error:

```
  File "c:\users\jstilley\codes\repos\armi\armi\__main__.py",
    code = ArmiCLI().run()
...
    assert ' '.join(opt_parts) == opt_usage
AssertionError
```

I can verify that the code runs ALL the way through this method without any errors, and then fails on `parse_args` in `entryPoint.py`.

My take-away from this is that the `nargs="*"`  is problematic.  You can see it [here](https://docs.python.org/3/library/argparse.html#nargs), and they don't make any warning about using the glob `*`, but that seems like the issue to me.

armi/cli/checkInputs.py
opotowsky: I am leaning towards `runLog.error`. 

When someone is programatically using `checkInputs`, they are looking for whether or not there are any issues tallied:
``` python
return not any(inspectorIssues)
```

When someone is using it from the command line, a table prints out:
```
Finding potential settings files matching ['armi/tests/armiRun.yaml'].
Checking for valid settings files.
loaded armi/tests/armiRun.yaml
----------------------------------------------------------------  -------  -----------  --------------------------
                                                                         case     can start    input is self consistent

       ----------------------------------------------------------------  -------  -----------  --------------------------
       C:\Users\aopotowsky\codes\nala\framework\armi\tests\armiRun.yaml  armiRun  UNKNOWN      PASSED
       ----------------------------------------------------------------  -------  -----------  --------------------------
```
So I'm just not convinced there needs to be anything more because the following step would be user-decided on the command line

armi/cli/clone.py
opotowsky: Ah ok that is somewhat helpful. I guess if you have `n` optional args you can't have an additional optional arg defined after that, or it would get processed with the first group of optional args...???

armi/cli/checkInputs.py
john-science: Cool. As long as you have the workflow in your mind.

If the user just wants to visually side "pass" or "fail" messages is different than... I don't know, say a CI job that needs to fail if the check inputs fails.

I haven't used the tool much in the wild, so the decision is yours.

armi/cli/clone.py
john-science: Yeah, my guess is it's something like that.

It is either a bug in the `argparse` standard library, or just the logical result of saying "we will take ANY extra args" with `nargs="*"`.  Like, maybe that's just logically sound, but limiting.

armi/bookkeeping/db/tests/test_comparedb3.py
```diff
@@ -86,7 +86,9 @@ def test_diffResultsBasic(self):
     def test_compareDatabaseDuplicate(self):
         """end-to-end test of compareDatabases() on a photocopy database"""
         # build two super-simple H5 files for testing
-        o, r = test_reactors.loadTestReactor(TEST_ROOT)
+        o, r = test_reactors.loadTestReactor(
+            TEST_ROOT, customSettings={"reloadDBName": "reloadingDB.h5"}
```
john-science, line 90: Love it!

armi/cli/tests/test_runEntryPoint.py
```diff
@@ -32,38 +33,71 @@
...
+                        f"implementation."
+                    ),
+                )
 
 
 class TestCheckInputEntryPoint(unittest.TestCase):
     def test_checkInputEntryPointBasics(self):
         ci = CheckInputEntryPoint()
         ci.addOptions()
-        ci.parse_args(["/path/to/fake.yaml"])
```
john-science, line 43: What's with this change?

I'm not complaining or anything.  I just don't understand.

opotowsky: I'm fairly certain it failed with the fake path, let me go double check that

opotowsky: Nope I'm wrong

That must have been a mistake! I'll push up a change. I feel like we should only use legit files when we need legit files

john-science: Your call. I was seriously just asking.

opotowsky: It was unnecessary. Leftover from troubleshooting most likely, already switched!

armi/cli/tests/test_runEntryPoint.py
```diff
@@ -32,38 +33,71 @@
...
+            self.assertIn(ARMI_RUN_PATH, mock._outputStream)
             self.assertIn("input is self consistent", mock._outputStream)
 
 
 class TestCloneArmiRunCommandBatch(unittest.TestCase):
     def test_cloneArmiRunCommandBatchBasics(self):
         ca = CloneArmiRunCommandBatch()
         ca.addOptions()
-        ca.parse_args(["--additional-files", "test"])
+        ca.parse_args([ARMI_RUN_PATH, "--additional-files", "test"])
```
opotowsky, line 100: Alternatively, this one needs to be a real file path

armi/cli/tests/test_runEntryPoint.py
```diff
@@ -237,7 +271,7 @@ class TestRunEntryPoint(unittest.TestCase):
     def test_runEntryPointBasics(self):
         rep = RunEntryPoint()
         rep.addOptions()
-        rep.parse_args([])
+        rep.parse_args([ARMI_RUN_PATH])
```
opotowsky, line 274: Same here, needs to be real since the settingsArgument is required here too

opotowsky:
The coverage drop looks like a fluke to me....
image

john-science (approved):

john-science:
@opotowsky But I see you fixed the coverage issue! Awesome!

PR #923, opened 2022-10-05
Remove deprecated coveragerc file

opotowsky:
## Description

There was a stale copy of a coverage config file in `armi/resources`. It's existence was only dreamed up to prevent the issues with dot-files on a Windows cluster. I changed the coverage work in `case.py` to copy the main coverage config file to a file name without a dot instead. Voila! No more forgetting to maintain multiple files.

---

## Checklist


armi/cases/case.py
```diff
@@ -425,6 +426,33 @@ def _endCoverage(cov=None):
...
+        ----------
+        makeCopy : bool (optional)
+            Whether or not to copy the coverage config file to an alternate file path
+
+        Returns
+        -------
+        covFile : str
+            path of coveragerc file
+        """
+        covRcDir = os.path.abspath(os.path.join(context.ROOT, ".."))
```
opotowsky, line 443: Any more robust suggestion here?

opotowsky:
@john-science My history is a little mucked up from the interactive rebase. Working on fixing

jakehader:
> Hey @john-science What happened with CI? Is there a way to kick it back off without making, e.g., an empty commit?

I think making an empty / trivial commit is the only way that I can see to easily kick off the CI again for GitHub. There isn't an easy "rerun all" since we have different GitHub actions that get fired off in parallel. That is my recommendation, but maybe @john-science has a more elegant solution.

armi/cases/case.py
jakehader: I think you could also reuse the `PROJECT_ROOT` in the test below.

opotowsky: I like this. on it.

opotowsky:
The coverage drop would be expected, since I added coverage-related code that can't be unit tested.

armi/cases/case.py
```diff
@@ -425,6 +426,33 @@ def _endCoverage(cov=None):
...
+        if platform.system() == "Windows":
+            if makeCopy == True:
+                # Make a copy of the file without the dot in the name
+                shutil.copy(
+                    os.path.join(covRcDir, ".coveragerc"),
+                    os.path.join(covRcDir, "coveragerc"),
+                )
+            covFile = os.path.join(covRcDir, "coveragerc")
+        else:
+            covFile = os.path.join(covRcDir, ".coveragerc")
```
john-science, line 453: It looks to me like you could make this simpler:

```python
if platform...
    if makeCopy...
        # Make ...

covFile = os.path.join(covRcDir, ".coveragerc")
```

It just _looks_ like that `else` is unnecessary, and you could have the `covFile =` line once instead of twice.  Right?

PR #924, opened 2022-10-05
Add writer style option to SettingsWriter pt1

opotowsky:
NOTE: the updates to the original PR on this topic requested by @keckler will be in a follow-up PR, namely making this argument available to `clone` / `writeInputs`.

## Description

As described in Issue #897, we would like a third option that preserves the original settings file even if a setting defined in it is a default value. Currently, `short` removes all settings that have the default value, and `full` prints out every possible setting to the file.

To accomplish that, I added a `medium` style option to `SettingsWriter` (which involved editing the handling of the strings of "acceptable" styles to constant strings as well as a helper function in the `caseSettings` module called `getSettingSetByUser` so the full list of user defined settings can be tracked.).

I made the `settingsWriteStyle` flag available to the `modify` EntryPoint, so one can alter the writing style from the command line.

The testing was a little difficult. `writeToYamlFile` method and `SettingsWriter` class aren't explicitly tested. I have the 3 writing style options tested, the new method (`getSettingsSetByUser`) in `caseSettings.py` is tested, and a sad attempt at adding this to the `modify` CLI testing. 

---

## Checklist

### DIFF
--- a/armi/settings/caseSettings.py
+++ b/armi/settings/caseSettings.py
@@ -357,7 +357,7 @@ class Settings:
 
         Notes
         -----
-        This resets the current CS's path to the newly written path.
+        This resets the current CS's path to the newly written absolute path.
 
         Parameters
         ----------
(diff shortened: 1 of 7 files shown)


armi/settings/caseSettings.py
```diff
@@ -368,13 +368,63 @@ def writeToYamlFile(self, fName, style="short"):
...
+
+        Parameters
+        ----------
+        stream : file object
+            Write-only file stream
+
+        style : str
+            Writing style for settings file. Can be short, medium, or full.
+
+        settingsSetByUser : list
```
keckler, line 417: Specify as optional.

armi/settings/caseSettings.py
```diff
@@ -368,13 +368,63 @@ def writeToYamlFile(self, fName, style="short"):
...
+        ----------
+        fPath : str
+            The absolute file path of the settings file
+
+        Returns
+        -------
+        userSettingsNames : list
+            The settings names read in from a yaml settings file
+        """
+        from ruamel.yaml import YAML
```
keckler, line 394: I am assuming there is some reason that this cannot be at the module level? If not, we should put it up there.

opotowsky: Yea there's no reason other than me not cleaning things up properly -- done!

armi/settings/settingsIO.py
```diff
@@ -320,7 +324,14 @@ def _getSettingDataToWrite(self):
         for _settingName, settingObject in iter(
```
jakehader, line 324: Recommend removing the `_` from `_settingName` now that it being used.

jakehader (changes requested):

jakehader (approved):

john-science (approved):
I love the test coverage!

PR #929, opened 2022-10-06
Restore the db `_fullPath` upon close()

keckler:
## Description

When a db open in write-mode is closed, it is moved out of the `FAST_PATH` back into the working directory. However, it's internal metadata `database._fullPath` was not updated to reflect this new path. 

This PR just updates the `_fullPath` attribute to be consistent.

This helps with keeping track of the DB during a run.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -705,7 +705,8 @@ class Database3:
 
         if self._permission == "w":
             # move out of the FAST_PATH and into the working directory
-            shutil.move(self._fullPath, self._fileName)
+            newPath = shutil.move(self._fullPath, self._fileName)
+            self._fullPath = os.path.abspath(newPath)
 
     def splitDatabase(
         self, keepTimeSteps: Sequence[Tuple[int, int]], label: str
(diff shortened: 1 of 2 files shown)


john-science:
I feel like this should break a lot of things.  Don't many people downstream of ARMI expect the output DB to be in a certain place after a run?

Have you tested this with our major downstream applications, to make sure it doesn't break any of the popular simulation workflows?

I ask because (a) that would be an important thing to know, and (b) this PR seems more like a "nice to have" than a "need".

keckler:
I'm not moving where the output DB is. It has always been moved out of the FAST_PATH and back to the working directory.

This did not change.

I simply updated the internal tracker to reflect this change. Before, the `database._fullPath` attribute was left in a state saying that the file was in the `FAST_PATH` even after it had been moved out. That was a bug. So I fixed that.

Further, this doesn't even happen until the database is closed, which in a normal run only happens in `interactEOL()` of the `database` interface, after which no further analysis is done during a typical run.

So I do not expect this to impact anybody's run negatively. The only reason I even found this bug was because I added a `UserPlugin` to my run which takes place _after_ the `database` interface at EOL. (It does not break my run when the `UserPlugin` is disabled either)

Does that address your concerns? Or if no, maybe we can talk offline.

john-science:
@keckler  Oh, sorry, that was a misread.  I thought you _added_ the `shutil.move()` line, and was immediately SO concerned I stopped reading and didn't notice that all you did was _change_ that line to include the return value.

Carry on.

john-science (approved):

PR #932, opened 2022-10-10
Add writer style option to SettingsWriter pt2 : Propagate to additional methods

opotowsky:
## Description

This propagates the writing style argument to `clone` and `writeInputs` (as well as the related CLI for `clone`). 

Sorry about the history. I was working on this branch in parallel to other branches. I'll clean it up in the squash n merge. 

---

## Checklist


armi/settings/caseSettings.py
```diff
@@ -367,10 +367,16 @@ def writeToYamlFile(self, fName, style="short"):
         style : str (optional)
             the method of output to be used when creating the file for the current
             state of settings
+        fromFile : str (optional)
+            if the source file and destination file are different (i.e. for cloning),
+            then this arg is used
         """
         self.path = pathTools.armiAbsPath(fName)
+        getSettingsPath = (
```
keckler, line 375: I would move this inside of the `if style == "medium" that takes place right after, since this is only used in there.

armi/settings/caseSettings.py
```diff
@@ -367,10 +367,16 @@ def writeToYamlFile(self, fName, style="short"):
         style : str (optional)
             the method of output to be used when creating the file for the current
             state of settings
+        fromFile : str (optional)
+            if the source file and destination file are different (i.e. for cloning),
```
keckler, line 371: AND if the writing style is "medium", right?

armi/cases/suite.py
```diff
@@ -266,17 +270,20 @@ def compare(
 
         return nIssues
 
-    def writeInputs(self):
+    def writeInputs(self, writeStyle="short"):
         """
         Write inputs for all cases in the suite.
 
+        writeStyle : str (optional)
+            Writing style for which settings get written back to the settings files.
```
keckler, line 278: Maybe want to add the options.

armi/cases/tests/test_cases.py
```diff
@@ -212,6 +212,41 @@ def test_run(self):
...
+        with directoryChangers.TemporaryDirectoryChanger():
+            cs = settings.Settings(ARMI_RUN_PATH)
+            setMasterCs(cs)
+            case = cases.Case(cs)
+            case.clone(writeStyle="medium")
+            clonedYaml = "armiRun.yaml"
+            self.assertTrue(os.path.exists(clonedYaml))
+            # validate a default value was removed
+            txt = open(clonedYaml, "r").read()
+            self.assertIn("availabilityFactor", txt)
```
keckler, line 248: I would add a comment in here as to why we would expect "availabilityFactor" to be in the written CS (i.e. because it is in the original CS but at it's default value, so it won't be in the short dump, but it will be in the medium dump)

Without seeing the original CS, it is not so obvious as to what is going on here.

keckler (approved):
Looks good, glad to have this new feature! Thanks for your work on it!

PR #936, opened 2022-10-11
Allow case coverage to use user-supplied coverage config file

opotowsky:
## Description

Someone might want to get code coverage on cases running with ARMI + plugins, in which case they would need to supply their own coverage configuration file. 

Does this look like an OK idea?

---

## Checklist

### DIFF
--- a/armi/settings/fwSettings/globalSettings.py
+++ b/armi/settings/fwSettings/globalSettings.py
@@ -230,7 +230,7 @@ def defineSettings() -> List[setting.Setting]:
         ),
         setting.Setting(
             CONF_COVERAGE_CONFIG_FILE,
-            default="",  # should default be empty string, or the armi .coveragerc file?
+            default="",
             label="File to Define Coverage Configuration",
             description="User-defined coverage configuration file",
         ),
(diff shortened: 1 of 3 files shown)


armi/settings/fwSettings/globalSettings.py
```diff
@@ -227,6 +228,12 @@ def defineSettings() -> List[setting.Setting]:
             "of code that execute during a run",
             isEnvironment=True,
         ),
+        setting.Setting(
+            CONF_COVERAGE_CONFIG_FILE,
+            default="",  # should default be empty string, or the armi .coveragerc file?
```
opotowsky, line 233: Commenting here so I remember to delete the comment. Also this comment on a comment is a question. What do you think, reviewer?

john-science: I guess it depends on how the default is used.

I kind of like the idea of having the ARMI `.coveragerc` as the default. But.. I think we should leave it as the empty string.  Because the ARMI `.coveragerc` file path would be defined as that stupid global variable. And these settings should be independent of the current state of ARMI global context variables.

john-science (approved):
Cool!

Maybe clean up your comment/note thing, but it looks good to me otherwise.

armi/settings/fwSettings/globalSettings.py
opotowsky: Good point. I will clean this up now!

Issue #930, opened 2022-10-10
Reduce or Eliminate getMasterCs and setMasterCs


john-science:
The `getMasterCs()` and `setMasterCs()` methods are extremely problematic. First off, obviously, they prevent ARMI from being able to do multiple things at once. We can not do the following currently, because they mean over-writing the "master cs":

* read two ARMI DBs at one time
* have two ARMI Reactors in memory at once
* All unit tests must share the same ARMI `App()`
* any `Block`s in memory must reference the same settings file
* All interfaces have access to their own settings object (`self.cs`), but can also reference the "master cs"

Here are all locations of usage:

### getMasterCs

* [x] `armi/bookkeeping/report/html.py`:    `cs = settings.getMasterCs()`
* [ ] `armi/interfaces.py`:        `self.cs = settings.getMasterCs() if cs is None else cs`
* [ ] `armi/interfaces.py`:        `self.cs = cs or settings.getMasterCs()`
* [ ] `armi/interfaces.py`:        `self.cs = settings.getMasterCs()`
* [ ] `armi/operators/operatorMPI.py`:        `cs = settings.getMasterCs()`
* [x] `armi/reactor/blocks.py`:        `if len(xsType) == 2 and len(settings.getMasterCs()["buGroups"]) == 1:`
* [x] `armi/reactor/blocks.py`:        `burnupPeakingFactor = settings.getMasterCs()["burnupPeakingFactor"]`
* [x] `armi/reactor/blocks.py`:        `if settings.getMasterCs()["gridPlateDpaXsSet"] and self.hasFlags(`
* [x] `armi/reactor/blocks.py`:            `dpaXsSetName = settings.getMasterCs()["gridPlateDpaXsSet"]`
* [x] `armi/reactor/blocks.py`:            `dpaXsSetName = settings.getMasterCs()["dpaXsSet"]`
* [x] `armi/reactor/reactors.py`:        `self.processLoading(settings.getMasterCs())`
* [x] `armi/reactor/reactors.py`:        `self.buildZones(settings.getMasterCs())`
* [x] `armi/reactor/reactors.py`:            `cs or settings.getMasterCs(), name=assemType`
* [ ] `armi/operators/tests/test_operators.py`:            `self.cs = settings.getMasterCs()`

If we can remove all the instances of `getMasterCs()`, we will be able to just delete all `setMasterCs()` for free.

jakehader:
[...]

PR #937, opened 2022-10-12
Removing usages of get/setMasterCs in getDpaXs

john-science:
## Description

As part of #930, this PR removes all 3 usages of `getMasterCs()` and `setMasterCs()` from `Block.getDpaXs()`.  This method is too specific to be on `Block` anyway, but if you move it to the `GlobalFluxInterface` (the only place it is every called), it becomes much simpler.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -307,6 +307,7 @@ class GlobalFluxOptions(executers.ExecutionOptions):
         self.aclpDoseLimit = None
         self.loadPadElevation = None
         self.loadPadLength = None
+        self.cs = None
 
         self._geomType: geometry.GeomType
         self.symmetry: str
@@ -600,6 +601,36 @@ class GlobalFluxResultMapper(interfaces.OutputReader):
                 "DPA cross section set {} does not exist".format(dpaXsSetName)
             )
 
+    def getBurnupPeakingFactor(self, b: Block):
+        """
+        Get the radial peaking factor to be applied to burnup and DPA for a Block
+
+        This may be informed by previous runs which used
+        detailed pin reconstruction and rotation. In that case,
+        it should be set on the cs setting ``burnupPeakingFactor``.
+
+        Otherwise, it just takes the current flux peaking, which
+        is typically conservatively high.
+
+        Parameters
+        ----------
+        b: Block
+            The block we want the peaking factor for
+
+        Returns
+        -------
+        burnupPeakingFactor : float
+            The peak/avg factor for burnup and DPA.
+        """
+        burnupPeakingFactor = self.cs["burnupPeakingFactor"]
+        if not burnupPeakingFactor and b.p.fluxPeak:
+            burnupPeakingFactor = b.p.fluxPeak / b.p.flux
+        elif not burnupPeakingFactor:
+            # no peak available. Finite difference model?
+            burnupPeakingFactor = 1.0
+
+        return burnupPeakingFactor
+
     def updateDpaRate(self, blockList=None):
         """
         Update state parameters that can be known right after the flux is computed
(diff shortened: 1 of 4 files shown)


armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -573,22 +576,44 @@ def _updateDerivedParams(self):
...
         updateFluenceAndDpa : uses values computed here to update cumulative dpa
-
         """
         if blockList is None:
             blockList = self.r.core.getBlocks()
+
         hasDPA = False
         for b in blockList:
-            xs = b.getDpaXs()
-            if not xs:
```
keckler, line 590: I notice that we used to allow for the possibility to not have a DPA XS set, but now we preclude that by raising an error in that case.

Have the implications of this been adequately considered?

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -597,13 +622,12 @@ def updateDpaRate(self, blockList=None):
...
-            self.r.core.getMaxBlockParam(
-                "detailedDpaPeakRate", typeSpec=Flags.GRID_PLATE, absolute=False
-            )
-            * 60.0
-            * units.SECONDS_PER_YEAR
+
+        peakRate = self.r.core.getMaxBlockParam(
+            "detailedDpaPeakRate", typeSpec=Flags.GRID_PLATE, absolute=False
         )
+        self.r.core.p.peakGridDpaAt60Years = peakRate * 60.0 * units.SECONDS_PER_YEAR
```
keckler, line 629: It is so odd IMO that this is a parameter in the framework... I am an advocate of moving it into our internal global flux plugin.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -120,6 +119,17 @@ def test_getIOFileNames(self):
         inf, _outf, _stdname = gfi.getIOFileNames(1, 2, 1)
         self.assertEqual(inf, "armi001_2_001.GlobalFlux.inp")
 
+    def test_getHistoryParams(self):
+        params = globalFluxInterface.GlobalFluxInterface.getHistoryParams()
+        self.assertEqual(len(params), 3)
+        self.assertIn("detailedDpa", params)
+
+    def test_checkEnergyBalance(self):
```
keckler, line 127: If possible, it would be useful to hack the underlying method to also force this method to fail, and to do an `assertRaises(ValueError)` from within this test.

keckler (approved):
Some minor comments. Awesome!

armi/physics/neutronics/globalFlux/globalFluxInterface.py
john-science: Yeah, I noticed that too.

I would be in favor of moving this parameter in another PR very soon.  It _feels_ very case specific to me. Though we could get Nick and Jake's opinion on that.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
john-science: There was never any way for this method to return False, None, or any empty/null value. The old code did not actually allow for that, but it pretended to.  

In all such cases, the method `getDpaXs()` always just raised an error (and still does), that not caught.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
keckler: @jakehader @ntouran Any thoughts here?

keckler:
I'm good on this, if you wanted to merge.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
mgjarrett: Personal opinion: It seems extremely specific and definitely odd to take a rate and multiply it by 60 years in _every_ case. Not every reactor even has a `Flags.GRID_PLATE`. You could store the rate as a parameter, and then somebody who wants to know the accumulated DPA can just... multiply by 60 years.

Devil's advocate: 60 years is a somewhat standard reactor life (40-year initial operating license + 20 year renewal), so you could argue that it's widely relevant.

PR #944, opened 2022-10-18
Allow for a reactor to be loaded without setting its cs to the master

keckler:
## Description

When loading a reactor from a DB, it has been the default behavior to apply the associated CS to be the "master" CS for the run. This is not always desirable. Just because you load a reactor doesn't mean you want to apply it to the entire run. This small PR simply allows for that possibility.

This is _related_, but not actively working against, #930 .

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -1085,6 +1085,7 @@ class Database3:
         statePointName=None,
         allowMissing=False,
         updateGlobalAssemNum=True,
+        updateMasterCs=True,
     ):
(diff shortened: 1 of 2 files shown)


john-science (approved):
I'm okay with this.  But:

1. I will have to remember to remove this when I remove `getMasterCs()` and `setMasterCs()`. Though, I will be fairly busy until Halloween, so sure.
2. Since this feature is being removed, I'm not sure we need the release notes.

keckler:
Yeah, its in the checklist for #930 already. So should be good to go wrt remembering.

PR #946, opened 2022-10-19
Removing two usages of getMasterCs

john-science:
## Description

As part of the on-going work on #930, this PR removes to usages of `getMasterCs()` in reports and reactor.py.

I have reviewed this PRs change on downstream repos, and there is only one place I can find this will make a difference, and it is a unit test. So I will open the fixing PR into that repo.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/report/html.py
+++ b/armi/bookkeeping/report/html.py
@@ -199,15 +199,14 @@ def encode64(file_path):
 # ---------------------------
 
 
-def writeStandardReportTemplate(f, report):
+def writeStandardReportTemplate(f, report, caseTitle=""):
     f.write(r"<!DOCTYPE html>" + "\n")
-    cs = settings.getMasterCs()
     with Html(f):
-
         with Head(f):
             f.write(r'<meta charset="UTF-8">' + "\n")
-            with Title(f):
-                f.write(cs.caseTitle)
+            if caseTitle:
+                with Title(f):
+                    f.write(caseTitle)
 
         with Body(f):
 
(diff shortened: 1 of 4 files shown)


armi/bookkeeping/report/html.py
```diff
@@ -264,8 +264,6 @@ def writeStandardReportTemplate(f, report):
                 with Div(f, attrs={"class": "page-header"}):
                     with H1(f):
                         f.write(report.title)
-                    with P(f):
-                        f.write(cs["comment"])
```
keckler, line 268: Don't know if you care or noticed, but just pointing out here that you removed functionality which I think _did_ work before.

Maybe you meant to add that as an argument as well?

keckler (approved):

armi/bookkeeping/report/html.py
john-science: Well, I can add it back in.  

But I couldn't find that many examples of people using this code, and the examples I _did_ find had empty "comment" settings.

If you can find me (offline) a counter example of someone putting a comment into a report, I can put it back in.  I just thought this looked like a mostly unused system, and I was trying to make it easier to support long term. 

I'm not married to this one change, if you have reason for me to put it back in.

keckler: I don't really know if people use the `comment` case setting anyways haha. 

I don't particularly care if it is removed, just seemed unnecessary. If you're willing to deal with whomever this might impact (if anyone), then go for it.

PR #952, opened 2022-10-27
Organize output files by cycle/node.

mgjarrett:
## Description

ARMI can output a large number of input files at each cycle and node. Most of these file names are re-used at different cycle/node points, so they get overwritten. To improve organization and retain all important inputs and outputs, the directory changer is being updated to copy these files to a permanent location in the working directly, under folders named `c0n0`, `c0n1`, etc, using a new setting `savePhysicsFiles`.

### Note

This PR does not change the functionality of `dumpSnapshot`, which is now used to run detailed physics calculations at selected time points.

---

## Checklist

### DIFF
--- a/armi/physics/executers.py
+++ b/armi/physics/executers.py
@@ -21,7 +21,7 @@ data pathways.
 import os
 import hashlib
 
-from armi.utils import directoryChangers
+from armi.utils import directoryChangers, pathTools
 from armi import runLog
 from armi.context import getFastPath, MPI_RANK
 
@@ -178,11 +178,18 @@ class DefaultExecuter(Executer):
             )
         self._performGeometryTransformations()
         inputs, outputs = self._collectInputsAndOutputs()
+        state = f"c{self.r.p.cycle}n{self.r.p.timeNode}"
+        outputDir = os.path.join(
+            pathTools.armiAbsPath(os.getcwd()), state, self.options.label
+        )
         # must either write input to CWD for analysis and then copy to runDir
         # or not list it in inputs (for optimization)
         self.writeInput()
         with directoryChangers.ForcedCreationDirectoryChanger(
-            self.options.runDir, filesToMove=inputs, filesToRetrieve=outputs
+            self.options.runDir,
+            filesToMove=inputs,
+            filesToRetrieve=outputs,
+            outputPath=outputDir,
         ) as dc:
             self.options.workingDir = dc.initial
             self._execute()
--- a/armi/physics/executers.py
+++ b/armi/physics/executers.py
@@ -56,6 +56,9 @@ class ExecutionOptions:
     interface : str
         A name for the interface calling the Executer that may be used to organize the
         input/output files generated within sub-folders under the working directory.
+    dumpSnapshot : bool
+        Dump the physics kernel I/O files from the execution to a dedicated directory that
+        will not be overwritten so they will be available after the run.
     applyResultsToReactor : bool
         Update the in-memory reactor model with results upon completion. Set to False
         when information from a run is needed for auxiliary purposes rather than progressing
@@ -74,6 +77,7 @@ class ExecutionOptions:
         self.interfaceName = None
         self.applyResultsToReactor = True
         self.paramsToScaleSubset = None
+        self.dumpSnapshot = False
 
     def __repr__(self):
         return f"<{self.__class__.__name__}: {self.label}>"
@@ -184,8 +188,7 @@ class DefaultExecuter(Executer):
         inputs, outputs = self._collectInputsAndOutputs()
         state = f"c{self.r.p.cycle}n{self.r.p.timeNode}"
         dirName = self.options.interfaceName or self.options.label
-        cycleNodeStamp = "{self.r.p.cycle:03d}{self.r.p.timeNode:03d}"
-        if cycleNodeStamp in self.cs["dumpSnapshot"]:
+        if self.options.dumpSnapshot:
             outputDir = os.path.join(pathTools.armiAbsPath(os.getcwd()), state, dirName)
         else:
             outputDir = None
--- a/armi/physics/executers.py
+++ b/armi/physics/executers.py
@@ -184,7 +184,11 @@ class DefaultExecuter(Executer):
         inputs, outputs = self._collectInputsAndOutputs()
         state = f"c{self.r.p.cycle}n{self.r.p.timeNode}"
         dirName = self.options.interfaceName or self.options.label
-        outputDir = os.path.join(pathTools.armiAbsPath(os.getcwd()), state, dirName)
+        cycleNodeStamp = "{self.r.p.cycle:03d}{self.r.p.timeNode:03d}"
+        if cycleNodeStamp in self.cs["dumpSnapshot"]:
+            outputDir = os.path.join(pathTools.armiAbsPath(os.getcwd()), state, dirName)
+        else:
+            outputDir = None
         # must either write input to CWD for analysis and then copy to runDir
         # or not list it in inputs (for optimization)
         self.writeInput()
--- a/armi/physics/executers.py
+++ b/armi/physics/executers.py
@@ -53,6 +53,9 @@ class ExecutionOptions:
         a shared network location. Auto-applied during execution by default.
     label : str
         A name for the run that may be used as a prefix for input/output files generated.
+    interface : str
+        A name for the interface calling the Executer that may be used to organize the
+        input/output files generated within sub-folders under the working directory.
     applyResultsToReactor : bool
         Update the in-memory reactor model with results upon completion. Set to False
         when information from a run is needed for auxiliary purposes rather than progressing
@@ -68,6 +71,7 @@ class ExecutionOptions:
         self.runDir = None
         self.workingDir = None
         self.label = label
+        self.interfaceName = None
         self.applyResultsToReactor = True
         self.paramsToScaleSubset = None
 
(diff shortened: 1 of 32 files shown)


keckler:
Yes, IT is going to love this! 😆

mgjarrett:
> Yes, IT is going to love this! 😆

Ideally it will be easier to delete old files you don't want because they'll be grouped into a handful of folders, rather than just having 1000+ loose files in the working directory. For now we still are copying to the working directory so the files are there for codes to interface with each other, but hopefully we can get to a point where the physics codes can find the files they need within these directories.

john-science:
Why did you change the `TemporaryDirectoryChanger`? The goal of that feature is to make sure that a test can create as many files as it needs, willy nilly, and be easily disposable. We do not want unit tests to be sharing a scratch space.

armi/physics/executers.py
```diff
@@ -178,11 +178,18 @@ def run(self):
             )
         self._performGeometryTransformations()
         inputs, outputs = self._collectInputsAndOutputs()
+        state = f"c{self.r.p.cycle}n{self.r.p.timeNode}"
+        outputDir = os.path.join(
+            pathTools.armiAbsPath(os.getcwd()), state, self.options.label
+        )
```
john-science, line 184: This change requires a change to our `.gitignore` file as well, right?

Because this change would definitely pollute the repo with all kinds of extra folders?

john-science:
@jakehader  Please do not merge this yet.

@mgjarrett  Have you tested this downstream in our other important repos and projects?  It appears to me that most of the projects that use ARMI to run simulations would have to change to meet your new change.  I would like you to propose this change at the next Monday meeting, and spearhead the work in what I am guessing is about 12 other pull requests in other repositories.   Thanks so much!

mgjarrett:
> Why did you change the `TemporaryDirectoryChanger`? The goal of that feature is to make sure that a test can create as many files as it needs, willy nilly, and be easily disposable. We do not want unit tests to be sharing a scratch space.

There are downstream repos that use the `TemporaryDirectoryChanger` for things other than testing. The temporary directory is still cleaned up, but this provides the option to provide an `outputPath` where the files will be transferred to. If no `outputPath` is passed, then there should be no additional permanent files created that weren't created before.

john-science:
@mgjarrett  Can you verify something for me?  Something that would be Bad would be if running 2 simulations in a row could now result in over-writing the first results because the second results are in the same folder.

That's not the case with your re-design here, right?  

Have you tested that?  
Thanks!

mgjarrett:
> @mgjarrett Can you verify something for me? Something that would be Bad would be if running 2 simulations in a row could now result in over-writing the first results because the second results are in the same folder.

This is the current behavior of ARMI, and this PR does not change that. If you re-run ARMI in the same folder where you have previously run it, it will overwrite any existing files as new ones are generated.

The benefit of this PR is that it allows you to store inputs/outputs from a given time step in the simulation. Previously, any auxiliary I/O from a physics kernel would be overwritten unless it was given a name with a unique cycle/step identifier.

mgjarrett:
After our internal user meeting, @ntouran brought up that what we are implementing here is basically the old behavior of the `dumpSnapshot` setting. `dumpSnapshot` is now used to run detailed physics calculations at selected time points.

I am proposing with this PR to change the name of the current `dumpSnapshot` setting to `runDetailedSnapshot`. This makes it more clear to the user that the setting is running detailed physics calculations at the specified "snapshots" (i.e., cycle/node time points). This setting is used during a `runType: Snapshots` run.

With this change, we can reclaim the `dumpSnapshot` setting for the functionality that has been re-introduced here: dumping all of the physics kernel I/O at selected "snapshot" statepoints during a `runType: Standard` run.

To summarize:
 - Old `dumpSnapshot` + `runType: Snapshots` will become `runDetailedSnapshot` + `runType: Snapshots`
 - New `dumpSnapshot` + `runType: Standard` will dump the I/O files into a folder at the specified cycle/node steps.
 
 For example:
```
dumpSnapshot:
  - 000000
  - 000001
  - 001000
runType: Standard
```
will dump I/O at `c0n0`, `c0n1`, and `c1n0`, but no other points. If no `dumpSnapshot` setting is provided, none of the additional I/O will be dumped. Without `dumpSnapshot`, the behavior of ARMI should be identical to before this PR.

jakehader:
This looks great to me! Holding on any approvals and landing until we get the greenlight from @john-science

john-science:
I am looking at the PR now.

I'm not 100% convinced about the naming.  Will a user understand the difference between "dumpSnapshot" and "runDetailedSnapshot" from the names?  Maybe "dumpSnapshot" and "dumpDetailedSnapshot" makes it more clear they are related and have only a difference in level of detail?

Naming is hard, I'm just thinking.  I'll be willing to expect the name if you think it's clear to the user.

(Also, I know changing the name now is 100 times easier than changing it a year from now, after everyone is already using it.)

john-science:
Sorry, I can't tell, did we get rid of the "dumpSnapshots" setting in this PR?

It looks like we did, but I thought we were keeping that and _adding_ a new level of snapshot detail?

(I just mention it because I quickly found 9 downstream repos that use that setting.  So, even just changing the spelling of that name would generate a lot of work that feels unnecessary if we can leave the name unchanged.)

mgjarrett:
> I am looking at the PR now.
> 
> I'm not 100% convinced about the naming. Will a user understand the difference between "dumpSnapshot" and "runDetailedSnapshot" from the names? Maybe "dumpSnapshot" and "dumpDetailedSnapshot" makes it more clear they are related and have only a difference in level of detail?
> 
> Naming is hard, I'm just thinking. I'll be willing to expect the name if you think it's clear to the user.
> 
> (Also, I know changing the name now is 100 times easier than changing it a year from now, after everyone is already using it.)

We can change the names to whatever we want. I am suggesting these names because `runDetailedSnapshot` actually runs additional calculations, while `dumpSnapshot` only dumps I/O but doesn't perform any additional calculation. `dumpDetailedSnapshot` is fine if we think that's better, but I was trying to highlight that it runs something, it doesn't just dump files.

mgjarrett:
> Sorry, I can't tell, did we get rid of the "dumpSnapshots" setting in this PR?

I am proposing renaming it to `runDetailedSnapshot`. We can keep it the same, or make it anything else we want; I'm putting that forth as a reasonable representation of what the setting does.

> It looks like we did, but I thought we were keeping that and _adding_ a new level of snapshot detail?

We're not adding the detailed calculation; the current setting `dumpSnapshot` does the detailed calculation. We're trying to add a setting that doesn't do the detailed stuff, which is what `dumpSnapshot` used to do before it was commandeered for its current purpose.

> (I just mention it because I quickly found 9 downstream repos that use that setting. So, even just changing the spelling of that name would generate a lot of work that feels unnecessary if we can leave the name unchanged.)

Yes, we'd have to update the name of the setting in downstream repos. They're trivial PRs, but it is unnecessary work. I am arguing that the current setting name is confusing because it doesn't just dump data, it also runs additional calculations. But that might not outweigh the burden of changing a setting name. If that's the case, we can come up with another name for what this PR is doing. Maybe `savePhysicsIO`?

john-science:
@mgjarrett Okay, this PR has gone on too long, sorry about that.

@ntouran Agrees though that we should either:

a. Leave the `dumpSnapshot` setting 100% alone and just add a new setting, or
b. Entirely remove the `dumpSnapthot`, and use [SettingRenamer](https://github.com/terrapower/armi/blob/bc3cf475b2158bc3be626287b18f2e21d5da52a7/armi/settings/settingsIO.py#L52).

Nick thinks you're probably right, that we should just rename the setting.

mgjarrett:
Okay, I am happy to change the settings names to whatever we all agree on.

For the new setting being implemented here, what is the favored name? I threw out `savePhysicsIO` but I am happy to hear other suggestions.

I'll leave `dumpSnapshot` alone in this PR; if we want to change it we could open another PR for that.

john-science:
> For the new setting being implemented here, what is the favored name? I threw out `savePhysicsIO` but I am happy to hear other suggestions.

It's up to you. It's your baby. I'm not sure `IO` means anything to me.  But I don't have like a _great_ alternative all lined up:

* `savePhysicsAtNodes`
* `dumpNodeSnapshots`
* `nodeSnapshots`

The only thing we _need_ is for the name (and the `description` field in [globalSettings.py](https://github.com/terrapower/armi/pull/952/files#diff-65b2fa0555ff1b7c5f4da946d3b799aaa4df8b32c41858191281bc575d2b02ed)) to be as descriptive and helpful as possible for users.

mgjarrett:
> > For the new setting being implemented here, what is the favored name? I threw out `savePhysicsIO` but I am happy to hear other suggestions.
> 
> It's up to you. It's your baby. I'm not sure `IO` means anything to me. But I don't have like a _great_ alternative all lined up:
> 
> * `savePhysicsAtNodes`
> * `dumpNodeSnapshots`
> * `nodeSnapshots`
> 
> The only thing we _need_ is for the name (and the `description` field in [globalSettings.py](https://github.com/terrapower/armi/pull/952/files#diff-65b2fa0555ff1b7c5f4da946d3b799aaa4df8b32c41858191281bc575d2b02ed)) to be as descriptive and helpful as possible for users.

Okay, I'm hesitant to use `snapShot` in the name now because there could be confusion between the functionality of `dumpSnapshot` vs. `dumpNodeSnapshots` or `nodeSnapshots`.

`IO` is supposed to mean Inputs and Outputs, but `savePhysicsFiles` might be more descriptive. I'll try that.

john-science (approved):

PR #959, opened 2022-11-05
Use r.core.axialMesh() for uniform mesh.

mgjarrett:
## Description

When making a uniform mesh assembly in the `UniformMeshConverter`, the freshly minted assembly does not have any submesh (i.e., `b.p.axMesh = 1`).

https://github.com/terrapower/armi/blob/bc3cf475b2158bc3be626287b18f2e21d5da52a7/armi/reactor/converters/uniformMesh.py#L380

Creating a uniform mesh assembly with the refAssem's block mesh (`r.core.refAssem.getAxialMesh()`) poses problems if some blocks in the `refAssem` have a submesh (i.e., `b.p.axMesh > 1`), because the uniform mesh assembly will not have a submesh. `r.core.p.axialMesh()` contains the computational neutronics mesh, which has all of the mesh subdivisions; this is the mesh that should be used for `makeAssemWithUniformMesh()`.

---

## Checklist

### DIFF
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -131,11 +131,12 @@ class UniformMeshGeometryConverter(GeometryConverter):
                 f"the core's reference assembly mesh: {r.core.refAssem.getAxialMesh()}"
             )
             self.convReactor = self._sourceReactor
+            self.convReactor.core.updateAxialMesh()
             self._setParamsToUpdate()
             for assem in self.convReactor.core.getAssemblies(self._nonUniformMeshFlags):
                 homogAssem = self.makeAssemWithUniformMesh(
                     assem,
-                    self.convReactor.core.refAssem.getAxialMesh(),
+                    self.convReactor.core.p.axialMesh[1:],
                     self.blockParamNames,
                 )
                 homogAssem.spatialLocator = assem.spatialLocator
--- a/armi/reactor/converters/tests/test_uniformMesh.py
+++ b/armi/reactor/converters/tests/test_uniformMesh.py
@@ -68,7 +68,6 @@ class TestAssemblyUniformMesh(unittest.TestCase):
                 )
             prevB = newB
 
-
         newAssemNumberDens = newAssem.getNumberDensities()
         for nuc, val in sourceAssem.getNumberDensities().items():
             self.assertAlmostEqual(val, newAssemNumberDens[nuc])
@@ -95,9 +94,10 @@ class TestAssemblyUniformMesh(unittest.TestCase):
 
         self.assertNotEqual(len(newAssem), len(sourceAssem))
         newHeights = [b.getHeight() for b in newAssem]
-        sourceHeights = [ b.getHeight() / b.p.axMesh for b in sourceAssem for i in range(b.p.axMesh)]
+        sourceHeights = [
+            b.getHeight() / b.p.axMesh for b in sourceAssem for i in range(b.p.axMesh)
+        ]
         self.assertListEqual(newHeights, sourceHeights)
-
 
     def test_makeAssemUniformMeshParamMappingSameMesh(self):
         """Tests creating a uniform mesh assembly while mapping both number densities and specified parameters."""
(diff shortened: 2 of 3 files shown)


mgjarrett:
I ran the tests locally and they passed, so... I don't know? The 3.7, 3.8, 3.9, and 3.10 unit tests all pass, but the "ARMI Windows tests" fails. I can't think of anything changed here that would be affecting the failing test.

john-science:
@mgjarrett I would guess that the issue is a race condition somewhere in `test_assemblies.py` or `textProcessors.py`.

```
    File "D:\a\armi\armi\armi\reactor\tests\test_assemblies.py", line 589, in _setup_blueprints
...
    File "D:\a\armi\armi\armi\utils\textProcessors.py", line 106, in _processIncludes
      raise ValueError(
  ValueError: The !included file, `refSmallSfpGrid.yaml` does not exist from D:\a\armi\armi\armi\tests\detailedAxialExpansion!
```

My assumption here is if you ran the test again, the bug would not appear.

I'll be honest, I've run the ARMI unit tests thousands of times, and not seen this one. So if I see it a second time, I'll have to fix it. But I'm hoping this is a one-off bug with the Windows file system. 

> Probably, your next commit will fix this, because it will force the tests to re-run.

armi/reactor/converters/uniformMesh.py
```diff
@@ -131,11 +131,12 @@ def convert(self, r=None):
                 f"the core's reference assembly mesh: {r.core.refAssem.getAxialMesh()}"
             )
             self.convReactor = self._sourceReactor
+            self.convReactor.core.updateAxialMesh()
             self._setParamsToUpdate()
             for assem in self.convReactor.core.getAssemblies(self._nonUniformMeshFlags):
                 homogAssem = self.makeAssemWithUniformMesh(
                     assem,
-                    self.convReactor.core.refAssem.getAxialMesh(),
+                    self.convReactor.core.p.axialMesh[1:],
```
john-science, line 139: Has the `core.p.axialMesh` included in parameters been thermally expanded?  (@albeanth My question is is if this axial mesh on the parameters is always correct at this point in time. I can't tell.)

armi/reactor/converters/uniformMesh.py
```diff
@@ -131,11 +131,12 @@ def convert(self, r=None):
                 f"the core's reference assembly mesh: {r.core.refAssem.getAxialMesh()}"
             )
             self.convReactor = self._sourceReactor
+            self.convReactor.core.updateAxialMesh()
             self._setParamsToUpdate()
             for assem in self.convReactor.core.getAssemblies(self._nonUniformMeshFlags):
                 homogAssem = self.makeAssemWithUniformMesh(
                     assem,
-                    self.convReactor.core.refAssem.getAxialMesh(),
```
john-science, line 138: Honest question:

@mgjarrett If this is all about submeshes, shouldn't this change of code (`getAxialMesh()` -> `p.axialMesh`) be wrapped in a check for submeshes?

(Something like `if block.p.axMesh > 1`)

Sorry, I'm just trying to make sure I fully understand the change.  Like: Should it _always_ be the new way, or does it still have to be the old way sometimes?

Sorry to bother!

mgjarrett: Good question. I cannot think of a reason why it would need to be the old way. If there is no submesh, `p.axialMesh` and `getAxialMesh()` should be equivalent.

Since the assembly created by `makeAssemWithUniformMesh` never has a submesh, we should homogenize it onto the `p.axialMesh` (which also has no submesh), instead of the `getAxialMesh()`.

armi/reactor/converters/uniformMesh.py
mgjarrett: Since we call `self.convReactor.core.updateAxialMesh()` immediately above, the `p.axialMesh` will be consistent with the `refAssem` axial mesh, which should be thermally expanded at this point.

https://github.com/terrapower/armi/blob/8757794cac263175430989da31bb504e345bdd0d/armi/reactor/reactors.py#L1830-L1866

john-science (approved):

Issue #830, opened 2022-08-11
Remove operator from reactor


jakehader:
There are only a handful of uses of `r.o`. it would be great to get rid of these, as:

- it is unclear why a reactor should have an operator
- reactor construction is muddled by being incomplete until the reactor has an operator
- many instances are only used for getting a `cs`
- removing the `r.o` would reduce the number of circular references `r.o` -> `o.r` -> `r.o` ...
- in almost all circumstances, a reactor and an operator are already in scope.

john-science:
Yeah, it's never made sense to me that the `Reactor` object should have an `Operator` on it.

It really seems like accidental design, not purposeful design.

john-science:
Here are all of the usages of this I could find in ARMI (though, of course, any number of plugins might be using this "feature"):

```python
armi/interfaces.py:  self.o = r.o if r else None
armi/mpiActions.py:  self.r.o = self.o

armi/operators/operator.py:  r.o = self
armi/operators/operator.py:  self.r.o = self
armi/operators/operator.py:  self.r.o = None

armi/physics/neutronics/globalFlux/globalFluxInterface.py:  daysIntoCycle = sum(self.r.o.stepLengths[cycle][:timeNode])
armi/physics/neutronics/macroXSGenerationInterface.py:  sum([self.r.o.burnSteps[i] + 1 for i in range(cycle)]) + self.r.p.timeNode

armi/reactor/converters/geometryConverters.py:  self._o = self._sourceReactor.o
armi/reactor/converters/parameterSweeps/generalParameterSweepConverters.py:  fh = r.o.getInterface("fuelHandler")

armi/reactor/converters/uniformMesh.py:  newReactor = Reactor(sourceReactor.o.cs.caseTitle, bp)
armi/reactor/converters/uniformMesh.py:  coreDesign.construct(sourceReactor.o.cs, bp, newReactor, loadAssems=False)
armi/reactor/converters/uniformMesh.py:  _geometryConverter = newReactor.core.growToFullCore(sourceReactor.o.cs)

armi/utils/reportPlotting.py:  dbi = reactor.o.getInterface("database")
armi/utils/reportPlotting.py:  dd = reactor.o.getInterface("ductDistortion")
armi/utils/reportPlotting.py:  history = reactor.o.getInterface("history")
```

john-science:
@jakehader  So far, there are no downstream impacts. I only did the removed the really low-hanging fruit versions of `r.o` (that I could fit into my day).

[...]

PR #961, opened 2022-11-10
Added reactor operator back into geometry converted reactor for model…

LMikeH:
…ing FPs, changed check for needing to model fission products within the lattice physics writer based on if FPs are present, which generalizes to allow blanket fuel to be depleted.

## Description

Users were running into issues with modeling MC2 1D CR with global flux solvers for self-shielding calculations. The error is caused by blocks not containing fission products due to the operator missing in the converted reactor.

---

## Checklist

- [ x ] This PR has only one purpose or idea.
- [ x ] Tests have been added/updated to verify that the new/changed code works.

- [ x ] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [ x ] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html).

- [ x ] The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any bug fixes or new features.
- [ x ] The documentation is still up-to-date in the `doc` folder.
- [ x ] The dependencies are still up-to-date in `setup.py`.

### DIFF
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -197,7 +197,9 @@ class UniformMeshGeometryConverter(GeometryConverter):
         coreDesign.construct(cs, bp, newReactor, loadAssems=False)
         newReactor.core.lib = sourceReactor.core.lib
         newReactor.core.setPitchUniform(sourceReactor.core.getAssemblyPitch())
-        newReactor.o = sourceReactor.o  # This is needed later for geometry transformation
+        newReactor.o = (
+            sourceReactor.o
+        )  # This is needed later for geometry transformation
 
         # check if the sourceReactor has been modified from the blueprints
         if sourceReactor.core.isFullCore and not newReactor.core.isFullCore:
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -197,6 +197,7 @@ class UniformMeshGeometryConverter(GeometryConverter):
         coreDesign.construct(cs, bp, newReactor, loadAssems=False)
         newReactor.core.lib = sourceReactor.core.lib
         newReactor.core.setPitchUniform(sourceReactor.core.getAssemblyPitch())
+        newReactor.o = sourceReactor.o  # This is needed later for geometry transformation
 
         # check if the sourceReactor has been modified from the blueprints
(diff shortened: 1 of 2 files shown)


keckler:
This change combined with the change to our internal lattice physics plugin seems to solve the issue that I was having.

jakehader:
@LMikeH can you run black on this before we can land?

jakehader:
@LMikeH you need to also link your TerraPower email to.your GitHub account so that you can sign the CLA before I land. Thank you for running black!

armi/reactor/converters/uniformMesh.py
```diff
@@ -197,6 +197,9 @@ def initNewReactor(sourceReactor, cs):
         coreDesign.construct(cs, bp, newReactor, loadAssems=False)
         newReactor.core.lib = sourceReactor.core.lib
         newReactor.core.setPitchUniform(sourceReactor.core.getAssemblyPitch())
+        newReactor.o = (
+            sourceReactor.o
+        )  # This is needed later for geometry transformation
```
john-science, line 202: This is fine as a quick fix.

Seems like the real problem is that a geometry converter (or downstream repo?) is still depending on `r.o`.

john-science:
If you just need a hot fix right now, I can support this PR. 


We/I will have to undo it real soon, of course.  We have [this open ticket](https://github.com/terrapower/armi/issues/830) that shows the actual cause of your problem, that `HexToRZThetaConverter` and `parameterSweepConverters` are some of the only places left that use this backwards `r.o`:

```
armi/reactor/converters/geometryConverters.py:  self._o = self._sourceReactor.o

armi/reactor/converters/parameterSweeps/generalParameterSweepConverters.py:  fh = r.o.getInterface("fuelHandler")
```

Both of those seem pretty straight forward to fix in ARMI, as they really aren't used much. Someone just got a little lazy about passing in the lumped fission products.  But I haven't looked in downstream projects to see what the LOE would be like there.  So, sure.

jakehader:
Thanks @john-science - Yeah we need this as a hot fix, but thank you for your work on removing the reactor's reference to the operator. I think one of the biggest things holding us back from removing the operator in some cases is that interfaces on the operator hold state that isn't recoverable in other ways and it's not easy to create a new interface on the fly and get that state back.

jakehader (approved):

PR #962, opened 2022-11-14
Moving FuelHandler Algorithms to separate files

john-science:
## Description

The `FuelHandler.py` file has a ton of algorithms of various kinds that are very specific to hex assembly reactors. This has caused lots of confusion for people who are implementing some sort of fuel handling logic.  It also muddies the waters significantly between general fuel handling logistics code and detailed hex assembly algorithms.

The code in the old `fuelHandler.py` has been broken up into three files:

* `fuelHandlers.py` - has `FuelHandler` class
* `hexAssemblyFuelMgmtUtils.py` - Assorted shuffling and misc algorithms for managing hex assembly reactors.
* `assemblyRotationAlgorithms.py` - Rotation algorithms called by `FuelHandler.outage()`.  (This _does_ seem to be specific to assembly/hex-assembly reactor, but `FuelHandler.outage()` will only call it if a particular setting is used, which does an okay job of keeping the `FuelHandler` class from being specific to hex assemblies.)

---

## Checklist


armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -121,16 +118,16 @@ def outage(self, factor=1.0):
 
         # do rotations if pin-level details are available (requires fluxRecon plugin)
         if self.cs["fluxRecon"] and self.cs["assemblyRotationAlgorithm"]:
-            # Rotate assemblies ONLY IF at least some assemblies have pin detail (enabled by fluxRecon)
+            # Rotate assemblies ONLY IF at least some assemblies have pin detail
             # The user can choose the algorithm method name directly in the settings
-            if hasattr(self, self.cs["assemblyRotationAlgorithm"]):
-                rotationMethod = getattr(self, self.cs["assemblyRotationAlgorithm"])
+            if hasattr(rotAlgos, self.cs["assemblyRotationAlgorithm"]):
+                rotationMethod = getattr(rotAlgos, self.cs["assemblyRotationAlgorithm"])
```
keckler, line 124: This change is kind of interesting, because it will prevent the user from adding and using their own custom rotation algorithm just via a subclass. Before, that would have been possibly by subclassing the `FuelHandler` class and adding their algorithm as a method. But now, the user will have to get their algorithm merged into the `assemblyRotationAlgorithms` module, or else this check will fail. 

Or they could subclass `FuelHandler` and then override this check, I suppose, but that's a little opaque.

Or am I misinterpreting that?

keckler (approved):
It makes a lot of sense to move those fuel handler methods off of the class anyways, or at least to make them into staticmethods, because none of them utilized the class/instance attributes at all.

armi/physics/fuelCycle/fuelHandlers.py
john-science: Yeah, I explicitly do not like that we had hex assembly rotation algorithms in the `FuelHandler` base class. It was just a bad design, and several people have run up against it as confusing.

Looking at the 8 years of Git history I have access to, no one has ever used this import magic "setting to rotation algorithm" feature in ARMI to extend and add more algorithms.  So I wasn't really too fussed with trying to support this import magic feature to it's fullest extent.

The new design would make it easier to add more algorithms to ARMI for real, not just ad-hoc.  They would be visible, and clearly defined as what they are.

I think it's a better choice, as it'll be easier for people to understand.  At some point, we could completely remove all the import magic, as we don't need it anywhere else.  But small steps.

keckler: Sure, just pointing it out.

john-science: That's fair.

PR #965, opened 2022-11-15
Replace yaml.load with yaml.YAML.load to get rid of deprecation warnings

opotowsky:
## Description

This is a common warning:

```python
armi\nucDirectory\nuclideBases.py:311: PendingDeprecationWarning:
  load will be removed, use

    yaml=YAML(typ='unsafe', pure=True)
    yaml.load(...)

  instead
    mc2Nuclides = yaml.load(mc2Nucs, yaml.RoundTripLoader)
'''

But it doesn't have to be that way. And the best part is that the warning told me how to fix it. Opening the PR to see how the unit tests do on different systems. 

---

## Checklist

<!--
    You (the pull requester) should put an `x` in the boxes below you have completed.
    If you're unsure about any of them, don't hesitate to ask. We're here to help!
    Learn what a "good PR" looks like here: https://terrapower.github.io/armi/developer/tooling.html#good-pull-requests
-->

<!-- Check the code quality -->

<!-- Check the project-level cruft -->

### DIFF
--- a/armi/cases/inputModifiers/tests/test_inputModifiers.py
+++ b/armi/cases/inputModifiers/tests/test_inputModifiers.py
@@ -200,7 +200,7 @@ class TestsuiteBuilderIntegrations(unittest.TestCase):
                 f"case-suite-testBPBM/000{case_nbr}/armi-000{case_nbr}-blueprints.yaml",
                 "r",
             )
-            bp_dict = yaml.load(yamlfile)
+            bp_dict = yaml.YAML(typ="unsafe", pure=True).load(yamlfile)
             yamlfile.close()
 
             self.assertEqual(bp_dict["blocks"]["fuel 1"]["clad"]["od"], 3.14)
--- a/armi/nucDirectory/nuclideBases.py
+++ b/armi/nucDirectory/nuclideBases.py
@@ -308,7 +308,7 @@ def __readMc2Nuclides():
     that have already been added from RIPL.
     """
     with open(os.path.join(context.RES, "mc2Nuclides.yaml"), "r") as mc2Nucs:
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: armi/cases/inputModifiers/tests/test_inputModifiers.py [hunk 1]
```diff
@@ -200,7 +200,7 @@ def SuiteNaming(index, _case, _mods):
                 f"case-suite-testBPBM/000{case_nbr}/armi-000{case_nbr}-blueprints.yaml",
                 "r",
             )
-            bp_dict = yaml.load(yamlfile)
+            bp_dict = yaml.YAML(typ="unsafe", pure=True).load(yamlfile)
```
john-science, line 203: What is this "unsafe" about?

opotowsky: It's been hard to find. I dont know why the deprecation warning printed that out. 

From the [docs](https://yaml.readthedocs.io/en/latest/basicuse.html): 

> typ='safe' accomplishes the same as what safe_load() did before: loading of a document without resolving unknown tags. Provide pure=True to enforce using the pure Python implementation, otherwise the faster C libraries will be used when possible/available but these behave slightly different (and sometimes more like a YAML 1.1 loader).

Seems like I could remove both the `typ` and `pure` args and it would all run fine.

opotowsky: I can't load the docstring, but the answer here explains the different options. Seems like we want just the default behavior though. https://stackoverflow.com/questions/51316491/ruamel-yaml-clarification-on-typ-and-pure-true

john-science (approved):

Issue #939, opened 2022-10-12
Parameters defined by UserPlugin do not make their way to the parameter collection


drewj-usnctech:
Very interested in the power of the `UserPlugin` interface! I'm trying to get a situation going where a `UserPlugin` defines some new parameters, maybe because it's an in-development feature not ready for prime time, and see how it works.

```python
from typing import List
from random import random

from armi.plugins import UserPlugin, HOOKIMPL
from armi.reactor.reactors import Core
from armi.interfaces import Interface, InterfaceInfo, STACK_ORDER
from armi.reactor.parameters import (
    ParameterDefinitionCollection,
    ParamLocation,
    Category,
)


class ExternalPlugin(UserPlugin):
    """Demo external plugin that exposes some trivial interfaces"""

    @staticmethod
    @HOOKIMPL
    def exposeInterfaces(cs) -> List[Interface]:
        return [
            InterfaceInfo(
                STACK_ORDER.BEFORE + STACK_ORDER.BOOKKEEPING, ExternalInterface, {}
            )
        ]

    @staticmethod
    @HOOKIMPL
    def defineParameters():
        import pdb

        # Set a breakpoint if we hit this call
        pdb.set_trace()

        return {
            Core: _coreRNGParam(),
        }


def _coreRNGParam() -> ParameterDefinitionCollection:
    params = ParameterDefinitionCollection()

    with params.createBuilder() as pb:
        pb.defParam(
            "rng",
            description="Random number",
            default=None,
            location=ParamLocation.VOLUME_INTEGRATED,
            categories=[Category.volumeIntegrated],
        )

    return params


class ExternalInterface(Interface):
    name = "Demo external interface"

    def interactEveryNode(self, cycle, node):
        self.r.core.p.rng = random()

```

When I run this, I see the `ExternalInterface` correctly in the stack
```
=========== 05 - Demo external interface        Init            ===========
```
but it falls with `ExternalInterface.interactEveryNode` for cycle 0, node 0
```
===========  Triggering EveryNode - cycle 0, node 0 Event ===========
=========== 01 - main                           EveryNode - cycle 0, node 0 ===========
=========== 02 - fissionProducts                EveryNode - cycle 0, node 0 ===========
=========== 03 - xsGroups                       EveryNode - cycle 0, node 0 ===========
=========== 04 - history                        EveryNode - cycle 0, node 0 ===========
=========== 05 - Demo external interface        EveryNode - cycle 0, node 0 ===========
[...]

PR #967, opened 2022-11-17
UserPlugins can't define Parameters

john-science:
## Description

Based on the discussion [in ticket #939](https://github.com/terrapower/armi/issues/939), I did some research and found the obvious truth that `UserPlugin`s can't (at least yet) define parameters.

Essentially, when we `armi.Configure(MyApp())`, we get the full set of `Parameter`s from the ARMI `App()`.  THEN, optionally, must later, we can read in a Settings file, that might include a `UserPlugin`.  And at that point, it's just too late to define a `Parameter`.  This is, I'm sure, fixable. But it would demand a high-level re-design of how and when we read plugins and parameters in ARMI.

This has always been a limit of my `UserPlugin` design. But no one had ever tested it before.

Due to Level-of-Effort, I think it makes more sense to leave this as a limitation of `UserPlugin`s, and anyone who needs this much power should just use a regular `ArmiPlugin`.

---

## Checklist

### DIFF
--- a/armi/plugins.py
+++ b/armi/plugins.py
@@ -640,6 +640,9 @@ class UserPlugin(ArmiPlugin):
         """
         if issubclass(self.__class__, UserPlugin):
             assert (
+                len(self.__class__.defineParameters()) == 0
+            ), "UserPlugins cannot define parameters, consider using an ArmiPlugin."
+            assert (
                 len(self.__class__.defineParameterRenames()) == 0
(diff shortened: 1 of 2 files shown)


john-science:
@drewejohnson @ntouran  Just FYI.

opotowsky (approved):
Most of the work was accomplished in the discussion, so I can't speak to that fully, but I certainly approve of the implementation!

PR #984, opened 2022-11-18
Remove MPI barriers in TemporaryDirectoryChanger and Output cache.

### LINKED ISSUE #982: Remove MPI awareness requirements from Temporary Directory Changer and Output Cache
#689 added some MPI awareness and barriers to a few commonly-used utility functions like the output cache and the directory changer. However, this has some unintended consequences when these utilities are being used by worker MPI nodes again and again in parallel. For example, if worker `X` has 10 tasks to do, all of which require tempdirs, and worker `Y` has 5 tasks, then there will be 10 `barrier` calls in `X` and 5 in Y`, which will unsync the whole MPI space-time continuum and cause confusing, difficult-to-debug lockups while things deadlock waiting for barriers that are never going to arrive.

Anyway, we want to remove those `waitAll` calls and their underlying `mpi.barrier` calls from those utilities.

If they are truly necessary to solve issues we were searching for, we will need to design and implement non-global MPI solutions, which should hopefully be possible.

ntouran:
## Description

These utilities aren't always used in a way that has all MPI nodes synced up. For example if worker x has 3 tasks and worker y has 5, that could lead to a deadlock if they use these utilities. Rather than requiring all clients of these utlities know the MPI global state, we just remove the barriers, and will deal with file lock issues in more forgiving ways.

Closes #982

---

## Checklist

### DIFF
--- a/armi/utils/tests/test_pathTools.py
+++ b/armi/utils/tests/test_pathTools.py
@@ -26,6 +26,7 @@ from armi.utils import pathTools
 from armi.utils.directoryChangers import TemporaryDirectoryChanger
 
 THIS_DIR = os.path.dirname(__file__)
+MPI_COMM = context.MPI_COMM
 
 
 class PathToolsTests(unittest.TestCase):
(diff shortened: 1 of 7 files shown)


armi/utils/tests/test_pathTools.py
```diff
@@ -26,6 +26,7 @@
 from armi.utils.directoryChangers import TemporaryDirectoryChanger
 
 THIS_DIR = os.path.dirname(__file__)
+MPI_COMM = context.MPI_COMM
```
ntouran, line 29: oops don't need this

john-science (approved):
The code looks good to me!

But I will need @onufer to test if this code solves his bug downstream.

onufer:
problem is solved.

Issue #985, opened 2022-11-18
Cumulative parameters as they relate to numerical diffusion during uniform mesh conversion


keckler:
During the uniform mesh conversion process, we know that there is some numerical diffusion going on when we map non-volume-integrated block parameters (a thermodynamicist would call these "intensive" properties) from the uniform mesh back to the non-uniform mesh. 

For parameters which are NOT time-integrated (i.e. reaction rates, power density, etc.), this numerical diffusion does exist, but it might not be blatant in the results, and further it does not compound with time because the parameter is recomputed at each time node.

But for parameters which are time-integrated, the numerical diffusion can sometimes be very obvious, for instance when a value that is physically monotonically increasing ends up decreasing. We have seen this sometimes happen with accumulated DPA, for example. The other reason it can be more important is because the time history matters for time-integrated parameters, meaning the numerical diffusion directly compounds.

We _sometimes_ designate these time-integrated parameters as "cumulative", as in here, but I find that to be inconsistent (for instance, number densities are time-integrated but not specified as such):
https://github.com/terrapower/armi/blob/36fe1bced3b817d6d76e66cf99dd355219993e63/armi/reactor/blockParameters.py#L128-L133

We really have nothing to guard against this numerical diffusion. I think that there are at least a few things that we can do:
1. For time-integrated parameters that are marked as "cumulative", we can skip mapping them to/from the uniform mesh. What we really are calculating on the uniform mesh are the _rates_ at which those time-integrated parameters change, and we can do the `rate * time` calculation after all the uniform mesh stuff is over with. This will require some restructuring, but will entirely avoid non-physical results for values that should be monotonic (obviously diffusion in the rates will still take place, but that is unavoidable).
2. If we don't want to do (1), then we can at least put a check in there that if a monotonically-increasing "cumulative" parameter _decreases_, we can warn the user that they are using too coarse of a mesh. This would require us to label "cumulative" parameters as either monotonically-increasing, decreasing, or neither.
[...]

PR #992, opened 2022-11-22
Fix numerical diffusion in uniformMesh converter

mgjarrett:
## Description

Mapping anything from a non-uniform assembly to a uniform assembly and then back will cause numerical diffusion of that value. This could be a block parameter (like mgFlux, detailedDpa, etc.) or number densities. This has been observed for neutron flux while using the `nonUniformAssemFlags` setting on control rods, and for `detailedDpa` when using the `detailedAxialExpansion` setting.

This commit deals with the issue of block parameters being mapped back and forth. The number density issue has already been addressed for the nonUniformAssemFlags case, but it seems to still be present for the detailedAxialExpansion case.

The selection of which parameters to map is very much up for discussion. **Please review the categories that have been set for `REACTOR_PARAMS_TO_MAP` and `BLOCK_PARAMS_TO_MAP`, and the `_setParamsToUpdate` method for both neutron and gamma transport.** Please leave comments on any parameters you think are missing or should be changed. My approach to these selections is summarized below:

- First and foremost, we want to ensure that we are not mapping any parameters both "into" and "out of" the uniform mesh assembly, because that will cause numerical diffusion.
- We also want to make sure that when a new parameter value is generated on the uniform mesh by some physics solver, that parameter is mapped back to the non-uniform mesh.
- We should always map number densities "into" the uniform assembly but never out of it. Other parameters (like power, temperature, etc.) generally don't need to be mapped in unless they are required to set up the input for the physics kernel. It will generally be harmless to include these additional parameters **so long as they are not re-mapped back to the non-uniform assembly on the way out.**

---

## Checklist


mgjarrett:
Not requesting review yet, just tagging a few interested parties @keckler @Nebbychadnezzar

mgjarrett:
@Nebbychadnezzar can't select you as a reviewer but I wanted to get your eyes on this as well.

armi/reactor/converters/uniformMesh.py
```diff
@@ -540,10 +522,6 @@ def setAssemblyStateFromOverlaps(
                 destBlock, updatedDestVals.values(), updatedDestVals.keys()
             )
 
-            UniformMeshGeometryConverter._applyCachedParamValues(
-                destBlock, blockParamNames, cachedParams
-            )
-
```
keckler, line 546: My understanding of this change is as follows:

- Before, we were clearing all of the block parameters on the new, converted assembly, while storing some of them temporarily in a "cache". Then, we were mapping the specified parameters to the new assembly, and the cached parameters are then used to put the other parameters back onto the blocks.
- Now, we just avoid clearing the parameters in the first place, so that the caching of parameters isn't needed for anything. All parameters that are not specified specifically to be mapped are just left as their initial values?

This seems like potentially a source of great speedup. But it makes me wonder, are we losing something by changing that flow? It seems to me that an expensive step, like clearing all the parameters in the first place, must have been done for a good reason. But I cannot see why that reason might have gone away now...

armi/reactor/converters/uniformMesh.py
```diff
@@ -673,7 +612,7 @@ def reset(self):
         self._cachedReactorCoreParamData = {}
         super().reset()
 
-    def _setParamsToUpdate(self):
+    def _setParamsToUpdate(self, direction):
         """Activate conversion of various paramters."""
         self.reactorParamNames = []
         self.blockParamNames = []
```
keckler, line 618: Should this method instead be changed to raise a `NotImplemetedError`, since we never want people to actually instantiate this base converter class?

I'm not sure, your call.

armi/reactor/converters/uniformMesh.py
```diff
@@ -814,21 +761,44 @@ def __init__(self, cs=None, calcReactionRates=True):
...
+        the NeutronicsUniformMeshConverter with different class attributes for parameters to
+        map.
+
+        Parameters
+        ----------
+        direction : str
+            "in" or "out". The direction of mapping; "in" to the uniform mesh assembly, or "out" of it.
+            Different parameters are mapped in each direction.
+        """
+        UniformMeshGeometryConverter._setParamsToUpdate(self, direction)
```
keckler, line 782: This call doesn't do anything, except set up some empty lists. Which is fine. But if you wanted to go the `NotImplementedError` route that I suggested in another comment, then this obviously won't work.

armi/reactor/converters/uniformMesh.py
```diff
@@ -867,6 +837,64 @@ def _mapStateFromReactorToOther(
...
+    }
+    BLOCK_PARAMS_TO_MAP = {
+        "in": [parameters.Category.detailedAxialExpansion],
+        "out": [
+            parameters.Category.detailedAxialExpansion,
+            parameters.Category.multiGroupQuantities,
+            parameters.Category.pinQuantities,
+        ],
+    }
+    # conditions on the output BLOCK_PARAMS_TO_MAP; non-cumulative only
```
keckler, line 857: I would move this comment to the docstring as a `Note`. It's useful for people to see.

keckler (approved):
I like it. I left a handful of comments and suggestions, but I don't see anything blatantly wrong. 

After this is merged, though, there is still work to do in our internal plugins to move the application/calculation of cumulative parameters outside of the geometry conversion entry/exit. For instance, I can say for certain that `detailedDpaRate` is calculated right now when the geometry is still converted. Likely the same for many of the `cumulative` parameters. So we'll have to hunt all of those down and update them, if necessary. 

For `cumulative` parameters that are being calculated within the framework itself, I would argue it might make sense to update those as part of this PR (if any of them exist).

armi/reactor/parameters/parameterDefinitions.py
```diff
@@ -60,12 +60,14 @@
...
     """A "namespace" for storing parameter categories."""
 
+    cumulative = "cumulative"
     assignInBlueprints = "assign in blueprints"
     retainOnReplacement = "retain on replacement"
     pinQuantities = "pinQuantities"
     fluxQuantities = "fluxQuantities"
     multiGroupQuantities = "multi-group quantities"
     neutronics = "neutronics"
+    gamma = "gamma"
```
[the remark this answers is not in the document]
mgjarrett: Yes, it is confusing. If only neutron transport is run without gamma, then the `power` and `pdens` parameters are populated. Once gamma is run, then `powerGamma` and `powerNeutron` are populated.

I think they have to stay in this category, so maybe I'll include a comment or update the parameter description.

armi/reactor/blocks.py
```diff
@@ -29,8 +29,10 @@
 
 from armi import runLog
 from armi import settings
+from armi.materials import Material
```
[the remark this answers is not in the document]
mgjarrett: This line looks like it's carried over from some experimental stuff. I was originally trying to do something more complicated with creating new blocks in `makeAssemWithUniformMesh` but I gave up on that for now.

armi/reactor/converters/uniformMesh.py
```diff
@@ -160,11 +173,10 @@ def convert(self, r=None):
         else:
             runLog.extra(f"Building copy of {r} with a uniform axial mesh.")
             self.convReactor = self.initNewReactor(r, self._cs)
-            self._setParamsToUpdate()
             self._computeAverageAxialMesh()
             self._buildAllUniformAssemblies()
             self._mapStateFromReactorToOther(
-                self._sourceReactor, self.convReactor, mapNumberDensities=False
+                self._sourceReactor, self.convReactor, mapNumberDensities=True
```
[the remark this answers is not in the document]
mgjarrett: Ah, this actually needs to be changed back.

`_buildAllUniformAssemblies()` calls `makeAssemWithUniformMesh()`. By default, `makeAssemWithUniformMesh()` was previously doing `mapNumberDensities=True`. I changed the default to `False` at some point during the work on this PR and then probably set this to `True` here so the number densities will still be mapped. But changing the default of `makeAssemWithUniformMesh()` turned out to be a mistake so it is now back to the original default. With the original default, `_buildAllUniformAssemblies()` takes care of number densities and this should be set to `False`.

armi/reactor/converters/uniformMesh.py
```diff
@@ -261,7 +276,7 @@ def applyStateToOriginal(self):
             self._cachedReactorCoreParamData = {}
             self._clearStateOnReactor(self._sourceReactor, cache=True)
             self._mapStateFromReactorToOther(
-                self.convReactor, self._sourceReactor, mapNumberDensities=True
+                self.convReactor, self._sourceReactor, mapNumberDensities=False
```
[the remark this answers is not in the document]
mgjarrett: The number densities are updated using `setNumberDensitiesFromOverlaps` which calls some specific functions that update number densities on all the components. The logic for updating number densities is a little different than the other parameters, so it probably makes sense to keep them separate.

armi/reactor/converters/uniformMesh.py
```diff
@@ -418,9 +433,8 @@ def setAssemblyStateFromOverlaps(
             assem that has the state
         destinationAssembly : Assembly
             assem that has is getting the state from sourceAssembly
-        blockParamNames : List[str], optional
-            A list of block parameter names to map between the source assembly and
-            the destination assembly.
+        blockParamNames : list
+            List of block parameter names to be updated
```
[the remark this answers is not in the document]
mgjarrett: Yeah the old description is better. I'm pretty sure what happened is that at some point I had deleted `blockParamNames` entirely, and then brought it back. So the less descriptive version was created from scratch, not an attempt to improve the original description.

armi/reactor/converters/uniformMesh.py
mgjarrett: That's a good point. But theoretically if you're only interested in geometry and not parameters, it might be valid to instantiate the base class and use it?

armi/reactor/converters/uniformMesh.py
mgjarrett: I initially thought there might be some parameters that we want to map in all cases that would be defined in the base class implementation. But I can't think of any good examples, so... ?

armi/reactor/converters/uniformMesh.py
mgjarrett: My understanding of the caching is basically the same as yours.

As for whether the caching is necessary, I would break it down into the two directions:

- Non-uniform source, uniform destination assembly: When the destination assembly is the uniform assembly, it's being created from the source assembly, so it shouldn't have any pre-existing parameters that need to be cached.
- Uniform source, non-uniform destination: When the destination assembly is the non-uniform assembly, parameter caching would ensure that we don't erase any of the parameters in `blockParamNames` on the non-uniform assembly by overwriting with null values from the source assembly. But, if we have defined the `blockParamNames` to be mapped back correctly, then we should only be mapping back parameters that have been set on the uniform assembly. If this `setAssemblyStateFromOverlaps` is trying to map back parameters that haven't been set, then either the `_setParamsToUpdate` was not defined correctly or maybe some analysis performed on the uniform reactor failed. `_applyCachedParamValues` would paper over that error by just retaining whatever value the non-uniform assembly already had for that parameter, but this would create a "silent" bug that would be very difficult to track down (e.g., you have a `b.p.power` in the database, but you have no way of verifying whether it was mapped back correctly or just retained by caching).

That's my take on it, but it's complicated and I'm definitely curious to hear the opinions of others. I think the short summary is that parameter caching would not be performing any detectable function when everything is working correctly. It would only act when there is a bug in the code or an error during execution, and it would effectively cover up the error without giving any warning or error message.

mgjarrett:
> After this is merged, though, there is still work to do in our internal plugins to move the application/calculation of cumulative parameters outside of the geometry conversion entry/exit. For instance, I can say for certain that `detailedDpaRate` is calculated right now when the geometry is still converted. Likely the same for many of the `cumulative` parameters. So we'll have to hunt all of those down and update them, if necessary.

I am aware of `detailedDpa`, and that looks like it's going to somewhat difficult to address. I don't know of anything else in the downstream plugins, though.

> For `cumulative` parameters that are being calculated within the framework itself, I would argue it might make sense to update those as part of this PR (if any of them exist).

I don't know of any parameter accumulation within the framework itself that needs to be updated. The most obvious cumulative parameter is depletion/burnup, and those are calculated on the non-uniform assembly. Aside from depletion/residence/burnup, fluence, DPA (mentioned above), I can't think of any other obvious cumulative parameters.

armi/reactor/converters/uniformMesh.py
```diff
@@ -792,12 +734,18 @@ class NeutronicsUniformMeshConverter(UniformMeshGeometryConverter):
...
-    BLOCK_PARAM_MAPPING_CATEGORIES = [
-        parameters.Category.detailedAxialExpansion,
-        parameters.Category.multiGroupQuantities,
-        parameters.Category.pinQuantities,
-    ]
+    REACTOR_PARAMS_TO_MAP = {
+        "in": [parameters.Category.neutronics],
+        "out": [parameters.Category.neutronics],
+    }
+    BLOCK_PARAMS_TO_MAP = {
```
keckler, line 741: I am wondering why the name of this dict was changed? To me, the original name `BLOCK_PARAM_MAPPING_CATEGORIES` was more descriptive of what is actually contained in the dict.

The new name `BLOCK_PARAMS_TO_MAP` indicates to me that it contains parameter names, but this is not the case.

armi/reactor/converters/uniformMesh.py
```diff
@@ -792,12 +734,18 @@ class NeutronicsUniformMeshConverter(UniformMeshGeometryConverter):
     back to the source reactor.
```
keckler, line 734: I think this note might be out of date? Or maybe just flat-out wrong? Mapping the parameters from and then back to a source reactor is exactly what we do _not_ want to do. 

Does this need to be updated?

armi/reactor/converters/uniformMesh.py
```diff
@@ -867,6 +838,64 @@ def _mapStateFromReactorToOther(
         self._cachedReactorCoreParamData = {}
 
 
+class GammaUniformMeshConverter(NeutronicsUniformMeshConverter):
+    """
+    A uniform mesh converter that specifically maps gamma parameters.
+    """
```
keckler, line 844: I think it would be extremely useful to outline in the docstring for this class as well as the `NeutronicsUniformMeshConverter` the conditions which determine which parameters are mapped in each direction. 

I foresee a case where a plugin is mistakenly calculating a parameter that is not mapped back to the source reactor, and then the user/developer would be left wondering why their updated value did not get reflected back on their original reactor.

Making it clear in the docstring that certain parameters are not mapped to/from would help to guard against that kind of situation.

keckler:
> I don't know of any parameter accumulation within the framework itself that needs to be updated. The most obvious cumulative parameter is depletion/burnup, and those are calculated on the non-uniform assembly. Aside from depletion/residence/burnup, fluence, DPA (mentioned above), I can't think of any other obvious cumulative parameters.

I agree with this sentiment. I just scanned through the framework parameters and couldn't find any which are cumulative that would also be mapped in the first place. There are some things in the `fuelPerformance` package that are cumulative (gasReleaseFraction, bondRemoved, cladWastage, totalCladStrain), but I guess those won't be mapped in the first place because they don't satisfy the category qualifications.

So we're probably good there!

keckler:
I pulled this PR into my ARMI app, reran some extensive calculations, and then looked at the calculated DPAs. 

I can confirm that this PR resolves the issues I was previously having with DPA possibly decreasing as time increases. Obviously that was non-physical, and these changes resolve that such that we are now getting physically-possible results for DPA.

I can also add another data point to add confidence to this change. The model that I ran is a benchmark of real reactor operations, and I was previously getting anomalously low reactivities in later cycles. I had suspected that this was due to the numerical diffusion of the uniform mesh generator, and it turns out that with this fix, core reactivities line up much more closely than before. 

So in my book, the changes in this PR are a resounding success. I approve!

mgjarrett:
#985

armi/reactor/parameters/parameterDefinitions.py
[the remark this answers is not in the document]
john-science: I would love a docstring or description in the /docs/ somewhere that explains to people what to use these two new categories for.

I mean, _I_ see your description above about going into/out of the uniform mesh.  But we want people who come to ARMI in a year to be able to understand this, right?

armi/reactor/converters/uniformMesh.py
john-science: My major concern with preserving the old, caching solution is backwards compatibility for our downstream projects.

But it doesn't appear well enough documented that I'm thrilled with having to support it forever. Michael's above description matches what I see in the code. But the original developer could have saved us several hours of digging if it had been better documented.

Let's do better!

john-science:
> I pulled this PR into my ARMI app, reran some extensive calculations, and then looked at the calculated DPAs.

So... above you were concerned that "there is still work to do in our internal plugins to move the application/calculation of cumulative parameters outside of the geometry conversion entry/exit".  

Are you concerns with that over?

armi/reactor/converters/tests/test_uniformMesh.py
```diff
@@ -28,6 +28,29 @@
...
+
+class TestConverterFactory(unittest.TestCase):
+    def setUp(self):
+        self.o, self.r = test_reactors.loadTestReactor(
+            inputFilePath=os.path.join(TEST_ROOT, "detailedAxialExpansion"),
+        )
+        self.dummyOptions = DummyFluxOptions()
+
+    def test_converterFactory(self):
+        print(self.dummyOptions)
```
john-science, line 44: Can we remove this `print()` statement?  We try not to sprinkle `print`s throughout ARMI.

mgjarrett: ah yes, good catch

keckler:
> Are you concerns with that over?

Yes, I am sufficiently convinced after both looking though our internal code and testing this PR that we are in the clear on that concern.

armi/reactor/parameters/parameterDefinitions.py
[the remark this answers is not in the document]
mgjarrett: I added some docstrings to describe most of the parameter categories in this class.

armi/reactor/converters/uniformMesh.py
mgjarrett: I never saw this comment somehow.

No good reason for it other than the order in which development happened. I started by trying to create a list of parameter names, not categories, so I changed the attribute name. And then I went back to categories.

Another good question is why is this snake case? I have no idea. I should probably change it to `blockParamMappingCategories`.

armi/reactor/converters/uniformMesh.py
mgjarrett: Yes it was just the old doc string, I had not updated it

armi/reactor/converters/uniformMesh.py
mgjarrett: Somehow I missed several of these comments the first time I looked through.

I added a lot of detail to the base class `UniformMeshGeometryConverter` docstring and to the derived class docstrings. I could definitely see that being a really frustrating error to contend with, so I'll try to put a clear warning in the docstring of each converter.

LMikeH:
I'm not sure why, but I'm getting an error in JOYO that I thought I fixed previously. I looked to make sure my previous changes were still there and they are. I don't know if any of you have seen this error before:

  File "\\albert\home\mhuang\site-packages\nala\__main__.py", line 26, in main
    code = ArmiCLI().run()
  File "\\albert\home\mhuang\site-packages\framework\armi\cli\__init__.py", line 202, in run
    return self.executeCommand(args.command, args.args)
  File "\\albert\home\mhuang\site-packages\framework\armi\cli\__init__.py", line 231, in executeCommand
    return cmd.invoke()
  File "\\albert\home\mhuang\site-packages\framework\armi\cli\run.py", line 33, in invoke
    inputCase.run()
  File "\\albert\home\mhuang\site-packages\framework\armi\cases\case.py", line 367, in run
    o.operate()
  File "\\albert\home\mhuang\site-packages\framework\armi\operators\operatorMPI.py", line 74, in operate
    Operator.operate(self)
  File "\\albert\home\mhuang\site-packages\framework\armi\operators\operator.py", line 320, in operate
    self._mainOperate()
  File "\\albert\home\mhuang\site-packages\framework\armi\operators\operator.py", line 329, in _mainOperate
    keepGoing = self._cycleLoop(cycle, startingCycle)
  File "\\albert\home\mhuang\site-packages\framework\armi\operators\operator.py", line 345, in _cycleLoop
    halt = self.interactAllBOC(self.r.p.cycle)
  File "\\albert\home\mhuang\site-packages\framework\armi\operators\operator.py", line 552, in interactAllBOC
    return self._interactAll("BOC", activeInterfaces, cycle)
  File "\\albert\home\mhuang\site-packages\framework\armi\operators\operator.py", line 423, in _interactAll
    halt = halt or interactMethod(*args)
  File "\\albert\home\mhuang\site-packages\framework\armi\utils\codeTiming.py", line 55, in time_wrapper
    return_value = func(*args, **kwargs)
  File "\\albert\home\mhuang\site-packages\framework\armi\physics\neutronics\latticePhysics\latticePhysicsInterface.py", line 108, in interactBOC
    self.updateXSLibrary(cycle)
  File "\\albert\home\mhuang\site-packages\framework\armi\physics\neutronics\latticePhysics\latticePhysicsInterface.py", line 129, in updateXSLibrary
    blockList=representativeBlocks, xsLibrarySuffix=self._getSuffix(cycle)
  File "\\albert\home\mhuang\site-packages\plugins\terrapower-partisn\terrapower\physics\neutronics\partisn\mc2PartisnInterface.py", line 78, in computeCrossSections
    executer.run()
  File "\\albert\home\mhuang\site-packages\plugins\terrapower-partisn\terrapower\physics\neutronics\partisn\partisnExecuters.py", line 417, in run
    output = globalFluxInterface.GlobalFluxExecuter.run(self)
  File "\\albert\home\mhuang\site-packages\framework\armi\physics\executers.py", line 187, in run
    self._performGeometryTransformations()
  File "\\albert\home\mhuang\site-packages\plugins\terrapower-partisn\terrapower\physics\neutronics\partisn\mc2PartisnExecuters.py", line 88, in _performGeometryTransformations
    self._computeFineGroupXS()
  File "\\albert\home\mhuang\site-packages\plugins\terrapower-partisn\terrapower\physics\neutronics\partisn\mc2PartisnExecuters.py", line 99, in _computeFineGroupXS
    blockList=self._xsgm.representativeBlocks.values()
  File "\\albert\home\mhuang\site-packages\plugins\terrapower-partisn\terrapower\physics\neutronics\partisn\mc2PartisnInterface.py", line 117, in computeFineGroupXs
    purgeFP=False,
  File "\\albert\home\mhuang\site-packages\plugins\terrapower-mc2\terrapower\physics\neutronics\mc2\mc2Interface.py", line 236, in _generateXsLibrary
    baseList, xsLibrarySuffix, blockList, writers
  File "\\albert\home\mhuang\site-packages\framework\armi\physics\neutronics\latticePhysics\latticePhysicsInterface.py", line 300, in generateLatticePhysicsInputs
    fromWriter = writer.write()
  File "\\albert\home\mhuang\site-packages\plugins\terrapower-mc2\terrapower\physics\neutronics\mc2\mc2Writers.py", line 411, in write
    self._writeComposition(f)
  File "\\albert\home\mhuang\site-packages\plugins\terrapower-mc2\terrapower\physics\neutronics\mc2\mc2Writers.py", line 601, in _writeComposition
    self._writeSingleComposition(fileObj, compNumber, component)
  File "\\albert\home\mhuang\site-packages\plugins\terrapower-mc2\terrapower\physics\neutronics\mc2\mc2Writers.py", line 606, in _writeSingleComposition
    allNuclides = self._getAllNuclideObjects(component)
  File "\\albert\home\mhuang\site-packages\framework\armi\physics\neutronics\latticePhysics\latticePhysicsWriter.py", line 187, in _getAllNuclideObjects
    nucs, fissProds = self._getAllNuclidesByCategory(component)
  File "\\albert\home\mhuang\site-packages\framework\armi\physics\neutronics\latticePhysics\latticePhysicsWriter.py", line 204, in _getAllNuclidesByCategory
    dfpDensities = self._getDetailedFPDensities()
  File "\\albert\home\mhuang\site-packages\framework\armi\physics\neutronics\latticePhysics\latticePhysicsWriter.py", line 370, in _getDetailedFPDensities
    dfpDensitiesByName = lfpCollection.getNumberDensities(self.block)
AttributeError: 'NoneType' object has no attribute 'getNumberDensities'

keckler:
@LMikeH I'll message you offline.

john-science (approved):
I think this is a good solution, in that it solves a semi-complicated problem.

The solution is as complicated as the problem.  But sometimes there is just no way around that.

As long as we are all convinced by the documentation we've put in place to explain it to future generations, I'm happy.

PR #993, opened 2022-11-22
Add ability to update grid plate pitch via temperature

albeanth:
## Description
This PR adds the ability to calculate and update the grid plate pitch based on temperature. It has two primary assumptions:
1. The grid plate material is uniform throughout the core. 
2. The temperature used to update the pitch is isothermal. 

ARMI allows for variability in block pitch, but the "2D spatial grid" does not. So one could in theory compute an assembly dependent grid plate pitch based on a nonuniform grid plate temperature distribution and assign it to the corresponding grid plate blocks. However, you would run into the problem of having the variable assembly pitch being disconnected from the 2D spatial grid.  

---

## Checklist


armi/reactor/reactors.py
```diff
@@ -2129,6 +2129,48 @@ def growToFullCore(self, cs):
 
         return converter
 
+    def calculateNewGridPlatePitchFromTemp(self, cs, isoTemp: float) -> float:
+        """calculate a grid plate pitch using a new bulk ave core inlet temperature in C"""
+        # retrieve the first grid plate (all the same material, so just get the first)
+        gridPlateBlock = self.getFirstBlock(blockType=Flags.GRID_PLATE)
```
keckler, line 2135: Since there is nothing currently enforcing that all GRID_PLATE blocks have the same material, it might be worthwhile to include a check here and throw a warning, if not.

armi/reactor/reactors.py
```diff
@@ -2129,6 +2129,48 @@ def growToFullCore(self, cs):
...
+        Notes
+        -----
+        1) all grid plate blocks are the same material
+        2) all grid plate blocks are isothermal and equal to cs["Tin"]
+        """
+        newGridPlatePitch = self.calculateNewGridPlatePitchFromTemp(cs, isoTemp)
+        # update the pitch
+        self.setPitchUniform(newGridPlatePitch)
+        # update the bulk core average inlet temperature
+        cs["Tin"] = isoTemp
```
keckler, line 2172: It is perhaps not desirable to set Tin to this value. I could see a likely scenario when a user just wants to perturb the pitch to see the reactivity effect without messing with other calcs (i.e. TH)

albeanth: How about an optional parameter to update the inlet temp? Set the default to False.

albeanth: https://github.com/terrapower/armi/pull/993/commits/191f193736fd82a9ad91ae6e3d7080156c86dd08

armi/reactor/reactors.py
albeanth: https://github.com/terrapower/armi/pull/993/commits/191f193736fd82a9ad91ae6e3d7080156c86dd08

armi/reactor/reactors.py
```diff
@@ -2129,6 +2129,77 @@ def growToFullCore(self, cs):
...
 
+    def checkGridPlateMaterialConsistency(self):
+        """check that the grid plates in the core are the same material"""
+        gridPlateBlocks = self.getBlocks(Flags.GRID_PLATE)
+        for ib, b in enumerate(gridPlateBlocks):
+            gridPlateComponent = b.getComponent(Flags.GRID_PLATE)
+            gridPlateMaterial = gridPlateComponent.material
+            if ib == 0:
+                refGridPlateMat = gridPlateMaterial
+                continue
```
jakehader, line 2140: Maybe just get this out of the loop and have `refGridPlateMat = self.getFirstBlock(Flags.GRID_PLATE).getComponent(Flags.GRID_PLATE).material`?

armi/reactor/reactors.py
```diff
@@ -2129,6 +2129,77 @@ def growToFullCore(self, cs):
...
+            an isothermal temperature in C to apply to cs["Tin"]
+        """
+        self.checkGridPlateMaterialConsistency()
+        # retrieve the first grid plate (all the same material, so just get the first)
+        gridPlateBlock = self.getFirstBlock(blockType=Flags.GRID_PLATE)
+        # retrieve common grid plate component, material, temp, and pitch
+        gridPlateComponent = gridPlateBlock.getComponent(Flags.GRID_PLATE)
+        gridPlateMaterial = gridPlateComponent.material
+        currentGridPlatePitch = self.getAssemblyPitch()
+        currentGridPlateTemp = cs["Tin"]
```
jakehader, line 2166: Is this true? Shouldn't this be the grid plate component temperature? Might it better if the initial grid plate temperature was just set on the blueprints as `Tinput` and then go from there?

albeanth: Discussed offline and addressed https://github.com/terrapower/armi/pull/993/commits/01ce852e7d7759ae463e6f3c612e031db1f77717

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/993/commits/01ce852e7d7759ae463e6f3c612e031db1f77717

armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -175,6 +175,9 @@ def _constructAssembly(self, cs, blueprint):
         aziMeshPoints = self.azimuthalMeshPoints or 1
         a.p.AziMesh = aziMeshPoints
 
+        # TODO: How can this be done better?
+        a.p.THcoolantInletT = cs["Tin"]
```
jakehader, line 179: It might be nice to have an assembly method in `assemblies.py` that is something like:

```
def inletTemperatureSetter(self, temperatureInC):
    a.p.THcoolantInletT = temperatureInC
```

In this way we can reuse the temperature setting functionality without directly setting the parameter.

armi/reactor/reactors.py
```diff
@@ -2129,9 +2129,83 @@ def growToFullCore(self, cs):
...
+        newGridPlatePitch = self.calculateNewGridPlatePitchFromTemp(inletTemp)
+        # update the pitch
+        self.setPitchUniform(newGridPlatePitch)
+        if updateInletTemp:
+            self.updateInletTemp(inletTemp)
+
+    def updateInletTemp(self, inletTemp):
+        """update core bulk average inlet temp for all assemblies"""
+        for a in self.getAssemblies():
+            a.p.THcoolantInletT = inletTemp
```
jakehader, line 2200: See above comment on having an assembly method to reuse.

jakehader (changes requested):

armi/reactor/reactors.py
albeanth: Addressed https://github.com/terrapower/armi/pull/993/commits/abfaede115874ce52ead2cf2f222640f0d5d4d12

armi/reactor/blueprints/assemblyBlueprint.py
albeanth: Discussed offline some and addressed here
 https://github.com/terrapower/armi/pull/993/commits/abfaede115874ce52ead2cf2f222640f0d5d4d12

jakehader (commented):
Thanks for the updates @albeanth - I will review soon.

armi/reactor/blueprints/assemblyBlueprint.py
jakehader: The implication of setting this should be very clear. For example, what do we do if the grid plate dimensions and the pitch of the assembly defined in the blueprints is not in sync with this temperature assignment? We now have two ways of setting up the reactor, one mechanism being that we do so with all cold dimensions and the other being setting up with hot dimensions for the blocks. The pitch of the assembly has typically been setup to assume that we are talking about hot dimensions (i.e., already expanded) at the inlet temperature (`cs["Tin"]`), but if this isn't the case then does this assignment made sense?

albeanth: > what do we do if the grid plate dimensions and the pitch of the assembly defined in the blueprints is not in sync with this temperature assignment?

By default, the grid plate dimensions and temperature of the grid plate should always be in sync. But as you point out, there is a chance that the grid plate dims/temp could be out of sync with the assembly inlet temperature and pitch, `cs["Tin"]` and `core.spatialGrid._unitSteps`, respectively. This seems problematic to me and an excellent point to bring up - not to mention unphysical (the grid plate and inlet temperature should probably be in sync). We should likely add checks to ensure alignment between the grid plate temp and assembly inlet temp at BOL and whenever the grid plate pitch is updated.

onufer: the intercoolant outer pitch is the dimension that determines the lattice.

It is defined at the hot temperature. even if the intercoolant changes temperature the OP should not change dimension even if the intercoolant temp changes. it should only change dimension if the grid plate temp changes. 

there is no way to check if they are consistent because the cold grid plate lattice is not defined anywhere.

jakehader: I think it's reasonable that the user should keep these consistent on initialization. Maybe just a runLog would help us keep a check on this. One other suggestion could be to define the cold grid plate pitch as a setting that could then be verified against the temperature assignment of Tin and the defined assembly pitch. 

I am fine with it being a an expectation that the user defines this correctly but some communication to the user in the stdout that says "hey, I'm initializing the core pitch to be X with an associated temperature of Y". This is something that we could check. Also, one thing that might help is a lot is to add a core parameter for pitch. It's actually not clear to me if loading a snapshot case after grid plate dimension changes will stick. I think they would because this would be reflected in dimension changes to components but I am not certain that I understand how this works.

Sorry for the non-answer, trying to think through it aloud and get more feedback before we decide what makes the most sense

armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -175,6 +175,9 @@ def _constructAssembly(self, cs, blueprint):
         aziMeshPoints = self.azimuthalMeshPoints or 1
         a.p.AziMesh = aziMeshPoints
 
+        # TODO: How can this be done better?
```
john-science, line 178: Let's not commit a `TODO` to the repo.

armi/reactor/reactors.py
```diff
@@ -2129,9 +2129,103 @@ def growToFullCore(self, cs):
...
+        ----------
+        inletTemp : float
+            an isothermal core bulk average inlet temperature in C
+        updateInletTemp: bool, optional
+            optional boolean to determine if the bulk average core inlet temperature should be updated to inletTemp
+
+        Notes
+        -----
+        1) all grid plate blocks are the same material
+        2) all grid plate blocks are isothermal and equal to inletTemp
```
john-science, line 2189: SUPER minor thought.

Can we put the word "assumption" in these notes?  I just want to make it clear to some stranger 5 years from now that these are our simplifying assumptions.  

Sorry to be nit-picky, I'm just want it to be transparent.

john-science:
It's a reasonably minor change, that I think add some utility to our `Flags.GRID_PLATE` logic.

If anything, I think ARMI is pretty light on `GRID_PLATE` tools people might want.

Only minor comments.

armi/reactor/blueprints/assemblyBlueprint.py
albeanth: The changes in this file have been removed altogether with so no more `TODO`. 
https://github.com/terrapower/armi/pull/993/commits/9107822341523e8c5f8e27f6e77f4d5e01fdca98

armi/reactor/reactors.py
albeanth: Great idea! I like it. 
https://github.com/terrapower/armi/pull/993/commits/43dea385bc0d24d075b241ab1e6399e1d9956a1f

PR #1001, opened 2022-11-29
Moving hyper-specific TH parameters out of ARMI

john-science:
## Description

After working with @andfranklin , I have determined that we should move the following 23 parameters out of ARMI and into a downstream Thermal Hydraulics Plugin.  The major motivation here is to help centralize the TH teams work. As these parameters are a bit too specific to be generally useful to the community. And more importantly, the TH team is putting a lot of effort into tweaking and re-aligning these parameters, which really makes it sound like these parameters should belong in their code.

- pinPeakingStdDev
- TH0SigmaFuelCenterlineT
- TH0SigmaOutletT
- TH2SigmaFuelCenterlineT
- TH2SigmaOutletT
- TH3SigmaFuelCenterlineT
- TH3SigmaOutletT
- THaveCoolantVel
- THaveInletVel
- THaveOutletVel
- THcoldChannel
- THcoldChanTemp
- THcoldPin
- THdeltaPBundle
- THdeltaPFormLoss
- THdeltaPGrav
- THdeltaPNoGrav
- THhotPin
- THinletDynamicPressure
- THmaxCoolantVel
- THmixingLength
- THorificeSetting
- THpeakingStdDev
- THradialPeakingFactor 

---

## Checklist

### DIFF
--- a/armi/physics/thermalHydraulics/parameters.py
+++ b/armi/physics/thermalHydraulics/parameters.py
@@ -46,19 +46,6 @@ def _getBlockParams():
     with pDefs.createBuilder(
         default=0.0, location=ParamLocation.AVERAGE, categories=["thermal hydraulics"]
     ) as pb:
-        pb.defParam(
-            "THaveInletVel",
-            units="m/s",
-            saveToDB=True,
-            description="Average fluid velocity taken at the inlet of the block.",
-        )
-
-        pb.defParam(
-            "THaveOutletVel",
-            units="m/s",
-            saveToDB=True,
-            description="Average fluid velocity taken at the outlet of the block.",
-        )
 
         pb.defParam(
             "THhotChannelCladMidwallT",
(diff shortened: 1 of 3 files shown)


opotowsky (approved):
No complaints here!

keckler (approved):

armi/physics/thermalHydraulics/parameters.py
```diff
@@ -134,13 +93,6 @@ def _getBlockParams():
             categories=["thInterface"],
         )
 
-        pb.defParam(
-            "THdeltaPBundle",
-            units="Pa",
-            description="Pressure difference in a bundle, including contributions from friction, acceleration, and gravity",
-            categories=["thInterface"],
-        )
-
```
jakehader, line 143: Arguably we can remove the other TH parameters too because these are also specific to our internal plugins. While these can be used by other folks it's more likely that someone who makes a TH plugin would just reimplement their own parameters. 

I think that in general we probably over committed to the number of parameters that are available on the vanilla ARMI framework and they probably don't have a lot of general use cases.

john-science: There might be other TH parameters we can use.

But they will all be more complicated, because they are used by multiple downstream projects.

These particular TH parameters are the easiest for me to vet, because they are only used by the one downstream repo.

Low hanging fruit.

PR #1005, opened 2022-12-01
Fix DB write out for coupling

albeanth:
## Description
This PR fixes a bug for how databases were written when using tight coupling. Previously, databases were written out on the `interactEveryNode` and then never written out/updated after the coupling iterations. So in effect, the updated values were never written to the DB. This PR resolves this by doing nothing on database3.py::Database::interactEveryNode if tight coupling is requested. Then only after the tight coupling iterations are complete is the database written out. 

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/databaseInterface.py
+++ b/armi/bookkeeping/db/databaseInterface.py
@@ -144,13 +144,20 @@ class DatabaseInterface(interfaces.Interface):
         Write to database.
 
         DBs should receive the state information of the run at each node.
+
+        Notes
+        -----
+        - if tight coupling is enabled, the DB will be written in operator.py::Operator::_timeNodeLoop
+          via writeDBEveryNode
         """
         if self.o.cs["numCoupledIterations"]:
+            # h5 cant handle overwriting so we skip here and write once the tight coupling loop has completed
             return
         self.writeDBEveryNode(cycle, node)
 
     def writeDBEveryNode(self, cycle, node):
-        """docstring"""
+        """write the database at the end of the time node"""
+        # skip writing for last burn step since it will be written at interact EOC
         if node < self.o.burnSteps[cycle]:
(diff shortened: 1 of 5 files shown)


armi/bookkeeping/db/databaseInterface.py
```diff
@@ -145,7 +145,12 @@ def interactEveryNode(self, cycle, node):
 
         DBs should receive the state information of the run at each node.
         """
-        # skip writing for last burn step since it will be written at interact EOC
```
onufer, line 148: please copy this comment over

onufer (changes requested):

armi/bookkeeping/db/databaseInterface.py
```diff
@@ -145,7 +145,12 @@ def interactEveryNode(self, cycle, node):
...
+        if self.o.cs["numCoupledIterations"]:
```
onufer, line 148: add comment saying that h5 cant handle writing more than once so we only write at the end

albeanth: Addressed https://github.com/terrapower/armi/pull/1005/commits/661e26a9a2a148719549ff717ce4fd0b745b2ef8

armi/bookkeeping/db/databaseInterface.py
albeanth: Addressed https://github.com/terrapower/armi/pull/1005/commits/661e26a9a2a148719549ff717ce4fd0b745b2ef8

onufer (approved):

john-science:
You can see [here](https://coveralls.io/builds/54797142/source?filename=armi%2Foperators%2Foperator.py#L384) that two of your new lines of code aren't covered by your unit tests.  Thus the coverage failure.

john-science (approved):

Issue #450, opened 2021-10-20
Remove references to geom files in the code


youngmit:
Back in the day, we used to specify core layout in what was called a facemap or geom file. While we retain the code that supports this input mechanism (see https://github.com/terrapower/armi/blob/master/armi/reactor/systemLayoutInput.py), it is mostly there to enable the migration performed here: https://github.com/terrapower/armi/blob/190c0caccfd36a0442118a1ae666c2a20c52a4d1/armi/reactor/blueprints/__init__.py#L524  

Instead we recommend using grid blueprints instead. However, when these were completely separate input files, they became part of some major ARMI function signatures where input stuff gets passed around, usually as `geom`. We should try to remove as many of these as we can.

Actually searching around, I'm not seeing as many as I remember (go us!), but a glaring example is in the Case class: https://github.com/terrapower/armi/blob/190c0caccfd36a0442118a1ae666c2a20c52a4d1/armi/cases/case.py#L73

john-science:
@drewejohnson @keckler @jakehader Would any of you notice or care if ARMI finally dropped all support for the old Geometry XML files?

My assumption is that everyone is using the Blueprint YAML file. I _assume_ I can nuke all references and support for the old Geometry XML files and it won't bother anyone.

Tell me if I'm wrong.

keckler:
In fact, I would love it if XML support was dropped.

john-science:
The code changes for this will be pretty minor.  I just want to note here that these files will need to be re-written in YAML and moved into a blueprints file:

* armi/tests/geom1Assem.xml
* armi/tests/geom.xml
* armi/tests/ThRZGeom.xml
* armi/tests/trz_geom.xml
* armi/tests/zpprTestGeom.xml

Though, maybe some/all of them will be straight deletable.

keckler:
So, can you remind me the status of this?

I know we removed support for XML case settings in #612 .

But did we ever remove support for the XML geom file? I know that [Jake said he would be fine with it](https://github.com/terrapower/armi/issues/608#issuecomment-1073043021), but did it ever actually happen? I can't find any record of that being finally deprecated.

john-science:
I would _love_ to remove geom files entirely.  As a bonus, to finally remove the last XML support inputs in ARMI.

There are still a couple of features blueprint files don't support that need translating over.  I can't find my notes at the moment, but if memory serves, ARMI only supports ThetaRZ geometries through geomFiles, and not blueprint YAML files.

[...]

PR #1009, opened 2022-12-05
added modern RZ theta unittest for ARMI

sombrereau:
## Description

The ability to generate RZ theta ARMI models systematically from other interfaces. Is not functional. Therefore, a better unittest needs to be developed to lock in the capability to initialize RZ theta ARMI models defined with radial segments and volume fractions.

---

## Checklist


john-science:
I am pretty uncertain about this.

I have been told, several times, that we only support RZ Theta blueprints in the old Geom XML file format, and not in the YAML format.  That is actually the only reason we haven't fully [dropped support for the old geom files](https://github.com/terrapower/armi/issues/450).

sombrereau:
I couldn't figure out how to go through the initialization with the geom file format, this passes with tests to check that perform operations that need to have the right block geometry in order to pass the unit tests.

jakehader:
@sombrereau - Godiva is a sphere model. Does it make sense here?

sombrereau:
@jakehader -- This input is demonstrating the capability of ARMI to model 3D features across a RZ theta geometry.

the central block (theta:0, radius:0, axial:2) has a volume fraction of HEU of 1 and in the upper corner (theta:0, radius:3, axial:4) the volume fraction is only 4.6% such that you define a pixelated "sphere".

I would be happy to entertain other RZ theta geometry benchmarks defined with this kind of strategy though. I just happened to have a Godiva model lying around.

jakehader:
Got it! I think it would be cool if the RZTheta grids could be updated to also just support 1D models, like sphere or slab but I see what you're getting at. Want to see if this model runs as-is with PARTISN and check k-eff?

jakehader:
> I couldn't figure out how to go through the initialization with the geom file format, this passes with tests to check that perform operations that need to have the right block geometry in order to pass the unit tests. I looked back and my team has been previously using yaml geomFiles 9 months ago, but when I looked at this a few weeks ago I could no longer navigate through the reactor initialization.

I think that the system for R, RZ, and RZT needs some love and that would a good PR as a predecessor to this in my opinion. See @john-science's comment on https://github.com/terrapower/armi/issues/450#issuecomment-1271951550

john-science:
I definitely agree with @jakehader that we need better support for R, Z, and RZTheta geometries in our YAML formats.

armi/tests/Godiva-blueprints.yaml
```diff
@@ -0,0 +1,733 @@
...
+        outer_radius: 9.0
+        inner_radius: 6.002
+        mult: 0.046154800802469254
+      compliment:
+        shape: RadialSegment
+        material: Void
+        Tinput: 0.0
+        Thot: 0.0
+        outer_theta: 0.7853981633974483
+        height: 3.5000000000000004
```
john-science, line 566: This screams "Python rounding error" to me.

john-science:
@ntouran What do you think about this?  We could add in a couple of other semi-real-world examples.  (We mentioned the Demon Core yesterday.)  And we could just store them all in the same folder.

john-science:
@sombrereau Seeing all these "void" really highlights that ARMI doesn't have an "air" material.  Nor even a "pure N gas" material, which would at least be 78% correct.  Thanks!

armi/tests/Godiva.armi.unittest.yaml
```diff
@@ -0,0 +1,24 @@
+metadata:
+  version: 0.2.5
```
john-science, line 2: I believe this has to be:

>   version: uncontrolled

Or the test will start to slide out of date as we bump the ARMI version number.

We always want to run this test against the latest ARMI version.

armi/tests/Godiva-blueprints.yaml
sombrereau: I moved to 3.5

sombrereau:
> 

I added a new issue. I might add something simple ... at sometime.

armi/tests/Godiva.armi.unittest.yaml
sombrereau: updated to uncontrolled version

john-science (approved):

PR #1014, opened 2022-12-09
Fixing bug with fp interface not in geomConverter

john-science:
## Description

There was a bug where an `reactor.Operator` was still being referenced in the RZ-Theta Geometry Converter.  Happily, this backwards import was never necessary to begin with, because all we used it for was getting the Lumped Fission Products with the Settings file.

While fixing this bug, I also noticed that we were loading an operator and going through all the work to re-generate the lumped fission products INSIDE A LOOP.  So, yeah, I pulled that out of the loop. Much more efficient.

---

## Checklist


john-science:
The unit tests are failing because GitHub changed how they deal with MPI?

Sigh...

Issue #1011, opened 2022-12-07
Misleading docstring for component.setDimension()


jasonbmeng:
Misleading docstring on `setDimension` method caused some minor confusion.

Testing the method indicates the `cold` flag sets the dimension of the component at cold temperature to `val`.
It does not set the dimension of the component at cold temperature such that the thermally expanded hot dimension matches `val`

https://github.com/terrapower/armi/blob/b1658e4a733751213c452b22661830e5ce2f9d9a/armi/reactor/components/component.py#L794-L796

john-science:
So not so much misleading as wrong?

john-science:
So, the only use of the `cold` flag on `Component.setDimension()` is this one line:

https://github.com/terrapower/armi/blob/b1658e4a733751213c452b22661830e5ce2f9d9a/armi/reactor/components/component.py#L809

So, the docstring doesn't _sound_ too far off. 

What do you think it _should_ say?

jasonbmeng:
The docstring for the `cold` flag should just say `If True sets the component cold dimension to the specified value.`

Currently, the flag description implies that `c.setDimension("op", 30.0, cold=True)` will set the cold dimension to a _thermally contracted_ value such that the hot dimension is 30.0 which is completely different behavior from what it actually does. It sets the component cold dimension to be 30.0 which is correct behavior imo. So no changes requested in the method, just the docstring.

john-science:
@jasonbmeng PR opened over [here](https://github.com/terrapower/armi/pull/1019).

PR #1019, opened 2022-12-11
Fixing docstring for setDimension

john-science:
## Description

This was a documentation upgrade suggested by @jasonbmeng in Issue #1011 

---

## Checklist

### DIFF
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -34,7 +34,6 @@ from armi.utils.units import C_TO_K
 from armi.materials import void
 from armi.nucDirectory import nuclideBases
 from armi import materials
-from armi.reactor.flags import Flags
 
 COMPONENT_LINK_REGEX = re.compile(r"^\s*(.+?)\s*\.\s*(.+?)\s*$")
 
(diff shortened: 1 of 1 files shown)


armi/reactor/components/component.py
```diff
@@ -792,8 +791,7 @@ def setDimension(self, key, val, retainLink=False, cold=True):
             If True, the val will be applied to the dimension of linked
             component which indirectly changes this component's dimensions.
         cold : book, optional
```
keckler, line 793: While we're at it, may as well fix the typo "book" to "bool"

keckler (approved):
I think this is a good change.

I would also propose that the way this method is written is kinda goofy. The logic makes it so we always calculate `expansionFactor`, whether or not it is used. This is (1) wasteful and (2) restricts us to only setting dimensions on components with materials that have the thermal expansion parameter defined.

Not super related to this PR, but just noticed it in the process of reviewing.

armi/reactor/components/component.py
john-science: Responded to both your thoughts.

They were both good ones!

keckler (approved):

PR #998, opened 2022-11-29
Extension of the framework nuclides

jakehader:
## Description

Update to the nuclides that are built into the ARMI framework. This expands the set of nuclide bases to include 4614 nuclides and imports the half-life data from the chart of the nuclides from a combination of the RIPL-3 database and from the IAEA chart of the nuclides.

Other Changes:

* Each `Element` now has a `phase` and a `group`.
* `Element.nuclideBases` becomes `Element.nuclides` - This is an API-breaking change.

---

## Checklist

- [x] This PR has only one purpose or idea.
- [x] Tests have been added/updated to verify that the new/changed code works.

- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html).

- [x] The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any bug fixes or new features.
- [x] The documentation is still up-to-date in the `doc` folder.
- [x] The dependencies are still up-to-date in `setup.py`.

PR #1022, opened 2022-12-13
Explicit Fission Product Modeling

jakehader:
## Description

Updates to the fission product model and nuclides to allow for users to model fission products explicitly based on which cross section libraries/software will be used to generate multi-group cross sections.

This implements a method that looks at the existing nuclides defined in the blueprints `nuclideFlags` and amends this to include all nuclides within regions that are determined to be `Depletable` based on which nuclides the user sets as being `burn=True`. This will not check that the user defines the correct depletable nuclides for this determination, but will only add all the nuclides that are missing to the depletable regions. This is reasonable since the behavior is similar to how it is when the user selects another fission product model option.

This change also introduces a mechanism to remove gaseous fission products from the core that can be called upon by a fuel performance or other plugin code. The previous implementation of this modified the fission yields as a surrogate for this removal, but upon further review it is better to update the number densities within the blocks directly compared to modifying the fission yields. This is currently left not implemented.

Finally, the fuel performance parameters were updated to add setters on the `gasReleaseFraction` and the `bondRemoved`. The description suggested that these values should only range from [0, 1] and any value outside of this range is not valid. The setters now check for this and raise a ValueError if these conditions are not met.

---

## Checklist


jakehader:
This PR depends on https://github.com/terrapower/armi/pull/998 and is basically just pulling out commits from there to isolate changes with added unit tests. This is functional for our internal needs

keckler:
@jakehader 

Can I make the request that we split this into two separate PRs: (1) for the explicit FP modeling and (2) for the FPs to be removed from the fuel.

It appears to me that these two capabilities are separable, and it would make it much easier during our internal integration testing in the process of pulling this into our app.

If you think this is possible, let me know if I can be any help.

jakehader:
> @jakehader
> 
> Can I make the request that we split this into two separate PRs: (1) for the explicit FP modeling and (2) for the FPs to be removed from the fuel.
> 
> It appears to me that these two capabilities are separable, and it would make it much easier during our internal integration testing in the process of pulling this into our app.
> 
> If you think this is possible, let me know if I can be any help.

I will consider it since these are two separate features.

armi/cases/case.py
```diff
@@ -563,38 +563,38 @@ def checkInputs(self):
...
+                runLog.info(
+                    tabulate.tabulate(
+                        queryData,
+                        headers=["Number", "Statement", "Question"],
+                        tablefmt="armi",
                     )
+                )
+            if context.CURRENT_MODE == context.Mode.INTERACTIVE:
+                # if interactive, ask user to deal with settings issues
+                inspector.run()
```
jakehader, line 597: This change provides some header in an interactive session to see the setting validators. I can pull out to a separate PR if needed, but this was a quality of life thing when I was testing.

armi/operators/settingsValidation.py
```diff
@@ -107,9 +105,7 @@ def resolve(self):
                         )
                         if make_correction:
                             self.correction()
-                            self._corrected = True
-                        else:
-                            self._passed = True
+                        self._passed = True
                     except RunLogPromptCancel as ki:
                         raise KeyboardInterrupt from ki
                 else:
```
jakehader, line 111: The `_corrected` attribute was not being used for anything and therefore it was removed. Related to debugging on the setting validators.

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -346,15 +346,7 @@ def _getAverageFuelLFP(self):
         """Compute the average lumped fission products."""
         # TODO: make do actual average of LFPs
         b = self.getCandidateBlocks()[0]
-        lfpCollection = b.getLumpedFissionProductCollection()
-        if lfpCollection:
-            lfpCollectionCopy = lfpCollection.duplicate()
-            fgRemoved = self._getAverageFissionGasRemoved()
-            lfpCollectionCopy.setGasRemovedFrac(fgRemoved)
```
jakehader, line 353: `setGasRemovedFrac` was removed because the removal of fission gases was not implemented correctly. In the current ARMI version we were modifying the fission product yields over time and NOT the number densities so this was actually not removing fission gases correctly. There is a much of clean up related to fixing the way the fission gases should actually be removed by modifying block/component number densities directly.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -85,64 +95,50 @@ def _fissionProductBlockType(self):
...
+                if not targetComponent:
+                    continue
+                ndens = targetComponent.getNumberDensities()
+                updatedNDens = {}
+                for nuc in self.r.blueprints.allNuclidesInProblem:
+                    if nuc in ndens:
+                        continue
+                    updatedNDens[nuc] = 0.0
+                targetComponent.updateNumberDensities(updatedNDens)
+        self._initialized = True
```
jakehader, line 141: This was added and paired with the addition of the `_autoUpdateNuclideFlags` method in `blueprints`. Essentially, moving between lumped fission product modeling and `explicitFissionProducts` would require the user to add 100s of fission products into the `nuclideFlags` section of the blueprints. This allows the same `blueprints` input to be used between two different cases without requiring a significant effort and migration of existing files.

This seemed like the best place to add the extra fission product nuclides to the blocks/components since this is the `FissionProductModel` interface. I tried a couple different ways to add these in the `componentBlueprints` or elsewhere and this seemed like the best fit so that the rest of the reactor construction doesn't have this one-off code based on a setting. 

Another point here is that settings are typically not passed all through the blueprints objects so it  made the code really messy to start passing a settings object around or adding more keyword arguments.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -205,60 +198,146 @@ def _getAllFissionProductNames(self):
...
-        anything else that may be interested in fission product information
-        (e.g. MCNP).
-
-        See Also
-        --------
-        armi.physics.neutronics.crossSectionGroupManager.AverageBlockCollection.getRepresentativeBlock
-
+        The current implementation will update the number density vector
+        of each of the blocks and the gaseous fission products are not
+        moved or displaced to another area in the core.
```
jakehader, line 217: This is a new implementation of this method that actually changes the number densities in the blocks/components where gaseous fission products should be removed. This call is paired with changes in the fuel performance plugin so that when the fuel performance plugin calculates how much gas to remove in a time-step/node it will only remove the specified amount of gas.

One limitation of this method is that it removes the gaseous fission products from existence rather than migrating them somewhere else. I think that this can be improved later on, but I didn't want to add a bunch of design-specific code that would not be generic here. For example, in some designs the fission gas may migrate to a plenum, but not all designs have this.

I can further clarify this caveat/note in the docstring if that is requested.

jakehader: This probably needs a few eyes on it because there is different treatment that is applied when the lumped fission product model is used versus the `explicitFissionProducts` model. The `explicitFissionsProducts` model is more straight-forward because we are just modifying the gaseous nuclides on the block/component directly. In the lumped fission product model, we instead have `LFP35`, `LFP38`, etc. in the block/component and in this case we need to reach into the lumped fission product class and look at the yields of each gaseous nuclide and then calculate the number density of the gaseous nuclides to subtract from the `LFP35`, `LFP38`, etc. number densities.

I have added unit tests for both of these model types and varied the gas removal fractions from 0.0, 0.5, and 1.0 to check that none, half, and all the gaseous fission products are removed, respectively.

armi/physics/neutronics/fissionProductModel/fissionProductModelSettings.py
```diff
@@ -28,13 +29,31 @@ def defineSettings():
             CONF_FP_MODEL,
             default="infinitelyDilute",
             label="Fission Product Model",
-            description="The fission product model to use in this ARMI run",
+            description="",
```
jakehader, line 32: Oops - I will add this back. Make a comment please :)

armi/physics/neutronics/fissionProductModel/tests/test_lumpedFissionProduct.py
```diff
@@ -98,55 +101,26 @@ def setUp(self):
...
-        lfp.setGasRemovedFrac(0.25)
-        gas2 = lfp[xe135]
-        self.assertAlmostEqual(gas1 * 0.75, gas2)
-
     def test_getYield(self):
         """Test of the yield of a fission product"""
         xe135 = nuclideBases.fromName("XE135")
         lfp = self.fpd.createSingleLFPFromFile("LFP39")
-        lfp[xe135] = 3
+        lfp[xe135] = 1.2
```
jakehader, line 108: This is changed since I put a validator on the yields that they must be within `[0.0, 2.0]`.

armi/physics/neutronics/fissionProductModel/tests/test_lumpedFissionProduct.py
```diff
@@ -98,55 +101,26 @@ def setUp(self):
...
-        self.assertEqual(val3, 3)
-        self.assertIsNone(lfp[5])
-
-    def test_getNumberFracs(self):
-        xe135 = nuclideBases.fromName("XE135")
-        lfp = self.fpd.createSingleLFPFromFile("LFP38")
-        numberFracs = lfp.getNumberFracs()
-        self.assertEqual(numberFracs.get(xe135), 1.0)
+        self.assertEqual(val3, 1.2)
+        self.assertEqual(lfp[5], 0.0)
```
jakehader, line 111: This is changed because it used to return a yield of `None` for a nuclide/key that didn't exist. This doesn't really make sense because `0.0` is physical.

armi/physics/neutronics/fissionProductModel/tests/test_lumpedFissionProduct.py
```diff
@@ -233,59 +227,6 @@ def test_getAllFissionProductNames(self):
...
-        lfps = fpd.createSingleLFPCollectionFromFile("LFP35")
-
-        massFrac = {
-            "U238": 24482008.501781337,
-            "LFP35": 0.0,
-            "CL35": 1617.0794376133247,
-            "CL37": 17091083.486970097,
-            "U235": 3390057.9136671578,
-            "NA23": 367662.29994516558,
-        }
```
jakehader, line 251: Collapsing lumped fission products into a single lumped fission product during a run was never used and this was just stale/dead code. I don't think there is any reason to keep it around.

armi/reactor/blueprints/__init__.py
```diff
@@ -342,6 +344,33 @@ def _assignTypeNums(self):
...
+            )
+            for nb in nbs:
+                nuc = nb.name
+                if nuc in self.nuclideFlags or elements.byZ[nb.z] in self.nuclideFlags:
+                    continue
+                nuclideFlag = isotopicOptions.NuclideFlag(
+                    nuc, burn=False, xs=True, expandTo=[]
+                )
+                self.nuclideFlags[nuc] = nuclideFlag
+
```
jakehader, line 373: See comment above on why this exists. Basically, it's for user convenience and allows easy switching between fission product model options without requiring the user to constantly update the `nuclideFlags` in the blueprints.

armi/reactor/tests/test_assemblies.py
```diff
@@ -704,7 +704,7 @@ def test_getBlockData(self):
             "residence": 3.145,
             "eqRegion": -1,
             "id": 299.0,
-            "bondRemoved": 33.7,
+            "bondRemoved": 0.337,
```
jakehader, line 707: This was changed because a setter was added to the `bondRemoved` and `gasReleaseFraction` parameters to enforce that these should be between `[0.0, 1.0]`. I believe anything less than 0.0 or greater than 1.0 is non-physical and should be an error.

armi/settings/fwSettings/globalSettings.py
```diff
@@ -134,7 +134,6 @@ def defineSettings() -> List[setting.Setting]:
             CONF_INITIALIZE_BURN_CHAIN,
             default=True,
             label="Initialize Burn Chain",
-            schema=vol.Coerce(type(bool)),
```
jakehader, line 137: This was implemented in #998 but this schema was an error. The schema is auto-detected as a boolean so this is not needed. This was setting the `initializeBurnChain` setting to `bool`.

armi/physics/neutronics/fissionProductModel/tests/test_lumpedFissionProduct.py
```diff
@@ -98,55 +101,26 @@ def setUp(self):
...
+        lfp[xe135] = 1.2
         val3 = lfp[xe135]
-        self.assertEqual(val3, 3)
-        self.assertIsNone(lfp[5])
-
-    def test_getNumberFracs(self):
-        xe135 = nuclideBases.fromName("XE135")
-        lfp = self.fpd.createSingleLFPFromFile("LFP38")
-        numberFracs = lfp.getNumberFracs()
-        self.assertEqual(numberFracs.get(xe135), 1.0)
```
jakehader, line 123: The `getNumberFracs` method was unused and stale/dead code.

armi/physics/neutronics/fissionProductModel/tests/test_lumpedFissionProduct.py
```diff
@@ -98,55 +101,26 @@ def setUp(self):
...
         lfp = self.fpd.createSingleLFPFromFile("LFP38")
         massVector = lfp.getExpandedMass(mass=0.99)
         self.assertEqual(massVector.get(xe135), 0.99)
 
-    def test_getGasFraction(self):
-        """Test of the get gas removal fraction"""
-        lfp = self.fpd.createSingleLFPFromFile("LFP35")
-        frac = lfp.getGasFraction()
-        self.assertGreater(frac, 0.0)
-        self.assertLess(frac, 1.0)
```
jakehader, line 136: The `getGasFraction` method was unused and stale/dead code.

armi/physics/neutronics/fissionProductModel/tests/test_lumpedFissionProduct.py
```diff
@@ -98,55 +101,26 @@ def setUp(self):
...
         _ = nuclideBases.fromName("XE135")
         lfp = self.fpd.createSingleLFPFromFile("LFP38")
         lfp.printDensities(10.0)
 
-    def test_getLanthanideFraction(self):
-        """Test of the lanthanide fraction function"""
-        lfp = self.fpd.createSingleLFPFromFile("LFP35")
-        frac = lfp.getLanthanideFraction()
-        self.assertGreater(frac, 0.0)
-        self.assertLess(frac, 1.0)
```
jakehader, line 148: The `getLanthanideFraction` method was unused and stale/dead code.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -344,133 +298,17 @@ def getMassFrac(self, oldMassFrac=None):
...
-    """
-
-    def __init__(self):
-        super(SingleLumpedFissionProductCollection, self).__init__()
-        self.collapsible = True
-
-    def getFirstLfp(self):
-        return list(self.values())[0]
-
-    def getName(self):
```
jakehader, line 391: This was stale/dead code that was implemented long ago and never used for anything practical.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -676,50 +516,6 @@ def expandFissionProducts(massFrac, lumpedFissionProducts):
...
-    collapsing only works if there is a 'single lumped fission product collection' -- otherwise its confusing to
-    determine how much of what isotope goes to which lumped fission products
-    """
-
-    assert isinstance(lumpedFissionProducts, SingleLumpedFissionProductCollection)
-    lfp = lumpedFissionProducts.getFirstLfp()
-    newMassFracs = {}
-    lumpedFissionProductsMassFracs = {}
-
-    for nb in lfp.keys():
```
jakehader, line 707: See comment on the tests. This was stale/dead code that was never used.

armi/resources/referenceFissionProducts.dat
```diff
@@ -1,719 +1,717 @@
...
+LFP35 GE76   1.5000E-04
+LFP35 AS75   7.0000E-05
+LFP35 SE77   2.9000E-04
+LFP35 SE78   5.6000E-04
+LFP35 SE79   9.1000E-04
+LFP35 SE80   1.5100E-03
+LFP35 SE82   3.9000E-03
+LFP35 BR81   2.4600E-03
+LFP35 KR82   6.0000E-05
+LFP35 KR83   6.0100E-03
```
jakehader, line 10: This is just a formatting change. The `13` input on the first two columns is actually CARD 13 from `MC2-2` and there is no reason to keep it around. Additionally, I updated the nuclide labels to be based on the nuclides that were implemented/updated in #998 rather than being the `MC2-3` labels.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -69,46 +67,45 @@ def __init__(self, name=None):
...
 
         Returns
         -------
-        yld : yield of the fission product. Defaults to None.
+        yld : yield of the fission product.
         """
-        yld = self.yld.get(fissionProduct, default)
-        if yld and isGas(fissionProduct):
-            yld *= self.gasRemainingFrac
-        return yld
```
jakehader, line 97: This was modified and replaced with the number density updates implemented in `removeFissionGasesFromBlocks` on the `FissionProductModel` interface. I cannot speak to why the yields were being updated but I believe this to be incorrect.

Essentially, the old implementation modified the yield fractions of gaseous nuclides in the lumped fission products rather than modifying the number densities directly in the blocks/components. In this way, the global flux calculation would have number densities that would not be reflected by the removal of the gaseous fission products. Instead, when `MC2-3` code was run, the microscopic XS for `LFP35`, `LFP38`, etc. were being augmented. 

By modifying the number densities directly this will capture both the cross section AND the flux solution effects together rather than only sort of treating one of the effects.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -205,60 +198,146 @@ def _getAllFissionProductNames(self):
...
+                remainingFraction = 1.0 - removedFraction
+                updatedNumberDensities = {}
+                for nuc, val in b.getNumberDensities().items():
+                    nb = nuclideBases.byName[nuc]
+                    updatedNumberDensities[nuc] = (
+                        remainingFraction * val
+                        if lumpedFissionProduct.isGas(nb)
+                        else val
+                    )
+                b.updateNumberDensities(updatedNumberDensities)
```
mgjarrett, line 284: I assume that the gas removal fractions are cumulative, but `_removeFissionGasesExplicitFissionProductModeling` is going to be called multiple times.

So if the gas removal fractions are 0.3, 0.4, 0.5 for three successive state points, you would end up with 0.7 * 0.6 * 0.5 = 0.21 of the initial fission gas instead of 0.5. Am I reading this wrong?

opotowsky (commented):
I didn't finish the code review yet but I figured I'd leave feedback tonight that you can work on first thing in the AM while I/we finish up our reviews. 

There is a significant coverage drop despite all of the testing work you did. There's two methods/functions that are newly uncovered, and a new one that needs to get tested. Since I can't comment on code that's not changed....well, you'll see what I did...

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -205,60 +192,146 @@ def _getAllFissionProductNames(self):
...
+
+        def _getGaseousFissionProductNumberDensities(b, lfp):
+            """Look into a single lumped fission product object and pull out the gaseous atom number densities."""
+            numberDensities = {}
+            for nb in lfp.keys():
+                if not lumpedFissionProduct.isGas(nb):
+                    continue
+                yld = lfp[nb]
+                ndens = b.getNumberDensity(lfp.name)
+                numberDensities[nb.name] = ndens * yld
```
keckler, line 249: I think this evaluation will only work as expected the before any gaseous fission products have been removed from the LFP.

For instance, you call this method from within `_removeFissionGasesLumpedFissionProductModeling()`. The first time you execute that method, this logic works fine. But once `_removeFissionGasesLumpedFissionProductModeling()` has actually modified the number density of the LFP, then yields are no longer representative of the isotopic split within the LFP. So if you call `_removeFissionGasesLumpedFissionProductModeling()` _again_, the value returned by this method will actually be over-representing the number of gaseous fission products in the LFP. Assuming that `_removeFissionGasesLumpedFissionProductModeling()` will be called on every time node, this will be a problem.

Does that make sense?

I think the only way around this is through some sort of caching from one time step to the next.

This is _maybe_ why the fission yields were being modified in the previous implementation, because that would account for the cumulative impact of the fission gases disappearing from the LFP.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -270,7 +225,6 @@ class LumpedFissionProductCollection(dict):
     """
 
     def __init__(self):
-        super(LumpedFissionProductCollection, self).__init__()
```
keckler, line 273: lol, what was going on here?

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -482,10 +320,6 @@ class FissionProductDefinitionFile:
     The path to this file name is specified by the
```
keckler, line 320: ...by the what?

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -580,28 +398,50 @@ def _readOneLFP(self, linesOfOneLFP):
...
+            nbs = nuclideBases.byMcc3Id.values()
+        else:
+            raise ValueError(
+                f"An option to handle the `fpModelLibrary` "
+                f"set to `{cs['fpModelLibrary']}` has not been "
+                f"implemented."
+            )
+    return nbs
+
+
```
keckler, line 444: This function doesn't appear to belong in this module.

keckler (changes requested):
Okay, here's some comments for you at this point.

I still have to review 2 files worth of tests and `referenceFissionProducts.dat`. So probably any further comments will be minor.

I have a few concerns, particularly related to the removal of fission gases. Those are captured in comments throughout.

I also am worried that  the changed handling of the gaseous fission products might surprise non-TP users of ARMI if all of the sudden their block number densities are different when they update their framework. 

I do not disagree that this change should be made, but I think it is likely to
cause some confusion down the line. That is really a product of having a bad
fission gas model in the first place, but still. Probably nothing can be done about that if we are intent on fixing the model.

armi/cases/case.py
jakehader: I can do this.

armi/nucDirectory/nucDir.py
```diff
@@ -362,7 +362,7 @@ def getThresholdDisplacementEnergy(nuc):
     nuc = getNuclide(nuc)
     el = elements.byZ[nuc.z]
     try:
-        ed = eDisplacement[el]
+        ed = eDisplacement[el.symbol]
```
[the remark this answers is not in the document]
jakehader: Sure!

armi/operators/settingsValidation.py
jakehader: I can add it back in. Nothing was using it. You're right that it's not a good place in this PR (same as above). I big thing here is that self._passed should be set to True regardless.

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -346,15 +346,7 @@ def _getAverageFuelLFP(self):
...
-        else:
-            lfpCollectionCopy = lfpCollection
-
-        return lfpCollectionCopy
+        return b.getLumpedFissionProductCollection()
```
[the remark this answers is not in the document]
jakehader: Good point! That should be removed now I think too.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -85,64 +95,50 @@ def _fissionProductBlockType(self):
...
+                f"Select another `fpModel` option."
+            )
         return blockType
 
     def interactBOL(self):
         interfaces.Interface.interactBOL(self)
         self.setAllBlockLFPs()
 
-    def setAllBlockLFPs(self, blockType=None, setMaterialsLFP=False):
+    def setAllBlockLFPs(self):
```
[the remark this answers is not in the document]
jakehader: I would prefer to keep this the same for now so I don't have to update more stuff that is dependent on it and possibly risk downstream impacts on others. I can make an issue.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -85,64 +95,50 @@ def _fissionProductBlockType(self):
...
+                targetComponent = b.getComponent(self._fissionProductBlockType)
+                if not targetComponent:
+                    continue
+                ndens = targetComponent.getNumberDensities()
+                updatedNDens = {}
+                for nuc in self.r.blueprints.allNuclidesInProblem:
+                    if nuc in ndens:
+                        continue
+                    updatedNDens[nuc] = 0.0
+                targetComponent.updateNumberDensities(updatedNDens)
```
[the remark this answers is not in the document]
jakehader: Yes, I see that now. Will move up into `if self._useGlobalLFPs` statement.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -205,60 +198,146 @@ def _getAllFissionProductNames(self):
...
-        runLog.debug("  Caching LFP densities of all blocks")
-        lfpDensities = {}
-        for b in blockList:
-            for lfpName in self._globalLFPs:
-                lfpDensities[lfpName, b.getName()] = b.getNumberDensity(lfpName)
-        return lfpDensities
+        return fissionProductNames
 
-    def updateFissionGasRemovalFractions(self):
+    def removeFissionGasesFromBlocks(self, gasRemovalFractions: dict):
```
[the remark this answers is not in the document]
jakehader: So I discussed this with @sammiller11235 and the fuel performance engine _may_ do this, but actually he would prefer if the core team managed where or how to remove things from the blocks/components and he just is just responsible for saying how much to move/remove.

Just responding that we thought about it and I do think it makes sense for the `FissionProductModel` to manage fission products in this way. Also, this was already implemented in the API to have this functionality/responsibility so I don't feel like we should change that just yet.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -205,60 +198,146 @@ def _getAllFissionProductNames(self):
...
-                totalWeight += weight
-            else:
-                # set individually
-                block.getLumpedFissionProductCollection().setGasRemovedFrac(
-                    block.p.gasReleaseFraction
+        if not self.cs["fgRemoval"]:
+            runLog.info(
+                "Skipping removal of gaseous fission products since the `fgRemoval` setting is disabled."
+            )
+            return
```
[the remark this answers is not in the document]
jakehader: Yeah I can do that. If fact, this is checked before this method would be called in our plugins so this is reasonable to just remove.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -205,60 +198,146 @@ def _getAllFissionProductNames(self):
...
+                )
+                updatedGasRemovalFractions[b] = 0.0
+            elif frac > 1.0:
+                runLog.warning(
+                    f"The fission gas removal fraction is greater than one for {b}. Setting to one."
                 )
+                updatedGasRemovalFractions[b] = 1.0
+            else:
+                updatedGasRemovalFractions[b] = frac
+        gasRemovalFractions.update(updatedGasRemovalFractions)
```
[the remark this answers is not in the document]
jakehader: Yeah, so I am going to update this based on some suggestions and information from @sammiller11235 and from @mgjarrett

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -205,60 +198,146 @@ def _getAllFissionProductNames(self):
...
+
+        def _getGaseousFissionProductNumberDensities(b, lfp):
```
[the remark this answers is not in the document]
jakehader: Will do. I am going to have another for explicit as well.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -205,60 +198,146 @@ def _getAllFissionProductNames(self):
...
+            """Look into a single lumped fission product object and pull out the gaseous atom number densities."""
+            numberDensities = {}
+            for nb in lfp.keys():
+                if not lumpedFissionProduct.isGas(nb):
```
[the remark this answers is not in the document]
jakehader: Yeah, so right _now_ the `gasRemovalFraction` doesn't say anything about how these gases are determined. In reality where will be some chemical equilibrium based on Gibbs Free Energy minimization (typically done in a code like FACTSAGE or something else). In this case we are just looking at the gases that will be gases, but there can be Cesium migration that wouldn't be accounted in higher temperatures where `CsI` is formed. This is an assumption that I will document in the docstring.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
jakehader: Fission yields are a generation term rather than a removal term. I am going to modify this implementation based on a conversation with @sammiller11235 and @mgjarrett. If it takes too long or isn't seemingly correct I might just remove this functionality altogether and then create an issue for determining the requirements. I think that there are enough "methodology" questions to be asked here that a generalized framework implementation might not cut it. Not sure yet.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -205,60 +192,146 @@ def _getAllFissionProductNames(self):
...
+
+                    # If the lumped fission products are global then we are going
+                    # release the average across all the blocks in the core and
+                    # this data is collected iteratively.
+                    if self._useGlobalLFPs:
+                        weight = b.getVolume() * (b.p.flux or 1.0)
+                        weightedGasRemoved[lfp] += (
+                            sum(ndens.values()) * removedFraction * weight
+                        )
+                        totalWeight[lfp] += weight
```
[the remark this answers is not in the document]
jakehader: Sure thing. I actually think that this caveat should be added to the `FissionProductModel` class docstring since LFPs that don't track `Xe` and `Sm` directly/explicitly would be wrong anyways.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -205,60 +192,146 @@ def _getAllFissionProductNames(self):
...
+                        continue
+
+                    for lfp in lfpCollection.values():
+                        # The updated number density is calculated as the current number density subtracting
+                        # the amount of gaseous fission products that are being removed.
+                        updatedLFPNumberDensity = b.getNumberDensity(lfp.name) - (
+                            weightedGasRemoved[lfp] / totalWeight[lfp]
+                        )
+                        numberDensities.update({lfp.name: updatedLFPNumberDensity})
+                        b.setNumberDensities(numberDensities)
```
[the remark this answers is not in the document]
jakehader: Yes, I think you are correct. There should be a condition on removal and also averaging. If there are no fission products then don't include them in the weighted sum and if there are no fission products don't modify.

armi/physics/neutronics/fissionProductModel/fissionProductModelSettings.py
```diff
@@ -28,29 +29,127 @@ def defineSettings():
...
                 "MO99",
+                "explicitFissionProducts",
+            ],
+        ),
+        setting.Setting(
+            CONF_FISSION_PRODUCT_LIBRARY_NAME,
+            default="",
+            label="Fission Product Library",
+            description=(
+                f"This setting is used when the `{CONF_FP_MODEL}` setting "
```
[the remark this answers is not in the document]
jakehader: It _is_ optional but the code will just crash without it being set, which is why I added a settingsValidator. I think a better default might be just `MC2-3`.

armi/physics/neutronics/fissionProductModel/fissionProductModelSettings.py
```diff
@@ -28,29 +29,127 @@ def defineSettings():
...
+                f"is set to `explicitFissionProducts` and is used to configure "
+                f"all the nuclides that should be modeled within the core. "
+                f"Setting this is equivalent to adding all nuclides in the "
+                f"selected code library (i.e., MC2-3) within the blueprints "
+                f"`nuclideFlags` to be [xs:true, burn:false]. This option acts "
```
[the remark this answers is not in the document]
jakehader: I will take a look again at the logic and see if `burn: false` or `burn: true` makes more sense. My reason for not adding `burn: true` is that this is used to determine which regions are depletable in the model. 

The main purpose of this implementation is to allow all the fission products to exist in the block/components that need to be there so that a code like `ORIGEN` could be run and put number densities from its separate solve on the blocks/components. This is not intended to be used with ARMI's built-in `burn-chain.yaml` file since that would then require a user to also build their own `burn-chain.yaml` file. If they want to do this then that is fine, but then they should also go and manually setup the `nuclideFlags` in the blueprints based on what they are trying to do.

armi/physics/neutronics/fissionProductModel/fissionProductModelSettings.py
```diff
@@ -28,29 +29,127 @@ def defineSettings():
...
+                f"as a short-cut so that analysts do not need to change their "
+                f"inputs when modifying the fission product treatment for "
+                f"calculations. This may be extended for other cross section "
+                f"generation codes."
```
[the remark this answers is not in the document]
jakehader: Yeah, I get in the habit of an all or nothing for f-strings. I can change it, but I like it?

armi/physics/neutronics/fissionProductModel/fissionProductModelSettings.py
```diff
@@ -28,29 +29,127 @@ def defineSettings():
...
+
+    queries.append(
+        Query(
+            lambda: inspector.cs["fpModel"] == "explicitFissionProducts"
+            and bool(inspector.cs["initializeBurnChain"]),
+            (
+                f"The explicit fission product model is enabled, but initializing the burn chain is also enabled. This will "
+                f"likely fail."
+            ),
+            (f"Would you like to disable the burn chain initialization?"),
```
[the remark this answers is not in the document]
jakehader: I can remove the f-strings here and the `()` for strings that aren't multiple lines.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -188,9 +166,7 @@ def getMassFrac(
         massFracDenom = self.getMassFracDenom(useCache=useCache, storeCache=storeCache)
         if not nuclideBase:
             nuclideBase = nuclideBases.byName[nucName]
-        return self.__getitem__(nuclideBase, default=0) * (
-            nuclideBase.weight / massFracDenom
-        )
+        return self.__getitem__(nuclideBase) * (nuclideBase.weight / massFracDenom)
 
     def getMassFracDenom(self, useCache=True, storeCache=True):
```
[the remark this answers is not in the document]
jakehader: Yeah.. I was thinking of just removing these but I think some code uses them. Will leave as-is.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -159,22 +153,6 @@ def getMassFracs(self):
...
-
-        numberFracs = {}
-        totalNumber = sum(self.yld.values())
-        for nuc, yld in self.yld.items():
-            numberFracs[nuc] = yld / totalNumber
-        return numberFracs
-
     def getMassFrac(
         self, nucName=None, nuclideBase=None, useCache=True, storeCache=True
     ):
```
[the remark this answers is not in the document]
jakehader: Yeah I have no idea and didn't look into it. Going to leave as-is for now.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -223,41 +199,20 @@ def getExpandedMass(self, mass=1.0):
...
 
-    def printDensities(self, lfpDens):
-        """Print densities of nuclides given a LFP density."""
+        Notes
+        -----
+        This multiplies the provided number density for the lumped fission
+        product by the yield of each nuclide.
+        """
         for n in sorted(self.keys()):
             runLog.info("{0:6s} {1:.7E}".format(n.name, lfpDens * self[n]))
```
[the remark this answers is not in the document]
jakehader: The `lfpDens` is what is entered by the user. This method is pretty silly and I might just remove if it's not used elsewhere. The `lfpDens` would in-fact be updated though.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
jakehader: Who knows! I will add more...

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
jakehader: Yeah I can move.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
```diff
@@ -614,8 +454,8 @@ def _buildMo99LumpedFissionProduct():
     for lfp in nuclideBases.where(
         lambda nb: isinstance(nb, nuclideBases.LumpNuclideBase)
     ):
-        # Not all lump nuclide bases defined are fission products, so ensure that only fission products are considered.
-        if not "FP" in lfp.name:
+        # Not all lump nuclides bases defined are fission products, so ensure that only fission products are considered.
+        if not ("FP" in lfp.name or "REGN" in lfp.name):
```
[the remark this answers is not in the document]
jakehader: Yeah this code is sort of silly and not needed. I don't know what it does or why someone would use it. Might delete altogether.

armi/physics/neutronics/fissionProductModel/tests/test_lumpedFissionProduct.py
keckler: We talked offline about the upper bound on the yields being 2, and about potentially removing that limitation.

armi/physics/neutronics/fissionProductModel/fissionProductModelSettings.py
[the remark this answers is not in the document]
keckler: Hm, I actually assumed that making it an f-string would actually make it slower (even if it is just a teeny tiny bit). Turns out it doesn't impact the speed at all, since there is no internal arguments passed here.

Keep it then!

armi/physics/neutronics/fissionProductModel/fissionProductModelSettings.py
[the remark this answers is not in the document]
keckler: Nah, I think your implementation with `burn: False` is the right way to go. The aspect about `getDefaultNuclideFlags()` wasn't clear to me at first.

It appears that a case will always have burnable actinides defined, unless the user specifically turns those off.

So I think this is fine.

armi/reactor/blueprints/__init__.py
```diff
@@ -342,6 +344,33 @@ def _assignTypeNums(self):
                     if bDesign not in self.blockDesigns:
                         self.blockDesigns.add(bDesign)
 
+    def _autoUpdateNuclideFlags(self, cs):
+        """
+        This method is responsible for examining the fission product model treatment
+        that is selected by the user and adding a set of nuclides to the `nuclideFlags`
+        list.
+        Notes
```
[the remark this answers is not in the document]
opotowsky: Where are you seeing the rendered doc strings, or are you doing that locally?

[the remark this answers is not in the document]
keckler: No, I'm not rendering it at all. I'm just making the comment based on my experience.

But actually, now that I look at the conventions in detail, maybe I'm wrong!

https://numpydoc.readthedocs.io/en/latest/format.html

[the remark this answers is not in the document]
keckler: But it would still be consistent with all our other docstrings to have a blank line there.

armi/physics/neutronics/fissionProductModel/tests/test_lumpedFissionProduct.py
jakehader: I tied this to a constant that was fixed to 2 in the code so at least its tied to another assumption.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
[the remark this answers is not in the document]
opotowsky: Issue is made: https://github.com/terrapower/armi/issues/1050

john-science:
It's a big PR.

I see changes to blueprints and settings.  Are these backwards compatible?  I just want to make sure we aren't breaking all of the current, really important, settings/blueprints YAML files we have lying around.

armi/settings/caseSettings.py
```diff
@@ -168,10 +168,10 @@ def _directAccessOfSettingAllowed(self, key):
...
         if key in SIMPLE_CYCLES_INPUTS and self.__settings["cycles"].value != []:
             err = ValueError(
-                "Cannot grab simple cycles information from the case settings"
-                " when detailed cycles information is also entered.\n In general"
-                " cycles information should be pulled off the operator or parsed"
-                " using the appropriate getter in the utils."
+                "Cannot grab simple cycles information from the case settings "
+                "when detailed cycles information is also entered. In general "
+                "cycles information should be pulled off the operator or parsed "
+                "using the appropriate getter in the utils. "
```
john-science, line 174: MINOR

Unnecessary space after the end of the sentence.

armi/cases/case.py
jakehader: Going to leave as is. This isnt a breaking change.

armi/reactor/blueprints/__init__.py
```diff
@@ -342,6 +344,33 @@ def _assignTypeNums(self):
                     if bDesign not in self.blockDesigns:
                         self.blockDesigns.add(bDesign)
 
+    def _autoUpdateNuclideFlags(self, cs):
```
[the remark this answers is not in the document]
jakehader: Added a test for this in `test_fissionProductModel.py`. See: `TestFissionProductModelExplicitMC2Library`.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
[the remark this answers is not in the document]
jakehader: Leaving as-is because it is paired with the composite structure.

armi/physics/neutronics/fissionProductModel/lumpedFissionProduct.py
jakehader: Moved to `isotopicOptions.py`

jakehader:
> It's a big PR.
> 
> I see changes to blueprints and settings. Are these backwards compatible? I just want to make sure we aren't breaking all of the current, really important, settings/blueprints YAML files we have lying around.

@john-science - existing blueprints and settings are not impacted by this change. This change only impacts modeling with `fpModel` is set to `explicitFissionProducts`, but the default is still to use lumped fission products.

keckler (approved):
@jakehader Thank you for all of your work on this. I think this puts us in a good place to be able to model the explicit fission products!

keckler:
I wanted to update this thread with the current state of what all was changed and/or added with this PR, since the target changed a couple times.

This PR enables the modeling of explicit fission products, as compared to previously being limited to using lumped fission products at best.

This PR does NOT implement the correct removal of fission gases from blocks. However, it does remove the previous implementation, which was modifying the yields of fission products to artificially exclude some fraction of gaseous fission products. Now, fission products are left in place, and a warning is issued if a user tries to remove them via the `removeFissionGasesFromBlocks()` method. If one wants to remove fission gases, they should overwrite that method from within their own plugin.

That is my understanding, but please feel free to chime in with more details if I am wrong.

opotowsky:
Thanks so much for this enormous lift @jakehader !

PR #1024, opened 2022-12-14
Fix power after snapshot load

onufer:
## Description

Fix a bug where power setting would not be respected during snapshot database load

---

## Checklist


onufer:
@john-science , we have no unit tests for the main interface at all and its part of db load which is a pretty heavy lift.  i think we should rely on internal in-use tests.

john-science (approved):
I don't see the _cause_ of the problem.  But the solution proposed seems fine, if the problem exists.

PR #1029, opened 2022-12-16
Speeding up unit tests

john-science:
## Description

Unfortunately, the `armiRun.yaml` test reactor that we use in so many of our unit tests wasn't designed for fast tests. It has 3 cycles, when 2 would be enough for a unit test.  It has 9 rings of assemblies, when we could have used 3 or 4 and still had all the same types of assemblies (fuel, reflector, etc).

So, before this PR, the unit tests for ARMI took about 22 minutes on GitHub Actions. If we made the test reactor smaller, I bet we could get that down to 10 minutes.  But that would mean fixing several hundred tests. And probably hundreds more down stream.

> For this PR, I looked at our repos longest-running unit tests and cut their run time in half.  It's not much, but it's a start.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/tests/test_databaseInterface.py
+++ b/armi/bookkeeping/db/tests/test_databaseInterface.py
@@ -292,10 +292,19 @@ class TestDatabaseReading(unittest.TestCase):
         del cls.r
         cls.r = None
 
+    def _fullCoreSizeChecker(self, r):
+        """TODO"""
+        self.assertEqual(r.core.numRings, 3)
+        self.assertEqual(r.p.cycle, 0)
+        self.assertEqual(len(r.core.assembliesByName), 19)
+        self.assertEqual(len(r.core.circularRingList), 0)
+        self.assertEqual(len(r.core.blocksByName), 95)
+
     def test_growToFullCore(self):
         with Database3(self.dbName, "r") as db:
             r = db.load(0, 0, allowMissing=True)
 
+        # test partial core values
         self.assertEqual(r.core.numRings, 3)
         self.assertEqual(r.p.cycle, 0)
         self.assertEqual(len(r.core.assembliesByName), 7)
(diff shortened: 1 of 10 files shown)


john-science:
Damn, I didn't see this code coverage change earlier.  I'll look into it.

armi/bookkeeping/db/tests/test_databaseInterface.py
```diff
@@ -343,11 +348,11 @@ def test_growToFullCoreFromFactoryWithCS(self):
...
         r.core.growToFullCore(self.cs)
 
-        self.assertEqual(r.core.numRings, 9)
+        self.assertEqual(r.core.numRings, 3)
         self.assertEqual(r.p.cycle, 0)
-        self.assertEqual(len(r.core.assembliesByName), 217)
+        self.assertEqual(len(r.core.assembliesByName), 19)
         self.assertEqual(len(r.core.circularRingList), 0)
-        self.assertEqual(len(r.core.blocksByName), 1085)
+        self.assertEqual(len(r.core.blocksByName), 95)
```
keckler, line 355: It could be pretty useful if we made all of these expected values into class variables in the `setUpClass()`, just to avoid the repetition.

We could even just define a little helper method with this block of code in it, to avoid repeating all of that entirely.

keckler: Not necessary, of course, just a suggestion.

armi/reactor/tests/test_reactors.py
```diff
@@ -200,6 +208,16 @@ def loadTestReactor(
...
 
+def reduceTestReactorRings(r, cs, maxNumRings):
+    """Helper method for the test reactor above
+    The goal is to reduce the size of the reactor for tests that don't neeed
+    such a large reactor, and would run much faster with a smaller one
+    """
+    if 9 > maxNumRings > 1:
+        for ring in range(9, maxNumRings, -1):
+            r.core.removeAssembliesInRing(ring, cs)
+
```
keckler, line 219: I get that this is a method that won't be touched too often, but do we want to make it a little more robust and general?

For instance:
- Telling the user when it doesn't do anything because the specified `maxNumRings` is out of range
- Ensuring that the upper limit is in line with the test reactor in the module, instead of hard-coding a value of `9` into this method

keckler (approved):
A good addition!

armi/reactor/tests/test_reactors.py
john-science: Agreed.

The reason it looks like this was because it was originally INSIDE the `loadTestRector()` method. Which was much easier to keep it clean. Unfortunately, it can't live there because that breaks all the tests when they are run in parallel.

Sure though.  5 minutes

john-science: done

armi/bookkeeping/db/tests/test_databaseInterface.py
john-science: Sure. Done.

opotowsky (approved):
Looks good to me! 

I like Chris' suggestions, but approve as is.

And thanks for adding that snapshot test. A "heavy lift" for me yesterday is a medium lift today...but is always a pinky lift for the venerable John Stilley!

john-science:
This takes the run time of all our code coverage (on GitHub Actions at least) from ~22 minutes, to ~20 minutes. That is not a huge improvement.

But if we use these tools and tricks, a few more PRs like this COULD make a big improvement.  #just-saying

PR #1040, opened 2022-12-20
Bug fix to correct difference tabulation 

opotowsky:
## Description

🐞 alert in `_compareSets`!

The length of a set difference was wrong when src has parameters not in ref. Easy little fix! @Nebbychadnezzar our discussion led to this bug find!

`test_compareDatabaseSim` failed after the bug fix because it relied on the previously faulty `len(src - ref)` check returning 0. Because the test added extra time steps to the database, the bug fix now results in `compareDatabases` returning an `nDiff` of `None`. I simplified the test to insert a single parameter difference. `nDiffs()` is 3 because the `mean(x)`, `mean(abs(x))`, and `max(abs(x))` are all counted.

---

## Checklist


jakehader:
I understand the fix but don't necessarily understand the optional name without looking at how this function is used in the codebase. This error message isn't really helpful as is and if I was to get this as a user I would not know what to do. 

Let me dig around some more but maybe @john-science can offer up an idea when he's got some time.

opotowsky:
> > Edit: I don't see anywhere in the wild where the optional name argument is not provided. This can probably be required? What was the case where you saw a None?
> 
> You're correct -- I think anywhere that uses this in the code provides that argument. I did not provide the `name` arg when I wrote the test initially, which is how I got the above failure. I'm fine making it required, since it doesn't break any other code.

Instead of making it required I fixed the string processing. Wasn't feeling strongly one way or the other

opotowsky:
Seems that `test_compareDatabaseSim` relied on this bug. Fixing.

armi/bookkeeping/db/compareDB3.py
```diff
@@ -493,8 +490,10 @@ def _compareComponentData(
 
         if srcSpecial ^ refSpecial:
             out.writeln(
-                "Could not compare data because one uses special formatting, "
-                "and the other does not. Ref: {} Src: {}".format(refSpecial, srcSpecial)
+                "Could not compare data for parameter {} because one uses special "
+                "formatting, and the other does not. Ref: {} Src: {}".format(
+                    paramName, refSpecial, srcSpecial
+                )
```
opotowsky, line 496: This was a tiny update I made and forgot to PR. Since it's in the same file, figured I could sneak it in 😉

jakehader (approved):

Issue #1039, opened 2022-12-20
Request for improved parameter lookup speed


mgjarrett:
https://github.com/terrapower/armi/blob/59b0cad5da848f8dfd1fa65e6cb858f42f964061/armi/reactor/parameters/parameterDefinitions.py#L399-L422

This implementation is painfully slow. In my experience experimenting with this, the "short circuit" option is not ever realized; there are always dozens of represented types. Instead, we look through _all_ of the parameters in the `ParameterDefinitionCollection` (644 of them in my test case) and perform a string comparison (`pd.name == name`) for all 644 of them, just to ensure that `pd.name == name` for exactly one case.

This should be replaced by a hashed lookup. Can we just use a combination of parameter name and represented type as the dictionary key? Looking up a value from a `dict` with `len(dict) == 644` should be orders of magnitude faster than performing `pd.name == name` 644 times.

These string operations cause the uniform mesh conversion to be very slow: it uses `__getitem__` to get parameters from the `sourceBlock` and set them on the `destBlock`. https://github.com/terrapower/armi/blob/59b0cad5da848f8dfd1fa65e6cb858f42f964061/armi/reactor/converters/uniformMesh.py#L994-L1034

If it is a problem to have a given parameter multiply defined on a collection, this is something that we should check for one time, when the `ParameterDefinitionCollection` is initialized, or when a new parameter is added. We shouldn't be re-performing this error check every single time we're trying to get a value for a parameter.

mgjarrett:
Related to #991

mgjarrett:
Okay I was slightly off target on the first assessment. It seems like it is not calls to `b.p[paramName]` that are expensive, but rather `b.p.pDefs[paramName]`.

It's looking like maybe we can get a big performance boost by just pulling out any of these `ParameterDefinitionCollection.__getitem__()` lookups out of nested loops. Right now, it's being performed ~2-3 million times for a single call to the uniform mesh converter on a reactor with 155 assemblies and ~40 blocks per assembly. If we can cut that down to about ~10,000 or fewer that's going to provide a ton of speedup without needing to refactor the `ParameterDefinitionCollection` itself.

jakehader:
@mgjarrett - I think this is important enough to do in the short term. Let's brainstorm tomorrow on how this can be done or I can help. This is probably some of my doing with the uniform mesh refactoring? Not sure how long this was like this though.

mgjarrett:
[...]

Issue #991, opened 2022-11-21
Request: Speed up the uniform mesh converter


keckler:
The uniform mesh converter can sometimes take a very long time. However the majority of what it does is loop over all the assemblies in the core and assign parameters back and forth between blocks. 

Calls to [`setAssemblyStateFromOverlaps`](https://github.com/terrapower/armi/blob/main/armi/reactor/converters/uniformMesh.py#L397) are all completely independent and could benefit from parallel execution. Based on looking at the uniform mesh generator, this should cover the majority of the work in that class and I think could speed up the conversion process a ton.

ntouran:
Can we at least run the operation through the profiler first to see exactly what the slow bottlenecks are with precision instrumentation? Then we can see if there's any superlow hanging fruit before parallelizing.

keckler:
Yes, a very good idea! This was just based on the keckler-profiler, aka my eyeballs looking at the exploding runtimes.

jakehader:
> Can we at least run the operation through the profiler first to see exactly what the slow bottlenecks are with precision instrumentation? Then we can see if there's any superlow hanging fruit before parallelizing.

I did this months ago but didn't write seem to write it down. It is slow on getting and setting number densities for each block in the core. I feel like this suggest keeps getting made. @john-science did timing tests of this already a couple months ago too.

john-science:
A good process for speeding up code goes something like:

1. Profile the code.
2. Review the code profile and see where the slow-downs are and what expensive operations are being called inside nested loops that can easily be improved.
3. See what data structures are being used.
4. See if we can lean more on NumPy versus pure Python
5. and on and on and on...
6. The very last act of desperation is deciding you have sped up the code as much as humanly possible and you need to parallelize it.

It is not appropriate to say "We need to parallelize this code" before you've done any work to improve it's performance.

john-science:
Also, and not for nothing, you want to parallelize the uniform mesh converter?  Even though the ARMI model is already run in parallel?

You can't spawn parallel code INSIDE code that's already parallel.

Is the uniform mesh converter run outside the already parallel code in the ARMI interface loop?

mgjarrett:
[...]

PR #1042, opened 2022-12-20
Efficiency improvement

mgjarrett:
## Description

The uniform mesh converter can take a very long time. Profiling the converter identified a couple of bottlenecks that turned out to be relatively low-hanging fruit. This is partially documented in #991 and #1039 .

These updates reduced uniform mesh conversion time by 80% by avoiding the `copy.deepcopy(block)`. This reduces the amount of detail stored on the uniform mesh reactor, but it is still sufficient for the physics that need to be performed on the uniform mesh reactor.

These updates also reduced operator initialization time by about 40% when thermal expansion is enabled.

---

## Checklist


armi/reactor/blocks.py
```diff
@@ -157,6 +158,47 @@ def __deepcopy__(self, memo):
 
         return b
 
+    def createBlankCopy(self):
```
jakehader, line 161: This might be better as a function somewhere else because this adds design details into the generic block class. This could be closer to where it's applied as a function for now and upgraded to a method later maybe.

armi/reactor/composites.py
```diff
@@ -1249,12 +1249,14 @@ def getNuclideNumberDensities(self, nucNames):
...
 
-        nucDensForEachComp = numpy.array(
-            [
-                [c.getNumberDensity(nuc) for nuc in nucNames]
-                for c in self.iterComponents()
-            ]
-        )  # c x n
+        densListForEachComp = []
+        for c in self.iterComponents():
+            numberDensityDict = c.getNumberDensities()
```
jakehader, line 1254: This is faster because of using `getNumberDensities()`, right?

armi/reactor/converters/uniformMesh.py
```diff
@@ -581,14 +604,17 @@ def clearStateOnAssemblies(assems, blockParamNames=None, cache=True):
 
         cachedBlockParamData = collections.defaultdict(dict)
 
-        blocks = []
-        for a in assems:
-            blocks.extend(a.getBlocks())
-        for b in blocks:
+        if len(assems) > 0:
```
jakehader, line 607: You might be able to save an indent if you do

```
if not assems:
   return cachedBlockParamData
```

armi/reactor/converters/uniformMesh.py
```diff
@@ -853,48 +913,13 @@ def _setParamsToUpdate(self, direction):
                     if not name in excludedParamNames
                 ]
             )
+        if direction == "in":
+            self.blockParamNames.extend(["molesHmBOL", "massHmBOL"])
```
jakehader, line 917: Should these be added to the parameter categories or why does it make sense to have this here?

armi/reactor/blocks.py
mgjarrett: This implementation did become specific to a `HexBlock`, but it's maybe a functionality that should exist for other types of blocks also.

armi/reactor/composites.py
mgjarrett: Yes. This is small compared to some of the other changes, but `composites.getNumberDensity()` calls 

* `getNuclideNumberDensities()` if the the `composite` is not a `component`, which then calls `getNumberDensities()` on the child components
*  `self.p.numberDensities.get(nucName, 0.0)` if it is a `component`


https://github.com/terrapower/armi/blob/15a710f0620ca9521370f11b828de352f7b939e4/armi/reactor/composites.py#L1234-L1235

armi/reactor/converters/uniformMesh.py
```diff
@@ -415,18 +421,24 @@ def makeAssemWithUniformMesh(
                 sourceBlock = blocks[0]
                 xsType = blocks[0].p.xsType
 
-            block = copy.deepcopy(sourceBlock)
+            block = sourceBlock.createBlankCopy()
             block.p.xsType = xsType
             block.setHeight(topMeshPoint - bottom)
             block.p.axMesh = 1
+            block.p.zbottom = bottom
```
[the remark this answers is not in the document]
mgjarrett: It would if the `block` was added to `newAssem` before `block.setHeight()` is called. Maybe it should be rearranged to make that happen.

armi/reactor/converters/uniformMesh.py
mgjarrett: They're already in the `depletion` category.

1. We could add `depletion` to the categories to be mapped, but then we'd also get a bunch of other parameters that we don't need.
2. We could create a new category `heavyMetalMass` or something along those lines and use it here. But I think we should avoid creating too many parameter categories for very specific use cases.

I'm trying to balance having a reasonable number of categories against having explicit parameter mapping specifications like this. If it was 10 parameters I would create a category, but for two parameters it seems better to just explicitly name them.

armi/materials/mixture.py
```diff
@@ -0,0 +1,30 @@
...
+from armi import materials
+
+
+class Mixture(materials.Material):
+    """
+    Homogenized mixture of materials
+
+    .. warning:: This class is meant to be used for homogenized block models for neutronics and other
+       physics solvers.
+    """
```
john-science, line 28: It feels like we should mention this is her for performance reasons, and not physics one.  

Opinion?

armi/physics/neutronics/parameters.py
```diff
@@ -579,6 +579,7 @@ def linPowByPinGamma(self, value):
     with pDefs.createBuilder(
         default=0.0,
         location=ParamLocation.AVERAGE,
+        categories=[parameters.Category.detailedAxialExpansion],
```
john-science, line 582: Is this related to this PR, or just a bonus?

armi/reactor/blocks.py
```diff
@@ -1519,6 +1533,58 @@ def coords(self, rotationDegreesCCW=0.0):
             round(y, units.FLOAT_DIMENSION_DECIMALS),
         )
 
+    def createCopy(self):
+        """
+        Create a blank copy of a block (ligher weight than deepcopy)
```
john-science, line 1538: "lighter weight" might be more clear if it said "faster".  Just IMO. I won't fight for it.

armi/reactor/blocks.py
```diff
@@ -1519,6 +1533,58 @@ def coords(self, rotationDegreesCCW=0.0):
...
+
+        b = self.__class__(self.getName(), height=self.getHeight())
+        b.setType(self.getType())
+
+        # assign macros and LFP
+        b.macros = self.macros
+        b._lumpedFissionProducts = self._lumpedFissionProducts
+        b.p.buGroup = self.p.buGroup
+
+        hexComponent = Hexagon(
```
john-science, line 1553: Sorry, stupid question, does this work for non-hex blocks?

Because 2/3 of our users don't have hex blocks.

armi/reactor/converters/uniformMesh.py
```diff
@@ -400,13 +405,13 @@ def makeAssemWithUniformMesh(
...
                     totalHeight = sum(typeHeight.values())
-                    runLog.extra(
+                    runLog.debug(
                         f"Multiple XS types exist between {bottom} and {topMeshPoint}. "
                         f"Using the XS type from the largest region, {xsType}"
                     )
                     for xs, h in typeHeight.items():
                         heightFrac = h / totalHeight
-                        runLog.extra(f"XSType {xs}: {heightFrac:.4f}")
+                        runLog.debug(f"XSType {xs}: {heightFrac:.4f}")
```
john-science, line 414: I think we need more than this.

Just because a single log line was useful to someone 8 years ago, doesn't mean anyone has looked at it in the last 7 years.

And I _definitely_ think we have a lot of `debug` log lines masquerading as `extra` and even `info`.

Nice!

armi/reactor/converters/uniformMesh.py
john-science: Could we at least save those strings off as constants at the top of the file?

```python
HEAVY_METAL_PARAMS = ["molesHmBOL", "massHmBOL"]
```

Or some such.

armi/reactor/blocks.py
```diff
@@ -157,6 +158,19 @@ def __deepcopy__(self, memo):
...
+    def createCopy(self):
+        """
+        Create a copy of a block
+
+        Used to implement a copy function for a specific block types that is lighter
+        than a deepcopy. This base class implementation is just a deepcopy.
+
+        Notes
+        -----
+        Only impelemented for HexBlock right now.
```
jakehader, line 170: `implemented`

armi/reactor/blocks.py
jakehader: I think that this is addressed on the general block as just a deepcopy now.

armi/materials/mixture.py
mgjarrett: Yes, I will add more detail to this docstring.

armi/physics/neutronics/parameters.py
mgjarrett: It's part of this PR. Some of these reaction rate parameters need to be mapped when performing gamma transport using a downstream plugin.

Before this PR, the uniform mesh converter would `deepcopy` blocks from the source assembly to the destination assembly in `makeAssemWithUniformMesh`, and these parameters would have values acquired from the `deepcopy`. Now that we have replaced the `deepcopy` with a less comprehensive copy function, the parameters need to be explicitly mapped so that the values are present on the uniform mesh reactor.

armi/reactor/blocks.py
mgjarrett: I agree. I added detail to the docstring to make it clear what this is for and why it would be used.

I originally said "lighter weight" to indicate that the block created from this function has less data associated with it than one created using `deepcopy`. But yes, the main goal is for `createCopy` to be significantly faster than a full `deepcopy` of the block.

armi/reactor/blocks.py
mgjarrett: That's correct @jhader. A faster/lighter copy function could be written for Cartesian or other blocks and it would probably look quite similar to this one, but for now I'm trying to minimize the scope of the change.

I don't have enough knowledge the use cases for other blocks to write a good `createCopy` that generalizes to all block types, but someone who uses another block type frequently should be able to implement a function modeled on this one without much trouble.

armi/reactor/converters/uniformMesh.py
mgjarrett: Thanks! Unfortunately, I think I'm the one who initially put this in as an `extra` just a couple months ago. I didn't realize how much obnoxious output it creates until I started testing with `detailedAxialExpansion: true`

armi/reactor/blocks.py
```diff
@@ -1519,6 +1536,84 @@ def coords(self, rotationDegreesCCW=0.0):
...
+        # only works if there is a single clad component that has a MultiIndexLocation
+        if self.hasComponents(Flags.CLAD):
+            pinComponent = Circle(
+                "voidPin",
+                "Void",
+                self.getAverageTempInC(),
+                self.getAverageTempInC(),
+                0.0,
+            )
+            clad = self.getComponent(Flags.CLAD)
```
keckler, line 1601: This is causing an error for me.

I have blocks that have multiple components with the `CLAD` flag, because I have some assemblies that have multiple fuel types loaded into it.

keckler: I guess you note this in a comment above...

armi/reactor/blocks.py
```diff
@@ -1587,27 +1587,29 @@ def _createHomogenizedCopy(self):
...
-            b.add(pinComponent)
+        if pinSpatialLocators:
+            # create a null component with cladding flags and spatialLocator from source block's
+            # clad component in case pin locations need to be known for physics solver
+            # only works if there is a single clad component that has a MultiIndexLocation
+            if self.hasComponents(Flags.CLAD):
+                cladComponents = self.getComponents(Flags.CLAD)
+                for i, clad in enumerate(cladComponents):
+                    pinComponent = Circle(
+                        "voidPin{i}",
```
keckler, line 1598: Does this need to be an f-string?

armi/materials/mixture.py
```diff
@@ -0,0 +1,41 @@
...
+# limitations under the License.
+
+"""
+Homogenized mixture material
+"""
+
+from armi import materials
+
+
+class Mixture(materials.Material):
```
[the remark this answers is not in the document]
mgjarrett: I would probably agree that we generally wouldn't want users to use this material. There could be a situation where a user has homogenized number densities for a material they want to model, but I guess the user could just create a custom material for that. Custom materials are a little outside my typical user experience with ARMI, so I would defer on this.

armi/reactor/blocks.py
```diff
@@ -1587,27 +1587,29 @@ def _createHomogenizedCopy(self):
...
-                for i1, i2 in zip(
-                    list(pinComponent.spatialLocator), list(clad.spatialLocator)
-                ):
-                    i1.associate(i2.grid)
-            pinComponent.setDimension("mult", clad.getDimension("mult"))
-            b.add(pinComponent)
+        if pinSpatialLocators:
+            # create a null component with cladding flags and spatialLocator from source block's
+            # clad component in case pin locations need to be known for physics solver
+            # only works if there is a single clad component that has a MultiIndexLocation
```
keckler, line 1593: Comment is out of date

armi/reactor/blockParameters.py
```diff
@@ -93,14 +93,12 @@ def detailedNDens(self, value):
...
             units="mole",
             description="Total number of atoms of heavy metal at BOL assuming a full assembly",
-            default=None,
         )
 
         pb.defParam(
             "massHmBOL",
             units="grams",
             description="Mass of heavy metal at BOL",
-            default=None,
```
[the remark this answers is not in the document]
mgjarrett: These are scalar values, so the default should be 0.0 rather than `None`. I don't have it documented, but I encountered errors with the defaults as `None`.

armi/reactor/blocks.py
mgjarrett: yeah...

armi/reactor/blocks.py
```diff
@@ -1519,6 +1535,85 @@ def coords(self, rotationDegreesCCW=0.0):
...
+        hexComponent = Hexagon(
+            "homogenizedHex",
+            "Mixture",
+            self.getAverageTempInC(),
+            self.getAverageTempInC(),
+            self._pitchDefiningComponent[1],
+        )
+        emptyNDens = {nuc: 0.0 for nuc in self.getNuclides()}
+        hexComponent.setNumberDensities(emptyNDens)
+        b.add(hexComponent)
```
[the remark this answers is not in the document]
mgjarrett: I renamed it to `_createHomogenizedCopy()`.

It's a bit of a misnomer for the base method `Block._createHomogenizedCopy()`, which just performs a `deepcopy` of the block, but that's probably okay. The non-homogenized block will still perform the correct functions, it will just be less efficient than `HexBlock._createHomogenizedCopy()`.

armi/reactor/blocks.py
mgjarrett: We spoke offline; updated to handle multiple clad components

armi/reactor/converters/uniformMesh.py
```diff
@@ -680,6 +699,9 @@ def _computeAverageAxialMesh(self):
         refNumPoints = len(src.core.findAllAxialMeshPoints([refAssem])) - 1
         allMeshes = []
         for a in src.core:
+            if a.hasFlags(Flags.CONTROL):
+                # don't include control assemblies; their mesh lines move a bunch
+                continue
```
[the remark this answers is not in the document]
mgjarrett: Removed

armi/reactor/converters/uniformMesh.py
```diff
@@ -307,13 +312,13 @@ def applyStateToOriginal(self):
         runLog.extra(
             f"Parameter remapping time: {completeEndTime-completeStartTime} seconds"
         )
-        self.reset()
```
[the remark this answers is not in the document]
mgjarrett: There is a `meshConverter.reset()` call after `applyStateToOriginal` is called here:

https://github.com/terrapower/armi/blob/9d6a405f10a0084b2fea8721f2b6d08c16a0b7a8/armi/physics/neutronics/globalFlux/globalFluxInterface.py#L485

The same pattern exists in the one place I could find `applyStateToOriginal` in internal code.

You could make the argument that it's better to keep the one that's within `applyStateToOriginal` because then we wouldn't ever forget to call it. But I don't see why it's absolutely necessary to run `reset()` immediately after `applyStateToOriginal`. They seem like independent actions, so I think it's okay to keep them separate.

armi/reactor/converters/uniformMesh.py
```diff
@@ -171,6 +176,7 @@ def convert(self, r=None):
                     assem,
                     self.convReactor.core.p.axialMesh[1:],
                     self.blockParamNames,
+                    self.bpm,
```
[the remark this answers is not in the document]
mgjarrett: I did, and I also updated the name to `self.paramMapper` to be more clear, and moved `self.blockParamNames` and `self.reactorParamNames` into the `self.paramMapper`

armi/physics/neutronics/parameters.py
mgjarrett: Yes, I think we really only need `rateFis` and `rateProdFis` for the particular use case that I'm thinking of. But there could be other cases where other reaction rates are desired. I didn't know exactly where to draw the line so I just included all of them instead of being picky. But if we're really trying to drill down on performance then we could limit it to just those fission rate parameters.

armi/reactor/blocks.py
```diff
@@ -1519,6 +1535,85 @@ def coords(self, rotationDegreesCCW=0.0):
...
+        by omitting this detailed data and only providing the necessary level of detail for
+        the uniform mesh reactor: number densities on each block.
+
+        .. note: Individual components within a block can have different temperatures, and this
+        can affect cross sections. This temperature variation is captured by the lattice physics
+        module. As long as temperature distribution is correctly captured during cross section
+        generation, it doesn't need to be transferred to the neutronics solver directly through
+        this copy operation.
+
+        .. note: If you make a new block, you must add it to an assembly and a reactor.
```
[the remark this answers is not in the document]
jakehader: I think it does!

armi/reactor/blocks.py
john-science: Yeah, we will need to find a way to fix this.  I don't like the idea that we are forcing the uniform mesh to only work for hex assemblies now.  (Because this is used pretty high-level in `makeAssemWithUniformMesh()`).

armi/reactor/blocks.py
```diff
@@ -1570,6 +1570,7 @@ def _createHomogenizedCopy(self, pinSpatialLocators=False):
 
         b = self.__class__(self.getName(), height=self.getHeight())
         b.setType(self.getType())
+        b.p.flags = self.p.flags
```
keckler, line 1573: I think setting the flags from within the `setType()` call itself would be more efficient, because it would avoid the internal call to `Flags.fromStringIgnoreErrors(typ)`.

https://github.com/terrapower/armi/blob/main/armi/reactor/composites.py#L757-L784

keckler: But either should work. I'm testing the way that you have it right now.

mgjarrett: Updated!

armi/reactor/blocks.py
mgjarrett: It still works with non-hex assemblies, it just uses `copy.deepcopy(block)`. It's the same behavior that existed before, which is less efficient but still works.

keckler:
Just putting this here to help back up this PR.

I've been testing locally with the version at the following hash: https://github.com/terrapower/armi/pull/1042/commits/81f1a81cb7891987777c522f0c49171fbce2e69f

This appears to speed up my calculation _a lot_. A calculation which previously took nearly 5 days is on track now to take only about 14 hrs.

A little bit of information about that run to help put that in context:
- I am using DIF3D/VARIANT P1P0 and MC2, so not the heaviest physics kernels
- I am doing critical control rod iterations at every step
- Anywhere between 1-4 depletion steps at each cycle, for nearly 50 cycles
- No other physics
- `detailedAxialExpansion: True`

This run was previously spending a lot of time doing uniform mesh conversions. That time has been reduced substantially, so the overall calculation time is now reduced by a factor of ~8-10.

Awesome!!

keckler (commented):
The changes _look_ good to me. I am running a test case and will comment again with the results of that.

keckler:
My test case worked well here. No complaints from me!

jakehader (approved):
Looks like this is functional and I propose to merge as soon as possible if @keckler and @john-science agree.

armi/reactor/blocks.py
```diff
@@ -1518,6 +1534,87 @@ def coords(self, rotationDegreesCCW=0.0):
...
+        b.setType(self.getType(), self.p.flags)
+
+        # assign macros and LFP
+        b.macros = self.macros
+        b._lumpedFissionProducts = self._lumpedFissionProducts
+        b.p.buGroup = self.p.buGroup
+
+        hexComponent = Hexagon(
+            "homogenizedHex",
+            "_Mixture",
```
keckler, line 1580: @mgjarrett Can you help me understand why this new `_Mixture` material is needed here? I see that it is just a subclass of `Material` with no differences. Couldn't we have just built a Hexagon with `Material` instead?

PR #1046, opened 2022-12-20
Bug fix for global material namespace order

albeanth:
## Description
This PR fixes a bug in which the global material namespace order was being set in test_axialExpansionChanger. This namespace change is for testing only local to this file. So while tests are run, we change it, but then via `tearDown` we reset the namespace order. This prevents downstream issues where the namespace is incorrectly set to this testing order. 

---

## Checklist

### DIFF
--- a/armi/reactor/converters/tests/test_axialExpansionChanger.py
+++ b/armi/reactor/converters/tests/test_axialExpansionChanger.py
@@ -45,7 +45,7 @@ from armi.tests import mockRunLogs
 from armi.utils import units
 
 
-class Base(unittest.TestCase):
+class AxialExpansionTestBase(unittest.TestCase):
     """common methods and variables for unit tests"""
 
     Steel_Component_Lst = [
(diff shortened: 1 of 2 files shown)


opotowsky (approved):
Wowza, nice catch! 🎣 

Looks like a really straightforward fix.

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -39,16 +39,11 @@
...
 from armi.tests import mockRunLogs
 from armi.utils import units
 
-# set namespace order for materials so that fake HT9 material can be found
-materials.setMaterialNamespaceOrder(
-    ["armi.reactor.converters.tests.test_axialExpansionChanger", "armi.materials"]
-)
-
 
 class Base(unittest.TestCase):
```
john-science, line 48: In retrospect, I think `Base` is too generic a name.  Tests are importable from ARMI, so this means the ARMI API has one "Base" class people could use, and it's this one. I expect that would surprise the users.

Can we rename this `AxialExpansionTestBase` (or something similar)?

albeanth: Great idea. I like it. 
https://github.com/terrapower/armi/pull/1046/commits/f5226c475ba0ebf6856c0a5d7556859a58b9fd47

PR #1054, opened 2022-12-30
Restart docs

### LINKED ISSUE #1000: We should have a "how to do a restart case" page in the docs
Occasionally I am asked how to run a restart case. As far as I know, a page describing this process does not exist in our docs.

I think we should have a simple page in the docs that describes the various settings and options that one has when performing a restart.

keckler:
## Description

This closes #1000 .

---

## Checklist


keckler:
@jakehader @monufer @ntouran Can I get a brief look at this from you all to confirm that the info added in there is still correct?

Some of it came from our old internal docs, and it's not clear to me if it is still true or not.

john-science (approved):
I like it.

Chris, I'll leave it to your judgement, but could/should we have another `.. note:` at the bottom, where we mention that if you restart a run on a different commit than it was run, you can get strange/undefined behavior? 

I know that might be obvious to you, but I just feel like maybe it bears repeating. The database does not include the code, and ARMI always has new users that will have to encounter that idea.

PR #1056, opened 2023-01-03
Move output cache runLog level from extra to info

opotowsky:
## Description

Just a tiny update to get more information automatically printed to a stdout for a case run (rather than requiring a verbosity update).

---

## Checklist

### DIFF
--- a/armi/utils/outputCache.py
+++ b/armi/utils/outputCache.py
@@ -169,7 +169,7 @@ def store(exePath, inputPaths, outputFiles, cacheDir):
         cachedLoc = os.path.join(folderLoc, baseName)
         shutil.copy(outputFile, cachedLoc)
 
-    runLog.extra("Added outputs for {} to the cache.".format(exePath))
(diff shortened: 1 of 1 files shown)


opotowsky:
@onufer Or did you think moving this one from extres to info would be more useful?

https://github.com/terrapower/armi/blob/52c3717e72d6587eb023717104c0c3d42d968ea5/armi/utils/outputCache.py#L71-L73

Not sure we need both

onufer (approved):

onufer:
> @onufer Or did you think moving this one from extres to info would be more useful?
> 
> https://github.com/terrapower/armi/blob/52c3717e72d6587eb023717104c0c3d42d968ea5/armi/utils/outputCache.py#L71-L73
> 
> Not sure we need both

we definitely need the first. the second one would also be useful probably just to remind folks its being used

jakehader:
Coverage is broken with fix in #1059. This won't impact coverage so landing anyways.

PR #1062, opened 2023-01-04
Refactoring Material to not subclass Composite

john-science:
## Description

This PR refactors the `Material` class to no longer be a subclass of `Composite`.

It also removes (most, not all) of the statefulness in ARMI Materials. For instance, I removed all `Parameter`s from `Material`.  Some of those I copied over to a simpler class attribute (`mat.p.massFrac` became `mat.massFrac`).  And some I was able to delete entirely (like `yieldStrength` and `atomFracDenom`).

Also, I should note that three parameters (`uFrac`, `puFrac`, and `zrFrac`) applied only to FUEL materials, but were defined by default for all materials.  That was fixed.

Stateful things I haven't removed in this PR (because they are going to be tricky):

* ARMI supports `Custom` materials, which are obviously stateful by design.
* ARMI supports custom isotopics for fuels, which again is a stateful idea.
* The `Material` class still has a `self.parent`, which is necessary for a whole LOAD of logic to do with `Material.getNuclides()`. (@ntouran Please note this proved to be a big enough change that if we want to remove it, it will be it's own PR.)

@drewj-usnctech @sombrereau  In any repo I can see, I am making PRs to match this one. But if you have a private repo I can't see, 99% of the changes are just to switch calls to `mat.p.thing` to `mat.thing`. (Once, a material outside ARMI needed an `__init__()` method, that it didn't have before.)

---

## Checklist


john-science:
@drewj-usnctech Just FYI

drewj-usnctech:
Thanks @john-science 

Quick skim and nothing seems to jump out to me here

armi/materials/material.py
```diff
@@ -32,16 +33,29 @@
...
 from armi.utils.units import getTk, getTc
 
 # globals
 FAIL_ON_RANGE = False
 
 
-class Material(composites.Composite):
+class MaterialMetaType(resolveCollections.ResolveParametersMeta):
+    """
+    Metaclass for automatically defining ParameterCollection classes.
```
ntouran, line 47: Would it make more sense to not use parameters at all for this? Like, just use private variables on the materials that have some extra attributes?

If not, we need to add some more implementation documentation here explaining.

* Even though Material is not an ArmiObject, it still has params
* Are these params stateful? Do they make it into and out of the db?
* What guidelines to devs should we put here?

armi/materials/material.py
```diff
@@ -73,7 +87,8 @@ class Material(composites.Composite):
     with information about thermal scattering."""
 
     def __init__(self):
-        composites.Composite.__init__(self, self.__class__.name)
+        self.parent = None
```
ntouran, line 90: Why do materials still have parents? They shouldn't, since they're no longer in the hierarchical tree.

Remove or justify with implementation docs.

armi/materials/material.py
```diff
@@ -101,6 +122,26 @@ def getChildrenWithFlags(self, typeSpec: TypeSpec, exactMatch=True):
...
+        """Empty because statelss materials shouldn't use cache"""
+        pass
+
+    def _getCached(self, name):
+        """Empty because statelss materials shouldn't use cache"""
+        return None
+
+    def _setCache(self, name, val):
+        """Empty because statelss materials shouldn't use cache"""
+        pass
```
ntouran, line 143: Can we just delete all these?

Spelling error: `stateless`

Big picture: have you measured the performance impact of not caching the material properties vs. temperature? That was a hugely major performance bottleneck we identified through formal profiling of test cases. 

We need to see a profiler trace of loading a reactor and printing out all number densities in order to judge this. Can you paste the before/after highlights of the profiler results as a comment to this PR?

armi/materials/mgO.py
```diff
@@ -30,8 +30,8 @@ class MgO(Material):
     def __init__(self):
         Material.__init__(self)
         """same reference as linear expansion. Table II.
-        Reference density is from Wolfram Alpha At STP (273 K)"""
-
+        Reference density is from Wolfram Alpha At STP (273 K)
+        """
```
ntouran, line 34: while editing, might as well bring up to formatting standards. 

Single line starting with capital letter, ending with `.`, blank line, additional info.

armi/materials/molybdenum.py
```diff
@@ -23,7 +23,7 @@ class Molybdenum(SimpleSolid):
     name = "Molybdenum"
 
     def setDefaultMassFracs(self):
-        """Moly mass fractions."""
+        """Moly mass fractions"""
```
ntouran, line 26: The period is generally required by docstring checkers.

armi/materials/thoriumOxide.py
```diff
@@ -106,8 +107,6 @@ def meltingPoint(self):
 
 
 class ThO2(ThoriumOxide):
-    """
-    Another name for ThoriumOxide.
-    """
+    """Another name for ThoriumOxide"""
```
ntouran, line 110: Needs  period. (we need to make pydocstyle part of our auto-linting! http://www.pydocstyle.org/en/stable/)

ntouran (changes requested):

armi/materials/material.py
john-science: It _appears_ to me that the mass fractions in all ARMI materials are designed to be changed by the user's choice of nuclides.  And right now those nuclide choices come from the parent.

Can you enlighten me on this?  Because these nuclides are also the reason materials have parameters.

ARMI materials seem to change a LOT based on the nuclides present in ARMI, and those nuclide are not static in ARMI.

armi/materials/molybdenum.py
john-science: Wait, really?

Which docstring checkers?  I've never seen that. They render fine in our docs.

The period there is grammatically incorrect.  I've always prescribed to the "don't put punctuation in code comments, unless they are complete sentences".  Which... I'm sure is a rule I have applied far and wide in ARMI.

armi/materials/material.py
john-science: Two questions:

1. Isn't a cache a "state" for the class?
2. In downstream repos, we have temperature-dependent densities.  These caches will preserve the first value they get forever. Shouldn't that lead to incorrect material densities?

armi/materials/material.py
john-science: Please see my question/note below about nuclides.

armi/materials/material.py
ntouran: Materials used to be *the place* where composition (defined as number density vectors of elements and nuclides) was stored. We moved that to the `Component`'s `numberDensities` param long ago. 

However, we still initialize this param based on the `Material`s that users add to their problem via the `Material.setDefaultMassFracs` method, which sets `Material.p.massFrac`. This is indeed (and unfortunately) a bit coupled to nuclide choices, since sometimes users want to model *Carbon* while other times they want to model *Carbon-12*. 

So may be we need an intermediate ticket here, with the goal of: "Decouple composition definitions from neutronics library assumptions". 

Material definitions _should_ be static and evergreen, not dependent in any way on user modeling settings. This should include initial material composition, and probably also the process of building Component `numberDensities`, and potentially even the evolution of composition vs. time from transmutation and decay and reprocessing and everything else. The user modeling settings that map compositions to specific subsets of nuclides based on neutronics library needs should be isolated to the neutronics physics modules. For example, if a neutronics input is being built for a code that only has C-12 but the composition has Carbon, a mapping between carbon and C-12 should happen as the input is being written.

At this point, changing `setDefaultMassFracs` to `getDefaultMassFracs` in a way that just returns the data might be workable, and just eliminate `Material.p.massFrac` altogether. We'll have to see how much code accesses `p.massFrac` and assess how exactly to pull off an upgrade like this.

john-science: Okay, I have made progress. I have removed all `Parameter`s from `Material` and `FuelMaterial` except:

* `massFrac`
* `puFrac`
* `uFrac`
* `zrFrac`

Frankly, those last three seem like a terrible idea to have in EVERY MATERIAL anyway.  

In the process of doing this cleaning I also fixed a couple of latent bugs in the materials, so that nice.

I'm hoping I'll remove these last 4 Paramters today, and we can re-open this PR.

armi/materials/material.py
john-science: I will look into this more once I have fully removed the Parameters from `Material`s.

john-science: Done.

Since `Material` no longer subclasses `Composite`, I had to mostly copy them over from the `Composite` class.  But now they aren't blank here any more.

armi/materials/material.py
john-science: Done.

Okay, I removed a lot of code, so I'm sure the code coverage will go down. BUT `Material` not longer has any `Parameter`s.

armi/materials/thoriumOxide.py
john-science: fixed

armi/materials/molybdenum.py
john-science: fixed

armi/materials/mgO.py
john-science: fixed

drewj-usnctech:
>  In any repo I can see, I am making PRs to match this one. But if you have a private repo I can't see, 99% of the changes are just to switch calls to mat.p.thing to mat.thing. (The only other change I've seen is a couple materials outside ARMI needed an __init__() method, that they didn't before.)

Thanks for the heads up @john-science 👍

armi/materials/custom.py
```diff
@@ -40,7 +40,7 @@ def __init__(self):
         by a constant user-input density.
         """
         Material.__init__(self)
-        self.p.density = 1.0
+        self.customDensity = 1.0
```
ntouran, line 43: kind of a minor point but if we're renaming things like this it might be wise to make them supernames like `customDensityInGPerCC`. As it turns out, putting units in names is one honkin good idea, especially for super generic terms like 'density'

armi/materials/material.py
ntouran: ok, please just add docs in the code explaining why there's a parent and what's to be done about it.

armi/materials/material.py
```diff
@@ -464,11 +482,11 @@ def getLifeMetalConservativeFcciCoeff(self, Tk: float) -> float:
 
     def yieldStrength(self, Tk: float = None, Tc: float = None) -> float:
         r"""returns yield strength at given T in MPa"""
-        return self.p.yieldStrength
+        pass
 
     def thermalConductivity(self, Tk: float = None, Tc: float = None) -> float:
         r"""thermal conductivity in given T in K"""
-        return self.p.thermalConductivity
+        pass
```
ntouran, line 489: should be NotImplementedError? If not, explain in code comments.

armi/materials/material.py
```diff
@@ -619,6 +628,7 @@ def densityTimesHeatCapacity(self, Tk: float = None, Tc: float = None) -> float:
 
     def getNuclides(self):
         warnings.warn("Material.getNuclides is being deprecated.", DeprecationWarning)
+        # NOTE: Material still has a self.parent because of this method.
```
ntouran, line 631: that is probably worth elaborating even further in the comments.

ntouran (approved):
looks great, thanks! Just some minor comments.

armi/materials/material.py
john-science: The `Material` class has `pass` methods, meant to be subclassed (and did before this PR).

For instance, `def applyInputParams()` is `pass` because if it was `NotImplementedError`, nearly all of the Materials would fail. The default functionality is to do nothing, because these `Material` methods get called willy-nilly by downstream code.

I tried, and `Material` is not an abstract class, but it's close. 

I can add some sort of comment to this affect.

john-science: In retrospect, this is part of the "materials need to be more consistent" issue. Discussed more [here](https://github.com/terrapower/armi/discussions/767).

This PR aims to make Materials less stateful, and to not subclass deposit.

But I can see your issue.  This PR won't fix _everything_ about our `Material`s, and speaking to the broad user base, most issues people have come down to consistency.

PR #1069, opened 2023-01-05
Cleaning out duplicate context import in bootstrap

john-science:
## Description

I ran some case suites and integration tests in downstream repos and was able to determine that we can safely remove this line.

**NOTE**: Several of the items on our ARMI Road Map would be easier if ARMI ran less code at import time. At some point, we will need to reduce the work done in `context.py` and `_boostrap.py` so that people can use ARMI in a more piecemeal fashion.

---

## Checklist

### DIFF
--- a/armi/_bootstrap.py
+++ b/armi/_bootstrap.py
@@ -34,9 +34,6 @@ if (
     )
 
 
-from armi import context
-
-
 def _addCustomTabulateTables():
     """Create a custom ARMI tables within tabulate."""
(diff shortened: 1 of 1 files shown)


john-science:
The code coverage change is because I removed one line of code that was previously covered.  Meh.

@jakehader Thanks for letting me test this change before merging it.

opotowsky (approved):
Hey, I like it!

PR #1067, opened 2023-01-05
Fission Product Model with Explicit Fission Products for all Depletable Components

jakehader:
## Description

Update the fission product model to fix a bug where all nuclides were not being initialized in non-fuel depletable components when using the `explicitFissionProducts` model. Add testing for the updated implementation.

It also ensures dummy nuclides are added for all XS types during `mergeXSLibrariesInWorkingDirectory`, with added testing.

---

## Checklist

- [x] This PR has only one purpose or idea.
- [x] Tests have been added/updated to verify that the new/changed code works.

- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html).

- [x] The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any bug fixes or new features.
- [x] The documentation is still up-to-date in the `doc` folder.
- [x] The dependencies are still up-to-date in `setup.py`.

PR #1081, opened 2023-01-10
XS Library Merge Bug

mgjarrett:
## Description

Two bugs introduced in #1067:

1. By changing to using an absolute file path for the current working directory, the XS library merge function was not correctly identifying all of the files to be merged. Needed to update regular expression to match the file names correctly with the full path included. Also needed to fix the processing related to suffixes so filepaths with dashes don't get excluded from the merge list. 

2. The FissionProductModel `interactBOC` was removed because the fission products are initialized on interactBOL. This was fine for standard runs but it caused snapShot runs to fail because the fission products were not initialized on the reactor loaded from the database.

---

## Checklist


opotowsky (approved):
Great work on the bug squash!

opotowsky:
There is one newly uncovered line in this PR from re-adding `interactBOC` (which wasn't tested before)

PR #1086, opened 2023-01-12
Defer setting material number in MCNP material card

mgjarrett:
## Description

This is a minor change that should not impact functionality of any existing code. This slight modification allows downstream code to generate MCNP inputs more effectively by avoiding duplication of identical materials many times over.

---

## Checklist

### DIFF
--- a/armi/utils/densityTools.py
+++ b/armi/utils/densityTools.py
@@ -192,7 +192,10 @@ def formatMaterialCard(
         for nuc in densities
     ):
         return []  # no valid nuclides to write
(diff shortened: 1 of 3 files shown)


mgjarrett:
The only potential pitfall is if somebody was calling this function and intentionally passing the default value `matNum=0`. `m0` is used in MCNP inputs to define a library to use for all materials. I don't think users would be using this function to create an `m0` card, but if they were then this would require them to make some minor changes.

onufer (approved):

PR #1087, opened 2023-01-13
Copy dummy micro XS.

mgjarrett:
## Description

Potential bug introduced by #1081 where the XS of dummy nuclides created in ISOTXS are not initialized. Most of them are approximately zero, as they should be, but sometimes a NaN value would show up in the ISOTXS file. This can cause an error when that microscopic XS is used in a reaction rate or macroscopic XS calculation.

---

## Checklist


opotowsky:
The coverage for this run isn't right (this PR wouldn't decrease coverage in `runLog.py`), but I can't kick it off again. This IS adding a line to a previously not-covered section of this function, though.

opotowsky:
I'm going to merge this even though the coverage run is messed up. There should be a MUCH smaller cov drop (one new uncovered line)

opotowsky (approved):

PR #1088, opened 2023-01-13
Add armi.reactor.compositeList.CompositeList : base class for Assemblies

### LINKED ISSUE #1030: Add a 1-D collection of composites to support assemblies and particle fuel compacts
Related to https://github.com/terrapower/armi/discussions/960 and #1008 

In some particle fuel compacts, there is some axial un-fueled regions where you just have matrix material. If we think about this as three regions
1. Lower cap
2. Matrix + particle fuel
3. Upper cap

Each of these could be a block-like thing containing one or more Components along a fixed axial grid. This is very similar to an assembly which means there's a chance for sharing code.

One could also use this to make a single few cm tall UO2 fuel pellet as a block-like thing, and then stack them into a fuel rod in a proper `Block` with this class

I'm proposing the creation of a special composite subclass that contains children along a fixed 1-D spatial grid. The grid should be monotonically increasing so iteration follows a natural order. It should support at least `__iter__` and `__getitem__` (hopefully in a consistent order -- #1018) There are probably other methods that make sense

The `Assembly` class would then inherit from this class. Most if not all of the parameters should be defined on this class first. Mainly to help w/ serialization. There might be some that make sense to move to the parent class but maybe later?

Possible names

- `CompositeStack` / `CompositeList`
  - maybe implies mutability more than desired?
- `CompositeArray`
  - maybe implies N-dimensions?
- `CompositeVector`

drewj-usnctech:
## Description

The intention of the abstract parent class is to shift most of the composite-tree traversal routines onto a generic parent class. It's abstract only because the Assembly does some re-naming of children during `Assembly.add` which subclasses might want to override. Might not be enough to justify all the metaclasses and just make that an empty method.

Later, another subclass of `CompositeList` could be used to model stacks of fuel pellets in a ``Block``. But that would require many blueprint side modifications that are to be discussed.

Some assembly methods like `getBlockAndZ` pass directly off to the parent method to maintain the same API. Wondering if there is room to make some of these iterator-based methods a generator? But that would also be an API change and should be done after some discussions.

Added a test for getBlocksAtElevation.

Closes #1030 

## Checklist

### DIFF
--- a/armi/reactor/compositeList.py
+++ b/armi/reactor/compositeList.py
@@ -1,0 +1,246 @@
+"""Collection of composites with a 1-D grid"""
+
+from abc import ABC, ABCMeta, abstractmethod
+from typing import Optional, Iterator, Tuple
+
+from numpy import empty, ndarray
+
+from armi import runLog
+from armi.reactor import flags
+from armi.reactor.grids import Grid, axialUnitGrid
+from armi.reactor.composites import Composite, CompositeModelType
+
+
+class AbstractCompositeMetaclass(ABCMeta, CompositeModelType):
+    """Metaclass that allows for abstract composites"""
+
+
+class AbstractComposite(ABC, Composite, metaclass=AbstractCompositeMetaclass):
+    """Composite that can have abstract methods
+
+    Useful for enforcing interfaces across generic subclasses
+    """
+
+
+class CompositeWithHeight(AbstractComposite):
+    """Interface for a composite with a height.
+
+    Like a block but more general
+
+    """
+
+    @abstractmethod
+    def getHeight(self) -> float:
+        """Return the height of this object"""
+
+
+class CompositeList(CompositeWithHeight):
+    """Axial collection of composites
+
+    Many things in an ARMI :class:`~armi.reactor.reactors.Reactor`
+    can be seen as things stacked on things. An
+    :class:`~armi.reactor.assemblies.Assemblies` is a 1-D stack of
+    :class:`~armi.reactor.blocks.Block` instances, each with an axial
+    location and height. Fuel pins are another concept that would benefit
+    from a 1-D stack of :class:`~armi.reactor.composites.Composite` like things.
+
+    This class seeks to separate out storing and maintaing the axial mesh
+    of things (iteration, item access, mesh updating, etc.) from
+    assembly-specific tasks (moving assemblies within the reactor,
+    looking for coolant channels, etc.)
+
+    """
+
+    spatialGrid: Optional[Grid]
+
+    def __init__(self, name):
+        super().__init__(name)
+        self.spatialGrid = None
+
+    def __iter__(self) -> Iterator[CompositeWithHeight]:
+        """Iterate over all children"""
+        return super().__iter__()
+
+    def __getitem__(self, index: int) -> CompositeWithHeight:
+        """Return the child at a specific index"""
+        return super().__getitem__(index)
+
+    def add(self, obj: CompositeWithHeight):
+        """Add an item to the end of the list
+
+        .. note::
+
+            Object **must** at least have a ``getHeight`` method
+            to understand how the spatial grid needs to be updated
+
+        Parameters
+        ----------
+        obj : Composite
+            Object to be added to the end.
+
+        """
+        super().add(obj)
+        nItems = len(self)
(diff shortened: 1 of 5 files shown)


armi/reactor/compositeList.py
```diff
@@ -0,0 +1,246 @@
...
+    of things (iteration, item access, mesh updating, etc.) from
+    assembly-specific tasks (moving assemblies within the reactor,
+    looking for coolant channels, etc.)
+
+    """
+
+    spatialGrid: Optional[Grid]
+
+    def __init__(self, name):
+        super().__init__(name)
```
john-science, line 57: Style points thing.

For reasons that pre-date my involvement in this project, we have [a style guide in the docs](https://terrapower.github.io/armi/developer/standards_and_practices.html?highlight=super#general-do-s-and-don-ts) that says to not use `super()`, in favor of more explicit calls to the super class `__init__()` function.

Just to be entirely clear, we use `super()` all over ARMI, and no one has bothered to fix it yet.  So, it's not mission-critical. But probably we should at least leave the `Assembly` one as-is.

armi/reactor/compositeList.py
```diff
@@ -0,0 +1,246 @@
...
+        self.spatialGrid = axialUnitGrid(len(self), self)
+        for zi, child in enumerate(self):
+            child.spatialLocator = self.spatialGrid[0, 0, zi]
+            self._renameChild(child, zi)
+
+    @abstractmethod
+    def _renameChild(self, child: Composite, index: int):
+        """Rename a child at a given index in the list"""
+
+    def calculateZCoords(self):
```
john-science, line 125: It would be cool if all these methods had their own unit tests.

armi/reactor/compositeList.py
drewj-usnctech: Roger. We will need to change the assembly `__init__` to not point to `Composite.__init__` instead the parent class, but I think the spirit of the request would still be preserved

armi/reactor/compositeList.py
drewj-usnctech: Excellent request. Since the `Assembly` is the primary concrete subclass, and it's well tested through `armi/reactor/tests/test_assemblies.py`, all if not most of these methods _should_ be tested.

The alternative approach could be to make another child subclass specific for testing the composite list methods. It would look similar to those in the assembly testing, but maybe not tied strictly to blocks? Ah but we need the children to have a `getHeight` method...

Thoughts?

drewj-usnctech:
Personal todo

- [ ] License header
- [ ] Changelog / release notes entry

armi/reactor/compositeList.py
john-science: Ah, you're counting on the `Assembly` subclasses unit tests. I see.

Well, in that case, your [unit test coverage](https://coveralls.io/builds/55940866/source?filename=armi%2Freactor%2FcompositeList.py) is pretty high for the new file (~91%).

drewj-usnctech: If you're alright with it, I'll extend the existing assembly tests to bump the coverage up. There's some edge cases in some pretty public methods (`getChildrenAndZ` -> `getBlocksAndZ`) that aren't touched

drewj-usnctech:
@john-science other than some conflicts in the release notes, is there anything else you're looking to see?

armi/reactor/compositeList.py
```diff
@@ -0,0 +1,260 @@
+# Copyright 2019 TerraPower, LLC
```
john-science, line 1: It's 2023 now!

hehe

armi/reactor/compositeList.py
drewj-usnctech: How do we feel about this in light of #1155?

armi/reactor/compositeList.py
drewj-usnctech: Updated copyright year

drewj-usnctech:
@john-science getting back to this. Is there anything you'd want to see before calling this change acceptable?

jakehader:
> @john-science getting back to this. Is there anything you'd want to see before calling this change acceptable?

@drewj-usnctech, @john-science bump!

drewj-usnctech:
Going to close this for now because it's dropped off our radar. We might re-investigate this after a few months

PR #1090, opened 2023-01-16
Write last time node of each cycle rather than EOC to database 

### LINKED ISSUE #1100: Operator should be responsible for advancing time
Right now, the the operator does not advance time. It does iterate through the interfaces at each time node, and call each interface but typically a specific interface has been responsible for advancing time. Internally, at TerraPower it has been our depletion interface. We realize that since the operator doesn't do it, other users of the framework may be advancing time differently, but it could be helpful to unify how we handle time.

Here is a table of how it is done for an example problem with 3 30-day depletion steps and a 10 day outage. I don't think its optimal though but providing as an initial strawman... see discussion below for why it is this way currently:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/monufer/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/monufer/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

<body link="#0563C1" vlink="#954F72">

Node | Time (days) | Written to DB
-- | -- | --
Beginning of Cycle(BOC) | 0 | no (same time as 0,0)
0,0 | 0 | x
0,1 | 30 | x
0,2 | 60 | x
0,3 | 90 | NO! (90 days isn't anywhere to be found)
End of Cycle(EOC) | 100 | x (as the 0,3 state point, but its not, its 10 days past..)

</body>

</html>

The first question is, Why is there no write at 0,3, but instead there is a write at EOC? This is so we can do restart runs from the end of a cycle. if you loaded from the data at 0,3 instead of EOC you and started the next cycle you would miss the decay step and the problem you would have run would diverge from the original by the decay step since currently when you load from 0,2 it jumps right into interact BOC for the next cycle (cycle 1 (1,0) ) rather than EOC from 0,2.

What's bad about this? If you want to retrieve the isotope distribution from 0,3 its not retrievable. only EOC is retrievable. Snapshot runs are much more interested in the 0,3 data than the EOC data. So the fundamental problem is the restarts want the EOC data but snapshots want the 0,3 data. We could potentially solve this by writing the 0,3 data instead of EOC and having restart run interact EOC when its restarting from the last node. I'm sure there are other solutions out there too such as redefining when the outage occurs... maybe it could be the first thing in a cycle and that might solve the misMatch, but its not super clear to me.

One solution that I discourage is writing both 0,3 AND EOC since this will increase the database size by a factor of ~1.25 (5/4) and haver very little valuable additional info. (was investigated here #1090 )

Just want to get discussion going.

onufer:
## Description

This PR Removes the `interactEOC` database write and forces the write at the last time step instead. The figure below shows the before and after. State points at red lines are written to database. State points with a circle around them are all at the same exact time in the reactor life (within the armi model). The bottom half is the new direction that this PR follows.

**Note**: After this PR, most databases will have 1 more time node, such that there is at least one accessible node for every time in the `Reactor`.

image

Close #1100 

---

## Checklist


albeanth:
@onufer what does this PR do? I can infer what the changes are to `loadOperator` in `armi/bookkeeping/db/__init__.py`, but it's not clear what's going on with the changes in `bookkeeping/db/databaseInterface.py`.

onufer:
closed in favor of #1100

john-science:
I can't wait!

We will want unit tests on this new feature at some point.  Should be straight forward.

john-science:
Can we add a release note under "API Changes"?

https://github.com/terrapower/armi/blob/b3986c182f50f2ea75605c0af690a5da10a39c79/doc/release/0.3.rst?plain=1#L20-L22

I would like it to say that we added a new time node, and slightly modified the meaning of one that existed. (Even if the change is small, this is the place where we tell people it happened.)

doc/release/0.3.rst
```diff
@@ -50,7 +50,8 @@ API Changes
 #. Removing the parameters ``outsideFuelRing`` and ``outsideFuelRingFluxFr``. (`PR#1700 <https://github.com/terrapower/armi/pull/1700>`_)
 #. Removing the setting ``doOrificedTH``. (`PR#1706 <https://github.com/terrapower/armi/pull/1706>`_)
 #. Changing the Doppler constant params to ``VOLUME_INTEGRATED``. (`PR#1659 <https://github.com/terrapower/armi/pull/1659>`_)
-#. Change ``Operator._expandCycleAndTimeNodeArgs`` to be a non-static method so it no longer needs to take arbitrary args (`PR#1766 <https://github.com/terrapower/armi/pull/1766>`_)
+#. Change ``Operator._expandCycleAndTimeNodeArgs`` to be a non-static method so it no longer needs to take arbitrary args. (`PR#1766 <https://github.com/terrapower/armi/pull/1766>`_)
```
onufer, line 53: just added period.

armi/bookkeeping/db/databaseInterface.py
```diff
@@ -151,36 +151,24 @@ def interactEveryNode(self, cycle, node):
         if self.o.cs["tightCoupling"]:
             # h5 cant handle overwriting so we skip here and write once the tight coupling loop has completed
             return
-        self.writeDBEveryNode(cycle, node)
+        self.writeDBEveryNode()
 
-    def writeDBEveryNode(self, cycle, node):
```
onufer, line 156: sending unused args in is unclear and frustrating for debugging. got rid of unsued.

armi/bookkeeping/db/__init__.py
```diff
@@ -78,7 +78,13 @@
 ]
 
 
-def loadOperator(pathToDb, loadCycle, loadNode, allowMissing=False):
+def loadOperator(
+    pathToDb,
+    loadCycle,
+    loadNode,
+    statePointName=None,
+    allowMissing=False,
```
john-science, line 86: I won't die on this hill, but we would have more backwards compatibility if the new method parameter (`statePointName`) came last.

armi/bookkeeping/db/databaseInterface.py
```diff
@@ -151,36 +151,24 @@ def interactEveryNode(self, cycle, node):
...
 
     def interactEOC(self, cycle=None):
-        """In case anything changed since last cycle (e.g. rxSwing), update DB. (End of Cycle)."""
-        # We cannot presume whether we are at EOL based on cycle and cs["nCycles"],
-        # since cs["nCycles"] is not a difinitive indicator of EOL; ultimately the
-        # Operator has the final say.
-        if not self.o.atEOL:
-            self.r.core.p.minutesSinceStart = (
-                time.time() - self.r.core.timeOfStart
-            ) / 60.0
```
john-science, line 175: Do we not want to update the time any more?

armi/bookkeeping/db/databaseInterface.py
john-science: I can confirm that no one calls this method outside ARMI.  I checked Nala and all submodules, Pylot, and all our benchmark reactors.

That makes this API change safely contained inside ARMI.

armi/bookkeeping/db/__init__.py
onufer: fair, done.

armi/bookkeeping/db/databaseInterface.py
onufer: we tend to only do it before db write since its not used anywhere in code.. not that it is stored in EOL anyway

john-science:
@onufer Can you re-run `ruff`? It looks like the linting CI is broken.

john-science:
When this PR is ready for final review, you can take it out of "Draft".

armi/bookkeeping/db/databaseInterface.py
john-science: Sorry, I still don't understand.

> we tend to only do it before db write since its not used anywhere in code.

So, did we add this somewhere else?

armi/bookkeeping/mainInterface.py
```diff
@@ -78,16 +78,25 @@ def _activateDB(self):
...
+            # will continue at the beginning of start cycle and start node, and the database contains
             # the values from the run at the end of the interface stack, which are what the start
             # start cycle and start node should begin with.
 
             # NOTE: this should be the responsibility of the database, but cannot because the
             # database is last in the stack and the MainInterface is first
             dbi.prepRestartRun()
+
+            if self.cs["startNode"] == 0:
+                # DB interface loaded the pervious time step (last time node of previous cycle), but
```
john-science, line 93:
```suggestion
                # DB interface loaded the previous time step (last time node of previous cycle), but
```

john-science (approved):

armi/bookkeeping/mainInterface.py
```diff
@@ -56,12 +56,12 @@ class MainInterface(interfaces.Interface):
...
         interfaces.Interface.interactBOL(self)
-        self._activateDB()
+        self._activateDBPrepRestart()
         self._moveFiles()
 
-    def _activateDB(self):
+    def _activateDBPrepRestart(self):
         """
-        Instantiate the database state.
+        Instantiate the database state, and add pervious time nodes for restart run.
```
keckler, line 64:
```suggestion
        Instantiate the database state, and add previous time nodes for restart run.
```

doc/release/0.3.rst
```diff
@@ -50,7 +50,8 @@ API Changes
...
+#. Database now writes state at the last time node of a cycle rather than during the `DatabaseInterface.interactEOC` interaction, making this node accessable since its the more useful node, and at a unique time. (`PR#1090 <https://github.com/terrapower/armi/pull/1090>`_)
```
keckler, line 54:
```suggestion
#. Database now writes state at the last time node of a cycle rather than during the `DatabaseInterface.interactEOC` interaction, making this node accessible since its the more useful node, and at a unique time. (`PR#1090 <https://github.com/terrapower/armi/pull/1090>`_)
```

keckler: Do we also want to add another sentence here that says something like "This will break backwards compatibility with using old databases for restart calculations from the last node in a cycle."? Seems like it would be useful to have such important information in the release notes.

keckler (approved):
Thanks Mark. This change was less painful to make than I expected.

armi/bookkeeping/db/databaseInterface.py
john-science: Mark's these is that we only need to update this parameter when we are writing to the DB, which is no longer happening here.

I can confirm that (in ARMI and the broader ecosystem), people are only using the parameter in reference to a DB.  If that helps.

onufer: yes i talked with john to explain. its only a valuable calculation when you are about to write a db. so only do the work then. its calculated at EOC and before all other writes 


ooooooHHHH, yes we still want it on the every node. not sure how that got lost.

onufer: its back now

doc/release/0.3.rst
onufer: done

keckler:
Sorry, last question -- Is there really not any documentation that needs to be updated with this change?

john-science:
> Sorry, last question -- Is there really not any documentation that needs to be updated with this change?

Well, as I noted somewhere (here or in the ticket?) there is no place in the docs where we explain the time steps in the DB. I _imagine_ that should go some where in `doc/user/input.rst`, but I'm not immediately sure _where_.

It's certainly a strange omission, but I think that needs to be fixed outside this PR. It does need to be fixed though.

Issue #1096, opened 2023-01-17
`getNumPins()` misinterprets some non-pinned blocks as having pins


dlangewisch:
As currently implemented, `armi.reactor.blocks.HexBlock.getNumPins()` misinterprets certain blocks that contain no pins but do contain other shapes of components with one or more of the flags currently listed in the `PIN_COMPONENTS` list. For example, an unpinned block containing a `SHIELD` component with shape `HoledHexagon` will be misinterpreted as a block containing 1 pin. 

https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/reactor/blocks.py#L1017-L1020

https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/reactor/blocks.py#L50-L58

PR #1098, opened 2023-01-18
Improve logic for getNumPins

mgjarrett:
## Description

#1096 

The logic for `Block.getNumPins()` will identify lots of different components as pins, whether or not they are indeed pins. If a component has `Flags.SHIELD`, for example, it will be considered a pin (no matter what the shape or intent of the actual component is).

It is a relatively safe bet to assume all pins are circular, so a filter on component shape is applied to weed out any non-pin components.

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_blocks.py
+++ b/armi/reactor/tests/test_blocks.py
@@ -23,7 +23,7 @@ import numpy
 from numpy.testing import assert_allclose
 
 from armi import materials, runLog, settings, tests
-from armi.reactor.components import basicShapes
+from armi.reactor.components import basicShapes, complexShapes
 from armi.nucDirectory import nucDir, nuclideBases
 from armi.nuclearDataIO.cccc import isotxs
 from armi.physics.neutronics import NEUTRON, GAMMA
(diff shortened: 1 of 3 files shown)


armi/reactor/tests/test_blocks.py
```diff
@@ -1242,6 +1242,11 @@ def test_getNumPins(self):
         emptyBlock = blocks.HexBlock("empty")
         self.assertEqual(emptyBlock.getNumPins(), 0)
 
+        comp = basicShapes.Hexagon("hexagon", "HT9", 1, 1, 1)
```
john-science, line 1245: I love it!

You've added so much logic to `getNumPins()` I wonder if we can test it a bit more.

For instance, can we have one test where it returns a NON-zero pin count, because it has a circular shape somewhere in it?  Perhaps we could do one with a `HoledRectangle`, and prove that the hole in the rectangle?

https://github.com/terrapower/armi/blob/20ca6877c2e2221bc3e3521cf9e827cf2d4a4376/armi/reactor/components/complexShapes.py#L94

mgjarrett: I added some more tests. Let me know if you think there are other corner cases we should cover!

john-science (approved):
Looks great!

PR #1099, opened 2023-01-18
Cleanup umc

mgjarrett:
## Description

Some subtle bugs in #1042 were missed until more detailed testing was performed.

1. There are two ways of getting the number of pins in a block: `block.getNumPins()` and `b.p.nPins`. `b.p.nPins` is set in the blueprints:

https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/reactor/blueprints/blockBlueprint.py#L180

It also needs to be set in `_createHomogenizedCopy`

2. `power` and `pdens` are initially set by the neutronics solver in a downstream workflow, but when gamma transport is performed these parameters are modified. Thus, they need to be mapped "out" from the `GammaUniformMeshConverter`. This is likely to be true of any implementation of a coupled neutronics/gamma transport solution.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/parameters.py
+++ b/armi/physics/neutronics/parameters.py
@@ -492,7 +492,10 @@ def _getNeutronicsBlockParams():
         )
 
         pb.defParam(
-            "pdens", units="W/cm$^3$", description="Average volumetric power density"
+            "pdens",
+            units="W/cm$^3$",
+            description="Average volumetric power density",
+            categories=[parameters.Category.neutronics],
         )
 
         pb.defParam(
@@ -620,7 +623,12 @@ def _getNeutronicsBlockParams():
             categories=[parameters.Category.gamma],
         )
 
-        pb.defParam("power", units="W", description="Total power")
+        pb.defParam(
+            "power",
+            units="W",
+            description="Total power",
+            categories=[parameters.Category.neutronics],
+        )
 
         pb.defParam("powerDecay", units="W", description="Total decay power")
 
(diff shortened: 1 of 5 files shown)


armi/physics/neutronics/parameters.py
```diff
@@ -492,7 +492,10 @@ def linPowByPinGamma(self, value):
         )
 
         pb.defParam(
-            "pdens", units="W/cm$^3$", description="Average volumetric power density"
+            "pdens",
+            units="W/cm$^3$",
+            description="Average volumetric power density",
+            categories=[parameters.Category.neutronics],
```
john-science, line 498: Sorry for being obtuse, but why are we adding these categories?

john-science (approved):
Approved.

@mgjarrett @jakehader @keckler  This PR worries me. It seems like this PR is only possible because we didn't do any downstream integration tests of the original Efficiency PR.  Is that correct?

If it IS correct, I would like to have a quick discussion about our PR process.

Thanks!

keckler:
To say we didn't do _any_ integration tests is most certainly not true. We did a lot of integration testing on #1042 , and caught a lot of problems in the process before merging that. 

I can't speak specifically to how these particular bugs made it through, though.

mgjarrett:
> To say we didn't do _any_ integration tests is most certainly not true. We did a lot of integration testing on #1042 , and caught a lot of problems in the process before merging that.
> 
> I can't speak specifically to how these particular bugs made it through, though.

The integration tests were run, but the test result differences had not yet been reviewed in detail. These issues were caught in that review process.

Issue #1095, opened 2023-01-17
Provide version-validating helper function


john-science:
In [this discussion](https://github.com/terrapower/armi/discussions/1066) @drewj-usnctech and I discussed the need for a simple helper function that could validate and match a version number.  For instance, it would need to be able to validate these cases:

* `1.2.3` meets `1.2.3`
* `1.2.3` does NOT meet `1.2.4`
* `1.2` meets `1.2.3`
* `1.3` does NOT meet `1.2.3`

The goal of this function is to provide a standard way for developer's plugins to check versions in settings files. Users could add a `Setting` for the version of their app, and then add a `SettingValidator` which would use this helper function to match the version.

drewj-usnctech:
Thanks @john-science! Is the intent still to have these "metadata" settings in the `metadata` block in the settings input file? Or would they reside in the main body of the input? To put it another way, are we looking for
```yaml
metadata:
  version: uncontrolled  # armi version
  myApp : 4.2  # primary app version
  myTransportKernel: 1.2.3
settings:
  neutronicsKernel: myTransportKernel
```
vs.
```yaml
metadata:
  version: uncontrolled  # armi version
settings:
  myAppVersion: 4.2
  myTransportKernelVersion: 1.2.3
  neutronicsKernel: myTransportKernel
```

My preference would be for the former: versions in the `metadata`. But I know much of the plugin interface is focused on adding content to the `settings` section

john-science:
I have a branch open for this helper method, Drew: https://github.com/terrapower/armi/commit/8f55cccea3f52b0616b3e6b09ce9f9f7ba737168

I am standing here looking at this dubious `metadata` section, and while I think the first option you have looks better:

```yaml
metadata:
  version: uncontrolled  # armi version
  myApp : 4.2  # primary app version
```

I'd like to support that for you, but it looks to be quite a large change. Where your second option is free:

```yaml
settings:
  myAppVersion: 4.2
```

Let me look at this a little longer.  I'd like to support the better option. But it appears we have no tooling build around the `metadata` section at all; it is just for this one `Setting`.  I don't like that that, but I'm thinking "perfect is the enemy of better.  

Let me keep poking around.

PR #1102, opened 2023-01-18
Adding a helper method to validate versions in settings files

john-science:
## Description

There was [a request](https://github.com/terrapower/armi/issues/1095) to validate the versions of the app or plugin from a custom setting.

This is a little tool the user can use as part of a setting validation, to validate their versions for controlled runs.  Please see [this armi-example-app PR](https://github.com/terrapower/armi-example-app/pull/3) for how to do that.  

After both of these PRs are merged, I could also add examples of custom settings to the ARMI docs, under tutorials.

### Docs

I have also added a basic introduction to how to add new `Settings` to you `Plugin` in the docs.

---

## Checklist

### DIFF
--- a/armi/operators/settingsValidation.py
+++ b/armi/operators/settingsValidation.py
@@ -746,9 +746,9 @@ def validateVersion(versionThis: str, versionRequired: str) -> bool:
     bool
         Does this version match the version in the Settings file/object?
     """
-    fullV = "[0-9]\.[0-9]\.[0-9]"
-    medV = "[0-9]\.[0-9]"
-    minV = "[0-9]"
+    fullV = "\d+\.\d+\.\d+"
+    medV = "\d+\.\d+"
+    minV = "\d+"
 
     if versionRequired == "uncontrolled":
         # This default flag means we don't want to check the version.
--- a/armi/operators/settingsValidation.py
+++ b/armi/operators/settingsValidation.py
@@ -737,7 +737,7 @@ def validateVersion(versionThis: str, versionRequired: str) -> bool:
     versionThis: str
         The version of this ARMI, App, or Plugin.
         This MUST be in the form: 1.2.3
-    versionRequired: str:
+    versionRequired: str
         The version to compare against, say in a Settings file.
         This must be in one of the forms: 1.2.3, 1.2, or 1
 
@@ -750,7 +750,10 @@ def validateVersion(versionThis: str, versionRequired: str) -> bool:
     medV = "[0-9]\.[0-9]"
     minV = "[0-9]"
 
-    if re.search(fullV, versionThis) is None:
+    if versionRequired == "uncontrolled":
+        # This default flag means we don't want to check the version.
+        return True
+    elif re.search(fullV, versionThis) is None:
         raise ValueError(
             "The input version ({0}) does not match the required format: {1}".format(
                 versionThis, fullV
--- a/armi/operators/settingsValidation.py
+++ b/armi/operators/settingsValidation.py
@@ -762,9 +762,9 @@ def validateVersion(versionThis: str, versionRequired: str) -> bool:
     elif re.search(fullV, versionRequired) is not None:
         return versionThis == versionRequired
     elif re.search(medV, versionRequired) is not None:
-        return versionThis[:-2] == versionRequired
+        return ".".join(versionThis.split(".")[:2]) == versionRequired
     elif re.search(minV, versionRequired) is not None:
-        return versionThis[0] == versionRequired
+        return versionThis.split(".")[0] == versionRequired
     else:
         raise ValueError(
             "The required version is not a valid format: {}".format(versionRequired)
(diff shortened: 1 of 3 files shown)


john-science:
@drewj-usnctech This is the PR for our little helper function.

Sorry, THIS we can do today. But it looks like it would take a medium-size overhaul of the settings system to write directly to the `metadata` header in the settings file.

We can talk about that, if it is high value to you.

armi/operators/settingsValidation.py
```diff
@@ -726,3 +727,45 @@ def createQueryRevertBadPathToDefault(inspector, settingName, initialLambda=None
...
+    elif re.search(fullV, versionRequired) is not None:
+        return versionThis == versionRequired
+    elif re.search(medV, versionRequired) is not None:
+        return versionThis[:-2] == versionRequired
+    elif re.search(minV, versionRequired) is not None:
+        return versionThis[0] == versionRequired
+    else:
+        raise ValueError(
+            "The required version is not a valid format: {}".format(versionRequired)
+        )
```
keckler, line 771: Is this ValueError duplicating the one above?

armi/operators/settingsValidation.py
```diff
@@ -726,3 +727,45 @@ def createQueryRevertBadPathToDefault(inspector, settingName, initialLambda=None
...
+        This must be in one of the forms: 1.2.3, 1.2, or 1
+
+    Returns
+    -------
+    bool
+        Does this version match the version in the Settings file/object?
+    """
+    fullV = "[0-9]\.[0-9]\.[0-9]"
+    medV = "[0-9]\.[0-9]"
+    minV = "[0-9]"
```
keckler, line 751: I think these patterns need to be extended to handle numbers with multiple digits?

armi/operators/settingsValidation.py
john-science: There are two `ValueError`s here. One for `versionThis` not being valid, and one for `versionRequired` not being valid.

armi/operators/settingsValidation.py
john-science: AH crap.

Yes.  You're right. Dammit. Thanks!

john-science: Fixed!

john-science:
@drewj-usnctech   I know it doesn't get us 100% the way to putting your version in `metadata`, but here is the PR.  Frankly, I think it would be easier to remove the `metadata` section from the `Settings` files entirely than to fully support them, so we could add more settings there.

Tell me what you think. Particularly, does my addition to the tutorial docs make enough sense?

opotowsky (approved):
looks good!

doc/tutorials/making_your_first_app.rst
```diff
@@ -164,6 +164,27 @@ add a little more to the plugin.
...
+runs.
+
+The following example boiler plate code defines three settings. We define two simple
+number settings (inlet and outlet temperatures), and we use :py:class:`Query 
+<armi.operators.settingsValidation .Query>` to define validation on those settings. Here,
+the validation isn't very exciting, we just make sure the temperatures are above zero.
+That's not particularly physically meaningful, but serves as a simple example. The next
+setting is a little more complicated, we define a setting ``myAppVersion`` that defines
+a specific version of our app that this setting file is valid for. And if you try to run
+a different version you get a nasty warning printed to the screen.
```
drewj-usnctech, line 181: Just so I'm clear, it's now on the plugin / app side to raise an error if they so choose if versions are unsupported?

- Plugin provides a setting for some version
- Plugin defines a setting validator

This works, just want to make sure I understand the paradigm

john-science: Right now all the `settingsValidation` does is give the option for the user to hand-change a setting interactively. They don't kill a run.  e.g.:

https://github.com/terrapower/armi/blob/e0b01d708eac7ba73c0caabde75158af97fa007e/armi/operators/settingsValidation.py#L51-L61

So, that is the default workflow that I want to show in the `armi-example-app`.

But, in YOUR case, I imagine you would do in the `condition` step of your `Query` is `raise ValueError()` before you `return False`.  That way your run would _actually_ stop if the version was wrong.  Remember, you can put arbitrary code into that first `"condition"` field in a `Query`.

Sorry, I was trying to tow the line between what a general user of the `armi-example-app` would like, and what I thought would be useful to you.

Does that make sense?  Does that seem like a pathway that gets you everything you need/want?

drewj-usnctech: Yeah that makes sense and should work. Thanks!

I think also if you use `correction=inspector.NO_ACTION` the `Query` validation will error for you

john-science: Oh, possibly you are right. I should try that.

How do you feel about this PR?

I know it's not 100% what you wanted, but will it be helpful?  Should we merge it?

drewj-usnctech: Oh yes, this is very helpful! The background, context, and implementation all makes a lot more sense.

I'm open to discussing how to improve this later, but I'm also more than content with where this is at

PR #1106, opened 2023-01-19
Using mockRunLogs in a better way

john-science:
## Description

It turns out, in our unit tests, people have been directly accessing `._outputStream` when there was already a method to access that value: `.getStdoutValue()`. Well, let's just use that method instead.

Also, I changed the name of that method to `.getStdout()` to make it easier to use and added a new method, because we tend to need it a lot: `.emptyStdout()`.

**NOTE**: This change was chosen because it does not break any downstream code.

---

## Checklist


opotowsky (approved):
This wins the award for most monotonous PR .

Hero's work, thanks!

PR #1108, opened 2023-01-20
Fix Issues with Peak Params

mgjarrett:
## Description

When mapping parameters between the non-uniform and uniform mesh, the mapper could handle volume-integrated or average values. There needs to also be a way to map "peak" or max parameters (`ParamLocation.MAX`) between the uniform and non-uniform mesh. Before they were just mapped the same as average values, which doesn't quite make sense. The "peak" values aren't associated with a location, so we still end up fudging a little bit when mapping between meshes.

Also, `cumulative` category was added to some params that are cumulative, and the `"cumulative over cycle"` category was formalized in `parameterDefinitions.Category`.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/parameters.py
+++ b/armi/physics/neutronics/parameters.py
@@ -677,6 +677,7 @@ def _getNeutronicsBlockParams():
             units="DPA/s",
             description="Peak DPA rate based on detailedDpaPeak",
             location=ParamLocation.MAX,
+            categories=["detailedAxialExpansion", "neutronics"],
         )
 
         pb.defParam(
--- a/armi/physics/neutronics/parameters.py
+++ b/armi/physics/neutronics/parameters.py
@@ -677,7 +677,10 @@ def _getNeutronicsBlockParams():
             units="dpa",
             location=ParamLocation.AVERAGE,
             description="Displacement per atom accumulated during this cycle. This accumulates over a cycle and resets to zero at BOC.",
-            categories=["cumulative over cycle", "detailedAxialExpansion"],
+            categories=[
+                parameters.Category.cumulativeOverCycle,
+                parameters.Category.detailedAxialExpansion,
+            ],
         )
 
         pb.defParam(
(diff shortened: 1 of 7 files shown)


armi/physics/neutronics/parameters.py
```diff
@@ -660,14 +684,16 @@ def linPowByPinGamma(self, value):
             "detailedDpaPeakRate",
             units="DPA/s",
             description="Peak DPA rate based on detailedDpaPeak",
-            location=ParamLocation.AVERAGE,
+            location=ParamLocation.MAX,
+            categories=["detailedAxialExpansion", "neutronics"],
```
opotowsky, line 688: do you want the string version here or `parameters.Category.detailedAxialExpansion`, etc?

armi/physics/neutronics/parameters.py
```diff
@@ -660,14 +684,16 @@ def linPowByPinGamma(self, value):
...
         )
 
         pb.defParam(
             "dpaPeakFromFluence",
             units="dpa",
             description="DPA approximation based on a fluence conversion factor set in the dpaPerFluence setting",
-            location=ParamLocation.AVERAGE,
+            location=ParamLocation.MAX,
+            categories=["cumulative", "detailedAxialExpansion"],
```
opotowsky, line 696: ditto above

armi/reactor/blockParameters.py
```diff
@@ -275,6 +275,7 @@ def buGroupNum(self, buGroupNum):  # pylint: disable=method-hidden
             units="dpa",
             description="displacements per atom with peaking factor",
             categories=["cumulative", "detailedAxialExpansion", "depletion"],
```
opotowsky, line 277: oh I guess you'd need to change it all over.... up to you

opotowsky (approved):
I don't see any issues, just optional comments

armi/physics/neutronics/parameters.py
mgjarrett: Yeah we should use `parameters.Category.detailedAxialExpansion` for anything new that's going in. I won't go through and change pre-existing string versions though.

armi/reactor/blockParameters.py
opotowsky: this isn't a new line!

PR #1113, opened 2023-01-23
Expose tight coupling to `OperatorSnapshots`

albeanth:
## Description
PR #1033 introduced the capability to measure the convergence of tight coupling and removed the `looseCoupling` setting. However, its exposure to the snapshot operator was missed. This PR serves to resolve that. 

---

## Checklist

### DIFF
--- a/armi/operators/tests/test_operators.py
+++ b/armi/operators/tests/test_operators.py
@@ -27,6 +27,8 @@ from armi.utils.directoryChangers import TemporaryDirectoryChanger
 from armi.physics.neutronics.globalFlux.globalFluxInterface import (
     GlobalFluxInterfaceUsingExecuters,
 )
+from armi.utils import directoryChangers
+from armi.bookkeeping.db.databaseInterface import DatabaseInterface
 
 
 class InterfaceA(Interface):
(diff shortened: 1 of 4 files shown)


opotowsky (approved):
I haven't been involved in this feature add, but this PR looks good to me!

john-science (approved):
It's a small change. And I see its value.

I'd like the change more if it broke a test. Or we added a test to cover it.

albeanth:
> I'd like the change more if it broke a test. Or we added a test to cover it.

I totally agree... Let me see if I can add/modify one.

albeanth:
Getting a decent test in is slightly tricky since `_performTightCoupling` calls on `interactAllCoupled` and tries to actually do stuff. So I'm working on getting an operator put together that avoids doing stuff but that can still write out a DB.

armi/operators/tests/test_operators.py
```diff
@@ -107,6 +109,25 @@ def test_couplingIsActive(self):
         self.o.cs["tightCoupling"] = True
         self.assertTrue(self.o.couplingIsActive())
 
+    def test_dbWriteForCoupling(self):
+        with directoryChangers.TemporaryDirectoryChanger():
+            self.o.cs["tightCoupling"] = True
+            self.dbWriteForCoupling(writeDB=True)
+            self.dbWriteForCoupling(writeDB=False)
```
john-science, line 116: There are no `assert` statements here.  We should add one.

albeanth: Ah. The assertion lives in the `dbWriteForCoupling` method.

PR #1114, opened 2023-01-23
`TD_frac` updates and consistency

### LINKED ISSUE #1112: `TD_frac` for `UraniumOxide` and `MOX` doesn't do anything
The `TD_frac` material modification doesn't do anything for `MOX` or `UraniumOxide`. To confirm this, try running the following:

```python
"""
Tests for Uranium Oxide.
"""

import unittest

from armi.materials.uraniumOxide import UO2
from armi.materials.tests.test_materials import _Material_Test


class UraniumOxide_TestCase(_Material_Test, unittest.TestCase):
    MAT_CLASS = UO2

    def setUp(self):
        _Material_Test.setUp(self)

    def test_applyInputParams(self):
        self.UO2_TD = UO2()
        original = self.UO2_TD.density(500)
        self.UO2_TD.applyInputParams(TD_frac=0.1)
        new = self.UO2_TD.density(500)
        ratio = new / original
        self.assertEqual(0.1, self.UO2_TD.getTD())
        self.assertEqual(ratio, 0.1)


if __name__ == "__main__":
    unittest.main()
```

You'll see that the test fails. This is a big problem!

This could be a problem for other material classes as well. The only other one I've looked at is B4C, which appears to be fine.

keckler:
## Description

This closes #1112  by ensuring that the `TD_frac` material modification on `UraniumOxide` and `MOX` both actually change the density as they are supposed to.

This is implemented mainly via changes to the `SimpleSolid` class and by making setting a default value of 1.0 to the `theoreticalDensityFrac` material parameter.

Other than that, the other material classes that were originally using some form of a `theoreticalDensityFrac` variable were modified to be consistent in their usage of that variable. Those other material classes included `ThoriumOxide` ~~and `B4C`~~.

This doesn't touch anything in the API, so users should not notice any changes except the fact that `TD_frac` is now correctly applied, so their results might change if they were using the framework oxide classes.

EDIT: This PR no longer makes any changes to B4C. A [new ticket](https://github.com/terrapower/armi/issues/1116) was opened to address that class.

---

## Checklist


keckler:
I am testing our internal projects against this change now, and will update when I know that those work.

keckler:
This appears to be working. I had to make one change, to the internal version of B4C. But that was easy to do, and now this should be ready to land.

keckler:
I see something is failing... grrr

keckler:
Okay @john-science , this is ready for review.

keckler:
And now I need to change this to account for https://github.com/terrapower/armi/pull/1062

armi/materials/materialParameters.py
```diff
@@ -0,0 +1,115 @@
+# Copyright 2019 TerraPower, LLC
```
john-science, line 1: This file can be deleted.

armi/materials/tests/test_uraniumOxide.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class UraniumOxide_TestCase(_Material_Test, unittest.TestCase):
+    MAT_CLASS = UO2
+
+    def setUp(self):
+        _Material_Test.setUp(self)
+
+    def test_applyInputParams(self):
+        # TODO add other matMods
```
john-science, line 34: We try not to add new `TODO` in a PR.  We should add more tests or remove the `TODO`, in an ideal world.

armi/materials/materialParameters.py
keckler: Done!

armi/materials/tests/test_uraniumOxide.py
keckler: Done!

keckler:
I have no idea why these tests are failing. They all pass locally, and looking at the ones that claim to be failing, they seem pretty random...

keckler:
@john-science Have you seen this before? The tests passed with Python3.9 but not 3.10...

armi/materials/material.py
```diff
@@ -669,6 +669,13 @@ def heatCapacity(self, Tk=None, Tc=None):
             f"Material {type(self).__name__} does not implement heatCapacity"
         )
 
+    def getTD(self):
+        return self.theoreticalDensityFrac
+
+    def adjustTD(self, val):
+        self.theoreticalDensityFrac = val
+        self.clearCache()
```
ntouran, line 677: these should have docstrings explaining the concept of TD

ntouran (changes requested):
would be good to update the user docs to explain this in the materials docs.

PR #1117, opened 2023-01-24
Update B4C docstring for assumed TD frac

### LINKED ISSUE #1116: B4C defaults to a 0.9 TD fraction
This isn't really a _Problem_, but it is unexpected, inconsistent, and potentially troublesome. 

The B4C class has a built-in non-unity theoretical density fraction that is applied by default when the material is constructed:
https://github.com/terrapower/armi/blob/37f547940501d5e55978cac2fdba37e10fbae0ff/armi/materials/b4c.py#L117-L145

This non-unity fraction is used whenever the material density is queried:
https://github.com/terrapower/armi/blob/37f547940501d5e55978cac2fdba37e10fbae0ff/armi/materials/b4c.py#L160-L178

There is _also_ the ability to specify the theoretical density fraction using the `TD_frac` material modification, which is a good thing:
https://github.com/terrapower/armi/blob/37f547940501d5e55978cac2fdba37e10fbae0ff/armi/materials/b4c.py#L31-L52

The bad part about that is that all other material classes return a fully-dense material by default. If you want a version of the material with lower density, generally you should use the associated material modification `TD_frac`. But for B4C, you actually need to use `TD_frac` to get a full-density material. So this means that the B4C class is inconsistent with our other materials, but it is not documented as such.

I did want to change this as part of #1114 , but since it would be an API-breaking change and it proved a bit more challenging than expected, I decided to just report it here for future work instead. B4C, being as ubiquitous as it is, is used allllllll over our repos, and this could be tricky to implement a change.

keckler:
## Description

This closes #1116 by updating the docstring. No other changes were made.

---

## Checklist


john-science:
Sorry, @keckler .  All your `TD_frac` PRs in ARMI (and elsewhere) need to have some conflicts resolved due to another Materials PR going into ARMI (and other downstream repos) this morning.

In most cases, it will be quite easy.  We removed all `Parameter`s from ARMI `Material`s. But we just moved all the ones that were still used to class attributes, so you can probably fix 99% of your errors by going from `mat.p.thing` to `mat.thing`.

Sorry for the hassle!

john-science (approved):

keckler:
Done!

armi/materials/b4c.py
```diff
@@ -12,8 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Boron carbide; a very typical reactor control material."""
-from armi import runLog
```
john-science, line 16: Whoops!

You deleted this import, which is used.

That's why the tests are failing.

keckler: Oops, this happened when I merged in main. Weird. Fixed!

PR #1120, opened 2023-01-25
Calculate fission rate density.

mgjarrett:
## Description

Calculate `b.p.fisDens` (fissions/cc/sec) in `calcReactionRates`

`fisDens` needs to be calculated here, after de-converison of uniform reactor back to non-uniform assemblies.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -1172,9 +1172,8 @@ def calcReactionRates(obj, keff, lib):
 
     if rate["rateFis"] > 0.0:
         vFuel = obj.getComponentAreaFrac(Flags.FUEL)
-        vol = obj.getVolume()
-        obj.p.fisDens = rate["rateFis"] / (vol * vFuel)
-        obj.p.fisDensHom = rate["rateFis"] / vol
+        obj.p.fisDens = rate["rateFis"] / vFuel
+        obj.p.fisDensHom = rate["rateFis"]
     else:
         obj.p.fisDens = 0.0
         obj.p.fisDensHom = 0.0
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -1170,10 +1170,6 @@ def calcReactionRates(obj, keff, lib):
     for paramName, val in rate.items():
         obj.p[paramName] = val  # put in #/cm^3/s
 
-    if rate["rateFis"] > 0.0:
-        vFuel = obj.getComponentAreaFrac(Flags.FUEL)
-        obj.p.fisDens = rate["rateFis"] / vFuel
-        obj.p.fisDensHom = rate["rateFis"]
-    else:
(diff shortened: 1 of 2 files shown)


jakehader (approved):
I might ask that we rename this function to be something else and propagate that change since this is doing a bit more than just reaction rates now. I think there is a bigger change that needs to come as we have discussed where this function might eventually calculate all flux-dependent parameters so we can leave as-is for now and then update later. 

Make sure to run `black`!

mgjarrett:
> I might ask that we rename this function to be something else and propagate that change since this is doing a bit more than just reaction rates now. I think there is a bigger change that needs to come as we have discussed where this function might eventually calculate all flux-dependent parameters so we can leave as-is for now and then update later.

Fission density is a reaction rate per unit volume, so it still seems appropriate. But once we expand it to calculate powers then that will be a new scope. `calculateFluxDerivedQuantities` or something like that?

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -1169,3 +1169,11 @@ def calcReactionRates(obj, keff, lib):
 
     for paramName, val in rate.items():
         obj.p[paramName] = val  # put in #/cm^3/s
+
+    if rate["rateFis"] > 0.0:
+        vFuel = obj.getComponentAreaFrac(Flags.FUEL)
+        obj.p.fisDens = rate["rateFis"] / vFuel
+        obj.p.fisDensHom = rate["rateFis"]
+    else:
```
john-science, line 1177: Sorry, naïve question.

Is this because `"rateFis"` might be zero?  Or are you trying to catch some crazy "negative rate" error?

mgjarrett: The goal was to not call `obj.getComponentAreaFrac(Flags.FUEL)` on a block unless it has a positive fission rate. If we call that on every block, then we'll get a bunch of warnings from this: https://github.com/terrapower/armi/blob/76a76b1b404a7beb38282c16405c25b9d54fe57a/armi/reactor/blocks.py#L1092-L1098

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -1169,3 +1169,11 @@ def calcReactionRates(obj, keff, lib):
...
+        obj.p.fisDens = 0.0
+        obj.p.fisDensHom = 0.0
```
mgjarrett, line 1179: An alternative implementation would be:
```suggestion
    vFuel = obj.getComponentAreaFrac(Flags.FUEL) if rate["rateFis"] > 0.0 else 1.0
    obj.p.fisDens = rate["rateFis"] / vFuel
    obj.p.fisDensHom = rate["rateFis"]
```

`vFuel` shouldn't matter for the `else` because `rateFis` should be 0. Just want to avoid divide by 0.

mgjarrett:
It passed CI before we ran into the `pip` failures today. I tested the new commit locally and the unit tests still passed. It's a pretty small commit, so I think it's safe to just squash and merge anyway. I don't know how to re-trigger the tests without making a commit.

PR #1122, opened 2023-01-25
Catch invalid material modification keys in the blueprints

### LINKED ISSUE #402: No warning when material modification is unused

keckler:
## Description

This closes #402  (finally).

This approach will catch invalid material modification names during construction of assemblies from the blueprints. It does this by inspecting the valid material modification options based on the materials in each block and comparing those against the material modification keys that a user specifies in the blueprints.

This seems like a pretty lightweight, one-and-done option that is completed at first parsing of the blueprints. But let me know if you see any major issues with this type of solution! I know we've seen a couple gotcha's in the past.

Tests are added for both normal and "by component" material modifications.

---

## Checklist


armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -16,14 +16,16 @@
 This module defines the ARMI input for a block definition, and code for constructing an ARMI ``Block``.
 """
 import collections
-
 import yamlize
 
+from inspect import signature
```
john-science, line 21: Can we move this up above line 18?

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -115,11 +117,26 @@ def construct(
...
             components[c.name] = c
+
+            # check that the mat mods for this component are valid options
+            # this will only examine by-component mods, block mods are done later
+            if isinstance(c, Component):
+                # there are other things like composite groups that don't get
+                # material modifications -- skip those
+                validMatModOptions = signature(
+                    c.material.applyInputParams
+                ).parameters.keys()
```
john-science, line 133: I haven't seen this inspector before.  It certainly seems to provide you with a way to solve the problem.  Interesting.

doc/release/0.2.rst
```diff
@@ -29,6 +29,7 @@ What's new in ARMI
 #. Refine logic for Block.getNumPins() to only count components that are actually pins. (`PR#1098 <https://github.com/terrapower/armi/pull/1098>`_)
 #. Improve handling of peak/max parameters by the UniformMeshConverter parameter mapper. (`PR#1108 <https://github.com/terrapower/armi/pull/1108>`_)
 #. Bug fix to expose new tight coupling functionality to ``OperatorSnapshots``. (`PR#1113 https://github.com/terrapower/armi/pull/1113`_)
+#. Check blueprints for invalid material modification keys
```
john-science, line 32: Can you add the link to this PR?

> (`PR#1122 https://github.com/terrapower/armi/pull/1122`_)

john-science:
This is a much cleaner solution than I expected.

It really seems like `from inspect import signature` is the star of the show here.  Nice find.

I'm trying to decide if this solution will always work in the situation where we have other materials libraries ingested in some big ARMI Application.  I think it will, because by the time you get to reading the blueprints, all the materials needs-must have been designed. Wherever they came from.

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -115,11 +117,26 @@ def construct(
...
-            filteredMaterialInput = self._filterMaterialInput(
+            filteredMaterialInput, byComponentMatModKeys = self._filterMaterialInput(
                 materialInput, componentDesign
             )
             c = componentDesign.construct(blueprint, filteredMaterialInput)
             components[c.name] = c
+
+            # check that the mat mods for this component are valid options
+            # this will only examine by-component mods, block mods are done later
+            if isinstance(c, Component):
```
john-science, line 128: I'm trying to decide if `isinstance(c, Component)` is correct, if it is more correct (and future-proof) to use `hasattr(c, "material")`.

Why did you choose to only worry about Components that might have materials, and not, say, `Composite`s?

Just trying to find possible edge cases.

armi/reactor/blueprints/tests/test_materialModifications.py
```diff
@@ -189,14 +189,56 @@ def test_filterMaterialInput(self):
             "fuel2": {"ZR_wt_frac": 0.3, "U235_wt_frac": 0.3},
         }
         componentDesign = a[0][0]
-        filteredMaterialInput = BlockBlueprint._filterMaterialInput(
+        filteredMaterialInput, _ = BlockBlueprint._filterMaterialInput(
```
john-science, line 192: Can we might as well test whatever `byComponentMatModKeys` get returned here?

It seems like a low-hanging fruit additional test, since your function now returns more info.

Sorry to bother!

armi/reactor/blueprints/blockBlueprint.py
keckler: Something that I stumbled on in the course of this work was the new `ComponentGroup`, which effectively sits at the same position in the reactor hierarchy as a `Component`:
https://github.com/terrapower/armi/blob/8b6ea1abcfe6ea4587ab0a2622f4d7c03e214e83/armi/reactor/blueprints/componentBlueprint.py#L160-L175

So when looping over all the "thing" in a block, you can get both `Component`s and `ComponentGroup`s (and maybe other things).

But it appeared to me that the only things that actually have `applyInputParam()` applied to them are `Component`s:
https://github.com/terrapower/armi/blob/4699c867e6b6fb7c1308a879e03c241f1d946b75/armi/reactor/blueprints/componentBlueprint.py#L238

So not knowing too much about these `ComponentGroup`s except that they don't receive material modifications, I just decided to test that the object was a `Component`.

But I guess your suggested approach would work as well.

After that explanation, do you want me to go in that direction? Happy to do so, just let me know.

john-science: Ah!

I see the issue.  Sure, `ComponentGroup` is new and not fully implemented everywhere.  That's fine then. No worries.

john-science (approved):

keckler:
Unfortunately, this plan has a flaw.

It is common practice for us to make derive a material class from `FuelMaterial`, which has it's own `applyInputParams()` that typically needs to be applied on top of the `applyInputParams()` in the derived class itself.

This means that, in order to find all the valid material modifications in situations like this, we'd have to traverse up the MRO to find the valid material modifications in the parents as well.

That starts to get kinda ugly, and I'm wondering if we want to go down that path or not.

Any thoughts @john-science ?

keckler:
This PR just helped me catch some bugs in my model! Woohoo!

Issue #1097, opened 2023-01-18
Issues found with 2D Density


john-science:
## Background

In the past, ARMI has had two methods:

* [Material.density()](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/materials/material.py#L357) - a 2D "pseudo" density (necessary for linear expansion)
* [Material.density3()](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/materials/material.py#L403) -  the usual physical density

This unfortunate naming convention lead to confusion.

## Potential Errors

While working on improving the naming convention, I found the following potential bugs:

* [x] `Material.densityTimesHeatCapacity()` [calls pseudo density](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/materials/material.py#L616), but I think it should call physical density.
* [x] `Component.density()` [calls pseudo density](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/reactor/components/component.py#L1162-L1170), but I think it should call physical density.
* [X] ~`Fluid(Material)` [calls pseudo density](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/materials/material.py#L651-L664)~ We define density and pseudodensity as the same for Fluids.
* Several materials only have pseudo density and not physical density, which suggests an error:
    * [x] [air](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/materials/air.py#L62) - Fluid (document it)
    * [x] [concrete.py](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/materials/concrete.py#L42-L43) - Broken ref, I can't tell if this is wrong.
    * [x] [copper.py](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/materials/copper.py#L32) - No reference, so I can't tell.
    * [x] [cs.py](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/materials/cs.py#L30) - Fluid (document it)
[...]

PR #1128, opened 2023-01-26
Fixing temperature range of Potassium

john-science:
## Description

@dpham-materials pointed out [here](https://github.com/terrapower/armi/issues/1097#issuecomment-1398849029) that the Potassium material had the `(min, max)` temperature ranges:

* `(63.38, 759) "K"`

But that reference is valid for:

* `(63.2, 1250), "C")`

Which is:

* `(336.35, 1523), "K")`

So I made that update, as part of [this umbrella ticket](https://github.com/terrapower/armi/issues/1097).

I expect the impact of this change to be small, because reactor cores are rarely at temperatures below 336K (which they would have to be for this change to break backwards compatibility).

---

## Checklist

### DIFF
--- a/armi/materials/potassium.py
+++ b/armi/materials/potassium.py
@@ -26,7 +26,7 @@ class Potassium(material.Fluid):
     """
 
     name = "Potassium"
-    propertyValidTemperature = {"density": ((63.38, 1250), "C")}
+    propertyValidTemperature = {"density": ((63.2, 1250), "C")}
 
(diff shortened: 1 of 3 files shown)


dpham-materials:
Minor change requested: Different sources will cite the melting point of potassium to be slightly different. Since we're referencing the NaK handbook, let's use the value it recommends on Page 10:
63.2°C = 336.35K

opotowsky (approved):

PR #1130, opened 2023-01-26
Implement timestep compare argument for database comparison

opotowsky:
## Description

There was some phantom `timestepMatchup` argument floating around that got my hopes up when I needed to compare two databases, one of which had additional time steps. 

I didn't fully implement the original intention of this argument since no one was using it in the first place and I'm not sure it's a useful feature. What I needed was a list of time steps to include in a comparison of two DBs. So I implemented that instead. Selfish? Maybe. Sorry? Not yet.

---

## Checklist

### DIFF
--- a/armi/cli/compareCases.py
+++ b/armi/cli/compareCases.py
@@ -68,14 +68,14 @@ class CompareCases(EntryPoint):
             help=("Patterns for parameters to ignore in comparisons"),
         )
         parser.add_argument(
-            "--timestepMatchup",
+            "--timestepCompare",
             default=None,
             action="store",
             nargs="+",
             help=(
-                "How to line up timesteps for comparisons. Note that any timestep not listed"
-                "will not be compared. Format the key and value separated by a period. "
-                "E.g. 0.0 1.2 2.1 3.3 will swap comparisons on the 1st and 2nd timenodes"
+                "List of timesteps to compare. Note that any timestep not listed will "
+                "not be compared. Format the cycle and node separated by a period. E.g. "
+                "0.0 0.1 1.2 3.3 will compare c0n0, c0n1, c1n2, c3n3 and skip all others"
             ),
         )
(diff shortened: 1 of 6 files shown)


armi/cli/compareCases.py
```diff
@@ -68,14 +68,14 @@ def _addComparisonOptions(self):
...
         parser.add_argument(
-            "--timestepMatchup",
+            "--timestepCompare",
             default=None,
             action="store",
             nargs="+",
             help=(
-                "How to line up timesteps for comparisons. Note that any timestep not listed"
-                "will not be compared. Format the key and value separated by a period. "
-                "E.g. 0.0 1.2 2.1 3.3 will swap comparisons on the 1st and 2nd timenodes"
```
opotowsky, line 78: I didn't implement the old version of this because I don't know any way to keep the input arg simple for this, and it seemed complicated to design for. If we wanted to actually match up different timesteps for comparison, it would have to look something like `1.0:0.2` to compare `c1n0` ref and `c0n2` src.

john-science: > Better is the enemy of perfect.

If you can get a feature together that works, that's better than having something broken and defunct lying around.

john-science:
Wait, how did the code coverage not go down?  I mean, I'm happy it didn't, but shouldn't it have?

opotowsky:
> Wait, how did the code coverage not go down? I mean, I'm happy it didn't, but shouldn't it have?

I changed something already there and already tested. The `if not timestepCompare` doesn't result in any new uncovered lines since it's the default. I didn't look at the compareDB tests yet...wanted a pre-review first!

john-science (approved):

opotowsky:
I see you approved this @john-science, but I added the arg to an existing test (not changing its behavior) just so it's in a unit test somewhere.

PR #1136, opened 2023-01-31
Calculate block masses at BOC

mgjarrett:
## Description

Block heavy metal and fissile masses are calculated in the depletion interface, which is the only interface where they would typically be affected. However, depletion is typically inactive at node 0 of a cycle. This PR adds a simple calculation for `b.p.kgHM` and `b.p.kgFis` on core loading and after shuffling.

## Checklist

### DIFF
--- a/armi/bookkeeping/db/tests/test_comparedb3.py
+++ b/armi/bookkeeping/db/tests/test_comparedb3.py
@@ -174,7 +174,6 @@ class TestCompareDB3(unittest.TestCase):
 
         # end-to-end validation that comparing a photocopy database works
         diffs = compareDatabases(dbs[0]._fullPath, dbs[1]._fullPath)
-        print(diffs.diffs)
         self.assertEqual(len(diffs.diffs), 462)
         # Cycle length is only diff (x3)
         self.assertEqual(diffs.nDiffs(), 3)
--- a/armi/bookkeeping/db/tests/test_comparedb3.py
+++ b/armi/bookkeeping/db/tests/test_comparedb3.py
@@ -174,7 +174,8 @@ class TestCompareDB3(unittest.TestCase):
 
         # end-to-end validation that comparing a photocopy database works
         diffs = compareDatabases(dbs[0]._fullPath, dbs[1]._fullPath)
-        self.assertEqual(len(diffs.diffs), 456)
+        print(diffs.diffs)
+        self.assertEqual(len(diffs.diffs), 462)
         # Cycle length is only diff (x3)
         self.assertEqual(diffs.nDiffs(), 3)
 
--- a/armi/bookkeeping/report/tests/test_report.py
+++ b/armi/bookkeeping/report/tests/test_report.py
@@ -177,7 +177,7 @@ class TestReportInterface(unittest.TestCase):
 
         self.assertEqual(repInt.fuelCycleSummary["bocFissile"], 0.0)
         repInt.interactBOC(1)
-        self.assertEqual(repInt.fuelCycleSummary["bocFissile"], 0.0)
+        self.assertEqual(repInt.fuelCycleSummary["bocFissile"], 726304.0175506991)
 
     def test_interactEOC(self):
         o, r = loadTestReactor()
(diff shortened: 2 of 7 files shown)


armi/bookkeeping/db/tests/test_comparedb3.py
```diff
@@ -174,7 +174,8 @@ def test_compareDatabaseSim(self):
 
         # end-to-end validation that comparing a photocopy database works
         diffs = compareDatabases(dbs[0]._fullPath, dbs[1]._fullPath)
```
opotowsky, line 176: heads up I just merged something into main affecting this line...but I don't think it should cause a merge conflict

opotowsky: nevermind I gave you a merge conflict. sorry!

mgjarrett: thanks for the heads up!

armi/reactor/reactors.py
```diff
@@ -374,6 +374,12 @@ def summarizeReactorStats(self):
             )
         )
 
+    def setBlockMassParams(self):
+        """Set the parameters kgHM and kgFis for each block"""
+        for b in self.getBlocks():
+            b.p.kgHM = b.getHMMass()
+            b.p.kgFis = b.getFissileMass()
```
opotowsky, line 381:
```suggestion
            b.p.kgHM = b.getHMMass() / units.G_PER_KG
            b.p.kgFis = b.getFissileMass() / units.G_PER_KG
```

armi/bookkeeping/report/tests/test_report.py
```diff
@@ -177,7 +177,7 @@ def test_interactBOC(self):
 
         self.assertEqual(repInt.fuelCycleSummary["bocFissile"], 0.0)
         repInt.interactBOC(1)
-        self.assertEqual(repInt.fuelCycleSummary["bocFissile"], 0.0)
+        self.assertEqual(repInt.fuelCycleSummary["bocFissile"], 726304.0175506991)
```
mgjarrett, line 180:
```suggestion
        self.assertEqual(repInt.fuelCycleSummary["bocFissile"], 726.3040175506991)
```

armi/bookkeeping/report/tests/test_report.py
```diff
@@ -177,7 +177,7 @@ def test_interactBOC(self):
 
         self.assertEqual(repInt.fuelCycleSummary["bocFissile"], 0.0)
         repInt.interactBOC(1)
-        self.assertEqual(repInt.fuelCycleSummary["bocFissile"], 0.0)
+        self.assertEqual(repInt.fuelCycleSummary["bocFissile"], 726.3040175506991)
```
mgjarrett, line 180:
```suggestion
        self.assertAlmostEqual(repInt.fuelCycleSummary["bocFissile"], 726.30401755)
```

armi/reactor/reactors.py
```diff
@@ -2280,6 +2286,8 @@ def processLoading(self, cs, dbLoad: bool = False):
 
         self.stationaryBlockFlagsList = stationaryBlockFlags
 
+        self.setBlockMassParams()
```
john-science, line 2289: So, are we duplicating effort now?

Sorry, I am just looking to see if we are slowing down the code by adding duplicate copies of this same logic. Presumably, these values were already being filled at other time steps. So, do we still need to calculate them at those timesteps in the exact same way?

opotowsky: This isn't actually duplicating effort, but filling in these calculations at time steps where they were previously missing. 

But, since you asked, the code is duplicated at those other time steps (at least internally).

mgjarrett: Until now, our neutronics code for SFRs was calculating and populating `b.p.kgHM` and `b.p.kgFis`. For anyone using a different plugin for neutronics, they would either need to implement a different mass calculation within one of their plugins, or those parameters would just be left as the default value of 0.0.

We have the functions to calculate these parameters in the framework, and it makes more sense to calculate them directly rather than outsource a nearly trivial mass calculation to an external code.

> Presumably, these values were already being filled at other time steps. So, do we still need to calculate them at those timesteps in the exact same way?

At time node 0, when fuel is first loaded into the core, we calculate the heavy metal and fissile mass in each block.
At non-zero time nodes there has been some depletion, so these masses need to be updated. The updated mass calculation should be handled by the depletion interface.

john-science (approved):

PR #1141, opened 2023-02-01
Have snapshots disable interfaces based on function as well

onufer:
## Description

---

## Checklist

- [] The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any bug fixes or new features.

### DIFF
--- a/armi/operators/snapshots.py
+++ b/armi/operators/snapshots.py
@@ -31,10 +31,7 @@ class OperatorSnapshots(operatorMPI.OperatorMPI):
         # disable fuel management and optimization
         # disable depletion because we don't want to change number densities for tn's >0 (or any)
         for toDisable in ["fuelHandler", "optimize", "depletion"]:
-            i = self.getInterface(name=toDisable)
-            if i:
-                i.enabled(False)
-            i = self.getInterface(function=toDisable)
+            i = self.getInterface(name=toDisable, function=toDisable)
             if i:
                 i.enabled(False)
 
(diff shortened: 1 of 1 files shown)


armi/operators/snapshots.py
```diff
@@ -31,7 +31,10 @@ def createInterfaces(self):
         # disable fuel management and optimization
         # disable depletion because we don't want to change number densities for tn's >0 (or any)
         for toDisable in ["fuelHandler", "optimize", "depletion"]:
-            i = self.getInterface(toDisable)
+            i = self.getInterface(name=toDisable)
+            if i:
+                i.enabled(False)
+            i = self.getInterface(function=toDisable)
             if i:
                 i.enabled(False)
```
opotowsky, line 39: Looks like this should be deleted

opotowsky: Oops now I get it no it shouldn't

opotowsky (approved):
Snapshots operator has been hard to test, so this code was already and still is uncovered (so cov went down when adding lines to an uncovered method).

armi/operators/snapshots.py
```diff
@@ -31,7 +31,10 @@ def createInterfaces(self):
         # disable fuel management and optimization
         # disable depletion because we don't want to change number densities for tn's >0 (or any)
         for toDisable in ["fuelHandler", "optimize", "depletion"]:
-            i = self.getInterface(toDisable)
+            i = self.getInterface(name=toDisable)
+            if i:
+                i.enabled(False)
+            i = self.getInterface(function=toDisable)
```
albeanth, line 37: Could this same thing be accomplished if you added `function=toDisable` onto the original call to `getInterface`? So it would be:
```python
i = self.getInterface(name=toDisable, function=toDisable)
```

Though I may have missed something as I do not know the context of the PR.

onufer: nope, thats a good idea. thanks!

albeanth:
Just curious, what is the context here? I get that you want the interfaces to be disabled on the interface `function` as well, but what was the context that spurred this?

Issue #1135, opened 2023-01-31
The `single` log messages should store a hash


john-science:
Right now, you can ask `runLog.py` to never repeat a particular log message by using `single=True`:

https://github.com/terrapower/armi/blob/8b6ea1abcfe6ea4587ab0a2622f4d7c03e214e83/armi/runLog.py#L445-L453

That's great. But testing for GIANT, long log messages can be slow, particularly when people put `single=True` all throughout their code. That's why we had to add the extra `label` feature.

We should just hash the message and skip the label. Pick a _fast_ hash algorithm, and don't worry about cryptographic security.

drewj-usnctech:
Would [builtin `hash`](https://docs.python.org/3.7/library/functions.html#hash) be sufficient? Just `hash(msg)`?

john-science:
> Would [builtin `hash`](https://docs.python.org/3.7/library/functions.html#hash) be sufficient? Just `hash(msg)`?

That would definitely work.

(I was just wondering if there was anything faster.  For instance, I know the SIMS use something like the [FNV hash](https://github.com/Cilyan/stblc/blob/master/STBLWriter.cs#L58) for string comparison. And [murmer3](https://stackoverflow.com/a/5400389/1287593) is fast I'm told, but also... I don't want another dependency. It's not important.)

PR #1145, opened 2023-02-02
Hashing log messages for quicker deduplication

john-science:
## Description

In references to [Issue #1135](https://github.com/terrapower/armi/issues/1135), I am just hashing the string message of a log statement, when we are trying to de-duplicate it. This requires less memory and is faster than storing and comparing the full strings.  

I decided it was time to do this because people are using this feature more and more often lately.

I could only speed-up the situation so much because we are currently saving all `Warning`/`Error` level messages and printing a collection of them at the end. So we need to save he entire log message. From my perspective, this is a nice-to-have feature that we will lose if people starting using this "don't duplicate this log message" feature a lot more in the future.

---

## Checklist

### DIFF
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -450,8 +450,9 @@ class DeduplicationFilter(logging.Filter):
         # If the message is set to "do not duplicate" we may filter it out
         if single:
             if record.levelno in (logging.WARNING, logging.CRITICAL):
-                # grab the custom label, if it exists, for warning logging later
+                # if this is from the custom logger, it will have a "label"
                 label = getattr(record, "label", msg)
+                # the "label" default is None, which needs to be replaced
                 label = msg if label is None else label
 
                 if label not in self.singleWarningMessageCounts:
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -383,23 +383,23 @@ def raw(msg):
     """
     Print raw text without any special functionality.
     """
-    LOG.log("header", msg, single=False, label=msg)
+    LOG.log("header", msg, single=False)
 
 
 def extra(msg, single=False, label=None):
-    LOG.log("extra", msg, single=single, label=label)
+    LOG.log("extra", msg, single=single)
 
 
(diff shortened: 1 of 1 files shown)


armi/runLog.py
```diff
@@ -446,18 +446,22 @@ def filter(self, record):
...
         single = getattr(record, "single", False)
-        label = getattr(record, "label", msg)
-        label = msg if label is None else label
 
         # If the message is set to "do not duplicate" we may filter it out
         if single:
             if record.levelno in (logging.WARNING, logging.CRITICAL):
+                # grab the custom label, if it exists, for warning logging later
+                label = getattr(record, "label", msg)
+                label = msg if label is None else label
```
keckler, line 455: Don't these two lines duplicate each other? Because you default to `msg` from `getattr()`

john-science: Well, the default `label` is `None`.  So I think `getattr(record, "label", msg)` returns `None` most of the time.

I could try it and find out though.

opotowsky: If the default label is `None`, wouldn't `getattr(record, "label", msg)` return `msg` instead of `None`?

john-science: @keckler @opotowsky 

The issue is this:

```python
>>> class X:
...     def __init__(self, x):
...         self.label = None
...         self.x = x
...
>>> x = X(7)
>>>
>>> getattr(x, "x", "hi")
7
>>> getattr(x, "label", "msg")
None
```

When `x.label` exists, but is set to the default value `None`, the `getattr()` correctly finds it and returns `None` as the value.

john-science: I have attempted to clarify the situation with better comments:

```python
# if this is from the custom logger, it will have a "label"
label = getattr(record, "label", msg)
# the "label" default is None, which needs to be replaced
label = msg if label is None else label
```

keckler (approved):

opotowsky (approved):

PR #1148, opened 2023-02-03
Opening up the disable snapshot interfaces list

john-science:
## Description

There was a hard-coded list in `snapshots.py` that recently became really important. People might need to change this list.  By preference, I believe this is not so common a need that it should be done as a `Setting`; it would be better as a class attribute. Even without this request, this list is better as a class attribute, because we don't like hard-coding data.

**NOTE**: Based on a discussion yesterday with @keckler, I am finally removing our rule against using the `super()` method. I have not found this rule to be productive or helpful.  It's high past time.

---

## Checklist

### DIFF
--- a/armi/operators/snapshots.py
+++ b/armi/operators/snapshots.py
@@ -26,11 +26,17 @@ class OperatorSnapshots(operatorMPI.OperatorMPI):
     This operator can be run as a restart, adding new physics to a previous run.
     """
 
-    def createInterfaces(self):
-        operatorMPI.OperatorMPI.createInterfaces(self)
+    def __init__(self, cs):
+        super().__init__(cs)
+
         # disable fuel management and optimization
         # disable depletion because we don't want to change number densities for tn's >0 (or any)
-        for toDisable in ["fuelHandler", "optimize", "depletion"]:
+        self.disabledInterfaces = ["depletion", "fuelHandler", "optimize"]
+
+    def createInterfaces(self):
+        operatorMPI.OperatorMPI.createInterfaces(self)
+
+        for toDisable in self.disabledInterfaces:
             i = self.getInterface(name=toDisable, function=toDisable)
             if i:
                 i.enabled(False)
--- a/armi/operators/tests/test_operatorSnapshots.py
+++ b/armi/operators/tests/test_operatorSnapshots.py
@@ -61,6 +61,34 @@ class TestOperatorSnapshots(unittest.TestCase):
         self.o._mainOperate()
         self.assertEqual(self.r.core.p.power, 100000000.0)
 
+    def test_createInterfaces(self):
+        self.assertEqual(len(self.o.interfaces), 0)
+        self.o.createInterfaces()
+
+        # If someone adds an interface, we don't want this test to break, so let's do >6
+        self.assertGreater(len(self.o.interfaces), 6)
+
+    def test_createInterfacesDisabled(self):
+        self.assertEqual(len(self.o.interfaces), 0)
+        allInterfaces = [
+            "main",
+            "fissionProducts",
+            "xsGroups",
+            "fuelHandler",
+            "history",
+            "database",
+            "memoryProfiler",
+            "snapshot",
+        ]
+        for i in allInterfaces:
+            self.o.disabledInterfaces.append(i)
+        self.o.createInterfaces()
+
+        # If someone adds an interface, we don't want this test to break, so let's do >6
+        self.assertGreater(len(self.o.interfaces), 6)
+        for i in self.o.interfaces:
+            self.assertFalse(i.enabled())
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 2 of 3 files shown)


doc/developer/standards_and_practices.rst
```diff
@@ -299,12 +299,8 @@ zero pylint warnings or errors.
 General do's and don'ts
 =======================
 
-do not use ``print``
+Do not use ``print``.
     ARMI code should not use the ``print`` function; use one of the methods within ``armi.runLog``.
 
-do not use ``super``
```
onufer, line 305: this is a pretty big difference in guidance. i thought we had good reason for it. it gets complex with multiple inheritance, which we also discourage. can you check with Nick? There are 2 ways to do it, and our code base is mostly limited to one way. it might be best to keep it that way

armi/operators/snapshots.py
```diff
@@ -26,11 +26,17 @@ class OperatorSnapshots(operatorMPI.OperatorMPI):
...
 
-    def createInterfaces(self):
-        operatorMPI.OperatorMPI.createInterfaces(self)
+    def __init__(self, cs):
+        super().__init__(cs)
+
         # disable fuel management and optimization
         # disable depletion because we don't want to change number densities for tn's >0 (or any)
-        for toDisable in ["fuelHandler", "optimize", "depletion"]:
+        self.disabledInterfaces = ["depletion", "fuelHandler", "optimize"]
```
onufer, line 34: you might consider a class variable. this is a singleton anyway....

armi/operators/tests/test_operatorSnapshots.py
```diff
@@ -61,6 +61,34 @@ def test_mainOperate(self):
         self.o._mainOperate()
         self.assertEqual(self.r.core.p.power, 100000000.0)
 
+    def test_createInterfaces(self):
```
albeanth, line 64: Love the test additions to get coverage for the disabled interfaces. 

Though it's not entirely clear to me how these assertions are covering the disabled interfaces? I very well may be missing something. 

Could we simplify and make into one test if we did something like below after a call to `self.o.createInterfaces()`? 
```python
disabledInterfaces = self.o.disabledInterfaces
for interface in self.o.interfaces:
    if i.name in disabledInterfaces or i.function in disabledInterfaces:
        self.assertFalse(i.enabled())
    else:
        self.assertTrue(i.enabled()) 
```

doc/developer/standards_and_practices.rst
albeanth: I too was a bit surprised by this. I'm not as opposed to multiple inheritance (as long as it's done carefully), but am certainly curious as to what the motivation was behind the change. I would have expected a larger discussion amongst the ARMI devs for a change to the standards and practices.

john-science: I can remove this from this PR and  make it it's own, if you like.

But I have never heard of a Python project with this rule. `super()` is just part of the Python language, and it makes long-term maintenance of OOP project significantly less tedious.

But, mostly, I am aware of no reason this rule would help us. And yesterday Chris K got bit by this problem. And frankly I think `super()` makes this PR cleaner.

armi/operators/snapshots.py
john-science: I considered it.  And that WOULD work, if you have strong opinions.

But, generally, we don't put things in class variables if we expect (or in this case _want_) people to edit them. 

It's certainly not a hard rule, so I'm open to the change. But people tend to use class variables for immutable data.

armi/operators/tests/test_operatorSnapshots.py
john-science: If you scroll down a few more lines, you will see the test for disabled interfaces.

doc/developer/standards_and_practices.rst
albeanth: I'm not necessarily opposed to the change, I think it's reasonable; `super()` is standard python practice for inheritance, as I understand it. I was just hoping for a bit more dialogue on it - or any change to the standards and practices, really. In its current form, I think it would slide under the radar and most of the regular ARMI contributors would miss it.

I would advocate for pulling it out of the PR and opening an issue or discussion on it so that there is transparency amongst everyone. Just my 2 cents though!

john-science: Well, the only reason I included it in this PR is because this PR makes a lot more sense with `super()`.  I can pull it out into it's own PR, and just wait on this one until the `super()` PR merges.

But here's the short version:  I do not see any benefit to the current opposition to `super()` use in our docs.  It seems like a "style points" rule made by someone with "strong opinions", and not a practical decision designed to help us out in any way.  After carefully researching it, I can't find any other Python open source projects using this rule. Because... why would they? What is the benefit of such a rule? 

Sure, we can open another ticket/discussion/PR.

john-science: Okay, I moved this to another PR.

albeanth (approved):

PR #1149, opened 2023-02-03
Pointing the Component.density method at 3D material density

john-science:
## Description

Presumably by accident, in the past the `Component.density()` method pointed at the 2D/linear expansion-only/pseudodensity, instead of the usual physical density of materials.  This PR corrects that issue.

@opotowsky 
**NOTE**: In testing, however, I found no materials in ARMI that explicitly defined separate `pseduodensity()` and `density()` method. So, using existing materials I could make a unit test that would fail after the change. But the moment I mocked up a material with clearly different `density` and `pseudodensity`, the test immediately failed.

---

## Checklist

### DIFF
--- a/armi/materials/uraniumOxide.py
+++ b/armi/materials/uraniumOxide.py
@@ -55,7 +55,7 @@ class UraniumOxide(material.FuelMaterial, material.SimpleSolid):
     propertyUnits = {"heat capacity": "J/mol-K"}
 
     propertyValidTemperature = {
-        "density": ((300, 3100), "K"),
+        "density3": ((300, 3100), "K"),
         "heat capacity": ((298.15, 3120), "K"),
         "linear expansion": ((273, 3120), "K"),
         "linear expansion percent": ((273, __meltingPoint), "K"),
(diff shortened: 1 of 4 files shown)


armi/materials/uraniumOxide.py
```diff
@@ -55,7 +55,7 @@ class UraniumOxide(material.FuelMaterial, material.SimpleSolid):
     propertyUnits = {"heat capacity": "J/mol-K"}
 
     propertyValidTemperature = {
-        "density": ((300, 3100), "K"),
+        "density3": ((300, 3100), "K"),
```
opotowsky, line 58: Is this the only material you found that needed this update?

opotowsky (approved):
Wow, fascinating that this didn't break for any of the materials. Seems like...luck?

Anyway, this turned out to be a lot simpler than I thought!

opotowsky:
This is fine to merge @john-science

armi/materials/uraniumOxide.py
john-science: Well, I randomly happened across this one. 

Perhaps we could look through all the materials for similar mis-namings.

PR #1151, opened 2023-02-03
Parse through applyInputParams in the MRO

keckler:
## Description

This is a fix to the issue pointed out in https://github.com/terrapower/armi/pull/1122#issuecomment-1416242677

This change now makes the error checking a little bit more lenient. For example, it will allow material modifications from a parent class to go through, even if those modifications are not applied in the child. I don't see any way around this. But we will still be protected from straight-up typos like `TD_frac` mistakenly being typed as `Td_frac`, which was really the point of this issue in the first place.

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/blockBlueprint.py
+++ b/armi/reactor/blueprints/blockBlueprint.py
@@ -21,6 +21,7 @@ from inspect import signature
 import yamlize
 
 from armi import getPluginManagerOrFail, runLog
+from armi.materials.material import Material
 from armi.reactor import blocks
 from armi.reactor import parameters
 from armi.reactor.flags import Flags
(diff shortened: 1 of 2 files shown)


john-science:
i feel like if we add 20 new lines of code, we should add a test.

Sure, your code coverage didn't go down. Granted. But you added 20 new lines of logic, so we must be adding some feature worth testing, right?

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -128,9 +129,14 @@ def construct(
...
                 # material modifications -- skip those
-                validMatModOptions = signature(
-                    c.material.applyInputParams
-                ).parameters.keys()
+                validMatModOptions = set()
+                for materialParentClass in c.material.__class__.__mro__:
+                    # we must loop over parents as well, since applyInputParams
+                    # could call to Parent.applyInputParams()
+                    if issubclass(materialParentClass, Material):
+                        validMatModOptions |= signature(
```
john-science, line 137: This might be a little easier to read as `validMatModOptions.update()` than `validMatModOptions |=`. 

It's an opinion, of course.

keckler:
I agree that this should be tested, and I'm not trying to drag my heels on this! I just got pulled into something urgent and haven't had the time. Will get to it in the next day or so.

keckler:
@john-science I just put up the relevant tests.

john-science (approved):

PR #1152, opened 2023-02-06
Add Copying Existing Flux Solutions to Cross Section Control Settings

jakehader:
## Description

Adding new functionality to the cross-section settings to copy existing flux solution files during lattice physics evaluations. This is useful for lattice physics codes that allow for an external solver solution file to be provided.

This breaks the crossSectionControl setting for existing applications with the use of pregenerated cross sections. The name of the setting was changed from `fileLocation` to `xsFileLocation` to allow the addition of the `fluxFileLocation` setting.

---

## Checklist

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -65,3 +65,5 @@ include armi/tests/tutorials/anl-afci-177.yaml
 include armi/tests/tutorials/data_model.ipynb
 include armi/tests/zpprTest.yaml
 include armi/tests/zpprTestGeom.xml
+include armi/physics/neutronics/tests/ISOXA
+include armi/physics/neutronics/tests/rzmflxYA
--- a/armi/bookkeeping/db/compareDB3.py
+++ b/armi/bookkeeping/db/compareDB3.py
@@ -164,6 +164,7 @@ def compareDatabases(
     srcFileName: str,
     exclusions: Optional[Sequence[str]] = None,
     tolerance: float = 0.0,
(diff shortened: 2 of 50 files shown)


mgjarrett:
I think I like this solution a little more than mine, but we should definitely add a migration from `fileLocation` to `xsFileLocation`, otherwise this is going to be a pesky break in backwards compatibility.

mgjarrett (approved):

jakehader:
Opened up https://github.com/terrapower/armi/issues/1166 for further clean-ups. Note to @mgjarrett - The `crossSectionControl` setting with `fluxFileLocation` option implies that the name of the file should contain `rzmflx` (case sensitive) for this to function on our end. The issue that is linked will aim to clean up these assumptions more broadly in the code base. There is no settings validation for `rzmflx` in the name because I don't think it's appropriate to mandate this since it should in the future be more general.

opotowsky (approved):
Approving the added test!

mgjarrett:
The unit tests worked when I ran locally, but they don't seem to be working here. I think the necessary `ISOXA` and `rzmflxYA` files are not being copied into the test directory.

opotowsky:
> The unit tests worked when I ran locally, but they don't seem to be working here. I think the necessary `ISOXA` and `rzmflxYA` files are not being copied into the test directory.

That reminds me -- they also need to be added to the `MANIFEST.in` file. I don't think that solves this problem, though

mgjarrett:
Looks great, nice job!

PR #1156, opened 2023-02-06
Update corner and edge DPA in DoseResultsMapper

mgjarrett:
## Description

Corner and edge DPA should be incremented in the dose results mapper. These were previously incremented in an internal physics plugin on the TerraPower side. Now a plugin just needs to set the DPA rate, and the incrementing is handled by the `DoseResultsMapper`, which is how other DPA parameters are handled.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -840,9 +840,13 @@ class DoseResultsMapper(GlobalFluxResultMapper):
             # this is specific to hex geometry, but they are general neutronics block parameters
             # if it is a non-hex block, this should be a no-op
             if not None in (b.p.pointsCornerDpa, b.p.pointsCornerDpaRate):
-                b.p.pointsCornerDpa = b.p.pointsCornerDpa + b.p.pointsCornerDpaRate * stepTimeInSeconds
+                b.p.pointsCornerDpa = (
+                    b.p.pointsCornerDpa + b.p.pointsCornerDpaRate * stepTimeInSeconds
+                )
             if not None in (b.p.pointsDpa, b.p.pointsEdgeDpaRate):
-                b.p.pointsEdgeDpa = b.p.pointsEdgeDpa + b.p.pointsEdgeDpaRate * stepTimeInSeconds
+                b.p.pointsEdgeDpa = (
+                    b.p.pointsEdgeDpa + b.p.pointsEdgeDpaRate * stepTimeInSeconds
+                )
 
             if self.options.dpaPerFluence:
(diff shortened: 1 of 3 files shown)


armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -836,6 +836,18 @@ def updateFluenceAndDpa(self, stepTimeInSeconds, blockList=None):
             b.p.detailedDpaPeak = b.p.detailedDpaPeak + newDPAPeak
             b.p.detailedDpaThisCycle = b.p.detailedDpaThisCycle + newDpaThisStep
 
+            # increment point dpas
+            # this is specific to hex geometry, but they are general neutronics block parameters
+            # if it is a non-hex block, this should be a no-op
+            if b.p.pointsCornerDpa is not None and b.p.pointsCornerDpaRate is not None:
```
Nebbychadnezzar, line 842: `pointsCornerDpa` and `pointsEdgeDpa` is `None` out of the gate. If we're moving this code from the internal plugin, how are these variables getting initialized to empty numpy arrays (since that plugin was doing that work)?

Nebbychadnezzar:
When we do coupled iterations, are we running the `apply` method the GlobalFluxInterface over and over? If so, isn't that going to cause code with `+=`  operations to accumulate over each iteration?

mgjarrett:
> When we do coupled iterations, are we running the `apply()` method the GlobalFluxInterface over and over? If so, isn't that going to cause variables with `+=` operations to accumulate over each iteration?

`DoseResultsMapper.apply()` is not called anywhere in the framework. We call it within our depletion plugin, which is only called after the coupling has converged. That's the motivation to move this calculation out of our internal plugin and here to the framework, where other DPAs are incremented.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
mgjarrett: Yes, they need to be initialized

Nebbychadnezzar:
> > When we do coupled iterations, are we running the `apply()` method the GlobalFluxInterface over and over? If so, isn't that going to cause variables with `+=` operations to accumulate over each iteration?
> 
> `DoseResultsMapper.apply()` is not called anywhere in the framework. We call it within our depletion plugin, which is only called after the coupling has converged. That's the motivation to move this calculation out of our internal plugin and here to the framework, where other DPAs are incremented.

Ah, okay! I didn't know this, I thought we were calling it after each flux solve.

albeanth (approved):
@Nebbychadnezzar you're up! 🙂

Nebbychadnezzar (approved):
👍🏼

PR #1161, opened 2023-02-06
Clarify definitions of `crInsertedElevation` and `crWithdrawnElevation`

keckler:
## Description

The definition of the `crInsertedElevation` parameter mistakenly referred to the length of the section with control material, when in reality it should have referred to the length of the moveable section.

This PR fixes this, and generally clarifies the definitions for these two important parameters.

---

## Checklist

### DIFF
--- a/armi/reactor/assemblyParameters.py
+++ b/armi/reactor/assemblyParameters.py
@@ -212,8 +212,9 @@ def getAssemblyParameterDefinitions():
             "crInsertedElevation",
             units="cm",
             description=(
-                "The elevation of the bottom of the bottom-most moveable block in a control rod assembly when fully inserted. Note that this should "
-                "be considered a lower elevation than the ``crWithdrawnElevation`` by definition and modeling semantics."
+                "The elevation of the furthest-most insertion point of a control rod assembly. For a control rod assembly "
+                "inserted from the top, this will be the lower tip of the bottom-most moveable section in the assembly when "
+                "fully inserted."
             ),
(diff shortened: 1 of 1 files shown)


armi/reactor/assemblyParameters.py
```diff
@@ -230,7 +230,7 @@ def _enforceNotesRestrictions(self, value):
             "crWithdrawnElevation",
             units="cm",
             description=(
-                "The initial starting elevation of the moveable section of a control rod assembly when fully withdrawn.  Note that this should "
+                "The elevation of the bottom of the bottom-most moveable section of a control rod assembly when fully withdrawn. Note that this should "
```
jakehader, line 233: The reason this said starting is because the direction is based on interpretation of the parameter within the context of it's definition (think rods that insert from the bottom of the core). In most cases this is the interpretation though but others may implement onto this more generically. I'm okay with this but wanted to provide some background

jakehader (approved):

armi/reactor/assemblyParameters.py
keckler: Oh, yes, you're totally correct. I will change!

keckler: Okay, what about now?https://github.com/terrapower/armi/pull/1161/commits/0e42c0801059a84ff2270b2a38038bc9c71329bb

jakehader: Nice! I'll merge

PR #1163, opened 2023-02-06
Changing the name Material.density3 to density

john-science:
## Description

In the past, ARMI has had two methods:

* [Material.density()](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/materials/material.py#L357) - a 2D "pseudo" density (necessary for linear expansion)
* [Material.density3()](https://github.com/terrapower/armi/blob/b5235f0872dafc1428c6209e20ec1365b1c35c92/armi/materials/material.py#L403) -  the usual physical density

This unfortunate naming convention lead to confusion.  (See [this issue](https://github.com/terrapower/armi/issues/1097).) This PR makes the API-breaking name changes:

* `density()` --> `pseudoDensity()`
* `density3()` --> `density()`

---

## Checklist


john-science:
@jakehader @drewj-usnctech @sombrereau @opotowsky 

Just FYI.

This PR is "draft" to give everyone good notice. Because this is an API-breaking change, I will wait a week before merging this PR into `main`.  To avoid current deadlines and work, let's pencil in Valentine's Day for this PR.

drewj-usnctech:
Thanks @john-science. Happy to see this change come through

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -109,7 +109,7 @@ def _storeTargetComponentMassAndDensity(self, c):
         tmp = array(
             [
                 c.getMass(),
-                c.material.getProperty("density", c.temperatureInK),
+                c.material.getProperty("pseudoDensity", c.temperatureInK),
```
john-science, line 112: @albeanth This looks a bit odd to me.  Is this the way you want it?

albeanth: It should be `density3` I think. But the density portion of that array looks like it's unused in the tests... So it could probably get removed.

albeanth: Er, in the case of this PR, not `density3`, but rather just `density`.

jakehader:
> @jakehader @drewj-usnctech @sombrereau @opotowsky
> 
> Just FYI.
> 
> This PR is "draft" to give everyone good notice. Because this is an API-breaking change, I will wait a week before merging this PR into `main`. To avoid current deadlines and work, let's pencil in Valentine's Day for this PR.

Thank you! Please send an email internally on impacted stakeholders so they are aware of this large change.

armi/reactor/converters/tests/test_axialExpansionChanger.py
john-science: Thanks, @albeanth !  Fixed!

armi/materials/tests/test_materials.py
```diff
@@ -601,15 +622,15 @@ def test_heatCapacity(self):
...
+            self.mat.pseudoDensity(Tc=expectedTemperature), 30.0
         )
         self.assertAlmostEqual(expectedTemperature, tAtTargetDensity)
 
     def test_getDensityExpansion3D(self):
         expectedTemperature = 100.0
 
         ref_density = 10.86792660463439e3
-        test_density = self.mat.density3KgM3(Tc=expectedTemperature)
+        test_density = self.mat.pseudoDensityKgM3(Tc=expectedTemperature)
```
albeanth, line 633: Looks like this should be `densityKgM3`, yes?

armi/reactor/tests/test_components.py
```diff
@@ -309,10 +309,12 @@ def test_densityConsistent(self):
```
albeanth, line 309: On line 306 there is a conditional that if `c` is a `Component` or `DerivedShape` that this test is skipped. Aren't all instances of `c` a `Component`? If so, none of the tests in `test_densityConsistent` are ever actually hit... Unless I'm misinterpreting?

armi/reactor/tests/test_components.py
```diff
@@ -336,7 +338,7 @@ class StrangeMaterial(Material):
             def density(self, Tk=None, Tc=None):
```
albeanth, line 338: Should this be `pseudodensity`?

doc/release/0.2.rst
```diff
@@ -8,6 +8,7 @@ Release Date: TBD
 
 What's new in ARMI
 ------------------
+#. The method ``Material.density3`` is now called ``density`` (`PR#1999 <https://github.com/terrapower/armi/pull/1999>`_)
```
albeanth, line 11: We should probably add a comment on `density` -> `pseudodensity`, yes?

albeanth (changes requested):
A few comments and suggestions to consider. Thanks for taking this on @john-science !

armi/materials/tests/test_materials.py
john-science: done

armi/reactor/tests/test_components.py
john-science: Yeah, this PR isn't supposed to do anything but change variable names.

But I tested it out and that test wasn't being run.  And when I tried to full reinstate it, it's clear it is used for LOTS of different types of `Component` subclasses that it couldn't possibly pass for (for instance, some components don't have a known volume).

So I deleted the broken half of this test, and revived the rest of it the best I could.

Done

armi/reactor/tests/test_components.py
john-science: Done!   Good catch.

doc/release/0.2.rst
john-science: Yeah, I improved the language some.  (Sorry, I'm always trying to keep these notes short for people.)

opotowsky (approved):

keckler:
This specific usage of `density()` did not change during this PR:
https://github.com/terrapower/armi/blob/5e808270f31504895559897c7112259c56fc035d/armi/reactor/composites.py#L1012

Here's the current git blame:
https://github.com/terrapower/armi/blame/5e808270f31504895559897c7112259c56fc035d/armi/reactor/composites.py#L1012

So if the call didn't change but the meaning of `density()` has changed... Did we introduce or eliminate a bug here?

I do not see this identified on the [list of bugs](https://github.com/terrapower/armi/issues/1097#issue-1537200189), so this seems inadvertent. 

Any insight here?

john-science:
@keckler Yeah, that was an attempt to fix a previous bug.  "rho" relates entirely to normal density, and not the 2D pseudo-density was specific solely for linear expansion.  At least, that was the idea.  I'm happy to be proven wrong.

keckler:
That's fine, I agree with the way that it is now. I was just getting confused, as I always do with this 2D density stuff...

Thanks for confirming!

armi/reactor/tests/test_components.py
```diff
@@ -304,40 +304,37 @@ def test_volumeAfterClearCache(self):
...
-        if c.is3D:
-            return  # no area defined
-        unexpandedHeight = c.parent.getHeight() / c.getThermalExpansionFactor()
-        self.assertAlmostEqual(
-            c.getArea(cold=True) * unexpandedHeight * c.material.refDens,
-            self.component.getMass(),
-        )
-        self.assertAlmostEqual(
-            c.getArea() * c.parent.getHeight() * c.density(), self.component.getMass()
-        )
```
onufer, line 329: @john-science, it seems like we lost some coverage here?

PR #1168, opened 2023-02-09
Fix Pu frac calculation

mgjarrett:
## Description

The previous puFrac calculation had a numerator that was adjusted for density changes (from block expansion) but a constant denominator that did not account for expansion.

The new formula calculates number of Pu atoms directly and does not account for density in either the numerator or denominator.

---

## Checklist


opotowsky (approved):
Checkout that test!

Thanks, and this looks good to me! Confirmed to work in practice, as well.

PR #1171, opened 2023-02-09
hexHoledCircle

Ashlita6:
## Description

This creates a new shape of a circle with a hexagonal hole in it. 

---

## Checklist

### DIFF
--- a/armi/reactor/components/complexShapes.py
+++ b/armi/reactor/components/complexShapes.py
@@ -127,7 +127,7 @@ class HexHoledCircle(basicShapes.Circle):
         )
 
     def getComponentArea(self, cold=False):
-        r"""Computes the area for the hexagon with n number of circular holes in cm^2."""
+        r"""Computes the area for the circle with one hexagonal hole."""
         od = self.getDimension("od", cold=cold)
         holeOP = self.getDimension("holeOP", cold=cold)
         mult = self.getDimension("mult")
--- a/armi/reactor/components/complexShapes.py
+++ b/armi/reactor/components/complexShapes.py
@@ -91,6 +91,62 @@ class HoledHexagon(basicShapes.Hexagon):
             return 0.0
 
 
+class HexHoledCircle(basicShapes.Circle):
+    """Circle with n uniform hexagon holes hollowed out of it."""
+
+    THERMAL_EXPANSION_DIMS = {"od", "holeOP"}
+
+    pDefs = componentParameters.getHexHoledCircleParameterDefinitions()
+
+    def __init__(
+        self,
+        name,
+        material,
+        Tinput,
+        Thot,
+        od,
+        holeOP,
+        mult=1.0,
+        modArea=None,
+        isotopics=None,
+        mergeWith=None,
+        components=None,
+    ):
+        ShapedComponent.__init__(
+            self,
+            name,
+            material,
+            Tinput,
+            Thot,
+            isotopics=isotopics,
+            mergeWith=mergeWith,
+            components=components,
+        )
+        self._linkAndStoreDimensions(
+            components, od=od, holeOP=holeOP, mult=mult, modArea=modArea
+        )
+
+    def getComponentArea(self, cold=False):
+        r"""Computes the area for the hexagon with n number of circular holes in cm^2."""
+        od = self.getDimension("od", cold=cold)
+        holeOP = self.getDimension("holeOP", cold=cold)
+        mult = self.getDimension("mult")
+        hexArea = math.sqrt(3.0) / 2.0 * (holeOP ** 2)
+        circularArea = math.pi * ((od / 2.0) ** 2)
+        area = mult * (circularArea - hexArea)
+        return area
+
+    def getHexInnerPitch(self, Tc=None, cold=False):
+        """
+        For the special case of only one single hole, returns the
+        diameter of that hole.
+
+        For any other case, returns 0.0 because an "circle inner diameter" becomes
+        undefined.
+        """
+        return 0.0
+
+
 class HoledRectangle(basicShapes.Rectangle):
     """Rectangle with one circular hole in it."""
 
--- a/armi/reactor/components/complexShapes.py
+++ b/armi/reactor/components/complexShapes.py
@@ -92,7 +92,7 @@ class HoledHexagon(basicShapes.Hexagon):
 
 
 class HexHoledCircle(basicShapes.Circle):
-    """Circle with a uniform hexagonal hole hollowed out of it."""
+    """Circle with a single uniform hexagonal hole hollowed out of it."""
 
     THERMAL_EXPANSION_DIMS = {"od", "holeOP"}
 
--- a/armi/reactor/components/complexShapes.py
+++ b/armi/reactor/components/complexShapes.py
@@ -92,7 +92,7 @@ class HoledHexagon(basicShapes.Hexagon):
 
 
 class HexHoledCircle(basicShapes.Circle):
-    """Circle with n uniform hexagon holes hollowed out of it."""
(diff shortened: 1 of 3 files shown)


keckler:
@john-science I'm taking a look at this now.

But as a side note, I'm wondering if it would make sense to implement some functionality in the future for creating "holed" components from other component shapes generically. We're starting to proliferate a bunch of "holed" things. So far, we have `HolexHexagon`, `HoledRectangle`, `HoledSquare`, and now `HexHoledCircle`. These basically all do the same thing under the hood.

If we could make a new shape like `HoledShape` that accepts `baseShape`, `holeShape`, and `nHoles`, we could get shapes of all varieties with holes in them with much less specific code.

armi/reactor/components/complexShapes.py
```diff
@@ -91,6 +91,62 @@ def getCircleInnerDiameter(self, Tc=None, cold=False):
...
+
+    def getHexInnerPitch(self, Tc=None, cold=False):
+        """
+        For the special case of only one single hole, returns the
+        diameter of that hole.
+
+        For any other case, returns 0.0 because an "circle inner diameter" becomes
+        undefined.
+        """
+        return 0.0
```
keckler, line 147: I think that you are intending to implement the `getCircleInnerDiameter()` method here? We have a requirement (for thermal expansion purposes) to implement a method on all component shapes called `getCircleInnerDiameter`, which basically returns the diameter of a circle that would be inscribed by the inner dimension.

For this case, I think you should (1) return 0.0 if there are multiple holes, or (2) return the diameter of the circle inscribed in the hexagonal hole like [this](https://media.geeksforgeeks.org/wp-content/uploads/Untitled-143-300x199.png).

If I'm interpreting incorrectly, and you actually _do_ want a method called `getHexInnerPitch()`, then the method above doesn't appear to do what you describe it to do. And, we should still implement a `getCircleInnerDiameter()` method as well.

armi/reactor/components/componentParameters.py
```diff
@@ -165,6 +165,15 @@ def getHoledHexagonParameterDefinitions():
     return pDefs
 
 
+def getHexHoledCircleParameterDefinitions():
+
+    pDefs = parameters.ParameterDefinitionCollection()
+    with pDefs.createBuilder(location=ParamLocation.AVERAGE, saveToDB=True) as pb:
+        pb.defParam("holeOP", units="cm", description="Pitch of interior hole(s)")
```
keckler, line 172: We should also add parameter definitions for `nHoles` and `od`.

armi/reactor/tests/test_components.py
```diff
@@ -1075,6 +1075,62 @@ def test_dimensionThermallyExpands(self):
             self.assertEqual(cur, ref[i])
 
 
+class TestHexHoledCircle(TestShapedComponent):
+    componentCls = HexHoledCircle
```
keckler, line 1079: You're going to have to add an import to this file for `HexHoledCircle`.

keckler (changes requested):
Your tests are failing right now, primarily because you need to add an import to your tests.

Let me know if you have any questions, and thanks for adding!

armi/reactor/components/complexShapes.py
```diff
@@ -91,6 +91,62 @@ def getCircleInnerDiameter(self, Tc=None, cold=False):
...
+
+    def __init__(
+        self,
+        name,
+        material,
+        Tinput,
+        Thot,
+        od,
+        holeOP,
+        mult=1.0,
```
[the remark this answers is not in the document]
mgjarrett: I suggested that we remove `nHoles`. I think we just need to update the docstring. A circle with multiple hexagonal holes would be a pretty weird shape. I can't really think of a use for it.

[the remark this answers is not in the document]
Ashlita6: I updated the docstring to reflect just having one hole. There was one other place that said hole(s) and so I also took out that (s).

armi/reactor/components/complexShapes.py
Ashlita6: I think the way I changed it meets the intent of (2), but let me know if I still didn't get it quite right! Thank you!

armi/reactor/components/complexShapes.py
```diff
@@ -91,6 +91,58 @@ def getCircleInnerDiameter(self, Tc=None, cold=False):
...
+            isotopics=isotopics,
+            mergeWith=mergeWith,
+            components=components,
+        )
+        self._linkAndStoreDimensions(
+            components, od=od, holeOP=holeOP, mult=mult, modArea=modArea
+        )
+
+    def getComponentArea(self, cold=False):
+        r"""Computes the area for the hexagon with n number of circular holes in cm^2."""
```
keckler, line 130: This docstring needs to be updated as well.

armi/reactor/components/componentParameters.py
```diff
@@ -165,6 +165,18 @@ def getHoledHexagonParameterDefinitions():
...
 
 
+def getHexHoledCircleParameterDefinitions():
+    """Return parameters for HexHoledCircle."""
+
+    pDefs = parameters.ParameterDefinitionCollection()
+    with pDefs.createBuilder(location=ParamLocation.AVERAGE, saveToDB=True) as pb:
+        pb.defParam("holeOP", units="cm", description="Pitch of interior hole")
+
+        pb.defParam("od", units="cm", description="Outer diameter")
```
keckler, line 175: Okay, apparently this line needs to be removed for your tests to pass: https://github.com/terrapower/armi/actions/runs/4145259916/jobs/7170527291#step:7:445

Sorry about that, cause it was me who recommended it in the first place :\

armi/reactor/tests/test_components.py
```diff
@@ -1075,6 +1076,62 @@ def test_dimensionThermallyExpands(self):
...
+        self.assertEqual(
+            self.componentDims["holeOP"],
+            simpleHexHoledCircle.getCircleInnerDiameter(cold=True),
+        )
+
+    def test_getArea(self):
+        od = self.component.getDimension("od")
+        opHole = self.component.getDimension("holeOP")
+        mult = self.component.getDimension("mult")
+        hexarea = math.sqrt(3.0) / 2.0 * (holeOP ** 2)
```
keckler, line 1112: This `holeOP` value is never assigned, so this test will error.

armi/reactor/tests/test_components.py
```diff
@@ -1075,6 +1076,62 @@ def test_dimensionThermallyExpands(self):
...
+            self.componentDims["holeOP"],
+        )
+        self.assertEqual(
+            self.componentDims["holeOP"],
+            simpleHexHoledCircle.getCircleInnerDiameter(cold=True),
+        )
+
+    def test_getArea(self):
+        od = self.component.getDimension("od")
+        opHole = self.component.getDimension("holeOP")
```
mgjarrett, line 1110:
```suggestion
        holeOP= self.component.getDimension("holeOP")
```

Ashlita6: Ah yes, that's exactly what it is! Thank you, Mike!

armi/reactor/components/complexShapes.py
```diff
@@ -91,6 +91,58 @@ def getCircleInnerDiameter(self, Tc=None, cold=False):
             return 0.0
 
 
+class HexHoledCircle(basicShapes.Circle):
+    """Circle with a uniform hexagonal hole hollowed out of it."""
```
onufer, line 95: optional: ad the word "single" before uniform or hexagonal. its an important distinction between this and holled hex since holed hex can have multiple holes with nHoles

keckler (approved):

Issue #1162, opened 2023-02-06
Tight Coupling: add runLog stating physics have converged


albeanth:
Feature request: It would be useful to add a runLog statement clearly indicating whether or not the physics have converged when considering tight coupling.

Issue #1089, opened 2023-01-13
Expose convergence norm to the user


keckler:
I'd also like to make the choice of norm exposed to the user. But that will be a future PR.

_Originally posted by @albeanth in https://github.com/terrapower/armi/pull/1033#discussion_r1064865646_

john-science:
@albeanth I believe this is already on your radar, right?  

Perhaps we should open a ticket or something that has your entire plan for the future (next PR) of the new tight coupling convergence?

albeanth:
@john-science yes, that is the plan. I just haven't gotten there yet. Soon!

PR #1172, opened 2023-02-09
Edit tight coupling `runLog`s + refactor tight coupling unit tests

### LINKED ISSUE #1115: Tight Coupling: Convergence Summary should not specify norm type
https://github.com/terrapower/armi/blob/b86547baa99ffc7a171b352c14d692277139f40e/armi/bookkeeping/report/reportingUtils.py#L241

The norm type for tight coupling convergence is not always an infinity norm and the summary-table should not state so.

albeanth:
## Description
This PR addresses the following:
1. Closes #1115 and #1162 
2. Corrects a minor typo from https://github.com/terrapower/armi/pull/1173/files#r1103086438. 
3. Refactors the growing number of tight coupling unit tests. 
    - Moves them to their own testing class that does not rely on `loadTestReactor`; these tests don't need a full reactor so there's no reason to build one. 

---

## Checklist

    - Probably not needed here. No new features or user-impacting bug fixes. 


armi/bookkeeping/report/reportingUtils.py
```diff
@@ -238,7 +238,7 @@ def getInterfaceStackSummary(o):
 
 
 def writeTightCouplingConvergenceSummary(convergenceSummary):
-    runLog.info("Tight Coupling Convergence Summary: Norm Type = Inf")
+    runLog.info("Tight Coupling Convergence Summary")
```
onufer, line 241: at some point we shoudl add what is converged to the table, maybe a (c) value for when its converged

onufer: doesn't have to be this PR

armi/operators/operator.py
```diff
@@ -393,7 +394,15 @@ def _performTightCoupling(self, cycle: int, timeNode: int, writeDB: bool = True)
                 self.r.core.p.coupledIteration = coupledIteration + 1
                 converged = self.interactAllCoupled(coupledIteration)
                 if converged:
+                    runLog.extra(
```
onufer, line 397: this should be at least info and my suggestion is "important".

onufer (changes requested):

armi/bookkeeping/report/reportingUtils.py
albeanth: Will address this with issue #1089.

armi/operators/operator.py
albeanth: Addressed. https://github.com/terrapower/armi/pull/1172/commits/7b7cbf86deb5494b16cb2ac7e6b6d9e1fe449df2

armi/operators/tests/test_operators.py
```diff
@@ -103,37 +105,101 @@ def test_loadStateError(self):
...
+
+    def test_snapshotRequest(self):
+        with TemporaryDirectoryChanger():
+            self.o.snapshotRequest(0, 1)
+
+
+class TestTightCoupling(unittest.TestCase):
+    def setUp(self):
+        self.cs = settings.Settings()
+        settings.setMasterCs(self.cs)
```
albeanth, line 127: @john-science this is the culprit.

john-science: I mean, if this HAS to go in for this PR, I will hold my nose and do it.

But we WILL have to add it to the list to remove ASAP:

https://github.com/terrapower/armi/issues/930

armi/operators/tests/test_operators.py
```diff
@@ -103,37 +111,96 @@ def test_loadStateError(self):
...
+                self.assertEqual(
+                    self.o.r.core.p.coupledIteration, hasCouplingInteraction
+                )
+
+    def test_performTightCoupling_NoWriteDB(self):
+        """ensure a tight coupling iteration accours and that a DB WILL NOT be written if requested"""
+        hasCouplingInteraction = 1
+        with directoryChangers.TemporaryDirectoryChanger():
+            self.dbWriteForCoupling(writeDB=False)
+            self.assertEqual(self.o.r.core.p.coupledIteration, hasCouplingInteraction)
```
opotowsky, line 199:
```suggestion
            with mockRunLogs.BufferLog() as mock:
                self.dbWriteForCoupling(writeDB=False)
                self.assertNotIn("Writing to database for statepoint:", mock.getStdout())
                self.assertEqual(self.o.r.core.p.coupledIteration, hasCouplingInteraction)
```

This is super optional...I just wanted to see some indication the db wasn't written

armi/operators/tests/test_operators.py
```diff
@@ -103,37 +111,96 @@ def test_loadStateError(self):
...
+        self.assertEqual(cs[CONF_RUN_TYPE], "Standard")
+
+    def test_snapshotRequest(self):
+        with TemporaryDirectoryChanger():
+            self.o.snapshotRequest(0, 1)
+
+
+class TestTightCoupling(unittest.TestCase):
+    def setUp(self):
+        self.cs = settings.getMasterCs()
```
opotowsky, line 132: did you try this with `settings.Settings()`?

armi/operators/tests/test_operators.py
```diff
@@ -103,37 +111,96 @@ def test_loadStateError(self):
...
+        self.o.cs = self.o.cs.modified(newSettings={"runType": "fake"})
+        self.assertEqual(self.o.cs[CONF_RUN_TYPE], "fake")
+
+        # validate the method works
+        cs = self.o.setStateToDefault(self.o.cs)
+        self.assertEqual(cs[CONF_RUN_TYPE], "Standard")
+
+    def test_snapshotRequest(self):
+        with TemporaryDirectoryChanger():
+            self.o.snapshotRequest(0, 1)
```
opotowsky, line 127:
```suggestion
            self.o.snapshotRequest(0, 1)
            self.assertTrue(os.path.exists("snapShot0_1"))
```

I abandoned a PR where I added an assert statement to this test. It's unrelated to your PR, but wanna sneak it in for me?

opotowsky (commented):
Some comments and suggestions

armi/operators/tests/test_operators.py
albeanth: Yeah... See https://github.com/terrapower/armi/pull/1172#discussion_r1103260925

I decided that maybe `getMasterCs` would be the lesser of the two evils between it and `setMasterCs`.

The problem is that if I don't call `setMasterCs` when getting the case settings, this method rears its head and causes the test to fail. 
https://github.com/terrapower/armi/blob/867ebecaa4f508239da4c3b4137c1b89beba8fbd/armi/operators/operator.py#L548-L563

albeanth: @john-science are you ok with the `getMasterCs` route? We'd just have to add it to the to-do list for #930.

opotowsky: Oh, darn! I did not read that clearly

I vote we get this in and you add it to the todo list. `setMasterCs` is used in every single unit test that calls `loadTestReactor`....so I'd imagine getting rid of it will take some work. Since there isn't a clear path to that yet, it shouldn't stop your test from being included in the code.

opotowsky (approved):
Thanks for adding a little extra pizazz to a unit test for me!

opotowsky:
I'm going to dismiss Mark's review since the changes he requested are addressed.

armi/operators/tests/test_operators.py
```diff
@@ -103,37 +111,96 @@ def test_loadStateError(self):
...
+        self.o.r = Reactor("empty", None)
+        self.o.r.core = Core("empty")
+
     def test_couplingIsActive(self):
-        """ensure that cs["tightCoupling"] controls couplingIsActive"""
-        self.assertFalse(self.o.couplingIsActive())
-        self.o.cs["tightCoupling"] = True
+        """ensure that cs[CONF_TIGHT_COUPLING] controls couplingIsActive"""
         self.assertTrue(self.o.couplingIsActive())
+        self.o.cs[CONF_TIGHT_COUPLING] = False
```
opotowsky, line 143: above you have `self.cs` defined than an operator init-ed from that, so can this be `self.cs[CONF_TIGHT_COUPLING] = False`?

armi/operators/tests/test_operators.py
opotowsky: I added it to https://github.com/terrapower/armi/issues/930

armi/operators/tests/test_operators.py
albeanth: It could since both `self.cs` and `self.o.cs` point to the same cs object in memory. My thoughts on the `self.o.cs` route: once we get rid of `set/getMasterCs`, `self.cs` and `self.o.cs` could be different. So the assertions should live on `self.o.cs` I think. What do you think?

PR #1177, opened 2023-02-10
Don't verify dimensions on irregular hex lattice.

mgjarrett:
## Description

`HexBlock.verifyDims()` verifies that the pins and wires defined for a block fit snugly into the duct. The logic for this check only works for a regular hex lattice. If an irregular hex lattice is defined using a grid in the blueprints, this won't work correctly. Adding code to skip the check in this case.

---

## Checklist


mgjarrett:
Closing this for now as it's not an immediate need and a simple fix like this probably won't cover all of the possible corner cases of non-regular hex lattices.

PR #1179, opened 2023-02-13
Remove unused HCFcoretype setting

mgjarrett:
## Description

The `HCFcoretype` setting is not used anywhere in the framework or in downstream TerraPower projects.

Since the only options are `TWR*`, I assume nobody outside of TerraPower is using this setting either.

TerraPower TH solvers use an assembly parameter to track hot channel factors: https://github.com/terrapower/armi/blob/05765608596bc907a88a6033cf5d4640f610ab7d/armi/reactor/assemblyParameters.py#L358-L366

I assume anybody else wanting to apply hot channel factors would want to use the same assembly parameter, and not this case setting.

---

## Checklist

### DIFF
--- a/armi/operators/tests/test_inspectors.py
+++ b/armi/operators/tests/test_inspectors.py
@@ -87,13 +87,7 @@ class TestInspector(unittest.TestCase):
 
     def test_assignCS(self):
         keys = sorted(self.inspector.cs.keys())
-        self.assertIn("HCFcoretype", keys)
-
-        self.assertEqual(self.inspector.cs["HCFcoretype"], "TWRC")
-        self.inspector._assignCS(
-            "HCFcoretype", "FAKE"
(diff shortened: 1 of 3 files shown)


mgjarrett:
The coverage reduction is from two uncovered lines here:

https://github.com/terrapower/armi/blob/05765608596bc907a88a6033cf5d4640f610ab7d/armi/runLog.py#L649-L651

I don't know how we were hitting the `FileExistsError` before, or how we could force it to be hit again. It's not clear how it relates to this PR.

opotowsky (approved):
Nice cleanup catch! Somehow I missed it when investigating the HCF settings...

That is a weird reason for a coverage drop. Sometimes it happens if the base branch isn't perfectly in sync with `main`, and even though github is smart enough to handle that situation, coveralls might not be.

PR #1181, opened 2023-02-13
Block Param Data Rotations

jasonbmeng:
## Description
Supporting mechanical team's use of `r.core.growToFullCore()`.

- Added block param rotation handling for non-list data types (ndarrays)
- Throws `TypeError` if attempting to rotate an unexpected data type
- Added unit testing for new data type handling

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_assemblies.py
+++ b/armi/reactor/tests/test_assemblies.py
@@ -38,7 +38,7 @@ from armi.reactor.assemblies import (
     numpy,
     runLog,
 )
-from armi.tests import TEST_ROOT
+from armi.tests import TEST_ROOT, mockRunLogs
 from armi.utils import directoryChangers
 from armi.utils import textProcessors
 from armi.reactor.tests import test_reactors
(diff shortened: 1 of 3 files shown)


opotowsky (approved):
Nice update and fantastic unit test. Looks good to me! 

I think once you sign the CLA we can get this merged when CI completes.

jasonbmeng:
Thanks @opotowsky! CLA signed.

opotowsky:
Hey @john-science! I think this PR looks great but there are a couple of `runLog` warnings that have been added. Since they aren't typically addressed in unit tests, I just wanted to make sure it's ok that this small cov drop is acceptable before merging.

And then there's this weird part of `runLog` itself that is uncovered -- I just saw it in a different PR too. So this PR is not at fault for that.

armi/reactor/tests/test_assemblies.py
```diff
@@ -1022,10 +1022,31 @@ def test_rotate(self):
...
+        self.assertEqual(a.getBlocks()[0].p.THcornTemp, 3)
+        b.p.THcornTemp = 4.0
+        a.rotate(math.radians(120))
+        self.assertEqual(a.getBlocks()[0].p.THcornTemp, 4.0)
+
+        # check that TypeError is raised for unexpected data type
+        b.p.THcornTemp = "bad data"
+        with self.assertRaises(TypeError):
+            a.rotate(math.radians(120))
+
```
jakehader, line 1049: I'm not sure if rotatePins is tested somewhere, but it might be good to test that as well. Also, you may want to add a test for the list equivalent since it looks like the rotate method works on lists or numpy arrays of length 6. Could you also add a test that fails to change a list of length != 6 and same for a numpy arrays?

jakehader (dismissed):

armi/reactor/tests/test_assemblies.py
opotowsky: `rotatePins` does have a unit test already

@jasonbmeng for writing tests where you need to test the output of the `runLog` like Jake is suggesting you would do the following:

```python
from armi.tests import mockRunLogs

b.p.THcornTemp = np.array([400, 450, 500]) 
    self.assertEqual("", mock.getStdout())
    a.rotate(math.radians(120))
    self.assertIn("No rotation performed", mock.getStdout())

# repeat this for a list with len 1-5 or > 6 too
```

jasonbmeng: Thanks for the feedback @jakehader 
@opotowsky mentioned there is already testing for `rotatePins`
The original unit test for `b.rotate()` has testing for lists.

Thanks for the mockRunLog info. I pushed a commit adding testing for lists of len 5 and 7, and ndarrays of len 5 and 7.

opotowsky:
Just to get this merged in faster, I'm dismissing @jakehader's change request since it has been addressed.

Issue #1166, opened 2023-02-08
Snapshot Request Improvements


jakehader:
Within the `operator.py` module, there is a method called `snapshotRequest`. This method makes some pretty specific assumptions about the data files that will exist in a user's directory, which are tied to TerraPower's specific use cases. It would be better to refactor this code base in a manner which allows plugins to define the which types of files to move for snapshots for their application. In general, there are several ways we are trying to handle moving data files around that can probably be cleaned up and unified.

Other examples:
- Setting `existingXSFiles`
- `specifyInputs` on the interfaces (run from `copyInterfaceInputs` in `case.py`
- Multiple locations in the code base on look for `ISOTXS` or wildcard / regex variants.

There are multiple use cases that should be considered:
- On case/case suite cloning we need the `Case` object to be able to detect which files to bring along. In most cases this can happen when a user uses a relative path versus an absolute path
- When a case is running and files are to be retrieved from a relative or absolute path. This can and does happen in the `crossSectionControl` setting.

Related Issues:
- [ ] https://github.com/terrapower/armi/issues/1160

PR #1182, opened 2023-02-14
Fix file copy of rzmflx files for snapshot runs

opotowsky:
## Description

The previous implementation of the file copy for rzmflx file presumed the files were already in the case run directory. But when absolute file paths are given for the `fluxLocation` setting in `crossSectionControl`, the files are very possibly not in the run directory already. 

This PR changes the snapshot file copy step to parse the whole of the `crossSectionControl` setting to copy any file to the snapshot directory, whether relative or absolute file path.

This is not an elegant way to handle this, but the issue is known (#1166) and will be addressed in the future.

---

## Checklist

### DIFF
--- a/armi/operators/tests/test_operators.py
+++ b/armi/operators/tests/test_operators.py
@@ -190,7 +190,7 @@ class OperatorTests(unittest.TestCase):
             fluxFile = "rzmflxYA"
             # update the case settings object so that the snapshot request will copy a file
             self.o.cs[CONF_CROSS_SECTION]["YA"].fluxFileLocation = os.path.join(
-                THIS_DIR, "../../physics/neutronics/tests", fluxFile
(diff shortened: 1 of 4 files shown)


armi/operators/tests/test_operators.py
```diff
@@ -183,7 +187,13 @@ def test_setStateToDefault(self):
 
     def test_snapshotRequest(self):
         with TemporaryDirectoryChanger():
+            fluxFile = "rzmflxYA"
+            self.o.cs[CONF_CROSS_SECTION]["YA"].fluxFileLocation = os.path.join(
+                THIS_DIR, "../../physics/neutronics/tests", fluxFile
+            )
```
opotowsky, line 193: stealing an empty file from elsewhere rather than adding one to this test dir. probably preferable to do the latter?

PR #1183, opened 2023-02-14
Fix user plugins case suite

bsculac:
## Description

When submitting a `case-suite` that uses `userPlugin`s, registration of the `userPlugin` with the plugin manager is attempted multiple times. The plugin manager does not allow this, and the `case-suite` fails.

To fix this, the registration of `userPlugin`s now checks if the user plugin name has already been registered with the plugin manager.

---

## Checklist

### DIFF
--- a/armi/apps.py
+++ b/armi/apps.py
@@ -283,6 +283,16 @@ class App:
 
         or on Windows:
         ``C:\\path\\to\\pluginMod.py:pluginCls``
+
+        Parameters
+        -----------
+        pluginPath : str
+            String path to a userPlugin.
+
+        Returns
+        --------
+        bool
+            Whether or not the plugin name is already registered with the manager.
         """
 
         if ":" in pluginPath:
--- a/armi/apps.py
+++ b/armi/apps.py
@@ -269,7 +269,7 @@ class App:
                 # The path is of the form: armi.thing.what.MyPlugin
                 self.__registerUserPluginsInternalImport(pluginPath)
 
-    def _isPluginRegistered(self, pluginPath:str):
+    def _isPluginRegistered(self, pluginPath: str):
         """
         Check if the plugin at the provided path is already registered.
 
@@ -284,11 +284,12 @@ class App:
         or on Windows:
         ``C:\\path\\to\\pluginMod.py:pluginCls``
         """
+
         if ":" in pluginPath:
             pluginName = pluginPath.strip().split(":")[-1]
         else:
             pluginName = pluginPath.strip().split(".")[-1]
-        
+
         return self._pm.has_plugin(pluginName)
 
     def __registerUserPluginsAbsPath(self, pluginPath):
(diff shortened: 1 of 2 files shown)


keckler:
Somehow I have never run into this issue... I wonder what is different between your case suite and mine.

armi/apps.py
```diff
@@ -260,13 +260,38 @@ def registerUserPlugins(self, pluginPaths):
         restrict their flexibility and power as compared to the usual ArmiPlugins.
         """
         for pluginPath in pluginPaths:
+            if self._isPluginRegistered(pluginPath):
```
john-science, line 263: The thing that convinces me that this PR is okay is that this is only implemented for `UserPlugin`s, not ALL `Plugin`s.

armi/apps.py
```diff
@@ -260,13 +260,38 @@ def registerUserPlugins(self, pluginPaths):
...
+                continue
             if ".py:" in pluginPath:
                 # The path is of the form: /path/to/why.py:MyPlugin
                 self.__registerUserPluginsAbsPath(pluginPath)
             else:
                 # The path is of the form: armi.thing.what.MyPlugin
                 self.__registerUserPluginsInternalImport(pluginPath)
 
+    def _isPluginRegistered(self, pluginPath: str):
```
john-science, line 272: I love the PR!

Do you think we could get a unit test of this new method?

If it helps, we already have a lot of other [UserPlugin unit tests](https://github.com/terrapower/armi/blob/main/armi/tests/test_user_plugins.py).

Or would you like a little help writing that test?

bsculac: I can take a look at that link and see about putting together a unit test. I wanted to see if this would be how you wanted to handle the problem :)

armi/tests/test_user_plugins.py
```diff
@@ -281,3 +281,29 @@ def test_userPluginWithInterfaces(self):
...
+        cs = caseSettings.Settings().modified(
+            caseTitle="test_registerUserPluginsFromSettings",
+            newSettings={
+                "userPlugins": [
+                    "armi.tests.test_user_plugins.UserPluginFlags3",
+                    "armi.tests.test_user_plugins.UserPluginFlags3",
+                ],
+            },
+        )
+        cs.registerUserPlugins()
```
john-science, line 309: Should we add some sort of `self.assert` statement to the end, proving that we only have one copy of the `UserPluginFlags3`?

Other than that, I really like the test!

It's clear and as concise as I think we're likely to get.

armi/apps.py
john-science: You made a great test! Thanks!

armi/tests/test_user_plugins.py
bsculac: Added a check for the number of `userPlugin`s in each stage of the test.

Issue #1198, opened 2023-02-24
Type hint conflicts with docstring in `densityTools`


mgjarrett:
Type hint indicates that `massFrac` is a `dict`, but the doc string says it is a `float`. The usage in the function seems consistent with a float.

https://github.com/terrapower/armi/blob/3e47ec6e6359870866502c9356747cf1da4f9515/armi/utils/densityTools.py#L316-L336

john-science:
We actually have a unit test for this:

https://github.com/terrapower/armi/blob/3e47ec6e6359870866502c9356747cf1da4f9515/armi/utils/tests/test_densityTools.py#L71-L84

So, you are correct. It appears the confusion was that the RETURN TYPE is a dict, but yes the input `massFrac` is a float.

PR #1199, opened 2023-02-24
Enforce the assignment of a mass fraction to a material to be a float

jakehader:
## Description

This pull request updates the `setMassFrac` implementation on the generic material class to ensure that a `float` is assigned in the material's `massFrac` dictionary. This is the expected behavior of the method based on the type hints in the method.

---

## Checklist


armi/materials/material.py
```diff
@@ -240,6 +240,37 @@ def setDefaultMassFracs(self):
...
+        """
+        try:
+            massFrac = float(massFrac)
+        except Exception as ee:
+            raise TypeError(
+                f"Error is converting the mass fraction of {massFrac} "
+                f"for nuclide {nucName} in {self} to a float. "
+                f"Exception: {ee}"
+            )
+        if massFrac < 0.0:
```
john-science, line 260: I take it we've seen bad mass fractions in the wild lately?

jakehader: I haven't seen it, but I am trying to guard against something to make sure we don't have an error like that. Also, it was trying to address https://github.com/terrapower/armi/issues/1198 and I was digging into the mass fractions and how they were being set for materials.

armi/materials/material.py
```diff
@@ -240,6 +240,37 @@ def setDefaultMassFracs(self):
...
+            raise ValueError(
+                f"Cannot assign a negative mass "
+                f"fraction of {massFrac} to {nucName} "
+                f"within {self}"
```
john-science, line 264: I think this `{self}` will print something hideous.  Have you tried it, to know you like it?

You could also try: `{self.__class__.__name__}`, though I'm not sure f-string formatting would allow that, that's the prettier version for the material class.

jakehader: I think that `self` will print:

image

armi/materials/material.py
```diff
@@ -240,6 +240,37 @@ def setDefaultMassFracs(self):
...
+        -----
+        This will try to convert the provided ``massFrac`` into a float
+        for assignment. If the conversion cannot occur then an error
+        will be thrown.
+        """
+        try:
+            massFrac = float(massFrac)
+        except Exception as ee:
+            raise TypeError(
+                f"Error is converting the mass fraction of {massFrac} "
```
keckler, line 256: Typo "is" -> "in"

armi/materials/material.py
```diff
@@ -240,6 +240,37 @@ def setDefaultMassFracs(self):
...
+        if massFrac < 0.0:
+            raise ValueError(
+                f"Cannot assign a negative mass "
+                f"fraction of {massFrac} to {nucName} "
+                f"within {self}"
+            )
+
+        if massFrac > 1.0:
+            raise ValueError(
+                f"Cannot assign a negative mass "
```
keckler, line 269: This error message is not correct.

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -234,7 +234,7 @@ def _constructMaterial(self, blueprint, matMods):
         if len(matMods) > 1:
             # don't apply if only customIsotopics is in there
             try:
-                # update material with updated input params from blueprints file.
+                # update material with updated input params from blueprints file.]
```
john-science, line 237: Added a random character to the end of this line.

john-science: This is minor, but let's fix it and we can Squash&Merge this PR.

jakehader:
@keckler, @john-science - I believe this is ready for merging now.

keckler:
I've been wondering why this PR is being put through. Could you elaborate on why these checks are being added? It just seems a bit random without any context.

At first I was concerned that this added type check would have performance implications because I assumed that `setMassFrac()` was being used anytime the isotopics were being changed (i.e. burnup). But I was surprised when I looked into the usage of `setMassFrac()`. It's basically only used to initialize materials, meaning only at the beginning of a run.

So I'm fine with this, but I don't see right away why it's needed all of the sudden.

jakehader:
@keckler - This point of this change is to fix the type for the massFrac and then to provide some defensive checks against implementing a material that does not meet the expected constraints on fractions. Ideally we would be able to resolve that all mass fractions sum 1.0 or to enforce a normalization. I didn't go as far as to change the behavior so that you can to set all mass fractions at once and check it all because it would be a significant API change and impact many materials downstream.

With that being said, I think that this is still a reasonably small change that should help without much performance impact since as you said this only happens on material instantiation. What do you think @keckler and @john-science?

keckler:
I think John and I understand the change, but were a bit confused why it
was being implemented. It seems not pythonic to add a check like this. One
could then wonder why we dont implement more type checks throughout ARMI.

I dont really care all that much, just seems odd. I'll let John be the
gatekeeper here.

john-science:
> I think John and I understand the change, but were a bit confused why it was being implemented. It seems not pythonic to add a check like this. One could then wonder why we dont implement more type checks throughout ARMI. I dont really care all that much, just seems odd. I'll let John be the gatekeeper here.

Yeah, I have been holding off on this PR, not because I think it's wrong or broken, but because I don't see the need.  Python has a really soft type system, so it is considered the classic wrong idea to check the types of everything in your code base all the time.  As Chris said "it's not Pythonic".

It _feels_ like all we need is this one line:

https://github.com/terrapower/armi/blob/massFrac/armi/materials/material.py#L246

And if _that_ fails, great. Let it fail. The error will be obvious to everyone who sees it.

jakehader:
> > I think John and I understand the change, but were a bit confused why it was being implemented. It seems not pythonic to add a check like this. One could then wonder why we dont implement more type checks throughout ARMI. I dont really care all that much, just seems odd. I'll let John be the gatekeeper here.
> 
> Yeah, I have been holding off on this PR, not because I think it's wrong or broken, but because I don't see the need. Python has a really soft type system, so it is considered the classic wrong idea to check the types of everything in your code base all the time. As Chris said "it's not Pythonic".
> 
> It _feels_ like all we need is this one line:
> 
> https://github.com/terrapower/armi/blob/f3f054b2120ac90997deb8b1c92c7cb479de7b2f/armi/materials/material.py#L246
> 
> And if _that_ fails, great. Let it fail. The error will be obvious to everyone who sees it.

That is fair to remove the type try/except but I still see the check on the range of the float bring valuable to guard against accidentally implementing a nuclide in a material with negative or greater than 1 mass fraction. Maybe the argument is that this would be verified in testing of a material class though.

PR #1203, opened 2023-03-02
Speed up runLog

opotowsky:
## Description

(trying to avoid tagging stilley for reviews if possible)

A profile of a long case run showed that **4.5 hours** was being spent inside `runLog`, and digging further in, specifically evaluating `os.path.exists`. 

~This proposed change cut that time down to 2.3 hours (just not running `os.path.exists` if the log level isn't applicable)~ Chris' suggestion that `os.path.exists` can be module level has the time spent in `runLog` down to **9 minutes**.

~However, I'd like to consider other options on how this speedup can happen. Would moving it to `__init__` be possible? Or is running it only one time problematic? There are some places where this gets evaluated thousands of times in a loop.~

Update: This was evaluated using a pretty hefty run that included tight coupling. The time spent in `runLog` without tight coupling enabled was never all that bad / didn't see much improvement. But yay for the tight coupling cases!

---

## Checklist

### DIFF
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -522,12 +522,6 @@ class RunLogger(logging.Logger):
         # Determine the log level: users can optionally pass in custom strings ("debug")
         msgLevel = msgType if isinstance(msgType, int) else LOG.logLevels[msgType][0]
 
-        # os.path.exists is slow, only call this if writing is needed
-        if self.isEnabledFor(msgLevel):
-            # If the log dir hasn't been created yet, create it.
-            if not os.path.exists(LOG_DIR):
-                createLogDir(LOG_DIR)
-
         # Do the actual logging
         logging.Logger.log(
             self, msgLevel, str(msg), extra={"single": single, "label": label}
@@ -663,6 +657,8 @@ def createLogDir(logDir: str = None) -> None:
         time.sleep(secondsWait)
 
 
+if not os.path.exists(LOG_DIR):
+    createLogDir(LOG_DIR)
 # ---------------------------------------
 
 
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -519,12 +519,14 @@ class RunLogger(logging.Logger):
         In this situation, we do the mangling needed to get the log level to the correct number.
         And we do some custom string manipulation so we can handle de-duplicating warnings.
         """
-        # If the log dir hasn't been created yet, create it.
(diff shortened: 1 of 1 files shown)


john-science:
That's awesome!

It would also help to speed up `runLog` if we weren't logging a million useless things every time node.  Everyone seems convinced while they are working on a PR that their new log line needs to be info, not debug. Because it was really interesting to them that day. But they never look at the log line again. 99.9% of the stuff we log at info level is read less than once a year and should be moved to debug.

keckler:
I would agree that, just by looking at the code, it appears the logDir could be created during the `__init__()`.

I even think it could go up higher somewhere? So when `createLogDir()` is called in this instance, it always passes in `LOG_DIR`:
https://github.com/terrapower/armi/blob/8d3c64ff5899f40a03c3ac56ba6412c552eb4a34/armi/runLog.py#L526-L529

But `LOG_DIR` is a _module_ variable:
https://github.com/terrapower/armi/blob/8d3c64ff5899f40a03c3ac56ba6412c552eb4a34/armi/runLog.py#L67-L71

If `LOG_DIR` never changes during the run, which it doesn't, then the call to `createLogDir(LOG_DIR)` will always produce the same result. So I think it should be safe just to call `createLogDir(LOG_DIR)` from the module level and hold that as a module object, right?

I don't know... I think what I said is correct, but then it makes it seem odd that one is able to supply a `logDir` parameter to `createLogDir()` in the first place, because I don't see that ability being exercised anywhere...

opotowsky:
Thanks for looking into it, @keckler. I agree. The only issue is if someone deletes the log dir during the run, which would just be weird and not a situation I think we should bother coding for. Just wondering if there's some other situation I'm not seeing that would require this to be done at this stage. I'm going to put it at the module level and run another (much shorter) profile comparison.

opotowsky:
Ok, @keckler and @albeanth ....took awhile to get the testing done right because I learned the hard way this isn't a problem for non-tight coupling cases. (Plus, all kinds of other  slowdowns that need not be mentioned). Updated the PR description and I'm ready for review!

@monufer tl;dr runLog on tight coupling cases is 4.5 hrs --> 9 minutes. Great catch!

keckler (approved):
Such a tiny change for such a big impact.

Thanks for discovering this!

albeanth (approved):

Issue #935, opened 2022-10-11
Formal Documentation of Axial Expansion in ARMI


albeanth:
Can we have some documentation added to the User Docs that details all that one needs to know related to thermal expansion in ARMI? I'm thinking an aggregation of all of:
- relevant case settings
- relevant blueprints attributes
- relevant flags
- how to use it
- how to turn it off
- what to expect when it is on/off
- what it can/can't handle
- assumptions
- others...

To me, this would be incredibly valuable and would save many users lots of time.

I don't think such a thing should be part of the current PR, but reading through this PR left me craving exactly such a compilation because it is difficult (for me) to keep track of all of the many pieces.

_Originally posted by @keckler in https://github.com/terrapower/armi/issues/931#issuecomment-1275240939_

albeanth:
I have started a feature branch that contains a standalone jupyter notebook that attempts to address some of these... But I think adding all of the other features that @keckler has requested here is likely outside the scope of a single jupyter notebook. And it's not immediately clear to me where we would put such documentation in the [ARMI User Docs](https://terrapower.github.io/armi/user/index.html). 

@john-science @ntouran @opotowsky Do you have any preferences?

keckler:
Perhaps here: https://terrapower.github.io/armi/user/inputs/settings.html#some-special-settings

albeanth:
> Perhaps here: https://terrapower.github.io/armi/user/inputs/settings.html#some-special-settings

Ooh, I like it.

opotowsky:
It might also fit as a [Tutorial](https://terrapower.github.io/armi/tutorials/index.html)? I know they are more interaction focused but for something move involved like this I could see an argument for that

albeanth:
@opotowsky that was what I initially intended with the jupyter notebook. There might be an argument to have documentation in multiple places.

jakehader:
I would suggest the gallery for examples, since that is where demonstration scripts should go.

john-science:
I think a Jupyter notebook is a great idea. As long as it has (or comes with) actual test describing things.  Working examples are great, when they come with lots of conversation.  I would add an RST file along with the Notebook.

albeanth:
Since GitHub doesn't feel like linking this.... 

https://github.com/terrapower/armi/discussions/1154#discussioncomment-4959351

albeanth:
@john-science @mgjarrett @keckler 

What do you think about adding a section in the ARMI docs (probably in the user-docs) that describes:
[...]

PR #1204, opened 2023-03-07
Add Thermal Expansion/Contraction Methodology to Docs

albeanth:
## Description
This is the first of a series of PRs that will add documentation for radial and axial expansion methodologies in ARMI.

This PR adds the methodology used in ARMI to perform linear expansion by creating a new page/section within `doc/user` and cleans up the docstring for `Material.linearExpansionFactor`. 

Cross linking #935 as it is related. 

Reviewers, please pull the feature branch and compile the ARMI docs to confirm that they build correctly. There are no new dependencies, but it will be good to confirm. 

---

## Checklist

### DIFF
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -212,7 +212,7 @@ class Material:
 
         See Also
         --------
-        :py:meth: `armi.materials.material.Material.linearExpansionPercent`
+        linearExpansionPercent
         """
         dLLhot = self.linearExpansionPercent(Tc=Tc)
         dLLcold = self.linearExpansionPercent(Tc=T0)
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -190,19 +190,13 @@ class Material:
 
     def linearExpansionFactor(self, Tc: float, T0: float) -> float:
         """
-        Return a dL/L factor relative to T0 instead of to the material-dependent reference
-        temperature. This factor dL/Lc is a ratio and will be used in dimensions through the
-        formula::
+        Return a dL/L factor relative to T0 instead of the material-dependent reference
+        temperature.
 
-            dim = dim0*(1+dLL).
-
-        If there is no dLL, it should return 0.0
-
-        calculate thermal expansion based on dL/L0, which is dependent on the mat-dep ref temp.::
-
-             L(T) = L0(1+dL/L0)
-             (Lh-Lc)/L0 = dL/L0(Th) - dL/L0(Tc)
-             (Lh-Lc)/Lc = (Lh-Lc)/L0 * L0/Lc = (dL/L0(Th)-dL/L0(Tc)/L0 / (dL/L0(Tc)+1.0)
+        Notes
+        -----
+        For a detailed description of the linear expansion methodology, see
+        :ref:`thermalExpansion` in the documentation.
 
         Parameters
         ----------
@@ -213,13 +207,12 @@ class Material:
 
         Returns
         -------
-        dL/L_fromCold : float
-            The average thermal expansion between T_current and T0
+        dLL: float
+            The average thermal expansion between Tc and T0
 
         See Also
         --------
-        linearExpansionPercent
-        components.Component.getThermalExpansionFactor
+        :py:meth: `armi.materials.material.Material.linearExpansionPercent`
         """
         dLLhot = self.linearExpansionPercent(Tc=Tc)
         dLLcold = self.linearExpansionPercent(Tc=T0)
--- a/doc/user/radial_and_axial_expansion.rst
+++ b/doc/user/radial_and_axial_expansion.rst
@@ -2,26 +2,85 @@
 Radial and Axial Expansion and Contraction
 *******************************************
 
-ARMI natively supports both one-dimensional radial and axial expansion. These expansion types function independently of one another and each have their own set of underlying assumptions and use-cases. In the following sections we describe their design, limitations, and intended functionality.
+ARMI natively supports both one-dimensional radial and axial expansion. These expansion types function independently of one another and each have their own set of underlying assumptions and use-cases. The remainder of this section is described as follows: in Section :ref:`thermalExpansion` the methodology used for thermal expansion within ARMI is described; in Sections :ref:`radialExpansion` and :ref:`axialExpansion`, we describe the design, limitations, and intended functionality of radial and axial expansion, respecitely.
+
+.. _thermalExpansion:
 
 Thermal Expansion
 -----------------
 ARMI treats thermal expansion as a linear phenomena using the standard linear expansion relationship,
 
 .. math::
-    :linearExp:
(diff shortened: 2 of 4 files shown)


albeanth:
@john-science @alexhjames @onufer @jakehader just a heads up.

armi/materials/material.py
```diff
@@ -190,19 +190,13 @@ def linearExpansionPercent(self, Tk: float = None, Tc: float = None) -> float:
...
         """
-        Return a dL/L factor relative to T0 instead of to the material-dependent reference
-        temperature. This factor dL/Lc is a ratio and will be used in dimensions through the
-        formula::
+        Return a dL/L factor relative to T0 instead of the material-dependent reference
+        temperature.
 
-            dim = dim0*(1+dLL).
-
-        If there is no dLL, it should return 0.0
```
keckler, line 199: Unless the statement on this line is no longer true, I think this line should stay.

doc/user/radial_and_axial_expansion.rst
```diff
@@ -0,0 +1,86 @@
+*******************************************
+Radial and Axial Expansion and Contraction
+*******************************************
+
+ARMI natively supports both one-dimensional radial and axial expansion. These expansion types function independently of one another and each have their own set of underlying assumptions and use-cases. The remainder of this section is described as follows: in Section :ref:`thermalExpansion` the methodology used for thermal expansion within ARMI is described; in Sections :ref:`radialExpansion` and :ref:`axialExpansion`, we describe the design, limitations, and intended functionality of radial and axial expansion, respecitely.
```
keckler, line 5: "respectively" is misspelled.

I might also suggest to remove the phrase "one-dimensional", just because I find it kinda confusing to read. I don't know whether you mean to apply that to "radial" or to both "radial" and "axial".

doc/user/radial_and_axial_expansion.rst
```diff
@@ -0,0 +1,86 @@
...
+    \frac{\Delta L}{L_0} = \alpha(T) \Delta T,
+    :label: linearExp
+
+where, :math:`\Delta L` and :math:`\Delta T` are the change in length and temperature from the reference state, respectively, and :math:`\alpha` is the thermal expansion coefficient relative to :math:`T_0`. Expanding Equation :eq:`linearExp` we obtain,
+
+.. math::
+    \frac{L_1 - L_0}{L_0} = \alpha(T_1)\left(T_1 - T_0\right).
+    :label: linearExp2
+
+Rearranging Equation :eq:`linearExp2`, we can obtain an expression for the new length, :math:`L_1`,
```
keckler, line 23: I think you could probably just skip directly to this equation.

keckler: Up to you.

doc/user/radial_and_axial_expansion.rst
```diff
@@ -0,0 +1,86 @@
...
+    :label: linearExpansionFactor
+
+Equation :eq:`linearExpansionFactor` is the expression used by ARMI in :py:meth:`linearExpansionFactor <armi.materials.material.Material.linearExpansionFactor>`.
+
+.. note::
+    :py:meth:`linearExpansionPercent <armi.materials.material.Material.linearExpansionPercent>` returns :math:`\frac{L - L_0}{L_0}` in %. The expression used in :py:meth:`linearExpansionFactor <armi.materials.material.Material.linearExpansionFactor>` can be found as follows,
+    
+    .. math::
+        \frac{L_h - L_c}{L_c} &= \frac{\frac{\alpha(T_h) \left(T_h - T_0\right)}{100} - \frac{\alpha(T_c)\left(T_c - T_0\right)}{100}}{\frac{1 + \alpha(T_c)\left(T_c - T_0\right)}{100}}, \\
+        &= \frac{\alpha(T_h) \left(T_h - T_0\right) - \alpha(T_c)\left(T_c - T_0\right)}{100 + \alpha(T_c)\left(T_c - T_0\right)}.
```
keckler, line 76: I don't think writing out this math is really needed, as it's just multiplying and dividing by 100 and then simplifying that factor. Probably easier for the reader to save the text, IMO.

keckler (approved):
Absolute poetry, I can't wait to read the next chapter!

opotowsky (approved):
Sorry for the delay! This looks great and is easy to follow. 

I don't think I noticed any errors that Chris didn't already point out. Also, I agree you don't need the extra line of math getting from pct to frac. 

We need a lot more of this, so thanks for leading by example!

doc/user/radial_and_axial_expansion.rst
albeanth: How about changing "one-dimensional" to linear? The goal here is to show that we aren't doing true volumetric expansion, but rather approximating it via linear expansion.

albeanth: https://github.com/terrapower/armi/pull/1204/commits/656cffbcf8170112b6d8f96d0b2c710d73115ef1

doc/user/radial_and_axial_expansion.rst
albeanth: https://github.com/terrapower/armi/pull/1204/commits/656cffbcf8170112b6d8f96d0b2c710d73115ef1

doc/user/radial_and_axial_expansion.rst
albeanth: https://github.com/terrapower/armi/pull/1204/commits/656cffbcf8170112b6d8f96d0b2c710d73115ef1

armi/materials/material.py
albeanth: So this statement is a little strange to me. 

It's not necessarily _wrong_, but I think is trivial. I interpret it as "if there is no thermal expansion, the linear expansion factor will be 0.0". Which is obvious, no?

keckler: I think it is important so that people don't create a new material and have it return `None` or something like that. 

We should be able to rely upon an actual float being returned, otherwise downstream things will break.

To me, that's the importance.

doc/user/radial_and_axial_expansion.rst
keckler: Sure, then my suggested rephrase would be something like

"ARMI natively supports linear expansion in both the radial and axial dimensions."

keckler:
I'm good with the changes.

doc/user/radial_and_axial_expansion.rst
albeanth: I like your suggestion. That's much better. 
https://github.com/terrapower/armi/pull/1204/commits/01ce52ae8e303949f4b776ccd870764f361bceb8

armi/materials/material.py
albeanth: https://github.com/terrapower/armi/pull/1204/commits/45ea38345dd50fa9e85d720a851c27704432b179

PR #1206, opened 2023-03-07
Fix tight coupling on lattice interface for Snapshots

albeanth:
## Description
For tight coupling, setting the requirement that cross sections be regenerated only when timeNode == 0 was made with only Standard run types with constant power, flow, and/or temperature in mind. When doing snapshot runs with tight coupling, this assumption becomes invalid and it is desirable to _always_ regenerate cross sections. This PR addresses this and modifies the unit testing for the lattice physics interface to test for this. 

The unit tests within the lattice physics interface were also somewhat lacking. So some additional unit tests were added to get the coverage up. 

A few slight edits to documentations were also added. 

---

## Checklist


albeanth:
@onufer @jakehader a heads up.

armi/physics/neutronics/latticePhysics/latticePhysicsInterface.py
```diff
@@ -224,7 +228,7 @@ def interactCoupled(self, iteration):
         iteration : int
             This is unused since cross sections are generated on a per-cycle basis.
         """
-        if self.r.p.timeNode == 0:
+        if self.cs[CONF_RUN_TYPE] == "Snapshots" or self.r.p.timeNode == 0:
```
onufer, line 231: Add a comment describing why this is nesseasry and the logic

something like snapshots may be lower power and different temperature destruction, so it must run in coupled since the interactBOC call runs prior to TH temperature updates.

albeanth: Addressed in https://github.com/terrapower/armi/pull/1206/commits/65edad525798b462c55fc0ce7d1bf7bb10fdb1b9. Let me know what you think?

mgjarrett:
We're adding this because we're using snapshots to reload from a given statepoint and then run at a different power/flow/temperature condition from the original case.

For now, we are still only updating XS at time node 0 for standard runs because the assumption is that a standard run will have similar power/flow/temperature conditions at all time nodes. This is not necessarily true. With detailed cycle histories, one could burn a reactor at 100% power / 100% flow for 50 days, then drop to 50% power / 50% flow and burn for another 50 days, then drop to 5% power / 20% flow and burn 10 days, etc. If we were to deviate far from the power/flow ratio = 1.0, we would likely want to update XS.

It feels like this caveat above should be documented somewhere, but I don't know the best place. Could be the docstring for this `interactCoupled`, or the user docs for tight coupling (or both).

armi/physics/neutronics/latticePhysics/tests/test_latticeInterface.py
```diff
@@ -19,38 +19,71 @@
...
+        cls.o = Operator(settings.Settings())
+        cls.o.r = Reactor("empty", None)
+        cls.o.r.core = Core("empty")
+        xsGroupInterface = CrossSectionGroupManager(cls.o.r, cls.o.cs)
+        cls.latticeInterface = LatticeInterfaceTester(cls.o.r, cls.o.cs)
+        cls.o.addInterface(xsGroupInterface)
+        cls.o.addInterface(cls.latticeInterface)
+
+    def setUp(self):
+        self.o.r.core.lib = "Nonesense"
```
mgjarrett, line 58: not important, but
```suggestion
        self.o.r.core.lib = "Nonsense"
```
yay or nay?

armi/physics/neutronics/latticePhysics/tests/test_latticeInterface.py
```diff
@@ -19,38 +19,71 @@
...
+        self.latticeInterface.interactCoupled(iteration=0)
+        self.assertIsNone(self.o.r.core.lib)
 
-        i.updateXSLibrary(0)
-        self.assertEqual(len(i._oldXsIdsAndBurnup), 0)
+    def test_interactCoupled_TimeNode1(self):
+        """make sure updateXSLibrary is NOT run"""
+        self.o.r.p.timeNode = 1
+        self.latticeInterface.interactCoupled(iteration=0)
+        self.assertEqual(self.o.r.core.lib, "Nonesense")
```
mgjarrett, line 86:
```suggestion
        self.assertEqual(self.o.r.core.lib, "Nonsense")
```

keckler:
What @mgjarrett said about varying power levels in a cycle is incredibly important for the user to understand, and is probably something we need to add better treatment for.

albeanth:
> For now, we are still only updating XS at time node 0 for standard runs because the assumption is that a standard run will have similar power/flow/temperature conditions at all time nodes. This is not necessarily true. With detailed cycle histories, one could burn a reactor at 100% power / 100% flow for 50 days, then drop to 50% power / 50% flow and burn for another 50 days, then drop to 5% power / 20% flow and burn 10 days, etc. If we were to deviate far from the power/flow ratio = 1.0, we would likely want to update XS.

This is an excellent point. I think it would be a worthy feature request in an issue. 

> It feels like this caveat above should be documented somewhere, but I don't know the best place. Could be the docstring for this interactCoupled, or the user docs for tight coupling (or both).

@jhader and I attempted to provide high-level justification for the time node == 0 case in the docstring for `interactCoupled` when the method was introduced. We could probably be more precise by calling out the expectation of consistent hot full power. Ironically, the tight coupling documentation directly contradicts this and shows the cross sections being updated at each time step.

keckler:
We can put a warning here: https://terrapower.github.io/armi/user/inputs/settings.html#cycle-history

albeanth:
Looks like we're getting a massive drop in coverage for `latticePhysicsInterface`. I think this is due to simply removing the call to `loadTestReactor`. So the drop may not actually be all that bad... The unit tests for the `latticePhysicsInterface` prior to this PR were very bare. 

I will investigate, but I think we have two options: 
1. bring back `loadTestReactor` . Which is a bit unnecessary since we don't need a full reactor for the existing interface tests and gives a bit of false coverage, in my opinion.
2. Add a bunch of dedicated unit tests for `latticePhysicsInterface`. That can either be in this PR or a separate PR. 

@john-science do you have a preference?

albeanth:
@mgjarrett @keckler - I added a couple of commits reflecting your input on the clarity of the docstring and adding a note to the detailed cycle history docs. 
 https://github.com/terrapower/armi/pull/1206/commits/af765f4e9920171a8903a2587af5b5a4593d1598
https://github.com/terrapower/armi/pull/1206/commits/78547da22a87e37ae05025aa10ce1f12338cb52e

armi/physics/neutronics/latticePhysics/latticePhysicsInterface.py
onufer: comment above the line of code would be more helpful in my opinion.

albeanth: https://github.com/terrapower/armi/pull/1206/commits/66eed254772a3cea047f9a29d9973becfe0a487c

onufer (approved):

keckler (approved):

Issue #1078, opened 2023-01-09
Resolve mix of lower and upper case file extensions


opotowsky:
The following file extensions are defined as upper case:

https://github.com/terrapower/armi/blob/42726110a72606b9a45e64238582a2841e889f2e/armi/physics/neutronics/__init__.py#L138-L142

But there are [test files that are lower case](https://github.com/terrapower/armi/tree/main/armi/nuclearDataIO/tests/fixtures). This mix causes issues on Linux (but not on Windows, where file paths are not case sensitive). One standard should be chosen and implemented to avoid failure/confusion on linux unit testing. 

If we were to edit these defaults to be lower case, would any linux users be affected? @jakehader @john-science

john-science:
Ah, some old-school, niche file extensions were saved as upper case.

By preference, file extensions should be lowercase.  

@jakehader Thoughts?

mgjarrett:
Note that these constants are not necessarily always used as extensions. DIF3D generates files with these exact names, upper case, without an extension. So maybe we need to split into two constants: one for a filename, and one for a file extension? That's kind of complicated but I think we'll find other problems if we change these constants to be lower case.

jakehader:
File extensions as lower case make sense to me, but the files themselves like ISOTXS, COMPXS, PMATRX, etc are always upper case coming from ANL software.

opotowsky:
Oh, yikes. That is complicated then. Hm, I'm leaning towards following Mike's suggestion in that case.

john-science:
Well, Jake throws the monkey wrench into the easy fix, Arrielle.  BUT it's still fixable.  And we just love any problem that has a solution.

opotowsky:
I'll be on this ~soon... It's a medium term priority

opotowsky:
> I'll be on this ~soon... It's a medium term priority

I'll be on this today, it's a today priority 😄

PR #1216, opened 2023-03-13
Force GAMISO/PMATRX file path extensions to be lower case

opotowsky:
## Description

Addresses #1078 to provide support for linux

---

## Checklist

### DIFF
--- a/armi/nuclearDataIO/__init__.py
+++ b/armi/nuclearDataIO/__init__.py
@@ -115,9 +115,7 @@ def getExpectedGAMISOFileName(cycle=None, suffix=None, xsID=None):
     getExpectedCOMPXSFileName
     getExpectedPMATRXFileName
     """
-    return _findExpectedGammaFileName(
-        neutronics.GAMISO, _getGammaKeywords(cycle, suffix, xsID)
-    )
+    return _findExpectedGammaFileName("gamiso", _getGammaKeywords(cycle, suffix, xsID))
 
 
 def getExpectedPMATRXFileName(cycle=None, suffix=None, xsID=None):
@@ -135,9 +133,7 @@ def getExpectedPMATRXFileName(cycle=None, suffix=None, xsID=None):
     getExpectedCOMPXSFileName
     getExpectedGAMISOFileName
     """
-    return _findExpectedGammaFileName(
-        neutronics.PMATRX, _getGammaKeywords(cycle, suffix, xsID)
-    )
+    return _findExpectedGammaFileName("pmatrx", _getGammaKeywords(cycle, suffix, xsID))
 
 
 def _findExpectedGammaFileName(fileType, fileNameKeywords):
--- a/armi/nuclearDataIO/__init__.py
+++ b/armi/nuclearDataIO/__init__.py
@@ -115,7 +115,13 @@ def getExpectedGAMISOFileName(cycle=None, suffix=None, xsID=None):
     getExpectedCOMPXSFileName
     getExpectedPMATRXFileName
     """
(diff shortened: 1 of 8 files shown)


albeanth:
@mgjarrett feel free to remove yourself as a reviewer if you don't want to/have time to review. I just figured having someone review who is a regular user of the affected downstream projects would be a good idea.

opotowsky:
> @mgjarrett feel free to remove yourself as a reviewer if you don't want to/have time to review. I just figured having someone review who is a regular user of the affected downstream projects would be a good idea.

Thanks @albeanth! I'm trying to tag ppl besides John for review for the next few weeks, and Mike makes sense. 

@mgjarrett I did run a test case internally to be sure this didn't break it.

albeanth (approved):
Approving, though I am less familiar with why DIF3D/MC2 creates files called GAMISO and then files with the extension *.gamiso. That seems like a rather error-prone approach. But I'm sure there are good reasons.

Before merging, please ensure that the usual downstream internal tests are not affected. Thanks, @opotowsky !

armi/nuclearDataIO/__init__.py
```diff
@@ -115,9 +115,13 @@ def getExpectedGAMISOFileName(cycle=None, suffix=None, xsID=None):
     getExpectedCOMPXSFileName
     getExpectedPMATRXFileName
     """
-    return _findExpectedGammaFileName(
-        neutronics.GAMISO, _getGammaKeywords(cycle, suffix, xsID)
-    )
+    if any(i is not None for i in (cycle, suffix, xsID)):
```
keckler, line 118: The list comprehension here is not needed, I think. You can just do `any()` on the tuple directly.

keckler: Same comment on instance below.

keckler (approved):

armi/nuclearDataIO/__init__.py
opotowsky: Oh my goodness, way to overcomplicate it.... Thanks so much!

opotowsky, line 118: Actually, the complexity is needed....if `cycle=0` then `any((0, None, None)) = False`. Any/All apparently check for false-y stuff, not `None` explicitly.

opotowsky: Actually, the complexity is needed....if `cycle=0` then `any((0, None, None)) = False`. Any/All apparently check for false-y stuff, not `None` explicitly.

mgjarrett:
> Approving, though I am less familiar with why DIF3D/MC2 creates files called GAMISO and then files with the extension *.gamiso. That seems like a rather error-prone approach. But I'm sure there are good reasons.

DIF3D/MC2 binary outputs are typically 6 upper-case characters: ISOTXS, GAMISO, PMATRX, etc.. We wanted to create multiple different files for different XS types and store them in the same directory, so we started using the convention of `AA.gamiso`, `AB.gamiso`, etc. It seems like ARMI got a bit sloppy with respect to the case-sensitivity of the `*.gamiso` and `*.pmatrx` extensions, probably because all of the testing/development was done on Windows for a long time.

mgjarrett:
Looks good to me!

If we were starting from scratch, I would advocate for the extensions to be upper case, but we have a whole bunch of files that have already been generated in lower case, so we would be causing ourselves an enormous headache without much of a good reason. Best to stick with the lower case extensions unless we find a very good reason to switch.

mgjarrett (approved):

PR #1218, opened 2023-03-20
Add copyOutput option to globalFluxInterface

mgjarrett:
## Description

Add option not to copy the output back from globalFluxInterface.

Sometimes the `globalFluxInterface` is only run to get a k-eigenvalue for a given case. Even if this is the case, the `globalFluxInterface` will still copy the full output from the physics kernel back to the working directory, which can be 50 MB+. When running many hundreds or thousands of cases to calculate k-effective from various perturbations, this generates an unwieldly number and total size of output files that aren't needed. The `noOutput` option would allow a case to call the `globalFluxInterface`, grab k-effective from the output, and then skip the step of copying it back to the working directory.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
@@ -14,7 +14,6 @@
 """Tests for generic global flux interface"""
 # pylint: disable=missing-function-docstring,missing-class-docstring,protected-access,invalid-name,no-self-use,no-method-argument,import-outside-toplevel
 import unittest
-import os
 
 import numpy
 


mgjarrett:
Only a single line was added, and it is in an `__init__` function that is already covered. No new unit tests are needed to cover this feature.

albeanth:
@mgjarrett and I are currently discussing the effects this has on downstream applications. Once that has been settled, we will move forward with this PR.

doc/release/0.2.rst
```diff
@@ -10,6 +10,7 @@ What's new in ARMI
 ------------------
 #. The method ``Material.density3`` is now called ``density``, and the old ``density`` is now called ``pseudoDensity``. (`PR#1163 <https://github.com/terrapower/armi/pull/1163>`_)
 #. Added documentation for the thermal expansion approach used in ARMI. (`PR#1204 <https://github.com/terrapower/armi/pull/1204>`_)
+#. New option ``noOutput`` for globalFluxInterface to not copy output back to working directory. (`PR#1218 <https://github.com/terrapower/armi/pull/1218>`_)
```
mgjarrett, line 13:
```suggestion
#. New option ``copyOutput`` for globalFluxInterface to not copy output back to working directory. (`PR#1218 <https://github.com/terrapower/armi/pull/1218>`_)
```

armi/physics/executers.py
```diff
@@ -210,10 +210,21 @@ def run(self):
...
     def _collectInputsAndOutputs(self):
-        """Get total lists of input and output files."""
+        """
+        Get total lists of input and output files.
+
+        If self.options.copyOutput is false, don't copy the main `outputFile` back from the working directory.
+
+        In some ARMI runs, the executer can be run hundreds or thousands of times and generate many output files that
+        aren't strictly necessary to keep around. One can save space by choosing not to copy the outputs back in these
+        special cases. extraOutputFiles are typically controlled by the subclass, so
```
albeanth, line 220: Docstring looks incomplete?

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -339,6 +339,7 @@ def __init__(self, label: Optional[str] = None):
         # can happen in eig if Fredholm Alternative satisfied
         self.includeFixedSource = False
         self.eigenvalueProblem = True
+        self.copyOutput = True
```
albeanth, line 342: Should this be added to armi/physics/neutronics/settings.py? Otherwise to change the setting a user would have to go into the source code.

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -223,6 +224,22 @@ def _setTightCouplingTrue(self):
...
+    def test_collectInputsAndOutputs(self):
+        """
+        Verify that the executer can select to not copy back output.
+        """
+        executer = self.gfi.getExecuter()
+        cs = settings.Settings()
+        executer.options.fromUserSettings(cs)
+        executer.options.inputFile = "test.inp"
+        executer.options.outputFile = "test.out"
+        executer.options.copyOutput = False
```
albeanth, line 236: To keep coverage up, you'll need to add a unit test for `copyOutput = True`.

albeanth (changes requested):

armi/physics/executers.py
```diff
@@ -210,10 +210,21 @@ def run(self):
...
+        """
         inputs = [self.options.inputFile] if self.options.inputFile else []
         inputs.extend(self.options.extraInputFiles)
-        outputs = [self.options.outputFile] if self.options.outputFile else []
+        if self.options.outputFile and self.options.copyOutput:
+            outputs = [self.options.outputFile]
+        else:
+            outputs = []
         outputs.extend(self.options.extraOutputFiles)
```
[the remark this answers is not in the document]
mgjarrett: The `extraOutputFiles` could be a variety of things depending on the `Executer`. We don't have a ton of examples in our internal repos, but in the main example the `outputFile` is the main ASCII output and `extraOutputFiles` are auxiliary binary outputs. The `extraOutputFiles` don't have unique file names from case-to-case, so if hundreds or thousands of cases are run the file would be overwritten each time and you wouldn't end up with that many files. Furthermore, these binary outputs are sometimes needed for later analysis.

For the main example alluded to above, the `extraOutputFiles` are controlled by a separate setting. So the user has more control if we leave the `extraOutputFiles` independent of the `copyOutput` setting.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
mgjarrett: A user would be extending `globalFluxInterfaceUsingExecuters` for their use case, and they can set `copyOutput` when they configure the options for that specific type of executer. The user could add a case setting to control the `copyOutput` setting for their specific executer. That's what I've done with two of our internal repos. This provides finer grain control because you can have several different interfaces run executers during a run, and you might want some to copy their output back but not others. A general case setting that forces all interfaces to adopt the same `copyOutput` value would be less versatile.

albeanth: Ok got it. 

This is likely a separate issue, but are the options in the initializer/constructor of globalFluxInterface.py::GlobalFluxOptions documented anywhere?

albeanth: I think it would be a good idea. But we can do that in a separate issue and PR.

albeanth (approved):

armi/physics/neutronics/globalFlux/tests/test_globalFluxInterface.py
```diff
@@ -14,6 +14,7 @@
 """Tests for generic global flux interface"""
 # pylint: disable=missing-function-docstring,missing-class-docstring,protected-access,invalid-name,no-self-use,no-method-argument,import-outside-toplevel
 import unittest
+import os
```
mgjarrett, line 17:
```suggestion
```

PR #1219, opened 2023-03-21
Use TempraryDirectoryChanger for Executer.run()

mgjarrett:
## Description
---

ForcedCreationDirectoryChanger doesn't clean up after itself. For big runs, this can end up creating a huge amount of data in temporary folders. ARMI will clean up after the run completes, but it's possible that it could fill up the entire hard drive on some systems and cause the run to crash. TemporaryDirectoryChanger cleans up on `__exit__()`, which will prevent too much temporary directory memory using during a run.

## Checklist

### DIFF
--- a/armi/physics/executers.py
+++ b/armi/physics/executers.py
@@ -99,7 +99,7 @@ class ExecutionOptions:
 
         This is optional (you can set runDir to whatever you want). If you
         use this, you will get a relatively consistent naming convention
-        for your fast-past folders.
+        for your fast-path folders.
         """
         # This creates a hash of the case title plus the label
         # to shorten the running directory and to avoid path length
@@ -196,12 +196,13 @@ class DefaultExecuter(Executer):
         # or not list it in inputs (for optimization)
         self.writeInput()
         with directoryChangers.TemporaryDirectoryChanger(
-            self.options.runDir,
+            root=self.options.runDir,
             filesToMove=inputs,
             filesToRetrieve=outputs,
             outputPath=outputDir,
         ) as dc:
             self.options.workingDir = dc.initial
+            self.options.runDir = dc.destination
             self._execute()
             output = self._readOutput()
             if self.options.applyResultsToReactor:
--- a/armi/physics/executers.py
+++ b/armi/physics/executers.py
@@ -195,7 +195,7 @@ class DefaultExecuter(Executer):
         # must either write input to CWD for analysis and then copy to runDir
         # or not list it in inputs (for optimization)
         self.writeInput()
-        with directoryChangers.ForcedCreationDirectoryChanger(
+        with directoryChangers.TemporaryDirectoryChanger(
             self.options.runDir,
             filesToMove=inputs,
             filesToRetrieve=outputs,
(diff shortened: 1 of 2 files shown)


armi/physics/executers.py
```diff
@@ -195,13 +195,14 @@ def run(self):
...
-        with directoryChangers.ForcedCreationDirectoryChanger(
-            self.options.runDir,
+        with directoryChangers.TemporaryDirectoryChanger(
+            root=self.options.runDir,
             filesToMove=inputs,
             filesToRetrieve=outputs,
             outputPath=outputDir,
         ) as dc:
             self.options.workingDir = dc.initial
+            self.options.runDir = dc.destination
```
mgjarrett, line 205: @opotowsky sorry I marked this ready for review before it was working. I had to add this line to get it to work, but this feels sort of strange to me. What are your thoughts?

mgjarrett:
The alternative to this approach is to continue using the `ForcedCreationDirectoryChanger` and update its `__exit__` method to delete the directory in the same fashion that the `TemporaryDirectoryChanger` does. I initially backed off of this because I didn't want to impact other use cases of `ForcedCreationDirectoryChanger`. But I am concerned that the naming of the `runDir` will not be as consistent anymore when using the `TemporaryDirectoryChanger`. Right now; the naming of the `runDir` is based on a hash of the `caseTitle`, so when you run a case multiple times the names of the `runDir` will be consistent. Using a random  directory might be harder for users to keep track of when debugging.

armi/physics/executers.py
opotowsky: That's interesting this is needed for the temp changer but not the forced. Wouldn't you need to change the runDir back after this?

opotowsky: oh goodness I'm being dense, ignore that first sentence. this is a random dir created with the runDir as root. 

But still, would changing the runDir in this step have any downstream impacts for a case run?

opotowsky:
> The alternative to this approach is to continue using the `ForcedCreationDirectoryChanger` and update its `__exit__` method to delete the directory in the same fashion that the `TemporaryDirectoryChanger` does. I initially backed off of this because I didn't want to impact other use cases of `ForcedCreationDirectoryChanger`

It sounds like the tradeoff here is having a better named temp dir upon failures versus potentially impacting other users. If you think the better named dir is important we could definitely survey everyone to be sure no one is relying on a workflow that leaves files behind. I think since the temp dir changer has the runDir as root, the random name shouldn't cause too many issues

mgjarrett:
I agree that it's probably better to use the `TemporaryDirectoryChanger` here. I think this PR is ready to go, then.

armi/physics/executers.py
mgjarrett: > But still, would changing the runDir in this step have any downstream impacts for a case run?

No, changing the `runDir` shouldn't have any effects. We just need the `DirectoryChanger` and the `Executer` to agree about where the executable is going to be run.

opotowsky: @mgjarrett I just wanted to clear up the above question. I'm sure this is fine but I'm not visualizing where this happens in the run.

PR #1225, opened 2023-03-28
Bug fix for precision issue with axial submesh

albeanth:
## Description
A bug was identified in downstream analysis when using the uniform mesh converter. Specifically, the uniform mesh, computed here,
https://github.com/terrapower/armi/blob/14e879896a1fe1b22b9b73c89dd028eef697bad0/armi/reactor/converters/uniformMesh.py#L701
was found to have a repeated mesh point. This led to the uniform mesh converter trying to create a uniform mesh for a block with essentially a zero height - which results in a fatal error. 

This PR aims to resolve this bug by coarsening the tolerance set forth here,
https://github.com/terrapower/armi/blob/14e879896a1fe1b22b9b73c89dd028eef697bad0/armi/reactor/reactors.py#L1828-L1831

It was found that a rounding tolerance of `units.FLOAT_DIMENSION_DECIMALS` (10) was too restrictive based on the precision of the division done here,
https://github.com/terrapower/armi/blob/14e879896a1fe1b22b9b73c89dd028eef697bad0/armi/reactor/reactors.py#L1826

Given a a block height, when you have a "nice" number of subdivisions, a precision of 10 digits works. But when the number of subdivisions is not "nice", 10 digits of precision it too restrictive and the same axial mesh point is tacked on to the set, `collection`, resulting in a block height on the order of 1e-10 cm and the uniform mesh converter failing. 

I think there is reason to believe that a coarser tolerance could be used (e.g., 4 or 5, possibly even 0 digits for some cases), but to limit the potential impact of other downstream applications, 8 digits of precision was chosen as an adequate balance.

---

## Checklist

### DIFF
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -1802,6 +1802,11 @@ class Core(composites.Composite):
         from fuel performance, an imbalanced axial mesh may result.
 
         There is a challenge with TRZ blocks because we need the mesh centroid in terms of RZT, not XYZ
+
+        When determining the submesh, it is important to not use too small of a rounding precision. It was
+        found that when using a precision of units.FLOAT_DIMENSION_DECIMALS, that the division in `step`
+        can produce mesh points that are the same up to the 9th or 10th digit, resulting in a repeated
+        mesh point. This repetition results in problems in downstream methods, such as the uniform mesh converter.
         """
         runLog.debug("Finding all mesh points.")
         if assems is None:
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -1825,10 +1825,10 @@ class Core(composites.Composite):
                     axisVal = float(base[axis])  # convert from numpy.float64
                     step = float(top[axis] - axisVal) / subdivisions
                     for _subdivision in range(subdivisions):
-                        collection.add(round(axisVal, units.FLOAT_DIMENSION_DECIMALS))
+                        collection.add(round(axisVal, 8))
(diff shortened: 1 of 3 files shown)


mgjarrett (approved):
Great find @albeanth!

armi/reactor/reactors.py
```diff
@@ -1825,10 +1825,10 @@ def findAllMeshPoints(self, assems=None, applySubMesh=True):
                     axisVal = float(base[axis])  # convert from numpy.float64
                     step = float(top[axis] - axisVal) / subdivisions
                     for _subdivision in range(subdivisions):
-                        collection.add(round(axisVal, units.FLOAT_DIMENSION_DECIMALS))
+                        collection.add(round(axisVal, 8))
                         axisVal += step
                     # add top too (only needed for last point)
-                    collection.add(round(axisVal, units.FLOAT_DIMENSION_DECIMALS))
+                    collection.add(round(axisVal, 8))
```
keckler, line 1831: I think a comment explaining the problem that was uncovered here could be useful. Or else we are liable to make the same mistake again sometime in the future.

keckler (approved):

armi/reactor/reactors.py
albeanth: Addressed in https://github.com/terrapower/armi/pull/1225/commits/72139e657b7e1a02aa4a41e65b659ab47339f4c9

keckler (approved):

PR #1229, opened 2023-03-31
Control decusping

mgjarrett:
## Description

The performance of the `UniformMeshGeometryConverter` depends significantly on how good of a common uniform mesh can be generated. The uniform mesh converter will homogenize materials in neighboring blocks, which can be a problem at material boundaries between fuel or absorber and another material. One manifestation of this issue is a significant control rod "cusping" effect that has been observed when calculating differential rod worth.

The existing method `_computeAverageAxialMesh()` is too simplistic and is totally agnostic to material boundaries. This PR introduces a new class, `UniformMeshGenerator`, that has a set of methods to modify the mesh produced by `_computeAverageAxialMesh` so that it does a better job of representing important material boundaries. The main goal of this PR is to eliminate control rod cusping effects, but it will also help when modeling a core that has fuel assemblies at different heights.

## Checklist

### DIFF
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -665,6 +665,16 @@ class UniformMeshGeometryConverter(GeometryConverter):
         runLog.debug(f"Creating a uniform mesh of {newAssem}")
         bottom = 0.0
 
+        def checkPriorityFlags(b):
+            """
+            Check that a block has the flags that are prioritized for uniform mesh conversion
+
+            Also check that it's not different type of block that is a superset of the
+            priority flags, like "Flags.FUEL | Flags.PLENUM"
+            """
+            priorityFlags = [Flags.FUEL, Flags.CONTROL, Flags.SHIELD | Flags.RADIAL]
+            return b.hasFlags(priorityFlags) and not b.hasFlags(Flags.PLENUM)
+
         for topMeshPoint in newMesh:
             overlappingBlockInfo = sourceAssem.getBlocksBetweenElevations(
                 bottom, topMeshPoint
@@ -690,10 +700,9 @@ class UniformMeshGeometryConverter(GeometryConverter):
             #     use the xs type that represents the largest fraction of the destination block.
             typeHeight = collections.defaultdict(float)
             blocks = [b for b, _h in overlappingBlockInfo]
-            priorityFlags = [Flags.FUEL, Flags.CONTROL, Flags.SHIELD | Flags.RADIAL]
-            fuelOrAbsorber = any(b.hasFlags(priorityFlags) for b in blocks)
+            fuelOrAbsorber = any(checkPriorityFlags(b) for b in blocks)
             for b, h in overlappingBlockInfo:
-                if b.hasFlags(priorityFlags) or not fuelOrAbsorber:
+                if checkPriorityFlags(b) or not fuelOrAbsorber:
                     typeHeight[b.p.xsType] += h
 
             sourceBlock = None
@@ -702,7 +711,7 @@ class UniformMeshGeometryConverter(GeometryConverter):
                 k for k, v in typeHeight.items() if v == max(typeHeight.values())
             )
             for b in blocks:
-                if b.hasFlags(priorityFlags) or not fuelOrAbsorber:
+                if checkPriorityFlags(b) or not fuelOrAbsorber:
                     if b.p.xsType == xsType:
                         sourceBlock = b
                         break
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -88,12 +88,11 @@ def converterFactory(globalFluxOptions):
 
 class UniformMeshGenerator:
     """
-    The :py:class:`UniformMeshGenerator <armi.reactor.converters.UniformMeshGenerator>`
-    generates a common axial mesh to for the uniform mesh converter to use. The generation
-    algorithm starts with the simple ``average1DWithinTolerance`` utility function to compute
-    a representative "average" of the assembly meshes in the reactor. It then modifies that
-    mesh to more faithfully represent important material boundaries of fuel and control absorber
-    material.
+    This class generates a common axial mesh to for the uniform mesh converter to use. The
+    generation algorithm starts with the simple ``average1DWithinTolerance`` utility function
+    to compute a representative "average" of the assembly meshes in the reactor. It then modifies
+    that mesh to more faithfully represent important material boundaries of fuel and control
+    absorber material.
 
     The decusping feature is controlled with the case setting ``uniformMeshMinimumSize``. If no
     value is provided for this setting, the uniform mesh generator will skip the decusping step
@@ -141,6 +140,15 @@ class UniformMeshGenerator:
         this, the proposed uniform mesh will be some average of many assemblies in the core.
         The reason for this is to account for the fact that multiple assemblies (i.e., fuel assemblies)
         may have a different mesh due to differences in thermal and/or burn-up expansion.
+
+        Averaging all the assembly meshes that have the same number of points can be undesirable
+        in certain corner cases because no preference is assigned based on assembly type. For
+        example: if the reflector assemblies have the same number of mesh points as the fuel
+        assemblies but the size of the blocks is slightly different, the reflector mesh can influence
+        the uniform mesh and effectively pull it away from the fuel mesh boundaries, potentially
+        resulting in smearing (i.e., homogenization) of fuel with non-fuel materials. This is an
+        undesirable outcome. In the future, it may be advantageous to determine a better way of
+        sorting and prioritizing assembly meshes for generating the uniform mesh.
         """
         src = self._sourceReactor
         refAssem = src.core.refAssem
@@ -214,6 +222,7 @@ class UniformMeshGenerator:
             preference="top",
         )
         # combine all mesh points using all material boundaries as anchors with top preference
+        # top vs. bottom preference is somewhat arbitrary here
         combinedMesh = self._filterMesh(
             list(set(meshWithBottoms + meshWithTops)),
             self.minimumMeshSize,
@@ -249,6 +258,10 @@ class UniformMeshGenerator:
             meshList = sorted(list(set(meshList)))
         elif preference == "top":
             meshList = sorted(list(set(meshList)), reverse=True)
+        else:
+            raise ValueError(
+                f"Mesh filtering preference {preference} is not an option! Preference must be either bottom or top"
+            )
 
         while True:
             for i in range(len(meshList) - 1):
@@ -258,7 +271,7 @@ class UniformMeshGenerator:
                         errorMsg = (
                             "Attempting to remove two anchor points!\n"
                             "The uniform mesh minimum size for decusping is smaller than the "
-                            "gap between anchor points:\n"
+                            "gap between anchor points, which cannot be removed:\n"
                             f"{meshList[i]}, {meshList[i+1]}, gap = {abs(meshList[i]-meshList[i+1])}"
                         )
                         runLog.error(errorMsg)
@@ -356,46 +369,6 @@ class UniformMeshGenerator:
             topBoundList, self.minimumMeshSize, fuelTops, preference="top", warn=True
         )
         return filteredBottomCtrl, filteredTopCtrl
-
-    @staticmethod
-    def _determineIndexToRemove(ileft, iright, meshPoints, anchorPoints):
-        """
-        Determine which mesh to remove when two points are too close together
-
-        Rules:
-            1. Always keep anchorPoints
-            2. If neither mesh point is in anchorPointws, keep the "left" one
-
-        Notes
-        -----
-        ``UniformMeshGenerator._filterMesh`` sorts the mesh based on the option
-        passed for ``preference`` (``bottom`` or ``top``) so in this method,
-        ``ileft`` always has preference over ``iright``.
-
-        Parameters
-        ----------
-        ileft : int, required
-            "left" index (bottom if preference is "bottom", top if preference is "top"
-        iright : int, required
-            "right" index (opposite of the "left" index)
-        meshPoints : [float, float], required
-            The two mesh points within minimumMeshSize of each other
-        anchorPoints : list[float], required
-            The anchor points that will not be removed from the mesh
-        """
-        meshLeft, meshRight = tuple(meshPoints)
-        if meshRight in anchorPoints:
-            if meshLeft in anchorPoints:
-                runLog.error(
-                    "Attempting to remove two anchor points!\n"
-                    "The uniform mesh minimum size for decusping may be smaller than the "
-                    "gap between anchor points:\n"
-                    f"{meshLeft}, {meshRight}, gap = {abs(meshLeft-meshRight)}"
-                )
-                raise ValueError
-            return ileft
-        else:
-            return iright
 
 
 class UniformMeshGeometryConverter(GeometryConverter):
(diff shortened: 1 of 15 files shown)


armi/settings/fwSettings/globalSettings.py
```diff
@@ -158,6 +159,14 @@ def defineSettings() -> List[setting.Setting]:
             "block. Used for axial mesh refinement.",
             schema=vol.All(vol.Coerce(int), vol.Range(min=0, min_included=False)),
         ),
+        setting.Setting(
+            CONF_UNIFORM_MESH_MINIMUM_SIZE,
```
onufer, line 163: when it is left as None, i think it goes back to the old behavior of not preserving fuel boundaries as well, so it might be worthwhile to call out that setting any number here will do a better job of preserving material boundaries.

armi/reactor/converters/uniformMesh.py
```diff
@@ -806,6 +1106,28 @@ def _calculateReactionRates(lib, keff, assem):
                 continue
             globalFluxInterface.calcReactionRates(b, keff, lib)
 
+    def updateReactionRates(self):
```
onufer, line 1109: this is not called or tested anywhere. i think i might understand what its used for, but can you add a little more clarification? something like:

useful for testing xyz to compare abc on a consistent mesh...

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,317 @@ def converterFactory(globalFluxOptions):
...
+    a representative "average" of the assembly meshes in the reactor. It then modifies that
+    mesh to more faithfully represent important material boundaries of fuel and control absorber
+    material.
+
+    The decusping feature is controlled with the case setting ``uniformMeshMinimumSize``. If no
+    value is provided for this setting, the uniform mesh generator will skip the decusping step
+    and just provide the result of ``_computeAverageAxialMesh``.
+    """
+
+    FLOAT_ROUNDING_DECIMALS = 12
```
onufer, line 103: maybe bring this down to 10. i am nervous that since it is not configurable, if you had bigger numbers  (>10kl) it would error out. either that or add comment for basis of 12 places

armi/reactor/converters/uniformMesh.py
```diff
@@ -383,21 +714,22 @@ def makeAssemWithUniformMesh(
             #     3. Use the first block of the majority XS type as the representative block.
             typeHeight = collections.defaultdict(float)
             blocks = [b for b, _h in overlappingBlockInfo]
+            priorityFlags = [Flags.FUEL, Flags.CONTROL, Flags.SHIELD | Flags.RADIAL]
+            fuelOrAbsorber = any(b.hasFlags(priorityFlags) for b in blocks)
             for b, h in overlappingBlockInfo:
-                if b.hasFlags([Flags.FUEL, Flags.CONTROL]):
+                if b.hasFlags(priorityFlags) or not fuelOrAbsorber:
```
onufer, line 720: the `or not fuelOrAbsorber` is confusion to me. can you add a comment explaining what this is trying to do.

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,317 @@ def converterFactory(globalFluxOptions):
...
+        This iterates over all the assemblies in the core and collects all assembly meshes
+        that have the same number of fine-mesh points as the `refAssem` for the core. Based on
+        this, the proposed uniform mesh will be some average of many assemblies in the core.
+        The reason for this is to account for the fact that multiple assemblies (i.e., fuel assemblies)
+        may have a different mesh due to differences in thermal and/or burn-up expansion.
+        """
+        src = self._sourceReactor
+        refAssem = src.core.refAssem
+
+        refNumPoints = len(src.core.findAllAxialMeshPoints([refAssem])) - 1
```
onufer, line 150: add a comment above explaning the -1 is for non-zero points or calculate it the same way (with `[:1]`)

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,317 @@ def converterFactory(globalFluxOptions):
...
+        allMeshes = []
+        for a in src.core:
+            # Get the mesh points of the assembly, neglecting the first coordinate
+            # (typically zero).
+            aMesh = src.core.findAllAxialMeshPoints([a])[1:]
+            if len(aMesh) == refNumPoints:
```
onufer, line 156: add a comment calling out that this is a pretty bad check since its intending to check that they have simlar mesh but is only checking that the len() is equal. it could fail pretty spectacularly in corner cases where len is equal but mesh is much different. i know you inhered the code though, so we dont need to fix it.

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,317 @@ def converterFactory(globalFluxOptions):
...
+            preference="bottom",
+        )
+        # combine material top boundaries with full mesh using top preference
+        meshWithTops = self._filterMesh(
+            list(self._commonMesh) + materialTops,
+            self.minimumMeshSize,
+            materialTops,
+            preference="top",
+        )
+        # combine all mesh points using all material boundaries as anchors with top preference
```
onufer, line 220: it seems like this step top vs bottom is somewhat arbitrary. is that true? if so say that its arbitrary. if not explain why its not? maybe that you know 0.0 is already in there so you want 0.0 + tops

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,317 @@ def converterFactory(globalFluxOptions):
...
+        warn : bool, optional
+            Whether to log a warning when a mesh is removed. This is true if a
+            control material boundary is removed, but otherwise it is false.
+        """
+        if preference == "bottom":
+            meshList = sorted(list(set(meshList)))
+        elif preference == "top":
+            meshList = sorted(list(set(meshList)), reverse=True)
+
+        keepChecking = True
```
onufer, line 257: could delete this and make it `while True`

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,317 @@ def converterFactory(globalFluxOptions):
...
+            preference="bottom",
+            warn=True,
+        )
+        filteredTopCtrl = self._filterMesh(
+            topBoundList, self.minimumMeshSize, fuelTops, preference="top", warn=True
+        )
+        return filteredBottomCtrl, filteredTopCtrl
+
+    @staticmethod
+    def _determineIndexToRemove(ileft, iright, meshPoints, anchorPoints):
```
onufer, line 360: I think I would prefer this logic back in the original function...
instead of 
```
                    removeIndex = self._determineIndexToRemove(
                        i, i + 1, meshList[i : i + 2], anchorPoints
                    )
```


It would be 
```
                    if meshList[i] in anchorPoints and meshList[i+1] in anchorPoints:
                        raise ValueError
                    if meshList[i] in anchorPoints:
                        removeIndex = i
                    if meshList[i+1] in anchorPoints:
                        removeIndex = i +1
```
would would seem more clear to me. The logic here does not seem complicated enough that hit needs a separate static method.

armi/reactor/converters/uniformMesh.py
mgjarrett: I was initially rounding to 5 decimal places because I thought anything less than that is just insignificant for meshing, and that would reduce the amount of redundant mesh points that need to be filtered out.

But I ran into a problem because the `b.p.zbottom` and `b.p.ztop` are not rounded. So if, for example, `fuelBlock.p.ztop = 100.1234567`, you'd get a mesh point at `100.12345`, and then 0.0000067 cm of that fuel would get blended into the next block above it. I went to 12 to avoid this rounding problem. But rounding to 12 probably isn't serving any purpose. I will test without the rounding and see if the functionality is the same.

armi/reactor/converters/uniformMesh.py
mgjarrett: This is not directly related to this PR, but it's to address an issue that I encountered while testing this PR. When determining which `xsType` to use for the destination block in a uniform mesh conversion, it would always take the `b.p.xsType` that accounts for the majority of overlap with the destination block that is associated with a `Flags.FUEL` block or `Flags.CONTROL` block. If there are no fuel or control flags, it would take the first block. This is updating the converter to also take the `b.p.xsType` that covers the largest fraction of the destination block even if there are no `Flags.FUEL` or `Flags.CONTROL` blocks.

I also added `Flags.SHIELD | Flags.RADIAL` to the "special" block types to preserve `b.p.xsType`.

armi/reactor/converters/uniformMesh.py
mgjarrett: I think it intends to do exactly what it does, which is check whether the two meshes have the same number of points. If the mesh does have the correct number of points, it is added to the list to be averaged together with the other meshes.

The corner case where the mesh is much different is handled by `average1DWithinTolerance`. The tolerance is not a user-configurable setting, but it defaults to 20%. If the mesh size for any block in a given assembly mesh differs from the overall average by more than 20%, that assembly mesh will be removed and the overall average will be recalculated.

armi/reactor/converters/uniformMesh.py
mgjarrett: It is arbitrary. I've gone back and forth on bottom vs. top preference.

The fuel and control material boundaries are all `anchorPoints` by now, so this is more about adjusting the mesh points away from boundaries if they are smaller than the `minimumMeshSize`. In most cases, I would expect that this filtering step doesn't actually filter out any mesh points, but there could be a situation where it becomes necessary. It's hard to say whether `bottom` or `top` is better in general here; it would depend on the specific circumstances of a given case.

armi/reactor/converters/uniformMesh.py
onufer: "If the mesh does have the correct number of points, it is added to the list to be averaged together with the other meshes." 

I think the subtext of this is that its testing "are these meshes similar enough to reasonably average. Length is not sufficient for this. For example if you had reflector meshes within 5%, you wouldn't want tat mesh to drag the uniform mesh, but it would with the current code.

I think there is room for improvement here, eg checking assembly type or something a little more rigorous than equal length. 

not part of this PR though.

armi/reactor/converters/uniformMesh.py
onufer: sorry, i meant to say add a comment in the code that it is arbitrary if its arbitrary. helps next person to figure out if they are breaking past intent

armi/reactor/converters/uniformMesh.py
onufer: sorry, I was asking you to add a comment in the code describing what this is accomplishing/why its needed.

or update the doc string to capture the new intended behavior.

onufer: maybe update the above comments to reflect the current state. EG add radial | shield to 2

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,321 @@ def converterFactory(globalFluxOptions):
...
+        modifies the uniform mesh to honor fuel and control rod material boundaries while avoiding excessively
+        small mesh sizes.
+
+        If adding control rod material boundaries to the mesh creates excessively small mesh regions,
+        this function will move internal fuel region boundaries to make room for the control rod boundaries.
+
+        This function operates by filtering out mesh points that are too close together while always holding on
+        to the specified "anchor" points in the mesh. The anchor points are built up progressively as the
+        appropriate bottom and top boundaries of fuel and control assemblies are determined.
+        """
```
onufer, line 176: I really like how clear this code is an how you were able to re-use filter!

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,321 @@ def converterFactory(globalFluxOptions):
...
+            When neither mesh point is in the list of ``anchorPoints``, which mesh point is given preference
+            ("bottom" or "top")
+        warn : bool, optional
+            Whether to log a warning when a mesh is removed. This is true if a
+            control material boundary is removed, but otherwise it is false.
+        """
+        if preference == "bottom":
+            meshList = sorted(list(set(meshList)))
+        elif preference == "top":
+            meshList = sorted(list(set(meshList)), reverse=True)
```
onufer, line 251: oh, we probably want it to fail if its non bottom or top.

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,321 @@ def converterFactory(globalFluxOptions):
...
+
+        while True:
+            for i in range(len(meshList) - 1):
+                difference = abs(meshList[i + 1] - meshList[i])
+                if difference < minimumMeshSize:
+                    removeIndex = self._determineIndexToRemove(
+                        i, i + 1, meshList[i : i + 2], anchorPoints
+                    )
```
onufer, line 259: delete this now

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,321 @@ def converterFactory(globalFluxOptions):
...
+            preference="bottom",
+            warn=True,
+        )
+        filteredTopCtrl = self._filterMesh(
+            topBoundList, self.minimumMeshSize, fuelTops, preference="top", warn=True
+        )
+        return filteredBottomCtrl, filteredTopCtrl
+
+    @staticmethod
+    def _determineIndexToRemove(ileft, iright, meshPoints, anchorPoints):
```
onufer, line 364: this can now be deleted.

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,321 @@ def converterFactory(globalFluxOptions):
         )
 
 
+class UniformMeshGenerator:
+    """
+    The :py:class:`UniformMeshGenerator <armi.reactor.converters.UniformMeshGenerator>`
```
john-science, line 91: I don't think we need to say the name of the class in the class docstring.  Redundant.

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,321 @@ def converterFactory(globalFluxOptions):
...
+                    )
+                    if meshList[i] in anchorPoints and meshList[i + 1] in anchorPoints:
+                        errorMsg = (
+                            "Attempting to remove two anchor points!\n"
+                            "The uniform mesh minimum size for decusping is smaller than the "
+                            "gap between anchor points:\n"
+                            f"{meshList[i]}, {meshList[i+1]}, gap = {abs(meshList[i]-meshList[i+1])}"
+                        )
+                        runLog.error(errorMsg)
+                        raise ValueError(errorMsg)
```
john-science, line 268: As is common around these parts, you added this exception, but it is not covered by unit tests.

armi/reactor/converters/uniformMesh.py
```diff
@@ -85,6 +86,321 @@ def converterFactory(globalFluxOptions):
...
+        meshLeft, meshRight = tuple(meshPoints)
+        if meshRight in anchorPoints:
+            if meshLeft in anchorPoints:
+                runLog.error(
+                    "Attempting to remove two anchor points!\n"
+                    "The uniform mesh minimum size for decusping may be smaller than the "
+                    "gap between anchor points:\n"
+                    f"{meshLeft}, {meshRight}, gap = {abs(meshLeft-meshRight)}"
+                )
+                raise ValueError
```
john-science, line 398: As is common around these parts, you added this exception, but it is not covered by unit tests.

armi/reactor/converters/uniformMesh.py
mgjarrett: I updated the comment.

armi/reactor/converters/uniformMesh.py
mgjarrett: I added a unit test to cover this error

onufer (approved):

onufer:
This looks great @mgjarrett! thanks!

@john-science take a look at the internal results shared on our reactivity control plugin. its hard to evaluate performance of de-cusping without an actual flux solver and we decided it would be better to share that evaluation internally.

mgjarrett:
> @mgjarrett You mention performance. This certainly adds a lot of code, do we know how this affects our performance?

The goal here was to improve performance with respect to accuracy. I don't think this will significantly impact the runtime performance.

I have not profiled it at all, but I just ran an example case to measure timing and found that it consistently takes about 0.6-0.7 s for the `UniformMeshGenerator` class to run `_generateUniformMesh()` on a Natrium model. The entire `UniformMeshGenerator` typically takes tens of seconds at least, so this is a minor increase in run time overall.

PR #1230, opened 2023-03-31
Cleanup of determineTargetComponent

### LINKED ISSUE #1197: Single component block axial expansion
Related to https://github.com/terrapower/armi/issues/1193

The axial expander should be able to recognize when there is one solid component in a block without extra blueprint flagging to make it recognize it's target.

albeanth:
## Description
This PR brings in various improvements to axialExpansionChanger.py::ExpansionData::determineTargetComponent. 
1. ~if Flags do not produce a target component, the component and block types are now checked. This is useful for the case where you have a target component that does not have a flag defined.~ 
   Rolled this back and users should use the block attribute `axial expansion target component` introduced in #777 .
3. Closes #1197.
4. Overhauled and improved the unit testing for the method.

---

## Checklist

### DIFF
--- a/armi/reactor/converters/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger.py
@@ -850,11 +850,6 @@ class ExpansionData:
         else:
             componentWFlag = [c for c in b.getChildren() if c.hasFlags(flagOfInterest)]
         if len(componentWFlag) == 0:
-            # didn't match flags so check type before failing
-            componentWFlag = [
-                c for c in b if (c.p.type in b.p.type or b.p.type in c.p.type)
-            ]
-        if len(componentWFlag) == 0:
             # if only 1 solid, be smart enought to snag it
             solidMaterials = list(
                 c for c in b if not isinstance(c.material, material.Fluid)
--- a/armi/reactor/converters/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger.py
@@ -851,7 +851,9 @@ class ExpansionData:
             componentWFlag = [c for c in b.getChildren() if c.hasFlags(flagOfInterest)]
         if len(componentWFlag) == 0:
             # didn't match flags so check type before failing
-            componentWFlag = [c for c in b if c.p.type == b.p.type]
+            componentWFlag = [
+                c for c in b if (c.p.type in b.p.type or b.p.type in c.p.type)
+            ]
         if len(componentWFlag) == 0:
             # if only 1 solid, be smart enought to snag it
             solidMaterials = list(
--- a/armi/reactor/converters/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger.py
@@ -854,10 +854,9 @@ class ExpansionData:
             componentWFlag = [c for c in b if c.p.type == b.p.type]
         if len(componentWFlag) == 0:
             # if only 1 solid, be smart enought to snag it
-            solidMaterials = []
-            for c in b:
-                if not isinstance(c.material, material.Fluid):
-                    solidMaterials.append(c)
+            solidMaterials = list(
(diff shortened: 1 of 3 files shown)


armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -850,13 +850,23 @@ def determineTargetComponent(self, b, flagOfInterest=None):
         else:
             componentWFlag = [c for c in b.getChildren() if c.hasFlags(flagOfInterest)]
         if len(componentWFlag) == 0:
-            raise RuntimeError("No target component found!\n   Block {0}".format(b))
+            # didn't match flags so check type before failing
+            componentWFlag = [
+                c for c in b if (c.p.type in b.p.type or b.p.type in c.p.type)
+            ]
```
keckler, line 856: I'm not sure I understand what the underlying assumption is here that makes this a valid thing to do. Basically you're saying that if a component in the block has the same name as the block itself then it should be the target component?

jakehader: It might be good to consider components by name since there is no guarantee that this will work for all block types and might be fragile. Is it worth considering a refactor for always using a component name or first flags and then name?

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -722,103 +688,148 @@ class TestDetermineTargetComponent(AxialExpansionTestBase, unittest.TestCase):
...
-        # check for no target component found
+    def test_specifyTargetComponent_viaType(self):
+        """set target component via type and not flags
+
+        Notes
+        -----
+        For the case in which a block contains a component that does not have Flags defined.
+        So instead of matching on flags, we match on type.
+        """
+        b = HexBlock("detector", height=10.0)
```
albeanth, line 770: @keckler @jakehader Here is the unit test that leverages the type checking. 

This block, "detector", contains a "detector" component. Since "detector" is not a registered flag the flag-based approach won't work. So our options are to either 1) modify source code to add a new flag, or 2) add a new capability to infer target components based on something other than flags. I got feedback that option 2 was more desirable and that checking types could be an option. 

The logic for `c.p.type in b.p.type or b.p.type in c.p.type` is to try and accommodate the case where the strings that define the component and block types aren't exactly the same. In this test, the component type is `detector neutron` and the block type is `neutron`. In this case the component type isn't "in" the block type, but the block type is "in" the component type. If the types were reversed, the opposite would be true (i.e., the component type would be "in" the block type).

Hopefully that's not as clear as mud...

keckler: Isn't there a third option? To specify the target component on the blueprints using `axial expansion target component`? I'm just kinda confused why this change is necessary in the first place, but maybe I'm missing something.

albeanth: Yeah, that is a third, and very valid, option.

This PR is in response to a series of requests to not have to use that blueprints attribute.

armi/reactor/converters/axialExpansionChanger.py
albeanth: After some offline discussions, this type check was rolled back in https://github.com/terrapower/armi/pull/1230/commits/d9d138b9b40264ba9e82639e810235d67f12175d.

armi/reactor/converters/tests/test_axialExpansionChanger.py
albeanth: After some offline discussions, this type check was rolled back in https://github.com/terrapower/armi/commit/d9d138b9b40264ba9e82639e810235d67f12175d.

keckler (approved):

albeanth:
This PR is ready, but I am converting to draft while we figure out downstream merge dependencies.

PR #1231, opened 2023-04-04
Remove outdated equilibrium setting validator

keckler:
## Description

There was a setting validator that's purpose was to check if a user had specified the cycle history using the [detailed cycles history](https://terrapower.github.io/armi/user/inputs/settings.html#cycle-history) while performing an equilibrium calculation. This need derived from one of our internal plugins, which lead to the incompatibility.

That internal plugin has now been updated, and so this restriction no longer needs to be enforced.

Side note: this setting validator should have resided in our internal plugin in the first place, I'm not sure why I put it in the framework back during #637 .

---

## Checklist

### DIFF
--- a/armi/operators/settingsValidation.py
+++ b/armi/operators/settingsValidation.py
@@ -644,16 +644,6 @@ class Inspector:
         )
 
         self.addQuery(
-            lambda: self.cs["runType"] == operators.RunTypes.EQUILIBRIUM
-            and self.cs["cycles"],
-            "Equilibrium cases cannot use the `cycles` case setting to define detailed"
-            " cycle information. Try instead using the simple cycle history inputs"
-            " `cycleLength(s)`, `burnSteps`, `availabilityFactor(s)`, and/or `powerFractions`",
-            "",
-            self.NO_ACTION,
-        )
-
-        self.addQuery(
             lambda: self.cs["skipCycles"] > 0
             and not os.path.exists(self.cs.caseTitle + ".restart.dat"),
             "This is a restart case, but the required restart file {0}.restart.dat is not found".format(


doc/release/0.2.rst
```diff
@@ -14,6 +14,7 @@ What's new in ARMI
 #. Use TemporaryDirectoryChanger for executer.run() so dirs are cleaned up during run. (`PR#1219 <https://github.com/terrapower/armi/pull/1219>`_)
 #. New option ``copyOutput`` for globalFluxInterface to not copy output back to working directory. (`PR#1218 <https://github.com/terrapower/armi/pull/1218>`_, `PR#1227 <https://github.com/terrapower/armi/pull/1227>`_)
 #. `Executer` class has a ``dcType`` attribute to define the type of ``DirectoryChanger`` it will use. (`PR#1228 <https://github.com/terrapower/armi/pull/1228>`_)
+#. Removed restriction on using detailed cycle histories in equilibrium mode (`PR#1231 <https://github.com/terrapower/armi/pull/1231>`_)
```
onufer, line 17: seems like this is pretty in the weeds and not worth documenting here. there is still a restriction that it will only use the first index...

onufer (approved):

doc/release/0.2.rst
keckler: That restriction is in our internal plugin, and it is now documented and checked by the internal plugin.

john-science: Yeah, the commit history exists for detail. But the release notes are supposed to be high-level changes that are easy for casual readers to understand the major changes in the release.

That has not been enforced well in the past, but that has always been the idea.  I'm trying to do better at cleaning up the notes before a release.

keckler: Fine, it's been removed.

PR #1235, opened 2023-04-05
Add setting for some assemblies to skip axial expansion

albeanth:
## Description
It may be desired to have some assemblies skip axial expansion routines. This PR adds a setting, `assemFlagsToSkipAxialExpansion` to accomplish this. 

Users can provide assembly flags in a list in the input setting. If an assembly contains a flag that matches those listed, it is skipped for axial expansion.

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/__init__.py
+++ b/armi/reactor/blueprints/__init__.py
@@ -92,9 +92,15 @@ from armi.reactor import systemLayoutInput
 from armi.scripts import migration
 
 from armi.utils import textProcessors
-from armi.physics.neutronics.fissionProductModel import lumpedFissionProduct
+from armi.settings.fwSettings.globalSettings import (
+    CONF_DETAILED_AXIAL_EXPANSION,
+    CONF_ASSEM_FLAGS_SKIP_AXIAL_EXP,
+    CONF_INPUT_HEIGHTS_HOT,
+    CONF_NON_UNIFORM_ASSEM_FLAGS,
+    CONF_ACCEPTABLE_BLOCK_AREA_ERROR,
+    CONF_GEOM_FILE,
+)
 from armi.physics.neutronics.settings import CONF_LOADING_FILE
-from armi.nucDirectory import elements
 
 # NOTE: using non-ARMI-standard imports because these are all a part of this package,
 # and using the module imports would make the attribute definitions extremely long
@@ -329,21 +335,23 @@ class Blueprints(yamlize.Object, metaclass=_BlueprintsPluginCollector):
             runLog.header("=========== Verifying Assembly Configurations ===========")
             self._checkAssemblyAreaConsistency(cs)
 
-            if not cs["detailedAxialExpansion"]:
+            if not cs[CONF_DETAILED_AXIAL_EXPANSION]:
                 # this is required to set up assemblies so they know how to snap
                 # to the reference mesh. They wont know the mesh to conform to
                 # otherwise....
                 axialExpansionChanger.makeAssemsAbleToSnapToUniformMesh(
-                    self.assemblies.values(), cs["nonUniformAssemFlags"]
+                    self.assemblies.values(), cs[CONF_NON_UNIFORM_ASSEM_FLAGS]
                 )
-            if not cs["inputHeightsConsideredHot"]:
+            if not cs[CONF_INPUT_HEIGHTS_HOT]:
                 runLog.header(
                     "=========== Axially expanding all assemblies (except control) from Tinput to Thot ==========="
                 )
                 # expand axial heights from cold to hot so dims and masses are consistent
                 # with specified component hot temperatures.
                 axialExpansionChanger.expandColdDimsToHot(
-                    self.assemblies.values(), cs["detailedAxialExpansion"]
+                    self.assemblies.values(),
+                    cs[CONF_DETAILED_AXIAL_EXPANSION],
+                    cs[CONF_ASSEM_FLAGS_SKIP_AXIAL_EXP],
                 )
 
             # pylint: disable=no-member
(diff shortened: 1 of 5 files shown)


albeanth:
@onufer FYI

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -64,26 +70,25 @@ def expandColdDimsToHot(assems, isDetailedAxialExpansion, referenceAssembly=None
     assems: list[:py:class:`Assembly <armi.reactor.assemblies.Assembly>`]
         list of assemblies to be thermally expanded
     isDetailedAxialExpansion: bool
-        If true assemblies will be forced to conform to the reference mesh after expansion
+        If False, assemblies will be forced to conform to the reference mesh after expansion
+    assemsToSkip: List[str]
+        list of strings to be converted to flags to indicate which assemblies to skip axial expansion
```
keckler, line 75: It would seem better to me to just pass the flags directly. That way someone who has the flags doesn't need to de-convert them into strings in order to use this function.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -64,26 +70,25 @@ def expandColdDimsToHot(assems, isDetailedAxialExpansion, referenceAssembly=None
...
-            # this doesn't get applied to control assems, so CR will be interpreted
-            # as hot. This should be conservative because the control rods will
-            # be modeled as slightly shorter with the correct hot density. Density
-            # is more important than height, so we are forcing density to be correct
-            # since we can't do axial expansion (yet)
-            axialExpChanger.applyColdHeightMassIncrease()
-            axialExpChanger.expansionData.computeThermalExpansionFactors()
-            axialExpChanger.axiallyExpandAssembly()
+        if a.hasFlags(Flags.CONTROL) or any(a.hasFlags(aFlags) for aFlags in aToSkip):
+            continue
```
keckler, line 87: ?

https://github.com/terrapower/armi/pull/1226

armi/reactor/blueprints/__init__.py
```diff
@@ -329,21 +335,23 @@ def _prepConstruction(self, cs):
...
                 runLog.header(
                     "=========== Axially expanding all assemblies (except control) from Tinput to Thot ==========="
                 )
                 # expand axial heights from cold to hot so dims and masses are consistent
                 # with specified component hot temperatures.
                 axialExpansionChanger.expandColdDimsToHot(
-                    self.assemblies.values(), cs["detailedAxialExpansion"]
+                    self.assemblies.values(),
+                    cs[CONF_DETAILED_AXIAL_EXPANSION],
+                    cs[CONF_ASSEM_FLAGS_SKIP_AXIAL_EXP],
```
keckler, line 354: This is not incredibly important, but...

With this change, you change the API of `expandCoddDimsToHot()`. And in fact you make the new parameter mandatory, so the change is also not backwards compatible.

My inclination would be to avoid changing the signature of `expandColdDimsToHot()`, and instead to just whittle down the list returned from `self.assemblies.values()` right here when you pass it in. Something like the following:

```python
assemsToExpand = [a for a in list(self.assemblies.values()) if not a.hasFlags(cs[CONF_ASSEM_FLAGS_SKIP_AXIAL_EXP])
axialExpansionChanger.expandColdDimsToHot(assemsToExpand, cs[CONF_DETAILED_AXIAL_EXPANSION])
```

Just a suggestion.

keckler (commented):
A few comments, one of which (with the skipping of the control assems) I would like to get answered before approval.

armi/reactor/converters/axialExpansionChanger.py
albeanth: Once #1226 is merged in, the check on `a.hasFlags(Flags.CONTROL)` will be removed. I left it in there to keep the PRs independent of one another.

keckler: Just take that part out now and put a comment on this PR saying not to merge until #1226 is merged?

keckler: Like, don't delete all those lines either. Github will handle the diffs properly. Then this PR will actually be independent of that one.

albeanth: I'll just pop off the check on `a.hasFlags(Flags.CONTROL)`.

keckler: I'm an idiot, ignore this comment thread!

armi/reactor/blueprints/__init__.py
albeanth: Addressed. https://github.com/terrapower/armi/pull/1235/commits/32ce614201e443c1a4e6db86c634f16c0f8856ad

armi/reactor/converters/axialExpansionChanger.py
albeanth: Addressed https://github.com/terrapower/armi/pull/1235/commits/32ce614201e443c1a4e6db86c634f16c0f8856ad

keckler (approved):

armi/reactor/converters/axialExpansionChanger.py
crswong888: Assuming it's okay to require exact type IDs here, e.g., `"duct"` wouldn't match `"innerduct"`, you could just do the following:

```suggestion
        if a.hasFlags(Flags.CONTROL) or a.hasFlags(aToSkip, exact=True):
            continue
```

albeanth: My concern with using exact if you have the following:

```python
import armi
armi.configure()

b0 = HexBlock("fuel")
b0.setType("fuel")
b1 = HexBlock("outer fuel")
b1.setType("outer fuel")

# Part 1: Use exact to get false negative
print(b1.hasFlags(typeID=[b0.p.flags]))
print(b1.hasFlags(typeID=[b0.p.flags], exact = True))

# Part 2: Gives false negative always
print(b0.hasFlags(typeID=[b1.p.flags]))
print(b0.hasFlags(typeID=[b1.p.flags], exact = True))

# Part 3: The correction for part 2
print(any(b0.hasFlags(f) for f in b1.p.flags))
print(any(b0.hasFlags(f, exact=True) for f in b1.p.flags))
```
which prints
```
True
False
False
False
True
True
```
So using the `any` approach is a bit more robust, I think. It's also used in Core::processLoading here, https://github.com/terrapower/armi/blob/c8a40c76d9da31e3f8322a51eb52d3864a3f68a6/armi/reactor/reactors.py#L2272-L2276

john-science:
@albeanth This is sitting in a queue for a downstream project. Merging one day soon, hopefully.

Issue #1212, opened 2023-03-10
stationary block misalignment with discharge swap can cause problems. 


onufer:
if your stationary blocks don't line up, after the swap the blocks in an assembly may not line up. this can cause problems. we should check that assembly block boundaries are maintained before and after shuffling.

onufer:
this would probably be a warning and not a full error.

onufer:
for example, grid blocks are typically not shuffled, so if the ztop of the block that was stationary is different from the zBot of the bottom block of the assembly being shuffled in. this inconsistency will cause a failure in the next interface

PR #1236, opened 2023-04-06
Implement warning for misaligned stationary blocks boundary

HunterPSmith:
Implement a warning callout for the case where the boundary of the stationary blocks between two assemblies have different heights. This PR seeks to resolve: https://github.com/terrapower/armi/issues/1212.

## Description

---

## Checklist

- [ ] ~The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any bug fixes or new features.~ Not needed, just adding a warning. No bugs or new features (TA).

### DIFF
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -775,7 +775,7 @@ class FuelHandler:
                     a2StationaryBlocks[-1][0].p.ztop,
                 )
             )
-            
+
         # swap stationary blocks
         for (assem1Block, assem1BlockIndex), (assem2Block, assem2BlockIndex) in zip(
             a1StationaryBlocks, a2StationaryBlocks
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -776,7 +776,6 @@ class FuelHandler:
                 )
             )
             
-
         # swap stationary blocks
         for (assem1Block, assem1BlockIndex), (assem2Block, assem2BlockIndex) in zip(
             a1StationaryBlocks, a2StationaryBlocks
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -775,7 +775,7 @@ class FuelHandler:
                     a2StationaryBlocks[-1][0].p.ztop,
                 )
             )
-            return
+            
 
         # swap stationary blocks
         for (assem1Block, assem1BlockIndex), (assem2Block, assem2BlockIndex) in zip(
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -762,6 +762,20 @@ class FuelHandler:
                     assembly1, a1StationaryBlocks, assembly2, a2StationaryBlocks
                 )
             )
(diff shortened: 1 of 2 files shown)


armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -762,6 +762,20 @@ def _transferStationaryBlocks(self, assembly1, assembly2):
...
+                 and {} (Stationary Blocks: {}, Elevation at top of stationary blocks {}))""".format(
+                    assembly1,
+                    a1StationaryBlocks,
+                    a1StationaryBlocks[-1][0].p.ztop,
+                    assembly2,
+                    a2StationaryBlocks,
+                    a2StationaryBlocks[-1][0].p.ztop,
+                )
+            )
+            return
```
onufer, line 778: probably delete the return statement, and write a test to cover this method if we dont have it.

onufer (changes requested):

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: Removed!

onufer (approved):

albeanth (changes requested):
@HunterPSmith looks like your code results in a failure of one of the unit tests. Before we merge, that will need to be resolved. 

Also, it would be great to add a unit test providing coverage for this new check. It will also help maintain and/or improve coverage for fuelHandlers.py. You can test to ensure that the `runLog.warning` hits by using `mockRunLogs`. [Here's an example](https://github.com/terrapower/armi/blob/c8a40c76d9da31e3f8322a51eb52d3864a3f68a6/armi/reactor/converters/tests/test_axialExpansionChanger.py#L818-L821) that uses it in a unit test. I'm happy to help you configure a test if needed. Thanks!

HunterPSmith:
@albeanth Thank you for the tip on the testing! I think everything should be good now!

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -594,6 +596,24 @@ def test_transferIncompatibleStationaryBlocks(self):
         with self.assertRaises(ValueError):
             fh._transferStationaryBlocks(a1, a2)
 
+        # re-initialize assemblies
+        self.setUp()
```
albeanth, line 600: Calling `self.setUp` suggests that the assertions below would be very well suited for a new test! I noticed that `self.setUp()` is called a few times in this test file. It would be a welcome addition if you could split off each of those into individual tests! 

If it ends up being more work/a pain to do, we can create an issue and punt. But we should definitely split off your newest addition into its own test.

albeanth (changes requested):

armi/physics/fuelCycle/tests/test_fuelHandlers.py
HunterPSmith: I split up the test into three separate tests!

albeanth: Looks like your unit tests are failing as the new tests that got split off don't have a `fuelHandler` object. I think if you add `fh = fuelHandlers.FuelHandler(self.o)` to the tests, they should fall into place.

HunterPSmith: Fixed!

albeanth (approved):
Nicely done. Thanks @HunterPSmith !

PR #1173, opened 2023-02-10
allow user to specify cycles to skip convergence

onufer:
## Description

allow user to specify cycles to skip convergence

---

## Checklist

- [x] This PR has only one purpose or idea.
- [x] Tests have been added/updated to verify that the new/changed code works.

- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html).

- [x] The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any bug fixes or new features.
- [x] The documentation is still up-to-date in the `doc` folder.
- [x] The dependencies are still up-to-date in `setup.py`.

PR #1239, opened 2023-04-11
New Option to Control Lattice Physics Update Frequency

onufer:
## Description
---

The lattice physics interface was hard-coded to update at BOC for a long time. With the recent addition of `tightCoupling`, we added some new logic to perform lattice physics updates in specific scenarios (on `Snapshots` runs, or on the first coupled iteration). Having the logic for when lattice physics is updated hard-coded for specific situations is undesirable.

This PR introduces a new setting, `latticePhysicsFrequency`, that controls when the lattice physics calculations are re-run. The options are:

0. never (could be used for custom scripts where user explicitly defines when to update XS)
1. BOL
2. BOC
3. everyNode
4. firstCoupledIteration
5. all

Note that lattice physics on `firstCoupledIteration` also runs at `everyNode`. Thus, each option is at least as frequent as all options above it.

These options will give users more control over when cross sections are updated, and that control is directly through the `caseSettings` without requiring a custom `LatticePhysicsInterface` or `UserPlugin`.

## Checklist


armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -617,6 +618,10 @@ def interactEOC(self, cycle=None):
         """
         self.clearRepresentativeBlocks()
 
+    def interactEveryNode(self, cycle=None):
+        if self.cs[CONF_FORCE_LATTICE_PHYSICS]:
```
onufer, line 622: we defiantly need this on every node so we can truncate the 1e-15 isotopes from the mc2 file. if its not in every node than the build up with depletion will cause no matching xs for those isotopes since they wont be there at BOC. 

Its optional to have it at interactCoupled, but when you do it with interactCoupled it does get you a temp update from the TH solve which this call wouldn't have.

armi/physics/neutronics/latticePhysics/latticePhysicsInterface.py
```diff
@@ -228,7 +234,12 @@ def interactCoupled(self, iteration):
             This is unused since cross sections are generated on a per-cycle basis.
         """
         # always run for snapshots to account for temp effect of different flow or power statepoint
-        if self.cs[CONF_RUN_TYPE] == "Snapshots" or self.r.p.timeNode == 0:
+        secondIterOnFirstPointWithUpdatedTemps = self.r.p.timeNode == 0
```
mgjarrett, line 237: This setting is named `secondIterOnFirstPointWithUpdatedTemps` but it seems like it would happen on every `iteration`, not just the second one.

armi/reactor/blockParameters.py
```diff
@@ -17,13 +17,12 @@
...
 
 from armi import runLog
-from armi.physics.neutronics import crossSectionGroupManager
-from armi.reactor.flags import Flags  # non-standard import to avoid name conflict below
 from armi.utils import units
 from armi.utils.units import ASCII_LETTER_A
 
 from armi.reactor import parameters
 from armi.reactor.parameters import ParamLocation, Parameter, NoDefault
+from armi.physics.neutronics import crossSectionGroupManager
```
mgjarrett, line 25:
```suggestion
```

armi/reactor/blockParameters.py
```diff
@@ -17,13 +17,12 @@
 import six
 
 from armi import runLog
-from armi.physics.neutronics import crossSectionGroupManager
-from armi.reactor.flags import Flags  # non-standard import to avoid name conflict below
 from armi.utils import units
```
mgjarrett, line 20:
```suggestion
from armi.physics.neutronics import crossSectionGroupManager
from armi.utils import units
```

armi/physics/neutronics/latticePhysics/latticePhysicsInterface.py
onufer: yep. bad name on my part.

armi/physics/neutronics/settings.py
```diff
@@ -324,6 +329,13 @@ def defineSettings():
             description="Method to determine broad group cross sections for assemblies",
             options=["", "MC2v2", "MC2v3", "MC2v3-PARTISN", "SERPENT"],
         ),
+        setting.Setting(
+            CONF_LATTICE_PHYSICS_UPDATE_FREQUENCY,
+            default="BOC",
+            label="Frequency of lattice physics updates",
+            description="Define the frequency at which cross sections are updated with new lattice physics interactions.",
+            options=["", "BOL", "BOC", "everyNode", "firstCoupled", "all"],
```
keckler, line 337: I would recommend pulling the options straight from the enum so that you don't have to keep this manually updated in the future.

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -1299,3 +1329,115 @@ def blockCollectionFactory(xsSettings, allNuclidesInProblem):
...
+class LatticePhysicsUpdateFrequency(IntEnum):
+    """
+    Enumeration for lattice physics update frequency options.
+    """
+
+    BEGINNING_OF_LIFE = 1
+    BEGINNING_OF_CYCLE = 2
+    EVERY_NODE = 3
+    FIRST_COUPLED = 4
+    ALL = 5
```
keckler, line 1343: I know its a bit unorthodox to have class variables that are not all caps, but in this case I'd like to propose an exception. 

If you were to instead change these enum attributes to the names that they correspond to (i.e. make them like `BOL = 1`, `everyNode=4`, etc), then you could avoid having your `_stringMapping()` method and also remove your `getStr()` method in favor of just calling `LatticePhysicsUpdateFrequency.everyNode.name` when you want the string. This seems more in line with the enum philosophy and makes more use of the infrastructure that is already in place.

Not really important, but just a thought.

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -1299,3 +1329,115 @@ def blockCollectionFactory(xsSettings, allNuclidesInProblem):
     return BLOCK_COLLECTIONS[blockRepresentation](
         allNuclidesInProblem, validBlockTypes=validBlockTypes
     )
+
+
+class LatticePhysicsUpdateFrequency(IntEnum):
```
keckler, line 1334: It seems possible that you could eliminate the word `Update` from this class name entirely and the meaning would still be clear. If you agree, that could help to reduce clutter where this is used, since it is pretty frequent that you type out `LatticePhysicsUpdateFrequency` and `black` forces you onto multiple lines.

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -748,9 +771,16 @@ def interactCoupled(self, iteration):
...
+        if iteration == 0:
+            targetFrequency = LatticePhysicsUpdateFrequency.FIRST_COUPLED
+        else:
+            targetFrequency = LatticePhysicsUpdateFrequency.ALL
+
+        if LatticePhysicsUpdateFrequency.checkFrequency(
+            self._latticePhysicsUpdateFrequency,
+            targetFrequency,
+        ):
+            self.createRepresentativeBlocks()
```
keckler, line 783:
```suggestion
        if (iteration == 0 and self._latticePhysicsUpdateFrequency == LatticePhysicsUpdateFrequency.FIRST_COUPLED) or (self._latticePhysicsUpdateFrequency == LatticePhysicsUpdateFrequency.ALL):
            self.createRepresentativeBlocks()
```

keckler: Obviously need to format that with `black`...

armi/physics/neutronics/crossSectionGroupManager.py
keckler, line 1334: Do we want this class to sit in here? I think maybe more appropriately sit in the `__init__` of this package, since it's not specifically related to the xs group manager (you make heavy use of it in the lattice physics interface as well)

keckler (commented):
Seems like this will work, and will be a big benefit to users.

I have a number of comments mainly related to the implementation of the new enum. None of those are really that important though.

Two other major comments:
(1) What is with the new `XSGeometryTypes` enum that is introduced? This seems like maybe it belongs to a different PR and was inadvertently added here? In any case, that class is now defined twice in `crossSectionSettings`...
(2) Now that there is the possibility to update the ISOTXS at every node, that file is going to be overwritten multiple times per cycle. We save the `ISOTXS-cX` files in the directory, and typically those are generated once per cycle and then left unchanged. Now we will be losing all of them that are generated except the last one in the cycle. I would guess we would want to add an index to those file names so that we can hold on to them.

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -468,25 +471,57 @@ def test_createRepresentativeBlocksUsingExistingBlocks(self):
...
+    def test_interactEveryNode(self):
+        """Test `everyNode` lattice physics update frequency"""
         self.blockList[0].r.p.timeNode = 1
-        self.csm.interactCoupled(iteration=0)
+        self.csm.cs[CONF_LATTICE_PHYSICS_UPDATE_FREQUENCY] = "BOC"
+        self.csm.interactBOL()
+        self.csm.interactEveryNode()
         self.assertFalse(self.csm.representativeBlocks)
+        self.csm.cs[CONF_LATTICE_PHYSICS_UPDATE_FREQUENCY] = "everyNode"
+        self.csm.interactBOL()
```
albeanth, line 497: So I walked this through a debugger and can see that for this test to pass the repeated calls to `interactBOL` are needed, but I don't know why. Can you explain?

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -468,25 +471,57 @@ def test_createRepresentativeBlocksUsingExistingBlocks(self):
...
+        self.csm.interactCoupled(iteration=0)
+        self.assertFalse(self.csm.representativeBlocks)
+        self.csm.cs[CONF_LATTICE_PHYSICS_UPDATE_FREQUENCY] = "firstCoupled"
+        self.csm.interactBOL()
         self.csm.interactCoupled(iteration=0)
         self.assertTrue(self.csm.representativeBlocks)
 
+    def test_interactAllCoupled(self):
+        """Test `all` lattice physics update frequency"""
+        self.blockList[0].r.p.timeNode = 1
```
albeanth, line 515: In the series of tests you added, it doesn't look like you need the `self.blockList[0].r.p.timeNode ...` options for the tests to pass.

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -748,9 +771,16 @@ def interactCoupled(self, iteration):
         --------
         :py:meth:`Assembly <armi.physics.neutronics.latticePhysics.latticePhysics.LatticePhysicsInterface.interactCoupled>`
         """
-        # always run for snapshots to account for temp effect of different flow or power statepoint
-        if self.cs[CONF_RUN_TYPE] == "Snapshots" or self.r.p.timeNode == 0:
-            self.interactBOC(cycle=None)
+        if iteration == 0:
```
albeanth, line 774: My earlier comment seems to have disappeared. I'll give the TL/DR version this time. 

I think we should leave the functionality here the same as it was and update the cross sections on each tight coupling iteration for timeNode 0. By only running on the zeroth iteration you are assuming that the temperature distribution won't change. Which I don't think we should assume in the general case.

armi/physics/neutronics/crossSectionSettings.py
```diff
@@ -454,6 +516,8 @@ def __init__(
         self.numExternalRings = numExternalRings
         self.mergeIntoClad = mergeIntoClad
         self.meshSubdivisionsPerCm = meshSubdivisionsPerCm
+        self.xsMaxAtomNumber = xsMaxAtomNumber
+        self.minDriverDensity = minDriverDensity
```
albeanth, line 520: Do these new additions get used downstream?

albeanth (changes requested):

armi/physics/neutronics/crossSectionGroupManager.py
mgjarrett: It's a user option; a user could also decide to run on every iteration. To me the old implementation doesn't make that much sense; it would update for every coupled iteration for BOC but then not update at all for any other state point.

albeanth: One slight clarification: the old implementation would update for every coupled iteration for timeNode 0 of every cycle and would not update at all for timeNode >= 1.  
https://github.com/terrapower/armi/blob/d5a74d1cc4754f357f6b7d24f3fd6e6d04e1bb52/armi/physics/neutronics/crossSectionGroupManager.py#L735-L745

armi/physics/neutronics/tests/test_crossSectionManager.py
mgjarrett: `interactBOL` sets the `self.csm._latticePhysicsFrequency`. If you just run `interactEverynode`, it doesn't update that attribute with the new setting.

armi/physics/neutronics/tests/test_crossSectionManager.py
mgjarrett: That's a good point. I was trying to make a test that would have failed with the old code that checked `self.r.p.timeNode`, but once that old code is deleted that piece of the test is effectively obsolete.

onufer:
the failing check here seems to be Framework wide @john-science 

this code change is completely different from #1244 and both are having the same failure.

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -1172,3 +1202,115 @@ def blockCollectionFactory(xsSettings, allNuclidesInProblem):
...
+        if typeSpec not in reverseMapping:
+            raise TypeError(f"{typeSpec} not in {reverseMapping.keys()}")
+        return reverseMapping[typeSpec]
+
+    @classmethod
+    def checkFrequency(
+        cls,
+        evalFrequency: [IntEnum, str],
+        targetFrequency: [IntEnum, str],
+        exact: Optional[bool] = False,
```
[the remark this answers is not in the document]
mgjarrett: I went with just implementing the logic explicitly everywhere it is needed. Makes the `LatticePhysicsFrequency` class much simpler.

armi/physics/neutronics/settings.py
mgjarrett: Good idea!

armi/physics/neutronics/crossSectionGroupManager.py
mgjarrett: Great idea, this makes the class much simpler.

armi/physics/neutronics/settings.py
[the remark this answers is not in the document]
mgjarrett: I was kind of following some other options like `CONF_GLOBAL_FLUX_ACTIVE` and `CONF_GEN_XS` that allow empty, but that would probably cause an error if you were to actually try to use it. I could add "Never" as an option instead?

mgjarrett:
> Seems like this will work, and will be a big benefit to users.
> 
> I have a number of comments mainly related to the implementation of the new enum. None of those are really that important though.
> 
> Two other major comments: (1) What is with the new `XSGeometryTypes` enum that is introduced? This seems like maybe it belongs to a different PR and was inadvertently added here? In any case, that class is now defined twice in `crossSectionSettings`... (2) Now that there is the possibility to update the ISOTXS at every node, that file is going to be overwritten multiple times per cycle. We save the `ISOTXS-cX` files in the directory, and typically those are generated once per cycle and then left unchanged. Now we will be losing all of them that are generated except the last one in the cycle. I would guess we would want to add an index to those file names so that we can hold on to them.

(1) I think I fat-fingered a merge conflict resolution, there was a whole bunch of random duplicate or missing code. I believe I have that all fixed now.
(2) Good point, I still need to get to this. I'm not sure whether we were already overwriting the `ISTOXS-cX` files when doing tight coupling, but I'll check on that.

requirements-testing.txt
```diff
@@ -1,4 +1,5 @@
 numpy>=1.21,<=1.23.5
+ipython>=7.34.0,<=8.12.0
```
john-science, line 2: I am opening a separate PR for this here:

https://github.com/terrapower/armi/pull/1250

Fingers crossed!

armi/physics/neutronics/crossSectionSettings.py
mgjarrett: Yes

armi/physics/neutronics/crossSectionGroupManager.py
mgjarrett: What is the difference between: 
```
 it would update for every coupled iteration for BOC but then not update at all for any other state point.
```

and

```
the old implementation would update for every coupled iteration for timeNode 0 of every cycle and would not update at all for timeNode >= 1.
```
?

mgjarrett:
I've made a few more updates to address the review comments:

@keckler : I added the `ISOTXS-c{0}n{1}` formatting when XS are generated at every time node. I didn't add one for coupled iterations because it seemed too excessive and we weren't doing it previously. If the point of naming and keeping these files around is for reproducibility, I presume you would want to reproduce the last iteration and there's not as much interest in reproducing intermediate iterations.

On a slight tangent, the fact that the ISOTXS/GAMISO/PMATRX filenames are so baked in to the framework is pretty weird and unfortunate.

@albeanth let me know if you are happy with the `firstCoupled` and `all` options for XS updates or if you think there's another option that should be added.

keckler:
> @keckler : I added the ISOTXS-c{0}n{1} formatting when XS are generated at every time node. I didn't add one for coupled iterations because it seemed too excessive and we weren't doing it previously. If the point of naming and keeping these files around is for reproducibility, I presume you would want to reproduce the last iteration and there's not as much interest in reproducing intermediate iterations.

I agree with this approach.

armi/physics/neutronics/settings.py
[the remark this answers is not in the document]
albeanth: Hmmmm. Are there any use cases for a "Never" option? If a user didn't want to use cross section generation they could just not activate the interface?

armi/physics/neutronics/crossSectionGroupManager.py
albeanth: > What is the difference between: ....

BOC is technically a different statepoint than timeNode 0. We don't do coupling at BOC. We do coupling on the timeNodes. I just wanted to be clear on that distinction. 

> To me the old implementation doesn't make that much sense; it would update for every coupled iteration for BOC but then not update at all for any other state point.

So the new approach would be either:
1. Update on all iterations where tight coupling is used
    - This is expensive but users can pick and choose where to do coupling via the `cyclesSkipTightCouplingInteraction` setting introduced in #1173. 
3. Only update on the first coupling iteration where tight coupling is used 
    - I don't see the general utility in this as it assumes that the temperature distribution won't change after the zeroth iteration. Which I don't think we should assume in the general case.

If that's the case, I would advocate for the first option and removing the second in the interest of clarity/simplicity. What do you think?

armi/physics/neutronics/crossSectionSettings.py
albeanth: Does it makes sense to/can we move them to the downstream plugin then? Maintaining things at the FW level that only get used in downstream plugins seems iffy/icky to me (I realize that we do this a lot throughout currently, but I think we should limit the amount of new stuff that we inject into the FW that only gets used in a downstream plugin).

mgjarrett: It's not a good pattern, but it's what exists. These are not the first settings to be defined like this. `nuclideReactionDriver`, `externalDriver`, and `useHomogenizedBlockComposition` are only used in downstream plugins (Also, the recently added `xsExecuteExclusive` and `xsPriority`). Additionally, many of the settings that are used in the framework are only used in methods that are only called by downstream plugins (`numInternalRings`, `numExternalRings`, `mergeIntoClad`, and `meshSubdivisionsPerCm` ).

It would be a bit messy because we'd have to subclass `XSModelingOptions` and `XSSettings` in our internal lattice physics interfaces and then use factories to get the correct types of each. I haven't attempted it, but at first glance it seems like a somewhat heavy lift. I would prefer to do it in a separate PR if possible, and then we could move all of the settings that don't belong in the framework.

armi/physics/neutronics/crossSectionGroupManager.py
mgjarrett: I think `firstCoupled` is useful because you could have a spatially uniform temperature distribution from the blueprints on the initial `everyNode` interaction (or when loading from a database of an un-coupled case). On the first coupled iteration, you would be capturing a potentially large change in temperature distribution. The change between those first two states could be orders of magnitude larger than the changes in subsequent coupled iterations.

We are not assuming anything is true in the general case, we're just providing an option to speed up calculations when, in the user's judgement, there is a lot of benefit to be had from updating XS after the zeroth coupled iteration and less benefit to be had after that.

armi/physics/neutronics/crossSectionSettings.py
albeanth: Sounds good. I wonder if it would be of interest for us to start outlining what this sort of refactor would require. @john-science any strong-feelings here?

armi/physics/neutronics/crossSectionGroupManager.py
albeanth: > On the first coupled iteration, you would be capturing a potentially large change in temperature distribution. The change between those first two states could be orders of magnitude larger than the changes in subsequent coupled iterations.

I totally agree that you would capture a potentially large change and that it would potentially be larger than subsequent calculations. But I would counter with that to really capture the nonlinearity, you would need to continue updating the cross sections for subsequent iterations. Otherwise you're missing the feedback mechanism, not really resolving the nonlinearity adequately. and providing a false convergence. This feedback could be particularly strong for varying types of reactors. 

But I think it's fine to provide the flexibility. My only request would be to rename `firstCoupled` to `firstCoupledIteration` and provide a description of it's intended use case.

armi/physics/neutronics/settings.py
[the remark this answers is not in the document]
mgjarrett: I guess you could also have `existingXSFiles` and you might not want to run any lattice physics during a run.

albeanth (approved):

mgjarrett:
@albeanth I've added 7 commits since you approved but it's all just merge, black formatting, etc. The only real change is commit 85bf8c9 where I changed a default from `None` to `0.0`.

Issue #1191, opened 2023-02-22
Define flags doesn't work for a normal armi submit run


onufer:
This error is never hit when i run armi locally on our cluster. contact me for the directory for full example. 

![image](https://user-images.githubusercontent.com/50242044/220761378-a073100c-55d5-4ec7-aeaf-0a4185a01266.png)

john-science:
Mark's result here is surprising, since we have [significant testing](https://github.com/terrapower/armi/blob/3e47ec6e6359870866502c9356747cf1da4f9515/armi/tests/test_user_plugins.py#L32) in ARMI for the `defineFlags()` feature on `UserPlugins`.

The tests even build an `App` from scratch and then load plugins.

Should be interesting.

PR #1241, opened 2023-04-12
Fixing UserPlugins.defineFlag

john-science:
## Description

This is in response to #1191.  @onufer I was able to reproduce your bug locally (though using a stand-alone App), and this fixed the bug for me.  Before we merge this PR I was hoping you would get to run this commit/change through your workflow to make sure this fixes the issue for you as well.  Thanks!

---

## Checklist

### DIFF
--- a/armi/apps.py
+++ b/armi/apps.py
@@ -35,8 +35,9 @@ import sys
 
 from armi import context, plugins, pluginManager, meta, settings
 from armi.reactor import parameters
-from armi.settings import Setting
+from armi.reactor.flags import Flags
 from armi.settings import fwSettings
+from armi.settings import Setting
 
 
 class App:
@@ -294,7 +295,6 @@ class App:
         bool
             Whether or not the plugin name is already registered with the manager.
         """
-
         if ":" in pluginPath:
             pluginName = pluginPath.strip().split(":")[-1]
         else:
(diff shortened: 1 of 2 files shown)


bsculac (approved):

keckler (approved):
So this appears to have just been an oversight before, not really a "bug"? We just forgot to actually load the flags when we imported the user plugins?

Either way, looks fine now.

armi/apps.py
```diff
@@ -332,6 +337,11 @@ def __registerUserPluginsInternalImport(self, pluginPath):
         assert issubclass(plugin, plugins.UserPlugin)
         self._pm.register(plugin)
 
+        # ensure UserPlugin flags are loaded
+        newFlags = plugin.defineFlags()
+        if newFlags:
+            Flags.extend(newFlags)
+
```
onufer, line 344: this works. there is a bit of duplicate code here, so maybe eventually we should make a `registerUserPlugin` method where everything from 
assert issubclass(plugin, plugins.UserPlugin)
down goes in

armi/apps.py
```diff
@@ -35,8 +35,9 @@
 
 from armi import context, plugins, pluginManager, meta, settings
 from armi.reactor import parameters
-from armi.settings import Setting
+from armi.reactor.flags import Flags
 from armi.settings import fwSettings
+from armi.settings import Setting
```
onufer, line 40: appears unused?

onufer (approved):

armi/apps.py
onufer:
```suggestion
```

john-science: No, this import is needed for line 121.

Issue #1221, opened 2023-03-23
Settings validation resolution appears to not work when submitted via cmd line


onufer:
validation resolution such as
![image](https://user-images.githubusercontent.com/50242044/227262771-56d55621-11f8-4593-b5a6-260fe9fe3e3c.png)

doesnt seem to work in cmd line  even if you say `yes`
![image](https://user-images.githubusercontent.com/50242044/227263002-e51b94ff-0e0f-43af-9757-f356eea327a2.png)

onufer:
yep correct. when you do it with the gui, it actually changes the settings files, saves new ones, and submits those, so that should be the actual behavior

john-science:
Are we at all worried that we are over-writing the original setting file, and that is unrecoverable? 

For instance, if the settings file is on a share folder, and multiple people are using it for different purposes, then this gives someone the power to change the settings file out from underneath another user.

Just asking.

onufer:
there could be some worry there. but the message warns you that this is what it will do. we could additionally rename the old file and write a new one.

PR #1244, opened 2023-04-14
Update settings file when user agrees to corrective query settings

HunterPSmith:
## Description
Seeks to resolve https://github.com/terrapower/armi/issues/1221. 

When a corrective query is resolved per the discretion of the use, the inspector now prompts the user whether to preserve the original settings. If yes, then a copy of the original settings is saved into the same filename but with an additional suffix: `_old.yaml`. If no, then the settings file is overwritten with the new settings. If an `_old.yaml` file exists in the repository, then the inspector will prompt the user if the `_old.yaml` file should be overwritten.

Testing has been done to make sure that the settings file is overwritten with new changes and the original settings are preserved correctly in the `_old.yaml` file.

---

## Checklist

### DIFF
--- a/armi/operators/settingsValidation.py
+++ b/armi/operators/settingsValidation.py
@@ -200,7 +200,7 @@ class Inspector:
             )
 
         else:
-            # preserve original settings
+            # preserve original settings in case a corrective query is resolved
             if any(query.isCorrective() for query in self.queries):
                 self.csOriginal = self.cs.duplicate()
 
@@ -224,13 +224,13 @@ class Inspector:
                 )
             runLog.debug("{} has finished querying.".format(self.__class__.__name__))
 
-        # Decide what to do with original settings
+        # decide what to do with original settings
         if correctionsMade:
             self.cs.writeToYamlFile(self.cs.path)
             oldFilePath = "{}_old.yaml".format(self.cs.path.split(".yaml")[0])
             if os.path.exists(oldFilePath):
                 overwrite_file = prompt(
-                    'INSPECTOR: At least one corrective query has been resolved and an "_old" settings file {} exists.'.format(
+                    'INSPECTOR: At least one corrective query has been resolved and an "_old.yaml" settings file {} exists.'.format(
                         oldFilePath
                     ),
                     'Overwrite "_old.yaml" settings file to preserve originally submitted settings?',
--- a/armi/operators/settingsValidation.py
+++ b/armi/operators/settingsValidation.py
@@ -226,9 +226,9 @@ class Inspector:
             strSkeleton = "{}_old".format(self.cs.path.split(".yaml")[0])
             for num in itertools.count():
                 if num == 0:
-                    renamePath = "{strSkeleton}.yaml"
+                    renamePath = f"{strSkeleton}.yaml"
                 else:
-                    renamePath = "{strSkeleton}{num}.yaml"
+                    renamePath = f"{strSkeleton}{num}.yaml"
                 if not self._csRelativePathExists(renamePath):
                     break
             # preserve old file before saving settings file
--- a/armi/operators/settingsValidation.py
+++ b/armi/operators/settingsValidation.py
@@ -251,8 +251,9 @@ class Inspector:
                 )
                 if preserve_settings:
                     self.csOriginal.writeToYamlFile(oldFilePath)
-        # overwrite settings file
-        self.cs.writeToYamlFile(self.cs.path)
+            # overwrite settings file
+            self.cs.writeToYamlFile(self.cs.path)
+
         return correctionsMade
 
     def addQuery(self, condition, statement, question, correction):
--- a/armi/operators/settingsValidation.py
+++ b/armi/operators/settingsValidation.py
@@ -237,7 +237,9 @@ class Inspector:
 
             # preserve old file before saving settings file
             if not self._csRelativePathExists(oldFilePath) or saveOldSettings:
-                runLog.extra(f"Preserving originally submitted settings file in `{oldFilePath}`")
+                runLog.extra(
+                    f"Preserving originally submitted settings file in `{oldFilePath}`"
+                )
                 shutil.copy(self.cs.path, oldFilePath)
 
             # save settings file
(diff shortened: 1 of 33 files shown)


john-science:
Are we at all worried that we are over-writing the original setting file, and that is unrecoverable? 

For instance, if the settings file is on a share folder, and multiple people are using it for different purposes, then this gives someone the power to change the settings file out from underneath another user.

Just asking.

armi/operators/tests/test_inspectors.py
```diff
@@ -52,6 +57,50 @@ def defdef(x, y, z):
...
+        self.cs.loadFromInputFile(self.filepathYaml)
+
+        self.assertEqual(self.cs["cycleLength"], 666)
+
+        # reset prompt function to nominal
+        settingsValidation.prompt = nominalPromptFunction
+
+        # reset test settings file
+        self.cs["cycleLength"] = 300.0
+        self.cs.writeToYamlFile(self.filepathYaml)
```
john-science, line 102: Okay, this block is super important.  I like that it exists.

I almost feel like we should wrap the rest of this test in a try/except/finally, and put this in the finally, to make sure it gets run, even if the test fails.

Thoughts?

HunterPSmith:
> Are we at all worried that we are over-writing the original setting file, and that is unrecoverable?
> 
> For instance, if the settings file is on a share folder, and multiple people are using it for different purposes, then this gives someone the power to change the settings file out from underneath another user.
> 
> Just asking.


It's a good question. I have thought about it, but from my experience, if someone is helping me out with trouble shooting or if I'm sharing it with someone who is interested in the case I ran, they usually just copy all the input files including the settings file over into their own folder and then run it or make changes.  I can't really think of a situation where multiple people would need to point to the same settings file. 

Also, if multiple people are working from the same settings file, the user trying to run the settings is being asked by the inspector if they want to change the settings, so maybe we leave that up to them to decide if it is appropriate or if they should notify the other users. Currently, when the inspector prompts the user, it implies that it is fixing the files for you, so the settings get changed in memory but not in the settings file which is problematic because then you have an output that doesn't actually reflect the "true" settings file.

armi/operators/tests/test_inspectors.py
HunterPSmith: Now that I am looking at it, maybe we don't actually need this block. The test starts by generating the test settings file, so even if it wasn't overwritten, the test starts by writing the file anyways. What do you think?

john-science: If the test runs 2 or 3 times in a row without this block, great, remove it. Sure.

armi/operators/settingsValidation.py
```diff
@@ -203,6 +203,11 @@ def run(self, cs=None):
                 query.resolve()
                 if query.corrected:
                     correctionsMade = True
+                if correctionsMade:
+                    runLog.extra(
+                        f"At least one setting correction detected. Overwriting settings file `{self.cs.path}`."
```
onufer, line 208: can we rename the old settings file as `fileName_old.yaml` or something like that so we retain the old file?

probably also check if that path exists and avoid overwitting something that alread exists if `fileName_old.yaml` exists

HunterPSmith: Just updated the code to handle this better I think. Going to fix the tests now.

HunterPSmith:
Ok, this looks a bit different now. I updated the PR description. But now I think it now handles it in a way that makes it very clear to the user what is going on.

armi/operators/settingsValidation.py
```diff
@@ -224,6 +224,36 @@ def run(self, cs=None):
...
+                    'INSPECTOR: At least one corrective query has been resolved and an "_old.yaml" settings file {} exists.'.format(
+                        oldFilePath
+                    ),
+                    'Overwrite "_old.yaml" settings file to preserve originally submitted settings?',
+                    "YES_NO",
+                    "NO_DEFAULT",
+                    "CANCEL",
+                )
+                if overwrite_file:
+                    self.csOriginal.writeToYamlFile(oldFilePath)
```
onufer, line 241: please change this just to a rename. loading int memory and then rewritting will suppress default settings and a may delete comments/change formatting. we want the exact original file.

HunterPSmith: Ok I have changed this to copy the file and rename instead of reading into memory.

armi/operators/settingsValidation.py
```diff
@@ -198,7 +199,13 @@ def run(self, cs=None):
                     self.__class__.__name__
                 )
             )
+
         else:
+            # preserve original settings file in case a corrective query is resolved
+            if any(query.isCorrective() for query in self.queries):
+                tempFilePath = "{}_temp.yaml".format(self.cs.path.split(".yaml")[0])
+                shutil.copyfile(self.cs.path, tempFilePath)
```
onufer, line 207: just do shutil.copy instead of copy file. copy has the better beahvior i think 
image

onufer: Also, why can't this code be moved down to the other code. is there a reason the temp file is necessary? cant we just check if the _old exists and copy it if it does not?

armi/operators/settingsValidation.py
```diff
@@ -219,6 +226,38 @@ def run(self, cs=None):
...
+            else:
+                preserve_settings = prompt(
+                    "INSPECTOR: At least one corrective query has been resolved.",
+                    "Preserve originally submitted settings in: {}?".format(
+                        oldFilePath
+                    ),
+                    "YES_NO",
+                    "NO_DEFAULT",
+                    "CANCEL",
+                )
```
onufer, line 253: I recommend always preserving and deleting the prompt. its another yes/no that someone will be asked every time... just make it  ( can give a message it is being made) and if the user doesn't want it, they can delete manually which is probably easier than another yes/no.

armi/operators/settingsValidation.py
```diff
@@ -219,6 +226,38 @@ def run(self, cs=None):
...
+                if preserve_settings:
+                    os.replace(tempFilePath, oldFilePath)
```
onufer, line 255: here is where i would recommend putting hte shUtil instead of replace and then you dont need the code above.

onufer (approved):

PR #1253, opened 2023-05-03
Make eps an optional parameter for getBlocksBetweenElevations

mgjarrett:
## Description
---

The `EPS` value for filtering out small blocks from `assemblies.getBlocksBetweenElevations()` was hard-corded to 1-e10. This PR makes it an optional parameter with a default of 1e-10, which gives the user more flexibility to get blocks with a different tolerance.

## Checklist

### DIFF
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -912,7 +912,7 @@ class Assembly(composites.Composite):
                 return bIndex
         return -1  # no block index found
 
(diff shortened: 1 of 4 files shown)


albeanth (approved):
Coverage failure looks like coverage spray to me. 

Just need a release notes update and this is good to go.

mgjarrett:
Don't need this; #1257 solves the same problem but it's better

PR #1248, opened 2023-04-20
Filter by minimum nuclide dens for generating macros

mgjarrett:
## Description
---

The `minimumNuclideDensity` setting has existed for a long time, but it was previously only used to set the initial nuclide density for trace isotopes in the lattice physics calculations.

This PR is part of an effort to make the `minimumNuclideDensity` a threshold below which nuclides are not included in global flux or reaction rate calculations. This will allow us to easily ignore isotopes at very small concentrations (e.g., below 1e-12 atoms/b-cm) if we don't have XS for them in the XS library.

This change does not _break_ the API, per se, but it does add the optional `minimumNuclideDensity` parameter argument to the `__init__` of certain classes, such as `macroXSGenerationInterface.MacroXSGenerator` and `xsCollections.MacroscopicCrossSectionCreator`. Since these arguments are optional and default to 0.0, it should affect any existing implementations.

## Checklist

- [x] This PR has only one purpose or idea.
- [x] Tests have been added/updated to verify that the new/changed code works.

- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html).

- [ ] The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any bug fixes or new features.
- [ ] The documentation is still up-to-date in the `doc` folder.
- [x] The dependencies are still up-to-date in `setup.py`.

PR #1257, opened 2023-05-04
Add detailed isotopes to all blocks when non-uniform reactor

onufer:
## Description

Code update to add all isotopes to all blocks when the reactor is non-uniform.

This is a more stable way of fixing isotopes missing in cross section ids than using tolerancing (eg #1248)

it maybe make some lattice physics computations a little more expensive, so the ultimate solution might be to allow blocks to have 2+ xsIDs to check, and if the nuc wasn't in the first one go down the list and check the next XSID, but this would be a much more significant framework change.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/fissionProductModel/fissionProductModel.py
+++ b/armi/physics/neutronics/fissionProductModel/fissionProductModel.py
@@ -104,6 +104,7 @@ from armi.physics.neutronics.fissionProductModel.fissionProductModelSettings imp
     CONF_FP_MODEL,
     CONF_MAKE_ALL_BLOCK_LFPS_INDEPENDENT,
 )
+from armi.settings.fwSettings.globalSettings import CONF_DETAILED_AXIAL_EXPANSION
 
 NUM_FISSION_PRODUCTS_PER_LFP = 2.0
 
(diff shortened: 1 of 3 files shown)


armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -185,7 +192,25 @@ def setAllComponentFissionProducts(self):
...
+            if allBlocksNeedAllNucs and not compsToAddIso:
+                # add the isotopics to the smallest solid since that is usually the most "interesting"
+                # sorted() calls getBoundingCircleOuterDiameter under the hood
+                solidsOrderedBySize = sorted(c for c in b if c.containsSolidMaterial())
+                if solidsOrderedBySize:
+                    compsToAddIso = [solidsOrderedBySize[0]]
+                else:
+                    # no solids, so just add to smallest component
+                    compsToAddIso = [
+                        sorted(c for c in b if c.containsSolidMaterial())[0]
```
mgjarrett, line 211: This line seems to produce the same exact `compsToAddIso` as the one in the `if` part of the block. Should `if c.containsSolidMaterial()` be removed here?

onufer: yes! thanks!

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -162,21 +163,27 @@ def interactBOL(self):
 
     def setAllComponentFissionProducts(self):
         """
-        Initialize all blueprint nuclides for each ``DEPLETABLE`` component in the core.
+        Initialize all nuclides for each ``DEPLETABLE`` component in the core, or all blocks if mesh conversion.
```
mgjarrett, line 166:
```suggestion
        Initialize all nuclides for each ``DEPLETABLE`` component in the core, or all blocks if detailedAxialExpansion is enabled.
```

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -162,21 +163,27 @@ def interactBOL(self):
...
         This should be called when explicit fission product modeling is enabled to
-        ensure that fission products are initialized on the depletable components within
-        the reactor data model.
+        ensure that all isotopes are initialized on the depletable components within
+        the reactor data model so that cross sections are created during depletion.
+
+        When detailedAxialExpansion is also enabled, all regions will have fission/activation
+        products added to avoid missing cross sections during mesh conversion (since converted
+        blocks only have one xsID but may have isotopes from multiple blocks with different IDs.)
+        Setting density to zero her enables small number densities is XS generation.
```
mgjarrett, line 195:
```suggestion
        Setting density to zero here enables small number densities is XS generation.
```

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -185,7 +192,23 @@ def setAllComponentFissionProducts(self):
         """
         for b in self.r.core.getBlocks(includeAll=True):
             b.setLumpedFissionProducts(None)
-            for c in b.getComponents(Flags.DEPLETABLE):
+
+            # If detailed axial expansion is active, mapping between blocks occurs on uniform mesh
+            # and this can cause blocks to have isotopes that they dont have cross sections for/
```
mgjarrett:
```suggestion
            # and this can cause blocks to have isotopes that they dont have cross sections for
```

mgjarrett (approved):

mgjarrett (approved):
I've tested this with explicit fission products downstream and it works. Haven't tested with lumped fission products, but it uses the same code and looks like it would also work.

PR #1258, opened 2023-05-04
Fix a bug in database compare.

mgjarrett:
## Description
---

There's a line in compareDB3.py that tries to call a method that doesn't exist, `DiffResults.add()`

This has been around for a long time, but we have somehow just coincidentally not been hitting it.

Also, another call to `addDiff` needed to be updated to pass `numpy.inf` as a scalar instead of a list.

## Checklist

### DIFF
--- a/armi/bookkeeping/db/compareDB3.py
+++ b/armi/bookkeeping/db/compareDB3.py
@@ -377,7 +377,9 @@ def _diffSpecialData(
         if isinstance(dSrc, numpy.ndarray) and isinstance(dRef, numpy.ndarray):
             if dSrc.shape != dRef.shape:
                 out.writeln("Shapes did not match for {}".format(refData))
-                diffResults.addDiff(compName, paramName, numpy.inf, numpy.inf, numpy.inf)
+                diffResults.addDiff(
+                    compName, paramName, numpy.inf, numpy.inf, numpy.inf
+                )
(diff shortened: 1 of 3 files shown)


armi/bookkeeping/db/compareDB3.py
```diff
@@ -329,7 +329,7 @@ def _diffSpecialData(
     diffResults.addStructureDiffs(nDiffs)
 
     if not keysMatch:
-        diffResults.addDiff(name, name, [numpy.inf], [numpy.inf], [numpy.inf])
```
onufer, line 332: can you describe in the PR why the brackets were removed?

onufer (approved):

mgjarrett:
I was able to add a unit test for one of the bad function calls.

For the second one, where the name of the function was wrong (`add()` instead of `addDiff()`), I have not been able to get the unit test to hit the line. To get here, we have to read some data `src` from an HDF5 dataset, iterate throught `src.tolist()` which ostensibly creates regular python `list`s, and then check if any of the items created by `tolist()` are `numpy.ndarray` instead of a regular `list`. I don't know how to make this happen, and I'm not sure how I happened upon it yesterday. For now I'll settle for leaving the line uncovered; I know the fix works because I tried it on my in-use case that was failing.

PR #1207, opened 2023-03-08
Update way that settings report defaults are printed

keckler:
## Description

The settings report in the docs is supposed to print out the default values. For many boolean settings that default to `False`, nothing is printed in the default column.
An example:
image

The default value for `trackAssems` is `False`:
https://github.com/terrapower/armi/blob/9de4be8b706a3b4ef96d78b57632fb3130bf3114/armi/settings/fwSettings/globalSettings.py#L695-L703

Turns out there is a small piece of the logic that generates the table that will turn values of `False` into empty strings.

This fixes that.

---

## Checklist

- [x] This PR has only one purpose or idea.
- [ ] Tests have been added/updated to verify that the new/changed code works.

- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html).

- [x] The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any bug fixes or new features.
- [x] The documentation is still up-to-date in the `doc` folder.
- [x] The dependencies are still up-to-date in `setup.py`.

Issue #1249, opened 2023-04-25
Settings Report page is broken


keckler:
Sometime in the past week or two, the "Settings Report" page of our docs appears to have broken. It no longer renders the big table of all the settings. This is what I see when I go to the page:
![image](https://user-images.githubusercontent.com/21959600/234339385-0f78e521-f86f-4b76-a4ec-9a96fc3c120e.png)

john-science:
I guess we need to find the commit that broke this.  Damn.

PR #1263, opened 2023-05-16
Fix #1207

keckler:
## Description

#1207 included a syntax bug, namely a missing closing parenthesis. This was a bit difficult to see originally because python syntax highlighting won't work in this instance of code embedded in an rst file.

It was discovered that the logic implemented in #1207 also didn't work anyways, so this PR fixes it.

Before #1207 , an empty string was printed if the default value was `None`. This was the initial problem that was identified. Now `None` values are printed when there is no default, `None` values are printed if the default is `None`, and an empty string will be printed if the default is an empty string. This is what is desired.

---

## Checklist


keckler:
@john-science Okay, I did some offline testing, this should be good to go now. This particular piece seems to be working, but due to #1249 (settings report page is broken in general right now) I cannot fully generate the docs on my machine right now.

john-science (approved):

PR #1265, opened 2023-05-17
Block collection burnup

mgjarrett:
## Description
---

Calculate the `b.p.percentBu` for the representative block in a `BlockCollection`. This is only used for logging purposes; the `percentBu` parameter does not factor into any physics calculations. Without adding this calculation, the representative block previously just had the `percentBu` from the original block that it is copied from.

## Checklist


onufer:
can you clarify the problem? i think it was that we were just using the burnup of the original copied block rather than average burnup of the blocks used to determine the composition?

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -238,6 +238,23 @@ def getCandidateBlocks(self):
...
+
+        Notes
+        -----
+        - Only used for logging purposes
+        - Burnup needs to be weighted by heavy metal mass instead of volume
+        """
+        weightedBurnup = 0.0
+        totalWeight = 0.0
+        for b in self:
+            weighting = b.p.massHmBOL * self.getWeight(b) / b.getVolume()
```
onufer, line 253: why divide by volume here?

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -392,6 +395,22 @@ def test_addXsGroupsFromBlocks(self):
         self.assertEqual(len(blockCollectionsByXsGroup), 4)
         self.assertIn("AB", blockCollectionsByXsGroup)
 
+    def test_calcWeightedBurnup(self):
+        self.blockList[1].p.percentBu = 3.1
+        self.blockList[2].p.percentBu = 10.0
```
onufer, line 400: maybe assert they all have the same volume or weight factor  since we need to know that to know its a simple avg

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -392,6 +395,22 @@ def test_addXsGroupsFromBlocks(self):
...
+            b.p.percentBu = 0.0
+        self.csm._updateBurnupGroups(self.blockList)
+        blockCollectionsByXsGroup = {}
+        blockCollectionsByXsGroup = self.csm._addXsGroupsFromBlocks(
+            blockCollectionsByXsGroup, self.blockList
+        )
+        self.assertEqual(
+            blockCollectionsByXsGroup["AA"]._calcWeightedBurnup(), 1 / 12.0
+        )
+        self.assertEqual(blockCollectionsByXsGroup["AB"]._calcWeightedBurnup(), 6.55)
```
onufer, line 412: can you test AA too?

armi/physics/neutronics/crossSectionGroupManager.py
mgjarrett: Yes, `getWeight` multiplies the weighting by volume. I wanted to just multiply the mass by the XSGM weighting factor.

armi/physics/neutronics/tests/test_crossSectionManager.py
mgjarrett: I added an `assertEqual` to confirm `self.blockList[1]` and `self.blockList[2]` have the same weighting factor.

armi/physics/neutronics/tests/test_crossSectionManager.py
mgjarrett: I set a single block to 1.5 percent burnup and then zeroed out all of the others. There are 18 blocks in `AA`, so the average burnup is 1.5/18 = 1/12. That's what that assert is for.

armi/physics/neutronics/crossSectionGroupManager.py
onufer: sorry, i was suggesting comment. if you prefer doc string thats fine but added suggestion

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -238,6 +238,26 @@ def getCandidateBlocks(self):
...
+        - Only used for logging purposes
+        - Burnup needs to be weighted by heavy metal mass instead of volume
+          self.getWeight(b) incorporates the volume, so we divide that weight
+          by volume so that volume is not effectively double-counted through
+          both the the b.p.massHmBol and the self.getWeight(b)
+        """
+        weightedBurnup = 0.0
+        totalWeight = 0.0
+        for b in self:
+            weighting = b.p.massHmBOL * self.getWeight(b) / b.getVolume()
```
onufer, line 256:
```suggestion
        """
        weightedBurnup = 0.0
        totalWeight = 0.0
        for b in self:
            # self.getWeight(b) incorporates the volume as does mass, so divide by volume not to double-count
            weighting = b.p.massHmBOL * self.getWeight(b) / b.getVolume()
```

armi/physics/neutronics/tests/test_crossSectionManager.py
onufer: doh!

onufer (approved):

PR #1271, opened 2023-05-18
block converter preserve flags

onufer:
## Description

Fixed a bug where block converter did not preserve flags
---

## Checklist

### DIFF
--- a/armi/reactor/converters/blockConverters.py
+++ b/armi/reactor/converters/blockConverters.py
@@ -639,7 +639,6 @@ class HexComponentsToCylConverter(BlockAvgToCylConverter):
             bigComponent = self._addSolidMaterialRing(
                 bcs, innerDiameter, outerDiameter, name
             )
-            bigComponent.p.flags = bcs.p.flags
             self.convertedBlock.add(bigComponent)
             innerDiameter = outerDiameter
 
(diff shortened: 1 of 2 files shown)


armi/reactor/converters/blockConverters.py
```diff
@@ -637,6 +639,7 @@ def _buildNthRing(self, pinComponents, ringNum):  # pylint: disable=too-many-loc
             bigComponent = self._addSolidMaterialRing(
                 bcs, innerDiameter, outerDiameter, name
             )
+            bigComponent.p.flags = bcs.p.flags
```
mgjarrett, line 642: Can the transferring of flags be moved into `_addSolidMaterialRing`? That would be easier to track and less error prone if somebody were to use `_addSolidMaterialRing` somewhere else in the future.

armi/reactor/converters/blockConverters.py
```diff
@@ -602,10 +603,11 @@ def _classifyComponents(self):
 
     def _buildFirstRing(self, pinComponents):
         """Add first ring of components to new block."""
-        newComps = copy.deepcopy(pinComponents)
-        for c in newComps:
+        for oldC in pinComponents:
+            c = copy.deepcopy(oldC)
```
mgjarrett, line 607: I also considered using `composites.Composite.setName()` instead of `component.setName()` here to avoid clobbering the flags. It would be a one-line change instead of 3, but then we have to import `composites` here so maybe this is better.

mgjarrett (approved):

PR #1273, opened 2023-05-22
Adding to GlobalFluxOptions docstring

john-science:
## Description

This PR attempts to document the over-the-top number of class attributes in `GlobalFluxOptions`. 

In the process, I was able to remove two entirely unused class attributes. But it turns out there was one more secret one that wasn't defined in the constructor to begin with.

I also found a few of these aren't used in ARMI at all, but only in downstream plugins.  So, I couldn't do as much cleanup as I was hoping.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -32,6 +32,7 @@ from armi.reactor.blocks import Block
 from armi.reactor.converters import geometryConverters
 from armi.reactor.converters import uniformMesh
 from armi.reactor.flags import Flags
+from armi.settings.caseSettings import Settings
 from armi.utils import units, codeTiming, getMaxBurnSteps
 from armi.physics.neutronics.settings import (
     CONF_BOUNDARIES,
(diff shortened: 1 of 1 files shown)


john-science:
I probably didn't need to add so many reviewers. It's Tony's ticket.

But there are a lot of descriptions there, so who knows. I tried to be clear, and use existing setting descriptions where available.

Also... this class has SO MANY attributes with no defaults.  That just feels... really strange to me.  Like, half of those are booleans. We seriously can't default those?  hmmmmmm

albeanth:
> Also... this class has SO MANY attributes with no defaults. That just feels... really strange to me. Like, half of those are booleans. We seriously can't default those? hmmmmmm

I'd love to move all of the attributes that are only used in downstream plugins to those plugins, but I've been told that's not as easy as it seems....

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -345,49 +346,99 @@ def getLabel(caseTitle, cycle, node, iteration=None):
...
+    kernelName : str
+        The neutronics / depletion solver for global flux solve.
+    loadPadElevation : float
+        The elevation of the bottom of the above-core load pad (ACLP) from
+        the bottom of the upper grid plate (in cm).
+    loadPadLength : float
+        The length of the load pad. Used to compute average and peak dose.
+    maxOuters : int
+        XY and Axial partial current sweep max outer iterations.
+    savePhysicsFilesList : bool
```
mgjarrett, line 402: `savePhyiscsFilesList` is the list of timestamps from the settings file. `self.savePhysicsFiles` is a boolean derived from `self.savePhysicsFilesList`

john-science: (This is the "Attributes" section of the docstring, so everything listed here is a "self.something".)

john-science:
> I'd love to move all of the attributes that are only used in downstream plugins to those plugins,

True. But this is supposed to just be a docstring PR, not a change to the API.

Still, we could look into that next.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -345,49 +346,99 @@ def getLabel(caseTitle, cycle, node, iteration=None):
...
+        in certain global flux implementations if reaction rates are not
+        required, but by default it is enabled.
+    eigenvalueProblem : bool
+        Whether this is a eigenvalue problem or a fixed source problem
+    includeFixedSource : bool
+        This can happen in eig if Fredholm Alternative satisfied.
+    photons : bool
+        Run the photon/gamma uniform mesh converter?
+    real : bool
+        True if  ``CONF_NEUTRONICS_TYPE`` setting is set to ``real``.
```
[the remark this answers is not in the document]
john-science: I added that to the docstring.

albeanth:
> > I'd love to move all of the attributes that are only used in downstream plugins to those plugins,
> 
> True. But this is supposed to just be a docstring PR, not a change to the API.
> 
> Still, we could look into that next.

Agreed. 100%.

PR #1279, opened 2023-05-25
Update Thermal Expansion Routines to Not Use Linearized Form

albeanth:
## Description
This PR removes the unnecessary use of the linearized form of thermal expansion in ARMI. There are several to-do items remaining before this PR is ready to open:
- [x] ARMI unit tests (coverage failure in CI looks like coverage spray)
- [ ] fix unit testing in downstream applications
- [x] update ARMI docs 
- [ ] determine merge order for downstream application updates

During downstream testing, a corner case was found that was not previously accounted for. When a block is not expanded/contracted, but its volume changes due to the changes from the blocks below it, its number densities were not updated. E.g., using one of the test fuel assemblies defined in `armi/tests/refSmallReactorBase.yaml`, if only the fuel components are artificially expanded, the plenum block will have its volume shrunk but not have the number densities updated (no growth factor for the components within it, but the number densities have changed). Commit https://github.com/terrapower/armi/pull/1279/commits/312aed7bdf0e5344b1d47586535f4c074c07284d addresses this by setting the growth factor equal to the change in volume of the block. 

---

## Checklist


albeanth:
@alexhjames, please provide a review when available to.

albeanth:
@onufer @ntouran @jakehader fyi. The origin of this work is that it's an identified update for a release for a downstream application.

albeanth:
@sombrereau if this affects your project(s), let me know and I am happy to help resolve and provide necessary updates. 

@drewj-usnctech a heads up, this will likely affect any internal testing suites you may have. I'd offer to help resolve, but I doubt that I can.

armi/materials/material.py
```diff
@@ -214,19 +215,19 @@ def linearExpansionFactor(self, Tc: float, T0: float) -> float:
         --------
         linearExpansionPercent
         """
-        dLLhot = self.linearExpansionPercent(Tc=Tc)
-        dLLcold = self.linearExpansionPercent(Tc=T0)
+        dLLhot = self.linearExpansionPercent(Tc=Tc) / 100.0
+        dLLcold = self.linearExpansionPercent(Tc=T0) / 100.0
 
-        return (dLLhot - dLLcold) / (100.0 + dLLcold)
+        return exp(dLLhot) / exp(dLLcold)
```
alexhjames, line 221: It probably is faster to just do `exp(dLLhot - dLLcold)`

alexhjames, line 221: Also is there a reason that we removed the "-1" so that this is centered around zero? I don't really care either way, but it's certainly less impactful to downstream code (and would remove a lot of the changes in this PR) if the definition of what was being returned as kept consistent.

alexhjames:
@albeanth Just two comments. I didn't really look through the test changes outside of `linearExpansionFactor` as I'm not well suited to reviewing those.

armi/materials/material.py
albeanth: Yeah that's a good perspective on minimizing the the impacts on downstream code. 

Looking back I suppose I ended up taking it out as I thought it simplified things by removing the need for `+ 1`s all over the place. 

Minimizing impacts on downstream code is probably more important. I'll put it back.

albeanth: Addressed https://github.com/terrapower/armi/pull/1279/commits/014a53503c227f55ade39d92cc818e3d3e381f34

jakehader:
@albeanth, can you explain a bit more in the PR description of what the change is, what it means, and why it's needed, and is it generally correct for all materials?

sammiller11235:
I'm not sure this is physically correct. Thermal expansion data measured by dilatometer is almost always reported at deltaL/L0, so essentially "engineering strain" form, should be used for consistency with how the material properties were measured and are defined. Some discussion: https://www.continuummechanics.org/hydrodeviatoricstrain.html#volume_strains and https://www.continuummechanics.org/truestrain.html

jakehader:
@albeanth / @alexhjames please set up a time to discuss. Also, I'm not sure if the PR description about the change in volume for the plenum makes sense to me just yet but I think these are two separate changes in a single PR that may be independent?

Issue #1018, opened 2022-12-11
reactor indexing is not same order as iteration


onufer:
```python
for i,a in enumerate(o.r.core):
    if someFunctionAssemblyMatches:
        break
assert o.r.core[i] == a  # this is false
```

I would expect the index order to be the same as iteration order but it is not.

another desciption if the top is not clear

```python
for a in o.r.core:
    print(a)
    break
assert o.r.core[0] == a  # this is false
```

john-science:
Well, let's see. The reason _how_ it is that was is easy to understand.  For instance, when you do `core[i]`, you call this code:

https://github.com/terrapower/armi/blob/23bb1c3cd2b1322dafc4f025d7f9c5dbf11b3665/armi/reactor/composites.py#L2656-L2657

That retrieves the `Assembly` from the `Reactor` by the order it was added to the `Reactor`.  We have a _lot_ of downstream code that people seem to love that depends on ARMI physical objects having an order determined by the order they were added to the `Reactor`. For better or worse, people seem to demand that as a feature.  

But when you did `for i, a in enumerate(core)`, you called this method on `class Core`:

https://github.com/terrapower/armi/blob/23bb1c3cd2b1322dafc4f025d7f9c5dbf11b3665/armi/reactor/reactors.py#L243-L251

Now, from the comment we can clearly see this was meant to always iterate over the highest-level objects in a `Core` (in this case probably Aseemblies) in the same order.  Without this method, we would iterate through the children in random order every time:

https://github.com/terrapower/armi/blob/23bb1c3cd2b1322dafc4f025d7f9c5dbf11b3665/armi/reactor/composites.py#L2662-L2663

Now, perhaps @ntouran can shed light on why we end up where we are.  But I can definitely say a couple of things:

1. We are here on purpose. (Take that for what it's worth.)
2. Changing this would cause a lot of rewrites and bug fixes downstream.e

IF we are to go about changing this situation, I would say that `Core.__iter__()` returning things in a sorted order is definitely the way to go, and we should change `Composite.__iter__()` to do the same.  In this solution, I would also want to change `Composite.__getitem__()` to get an item by it's position in the SORTER order, not that arbitrary order the `Core` was built in.

Having said _all_ that, I don't think we should close this ticket just yet, I think it would have pretty broad impact on ARMI and the ARMI ecosystem; this isn't just a small PR I can make without breaking people's plugins.

Thoughts? Opinions?

onufer:
[...]

PR #1280, opened 2023-05-30
Unifying the Reactor sorting

john-science:
## Description

In #1018, it was noted that if you iterate through the parts of a Reactor using different helper methods, you get the reactor parts in different order. 

This is obviously not idea. But it was a side effect of an early attempt at sorting the Reactor pieces by over-riding the `__iter__()` method in the `Core` object:

https://github.com/terrapower/armi/blob/23bb1c3cd2b1322dafc4f025d7f9c5dbf11b3665/armi/reactor/reactors.py#L243-L251

I _like_ the idea of sorting the Reactor Composites (parts), but I don't like the inconsistency of iteration order that arrises.

My first idea was to force the Reactor to always be fully sorted. This seems ideal. But that leads to some REAL headaches every time someone tries to `Reactor.add()` or `Assembly.add()`.  Aside from slowing down the code, it means if you do a `Block.add()`, you need to reach up to the entire Reactor to re-sort it. And that kind of backwards code design leads to impenetrable nightmares of code. 

The solution given here means:

1. In 99% of cases, the Reactor will be fully sorted.
2. At any time, the user / developer can fully sort the Reactor (or any sub-part) with `x.sort()`.

---

## Checklist

### DIFF
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -47,10 +47,10 @@ from armi.reactor import geometry
 from armi.reactor import grids
 from armi.reactor import parameters
 from armi.reactor import reactorParameters
-from armi.reactor import systemLayoutInput
 from armi.reactor import zones
 from armi.reactor.converters.axialExpansionChanger import AxialExpansionChanger
 from armi.reactor.flags import Flags
+from armi.reactor.systemLayoutInput import SystemLayoutInput
 from armi.settings.fwSettings.globalSettings import CONF_MATERIAL_NAMESPACE_ORDER
 from armi.utils import createFormattedStrWithDelimiter, units
 from armi.utils import directoryChangers
@@ -81,7 +81,7 @@ class Reactor(composites.Composite):
         self.blueprints = blueprints
 
     def __getstate__(self):
-        r"""applies a settings and parent to the reactor and components."""
+        """Applies a settings and parent to the reactor and components."""
         state = composites.Composite.__getstate__(self)
         state["o"] = None
         return state
@@ -110,9 +110,19 @@ class Reactor(composites.Composite):
             self.core = cores[0]
 
 
-def loadFromCs(cs):
+def loadFromCs(cs) -> Reactor:
     """
     Load a Reactor based on the input settings.
+
+    Parameters
+    ----------
+    cs: CaseSettings
+        A relevant settings object
+
+    Returns
+    -------
+    Reactor
+        Reactor loaded from settings file
     """
     from armi.reactor import blueprints
 
@@ -120,7 +130,7 @@ def loadFromCs(cs):
     return factory(cs, bp)
 
 
-def factory(cs, bp, geom: Optional[systemLayoutInput.SystemLayoutInput] = None):
+def factory(cs, bp, geom: Optional[SystemLayoutInput] = None) -> Reactor:
     """Build a reactor from input settings, blueprints and geometry."""
     from armi.reactor import blueprints
 
@@ -147,6 +157,8 @@ def factory(cs, bp, geom: Optional[systemLayoutInput.SystemLayoutInput] = None):
                 structure.construct(cs, bp, r)
 
     runLog.debug("Reactor: {}".format(r))
+
+    r.sort()
     return r
 
 
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -164,10 +164,11 @@ def factory(cs, bp, geom: Optional[SystemLayoutInput] = None) -> Reactor:
         r.sort()
     else:
         runLog.warning(
-            "DeprecationWarning: This Reactor is not being sorted on DB load. "
+            "DeprecationWarning: This Reactor is not being sorted on blueprint read. "
             f"Due to the setting {CONF_SORT_REACTOR}, this Reactor is unsorted. "
             "But this feature is temporary and will be removed by 2024."
         )
+
     return r
 
 
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -52,6 +52,7 @@ from armi.reactor.converters.axialExpansionChanger import AxialExpansionChanger
 from armi.reactor.flags import Flags
 from armi.reactor.systemLayoutInput import SystemLayoutInput
 from armi.settings.fwSettings.globalSettings import CONF_MATERIAL_NAMESPACE_ORDER
+from armi.settings.fwSettings.globalSettings import CONF_SORT_REACTOR
 from armi.utils import createFormattedStrWithDelimiter, units
 from armi.utils import directoryChangers
 from armi.utils.iterables import Sequence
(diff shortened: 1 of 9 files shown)


john-science:
The unit tests pass on my laptop.  I'm working on it.

john-science:
It turns out this PR didn't break anything: the `main` branch broke over the weekend.

(No commits were made over the weekend, so there was a change in one of our YAML dependencies. I'm on it.)

armi/reactor/tests/test_reactors.py
```diff
@@ -211,6 +212,7 @@ def loadTestReactor(
         # protocol=2 allows for classes with __slots__ but not __getstate__ to be pickled
         TEST_REACTOR = cPickle.dumps((o, o.r, assemblies.getAssemNum()), protocol=2)
 
+    o.r.sort()
```
albeanth, line 215: Is this needed? `r.sort()` gets called in the `isPickledReaactor` above and via `reactors.loadFromCs(cs)` so I'm not sure that it is?

keckler (approved):
Seems almost too easy, but I think it looks correct for what it's trying to accomplish.

john-science:
@opotowsky My only "concern" here is that it is _possible_ that any downstream repo could technically be explicitly testing some part of the reactor based on its position in iteration.  So, if someone wrote such a test, it might break (depending on which of the old iteration methods they use).

There's a lot of "ifs" there, but it's possible.

albeanth (commented):
> Aside from slowing down the code, it means if you do a Block.add(), you need to reach up to the entire Reactor to re-sort it. And that kind of backwards code design leads to impenetrable nightmares of code.

So if someone calls `a.add()` without calling `r.sort()` afterwards, then the reactor will be left in an unsorted state, right? If so, then we would need something to account for fuel shuffling over in armi/physics/fuelCycle/fuelHandlers.py, right?

albeanth:
>  it is possible that any downstream repo could technically be explicitly testing some part of the reactor based on its position in iteration.

Have you run the "big" downstream tests to double check? I don't _think_ this should be a problem for any of our internal testing, but you never know what you might uncover....

armi/reactor/tests/test_reactors.py
john-science: I believe it _is_ necessary, yes.  Because just above, we are manually adding assemblies one-at-a-time to the `Reactor`:

https://github.com/terrapower/armi/blob/996341f33f9629796e959f7110e8a676d31eb648/armi/reactor/tests/test_reactors.py#L204-L206

And those are not necessarily added in their geometric order, right?

opotowsky:
> @opotowsky My only "concern" here is that it is _possible_ that any downstream repo could technically be explicitly testing some part of the reactor based on its position in iteration. So, if someone wrote such a test, it might break (depending on which of the old iteration methods they use).
> 
> There's a lot of "ifs" there, but it's possible.

I will look into this!

armi/reactor/tests/test_reactors.py
albeanth: That's true. They are going into the spent fuel pool though, not the core. If we're interested in ordering the sfp only, should/could the call be `o.r.core.sfp.sort()`? That way we aren't sorting through the already sorted core too (if I understand right).

onufer:
our 1D slab cases do depend on order of components being consistent with blueprints. this as an undocumented feature and is not preserved after a DB load. The components like need an addition optional param for preserving order. For this reason i recommend making a setting `sortReactor` that defaults to true so that these cases can still function while we develop a mitigation plan.

the `sortReactor` setting should give a deprecation warning to indicate its just to allow for the transition.

albeanth:
Converting to draft while we figure out downstream merge order dependencies for a downstream Application.

john-science:
@keckler @albeanth The PR has doubled in size, but it is ready.  We just had to add a temporary setting so people could turn the setting off.

It is REAL easy to add totally unused settings into ARMI, or add settings that _become_ unused over time.  (Same with Parameters.)  And all this unused / untested code becomes a real weight to the project (in that they take a huge portion of my time to support).  So I will be removing this Setting by 2024-01-01 at the latest.

armi/reactor/reactors.py
```diff
@@ -158,7 +159,15 @@ def factory(cs, bp, geom: Optional[SystemLayoutInput] = None) -> Reactor:
...
-    r.sort()
+    # return a Reactor object
+    if cs[CONF_SORT_REACTOR]:
+        r.sort()
+    else:
+        runLog.warning(
+            "DeprecationWarning: This Reactor is not being sorted on DB load. "
+            f"Due to the setting {CONF_SORT_REACTOR}, this Reactor is unsorted. "
+            "But this feature is temporary and will be removed by 2024."
+        )
```
albeanth, line 170: A copy paste error here I think. Should be "DeprecationWarning: This reactor is not being sorted on reactor construction due to the setting..."

armi/reactor/tests/test_reactors.py
```diff
@@ -252,6 +253,25 @@ def setUp(self):
...
+        r1 = reactors.loadFromCs(cs)
+
+        # the reactor / core should be the same size
+        self.assertEqual(len(r0), len(r1))
+        self.assertEqual(len(r0.core), len(r1.core))
+
+        # the reactor / core should be in a different order
+        a0 = [a.name for a in r0.core]
+        a1 = [a.name for a in r1.core]
+        self.assertNotEqual(a0, a1)
```
albeanth, line 274: Thoughts on including this assertion in the other test over in test_database3.py as well?

albeanth (changes requested):
Just a couple minor comments. 

Want to update the release notes since this has a new setting? Even if it is temporary, it might be around for ~6 months

armi/reactor/tests/test_reactors.py
john-science: It wouldn't work the same. The assemblies in THIS test file are added to the reactor in order.  Thus, I would have to change the tests here to have them create a Reactor in a non-sorted order by default.  But there are plugins that import these tests, and I'm _really_ trying to keep this PR from requiring PRs in downstream plugins. (Especially for this "Setting" bonus feature that I'm just going to want to rip our in a few months.)

armi/reactor/reactors.py
john-science: Ah, right you are. 1 sec.

john-science: Fixed! Thanks!

armi/reactor/tests/test_reactors.py
albeanth: >  I'm really trying to keep this PR from requiring PRs in downstream plugins.

Great call.

albeanth (approved):

albeanth:
Converting to draft while we figure out merge order dependencies for a downstream Application. Thanks for understanding @john-science. I think we should be able to get this in on Monday!

armi/reactor/tests/test_reactors.py
john-science: I understand it's not desperately important, but this test reactor is used in a LOT of places, and I want to keep the entire `Reactor` sorted.  I mean, "it's just the Spent Fuel Pool" is probably fair, but somewhere out there, someone cares about the SFP.

PR #1282, opened 2023-05-30
Apply decusping to individual assemblies.

mgjarrett:
## Description
---

Grouping assemblies by type does not always work because assemblies of a single type can vary significantly in height over their lifetime in a reactor.
 
The particular example that motivates this bug fix is axial expansion caused by irradiation-driven swelling.

## Checklist


mgjarrett:
I believe the coverage decrease is only because I deleted covered lines. No new uncovered lines were introduced, and unit tests were enhanced to provide more coverage of the new feature.

armi/reactor/converters/uniformMesh.py
```diff
@@ -304,29 +302,19 @@ def _getFilteredFuelTopAndBottom(self):
...
-            for assemFlags in assemblyTypeFlags
+            a.getBlocks(Flags.FUEL)[-1].p.ztop
+            for a in self._sourceReactor.core.getAssemblies(Flags.FUEL)
         ]
         filteredBottoms = self._filterMesh(
             fuelBottoms,
             self.minimumMeshSize,
             [min(fuelBottoms)],
             preference="bottom",
-            warn=True,
```
keckler, line 329: Why do we want to remove this `warn=True` from here, but not from the determination of `filteredTops` below?

armi/reactor/converters/uniformMesh.py
```diff
@@ -185,11 +185,9 @@ def _decuspAxialMesh(self):
...
-            filteredBottomControl,
-            filteredTopControl,
+            materialBottoms,
+            materialTops,
         ) = self._getFilteredControlTopAndBottom(filteredBottomFuel, filteredTopFuel)
-        materialBottoms = filteredBottomFuel + filteredBottomControl
-        materialTops = filteredTopFuel + filteredTopControl
 
         # combine the bottoms and tops into one list with bottom preference
         allMatBounds = materialBottoms + materialTops
```
keckler, line 193: Am I correct in understanding that this list now only includes bottoms and tops from the control assemblies? I can't quite tell. 

If that understanding is not correct, then I don't see why it used to be necessary to combine the lists, but now is not.

keckler (commented):
There are a couple questions that'd I'd like to get clarified before I approve this. 

But nominally I think it _looks_ fine.

Is there a way that you can show me how you are convinced that this is solving the problem that you're trying to fix?

And finally, are there performance or other issues that this will now cause that were leading us to not just go this route in the first place?

armi/reactor/converters/uniformMesh.py
mgjarrett: It wasn't necessary before. Maybe we should change the name of `getFilteredControlTopAndBottom`. It returns lists that include whatever was already in `filteredBottomFuel` and `filteredTopFuel`, respectively.

mgjarrett:
> Is there a way that you can show me how you are convinced that this is solving the problem that you're trying to fix?

I will send you some evidence.

> And finally, are there performance or other issues that this will now cause that were leading us to not just go this route in the first place?

I don't think there's a significant performance hit, but it does require paring down from a much longer initial list of material boundaries. The `meshList` length here is now going to be determined by the total number of assemblies instead of unique assembly types. In most cases all but 1 or 2 of them will eventually get `pop`'ed.

https://github.com/terrapower/armi/blob/320778dc684977b1092e134b19dcf60de530ded9/armi/reactor/converters/uniformMesh.py#L266-L294

armi/reactor/converters/uniformMesh.py
mgjarrett: We had an offline conversation about consolidating and simplifying into a single `_getFilteredMeshTopAndBottom` method.

mgjarrett:
The coverage decrease is again from removing previously covered lines. All of the new lines added are covered: 
image

armi/reactor/converters/uniformMesh.py
```diff
@@ -293,82 +292,47 @@ def _filterMesh(
...
             meshList.pop(removeIndex)
 
-    def _getFilteredFuelTopAndBottom(self):
+    def _getFilteredMeshTopAndBottom(self, flags, bottoms=None, tops=None):
         """
         Get the bottom and top boundaries of fuel assemblies and filter them based on the ``minimumMeshSize``.
 
+        Parameters
+        ----------
+        flags : armi.reactor.flags.Flags
```
keckler, line 301: Not super important since this is a private method, but could be helpful to specify if these input flags will be used exactly or greedily.

keckler (approved):

armi/reactor/converters/uniformMesh.py
```diff
@@ -293,82 +292,47 @@ def _filterMesh(
...
+            The assembly and block flags for which to preserve material boundaries
```
mgjarrett, line 302:
```suggestion
            The assembly and block flags for which to preserve material boundaries
            ``getAssemblies()`` and ``getBlocks()`` are both called with the default, ``exact=False``
```

PR #1283, opened 2023-05-30
Fixed float decimals v3

jyang-TP:
## Description

A RuntimeError was identified when generating MCNP inputs at BOEC/MOEC/EOEC for a downstream application. Specifically, in getting the top surface number in the global dictionary. 

image
image

The issue was caused by the float decimals.  MCNP method `getSurfNum` returns a surface number by the surface type and the associated value with 10 decimals (`units.FLOAT_DIMENSION_DECIMALS`).

This PR aims to resolve the float decimals issue.  This RP has also updated the global parameter FLOAT_DIMENSION_DECIMALS to 8 digits in order to resolve the conflict with #1225.


albeanth:
@jyang-TP it looks like all of the proposed changes here are also within #1268. That said, can you/we close #1268 and just use this PR?

armi/reactor/converters/meshConverters.py
```diff
@@ -24,7 +24,7 @@
 
 from armi import runLog
 from armi.reactor import grids
-
+from armi.utils import units
```
keckler, line 27: There shouldn't be a blank line above this.

armi/utils/units.py
```diff
@@ -80,7 +80,7 @@
 MIN_FUEL_HM_MOLES_PER_CC = 1e-10
 
 # More than 10 decimals can create floating point comparison problems in MCNP and DIF3D
-FLOAT_DIMENSION_DECIMALS = 10
+FLOAT_DIMENSION_DECIMALS = 8
 EFFECTIVELY_ZERO = 10.0 ** (-1 * FLOAT_DIMENSION_DECIMALS)
```
keckler, line 84: This `EFFECTIVELY_ZERO` value appears to be used in one of our downstream plugins. The value here will change, but I don't expect the change to have any practical impact.

keckler (approved):

jyang-TP:
@albeanth Thanks.  I will close PR #1268.

albeanth (changes requested):
@jyang-TP, please run black on your files changed to get CI to pass. also, please update the release notes. If you need help with either of these tasks, feel free to reach out here or offline.

armi/utils/units.py
jyang-TP: Agree. The `EFFECTIVELY_ZERO` is only used in downstream MCNP plugin.

albeanth (approved):

PR #1291, opened 2023-06-02
Cleaning up TODOs

john-science:
## Description

ARMI has like 50 `TODO`s, which I want to remove and make tickets (or just fix).  But since we have a code freeze, that will have to wait.  

We also had like 3 `FIXME`s and a few `XXX`s.   But some of those were meant to be `TODO`s, and some were meant to be `NOTE`s.  Since even _I_ found it confusing, I am just replacing all `FIXME` and `XXX` with `NOTE` or `TODO`, and putting it in the docs that we don't allow `FIXME` or `XXX` any more.

I am just aiming to make the project more clear.

---

## Checklist

### DIFF
--- a/armi/utils/tests/test_triangle.py
+++ b/armi/utils/tests/test_triangle.py
@@ -11,8 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-r""" Test the basic triangle math.
-"""
+""" Test the basic triangle math."""
 # pylint: disable=missing-function-docstring,missing-class-docstring,abstract-method,protected-access,unused-variable
 import unittest
 from armi.utils import triangle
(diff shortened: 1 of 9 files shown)


armi/utils/tests/test_triangle.py
```diff
@@ -88,7 +87,9 @@ def test_checkIfPointIsInTriangle(self):
         y3 = 0.376
         xP = 0.0
         yP = 0.17
-        # FIXME: generalTriangleInOrOut = triangle.checkIfPointIsInTriangle(xT1, yT1, xT2, yT2, xT3, yT3, xP, yP)
+        generalTriangleInOrOut = triangle.checkIfPointIsInTriangle(
+            xT1, yT1, xT2, yT2, xT3, yT3, xP, yP
+        )
```
opotowsky, line 92: This cracks me up

opotowsky: hashtag fixed

opotowsky (approved):
Easy little fixxer upper!

armi/utils/tests/test_triangle.py
john-science: lol

Yeah, my guess is the concerning bug was fixed years ago, probably right after this test line was commented out.

alexhjames:
@john-science @opotowsky Maybe I missed it in the communications but this type of change would be nice to socialize a bit more. Oxbow also points to this pylintrc file because we stylistically try to maintain similarity to ARMI and also the whole pylint license issue. This change caused all of the FIXMEs in Oxbow to disappear in pylint and took a little bit to track down (I thought it was my pylint setup at first, not a PR change)

opotowsky:
Oh I had no idea this could have any downstream impact!!

There is [an upcoming PR](https://github.com/terrapower/armi/pull/1296) that will remove `pylint` in favor of `ruff`. Would you like us to hold off? Can you copy the ARMI pylintrc file (minus this change) for your use?

PR #1298, opened 2023-06-09
undo setting ndense to 0 for xs gen; move feature to xs gen

onufer:
## Description

Undo #1257 since it impacts memory and database. Add features to lattice physics

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/fissionProductModel/fissionProductModel.py
+++ b/armi/physics/neutronics/fissionProductModel/fissionProductModel.py
@@ -161,17 +161,6 @@ class FissionProductModel(interfaces.Interface):
         else:
             self.setAllBlockLFPs()
 
-    def _getComponentToInitDensities(self, b):
-        """Get the component to initialize ndens keys on."""
-        # add the isotopics to the smallest solid since that is usually the most "interesting"
-        # sorted() calls getBoundingCircleOuterDiameter under the hood
-        solidsOrderedBySize = sorted(c for c in b if c.containsSolidMaterial())
-        if solidsOrderedBySize:
-            return solidsOrderedBySize[0]
-        else:
-            # no solids, so just add to smallest component
-            return sorted(c for c in b)[0]
-
     def setAllComponentFissionProducts(self):
         """
         Initialize all nuclides for each ``DEPLETABLE`` component in the core.
--- a/armi/physics/neutronics/fissionProductModel/fissionProductModel.py
+++ b/armi/physics/neutronics/fissionProductModel/fissionProductModel.py
@@ -174,7 +174,7 @@ class FissionProductModel(interfaces.Interface):
 
     def setAllComponentFissionProducts(self):
         """
-        Initialize all nuclides for each ``DEPLETABLE`` component in the core, or all blocks if detailedAxialExpansion is enabled.
+        Initialize all nuclides for each ``DEPLETABLE`` component in the core.
 
         Notes
         -----
--- a/armi/physics/neutronics/fissionProductModel/fissionProductModel.py
+++ b/armi/physics/neutronics/fissionProductModel/fissionProductModel.py
@@ -104,7 +104,7 @@ from armi.physics.neutronics.fissionProductModel.fissionProductModelSettings imp
     CONF_FP_MODEL,
     CONF_MAKE_ALL_BLOCK_LFPS_INDEPENDENT,
 )
-from armi.settings.fwSettings.globalSettings import CONF_DETAILED_AXIAL_EXPANSION
+
 
 NUM_FISSION_PRODUCTS_PER_LFP = 2.0
 
@@ -125,10 +125,6 @@ class FissionProductModel(interfaces.Interface):
     def __init__(self, r, cs):
         interfaces.Interface.__init__(self, r, cs)
         self._globalLFPs = lumpedFissionProduct.lumpedFissionProductFactory(self.cs)
-        # If detailed axial expansion is active, mapping between blocks occurs on uniform mesh
-        # and this can cause blocks to have isotopes that they don't have cross sections for.
-        # Fix this by adding all isotopes to all blocks so they are present it lattice physics.
-        self.allBlocksNeedAllNucs = self.cs[CONF_DETAILED_AXIAL_EXPANSION]
 
     @property
     def _explicitFissionProducts(self):
@@ -156,10 +152,7 @@ class FissionProductModel(interfaces.Interface):
         products be consistent across all blocks, even if fission products are
         not generated when the block is depleted.
         """
-        if self.getInterface("mcnp") or self.allBlocksNeedAllNucs:
-            return None
-        else:
-            return Flags.FUEL
+        return None if self.getInterface("mcnp") is not None else Flags.FUEL
 
     def interactBOL(self):
         interfaces.Interface.interactBOL(self)
@@ -187,12 +180,7 @@ class FissionProductModel(interfaces.Interface):
         -----
         This should be called when explicit fission product modeling is enabled to
         ensure that all isotopes are initialized on the depletable components within
-        the reactor data model so that cross sections are created during depletion.
-
-        When detailedAxialExpansion is also enabled, all regions will have fission/activation
-        products added to avoid missing cross sections during mesh conversion (since converted
-        blocks only have one xsID but may have isotopes from multiple blocks with different IDs.)
-        Setting density to zero here enables small number densities is XS generation.
+        the reactor data model so that there is some density as a starting point.
 
         When explicit fission products are enabled and the user has not already included
         all fission products in the blueprints (in ``nuclideFlags``), the ``fpModelLibrary`` setting is used
@@ -210,28 +198,15 @@ class FissionProductModel(interfaces.Interface):
         """
         for b in self.r.core.getBlocks(includeAll=True):
             b.setLumpedFissionProducts(None)
-            self._initializeIsoDensities(b)
-
-    def _initializeIsoDensities(self, b):
-        """
-        Initialize isotope densities to include all on depletable components.
-
-        Notes
-        -----
-        In non-uniform reactor cases all blocks will get all isotopes so that the
-        global flux solve does not have any missing cross sections.
-        """
-        compsToAddIso = b.getComponents(Flags.DEPLETABLE)
-        if self.allBlocksNeedAllNucs and not compsToAddIso:
-            compsToAddIso = [self._getComponentToInitDensities(b)]
-        for c in compsToAddIso:
-            # Add all isotopes in problem at 0.0 density
-            updatedNDens = c.getNumberDensities()
-            for nuc in self.r.blueprints.allNuclidesInProblem:
-                if nuc in updatedNDens:
-                    continue
-                updatedNDens[nuc] = 0.0
-            c.updateNumberDensities(updatedNDens)
+            for c in b.getComponents(Flags.DEPLETABLE):
+                # Add all isotopes in problem at 0.0 density
+                updatedNDens = c.getNumberDensities()
+                # self.r.blueprints.allNuclidesInProblem contains ~everything in ENDF if _explicitFissionProducts
+                for nuc in self.r.blueprints.allNuclidesInProblem:
+                    if nuc in updatedNDens:
+                        continue
+                    updatedNDens[nuc] = 0.0
+                c.updateNumberDensities(updatedNDens)
 
     def setAllBlockLFPs(self):
         """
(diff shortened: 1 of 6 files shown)


armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -210,28 +198,15 @@ def setAllComponentFissionProducts(self):
...
-            compsToAddIso = [self._getComponentToInitDensities(b)]
-        for c in compsToAddIso:
-            # Add all isotopes in problem at 0.0 density
-            updatedNDens = c.getNumberDensities()
-            for nuc in self.r.blueprints.allNuclidesInProblem:
-                if nuc in updatedNDens:
-                    continue
-                updatedNDens[nuc] = 0.0
-            c.updateNumberDensities(updatedNDens)
+            for c in b.getComponents(Flags.DEPLETABLE):
```
onufer, line 201: this is moving it exactly back to what is was before #1257

armi/physics/neutronics/latticePhysics/latticePhysicsWriter.py
```diff
@@ -271,7 +280,7 @@ def _getAllNuclidesByCategory(self, component=None):
             nucCategory = ""
             # Remove nuclides from detailed fission product dictionary if they are a part of the core materials
             # (e.g., Zr in the U10Zr which is at fuel temperature and Mo in HT9 which is at structure temp)
-            if nuc in dfpDensities:
+            if nuc in list(dfpDensities):
```
onufer, line 283: del-ing a couple lines before end generally dont want to del as you iterate over keys

armi/reactor/blueprints/__init__.py
```diff
@@ -393,69 +394,77 @@ def _resolveNuclides(self, cs):
 
         actives = set()
         inerts = set()
-        undefBurnChainActiveNuclides = set()
         if self.nuclideFlags is None:
             self.nuclideFlags = isotopicOptions.genDefaultNucFlags()
-
+        originalFlagKeys = set(self.nuclideFlags)
         isotopicOptions.autoUpdateNuclideFlags(cs, self.nuclideFlags)
```
onufer, line 400: There might be an opportunity for a refeactor where we auto-update after the nuclide expansion code below (since the auto-updated nuclides almost certainly dont need expansion) but that would be a bit bigger of a change and could have negative unforeseen consequences . happy to try that direction though if someone wants to review it.

The upside is you wouldn't need to store `originalFlagKeys` and do all that testing as you are doing the expansion. you would.

1.  expand original flags
2.  store flags in nucsToForceInXsGen
3.  call autoUpdate

armi/reactor/blueprints/__init__.py
```diff
@@ -393,69 +394,77 @@ def _resolveNuclides(self, cs):
...
+                if wasInOriginalFlags:
+                    nucsToForceInXsGen.add(elemental.name)
                 continue
 
             self.elementsToExpand.append(elemental.element)
 
             if (
                 elemental.name in self.nuclideFlags
-                and self.nuclideFlags[elemental.name].expandTo
+                and self.nuclideFlags[elemental.name.symbol].expandTo
```
onufer, line 456: we know its an elemental already, and for elemental, name and symbol are same. rest of this uses symbol so symbol for consistency. otherwise the  developer has no  idea why its different here. (i double checked name== symbol for all ele)

armi/reactor/blueprints/__init__.py
onufer: @jakehader, do you know if there is a problem moving `autoUpdateNuclideFlags` (or some tweaked version of it) to after all the ele expansion code?

armi/reactor/blueprints/isotopicOptions.py
```diff
@@ -107,16 +107,15 @@ def __repr__(self):
             self.nuclideName, self.burn, self.xs
         )
 
-    def fileAsActiveOrInert(
-        self, activeSet, inertSet, undefinedBurnChainActiveNuclides
```
onufer, line 111: arg was unused and explicit returning seemed more clear

armi/physics/neutronics/latticePhysics/latticePhysicsWriter.py
```diff
@@ -241,11 +242,19 @@ def _getAllNuclidesByCategory(self, component=None):
...
+            # and this can cause blocks to have isotopes that they don't have cross sections for.
+            # Fix this by adding all isotopes so they are present it lattice physics.
+            if self.cs[CONF_DETAILED_AXIAL_EXPANSION]:
+                nuclides = self.r.blueprints.allNuclidesInProblem
+            else:
+                nuclides = sorted(objNuclides)
+        else:
+            nuclides = self.r.blueprints.allNuclidesInProblem
+
+        nuclides = nuclides.union(self.r.core.blueprints.nucsToForceInXsGen)
```
jakehader, line 256: I need some explanation of why this `nucsToForceInXsGen` is required still. I think this should be equivalent to just using `self.r.blueprints.allNuclidesInProblem`, so basically we would just have:

`nuclides = self.r.blueprints.allNuclidesInProblem` rather than having a branch for `nuclides = sorted(objNuclides)`, right?

armi/reactor/blueprints/__init__.py
jakehader: See comment below because we might be able to just remove all this new logic with a short change in the lattice physics writer. I could be wrong though.

armi/physics/neutronics/latticePhysics/latticePhysicsWriter.py
onufer: no its not possible with 1D control cross section generation, putting everything in the problem in there causes it to take way too long and use more memory than VDI has available. We need a way to disriminate between something the original problem request and `allNuclidesInProblem` which has 400+ isotopes from end 7.0

armi/reactor/blueprints/__init__.py
onufer: we cannot. see comment below. do you have any thoughts on the above question?

armi/physics/neutronics/latticePhysics/latticePhysicsWriter.py
mgjarrett: The iteration is over `for nucName, dens in zip(nuclides, numDensities):`, it's not iterating over `dfpDensities`?

This doesn't hurt but I don't see the need for it.

armi/physics/neutronics/fissionProductModel/fissionProductModel.py
```diff
@@ -210,28 +198,15 @@ def setAllComponentFissionProducts(self):
...
-        Initialize isotope densities to include all on depletable components.
-
-        Notes
-        -----
-        In non-uniform reactor cases all blocks will get all isotopes so that the
-        global flux solve does not have any missing cross sections.
-        """
-        compsToAddIso = b.getComponents(Flags.DEPLETABLE)
-        if self.allBlocksNeedAllNucs and not compsToAddIso:
-            compsToAddIso = [self._getComponentToInitDensities(b)]
```
mgjarrett, line 226: `_getComponentToInitDensities(b)` is no longer used

armi/physics/neutronics/latticePhysics/latticePhysicsWriter.py
```diff
@@ -241,11 +242,19 @@ def _getAllNuclidesByCategory(self, component=None):
...
-            else self.r.blueprints.allNuclidesInProblem
-        )
+        if self.explicitFissionProducts:
+            # If detailed axial expansion is active, mapping between blocks occurs on uniform mesh
+            # and this can cause blocks to have isotopes that they don't have cross sections for.
+            # Fix this by adding all isotopes so they are present it lattice physics.
+            if self.cs[CONF_DETAILED_AXIAL_EXPANSION]:
+                nuclides = self.r.blueprints.allNuclidesInProblem
+            else:
+                nuclides = sorted(objNuclides)
```
mgjarrett, line 252: I think we want to remove the `if` here and always use the `else`

Removing the dependency on `self.cs[CONF_DETAILED_AXIAL_EXPANSION]` for determining the set of nuclides to use is one of the purposes of this PR.

I think `objNuclides` will have the explicit fission products where we want in to (i.e., fuel) and not in other components.

armi/physics/neutronics/latticePhysics/latticePhysicsWriter.py
```diff
@@ -241,11 +242,19 @@ def _getAllNuclidesByCategory(self, component=None):
...
         # should be modeled.
-        nuclides = (
-            sorted(objNuclides)
-            if self.explicitFissionProducts
-            else self.r.blueprints.allNuclidesInProblem
-        )
+        if self.explicitFissionProducts:
+            # If detailed axial expansion is active, mapping between blocks occurs on uniform mesh
+            # and this can cause blocks to have isotopes that they don't have cross sections for.
+            # Fix this by adding all isotopes so they are present it lattice physics.
```
mgjarrett, line 248:
```suggestion
            # Fix this by adding all isotopes so they are present in lattice physics.
```

armi/reactor/blueprints/__init__.py
```diff
@@ -393,69 +394,77 @@ def _resolveNuclides(self, cs):
...
                 currentSet = inerts
-                inerts.remove(elemental.name)
             else:
                 # This was not specified in the nuclide flags at all.
                 # If a material with this in its composition is brought in
                 # it's nice from a user perspective to allow it.
                 # But current behavior is that all nuclides in problem
                 # must be declared up front.
+                if wasInOriginalFlags:
+                    nucsToForceInXsGen.add(elemental.name)
```
mgjarrett, line 449: I don't know if I follow how we get to line 449. If `wasInOriginalFlags == true`, it should be in `actives` or `inerts`, yes?

The line doesn't do any damage but I'm just curious whether it's reachable. Maybe there's a situation I haven't thought of.

armi/reactor/blueprints/__init__.py
mgjarrett: I'm confused by `elemental.name.symbol` vs. `elemental.element.symbol`

Could the usage in lines 456, 461, and 464 all be unified to the same attribute?

armi/physics/neutronics/latticePhysics/latticePhysicsWriter.py
onufer: too early in morning. thanks.

armi/physics/neutronics/latticePhysics/latticePhysicsWriter.py
onufer: this is needed when you have a depletable block next to non-depletable so that all can have all nuclides.

armi/reactor/blueprints/__init__.py
onufer: yes should be `elemental.element.symbol`

armi/reactor/blueprints/__init__.py
onufer: yeah not sure its reachable either. i can just delete. its not in xs if so so we shouldn't force.

armi/physics/neutronics/latticePhysics/latticePhysicsWriter.py
mgjarrett: So `if self.explicitFissionProducts` is true and `self.cs[CONF_DETAILED_AXIAL_EXPANSION]` is true, then `nuclides = self.r.blueprints.allNuclidesInProblem` which is going to include all of the detailed FPs.

So `nucsToForceInXsGen` isn't going to add any new isotopes to `nuclides` because it already has them all.

I thought the purpose of `nucsToForceInXsGen` was specifically for the case of `self.explicitFissionProducts == true` and `self.cs[CONF_DETAILED_AXIAL_EXPANSION] == true`, so that we could specify a limited set of nuclides to generate XS for with having to include all of the fission products.

mgjarrett: We talked offline; this makes sense to me in the context of the features from #1257 having been reverted in this PR.

Even if `self.r.blueprints.nucsToForceInXsGen` doesn't add any new nuclides here, it will be used later by a subclass of the `LatticePhysicsWriter`.

mgjarrett (approved):

opotowsky:
The lattice physics writer has a barely-test that covers it, so the small code cov drop is a result of the code in `latticePhysicsWriter.py` changing a little so 4 new lines are uncovered. Since this is mainly a feature reversion I don't want to hold up progress, and we need to get some real tests covering that module. Making an issue about it so I can merge this PR.

PR #1299, opened 2023-06-09
Improving HexBlock getWettedPerimeter calculation

john-science:
## Description

Improving HexBlock getWettedPerimeter calculation to include wire wrap

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -2211,12 +2211,27 @@ class HexBlock(Block):
                 6 * c.getDimension("ip") / math.sqrt(3) if c else 0.0
             )
 
+        # account for the wire warp in the wetted perimeter
+        try:
+            wire = self.getComponent(Flags.WIRE)
+            if wire is not None:
+                correctionFactor = numpy.hypot(
+                    1.0,
+                    math.pi
+                    * wire.getDimension("helixDiameter")
+                    / wire.getDimension("axialPitch"),
+                )
+            else:
+                correctionFactor = 1.0
+        except ValueError:
+            correctionFactor = 1.0
+
         # solid circle = od * pi
         # NOTE: since these are pin components, multiply by the number of pins
         wettedPinPerimeter = 0.0
         for c in wettedPinComponents:
             wettedPinPerimeter += c.getDimension("od") if c else 0.0
-        wettedPinPerimeter *= self.getNumPins() * math.pi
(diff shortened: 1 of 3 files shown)


armi/reactor/blocks.py
```diff
@@ -2211,12 +2211,27 @@ def getWettedPerimeter(self):
...
+        except ValueError:
+            correctionFactor = 1.0
+
         # solid circle = od * pi
         # NOTE: since these are pin components, multiply by the number of pins
         wettedPinPerimeter = 0.0
         for c in wettedPinComponents:
             wettedPinPerimeter += c.getDimension("od") if c else 0.0
-        wettedPinPerimeter *= self.getNumPins() * math.pi
+        wettedPinPerimeter *= self.getNumPins() * math.pi * correctionFactor
```
dlangewisch, line 2234: If I'm understanding this correctly, I don't think this is quite correct. This would calculate the total pin (plus wire) wetted perimeter as:
$$P_{wet} = N_{pins}\pi(D_{clad} + D_{wire})F_{corr}$$
where
$$F_{corr} = \sqrt{1 + \left(\frac{\pi(D_{clad} + D_{wire})}{H}\right)^2}$$
with $H$ being the wire axial pitch. But the correct expression should be
$$P_{wet} = N_{pins}\pi(D_{clad} + D_{wire} F_{corr})$$
Note that the correction factor only applies to wire diameter, _not_ the clad diameter. When the bundle is sliced by a plane orthogonal to the bundle axis, the pin intersections are circles but the wire intersections form ellipses, hence the correction factor.

dlangewisch: There may be better ways to go about it, but I would implement this correction using something like:
```python
wettedPinPerimeter = 0.0
for c in wettedPinComponents:
    if c is not None:
        correctionFactor = 1.0
        if isinstance(c, armi.reactor.components.complexShapes.Helix):
            correctionFactor = numpy.hypot(1.0, math.pi*c.getDimension("helixDiameter")/c.getDimension("axialPitch"))
        wettedPinPerimeter += c.getDimension("od")*correctionFactor
wettedPinPerimeter *= self.getNumPins() * math.pi
```

john-science: implemented

dlangewisch (approved):
This looks correct.

opotowsky (approved):
Approving purely from a code perspective, since Dustin approved the technical correctness for us :-)

PR #1305, opened 2023-06-13
Sort Assembly by ring position

onufer:
## Description

The sorted order of assemblies in a reactor doesn't make a whole lot of unitive sense to the user. the current ordering is dervied from the fact that its faster for python code to determine neighbors by completeIndicies. But complete indices are more of an implementation detail that is not exposed to users. When you print an assembly you get ring/position not indices. 

Here is a screenshot of old sort on left and new sort on right (for the test reactor). you can see how the numbers are increasing and make more sense, and the assembly types most folks care about (fuel and control) are at the front. It also makes spatial sense since its innermost to outermost.:

image

---

## Checklist


john-science:
1. This PR needs a description describing the change and why it's being made.
2. WHY are you making this change? The last 8 years ARMI has sorted Assemblies by X\Y position. This is a big change, and notably requires a lot more calculation.  WHY do you want this change?

john-science:
It is really important (and mandatory) that every ARMI PR comes with a good description:

https://github.com/terrapower/armi/blob/bc2763fe14fd4f9b8cb1b50b0c9b53dea25452c8/.github/pull_request_template.md?plain=1#L3

The goal of the description is to make it easy for reviewers to know: WHAT you are doing, and WHY you are doing it.

john-science:
@onufer  I will be really curious to hear WHY you want to make this change. Is it just because you like this way better, or is there a bug you're trying to solve before the release?  (IS this change part of the release?)


ANYWAY

My first thought on this change is that not all assemblies are in reactors that have rings. For instance, we can build a fun little example of Chicago Pile 1 with ARMI, and that has square assemblies that aren't in rings.

> Perhaps you meant to make these changes on `HexAssembly`?  It would seem a much stronger assumption that reactors with HEX-shaped assemblies have rings, right?

armi/reactor/assemblies.py
```diff
@@ -121,7 +121,7 @@ def __lt__(self, other):
         -----
         As with other ArmiObjects, Assemblies are sorted based on location. Assemblies
         are more permissive in the grid consistency checks to accomodate situations
-        where assemblies might be children of the same Core, but not in the same grid as
+        where assemblies might be children of the same Reactor, but not in the same grid as
```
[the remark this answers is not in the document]
onufer: I dont think so.  I thought SFP was a child of reactor not core. 

oh, wait, looking at the code SFP is a child of core... that is rather supprising to me. i will bring it up with others but fix for this diff.
image

armi/reactor/reactors.py
```diff
@@ -1149,7 +1149,7 @@ def genBlocksByLocName(self):
             block.getLocation(): block for block in self.getBlocks(includeAll=True)
         }
 
-    def getBlocks(self, bType=None, **kwargs):
+    def getBlocks(self, bType=None, exact=False, **kwargs):
```
onufer, line 1152: getFirstblock (line 1180) allowed exact, but this didnt...

onufer:
sorry this should be marked draft. i will work on the description. i hadn't assigned a reviewer yet so didnt mean to spin you up

i do believe all geometreis support getRingPos
image

armi/reactor/tests/test_reactors.py
```diff
@@ -812,7 +812,10 @@ def test_createAssemblyOfType(self):
         # basic creation
         aOld = self.r.core.getFirstAssembly(Flags.FUEL)
         aNew = self.r.core.createAssemblyOfType(aOld.getType(), cs=self.o.cs)
-        self.assertAlmostEqual(aOld.getMass(), aNew.getMass())
+        self.assertAlmostEqual(
```
john-science, line 815: Why is this change necessary?

onufer: the fresh assembly is not located in the core so its values are "full"

john-science (approved):

onufer:
this causes many internal test differences. i do think this is the long term best way to go, but abanoning for now in favor of #1320

PR #1311, opened 2023-06-14
Adding Boron 10 initial volume parameters to support b10 caps/cc. Derive Peak burnup from peak burnup rate.

onufer:
## What is the change?

Adding Boron 10 initial volume parameters to support b10 caps/cc. Derive Peak burnup from peak burnup rate.

This supports absorber performance evaluations and is part of our internal release.

---

## Checklist


armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -177,10 +178,11 @@ def _constructAssembly(self, cs, blueprint):
 
         # loop a second time because we needed all the blocks before choosing the
         # assembly class.
-        for axialIndex, block in enumerate(blocks):
-            b.p.assemNum = a.p.assemNum
```
onufer, line 181: the call in the line below does exactly this.

john-science:
I'm a little hesitant to have four parameters for one element in every `Block`.   My concern goes something like this:  

1. People see this feature.
2. Then someone creates 3 or 4 new parameters for U238, and puts them on every block.
3. Then later someone comes by and adds 3 or 4 parameters for U235 on every block.
4. Then a different person does the same for Plutonium.
5. Then Thorium.
6. Then we just stop using ARMI because we have thousands of by-nuclide parameters in ARMI and the system becomes unusable.

Is this crazy and alarmist?  ...Probably. 

So.  @onufer Can you talk me off a ledge?  What is so special about B10 that it gets 4 parameters on every `Block`, but not other nuclide/isotope/element will ever need such parameters?

Thanks!

armi/reactor/blocks.py
```diff
@@ -1650,6 +1650,43 @@ def getDuctOP(self):
...
+                    f"More than 1 component found  in {b.name} with different B10 material"
+                    f" or temperature, using {potentialC} params",
+                    single=True,
+                )
+            area += c.getArea(cold=True)
+
+        volume = area * self.getHeight()
+        if heightHot:
+            self.p.initialB10ComponentVol = (
+                volume / potentialC.getThermalExpansionFactor()
```
sammiller11235, line 1685: any concern about volume and linear expansion being 1D? From `component.getThermalExpansionFactor` it looks like this is a linear expansion factor. But volume should be factor^3. (e.g., if working with a unit cube, l, w, h = 1, vol=1. If the linear thermal expansion is 10%, then l,w,h = 1.1 and vol = 1.331).

jakehader: I think the idea here is that `getHeight` above returns the hot block height if this condition is hit. Since the area is cold here, the height is rescaled for this target component to be cold so that the initial volume is stored as cold. 

It might be clearer to have a comment here about this and possibly rescale height first based on this conditional and calculate volume afterwards.

```
if hotHeight:
    # add comment here...
    coldComponentHeight = self.getHeight() / c.getThermalExoansionFactor()
else:
    coldComponentHeight = self.gerHeight()

self.p.initialB10ComponentVol = area * coldComponentHeight
```

Do you agree @onufer / @sammiller11235 ?

onufer: sure i can tweak. we shouldn't be hitting hot height for any of our internal calcs anyway.

onufer:
moving most params to internal plugin

armi/reactor/blocks.py
```diff
@@ -1650,6 +1650,39 @@ def getDuctOP(self):
...
+        potentialC = None
+        for c in self:
+            if c.getNumberDensity("B10"):
+                potentialC = c
+                b10Comps.append(potentialC)
+                tHot = potentialC.temperatureInC
+                tCold = potentialC.inputTemperatureInC
+                mat = potentialC.material
+        if potentialC is None:
+            return
```
keckler:
```suggestion
        b10Comps = [c for c in self if c.getNumberDensity("B10")]]
        if not b10Comps:
            return
        
        potentialC = b10Comps[-1]
        tHot = potentialC.temperatureInC
        tCold = potentialC.inputTemperatureInC
        mat = potentialC.material
```

armi/reactor/blocks.py
```diff
@@ -1174,6 +1174,39 @@ def isPlenumPin(self, c):
             return True
         return False
 
+    def setB10VolParam(self, heightHot):
```
keckler, line 1177: Might be helpful to indicate in the docstring that this parameter is a boolean.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -903,24 +903,20 @@ def updateFluenceAndDpa(self, stepTimeInSeconds, blockList=None):
...
             # update detailed DPA based on dpa rate computed at LAST timestep.
-            dpaRateThisStep = b.p.detailedDpaRate
-            newDpaThisStep = dpaRateThisStep * stepTimeInSeconds
-            newDPAPeak = newDpaThisStep * burnupPeakingFactor
-            # track incremental increase for duct distortion interface (and eq)
-            b.p.newDPA = newDpaThisStep
-            b.p.newDPAPeak = newDPAPeak
+            # new incremental DPA increase for duct distortion interface (and eq)
+            b.p.newDPA = b.p.detailedDpaRate * stepTimeInSeconds
+            b.p.newDPAPeak = b.p.detailedDpaPeakRate * stepTimeInSeconds
```
keckler, line 914: Does this somehow account for the fact that the peak location could have changed? For instance if the assembly were shuffled/rotated, it is likely that the peak location is different, and thus just adding to the old value is not necessarily valid.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -946,7 +942,19 @@ def updateFluenceAndDpa(self, stepTimeInSeconds, blockList=None):
...
-            b.p.percentBuPeak = b.p.percentBu * burnupPeakingFactor
+            # b.p.percentBu is expected to have been updated elsewhere
+            peakRate = None
+            if b.p.buRatePeak:
+                # best case scenario, we have peak burnup rate
+                peakRate = b.p.percentBuPeak
+            elif b.p.buRate:
+                # use whatever peaking factor we can find
+                peakRate = b.p.buRate * self.getBurnupPeakingFactor(b)
+            if peakRate:
```
keckler, line 953:
```suggestion

            if peakRate:
```

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -946,7 +942,19 @@ def updateFluenceAndDpa(self, stepTimeInSeconds, blockList=None):
...
+                b.p.percentBuPeak = b.p.percentBuPeak + peakRate * stepTimeInSeconds
+            else:
+                # bad assumption.... assumes peaking is same at each position through shuffling/irradiation history...
+                b.p.percentBuPeak = b.p.percentBu * self.getBurnupPeakingFactor(b)
```
keckler, line 957: If we acknowlege this to be a bad assumption, perhaps we should add a runLog warning when it actually takes place?

Maybe there is a smart way to put a runLog statement in such a way that it doesn't get hit too many times.

keckler (approved):
I think this is fine overall. I made some suggestions throughout to make some bits a little more clear and to optimize.

As a general comment not really about your changes but related to this part of the code -- we have a lot of DPA block parameters, some of which feel redundant or unclear. This part of the codebase seems due for a cleaning in the future.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
onufer: I think fuel handler happens right before depletion on bol. this code is called by depletion. i don't know when another neutron flux solve would happen between the last nodes global flux and shuffling/depletion so i dont know how a different peak would be calculated. if it was, and peakFlux was updated, they could update `detailedDpaPeakRate` accordingly
image

Additionally, anytime the global flux interface is called after a shuffle or what nod 773 is called anyway so we should always be using the latest peaking from the latest flux solve. if you don't have a flux solve, how would you update peaking?

ultimately we always recalc flux after shuffling and before any subsequent (non-zero-time) depletion step.

i think at one time we transfer peak fluxes between blocks when we shuffled but that code was deleted, so this would no longer have value

armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -177,10 +178,11 @@ def _constructAssembly(self, cs, blueprint):
...
+        for axialIndex, b in enumerate(blocks):
             b.name = b.makeName(a.p.assemNum, axialIndex)
-            a.add(block)
+            a.add(b)
+            # set b10 volume cc since its a cold dim param
+            b.setB10VolParam(cs[CONF_INPUT_HEIGHTS_HOT])
```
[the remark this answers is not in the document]
onufer: no thats reasonable. moved.

armi/reactor/blocks.py
onufer: done

armi/physics/neutronics/globalFlux/globalFluxInterface.py
onufer: yep its been default behavior for a while but no reason to not add a warning.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
onufer: done and added more comments

armi/physics/neutronics/globalFlux/globalFluxInterface.py
onufer: I also am not fond of using a burnup peaking factor for dpa... This way if another interface calcs the dpa peaking (setting `detailedDpaPeakRate `) it can get integrated over time here

armi/reactor/blocks.py
```diff
@@ -1650,6 +1650,39 @@ def getDuctOP(self):
...
+                or mat != c.material
+            ):
+                runLog.warning(
+                    f"More than 1 component found  in {b.name} with different B10 material"
+                    f" or temperature, using {potentialC} params",
+                    single=True,
+                )
+            coldArea += c.getArea(cold=True)
+
+        coldFactor = potentialC.getThermalExpansionFactor() if heightHot else 1
```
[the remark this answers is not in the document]
onufer: i still think i want the warning because even if its cold but they are differnt temps, it means we will have mass flow of boron. ill clarify in the warning

armi/physics/neutronics/globalFlux/globalFluxInterface.py
onufer: single=true is the simple way....

keckler (approved):
Thanks for making the updates, approved!

keckler (approved):
Just a couple small suggestions.

Good find and fixes in the last few commits!

PR #1312, opened 2023-06-14
Fixing circular import in GlobalFluxInterface

john-science:
## What is the change?

We are moving the `CONF_X` string imports in `globalFluxInterface.py` from the module-level to the function level

This is being made because a downstream project was importing ARMI code and got some circular imports.  Also, making this change is 100% safe, so there isn't a strong reason not to make this change here, for future-proofing reasons.

---

## Checklist


opotowsky (approved):
Thanks for finding this fix!

PR #1313, opened 2023-06-14
Removing unnecessary unittest main functions

john-science:
## What is the change?

About half of the `test_whatever.py` files in ARMI had these empty placeholder `if __name__ == "__main__":` statements at the bottom.  I am removing these here.

As @opotowsky and I were discussing recently, these statements are completely unused and defunct.  Especially since we use `PyTest` in ARMI, and can run any single test via the commandline using something like:

```bash
pytest armi/physics/neutronics/latticePhysics/tests/test_latticeWriter.py::TestLatticePhysicsWriter::test_getDetailedFPDensitiesaile
```

And most people use an IDE for their development, where these `main()` statements are also completely bipassed.

I would rather remove these than have to keep explaining to people what defunct, historical use they once had for a small set of developers.

---

## Checklist


opotowsky (approved):
Easy one

PR #1318, opened 2023-06-16
Fixing broken docs try/except

john-science:
## What is the change?

This fixes a bug that was causing the docs build to fail.

In a recent (ruff) PR I moved a single try/except in the docs from bare to specific. But apparently, that try/except is hit several different ways while building the docs and needs to pass/fail with various errors during the doc build.  So I had to move it back to a bare except.

---

## Checklist


john-science:
Also, I added a bonus here.  Our `ruff.toml` file had a few invisible defaults. I am just calling them out here to make them more explicit.

opotowsky (approved):
glad you caught this so quickly! I did not think to test the docs build while reviewing your PR.

PR #1320, opened 2023-06-17
Add sortAssemsByRing to sort the reactor assemblies by ring

onufer:
## What is the change?
Fix bug in reactor sorting that didn't recursively sort
Add sortAssemsByRing method to sort assembly from inner to outer.

for ease of postprocessing. a full sort caused many internal unit test issues.
https://github.com/terrapower/armi/pull/1305

---

## Checklist


armi/reactor/reactors.py
```diff
@@ -272,7 +272,7 @@ def __iter__(self):
         --------
         getAssemblies()
         """
-        return iter(sorted(self._children))
+        return iter(self._children)
```
onufer, line 275: the sort would not matter with this sort here since it sorts by something else. i think with John's sorts, this should be good. not sorting itter will also guarantee list accessing and iteration will not be out of sync

armi/reactor/composites.py
```diff
@@ -2681,7 +2681,7 @@ def sort(self):
 
         # recursively sort the children below it.
         for c in self._children:
-            if issubclass(Composite, c.__class__):
+            if issubclass(c.__class__, Composite):
```
onufer, line 2684: this is a bug that caused no recursive sorting since its never true.

armi/reactor/tests/test_reactors.py
```diff
@@ -272,6 +270,25 @@ def test_factorySortSetting(self):
         a1 = [a.name for a in r1.core]
         self.assertNotEqual(a0, a1)
 
+    def test_sortChildren(self):
```
onufer, line 273: this test failed on main due to the `if issubclass(c.__class__, Composite):` issue

armi/reactor/tests/test_reactors.py
```diff
@@ -179,7 +179,6 @@ def loadTestReactor(
         assemblies.setAssemNumCounter(assemNum)
         settings.setMasterCs(o.cs)
         o.reattach(r, o.cs)
-        r.sort()
```
onufer, line 182: these are not needed since its sorted on initialization as the test below shows.

armi/tests/tutorials/data_model.ipynb
```diff
@@ -53,7 +53,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "o=armi.init(fName=\"anl-afci-177.yaml\");"
+    "o=armi.init(fName=\"anl-afci-177.yaml\");\n",
+    "o.r.core.sortAssemsByRing() # makes innermost assemblies appear first"
```
onufer, line 57: fixes broken unit test

opotowsky (commented):
For how difficult this was (mostly from the bug) the resulting code update is pretty simple. 

Assuming the downstream effects are OK, I think the PR looks good!

albeanth (approved):

opotowsky (approved):

PR #1327, opened 2023-06-26
Adding more ruff simplify rules

john-science:
## What is the change?

Adding the `ruff` simplify rules to the repo:

* **SIM2** - simplify code: complex equality statements
* **SIM3** - simplify code: yoda-conditions
* **SIM4** - simplify code: if-else-blocks
* **SIM9** - simplify code: dict-get-with-none-default

Ignoring the `ruff` simplify rule: `SIM118`. This checks against the `.keys()` method. But we overload that method in fun and unique ways that the linter just can't understand.

This is part of my on-going work to find the final set of `ruff` rules we want to use in ARMI.  Also, to figure out the sweet spot for what `ruff` rules we want to support.

---

## Checklist

### DIFF
--- a/armi/utils/plotting.py
+++ b/armi/utils/plotting.py
@@ -778,7 +778,7 @@ def plotAssemblyTypes(
     # Setup figure
     fig, ax = plt.subplots(figsize=(15, 15), dpi=300)
     for index, assem in enumerate(assems):
-        isLastAssem = True if index == (numAssems - 1) else False
+        isLastAssem = index == numAssems - 1
         (xBlockLoc, yBlockHeights, yBlockAxMesh) = _plotBlocksInAssembly(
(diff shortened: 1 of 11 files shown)


opotowsky:
yoda conditions 😆

armi/utils/plotting.py
```diff
@@ -778,7 +778,7 @@ def plotAssemblyTypes(
     # Setup figure
     fig, ax = plt.subplots(figsize=(15, 15), dpi=300)
     for index, assem in enumerate(assems):
-        isLastAssem = True if index == (numAssems - 1) else False
+        isLastAssem = index == numAssems - 1
```
albeanth, line 781: there's pythonic and then there's _pythonic_

dang lol

PR #1330, opened 2023-06-26
Expanding ruff application of docstring rules

john-science:
## What is the change?

I added blanket coverage to all `D2` and `D3` docstring rules in `ruff`.  

ALMOST all of these were fully automatable via `ruff --fix .`. 

This is part of the on-going work to get the ARMI `ruff.toml` up-to-date. But I am also doing it one (or two) rules at a time, so my reviewers can actually review all these changes.

---

## Checklist

### DIFF
--- a/armi/apps.py
+++ b/armi/apps.py
@@ -120,9 +120,7 @@ class App:
         return self._pm
 
     def getSettings(self) -> Dict[str, Setting]:
-        """
-        Return a dictionary containing all Settings defined by the framework and all plugins.
-        """
+        """Return a dictionary containing all Settings defined by the framework and all plugins."""
         # Start with framework settings
         settingDefs = {
             setting.name: setting for setting in fwSettings.getFrameworkSettings()
@@ -231,7 +229,7 @@ class App:
         return renames
 
     def registerUserPlugins(self, pluginPaths):
-        """
+        r"""
         Register additional plugins passed in by importable paths.
         These plugins may be provided e.g. by an application during startup
         based on user input.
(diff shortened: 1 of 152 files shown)


armi/apps.py
```diff
@@ -231,7 +229,7 @@ def getParamRenames(self) -> Dict[str, str]:
         return renames
 
     def registerUserPlugins(self, pluginPaths):
-        """
+        r"""
```
opotowsky, line 232: @john-science why is the `r` being added? I'm assuming the docs built fine before without it?

armi/materials/uraniumOxide.py
```diff
@@ -131,9 +131,7 @@ def applyInputParams(
         material.FuelMaterial.applyInputParams(self, *args, **kwargs)
 
     def setDefaultMassFracs(self) -> None:
-        r"""
-        UO2 mass fractions. Using Natural Uranium without U234.
-        """
+        r"""UO2 mass fractions. Using Natural Uranium without U234."""
```
opotowsky, line 134: Interesting that `ruff` adds `r` when needed but doesn't remove it when it's not.

Not asking you to comb through these, just making an observation :-)

opotowsky (approved):
Looks good to me!

armi/materials/uraniumOxide.py
john-science: So, I BELIEVE there is a different `ruff` rule for removing the unnecessary r's.  But that's just not part of this PR.

Soon!

opotowsky: <3

Issue #1325, opened 2023-06-23
SHA-1 hashes of files listed in `crossSectionControl`


keckler:
The stdout file currently contains SHA-1 hashes of most of the input files associated with a run. It looks something like this:
![image](https://github.com/terrapower/armi/assets/21959600/35960f06-8403-4f3d-b0c1-a86b425bb696)

I have noticed that the hashes for files that are in the `crossSectionControl` setting are not printed to this table. This means files like `ISOTXS`, `gamiso`, `pmatrx` don't get their hashes included.

We should add these files to the SHA-1 listing to improve traceability and repeatability of our calculations.

@Nebbychadnezzar Tagging since we talked about this offline.

Nebbychadnezzar:
Great find, Chris, and solid redaction work. :D

john-science:
Just as a priority-setter, @keckler , has this already been a problem for you / your team?

Or are you just pointing out something that could be improved.

Either way, it seems like an easy win.  Just asking.

keckler:
This ended up being a problem for us when we had to permanently store calculation results. Specifically, because relative file paths can become irrelevant when the files are moved from the original run location to the storage location, it is necessary to have a better system of relating the supporting files to the `stdout` to keep the traceability and repeatability of the run.

Hopefully that makes sense and adds the necessary context.

PR #1334, opened 2023-06-29
Adding XS files SHA1 hashes to welcome text

john-science:
## What is the change?

The SHA1 hashes have been added for the XS control files, in the welcome text.

This is part of #1325, and the idea is that we are recording the SHA1 hashes for all the other files referenced in the master settings files, but not these. So, adding them here would allow for greater reproducibility of ARMI-based simulations.

---

## Checklist


keckler (approved):
Thanks for this, and the robust test.

PR #1335, opened 2023-06-30
Automating the Material names

john-science:
## What is the change?

All materials in ARMI (or ARMI plugins) used to need a class attribute that defined their name.  But that name HAD to exactly match the name of the class.

That led to all kinds of silly problem, because of course it did.

So this PR automates that process.

BUT it also still allows developers to manually set the name of the material, if they want. So everyone's use cases should be covered.

We have run up against the problem of ARMI materials having the wrong `.name` attribute several times lately.  And we have had to go through various repos and manually fix things.  But this is a problem that can be solved by automation.

---

## Checklist

### DIFF
--- a/armi/materials/concrete.py
+++ b/armi/materials/concrete.py
@@ -24,8 +24,10 @@ from armi.materials.material import Material
 
 
 class Concrete(Material):
-    name = "Concrete"
-    """ http://jolissrch-inter.tokai-sc.jaea.go.jp/pdfdata/JAERI-Data-Code-98-004.pdf """
+    """Simple concreate material.
+
+    http://jolissrch-inter.tokai-sc.jaea.go.jp/pdfdata/JAERI-Data-Code-98-004.pdf
+    """
 
     def setDefaultMassFracs(self):
         self.setMassFrac("H", 0.023 / 2.302)
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -669,8 +669,6 @@ class Material:
 class Fluid(Material):
     """A material that fills its container. Could also be a gas."""
 
-    name = "Fluid"
-
     def getThermalExpansionDensityReduction(self, prevTempInC, newTempInC):
         """Return the factor required to update thermal expansion going from temperatureInC to temperatureInCNew."""
         rho0 = self.pseudoDensity(Tc=prevTempInC)
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -599,7 +599,7 @@ class Material:
         Tk : float, optional
             Temperature in Kelvin.
         Tc : float, optional
-            Temperature in degrees Celsius
+            Temperature in degrees Celsius.
 
         Returns
         -------
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -95,10 +95,22 @@ class Material:
 
     @property
     def name(self):
+        """Getter for the private name attribute of this Material."""
         return self._name
 
     @name.setter
     def name(self, nomen):
+        """Setter for the private name attribute of this Material.
+
+        Warning
+        -------
+        Some code in ARMI expects the "name" of a meterial matches its
+        class name. So you use this method at your own risk.
+
+        See Also
+        --------
+        armi.materials.resolveMaterialClassByName
+        """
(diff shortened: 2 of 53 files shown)


john-science:
@alexhjames Just FYI.

@sombrereau Just FYI, this will necessitate some (very small) changes in your downstream materials repo. All you really "need" to do is remove a line of code from each file.  I will show you a parallel PR to copy.

john-science:
The code coverage decreased because I removed lines of code.

armi/materials/material.py
```diff
@@ -88,15 +85,25 @@ def __init__(self):
...
     def __repr__(self):
-        return "<Material: {0}>".format(self.getName())
+        return f"<Material: {self._name}>"
+
+    @property
+    def name(self):
+        return self._name
+
+    @name.setter
+    def name(self, nomen):
```
alexhjames, line 101: If someone defines a custom name that isn't the class name, how do you intend for this to function with regards to `resolveMaterialClassByName()`?

armi/materials/material.py
```diff
@@ -584,14 +591,15 @@ def checkTempRange(self, minT, maxT, val, label=""):
...
+        """
+        Return heat capacity * density at a temperature.
+
         Parameters
         ----------
         Tk : float, optional
             Temperature in Kelvin.
         Tc : float, optional
-            Temperature in degrees Celsius.
+            Temperature in degrees Celsius
```
opotowsky, line 602: I've got a good eye for ruff-iness now that I've reviewed a few PRs. Why was the period removed here? Just curious, obvi not a PR blocker :-)

armi/materials/material.py
opotowsky: I second this question. Or maybe you could add a unit test for `resolveMaterialClassByName` that covers the custom name case?

armi/materials/material.py
john-science: huh.

Typo, I guess.

Fixing it.

armi/materials/material.py
alexhjames: Right now, every material class in armi can be instantiated by name. In the case of custom materials it is "Custom". 
image

If you assume, the user is going to rename the name on "Custom" materials to something useful, then sure it doesn't work for that right now. At least that is isolated to one material class that's an odd child in it's usage. By adding this method onto the base material class, you are encouraging people to rename any material class as an acceptable behavior. When that starts happening, then `resolveMaterialClassByName()` stops working for a larger number of classes. 
image

alexhjames: Maybe we could just shift this setter method to the custom material class from the base material class and call it good?

alexhjames: Yes, I was suggesting shifting it from the base material class like it is in this PR to the Custom material class (https://github.com/terrapower/armi/blob/main/armi/materials/custom.py). That would provide some contextual clues to a user that they shouldn't change the name of a non-custom material.

john-science: Okay, so here are some examples in the docs of `Custom` materials being used:

https://terrapower.github.io/armi/tutorials/walkthrough_lwr_inputs.html#setting-up-the-blueprints
https://terrapower.github.io/armi/user/inputs/blueprints.html#custom-isotopics

The important thing to note here is the name of every custom material is "Custom".  And then they are given specific isotopics.

I have checked with in-house TerraPower users of Custom materials, and this appears to be the only case.

As such, I believe the `resolveMaterialClassByName()` function does not need to be changed to work with this PR.

alexhjames (approved):

john-science:
The code coverage went down because I deleted many lines of code in this PR.

That's acceptable.

opotowsky (approved):
Seems like our questions are resolved so this one is good to go!

armi/materials/concrete.py
```diff
@@ -24,8 +24,10 @@
 
 
 class Concrete(Material):
-    name = "Concrete"
-    """ http://jolissrch-inter.tokai-sc.jaea.go.jp/pdfdata/JAERI-Data-Code-98-004.pdf """
+    """Simple concreate material.
```
dpham-materials, line 27: Minor typo here for "concrete"

PR #1336, opened 2023-06-30
Moving Spent Fuel pool off Core and onto Reactor

john-science:
## What is the change?

I moved the Spent Fuel Pool (SFP) object was on the `Core` instead of the `Reactor`.

I wanted to make the shift more cleanly than I did, but when you call `Core.removeAssembly()`, it wants to put the assembly in the SFP.  Which means the `Core` needs to at _least_ know the SFP is on the `Reactor`.  But fine.

For some reason the 

The SPF is, objectively, not inside the Core.  This was just wrong.  And more importantly, counterintuitive for the users.

---

## Checklist

### DIFF
--- a/armi/physics/fuelCycle/fuelHandlerInterface.py
+++ b/armi/physics/fuelCycle/fuelHandlerInterface.py
@@ -89,9 +89,7 @@ class FuelHandlerInterface(interfaces.Interface):
         timeYears = self.r.p.time
         # keep track of the EOC time in years.
         self.cycleTime[cycle] = timeYears
-        runLog.extra(
-            "There are {} assemblies in the Spent Fuel Pool".format(len(self.r.sfp))
-        )
+        runLog.extra(f"There are {len(self.r.sfp)} assemblies in the Spent Fuel Pool")
 
     def interactEOL(self):
         """Make reports at EOL."""
--- a/armi/physics/fuelCycle/fuelHandlerInterface.py
+++ b/armi/physics/fuelCycle/fuelHandlerInterface.py
@@ -90,9 +90,7 @@ class FuelHandlerInterface(interfaces.Interface):
         # keep track of the EOC time in years.
         self.cycleTime[cycle] = timeYears
         runLog.extra(
-            "There are {} assemblies in the Spent Fuel Pool".format(
-                len(self.r.core.sfp)
-            )
+            "There are {} assemblies in the Spent Fuel Pool".format(len(self.r.sfp))
         )
 
     def interactEOL(self):
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -78,6 +78,7 @@ class Reactor(composites.Composite):
         self.p.cycle = 0
         self.p.flags |= Flags.REACTOR
         self.core = None
+        self.sfp = assemblyLists.SpentFuelPool("Spent Fuel Pool", self)
         self.blueprints = blueprints
 
     def __getstate__(self):
@@ -88,6 +89,7 @@ class Reactor(composites.Composite):
 
     def __setstate__(self, state):
         composites.Composite.__setstate__(self, state)
+        self.sfp.parent = self
 
     def __deepcopy__(self, memo):
         memo[id(self)] = newR = self.__class__.__new__(self.__class__)
@@ -207,10 +209,6 @@ class Core(composites.Composite):
         self.spatialGrid = None
         self.xsIndex = {}
         self.p.numMoves = 0
-
-        # build a spent fuel pool for this reactor
-        runLog.debug("Building spent fuel pools")
-        self.sfp = assemblyLists.SpentFuelPool("Spent Fuel Pool", self)
         self._lib = None  # placeholder for ISOTXS object
         self.locParams = {}  # location-based parameters
         # overridden in case.py to include pre-reactor time.
@@ -252,7 +250,6 @@ class Core(composites.Composite):
 
     def __setstate__(self, state):
         composites.Composite.__setstate__(self, state)
-        self.sfp.parent = self
         self.regenAssemblyLists()
 
     def __deepcopy__(self, memo):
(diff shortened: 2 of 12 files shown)


armi/physics/fuelCycle/fuelHandlerInterface.py
```diff
@@ -90,9 +90,7 @@ def interactEOC(self, cycle=None):
         # keep track of the EOC time in years.
         self.cycleTime[cycle] = timeYears
         runLog.extra(
-            "There are {} assemblies in the Spent Fuel Pool".format(
-                len(self.r.core.sfp)
-            )
+            "There are {} assemblies in the Spent Fuel Pool".format(len(self.r.sfp))
```
albeanth, line 93: A bit off topic, but I'll ask anyway. Feel free to punt. 

Before the transition to ruff, pylint used to complain about using `.format()` instead of f-strings. Is this a rule we want to enforce moving forward, or no? 

https://pylint.pycqa.org/en/latest/user_guide/messages/convention/consider-using-f-string.html.

armi/reactor/reactors.py
```diff
@@ -527,7 +525,8 @@ def removeAllAssemblies(self, discharge=True):
         assems = set(self)
         for a in assems:
             self.removeAssembly(a, discharge)
-        self.sfp.removeAll()
+        if hasattr(self.parent, "sfp"):
+            self.parent.sfp.removeAll()
```
albeanth, line 529: Same comment here on `else` for failing conditional.

armi/reactor/reactors.py
```diff
@@ -466,7 +463,8 @@ def removeAssembly(self, a1, discharge=True):
         self.remove(a1)
 
         if discharge and self._trackAssems:
-            self.sfp.add(a1)
+            if hasattr(self.parent, "sfp"):
```
keckler, line 466: It is unclear to me why this `if` is needed. I think that `self.parent` (i.e. the reactor object) will always have a `sfp` based on the way that the reactor object is initialized. Is that wrong?

keckler: This comment also applies to a similar `if` statement below.

keckler: More importantly, if this `if` statement fails, what then happens to the assembly. I guess it will just poof disappear, without the user knowing about it. Because they have set `trackAssems: True`, this would be unexpected behavior for them that they should probably be warned about.

keckler (approved):
I don't see anything here that conflicts with my models which employ the SFP, so I'm good with these changes!

armi/physics/fuelCycle/fuelHandlerInterface.py
john-science: Yeah, as a rule we like f-strings better than the various other options.

Here, I didn't create the original `.format()`, of course.  So I was just making the necessary change.  But I can clean that up if you like.

albeanth: For sure. I suppose I was mostly curious if we should add the preference of f-strings over `format.()` into our linting? Recognizing that this will result in a gazillion changes since there are LOTS of `format.()` calls in ARMI.... What do you think?

armi/reactor/reactors.py
```diff
@@ -466,7 +463,8 @@ def removeAssembly(self, a1, discharge=True):
...
+                self.parent.sfp.add(a1)
```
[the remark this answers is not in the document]
john-science: Well, imagine we said:

```python
if self.parent is not None and  hasattr(self.parent, "sfp"):
```

It turns out `hasattr()` covers our bases so we don't need the "is not None" statement.

My major concerns are:

1.  This is Python, you can build a "Core" object in vaccuum with no "Reactor" object.
2. Maybe someone wants to build multiple cores for an analysis, that should be allowable with them having to build multiple reactors.
  a. People who complain about ARMI usually complain that everything is too complicated because they can't just do the one thing they want without: building a whole reactor or building an Operator/Plugin.  So I take it people would prefer they were freed from having to do so.

armi/physics/fuelCycle/fuelHandlerInterface.py
john-science: fixed

john-science: I'm not going to open a 200-file PR for the `.format()` thing, I think.

I mean, maybe if I could script it.

But even then, sometimes people have a LOT of code in their `.format()` statements, and the f-strings can only do extremely simple expressions, so you'd have to create temporary variables.  So there _might_ be a couple format statements that end up looking cleaner.  Maybe? I'm just spit-balling.

albeanth: Ok, I'm fine with it. I don't have a strong preference. Was just an inconsistency I realized. Figured I'd call it out.

armi/reactor/reactors.py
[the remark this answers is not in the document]
albeanth: Totally. I wasn't necessarily concerted with an `is not None` getting checked. I was thinking more about what happens in the case when there isn't a parent `Reactor` but a user still tries to add an assembly to `r.sfp`. In this current implementation, the user would never know that the assembly wasn't getting added as the code wouldn't tell them. The conditional would just evaluate to false and move on. So I was thinking maybe something akin to:

```python
if hasattr(self.parent, "sfp"):
    self.parent.sfp.add(a1)
else:
    runLog.warning(f"There is no parent Reactor object for {self}, so {a1} will not be added to the spent fuel pool")
```

I believe @keckler had a similar thought as well. https://github.com/terrapower/armi/pull/1336#discussion_r1253358584

armi/reactor/reactors.py
john-science: Okay, I added a log statement if the SFP doesn't exist, but `trackAssems` is on.  That's a good call.

PR #1338, opened 2023-07-05
Fix minor conservation issue + new diagnostic summary for conservation

### LINKED ISSUE #1048: Axial Expansion Changer + verbosity: debug
When the verbosity is set to debug, too much information is printed to the screen. It _really_ clutters up the stdout. This is particularly true for:
1. https://github.com/terrapower/armi/blob/6bd85c628715e0ac7b48984fd453a4d3176900b6/armi/reactor/converters/axialExpansionChanger.py#L425-L442
2. https://github.com/terrapower/armi/blob/6bd85c628715e0ac7b48984fd453a4d3176900b6/armi/reactor/converters/axialExpansionChanger.py#L479-L498

These are indeed useful but they print way too many times and after the first time they are printed, they are just a nuisance.

albeanth:
## What is the change?
Downstream testing illustrated some minor conservation issues with non-axial expansion target components (e.g., clad, wire, or duct within a fuel block) during thermal expansion. This PR addresses this and also introduces two things:
1. minor cleanup of simplifying some repeated and unnecessary code.
2. diagnostics that allows users and devs to identify mass conservation issues (these were used in place of the existing runLogs so their utility has been illustrated, at least in my opinion!). 

Some additional details on item 2. There are two tables that get printed. 
- `massConservationReport`: prints the pre- and post-axial expansion assembly mass (as well as the difference) by material and by component if the difference if greater than 1e-9 grams. This 1e-9 g threshold was chosen semi-arbitrarily as the differences between pre- and post-axial expansion were found in downstream testing to range from 0.0 to on the order of 1e-10 g.  
- `detailedMassConservationReport`: prints the following component-level information for an assembly: grow fraction, bottom, top, height, and mass for pre- and post-axial expansion. The post-pre component mass is also printed. This summary table gets printed only for debug runs. 

Closes #1048. 

The axial expansion changer should not lose mass post thermal expansion. This PR fixes that. There are some cases where this is inevitable and these should be documented in the ARMI docs (in a subsequent PR). 

---

## Checklist


armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -124,6 +126,8 @@ def __init__(self, detailedAxialExpansion: bool = False):
         self._detailedAxialExpansion = detailedAxialExpansion
         self.linked = None
         self.expansionData = None
+        self.massConservationReport = None
+        self.detailedMassConservationReport = None
```
keckler, line 130: It doesn't appear that there is any need to store these items on `self`, since they're only really used in one method of this class.

I think that trashing these dicts after each call to `axiallyExpandAssembly()` would be both safe and more memory efficient.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -235,25 +239,55 @@ def _isTopDummyBlockPresent(self):
 
     def axiallyExpandAssembly(self):
         """Utilizes assembly linkage to do axial expansion."""
+        # Set up conservation report information (useful for debugging)
+        self.massConservationReport = {}
```
keckler, line 243: It appears that a lot of work is done to build the `massConservationReport` and `detailedMassConservation` dicts.

I am concerned that this work will be performed on every call to `axiallyExpandAssembly()`, regardless of what the `runLog` verbosity is set to for the run.

It seems like there should be an elegant way to cut out all this work if the verbosity setting is low enough to not be printing these reports.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -267,32 +301,95 @@ def axiallyExpandAssembly(self):
...
+            "(prev) c.bottom",
+            "(prev) c.top",
+            "(prev) c.height",
+            "(post) c.bottom",
+            "(post) c.top",
+            "(post) c.height",
+            "(prev) mass",
+            "(post) mass",
+            "(post - pre) mass",
+        ]
```
keckler, line 391: This can be a (private) module variable to avoid reinstantiating it all the time.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -235,25 +239,55 @@ def _isTopDummyBlockPresent(self):
...
+        materialMasses = collections.defaultdict(float)
+        componentMasses = collections.defaultdict(float)
+        for b in self.linked.a:
+            for c in getSolidComponents(b):
+                materialMasses[c.material.name] += c.getMass()
+                componentMasses[c.name] += c.getMass()
+        self.massConservationReport[self.linked.a] = (
+            list(materialMasses.keys()) + [None] + list(componentMasses.keys())
+        )
+        self.massConservationReport["pre-exp"] = (
```
keckler, line 253: Might be clearer if you changed this header to `mass pre-exp`?

keckler: Same with the corresponding `post-exp` entry below.

keckler (changes requested):
I left some comments throughout the code, and have one more here:

In the PR description you say
> Downstream testing illustrated some minor conservation issues with non-axial expansion target components (e.g., clad, wire, or duct within a fuel block) during thermal expansion. This PR addresses this 

It is unclear to me exactly what has changed to fix the problem that you were seeing (partly because I don't know the exact problem you were seeing). 

Is the fix related to the change of `c.height = growFrac * blockHeight` to `c.height = growFrac * prevCompHeight`?

armi/reactor/converters/axialExpansionChanger.py
albeanth: I put them on `self` in https://github.com/terrapower/armi/pull/1338/commits/29c345e117746a5a4e69193153697d964f8e92d4 simply to accommodate testing. I could use the mock logs to rebuild them if you'd prefer... But that seemed clunky to me and the easier solution was just to make them class variables.

albeanth: That said, if we kept them as class instances, we could empty them after they are populated. Maybe a compromise of sorts?

armi/reactor/converters/axialExpansionChanger.py
albeanth: Addressed https://github.com/terrapower/armi/pull/1338/commits/80444f851a16baf40e0a2f54c301c84765f8d7d5

armi/reactor/converters/axialExpansionChanger.py
albeanth: Addressed https://github.com/terrapower/armi/pull/1338/commits/80444f851a16baf40e0a2f54c301c84765f8d7d5

albeanth:
> Is the fix related to the change of c.height = growFrac * blockHeight to c.height = growFrac * prevCompHeight?

@keckler, yes, exactly.

armi/reactor/converters/axialExpansionChanger.py
albeanth: Yeah, I agree. Working on this.

armi/reactor/converters/axialExpansionChanger.py
albeanth: Ok, so this should be resolved with https://github.com/terrapower/armi/pull/1338/commits/027db327772541229a33290a8116c7789109b807. All of the logging now lives in its own class and only gets populated if the `runLog` verbosity is either `extra` or `debug`. The dicts that get populated will go out of scope with the rest of the expansion changer and should follow suit with the expansion data, etc.

albeanth:
> The mass conservation tables look like the following:
> 
> ### massConservationReport
> Is assembly-wide and shows the pre- and post-axial expansion mass values for the unique materials and components.
> ...

Just a note, I did confirm that the summary tables are identical post https://github.com/terrapower/armi/pull/1338/commits/027db327772541229a33290a8116c7789109b807.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -800,3 +806,129 @@ def isTargetComponent(self, c):
             Component to check target component status
         """
         return bool(c in self._componentDeterminesBlockHeight)
+
+
+class AxialExpMassConservationSummaryLogger:
+    """Class to log and print axial expansion mass conservation."""
+
+    # do not do logging for verbosity levels greater than or equal to
+    _LOGGING_VERBOSITY_THRESHOLD = 20  # runLog.info
```
keckler, line 815: I think you can pull this value straight from the `runLog` module (or the `logging` module which it inherits from) in order to keep those consistent.

keckler (approved):
Thanks for that change, I think it streamlines this a lot!

armi/reactor/converters/axialExpansionChanger.py
albeanth: @john-science turns out that this approach is sensitive to parallelization and doesn't really work with our current testing environment.... particularly in regard to the global cs struggles.... 

It'd be great to be able to populate these summary reports based on the runLog verbosity, and in serial, I think we can. But since we struggle with parallelization in regard to changing case settings, I can't fiddle my way to get the tests to pass in parallel. 

Any suggestions would be welcomed!

armi/tests/detailedAxialExpansion/refSmallReactorBase.yaml
```diff
@@ -40,7 +40,7 @@ blocks:
             Thot: 450.0
             ip: grid.op
             mult: 1.0
-            op: 16.75
+            op: 16.8
```
jakehader, line 43: Why is this changing?

jakehader: Is it to fit the ACLP block?

albeanth: Yeah exactly.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -800,3 +806,140 @@ def isTargetComponent(self, c):
...
+        )
+        # check for differences in pre and post expansion mass greater than 1e-9 grams
+        diff = [
+            round(new - orig, 9) if new else None
+            for new, orig in zip(
+                self.massConservationReport["(post) mass"],
+                self.massConservationReport["(prev) mass"],
+            )
+        ]
+        self.massConservationReport["round(post - prev, 9)"] = diff
```
john-science, line 927: Why is this "self.massConservationReport"?  Isn't this just a local variable?  Is it used outside this method?

It appears to be a new variable that is local, but is initialized in the constructor.

john-science (approved):

armi/reactor/converters/axialExpansionChanger.py
albeanth: I put it on `self` to accommodate some unit testing.

armi/reactor/converters/axialExpansionChanger.py
john-science:
```python
from logging import INFO
```

Or:

```python
import logging

logging.INFO
```

albeanth: https://github.com/terrapower/armi/pull/1338/commits/9d8362b40d2be61277e8a43210f26636c1a8f785

armi/reactor/converters/axialExpansionChanger.py
onufer: prefer returning it if its not actually needing to be an attribute.

we would have WAY more attributes if we did this every time we wanted to test.

albeanth:
> > Is the fix related to the change of c.height = growFrac * blockHeight to c.height = growFrac * prevCompHeight?
> 
> @keckler, yes, exactly.

This is breaking downstream testing... Converting to draft while I debug.

john-science:
@albeanth This PR is almost a year old now.  What is its status?

albeanth:
That its almost had its first birthday! 😄 Feel free to close. I can dig it
up when the time comes.

john-science:
I have no problem with the idea behind this PR.  But it is a year old now, and feels abandoned.

Please bring this PR back the _moment_ you are willing to open it in non-`Draft` from.

Issue #957, opened 2022-11-01
Test support for Python 3.11


john-science:
This could be as easy as adding it to the GitHub workflow.

But some local testing wouldn't hurt.

john-science:
Currently, the blocker on this is `mpi4py` isn't compiling for Linux  on Python 3.11.

Please see [this](https://github.com/mpi4py/mpi4py/issues/273) bug report for mpi4py.

dalcinl:
@john-science [mpi4py 3.4.1](https://github.com/mpi4py/mpi4py/releases/tag/3.1.4) should resolve your issue.

john-science:
Thanks, @dalcinl !  That fixed the `mpi4py` issue!

john-science:
The current stopping block is [h5py](https://pypi.org/project/h5py/#files) does not have an official Python 3.11 release yet.

They have an [issue](https://github.com/h5py/h5py/issues/2146) open for this though.

john-science:
H5PY is still the limiting factor.  I can't seem to successfully [install it in Python 3.11](https://github.com/john-science/armi/actions/runs/3681061169/jobs/6227339055):

```
      Loading library to get build settings and version: libhdf5.so
      error: Unable to load dependency HDF5, make sure HDF5 is installed properly
      error: libhdf5.so: cannot open shared object file: No such file or directory
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for h5py
```

...

```
Building wheels for collected packages: h5py
  Building wheel for h5py (pyproject.toml): started
  Building wheel for h5py (pyproject.toml): finished with status 'error'
Failed to build h5py
mpitest: exit 1 (10.07 seconds) /home/runner/work/armi/armi> python -I -m pip install configparser coverage future 'h5py>=3.0' htmltree matplotlib numpy ordered-set pillow pluggy pyDOE pyevtk scipy tabulate voluptuous xlrd yamlize pid=3408
.pkg: _exit> python /opt/hostedtoolcache/Python/3.11.0/x64/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
  test: FAIL code 1 (50.61 seconds)
```

So, we can't just install the version-dependent tarball from the h5py team. We need to wait for a wheel from them. After much work, it appears that all our non-h5py unit tests pass in Python 3.11.  I think this is out last hold-out.

john-science:
[...]

PR #1341, opened 2023-07-10
Add Python 3.11 to GH Actions

opotowsky:
## What is the change?

Updating ARMI to run CI with python 3.11. 

Note, however, the docs GH action is remaining behind with python 3.9.

Python 3.11 offers a 10-40% speedup! Let's get with the program. 

Resolves Issue #957, but Opens Issues #1344 and #1377

---

## Checklist

- [x] ~Tests have been added/updated to verify that the new/changed code works.~


.github/workflows/black.yaml
```diff
@@ -8,10 +8,10 @@ jobs:
         runs-on: ubuntu-22.04
         steps:
             - uses: actions/checkout@v2
-            - name: Set up Python 3.10
+            - name: Set up Python 3.11
```
john-science, line 11: I'm not sure about this change.  I think for the `black` and `coverage` actions maybe we should leave them on the older versions, until people transition to the newer ones.

Thoughts?

opotowsky: I thought these GH actions, minus the python-version-specific unit tests, always used "the latest" python (since most of us are on 3.9 and this is being run on 3.10). Happy to leave it at 3.10 or switch it to 3.9.

opotowsky:
Just FYI there are a bunch of issues with the docs requirements that need updating. I can update all the packages to 3.11-compliant versions and hope that old bugs have been fixed that requires the version pin?

Or we could leave the GH action to build docs with python 3.9 like it does now. But someone wouldn't be able to build them locally if they were using 3.11. Thoughts @john-science?

john-science:
We can just continue to build the docs on GitHub with Python 3.9, and get the tests and downstream simulations up to 3.11.  Right?

And transfer the docs later?

They are fairly independent.

opotowsky:
> We can just continue to build the docs on GitHub with Python 3.9, and get the tests and downstream simulations up to 3.11. Right?

Yes I think it's fine to do that on CI. Someone building the docs locally will need to have a separate 3.9 venv, though. I've actually never built ARMI's locally I'm ashamed to say.

setup.py
```diff
@@ -76,8 +76,8 @@ def collectExtraFiles():
         "yamlize==0.7.1",
     ],
     extras_require={
-        "mpi": ["mpi4py==3.0.3"],
-        "grids": ["wxpython<=4.1.1"],
+        "mpi": ["mpi4py==3.1.4"],
```
opotowsky, line 79: I haven't tested this yet, so I think this PR should either remove the version pin to not break requirements downstream or it should hold off until I can test it.

opotowsky:
@john-science I have a few questions above. I can't test mpi4py update with our internal code for a week or three but I'm worried about pinned version dependency clashes, so I'm proposing removing the version pin altogether.

opotowsky:
I'm going to mark this as draft for now, until we resolve some stuff internally. I don't want to break things with new version pins.

john-science:
@opotowsky @ntouran Arrielle, it would be nice to allow people to check out ARMI and run it on Python 3.11.

Can we make this two PRs?

1. Get dependencies and unit tests up to date to just get Python 3.11 "unit tests" GH working.
2. Then get another PR to do the rest of this?  (ruff and docs)

opotowsky:
> @opotowsky @ntouran Arrielle, it would be nice to allow people to check out ARMI and run it on Python 3.11.
> 
> Can we make this two PRs?
> 
> 1. Get dependencies and unit tests up to date to just get Python 3.11 "unit tests" GH working.
> 2. Then get another PR to do the rest of this?  (ruff and docs)

Among others, I reverted the `ruff` version pin update. I don't know why I even updated it in the first place, since the existing version is 3.11-compatible.

setup.py
```diff
@@ -76,8 +76,8 @@ def collectExtraFiles():
         "yamlize==0.7.1",
     ],
     extras_require={
-        "mpi": ["mpi4py==3.0.3"],
```
opotowsky, line 79: Removing this version pin should allow the right max versions to get installed for tests. 

Testing internally to make sure this doesn't mess anything up!

requirements-testing.txt
```diff
@@ -1,7 +1,4 @@
 numpy>=1.21,<=1.23.5
 ipython>=7.34.0,<=8.12.0
 
-# docutils 0.17 introduced a bug that prevents bullets from rendering
-# see https://github.com/terrapower/armi/issues/274
-docutils <0.17
```
opotowsky, line 6: This could stay removed and I could instead update the requirement in `requirements-docs.txt` (which, oddly, it has `docutils <0.18` instead)

Whatcha think @john-science?

john-science: That's just for docs, right?

Or is it used outside the docs?

setup.py
opotowsky: Ok this is OK to remove, resolving

requirements-testing.txt
opotowsky: It's used in `utils`:

image

But I don't think bullet point rendering matters there ;-)

It needs to be unpinned in setup.py to be python 3.11 compatible.

john-science (approved):
The only question I have is the MPI version change.

Should we run some in-house integration tests to make sure that change doesn't break anything?

Other than that, great!

opotowsky:
> The only question I have is the MPI version change.
> 
> Should we run some in-house integration tests to make sure that change doesn't break anything?
> 
> Other than that, great!

Actually we have mpi4py version pinned in-house, so removing the version pin here won't change anything.

opotowsky:
@john-science @ntouran we have py311 success!

PR #1343, opened 2023-07-13
Removing test_gridGui.py to simplify command line testing

john-science:
## What is the change?

I removed the `test_gridGui.py` file.

(Also, I did some minor cleanup for black/ruff on a couple of recent PRs.)

The single test in this file have been broken since about a week after it was added to the repo.  And we have _never_ run the file in our dev or production environments.  We have also never run the test on GitHub, because it involves booting up a GUI.

---

## Checklist


john-science:
The coverage drop happened because I removed lines of code.  This is our one exceptional case where this is okay.

(I've been removing a lot of code lately.)

PR #1346, opened 2023-07-17
Moving missing file warning to debug

### LINKED ISSUE #1304: Dont check file paths in settings when doing case suite copying
There is a file-exists check that happens in [copyInterfaceInputs()](https://github.com/terrapower/armi/blame/main/armi/cases/case.py#L886); but this is not necessary when just copying, it is necessary when trying to run.

john-science:
## What is the change?

This moves a "missing file" warning from `INFO` to `DEBUG` when cloning a case.

> The only problem I have here, is I don't know how to test this with all the different workflows, to make sure it doesn't cause anyone problems.  Ideas?

This is to close #1304 .

It is a request, because users don't like seeing the extra warning on the command line.

---

## Checklist

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-r"""
+"""
 The ``Case`` object is responsible for running, and executing a set of user inputs.  Many
 entry points redirect into ``Case`` methods, such as ``clone``, ``compare``, and ``run``.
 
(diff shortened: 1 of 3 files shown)


keckler (commented):
Not sure whether or not this satisfies the original request from @onufer , so I'm going to abstain from approving.

opotowsky (approved):
Following Chris' comment, I have confirmed offline that this does satisfy Mark's request. He's happy to just have the warning print for higher verbosity runs.

The code change is easy enough, approved!

onufer:
this does not solve the issue. if the issue was just a message, i wouldn't have bothered to make a ticket, or i would have changed there verbosity myself. sorry the original issue was unclear... working on getting you a valid traceback. @opotowsky told me the issue so i think i need to talk with her

onufer (changes requested):
print message verbosity was not the problem. working to get you a traceback...

PR #1349, opened 2023-07-18
Make SFP a child of reactor.

mgjarrett:
## What is the change?

The spent fuel pool (`sfp`) is an attribute of the reactor, but not a child. As a result, the SFP does not get written to the database. Thus, the SFP will not be populated when loading from a database (e.g., restart or snapshots case). This PR makes the SFP a child so that it is returned by `reactor.__iter__()` and written to the database.

The one downside of this implementation is that a user will now have to define a spent fuel pool system in the blueprints (`type: sfp`) in order to track assemblies. The blueprints SFP is only required if the user wants to track assemblies. Before, the `r.core.sfp` (and then `r.sfp`) were automatically created. But the SFP definition in the blueprints is pretty simple. The minimum working example shown below should work in general:

```
systems:
  core:
    grid name: core
    origin:
      x: 0.0
      y: 0.0
      z: 0.0

  Spent Fuel Pool:
    type: sfp
    grid name: sfp
    origin:
      x: 0.0
      y: 0.0
      z: 0.0
```

Some workflows can put assemblies into the SFP and then later pull those assemblies out to load them back into the core. We'd like to be able to do this during a restart run. Or, somebody might just want to do some post-processing and have access to all of the discharged assemblies.

https://github.com/terrapower/armi/issues/876

---

## Checklist


mgjarrett:
@keckler @john-science this is ready for review

keckler:
I have not yet looked at this (about to), but one thought is that it might only make sense to require a SFP in the blueprints if the user turns on `trackAssems`.

keckler (commented):
Seems like everything you implemented should work, so from that perspective I approve.

However, I have a couple concerns that I'd like to discuss. Please feel free to correct me if I'm misunderstanding some of this, because the database system is somewhat magical to me.
1. I think most people would agree that it is not desirable for everybody going forward to have to define a spent fuel pool in their system definitions. People who need an SFP, yes, they should have to define it. But people who don't care about it should have the option of not having one. The way I understand your PR, it would appear that this would become mandatory. If I'm correct about that, then this will cause a lot of broken cases when people update their ARMI version.
2. Is this _really_ the path that needs to be taken when we want to get a new reactor system written into the database? If so, this is a problem, because anytime somebody needs a new system written to the database, it will require a new mandatory system for everybody, and that is not sustainable. I'm not very hip to how we ensure that specific items get written into the database, but this seems like an issue to me.

jakehader:
> Seems like everything you implemented should work, so from that perspective I approve.
> 
> However, I have a couple concerns that I'd like to discuss. Please feel free to correct me if I'm misunderstanding some of this, because the database system is somewhat magical to me.
> 
> 1. I think most people would agree that it is not desirable for everybody going forward to have to define a spent fuel pool in their system definitions. People who need an SFP, yes, they should have to define it. But people who don't care about it should have the option of not having one. The way I understand your PR, it would appear that this would become mandatory. If I'm correct about that, then this will cause a lot of broken cases when people update their ARMI version.
> 2. Is this _really_ the path that needs to be taken when we want to get a new reactor system written into the database? If so, this is a problem, because anytime somebody needs a new system written to the database, it will require a new mandatory system for everybody, and that is not sustainable. I'm not very hip to how we ensure that specific items get written into the database, but this seems like an issue to me.

I agree with this and think that we should move to allow users to add "n" systems to the reactor, by application or by plugin based on the application need. For example, the SFP is just one storage area in a plant and there could be others. 

I think the idea of leaving the SFP as is, is to not break fuel management capabilities with where core assemblies are discharged to by some flag/setting. I would propose that we make the SFP optional and pair this change with another update to fuel management that can either (a) discharge and the assemblies get removed from the system and cannot come back or (b) require that the user sets up some logic for where the assemblies go on discharge. If the user sets up where the assemblies go from discharge then we can check that the location defined does exist on the reactor or it fails.

If a system is defined as an extra place in the reactor and fuel management doesn't use this location then it can be okay to still exist.

I do think it's a good idea for any new system on the reactor to have the capability of being written to the database though and to have the framework set up to reload that system.

What do we think? Is this too difficult to do with the current structure of the code?

armi/tests/detailedAxialExpansion/refSmallReactor.yaml
```diff
@@ -7,6 +7,7 @@ systems:
             y: 0.0
             z: 0.0
     sfp:
+        type: sfp
         grid name: sfp
```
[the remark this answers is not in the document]
mgjarrett: 🤷 the only one that has to be literally "sfp" is the `type`. The `grid name` and the name of the `system` are arbitrary, I just made them `sfp` to be equivalent to the `core`.

For the `core` system, the `grid name` is also `core`, and the default type is `core`. So it's just a parallel of that.

mgjarrett:
> Seems like everything you implemented should work, so from that perspective I approve.
> 
> However, I have a couple concerns that I'd like to discuss. Please feel free to correct me if I'm misunderstanding some of this, because the database system is somewhat magical to me.
> 
> 1. I think most people would agree that it is not desirable for everybody going forward to have to define a spent fuel pool in their system definitions. People who need an SFP, yes, they should have to define it. But people who don't care about it should have the option of not having one. The way I understand your PR, it would appear that this would become mandatory. If I'm correct about that, then this will cause a lot of broken cases when people update their ARMI version.
> 2. Is this _really_ the path that needs to be taken when we want to get a new reactor system written into the database? If so, this is a problem, because anytime somebody needs a new system written to the database, it will require a new mandatory system for everybody, and that is not sustainable. I'm not very hip to how we ensure that specific items get written into the database, but this seems like an issue to me.

It was not my intention to make the SFP required in all cases. It should only be required when a user wants to track assemblies and store them. However, after testing I realized that non-SFP cases were not correctly functional. I have made a few more commits and am re-running tests to verify that ARMI doesn't crash for non-SFP cases.

mgjarrett:
> Seems like everything you implemented should work, so from that perspective I approve.
> 
> However, I have a couple concerns that I'd like to discuss. Please feel free to correct me if I'm misunderstanding some of this, because the database system is somewhat magical to me.
> 
> 1. I think most people would agree that it is not desirable for everybody going forward to have to define a spent fuel pool in their system definitions. People who need an SFP, yes, they should have to define it. But people who don't care about it should have the option of not having one. The way I understand your PR, it would appear that this would become mandatory. If I'm correct about that, then this will cause a lot of broken cases when people update their ARMI version.
> 2. Is this _really_ the path that needs to be taken when we want to get a new reactor system written into the database? If so, this is a problem, because anytime somebody needs a new system written to the database, it will require a new mandatory system for everybody, and that is not sustainable. I'm not very hip to how we ensure that specific items get written into the database, but this seems like an issue to me.

I needed to add a few more `if self.r.sfp is not None`'s for it to work without the SFP. With this fixed, I probably didn't need to add SFPs to all of the test reactors that I added them to, so I'll go back and check that tomorrow.

mgjarrett:
> I agree with this and think that we should move to allow users to add "n" systems to the reactor, by application or by plugin based on the application need. For example, the SFP is just one storage area in a plant and there could be others.
> 
> I think the idea of leaving the SFP as is, is to not break fuel management capabilities with where core assemblies are discharged to by some flag/setting. I would propose that we make the SFP optional and pair this change with another update to fuel management that can either (a) discharge and the assemblies get removed from the system and cannot come back or (b) require that the user sets up some logic for where the assemblies go on discharge. If the user sets up where the assemblies go from discharge then we can check that the location defined does exist on the reactor or it fails.
> 
> If a system is defined as an extra place in the reactor and fuel management doesn't use this location then it can be okay to still exist.
> 
> I do think it's a good idea for any new system on the reactor to have the capability of being written to the database though and to have the framework set up to reload that system. I am not sure if mandatory writing is required though. I think that the core writing itself to the database as mandatory makes sense, but maybe there is an opportunity for the database interface to be configurable?
> 
> What do we think? Is this too difficult to do with the current structure of the code?

There's a lot here, and I think some of it is beyond the scope of this PR. The intent of this PR is to make `r.sfp` a child of the reactor so that it is stored in the database. Yes, it would be good to have any arbitrary number of systems be able to be attached to the reactor, be stored in the database, and be reloaded from the database. But those systems would have to be developed first. There is already a `SpentFuelPool` class. This PR is just taking the small step of making it officially a "child" of the reactor. It's a much bigger task to add arbitrary systems as children of the reactor that are written to the database.

I agree that having an arbitrary list of systems attached to the reactor could be a useful future development. We should make a separate issue for this.

armi/reactor/reactors.py
```diff
@@ -463,7 +465,7 @@ def removeAssembly(self, a1, discharge=True):
         self.remove(a1)
 
         if discharge and self._trackAssems:
-            if hasattr(self.parent, "sfp"):
+            if self.parent.sfp is not None:
```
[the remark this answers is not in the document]
mgjarrett: I've updated the PR so it is allowed to not have an SFP. If there is no SFP, the discharged assembly will be lost.

john-science:
@mgjarrett I don't really understand this statement:

> The one downside of this implementation is that a user will now have to define a spent fuel pool system in the blueprints

First off, that's a pretty huge change. No joke, that'll mean downstream changes in nearly every repo and certainly all blueprint files (test and production).

Before I concede this point, I just want to understand it.  How did you come to this design?  Are default values not an option?

mgjarrett:
> @mgjarrett I don't really understand this statement:
> 
> > The one downside of this implementation is that a user will now have to define a spent fuel pool system in the blueprints
> 
> First off, that's a pretty huge change. No joke, that'll mean downstream changes in nearly every repo and certainly all blueprint files (test and production).
> 
> Before I concede this point, I just want to understand it. How did you come to this design? Are default values not an option?

The full sentence is:

> The one downside of this implementation is that a user will now have to define a spent fuel pool system in the blueprints (type: sfp) in order to track assemblies. 

It's only required when the user wants to track assemblies. The setting `trackAssems` defaults to `False`, so a user would only have to add this to the blueprints if they want `trackAssems: true`. We could add an SFP to the reactor by default, but maybe people don't always want to have an SFP on the reactor.

If you want to model a `core`, you have to define one in the blueprints. Why can't the same be true of an `sfp`? I agree it is somewhat unfortunate because we used to create a thing called `sfp` by default, but that thing wasn't tied to the blueprints in any way so the user would have no control over the configuration of it.

john-science:
@mgjarrett So, the Level Of Effort for forcing people (even just those who are tracking assemblies) to change their blueprints is pretty high.

It's not just that people would need to change their blueprint files, it's that we'd need to communicate that idea to everyone. Then everyone would have to find all their old blueprint files and... suddenly we are weeks into a very large communications and kitten herding project where (I promise you) people WILL get angry with you for "wasting" their time.

That is my experience.  If you do not "buy" that, please talk to me offline.


So, it would be better if things still worked if people didn't change their blueprints files.  I hazard a guess there is a way to _assume_ an empty SFP if one isn't defined in the blueprints.

mgjarrett:
> @mgjarrett So, the Level Of Effort for forcing people (even just those who are tracking assemblies) to change their blueprints is pretty high.
> 
> It's not just that people would need to change their blueprint files, it's that we'd need to communicate that idea to everyone. Then everyone would have to find all their old blueprint files and... suddenly we are weeks into a very large communications and kitten herding project where (I promise you) people WILL get angry with you for "wasting" their time.
> 
> That is my experience. If you do not "buy" that, please talk to me offline.
> 
> So, it would be better if things still worked if people didn't change their blueprints files. I hazard a guess there is a way to _assume_ an empty SFP if one isn't defined in the blueprints.

We can probably just make a default SFP and throw that on the reactor. The reason I initially was trying to avoid that is because the long-term vision for the `systems` is that they could be used to model all kinds of things that aren't part of the active core, like heat exchangers, pumps, or in-vessel assembly storage. The `SFP` is just the first system we've added in what is supposed to be a long line of systems, and I don't think we want to have all of them created on every reactor by default. Not all reactors are going to have every type of system.

The `SpentFuelPool` is an `AssemblyList`, so it only really makes sense for assembly-based reactors. If somebody is modeling a liquid or pebble fuel reactor they might want to implement a different kind of long-term storage object for their spent fuel. So those people might not be happy about an `AssemblyList` being created by default and attached to the reactor as `r.sfp`. But I can probably make a separate issue for that.

mgjarrett:
I understand that the issue is that people have existing blueprints that don't have an SFP, so the easiest thing is to just add it by default. That's the direction we should take.

But I do think it's important to also note that the docs, going back to at least 2019, show an `sfp` system being defined in the basic blueprints example:

image

mgjarrett:
@john-science and @keckler I added a blueprints method that creates a "default" SFP if one is not defined in the blueprints. The reactor factory calls this method before constructing the reactor from the blueprints.

keckler:
Side note -- `trackAssems` is a pretty vague name for the setting that now controls all this SFP machinery. 

Its not necessary, but maybe we'd want to deprecate that name in favor of something more descriptive.

Totally up to you.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -652,6 +652,9 @@ def _getAssembliesInRings(
         assemblyList : list
             List of assemblies in each ring of the ringList. [[a1,a2,a3],[a4,a5,a6,a7],...]
         """
+        if "SFP" in ringList and self.r.sfp is None:
+            raise ValueError(f"No SFP attached to the reactor {self.r}!")
```
keckler, line 656: Maybe add a suggestion to the message on how to fix the issue?

armi/reactor/reactors.py
```diff
@@ -156,7 +161,7 @@ def factory(cs, bp, geom: Optional[SystemLayoutInput] = None) -> Reactor:
         coreDesign.construct(cs, bp, r, geom=geom)
         for structure in bp.systemDesigns:
             if structure.name.lower() != "core":
-                structure.construct(cs, bp, r)
+                structure.construct(cs, bp, r, loadAssems=False)
```
keckler, line 164: Can you provide some context on this change for me? I don't immediately see why it was made.

keckler (approved):
You have some merge conflicts to address. Otherwise I think this looks good. Thanks for taking this on, it will surely be a big benefit to a lot of users!

mgjarrett:
> Side note -- `trackAssems` is a pretty vague name for the setting that now controls all this SFP machinery.
> 
> Its not necessary, but maybe we'd want to deprecate that name in favor of something more descriptive.
> 
> Totally up to you.

It seems like more hassle than it's worth to change the name of the setting. I think the `description` and `label` attributes of the setting have enough detail for anybody who wants to find out what the setting does.

armi/reactor/reactors.py
mgjarrett: `loadAssems` was causing failures in test_rz_reactors.py because there wasn't any `gridDesign` defined for the RZT reactors. But setting `loadAssems=False` was the wrong approach because we do actually want to load the assemblies from the blueprints, if they're there.

I changed the logic in reactorBlueprint.py to skip `loadAssems` if there is no `gridDesign` available, and now the tests pass without this argument.

keckler (approved):

PR #1351, opened 2023-07-18
Removing units.NOT_APPLICABLE in favor of UNITLESS

john-science:
## What is the change?

This PR removes the unit `NOT_APPLICABLE`, in favor of just using `UNITLESS`.

(Also, it fixes some `ruff` errors.)

I marked the `NOT_APPLICABLE` unit as "prefer UNITLESS" in a recent PR. But after looking in all downstream repos, I couldn't find a single instance of `NOT_APPLICABLE` being used. So I think it's less confusing to just have one of them.  This should be easier for developers to understand moving forward.

---

## Checklist

### DIFF
--- a/armi/utils/tests/test_textProcessors.py
+++ b/armi/utils/tests/test_textProcessors.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Tests for functions in textProcessors.py."""
-from io import StringIO, TextIOWrapper, TextIOBase
+from io import StringIO
 import os
 import pathlib
 import ruamel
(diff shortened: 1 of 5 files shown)


john-science:
Code coverage dropped because I removed a line of code.  That's the exception we're okay with.

armi/utils/tests/test_textProcessors.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Tests for functions in textProcessors.py."""
-from io import StringIO, TextIOWrapper, TextIOBase
+from io import StringIO
```
opotowsky, line 15: Oops, I missed this tidbit of cleanup in my PR -- thanks!

opotowsky (approved):

PR #1355, opened 2023-07-20
Update reactivity coefficient parameters

mgjarrett:
## What is the change?

Update the parameter definition for some reactivity coefficient parameters to get the correct uniform mesh mapping behavior (change from `ParamLocation.AVERAGE` to `ParamLocation.VOLUME_INTEGRATED`).

Also changed some parameter defaults (`None` for list/array-like, `0.0` for scalar floats).

Simplified the logic for getting parameter values to be mapped in the uniform mesh converter.

Certain parameters were not being mapped correctly because they were listed with the wrong `ParamLocation` tag.

---

## Checklist


armi/reactor/converters/uniformMesh.py
```diff
@@ -1317,7 +1317,7 @@ def paramSetter(block, vals, paramNames):
             if val is None:
                 continue
 
-            if isinstance(val, (list, numpy.ndarray)):
+            if isinstance(val, (tuple, range, list, numpy.ndarray)):
```
albeanth, line 1320: Is there an expectation that val would ever be a `range` object?

albeanth (commented):
Nice simplification in uniformMesh!

armi/reactor/converters/uniformMesh.py
mgjarrett: It came up in a unit test, but I fixed the unit test so I should just remove it

albeanth (approved):

armi/reactor/parameters/parameterDefinitions.py
```diff
@@ -69,6 +69,8 @@ class Category:
     * `neutronics` parameters are calculated in a neutronics global flux solve
     * `gamma` parameters are calculated in a fixed-source gamma solve
     * `detailedAxialExpansion` parameters are marked as such so that they are mapped from the uniform mesh back to the non-uniform mesh
+    * `reactivity coefficients` parameters are related to reactivity coefficient or kinetics parameters for kinetics solutions
+    * `thermal hydarulics` parameters come from a thermal hydraulics phyiscs plugin (e.g., flow rates, temperatures, etc.)
```
jakehader, line 74: Typo in `physics`

armi/reactor/converters/uniformMesh.py
```diff
@@ -1327,22 +1327,11 @@ def paramGetter(self, block, paramNames):
...
-            # checks if the value type is any of these and if so, the block-level parameter is treated as an
-            # array.
-            if isinstance(valType, (list, numpy.ndarray)) or isinstance(None, valType):
-                if val is None or len(val) == 0:
-                    paramVals.append(None)
-                else:
-                    paramVals.append(numpy.array(val))
-            # Otherwise, the parameter is treated as a scalar, like a float/string/integer.
+            # list-like should be treated as a numpy array
+            if isinstance(val, (tuple, range, list, numpy.ndarray)):
```
jakehader, line 1331: Should this `isinstance` match above exactly within the setter?

john-science:
So, quick question.  

These parameters need to be in ARMI and not a downstream reactivity coeffs plugin/repo?

albeanth:
And one the great debates (well, to me lol) rages on!

I'm of the opinion that these (amongst, many _many_ others) should be moved downstream. I recognize the other arguments for keeping these sorts of things on ARMI, but I currently think they should be moved downstream.

Just my two cents.

mgjarrett:
Yeah, these parameters could definitely be moved to a downstream plugin, but I was just trying to tread lightly here. The PR was mostly about fixing minor issues with default values or locations for a few parameters. I only added one new reactivity coefficients parameter, which can be migrated downstream with the others if/when that happens.

PR #1358, opened 2023-07-21
Ensuring we throw an error on bad YAML files

### LINKED ISSUE #433: Multiple components with same name in a block
As best as I can tell, it seems that if one defines multiple components that have the same name within given block, previous components with that name will be overwritten by components further down in the input file. Please correct me if I am wrong on this, though. I am having a hard time convincing myself, but some simple testing makes me inclined to believe.

Further, I believe that no warning is given to the user in this case.

The relevant code, I think, is [here](https://terrapower.github.io/armi/_modules/armi/reactor/blueprints/blockBlueprint.html#BlockBlueprint.construct), where components are being added to a KeyedList using their names. I think a check could be put in there to first see if a component with the same name has already been added.

Some might argue that this is a feature, rather than a bug. That could be debated, but my opinion is that a warning message would be helpful in this scenario for reducing input bugs.

john-science:
## What is the change?

Currently, ARMI will except bad YAML files. That is, if you have the same tag twice, it accepts the second one without warning.

This PR makes such bad inputs hard failures.

Chris opened ticket "4 4 3", where he points out you can have two components with the same name in a Block. Which, in the end, is an invalid YAML file. And we should catch that as a hard error.

Though that ticket was about a very specific problem, this solution handles a much wider range of problems. Like duplicate settings and the like.

Unfortunately, as described below, we can only solve the problem for top-level duplicate keys. Sigh.

---

## Checklist


keckler:
Is this PR complete? Looking at it, it is not immediately obvious to me how this fixes #433 . In particular, I don't see how this addresses the blueprints section of our input yaml files.

Maybe I'm missing something here or don't understand, so an example or a test explicitly on the blueprints would be helpful.

john-science:
Okay, I hate the people that wrote the `YAML` spec.

I thought the problem was `yamlize`, then I thought it was `ruamel.yaml`, but no. Apparently, the official YAML spec defines this as a duplicate key, and thus an error:

```yaml
blocks:
    # things
blocks:
    # stuff
```

But if the duplicate keys is NOT at top level, it isn't an error:

```yaml
blocks:
    fuel:
        duct:
            # things
        duct:
            # stuff
```

So, now we are in the extremely unenviable position that if we want to catch such things we need to write our own YAML-text-parsing tool to go through and find these not-top-level duplicates.

Because, after extensive testing, `ruamel.yaml` and `pyyaml` both silently pass on this and they have no tools to throw errors or even warnings in this case.

**NOTE**: This PR DOES fix the problem of duplicate keys at the top-level though, so we are still keeping this PR. I have tests to prove this works for top-level keys in the PR now.

john-science:
I also tried the `yamllint` library, to find our duplicates. But it follows the YAML spec, so it only sees top-level duplicates as errors. 

Sigh.

john-science:
@keckler This PR still has value, but I have removed the original ticket from it. As I don't think we can close that ticket.

armi/reactor/blueprints/tests/test_blueprints.py
```diff
@@ -199,6 +199,25 @@ class TestBlueprintsSchema(unittest.TestCase):
...
+        for sectionName in sections:
+            # modify blueprint YAML to duplicate this section
+            yamlString = str(self._yamlString)
+            i = yamlString.find(sectionName)
+            lenSection = yamlString[i:].find("\n\n")
+            section = yamlString[i : i + lenSection]
+            yamlString = yamlString[:i] + section + yamlString[i : i + lenSection]
+
+            # validate that this is now an invalid YAML blueprint
+            with self.assertRaises(Exception):
```
keckler, line 218: Is there not a more specific exception that we can assert here?

keckler (approved):

PR #1238, opened 2023-04-11
Add 1D Cylindrical Modeling to the XS Group Manager

jakehader:
## Description

Implement a block collection type for managing 1D cylindrical
control rod modeling to ensure that number densities do not migrate
to other components. This is avoided by building a representative
block and setting number densities directly onto the components
instead of using `b.setNumberDensities()`. Setting number densities
directly on the blocks is not a good technique if the blocks are to
be treated heterogeneously because there is a chance that material
will migrate to other regions based on the "dehomogenization" logic
that is applied.

---

## Checklist

- [x] This PR has only one purpose or idea.
- [x] Tests have been added/updated to verify that the new/changed code works.

- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html).

- [x] The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any bug fixes or new features.
- [x] The documentation is still up-to-date in the `doc` folder.
- [x] The dependencies are still up-to-date in `setup.py`.

PR #1363, opened 2023-07-26
Add nucTempHelper to CylindricalComponentsAverageBlockCollection

mgjarrett:
## What is the change?

This PR adds a `_getNucTempHelper` method to `CylindricalComponentsAverageBlockCollection`

The `CylindricalComponentsAverageBlockCollection` was implemented recently: https://github.com/terrapower/armi/pull/1238

It needs to have a `_getNucTempHelper` method like some of the other block collections, e.g., the `AverageBlockCollection`, in order to perform downstream analyses that require this method.

---

## Checklist


armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -295,6 +295,8 @@ def test_ComponentAverage1DCylinder(self):
 
         xsgm.interactBOL()
 
+        xsgm.updateNuclideTemperatures()
```
keckler, line 298: I guess this is just a check to see that the `NotImplementedError` doesn't arise? Not sure if it's easy to add more to this test?

keckler (approved):
This looks good.

At first I thought that the 1D slab class would need a similar `_getNucTempHelper` method. But looking at that class, it seems like it is hard-coded to only ever work for isothermal reactors, so such an averaging process isn't needed. Is that your assessment as well?

mgjarrett:
> This looks good.
> 
> At first I thought that the 1D slab class would need a similar `_getNucTempHelper` method. But looking at that class, it seems like it is hard-coded to only ever work for isothermal reactors, so such an averaging process isn't needed. Is that your assessment as well?

Yeah, the fact that it explicitly raises a `NotImplementedError` for `_getNucTempHelper` even though it inherits the same thing from `BlockCollection` makes me think that somebody really doesn't want it to be implemented. I don't think it would be harmful to implement one, but there's also no apparent need for it, so I'm happy to just maintain status quo here.

keckler:
Is there something blocking this from being merged? It is a feature that I think is needed for our current analyses and its been sitting up here for months now. I have been manually pulling this branch into my installation because I need it, so it'd be great if we can just merge it.

albeanth:
@keckler it's dependent on a larger downstream PR. It should get merged soon.

Issue #1134, opened 2023-01-31
Provide a Compact class to hold particle fuel and matrix


drewj-usnctech:
Something that came up in a chat today: having a `Compact` composite class with strictly two children: a matrix `Component` and something to represent the particle fuel - #1055. The matrix would define the shape of the compact (cylinder, potentially spherical for pebble bed reactors). The `Compact` would / could be a child of a `Block` along with it's sibling `Component` children. Would require what we're calling arbitrarily deep composite tree (#1008) or at least a little deeper.

Thinking of the MHTGR-350 reactor (#224) you could think of a fuel element as a `Block` with the following children

```
block
  - circular coolant component
  - fuel compact composite
    - circular matrix component
    - particle fuel composite
      - spherical fuel kernel
      - spherical buffer
      - spherical ipyc
      - ...
``` 
Then, if the compact is placed on a lattice grid, the multiplicity of the compact is easy to get: number of lattice sites. And the matrix could have a multiplicity of (1 - packing fraction) to account for the volume occupied by the particle fuel. And the particle fuel (and subsequent layers) would have a multiplicity on the order of thousands to reflect the number of particles in one compact

_Originally posted by @drewj-usnctech in https://github.com/terrapower/armi/discussions/960#discussioncomment-4792352_

drewj-usnctech:
@john-science this and #1055 are something we at Tech are interested in exploring

cc @slee-USNC @cbroman-usnctech

PR #1369, opened 2023-08-01
Don't validate material modifications if they are None

### LINKED ISSUE #1367: Material modifier validation fails if modifications are valid for some but not all blocks in an assembly
Consider an assembly with two blocks, one for fuel and one for an axial reflector. The material modifier for this assembly might look like
```yaml
U235_wt_frac:
  - 0.05
  -
```
As I understand how the yaml snippet above is intended to behave, no `U235_wt_frac` is passed to material modifications in the upper block. Or maybe it passes `U235_wt_frac=None`? This feels like it should be fine?

Changes from #1122 and #1151 help catch material modifiers that don't apply to any material in the block. Which is excellent, and has helped us uncover some bugs already! But in this specific case, we get an error saying
```
ValueError: Block reflector has invalid material modification key: U235_wt_frac
```
Which is technically correct, but does not allow us to have blocks in an assembly that have non-overlapping sets of modifiers.

 A more problematic case would be passing a non-None value
```yaml
U235_wt_frac:
  - 0.05
  - 0.05 # input error, copy-paste from fueled assembly
```
implying I want to modify U235 in the axial reflector, which does not contain materials that support that.

The logic in #1122 and #1151 is already complex, trying to gather and compare all material modifications block-by-block. Should it instead or in addition look at modifiers in the entire assembly?

drewj-usnctech:
## What is the change?

Material modifiers are not validated if the corresponding value is `None`

During block construction, an error can be raised if a material modifier does not apply to any components in the block. This is great, and has already caught some errors. However, if a material modifier applies to _some_ components in _other_ blocks, and it not applied to a block without components that support that modifier, an error is raised.

Consider two blocks in an assembly, one fuel and the other axial reflector. The enrichment could be set in the fueled block with `U235_wt_frac`, but setting a U235 enrichment in an the axial reflector does not make sense. The following YAML input
```yaml
U235_wt_frac:
  - 0.05
  -
```
would support this case, but the material validation fails because no component in the axial reflector supports the `U235_wt_frac` modifier. 

Closes #1367 

## Checklist

  - Open to talking testing

  - Maybe not necessary?


john-science:
Is `None` something we want to support? 

Why not throw an error here and force people to set it to `0.0`?

What is the benefit of supporting `None`?  It isn't even a valid YAML file any more.

john-science:
Okay, I can jive with the high-level idea for this PR.  Let's do it.

But this might be an invalid YAML file that `ruamel.yaml` isn't fighting us on. But even if that's so, I bet the `rumael.yaml` team won't change their opinion on this, and they will support it forever.

IF they do drop support for it, I won't be able to fight that.  We would then just lose this feature. But I am betting the chances of that are low.

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -241,6 +229,41 @@ def construct(
                 runLog.warning(str(e), single=True)
         return b
 
+    def _getBlockwiseMaterialModifierOptions(
```
john-science, line 232: Can this method get a docstring?

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -241,6 +229,41 @@ def construct(
...
+        self, children: Iterable[Composite]
+    ) -> Set[str]:
+        validMatModOptions = set()
+        for c in children:
+            perChildModifiers = self._getCompositeMaterialModifiers(c)
+            validMatModOptions.update(perChildModifiers)
+        return validMatModOptions
+
+    def _getCompositeMaterialModifiers(self, c: Composite) -> Set[str]:
```
john-science, line 241: Can this method get a docstring?

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -241,6 +229,41 @@ def construct(
...
+                        signature(
+                            materialParentClass.applyInputParams
+                        ).parameters.keys()
+                    )
+        # self is a parameter to class methods, so it gets picked up here
+        # but that's obviously not a real material modifier
+        perChildModifiers.discard("self")
+        return perChildModifiers
+
+    def _getMaterialsInComposite(self, child: Composite) -> Iterator[Material]:
```
john-science, line 258: Ditto. Can this method get a docstring?

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -241,6 +229,41 @@ def construct(
...
+                # we must loop over parents as well, since applyInputParams
+                # could call to Parent.applyInputParams()
+                if issubclass(materialParentClass, Material):
+                    perChildModifiers.update(
+                        signature(
+                            materialParentClass.applyInputParams
+                        ).parameters.keys()
+                    )
+        # self is a parameter to class methods, so it gets picked up here
+        # but that's obviously not a real material modifier
```
john-science, line 254: This. Is. Wild.

john-science:
@drewj-usnctech Looks good!  I know I technically made a bunch of comments. But they're all small.

Looks suuuper close to me.

armi/reactor/blueprints/blockBlueprint.py
drewj-usnctech: There are some unit tests that use the composite group / component group feature. So there are things that currently live on a block that are not all _strict_ `Component`s.

This also looks toward particle fuel modeling (#1134) where there could be a non-`Component` living on the block

armi/reactor/blueprints/blockBlueprint.py
drewj-usnctech: Yeah it was wild during debugging seeing `self` as modifier collected by looking at the signatures of the various `Material.applyInputParameter` methods

john-science:
@drewj-usnctech This PR is super close. Just FYI.

drewj-usnctech:
> @drewj-usnctech This PR is super close. Just FYI.

Thanks @john-science. I'll try to get back to it by end of the week

john-science (approved):

PR #1370, opened 2023-08-01
Removing global variable from runLog.py

john-science:
## What is the change?

I removed the global `_WHITE_SPACE` variable from `runLog.py`.

Other than getting rid of an ugly global variable, I was responding to [this dicussion](https://github.com/terrapower/armi/discussions/1360), started by Arrielle. Good call!

---

## Checklist

- [X] This PR has only one purpose or idea. (For real this time.)

### DIFF
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -61,7 +61,6 @@ _ADD_LOG_METHOD_STR = """def {0}(self, message, *args, **kws):
     if self.isEnabledFor({1}):
         self._log({1}, message, args, **kws)
 logging.Logger.{0} = {0}"""
-_WHITE_SPACE = " " * 6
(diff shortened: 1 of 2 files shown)


opotowsky (approved):
Wow, thanks for taking this on so quickly. 

Just a sanity check: this line gives a "default" for the global variable (`_WHITE_SPACE = " " * 6`), but that was never used, correct?

john-science:
> Just a sanity check: this line gives a "default" for the global variable (`_WHITE_SPACE = " " * 6`), but that was never used, correct?

That was never used directly. But it turns out that's the value that gets calculated for nearly all cases in the `MPI_RANK=0` parent process.

Issue #1278, opened 2023-05-24
Location of pins in tips-up hexagonal grid is off by 60 degrees


drewj-usnctech:
I have a tips up grid with fuel pins and one non-fuel pin on a tips-up hexagonal grid. The lattice map is
```yaml
  pins:
    geom: hex_corners_up
    symmetry: full
    lattice map: |
      - - O F F
       - F F F F
        F F F F F
         F F F F
          F F F
```
where `O` is the non-fuel and `F` is the fuel pin.

As I read it, the non-fuel pin should be in the top left of the hexagonal grid.  But, the locator attached to that component has a `getLocalCoordinates` of `[-2.0, 0, 0]`, which implies it is located at due-left, 60 degrees off of where it should be.

<details><summary>Blueprint</summary>

```yaml
blocks:
  fuel: &block_fuel
    grid name: pins
    duct:
      shape: Hexagon
      material: HT9
      Tinput: 450.0
      Thot: 450.0
      ip: 16
      op: 16.5
    coolant:
      shape: DerivedShape
      material: Sodium
      Tinput: 600.0
      Thot: 600.0
    fuel:
      shape: Circle
      material: UZr
      Tinput: 900
      Thot: 900
      od: 1
      latticeIDs: [F]
    other:
      shape: Circle
      material: HT9
      od: 1
      Tinput: 900
      Thot: 900
      latticeIDs: [O]
assemblies:
  fuel:
    specifier: F
    height: [10]
    axial mesh points: [1]
    blocks: [*block_fuel]
    xs types: [A]
systems:
  core:
    grid name: core
    origin:
      x: 0.0
      y: 0.0
      z: 0.0
grids:
  core:
    geom: hex
    symmetry: full
    lattice map: |
      F
  pins:
    geom: hex_corners_up
    symmetry: full
    lattice map: |
      - - O F F
       - F F F F
        F F F F F
         F F F F
          F F F
nuclide flags:
  NA: &nuc_flags {burn: false, xs: true}
  U: *nuc_flags
  ZR: *nuc_flags
  MO: *nuc_flags
  MN: *nuc_flags
  SI: *nuc_flags
  CR: *nuc_flags
  V: *nuc_flags
  C: *nuc_flags
  FE: *nuc_flags
  W: *nuc_flags
  NI: *nuc_flags
```

</details> 

<details><summary>Processing script</summary>

```python
import armi

armi.configure()

o = armi.init(fName="sample.yaml")

b = o.r.core[0][0]

[...]

PR #1373, opened 2023-08-04
Refactor armi.reactor.grids to wider submodule; Grids as ABC

drewj-usnctech:
## What is the change?

Mainly two changes here, one largely invisible to users/developers, one maybe controversial (in a312849fc95d2f539819d636e56357670c6d916d)

### Refactor

 Refactoring `armi/reactor/grids.py` into "wider" submodule `armi.reactor.grids`.  Dedicated files for specific grid types, e.g., `armi.reactor.grids.hexagonal` provides `HexGrid`. All classes and utilities are imported into `armi/reactor/grids/__init__.py` so there should be no API-breaking changes up to a312849fc95d2f539819d636e56357670c6d916d.

I tried to keep the commits small so you could see the process was essentially
- `git mv armi/reactor/grids.py armi/reactor/grids/__init__.py`
- copy some class from `armi/reactor/grids/__init__.py` to new file `armi/reactor/grids/class.py`
- reimport class from `armi/reactor/grids/class.py` into `armi/reactor/grids/__init__.py` so there are no API changes

In this process, I also added an `AxialGrid` class, the return type from the (proposed deprecated) `armi.reactor.grids.axialUnitGrid` function.

### Abstraction
Introduced in a312849fc95d2f539819d636e56357670c6d916d

Heavier lift, but the classes defined in `armi.reactor.grids`follow this inheritance pattern

- `Grid(abc.ABC)` :: abstract interface for _a_ grid, or things that reside in
  space
  - `StructuredGrid(Grid)` :: still abstract, but for things that have
    some structure, known to ARMI as unit steps and bounds
    - `HexGrid(StructuredGrid)`
    - `CartesianGrid(StructuredGrid)`
    - `ThetaRZGrid(StructuredGrid)` - though how "structured" is this
      really?
    - `AxialGrid(StructuredGrid)`

In some cases, the shared interface between structured grids falls
apart. For example, `HexGrid.pitch` returns a single float, while
`CartesianGrid.pitch` is a 2-ple of float, one for x and one for y.
And the concept of ring and position appears to be stressed since the
inception of cartesian grids. But parts of ARMI require an interface
and so an attempt was made to define that interface.

The decision was made to not have `StructuredGrid` the base class, and
to introduce `Grid` as the base intentionally. There are cases for some
multiphysics codes where a grid may still be "structured" but not with
the level of structure the previous `Grid` implementations expect. One
example would be an irregular cartesian grid, where the unit cell
spacing in one or more dimensions may change. So the cell spacing is a
function of the cell position in that dimension.

This is more a breaking change because classes that used to inherit from `armi.reactor.grids.Grid` may not have the same unit stepping / bounds behavior they would get instead by subclassing the new `armi.reactor.grids.StructuredGrid`

Primary motivation is #1278 : fixing orientation of tips up hex grids. My proposed plan, introduced in [this comment](https://github.com/terrapower/armi/issues/1278#issuecomment-1648732062) is to have `HexGrid` be "orientation agnostic" and instead provide two subclasses, e.g.,  `FlatsUpHexGrid` and `TipsUpHexGrid`. 

> Looking through the code, it seems the spatial locators attached to these components have correct (i, j) indices. But the problem falls through HexGrid.getRingPos as that seems (after minor investigation) to assume a flats-up orientation.

> Looking through more of the HexGrid implementation, making HexGrid.getRingPos more aware of the orientation seems to be the tip of the iceberg. It seems like multiple methods make this assumption

This also helps clients understand more simply the hex grid orientation, something that isn't really feasible now (see [this comment](https://github.com/terrapower/armi/pull/730#discussion_r913942220))

## Checklist

  - maybe 1.5 ideas? it's a refactor, but like a refactor+ 

  - TBD depending on acceptance of ABC grid. Otherwise it's relatively transparent from an API perspective 


armi/reactor/grids/tests/test_grids.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 TerraPower, LLC
+# Copyright 2023 TerraPower, LLC
```
john-science, line 1: Can you put this header in the other files, so CI passes?

armi/reactor/grids/tests/test_grids.py
```diff
@@ -12,14 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Tests for grids."""
+"""Tests for grids"""
```
john-science, line 15: Can you run `ruff` on this PR?

john-science: This is not mandatory.

armi/reactor/grids/hexagonal.py
```diff
@@ -0,0 +1,432 @@
+from math import sqrt
```
john-science, line 1: This needs one of the license headers at the top.

armi/reactor/grids/hexagonal.py
```diff
@@ -0,0 +1,432 @@
...
+        (+COS30, -SIN30),
+    ]
+)
+
+
+class HexGrid(StructuredGrid):
+    """
+    Has 6 neighbors in plane.
+
+    .. note::
```
john-science, line 37: The way the ARMI docs are build, we use this format:

```
Notes
-----
```

The `.. note::` doesn't usually render the way we like.

I Juuuuust noticed we have 13 of these left in the repo's docstrings.  Whoops.  I will have to go fix those.

armi/reactor/grids/locations.py
```diff
@@ -0,0 +1,480 @@
...
+
+from typing import Optional, TYPE_CHECKING, Union, Hashable, Tuple, List, Iterator
+from abc import ABC, abstractmethod
+import math
+
+import numpy
+
+if TYPE_CHECKING:
+    # Avoid some circular imports
+    from . import Grid
```
john-science, line 23: I see you.

This is one of the options we've discussed. As I say, at some point we'll have to make the decision and we will unify all documentation in ARMI. 

But there are some discussions ahead first.

armi/reactor/grids/locations.py
```diff
@@ -0,0 +1,480 @@
...
+
+
+IJType = Tuple[int, int]
+IJKType = Tuple[int, int, int]
+
+
+class LocationBase(ABC):
```
john-science, line 30: Being able to actually use `ABC` without breaking the API is so nice.  How lovely.

I just assumed trying this in the repo would break the API.  I guess I should look around to see where else I could do this safely.

Just wonderful.

john-science:
@drewj-usnctech Awesome! I love it.  I have a couple of little cleanup things above, then we can merge this.

drewj-usnctech:
Some weird pip behavior that feels like a github / docker / larger issue?
```
 .pkg: _exit> python /opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
  cov2: FAIL code 1 (1221.10=setup[62.95]+cmd[1158.15] seconds)
```

john-science:
> Some weird pip behavior that feels like a github / docker / larger issue?
> 
> ```
>  .pkg: _exit> python /opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
>   cov2: FAIL code 1 (1221.10=setup[62.95]+cmd[1158.15] seconds)
> ```

@opotowsky Is this a Python 3.11 issue?

opotowsky:
> > Some weird pip behavior that feels like a github / docker / larger issue?
> > ```
> >  .pkg: _exit> python /opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
> >   cov2: FAIL code 1 (1221.10=setup[62.95]+cmd[1158.15] seconds)
> > ```
> 
> @opotowsky Is this a Python 3.11 issue?

I'm seeing an actual failure, actually

But why in only python 3.11?

image

john-science:
@drewj-usnctech OR this might not be Python 3.11, this might just be a plain unit test failure:

https://github.com/terrapower/armi/actions/runs/5965597782/job/16183385153?pr=1373

```
  File "/home/runner/work/armi/armi/armi/reactor/grids/tests/test_grids.py", line 216, in test_ringPosFromIndicesIncorrect
    grid = grids.Grid(unitSteps=((1.0, 0.0, 0.0), (0.0, 1.0, 0.0), (0.0, 0.0, 1.0)))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Can't instantiate abstract class Grid with abstract methods __len__, backUp, getCellBase, getCellTop, getCoordinates, getSymmetricEquivalents, isAxialOnly, items, locatorInDomain, overlapsWhichSymmetryLine, reduce, restoreBackup
```

drewj-usnctech:
Nope that's on me. I just missed that `grids.Grid` in the tests 🤦

john-science (approved):
I love a good refactor!

john-science:
@drewj-usnctech It looks like the problem above that broke the unit tests also [broke the docs](https://github.com/terrapower/armi/actions/runs/5978350890/job/16220251937):

```
/home/runner/work/armi/armi/doc/gallery-src/framework/run_grids2_cartesian.py failed leaving traceback:
  File "/home/runner/work/armi/armi/doc/gallery-src/framework/run_grids2_cartesian.py", line 34, in <module>
    cartesian_grid = grids.Grid(
TypeError: Can't instantiate abstract class Grid with abstract methods __len__, backUp, getCellBase, getCellTop, getCoordinates, getSymmetricEquivalents, isAxialOnly, items, locatorInDomain, overlapsWhichSymmetryLine, reduce, restoreBackup
```

CI is currently broken in ARMI.

armi/reactor/grids/locations.py
albeanth: And so it begins! 😈 😅

armi/reactor/grids/structuredgrid.py
```diff
@@ -0,0 +1,511 @@
...
+    @staticmethod
+    def _meshBaseByBounds(index, bounds):
+        if index < 0:
+            raise IndexError("Bounds-defined indices may not be negative.")
+        return bounds[index]
+
+    @staticmethod
+    def getNeighboringCellIndices(i, j=0, k=0):
+        """Return the indices of the immediate neighbors of a mesh point in the plane."""
+        return ((i + 1, j, k), (1, j + 1, k), (i - 1, j, k), (i, j - 1, k))
```
mgjarrett, line 365: @drewj-usnctech the second element looks wrong. Should be `(i, j + 1, k)`, not `(1, j + 1, k)`, right?

I came across this while writing a unit test for this with a Cartesian grid: 

https://github.com/terrapower/armi/blob/14c951a05214154a1e6bfb43c728aa714084cea7/armi/reactor/reactors.py#L1676-L1678

mgjarrett: https://github.com/terrapower/armi/pull/1614

drewj-usnctech: https://github.com/terrapower/armi/pull/1614#issuecomment-1915149546

PR #1376, opened 2023-08-07
Add New Component Axial Linking Approach

### LINKED ISSUE #769: Resolving axial component linkages for blocks with multiple identical component definitions
In the current implementation of the `axialExpansionChanger`, linkages between axially-continuous components are inferred via a set of criteria:
https://github.com/terrapower/armi/blob/a4b54cac85dbf90ee8b28aa94fce51ab31593efc/armi/reactor/converters/axialExpansionChanger.py#L480-L529

However, at least one case slips through the cracks of this logic. If one defines a block with multiple independent pins that have the same exact geometry and material class, but that utilize different enrichments (or something else defined via the material modifications). 

In such a case, the `axialExpansionChanger` will not be able to disambiguate the identical components, and when trying to link them to the components in the blocks above/below, it won't know which ones to link up and it will crash.

This might sound like an unlikely scenario, but I am trying to model more than one assembly that has these properties.

My first thought is that in such a complicated scenario, we might force the user to use an explicitly-defined grid for their block, as is demonstrated in the LWR blueprints tutorial:
https://terrapower.github.io/armi/tutorials/walkthrough_lwr_inputs.html

That would make the various pin positions completely unambiguous, and the linkage relationships should be completely obvious.

@albeanth @jakehader @onufer

albeanth:
## What is the change?
Closes #769. A request was made to improve the axial linking logic to take advantage of the block grids system in ARMI. This PR aims to accomplish that. While previous behavior is maintained, there are several changes:
1. AssemblyAxialLinkage::_getLinkedComponents now allows for multiple component axial linkage. Any components with multiple linkages are resolved in AxialExpansionChanger::retrieveLinkedComponent. 
2. AxialExpansionChanger::retrieveLinkedComponent accounts for 3 cases to try and resolve the multiple axial linkage. See test_axialExpansionChanger.py::TestRetriveAxialLinkage for details. 
3. Tests have been added and modified to account for new grid block behavior.
     i. A block grid was added to armi/tests/detailedAxialExpansion/refSmallCoreGrid.yaml
     ii. Additional complexity was added to the simple assembly defined in test_blockBlueprints.py.  

To support a user request for downstream analyses.

---

## Checklist

- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html). Keckler let me know that commit history isn't important, so the world gets the working branch! 

- The documentation is still up-to-date in the `doc` folder. Another item for #935.... 


armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -488,96 +508,130 @@ def _getLinkedComponents(self, b, c):
...
 
-    Parameters
-    ----------
-    componentA : :py:class:`Component <armi.reactor.components.component.Component>`
-        component of interest
-    componentB : :py:class:`Component <armi.reactor.components.component.Component>`
-        component to compare and see if is linked to componentA
+        Notes
+        -----
+        If componentA and componentB are both solids and the same type, geometric overlap can be checked via
```
keckler, line 542: I think your change lost the check on whether the components are solids or not. Am I wrong about that?

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -488,96 +508,130 @@ def _getLinkedComponents(self, b, c):
...
+            ) and isinstance(componentB.spatialLocator, MultiIndexLocation):
+                ## Case 2 -- see docstring
+                componentAIndices = [
+                    list(index) for index in componentA.spatialLocator.indices
+                ]
+                componentBIndices = [
+                    list(index) for index in componentB.spatialLocator.indices
+                ]
+                # check for common indices between components. If either component has indices within its counterpart,
+                # then they are candidates to be linked and overlap should be checked.
```
keckler, line 580: Just thinking out loud here... It seems like this logic will not allow for partial-length fuel rods, if I'm reading it correctly. That's not an issue for me right now, but just trying to consider common cases (partial-length rods are common in BWRs).

armi/tests/detailedAxialExpansion/refSmallCoreGrid.yaml
```diff
@@ -19,3 +19,26 @@ core:
               IC  IC  MC  OC
             AF  IC  IC  PC  SH
   symmetry: third periodic
+
+fourPin:
+  # geom: hex
```
keckler, line 24:
```suggestion
```

keckler (changes requested):
This looks very good. I like that your new test provides a good blueprint for somebody to implement this in their model.

I think the only thing that requires real attention is that I think the logic no longer tests if the components are solids. It'd be great if you could check on that.

keckler:
I pushed some changes to this branch so that the axial expansion changer can properly handle blocks that don't have any solid components. This was leading to issues with assemblies accidentally changing overall height, which is not supposed to happen (the DUMMY block is supposed to compress to allow for the overall assembly height to remain unchanged).

I also added a test on this fix.

keckler:
I have absolutely no clue why some of the `bookkeeping` related tests are failing. I am not seeing any failures when I run locally on 3.9, and I don't think that I made any changes that are related to the operation of the database code.

keckler:
I do not know why this branch is now seeing some test failures in the `bookkeeping` module. I do not see any errors when I run those tests locally with 3.9, and I did not make any changes related to the `bookkeeping` module.

keckler:
Okay, after some offline discussion with @albeanth , the changes that I mentioned in the comment above have been rolled back. Instead of trying to handle them, we are explicitly going to block users from having blocks in their assemblies that don't have any solid components in them. In such scenarios, it is unclear how the axial expansion should be performed and so we just disallow it. The exceptions to this are:
1. The dummy block
2. An assembly with no solid components at all is allowed.

I added the appropriate logic and an associated test.

armi/reactor/converters/axialExpansionChanger.py
```diff
@@ -259,6 +265,33 @@ def _isTopDummyBlockPresent(self):
...
+            # the topmost block has already been confirmed as the dummy block
+            solidCompsInBlock = [
+                c
+                for c in b.iterComponents()
+                if not isinstance(c.material, material.Fluid)
+            ]
+            if len(solidCompsInBlock) == 0:
+                raise InputError(
+                    f"Assembly {self.linked.a} is constructed improperly for use with the axial expansion changer.\n"
+                    "Consider using the assemFlagsToSkipAxialExpansion case setting."
```
albeanth, line 292: Let's give a little more detail so users don't have to dig into the source code. E.g. "...is constructed improperly for use with the axial expansion changer; i.e., assemblies cannot have intermediate blocks void of a solid component. Consider using....

keckler: Well ideally there would be some documentation that outlines how to use the axial expansion capability ;-)

albeanth: Hah! Yeah yeah.... very true....

albeanth:
@keckler a subtle hint hint you want this sooner than later, eh? 😄

keckler:
I can't run my model without this branch, so at the very least I needed to get the merge conflicts resolved. But yes, sooner is better!

albeanth:
I will carve some out time this week to restart this effort. I know John would like to see this PR gone too!

albeanth:
I'll just close this for now. This PR is tracked elsewhere so it's not going to get lost. When this work gets prioritized I'll reopen it.

john-science:
@keckler @john-science We need to push this through.

maurelius167:
Hey @keckler & @albeanth,

Will you ask Dana and Casey if they can decrypt this: https://youtu.be/F_PmQFr5CTY?si=QsVNSxIgP8cNzO0S

They have no idea what's coming.

albeanth:
> @keckler @john-science We need to push this through.

this will come through ~6 months.

john-science:
> > @keckler @john-science We need to push this through.
> 
> this will come through ~6 months.

We will be bringing this one back!

albeanth:
And I was almost right with my 6 month estimate! 😄

PR #1382, opened 2023-08-09
Remove coveragepy-lcov dependency during CI.

### LINKED ISSUE #1377: Find a way to run `coverage` on python 3.11
Typically, github actions are set to run on the latest version of python that we have tested with ARMI. #1341 originally set out to bring all GH actions up to python 3.11, but a known issue with `coverage` (#1058) prevented us from updating the coverage dependency to a python 3.11 compatible version. 

We need to investigate the bug with the coverage version and see if it can be fixed.

I had coverage pinned to 7.2.1 in both setup.py and the coverage workflow, and the action failed. 

Update: I'm not sure what the above bug is all about, but `coveragepy_lcov` has not been updated in 1.5 years and is not python 3.11 compatible. So we will need to completely rework the cooverage step to make it run on python 3.11

ntouran:
## What is the change?

coveragepy-lcov is incompatible with coverage 7.0.0 and in fact is totally unnecessary now that pytest can write lcov format files directly. This switches pytest to write lcov directly. Now coverage should work in CI on python 3.11.

Fixes #1377

I have not fully tested this but tested the components. Will need to let CI run to see if it fully works I think.

We want coverage on python 3.11

---

## Checklist

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -31,17 +31,9 @@ jobs:
         run: tox -e cov1 || true
       - name: Run Coverage Part 2
         run: tox -e cov2
-      - name: Convert Coverage Results
-        env:
-          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
-        run: |
-          pip install click==8.1.3
-          pip install coverage==6.5.0
-          pip install coveragepy-lcov==0.1.2
-          coveragepy-lcov --data_file_path coverage_results.cov --output_file_path lcov.txt
       - name: Publish to coveralls.io
         uses: coverallsapp/github-action@v1.1.2
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          path-to-lcov: lcov.txt
+          path-to-lcov: coverage.lcov
 
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -8,19 +8,16 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-22.04
-
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
-
     steps:
       - uses: actions/checkout@v2
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: '3.10'
+          python-version: '3.11'
       - name: Update package index
         run: sudo apt-get update
       - name: Install mpi libs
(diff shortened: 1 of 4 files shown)


john-science:
I **LOVE** removing dependencies. Awesome.

And I note that the "coverage" step on GH Actions has already passed for this PR.

You rock, man. Thanks!

john-science (approved):

opotowsky (changes requested):
Two things that I can't comment inline:

1. In `.github/workflows/coverage.yaml` Can you update python 3.10 to 3.11?
2. In `setup.py`, you'll note a silly little python-version-specific coverage version update I made to get the first py311 PR merged. I think given this GH action cleanup we can revert that update, and just remove the coverage version pin altogether.

ntouran:
ok done. Part of me wonders if coverage belongs in the install_requires section at all, since it's not used at runtime, but I'm happy to clean out those pins for now.

john-science:
> ok done. Part of me wonders if coverage belongs in the install_requires section at all, since it's not used at runtime, but I'm happy to clean out those pins for now.

I turns out `coverage` is used in the base `Case` class:

https://github.com/terrapower/armi/blob/06bf1f1806dda8ea63ed832282693df3b768a37d/armi/cases/case.py#L41

https://github.com/terrapower/armi/blob/06bf1f1806dda8ea63ed832282693df3b768a37d/armi/cases/case.py#L66

https://github.com/terrapower/armi/blob/06bf1f1806dda8ea63ed832282693df3b768a37d/armi/cases/case.py#L353

And when I suggested we could pull that out because it's not used, I believe @jakehader lobbied quite hard that this was used and shouldn't be removed.  (I might be mis-remembering WHO lobbied so hard, but someone.)

opotowsky (approved):
Thanks for digging into this!

opotowsky:
> > ok done. Part of me wonders if coverage belongs in the install_requires section at all, since it's not used at runtime, but I'm happy to clean out those pins for now.
> 
> I turns out `coverage` is used in the base `Case` class:
> 
> https://github.com/terrapower/armi/blob/06bf1f1806dda8ea63ed832282693df3b768a37d/armi/cases/case.py#L41
> 
> https://github.com/terrapower/armi/blob/06bf1f1806dda8ea63ed832282693df3b768a37d/armi/cases/case.py#L66
> 
> https://github.com/terrapower/armi/blob/06bf1f1806dda8ea63ed832282693df3b768a37d/armi/cases/case.py#L353
> 
> And when I suggested we could pull that out because it's not used, I believe @jakehader lobbied quite hard that this was used and shouldn't be removed. (I might be mis-remembering WHO lobbied so hard, but someone.)

Actually this might have a viable use case and it's been on my ToDo backburner for a few months. So I'd also lobby to keep it until we know it's not useful!

PR #1383, opened 2023-08-09
Move max assembly number out of global scope

john-science:
## What is the change?

This PR moves the "maximum assembly number" out of the global scope and into a `Reactor` `Parameter`.  (The reason it's a `Parameter` and not just a class attribute is so we can store it in the Database.)

We have far too many global variables in ARMI.  Such a mess.  This particular change was spurred by a [discussion](https://github.com/terrapower/armi/discussions/1360) @opotowsky stared.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -727,6 +727,11 @@ class Database3:
         )
         root = comps[0][0]
 
+        if updateGlobalAssemNum:
+            raise DeprecationWarning(
+                "The method input `updateGlobalAssemNum` is no longer used."
+            )
+
         # return a Reactor object
         if cs[CONF_SORT_REACTOR]:
             root.sort()
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -101,18 +101,6 @@ def getH5GroupName(cycle: int, timeNode: int, statePointName: str = None) -> str
     return "c{:0>2}n{:0>2}{}".format(cycle, timeNode, statePointName or "")
 
 
-def updateGlobalAssemblyNum(r) -> None:
-    """
-    Updated the global assembly number counter in ARMI, using the assemblies
-    read from a database.
-    """
-    assemNum = r.core.p.maxAssemNum
-    if assemNum is not None:
-        assemblies.setAssemNumCounter(int(assemNum + 1))
-    else:
-        raise ValueError("Could not load maxAssemNum from the database")
-
-
 class Database3:
     """
     Version 3 of the ARMI Database, handling serialization and loading of Reactor states.
@@ -691,8 +679,7 @@ class Database3:
             Whether to emit a warning, rather than crash if reading a database
             with undefined parameters. Default False.
         updateGlobalAssemNum : bool, optional
-            Whether to update the global assembly number to the value stored in
-            r.core.p.maxAssemNum. Default True.
+            DeprecationWarning: This is unused.
         updateMasterCs : bool, optional
             Whether to apply the cs (whether provided as an argument or read from
             the database) as the primary for the case. Default True. Can be useful
@@ -740,10 +727,6 @@ class Database3:
             parameterCollections.GLOBAL_SERIAL_NUM, layout.serialNum.max()
         )
         root = comps[0][0]
-
-        # ensure the max assembly number is correct, unless the user says no
-        if updateGlobalAssemNum:
-            updateGlobalAssemblyNum(root)
 
         # return a Reactor object
         if cs[CONF_SORT_REACTOR]:
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -728,8 +728,9 @@ class Database3:
         root = comps[0][0]
 
         if updateGlobalAssemNum:
-            raise DeprecationWarning(
-                "The method input `updateGlobalAssemNum` is no longer used."
+            raise runLog.warning(
+                "The method input `updateGlobalAssemNum` is no longer used.",
+                single=True,
             )
 
         # return a Reactor object
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -70,7 +70,6 @@ from armi.bookkeeping.db.layout import (
 from armi.bookkeeping.db.typedefs import History, Histories
 from armi.nucDirectory import nuclideBases
 from armi.physics.neutronics.settings import CONF_LOADING_FILE
-from armi.reactor import assemblies
 from armi.reactor import grids
 from armi.reactor import parameters
 from armi.reactor import systemLayoutInput
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -728,7 +728,7 @@ class Database3:
         root = comps[0][0]
 
         if updateGlobalAssemNum:
-            raise runLog.warning(
+            runLog.warning(
                 "The method input `updateGlobalAssemNum` is no longer used.",
                 single=True,
             )
--- a/armi/bookkeeping/db/tests/test_database3.py
+++ b/armi/bookkeeping/db/tests/test_database3.py
@@ -81,7 +81,7 @@ class TestDatabase3(unittest.TestCase):
         self.assertEqual(sorted(self.db.h5db["c00n00"].keys()), sorted(keys))
 
         # validate availabilityFactor did not make it into the H5 file
-        rKeys = ["assemNum", "cycle", "cycleLength", "flags", "serialNum", "timeNode"]
+        rKeys = ["maxAssemNum", "cycle", "cycleLength", "flags", "serialNum", "timeNode"]
         self.assertEqual(
             sorted(self.db.h5db["c00n00"]["Reactor"].keys()), sorted(rKeys)
         )
--- a/armi/bookkeeping/db/tests/test_database3.py
+++ b/armi/bookkeeping/db/tests/test_database3.py
@@ -81,7 +81,14 @@ class TestDatabase3(unittest.TestCase):
         self.assertEqual(sorted(self.db.h5db["c00n00"].keys()), sorted(keys))
 
         # validate availabilityFactor did not make it into the H5 file
-        rKeys = ["maxAssemNum", "cycle", "cycleLength", "flags", "serialNum", "timeNode"]
+        rKeys = [
+            "maxAssemNum",
+            "cycle",
+            "cycleLength",
+            "flags",
+            "serialNum",
+            "timeNode",
+        ]
         self.assertEqual(
             sorted(self.db.h5db["c00n00"]["Reactor"].keys()), sorted(rKeys)
         )
--- a/armi/bookkeeping/db/tests/test_database3.py
+++ b/armi/bookkeeping/db/tests/test_database3.py
@@ -81,7 +81,7 @@ class TestDatabase3(unittest.TestCase):
         self.assertEqual(sorted(self.db.h5db["c00n00"].keys()), sorted(keys))
 
         # validate availabilityFactor did not make it into the H5 file
-        rKeys = ["cycle", "cycleLength", "flags", "serialNum", "timeNode"]
+        rKeys = ["assemNum", "cycle", "cycleLength", "flags", "serialNum", "timeNode"]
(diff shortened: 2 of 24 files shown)


armi/bookkeeping/db/database3.py
```diff
@@ -741,9 +727,11 @@ def load(
         )
         root = comps[0][0]
 
-        # ensure the max assembly number is correct, unless the user says no
         if updateGlobalAssemNum:
-            updateGlobalAssemblyNum(root)
+            runLog.warning(
+                "The method input `updateGlobalAssemNum` is no longer used.",
+                single=True,
+            )
```
ntouran, line 734: could also be a candidate for [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning)

armi/reactor/reactors.py
```diff
@@ -111,6 +111,40 @@ def add(self, container):
...
+
+        Returns
+        -------
+        int
+            The new max Assembly number.
+        """
+        self.p.maxAssemNum = 0
+
+        ind = self.core.normalizeNames(self.p.maxAssemNum)
+        self.p.maxAssemNum = ind
```
ntouran, line 141: why is this line done twice in sequence? Seems kinda confusing.

armi/bookkeeping/db/tests/test_database3.py
```diff
@@ -365,39 +371,6 @@ def test_loadSortSetting(self):
...
-        # is updated and equal to self.r.p.maxAssemNum + 1 which is equal to the number of
-        # assemblies in blueprints/core.
-        r = self.db.load(0, 0, allowMissing=True, updateGlobalAssemNum=True)
-        expected = len(self.r.core) + len(self.r.blueprints.assemblies.values())
-        self.assertEqual(15, expected)
-
-        # repeat the test above to show that subsequent db loads (with updateGlobalAssemNum=True)
-        # do not continue to increase the global assem num.
-        self.db.load(0, 0, allowMissing=True, updateGlobalAssemNum=True)
-        self.assertEqual(15, expected)
```
ntouran, line 399: A lot of this Assembly number management was designed to allow loading from a db multiple times in the same process. Can we maybe rename the test to be "loading multiple times works as expected" to cover the desire even though the implementation is changing? we don't want a regression in this capability.

armi/reactor/reactors.py
john-science: The first is `self.core` and the second is `self.sfp`.

As stated in the docstring.  The method does one first, and then the second, in an attempt to get all the Children (and other assemblies) in the Reactor.

armi/reactor/converters/geometryConverters.py
```diff
@@ -96,10 +96,6 @@ def reset(self):
         runLog.info(
             f"Resetting the state of the converted reactor core model in {self}"
         )
-        currentAssemCounter = assemblies.getAssemNum()
-        assemblies.setAssemNumCounter(
-            currentAssemCounter - len(self._newAssembliesAdded)
-        )
```
keckler, line 102: I understand that the `setAssemNumCounter()` machinery has been removed, so obviously it can no longer be called.

But don't we still need to fix the assembly number counter parameter on the reactor when moving from one core representation to another?

armi/reactor/reactors.py
```diff
@@ -111,6 +111,40 @@ def add(self, container):
...
+        Increase the max assembly number by one and return the current value.
+
+        Returns
+        -------
+        int
+            The new max Assembly number.
+        """
+        val = int(self.p.maxAssemNum)
+        self.p.maxAssemNum += 1
+        return val
```
keckler, line 125: This returns the old value, not the new one?

albeanth:
@john-science Though a big fan, has this been run through the slew of internal downstream tests? I imagine it might break some things. Not sure though.

armi/reactor/reactors.py
```diff
@@ -111,6 +111,40 @@ def add(self, container):
...
+        """
+        self.p.maxAssemNum = 0
+
+        ind = self.core.normalizeNames(self.p.maxAssemNum)
+        self.p.maxAssemNum = ind
+
+        ind = self.sfp.normalizeNames(self.p.maxAssemNum)
+        self.p.maxAssemNum = ind
+
+        return ind
```
keckler, line 146:
```suggestion
        self.p.maxAssemNum = self.sfp.normalizeNames(self.core.normalizeNames(0))
        
        return self.p.maxAssemNum
```

I _think_ that does the same thing that your code currently does. It might be more confusing though lol

At the very least, I think you can avoid the initial `self.p.maxAssemNum = 0` by just doing `ind = self.core.normalizeNames(0)` instead.

armi/reactor/reactors.py
```diff
@@ -531,6 +566,49 @@ def removeAllAssemblies(self, discharge=True):
...
+
+            ind += 1
+
+        # Update some bookkeeping dictionaries of assembly and block names in this Core.
+        self.assembliesByName = {}
+        self.blocksByName = {}
+        for assem in self:
+            self.assembliesByName[assem.getName()] = assem
+            for b in assem:
+                self.blocksByName[b.getName()] = b
```
keckler, line 609: Can we move these operations into the loop above? That would help us to avoid calling `*.getName()` many extra times.

armi/reactor/assemblies.py
```diff
@@ -93,11 +68,13 @@ def __init__(self, typ, assemNum=None):
...
-            The unique ID number of this assembly. If none is passed, the class-level
-            value will be taken and then incremented.
+            The unique ID number of this assembly. If None is provided, we generate a
+            random int. This makes it clear that it is a placeholder. When an assembly with
+            a negative ID is placed into a Reactor, it will be given a new, positive ID.
         """
+        # If no assembly number is provided, generate a random number as a placeholder.
         if assemNum is None:
-            assemNum = incrementAssemNum()
+            assemNum = randint(-9e12, -1)
```
keckler, line 77: Is there any reason to make this value a truly random integer? It seems like the fact that this is a placeholder is signified by the sign, not the randomness of the number. Generating a random number is unnecessarily expensive if we could instead just always pass a value of `-1`, for instance.

armi/reactor/tests/test_reactors.py
```diff
@@ -210,7 +208,7 @@ def loadTestReactor(
     if isPickeledReactor:
         # cache it for fast load for other future tests
         # protocol=2 allows for classes with __slots__ but not __getstate__ to be pickled
-        TEST_REACTOR = cPickle.dumps((o, o.r, assemblies.getAssemNum()), protocol=2)
+        TEST_REACTOR = cPickle.dumps((o, o.r, o.r.p.maxAssemNum), protocol=2)
```
keckler, line 211: I think it might be quite a headache to change, but I'm guessing that `o.r.p.maxAssemNum` no longer really needs to be a part of this, because it will be encompassed within `o.r` already, right?

keckler (commented):
This was more complicated than I expected. Some comments for you.

armi/reactor/reactors.py
john-science: This is definitely shorter. Is it more clear or easier to read?

I don't know.

I was actually thinking of going the opposite direction, and putting a comment on line 139 and 142 to describe the system more carefully.

> OH, also, no. You _need_ to reset `self.p.maxAssemNum = 0` before you do any of this logic. Or you will get edge cases where you start numbering at a non-zero value.

armi/reactor/reactors.py
john-science: I thought this was easier to read.

In fact, I was thinking of moving this "re-grab the names" block of code and moving it to a helper method.

You don't think it would  be an uglier loop the other way?  I guess I could do that, yeah.

keckler: I personally think the other way would be nicer to read, as it puts related operations closer to each other.

armi/reactor/assemblies.py
john-science: Hmm... this logic is used in a method called "make unique".  So I wanted to err on making each assembly _more_ unique.

You're right, of course. Much of the logic that exists when we add the Assembly to the Reactor would be the same if we used `-1`, or any negative number.

armi/reactor/tests/test_reactors.py
john-science: Oh, for sure. 

But a quick check says that one change breaks over 500 unit tests in downstream repositories.

So... it's a fine idea, but I would want to make that it's own PR, as it would be mostly an exercise in communication and kitten herding.

Happily, that is just cleaning up our tests, and can't affect anyone's runtime.

armi/reactor/converters/geometryConverters.py
john-science: Well, I didn't / don't think so.

I believe the chain of logic goes like this:

1. The original core gets built.
2. At this point, the `maxAssemNum` is whatever... above zero, Let's say "N".
3. Some converter (geometryConvert") is called
4. If the converter doesn't add new assemblies, we're fine: `maxAssemNum == N`
5. If the converter DOES add new assemblies (which is common), the new assembly numbers will start with `N` and go up.
6. The end result will be `maxAssemNum == M` where `M > N`.

I believe that situation is sufficient for our needs, correct, and the same as we have now.

BUT

If you think I'm just seeing the problem, speak up.  If I've missed something, I'm happy to improve the logic here.

keckler: I agree with your chain of logic steps.

But what about when the reactor gets converted back down? So maybe it started as a 1/3 core representation, we convert to full-core to do some CR worth calcs, and then bring it back down to 1/3 core to continue the remainder of the calculation sequence.

In that case, my expectation would be that `r.maxAssemNum` would be correspond to the 1/3 core representation for the remainder of the calculation sequence.

But that could be a fundamental misunderstanding of mine. I think both ways are legitimate, but I guess I don't know which one is "correct" for us.

I suppose in the past `maxAssemNum` was always monotonically increasing and it would include any temporary core conversions. Based on that, I guess that your current logic is correct.

jakehader:
Maybe way to late to the party, but why do we need a counter on the assembly numbers? Is this to give it a unique ID that has never existed before? Do we have a set of requirements for naming or numbering things that we can point to or develop as a part of this PR?

I am just thinking that maxAssemNumber is basically just the len(self). Is there a way to trigger parameter updates automatically like an event trigger?

john-science:
@jakehader 

> But why do we need a counter on the assembly numbers? Is this to give it a unique ID that has never existed before?

**I am not adding or removing any features from ARMI.**

The problem is that all assemblies and blocks in the reactor are numbered. But currently that number is kept in a global variable that is set when you do `import armi`.

And this is causing all kinds of problems:

1. It makes writing tests harder.
2. In particular, these global variables are causing testing trouble in some of our internal repos.
3. It means you can't have more than two reactors in Python at the same time.
    a . Or, you can, but you get impossible-to-solve invisible bugs that pop up during simulations.
4. It makes the code a LOT harder to debug.

So, chalk this up to "code quality and maintainability". None of this is a surprising: keeping important data in global variables is a classic bad code problem.

armi/reactor/converters/geometryConverters.py
john-science: ARMI does not currently preserve the assembly numbers when you go from 1/3 core to full core and back again.

In order to achieve that, we would need to do a lot of bookkeeping, At the very least we would need to keep a record of what conversions have been done.  And store a bunch of key information from the Reactor state before the conversion. Which we don't do.

Also, I imagine reverting a converter would be impossible after one time step of modeling takes place.

keckler (approved):

opotowsky:
@keckler @john-science @albeanth 

This PR was amazing! But it broke both unit and integration tests internally. I just wanted to give everyone some feedback on that so we can be better about knowing when to request that certain updates are tested internally. I definitely am also to blame -- I should have realized the potential for this PR to stir the pot.

armi/reactor/assemblies.py
jasonbmeng: Hey @john-science, this random number assignment for `makeUnique` is causing some regression test failures in OXBOW shuffling models. Could you help us understand what drove you towards random int assignment? Why not assign each assembly a unique and valid ID on instantiation (or when `makeUnique` is called) instead of overwriting the random negative int when the assembly is shuffled into the core?

cc @jeffbaylor @opotowsky

PR #1399, opened 2023-09-01
Removed getMasterCs and setMasterCs

### LINKED ISSUE #930: Reduce or Eliminate getMasterCs and setMasterCs

john-science:
## What is the change?

Removed all instances of `getMasterCs()` and `setMasterCs()`.

In particular, that meant changing the API:

* `Interface` has always taken `cs` as it's second parameter; this is no longer optional.
* `InputWriter` has always taken `cs` as it's second parameter; this is no longer optional.
* `OutputReader` now takes a fourth, named, parameter: `cs`.

In an effort to close #930, this PR removes a really problematic pieces of the ARMI API. Please see that ticket for an exhaustive discussion on WHY.

Also, this will be the last usage of the word "master" removed from the project. Finally.

---

## Checklist


john-science:
@sombrereau @opotowsky @drewj-usnctech 

This is an API-breaking change.  As such, I will not be merging it in willy-nilly.  I expect to help people understand the change and how to work with it. And I expect to do a lot of the fixing on internal repos myself.

armi/operators/operator.py
```diff
@@ -484,8 +483,6 @@ def _interactAll(self, interactionName, activeInterfaces, *args):
                     )
                 )
 
-            self._checkCsConsistency()
```
keckler, line 487: Instead of getting rid of this check entirely, do we instead just want to ensure that the `cs` on interface is consistent among them all?

armi/reactor/grids/__init__.py
```diff
@@ -62,6 +62,7 @@
 """
 from typing import Tuple, Optional
 
+# ruff: noqa: F401
```
keckler, line 65:
```suggestion
# noqa: F401
```
I think the syntax is this?

armi/settings/__init__.py
```diff
@@ -151,34 +151,3 @@ def promptForSettingsFile(choice=None):
...
-    intended to be shared by many other objects.
-
-    It can have multiple instances in multiprocessing cases.
-
-    Returns
-    -------
-    cs : Settings
-        The loaded cs object
-    """
-    cs = Settings.instance
```
keckler, line 170: Hmmm, what of this `Settings.instance` object? Should that go away as well?

keckler (commented):
Now that it's gone I have to wonder... Why was this feature here in the first place?

armi/reactor/grids/__init__.py
john-science: Actually, adding the prefix `ruff` makes the rule apply to the entire file.

And this file has about a dozen imports that aren't used in the FILE, but are part of the API.  So, we want to keep them.

I just think this is cleaner than having a dozen lines end with a `noqa` comment.

armi/settings/__init__.py
john-science: That's a really interesting point.  Let me try that.

armi/operators/operator.py
john-science: I can look at it again.

But it really looks like the only thing this function does is check that your CS object matches the master/global CS object.  So, I _expect_ that this function will suddenly do nothing once we remove the master/global CS object.  I'll look.

armi/settings/__init__.py
john-science: Done!

armi/operators/operator.py
john-science: Okay, unless you have another idea, this method simply compares `o.cs` to the global CS to see if they match.

Now that there is only one CS object possible, and not two, I don't think we need this method any more.

jakehader:
> Now that it's gone I have to wonder... Why was this feature here in the first place?

Convenience to get user configuration data anywhere, anytime, And all the time! It has caused some grief within Interfaces where they just change the behavior of the physics solvers (sometimes forgetting to turn them back to how the user wanted them!) to do things like have convergence criteria or boundary condition assumptions.

I think that removing global CS state marks a good level of maturing in the API and we should continue to codify the relationship between an `Interface`, the reactor data model, and an `Operator` on what they can and cannot do.

john-science:
> I'd be super surprised if this doesnt break something downstream for us...

@jakehader Yeah, I will not just merge this one willy-nilly.  I will look downstream and open a bunch of PRs to pre-empt this change before I merge it in.  For sure.  I won't take API changes lightly.

Thanks, man!

armi/bookkeeping/db/database3.py
```diff
@@ -680,10 +680,7 @@ def load(
         updateGlobalAssemNum : bool, optional
             DeprecationWarning: This is unused.
         updateMasterCs : bool, optional
-            Whether to apply the cs (whether provided as an argument or read from
-            the database) as the primary for the case. Default True. Can be useful
-            if you don't intend to use the loaded reactor as the basis for further
-            computations in the current operator.
+            TODO: Deprecated. Slated for removal.
```
opotowsky, line 683: does removing `updateMasterCs` here cause a lot of additional work? Not sure if it should be a part of the master cs cleanup or if it should wait.

opotowsky (approved):
Normally I'm more than OK with docstring/formatting fixes tossed into PRs, but I think they crowded out this very important/impactful PR a little bit. Not asking you to change anything, just some feedback for heavy hitting updates like this.

I am approving since I reviewed all the code, but I trust you'll get all the internal updates queued up as needed. VERY excited this is happening!

@albeanth When merged, will this PR fix your unit testing woes?

armi/bookkeeping/db/database3.py
john-science: I was trying to break the API in fewer ways, if possible. But I could look downstream and see how big a footprint adding this change in would make. Sure.

opotowsky: I'm assuming it isn't altered from the default ever/often but if it is then don't worry about it!

albeanth:
> @albeanth When merged, will this PR fix your unit testing woes?

I imagine so, yes. But only time will tell as that was a bug that seemed to only occasionally rear its head. 

@john-science before merging, please run this PR through the gamut of downstream tests. If they all come back clean, feel free to merge at will. Otherwise let's touch base about a merge plan. Thanks!

john-science:
> ...before merging, please run this PR through the gamut of downstream tests...

@albeanth I did that. But, of course, the downstream work is frozen for the moment, so we are freed from that constraint until February. 

Thanks!

PR #1401, opened 2023-09-05
Enforcing no-relative-imports via Ruff

john-science:
## What is the change?

This PR enforce the ruff rule to ban absolutely ALL relative imports. Also, there were a couple of docstrings violating our ruff rules, that I fixed.

A recent PR added some relative imports that @albeanth noted. So together we found a way to get ruff to enforce the relative imports rule we want.

---

## Checklist

### DIFF
--- a/armi/reactor/grids/__init__.py
+++ b/armi/reactor/grids/__init__.py
@@ -60,16 +60,17 @@ Throughout the module, the term **global** refers to the top-level coordinate sy
 while the word **local** refers to within the current coordinate system defined by the
 current grid.
 """
+# ruff: noqa: F401
 from typing import Tuple, Optional
 
-from .constants import (
+from armi.reactor.grids.constants import (
     BOUNDARY_CENTER,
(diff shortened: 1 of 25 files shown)


armi/reactor/grids/__init__.py
```diff
@@ -60,29 +60,30 @@
 while the word **local** refers to within the current coordinate system defined by the
 current grid.
 """
+# ruff: noqa: F401
```
albeanth, line 63: Do we need all of these unused imports?

john-science: Yeah. Totally.

There once was a single multi-thousand-line file: `grids.py` that was refactored beautifully by Drew into the dozen files in the directory you see here.

These "unused" imports are actually used ALL OVER ARMI and downstream repos.  They form the grids API.  If we removed these, there would be hundreds of changes in dozens of repos.

And this is just a nicer API to work with, this way.

albeanth (approved):

Issue #1354, opened 2023-07-20
Unreachable Code


albeanth:
Was stepping through a debugger and noticed that this was all unreachable code. It can be safely removed or if is something we actually want to keep should be relocated. 

https://github.com/terrapower/armi/blob/5912372a9c037f8b6caf52749ebe6dc28c06b873/armi/reactor/blocks.py#L1968-L1974

albeanth:
@john-science @jakehader

jakehader:
What a strange piece of code! Not even sure that method really should be on a block... Looks like a helper function if anything. Is this called on block initialization?

john-science:
Yeah, I'm in favor of just deleting this block of code.

mgjarrett:
It was probably just a simple indentation error. Is it possible that we actually wanted that error check, and it just needed to be bumped out one level of indentation?

PR #1403, opened 2023-09-07
fix indenting on deleted code block

albeanth:
## What is the change?
@mgjarrett brought up a decent point in #1354 after the fact. Though it hasn't been hit recently doesn't mean that it can't get hit. It seems reasonable enough to bring back.

I think it might be useful in case a user defines a pin design that won't fit in a duct. 

Edit: Turns out, the armi test reactor in armi/tests actually hit this ValueError! The blueprints had to be slightly modified. 

---

## Checklist

    - test_blocks.py could use some tidying... This is low priority so I'll leave this for now....

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -2023,7 +2023,7 @@ class HexBlock(Block):
         pinOuterFlatToFlat = (
             pinCenterFlatToFlat
             + clad.getDimension("od", cold=cold)
-            + wire.getDimension("od", cold=cold)
+            + 2.0 * wire.getDimension("od", cold=cold)
         )
         ductMarginToContact = duct.getDimension("ip", cold=cold) - pinOuterFlatToFlat
         pinToDuctGap = ductMarginToContact / 2.0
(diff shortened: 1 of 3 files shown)


armi/reactor/blocks.py
```diff
@@ -2018,7 +2023,7 @@ def getPinToDuctGap(self, cold=False):
         pinOuterFlatToFlat = (
             pinCenterFlatToFlat
             + clad.getDimension("od", cold=cold)
-            + 2.0 * wire.getDimension("od", cold=cold)
+            + wire.getDimension("od", cold=cold)
```
mgjarrett, line 2026: I guess this can go either way, but I thought the factor of 2.0 should remain here. I was considering the `pinToDuctGap < wireThicknesses` as the double-counting. It's algebraically the same as what I was thinking, but just a different way of defining `pinToDuctGap`

albeanth: Yeah you're right. That factor 2 should stay. Bonehead move on my part. 

Can you explain the double-counting thing some more? I'm not quite following.

PR #1404, opened 2023-09-08
Replace some cs["string"] with cs[CONF]

albeanth:
## What is the change?
Swapping out some (low priority) cs["string"] usages for cs[CONF_SETTING].

Consistency! Plus I'm tired of having to grep for the string when you can just alt-click on the CONF setting in vscode and jump straight to it. 

---

## Checklist

- ~Tests have been added/updated to verify that the new/changed code works.~

- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html). _well, the last commit, maybe not, but that's ok, right? lol_

-  ~The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any important changes.~

### DIFF
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -51,11 +51,17 @@ from armi.reactor.flags import Flags
 from armi.reactor.systemLayoutInput import SystemLayoutInput
 from armi.settings.fwSettings.globalSettings import (
     CONF_MATERIAL_NAMESPACE_ORDER,
+    CONF_FRESH_FEED_TYPE,
     CONF_SORT_REACTOR,
     CONF_GEOM_FILE,
     CONF_NON_UNIFORM_ASSEM_FLAGS,
     CONF_STATIONARY_BLOCK_FLAGS,
     CONF_ZONE_DEFINITIONS,
+    CONF_TRACK_ASSEMS,
+    CONF_CIRCULAR_RING_PITCH,
+    CONF_AUTOMATIC_VARIABLE_MESH,
+    CONF_MIN_MESH_SIZE_RATIO,
+    CONF_DETAILED_AXIAL_EXPANSION,
 )
 from armi.utils import createFormattedStrWithDelimiter, units
 from armi.utils import directoryChangers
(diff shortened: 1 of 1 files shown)


armi/reactor/reactors.py
```diff
@@ -49,8 +49,20 @@
...
+    CONF_TRACK_ASSEMS,
+    CONF_CIRCULAR_RING_PITCH,
+    CONF_AUTOMATIC_VARIABLE_MESH,
+    CONF_MIN_MESH_SIZE_RATIO,
+    CONF_DETAILED_AXIAL_EXPANSION,
+    CONF_GEOM_FILE,
+    CONF_NON_UNIFORM_ASSEM_FLAGS,
+    CONF_STATIONARY_BLOCK_FLAGS,
+    CONF_ZONE_DEFINITIONS,
+)
```
albeanth, line 65: This is admittedly a little bulky. Not terrible, but bulky. We could consider changing the API so that these setting defs live in their own file. That way we could just do something like:
```python
import armi.settings.fwSettings.globalSetting.definitions as defs
...
cs[defs.CONF_GEOM_FILE]
```
This has its own pros and cons. I'm definitely not attached to the idea. Just spitballing.

john-science:
I thought these were left out due to a circular import happening downstream.

Perhaps we should move the big block of imports into the method in question?

albeanth:
> I thought these were left out due to a circular import happening downstream.

I haven't noticed any issues with anything downstream. I haven't run the full gamut of tests, but so far it seems ok. I could very well be missing something though. 

> Perhaps we should move the big block of imports into the method in question?

Which method? Sorry, I'm confused on this one.

john-science:
> Which method? Sorry, I'm confused on this one.

You can see `def setOptionsFromCs(self, cs):` already has two imports for `CONF_` strings at the top of the method.  

As I (albeit vaguely) recall, this was because one of our internal processes imported ARMI and got a circular import error otherwise. Though, to be honest, I don't remember which project.

We can talk about testing offline, if you like.

john-science:
> Looks like #1185 dealt with the circular import issue.
> 
> Happy to chat offline or when @opotowsky gets back. Either way, this is so low on the priority list it's arguably not on there at all lol.

That PR fixed a circular import issue.

If we merged this PR it would cause the same circular import again.

So, can you please move the relevant imports you added so you can use them in `setOptionsFromCs()` inside the `setOptionsFromCs()` method?

Or, we can talk offline, sure.

john-science (approved):

PR #1410, opened 2023-09-13
Enforcing discrete options for neutronics BCs

### LINKED ISSUE #1407: Enforce boundary condition options
This logic here
https://github.com/terrapower/armi/blob/5c4c5cfd80434e66363bd849d4ac32d456277ea7/armi/physics/neutronics/settings.py#L144-L158

means that setting `boundaries: infinite` is not the same as setting `boundaries: Infinite`. This setting snuck through and caused a neutronics plugin to have weird behavior looking for exact matches to those settings.

Would it be acceptable to add `enforcedOptions=True` for this setting? Other physics plugins that need new boundary conditions should still be able to inject options by having their plugin `defineSettings` include an `Option` in the returned list

john-science:
## What is the change?

Enforcing the discrete list of a few options for the neutronics BCs.

**NOTE**: I also found some `ruff` errors from recent PRs that I fixed. And I removed a defunct `TODO`. The only real change is in the `settings.py` file.

To close #1407.  

As best I can tell, no one is using a BC for this setting that isn't in the list.  So, this setting would be safer if we constrained it to it's practical use case.

---

## Checklist


drewj-usnctech:
Thanks for hopping on #1407 so quick! I can't seem to find a fix in this MR though, it just appears to be docstring updates

john-science:
> Thanks for hopping on #1407 so quick! I can't seem to find a fix in this MR though, it just appears to be docstring updates

lawls.  I forgot to commit my actual fix change.  Hilarious.

Check the second commit on this PR.

PR #1411, opened 2023-09-13
Reducing warning spam from zones

### LINKED ISSUE #1408: Lots of warnings about not being able to find zones but nothing breaks
We leverage the zoning features for lots of physics routines, which means we often fall back to [`Zones.findZoneItIsIn`](https://github.com/terrapower/armi/blob/5c4c5cfd80434e66363bd849d4ac32d456277ea7/armi/reactor/zones.py#L375-L398) quite a lot.

This allows zones to not be found and will return `None`. Which isn't a problem for our applications because we don't do the thing if an assembly isn't in some zone. But this can flood the run log with messages like
```
[warn] Was not able to find which zone <fuel stack Assembly A0001 at 002-005> is in
[warn] Was not able to find which zone <fuel stack Assembly A0003 at 002-006> is in
...
```
due to

https://github.com/terrapower/armi/blob/5c4c5cfd80434e66363bd849d4ac32d456277ea7/armi/reactor/zones.py#L395-L398

Is the runlog call really necessary? Or should it be left up to the thing that uses `Zones.findZoneItIsIn` to determine if `None` is okay, worth reporting a warning, or raising an error?

john-science:
## What is the change?

Reducing some warning spam from the `findZoneItIsIn()` method.

A new user base has hit this message too many times. We were a bit too verbose about it to begin with, because we don't use this method much. close #1408 

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/assemblyBlueprint.py
+++ b/armi/reactor/blueprints/assemblyBlueprint.py
@@ -194,7 +194,7 @@ class AssemblyBlueprint(yamlize.Object):
 
     @staticmethod
     def _shouldMaterialModiferBeApplied(value) -> bool:
-        """Determine if a material modifier entry is applicable
+        """Determine if a material modifier entry is applicable.
 
         Two exceptions:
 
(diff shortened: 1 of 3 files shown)


albeanth (approved):
Easy-peasy. @drewj-usnctech sometimes the `single` option is magical, sometimes it's not. Let us know if this doesn't resolve the issue.

Issue #1374, opened 2023-08-07
[Suspect 🕵️‍♀️] - "De-homogenization" of Number Densities


jakehader:
For a long time now, we have allowed users and developers to use the `setNumberDensities` method that exists on the `composites` to set number densities on a `Component`, `Block`, `Assembly`, `Core` directly:

https://github.com/terrapower/armi/blob/e8ddcdf774f9eb3e842ae75948f095095fc2d48b/armi/reactor/composites.py#L1546-L1564

When this is called, the `updateNumberDensities` method is called and for non-components, the nuclides that are updated at distributed "evenly" throughout the children of the `composite`:

https://github.com/terrapower/armi/blob/e8ddcdf774f9eb3e842ae75948f095095fc2d48b/armi/reactor/composites.py#L1566-L1576

This works well when the user/developer does not care about maintaining the distribution of number densities below the level that the number densities are updated on, but can lead to unintended consequences if a down-stream user of the data model needs to know the material distributions.

I do not have a public example to show at the moment, but I will create one to demonstrate the issue. There are a couple solutions:

1. **Be opinionated in the implementation:** Only allow number densities to be set on an object that has no children. This would be a strict implementation and could lead to down-stream "breaking" changes but could also uncover hidden implementation bugs.
2. **Be flexible in the implementation:** Allow users/developers to communicate with the data model for _how_ they would like the number densities to be modified (i.e., provide a nuclide to children mapping as an input as optional, and default to "evenly" spreading out the nuclides as currently implemented).
3. **Explore another algorithm for the implementation:** It's not clear to me if the current implementation of using volume fractions of the children to "evenly" spread out nuclides is correct. In some cases, after you set number densities in this way you can ask a component for its `density()` and get a value that is much different than its `material.density(Tc=c.temperatureInC)`.

CC: @john-science, @drewejohnson, @albeanth, @opotowsky

drewj-usnctech:
Hey @jakehader, sorry for missing this. Might have been a wrong user-name tag 😜 

If I understand right, if I did something like
```python
core.setNumberDensities(data)
```
does it functionally unravel to
```python
for leaf in core.iterComponents():
    leaf.updateNumberDensities(data)
```
Oh but only updating nuclides that exist in those `leaf` children... I guess the name `update` was a choice vs. `set`?

Like this is a way to quickly build a homogenized core of a single material?

I think to get to your question, we're not using this so if it gets yanked / changed we won't be impacted

jakehader:
[...]

PR #1412, opened 2023-09-14
Representative block component temperature fix

mgjarrett:
## What is the change?

Number densities are set on a representative block to match the average number densities of all blocks in the `BlockCollection`. `block.setNumberDensities()` will smear the mass of a nuclide evenly over all components that contain the nuclide.

This PR aims to avoid the smearing when possible by setting the number densities on a component-by-component basis. This is only possible if all of the blocks in the `BlockCollection` have the same number of components and types of each component.

If tracking a large number of fission products, fuel components can often have trace amounts of isotopes that are typically associated with cladding, structural, or coolant material. When `block.setNumberDensities()` is called in this situation, the structural and coolant mass are smeared into the fuel components. A `BlockCollection` uses `block.setNumberDensities()` to set the number density on the representative block for that collection.

When the lattice physics writer later queries the representative block to get the average temperature of a structural nuclide, e.g., Fe-56, it performs a mass-weighted temperature of all the Fe-56 in the block. Since Fe-56 is in the fuel component of the representative block, the fuel component temperature will contribute to the average nuclide temperature of Fe-56. This is artificial and undesirable.

The solution here is to prevent the masses from being incorrectly smeared amongst the components by averaging number densities at the component level instead of the block level. The solution will not work if the `BlockCollection` has blocks of different types with different components; in this case, the code will revert to the old behavior.

This addresses one problem arising from the issue described in #1374, but this PR alone is not enough to close that issue. 

---

## Checklist


keckler:
Is this the problem case that #1374 was hinting at?

john-science:
@mgjarrett Have you tested to see what affect this has on our downstream modeling?  If you'd like, we can talk about that offline.

mgjarrett:
> Is this the problem case that #1374 was hinting at?

This is one instance of it. I can think of at least one other (uniform mesh converter), but there could also be more.

mgjarrett:
I updated one docstring and somehow all of the tests started failing. Is there some GH actions problem?

john-science:
> It looks like matplotlib 3.8.0 was released in the last couple days...

Yeah, that fix is in ARMI `main`. It looks like your PR's CI is running again.

albeanth:
> If tracking a large number of fission products, fuel components can often have trace amounts of isotopes that are typically associated with cladding, structural, or coolant material. 

I might get the answer to this during my review, but what causes this?

jakehader:
@albeanth this is related to my implementation changes to the fission product model with explicit fission products. Basically if anything is depletable then it gets all trace nuclides since the depletion chain isn't generically represented in the framework and is offloaded to a code like ORIGEN. To fix this we can make some assumptions about what should go into fuel versus other components but the most comprehensive way would be to have a detailed decay and transmutation chain implemented in the framework that could be traversed to ensure fission products are not assigned to clad, coolant, and other structural components.

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -119,6 +135,79 @@ def test_createRepresentativeBlock(self):
...
+        self.assertAlmostEqual(newBc.avgNucTemperatures["FE56"], 502.0)
+        self.assertAlmostEqual(newBc.avgNucTemperatures["NA23"], 402.0)
+
+    def test_createRepresentativeBlockDissimilar(self):
+
+        # add a different type of block
+        _o, r = test_reactors.loadTestReactor(
+            os.path.join(TEST_ROOT, "detailedAxialExpansion"),
+            inputFileName="armiRun.yaml",
+        )
```
albeanth, line 155: Instead of loading up an entire reactor for a single block you can use `loadTestBlock`
https://github.com/terrapower/armi/blob/d341941e570dc731a4b0ab9017acc65c35d43098/armi/reactor/tests/test_blocks.py#L75

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -119,6 +135,79 @@ def test_createRepresentativeBlock(self):
...
+
+        with mockRunLogs.BufferLog() as mock:
+            avgB = self.bc.createRepresentativeBlock()
+            self.assertIn(
+                "Non-matching block in AverageBlockCollection", mock.getStdout()
+            )
+
+        self.assertNotIn(avgB, self.bc)
+        # 0 + 1/4 + 2/4 + 3/4 + 4/4 =
+        # (0 + 1 + 2 + 3 + 4 + 8) / 6 = 18/6 = 3.0
```
albeanth, line 179: Can you improve the comment here?

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -319,7 +319,18 @@ def _makeRepresentativeBlock(self):
...
         newBlock.setLumpedFissionProducts(lfpCollection)
-        newBlock.setNumberDensities(self._getAverageNumberDensities())
+        # check if components are similar
+        if self._checkBlockSimilarity():
+            # set number densities on a component basis
+            for compIndex, c in enumerate(sorted(newBlock.getComponents())):
+                c.setNumberDensities(
+                    self._getAverageComponentNumberDensities(compIndex)
+                )
+                c.temperatureInC = self._getAverageComponentTemperature(compIndex)
```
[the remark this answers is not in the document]
mgjarrett: `setTemperature` will update the number densities for thermal expansion, but we just set the number densities to the value we wanted in the previous line.

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -359,6 +370,74 @@ def _getNucTempHelper(self):
...
+    def _getAverageComponentNumberDensities(self, compIndex):
+        """
+        Get weighted average number densities of a component in the collection.
+
+        Returns
+        -------
+        numberDensities : dict
+            nucName, ndens data (atoms/bn-cm)
+        """
+        nuclides = self.allNuclidesInProblem
```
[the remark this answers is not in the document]
mgjarrett: I think I was just following the pattern used for the `_getNucTempHelper` methods; I don't know if I had a particular reason when I initially did it.

Looking at it now, I could come up with a possible reason: you need all of the nuclides in all of the candidate blocks. It's possible that some blocks would have nuclides that others don't have. So you'd have to iterate over every block and update the list of nuclides.

This could still be done, and maybe it's better to do it that way. A little more code than the single line here, but it could be more efficient in some cases.

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -359,6 +370,74 @@ def _getNucTempHelper(self):
...
+            return numpy.mean(numpy.array([c.temperatureInC for c in components]))
+        else:
+            return (
+                weights.dot(
+                    numpy.array([c.temperatureInC * c.getMass() for c in components])
+                )
+                / avgComponentMass
+            )
+
+    def _checkBlockSimilarity(self):
```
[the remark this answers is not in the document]
mgjarrett: added a unit test

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -359,6 +370,74 @@ def _getNucTempHelper(self):
...
+        """
+        nuclides = self.allNuclidesInProblem
+        blocks = self.getCandidateBlocks()
+        weights = numpy.array([self.getWeight(b) for b in blocks])
+        weights /= weights.sum()  # normalize by total weight
+        components = [sorted(b.getComponents())[compIndex] for b in blocks]
+        ndens = weights.dot([c.getNuclideNumberDensities(nuclides) for c in components])
+        return dict(zip(nuclides, ndens))
+
+    def _getAverageComponentTemperature(self, compIndex):
```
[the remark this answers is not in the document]
mgjarrett: added a unit test

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -119,6 +135,79 @@ def test_createRepresentativeBlock(self):
         self.assertAlmostEqual(avgB.getNumberDensity("U235"), 2.0)
```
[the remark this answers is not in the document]
mgjarrett: I updated the comments to reflect how the reference values are calculated

mgjarrett:
@albeanth I added several unit tests and updated some of the docstrings and comments

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -359,6 +370,74 @@ def _getNucTempHelper(self):
             nv += nvBlock * wt
         return nvt, nv
 
+    def _getAverageComponentNumberDensities(self, compIndex):
```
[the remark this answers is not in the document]
mgjarrett: added a unit test

armi/physics/neutronics/crossSectionGroupManager.py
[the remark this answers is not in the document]
albeanth: I imagine the LFP doesn't have an actual material either so this probably wouldn't work. Ok cool. 
https://github.com/terrapower/armi/blob/4e4451ba7bdd00696ed1419d2c3acf24504482ca/armi/reactor/components/component.py#L597-L599

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -359,6 +370,74 @@ def _getNucTempHelper(self):
...
+        numberDensities : dict
+            nucName, ndens data (atoms/bn-cm)
+        """
+        blocks = self.getCandidateBlocks()
+        weights = numpy.array([self.getWeight(b) / b.getHeight() for b in blocks])
+        weights /= weights.sum()  # normalize by total weight
+        components = [sorted(b.getComponents())[compIndex] for b in blocks]
+        avgComponentMass = sum(c.getMass() for c in components) / len(components)
+        if avgComponentMass == 0.0:
+            # if there is no compponent mass (e.g., gap), do a regular average
```
albeanth, line 413: misspelled component in the comment

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -104,21 +105,198 @@ def setUp(self):
...
+        massWeightedIncrease = 5.0 / 3.0
+        baseTemps = [600, 400, 500, 500, 400, 500, 400]
+        expectedTemps = [t + massWeightedIncrease for t in baseTemps]
+        for compIndex, c in enumerate(b.getComponents()):
+            avgTemp = self.bc._getAverageComponentTemperature(compIndex)
+            self.assertAlmostEqual(
+                expectedTemps[compIndex],
+                avgTemp,
+                msg=f"{c} avg temperature {avgTemp} not equal to expected {expectedTemps[compIndex]}!",
+            )
```
albeanth, line 279: The test you provided here doesn't give coverage for the first block of the conditional (`if avgComponentMass = 0.0`) in getAverageComponentTemperature. If you swap out `b` with `test_block.loadTestBlock` and self.bc` with a local instance of `AverageBlockCollection`, then you can get full coverage.

john-science:
@mgjarrett The CI has been red on this PR for a week or two, so I hadn't been looking at it.  (My eyes tend to skip over PRs where the CI is red, it's not on purpose.)

It turns out it was just a merge conflict on the release notes. Easy.

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -104,21 +105,198 @@ def setUp(self):
...
+        self.assertAlmostEqual(newBc.avgNucTemperatures["U235"], 600.0)
+        self.assertAlmostEqual(newBc.avgNucTemperatures["FE56"], expectedIronTemp)
+        self.assertAlmostEqual(newBc.avgNucTemperatures["NA23"], expectedSodiumTemp)
+
+
+class TestComponentAveraging(unittest.TestCase):
+
+    def setUp(self):
+        fpFactory = test_lumpedFissionProduct.getDummyLFPFile()
+        self.blockList = makeBlocks(3)
```
albeanth, line 225: You can improve performance a little by moving `makeBlocks` to a `setUpClass` method. `makeBlocks` calls the `loadTestReactor` method. From the looks of it, that only needs to be called once here.

john-science:
@mgjarrett Since this branch is in your forked repo, I can fixed the `black` error:

```bash
would reformat armi/physics/neutronics/tests/test_crossSectionManager.py
````

File: armi/physics/neutronics/tests/test_crossSectionManager.py (hunk 10 above)
Comment (mgjarrett) [reply armi/physics/neutronics/tests/test_crossSectionManager.py:279]: I'm pretty sure it does cover it. I tested it by putting a print statement in the `if avgComponentMass == 0.0` block and it does enter the block.

File: armi/physics/neutronics/tests/test_crossSectionManager.py (hunk 11 above)
Comment (mgjarrett) [reply armi/physics/neutronics/tests/test_crossSectionManager.py:225]: Got it down from 2.5 s to 1.66 s for the 3 new tests!

Comment (mgjarrett) [reply armi/physics/neutronics/tests/test_crossSectionManager.py]: I used `setUpClass` for the other test classes also.

File: armi/physics/neutronics/tests/test_crossSectionManager.py (hunk 10 above)
Comment (albeanth) [reply armi/physics/neutronics/tests/test_crossSectionManager.py:282]: There is literally another test to cover this exact thing via test_getAverageComponentTemperatureNoMass`. 

Doh!

File: armi/physics/neutronics/crossSectionGroupManager.py [hunk 12]
```diff
@@ -370,22 +449,27 @@ def getBlockNuclideTemperatureAvgTerms(block, allNucNames):
...
     """
 
-    def getNumberDensityWithTrace(component, nucName):
-        # needed to make sure temperature of 0-density nuclides in fuel get fuel temperature
-        try:
-            dens = component.p.numberDensities[nucName] or TRACE_NUMBER_DENSITY
-        except KeyError:
-            dens = 0.0
-        return dens
+    def getNumberDensitiesWithTrace(component, allNucNames):
```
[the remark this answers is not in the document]
albeanth: we'll chat offline

albeanth (approved):

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -408,19 +411,30 @@ def _getAverageComponentTemperature(self, compIndex):
...
             return (
                 weights.dot(
                     numpy.array([c.temperatureInC * c.getMass() for c in components])
                 )
-                / avgComponentMass
+                / weightedAvgComponentMass
             )
 
-    def _checkBlockSimilarity(self):
+    def _averageByComponent(self):
```
albeanth, line 428: I think it's more clear to leave this as `_checkBlockSimilarity`. Having a method called `_averageByComponent` and an attribute called `averageByComponent` is a little confusing. Plus this methods functionality remains mostly unchanged so leaving its name makes sense to me.

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -281,6 +282,26 @@ def test_getAverageComponentTemperature(self):
...
 
+    def test_getAverageComponentTemperatureVariedWeights(self):
+        """
+        Test mass-weighted component temperature averaging with variable weights
+        """
+        # make up a fake weighting with power param
+        self.bc.weightingParam = "power"
+        for i, b in enumerate(self.bc):
+            b.p.power = i
+            print(b.getMass())
```
albeanth, line 293: print statement -> gotta go

albeanth (changes requested):

armi/physics/neutronics/crossSectionGroupManager.py
mgjarrett: I thought "_checkBlockSimilarity" didn't fully encompass what it was doing since it now checks `self.averageByComponent`

mgjarrett: I broke it out into a separate function

albeanth (approved):

PR #1416, opened 2023-09-26
copy interface inputs to return relative paths

opotowsky:
## What is the change?

This update is twofold:

1. Copy Interface Inputs now returns a relative file path when a file is copied rather than an absolute file path
2. Moves warnings about file copy to DEBUG 

1. When the directory structure changes, a settings file will no longer be valid. This prevents that
2. A lot of warnings print with case suites that have dependencies because the files don't yet exist. This moves those print outs to a more verbose run log level for user convenience.

Resolves Issue #1304 

---

## Checklist


john-science (approved):

opotowsky:
@john-science could you see this impacting anyone's workflow externally? Internally this is both tested and improves our workflow. This can be merged immediately if we don't think we need to warn anyone about this update.

john-science:
I merged this branch with `main`, and handled the merge conflicts.

I will probably merge this PR as soon as the CI passes again.

PR #1425, opened 2023-10-05
(API Breaking) Finish removal of block.r

albeanth:
## What is the change?
This follows up on #909 and fully removes the `block.r` property. Leaving it causes some (imo) unnecessary tangling. So instead of the current
```text
Reactor <--> Core <--> Assembly <--> Block <--> Component
   ^           ^                       |
   |-----------|-----------------------|
```
the following would be the new Composite relationships via properties
```text
Reactor <--> Core <--> Assembly <--> Block <--> Component
               ^                       |
               |-----------------------|
```
 

I can whip up some examples, but it causes some tangles with testing. Plug in the docstring it even says we should probably get rid of it. Plus you can still access the reactor object via `b.core`.

---

## Checklist


albeanth:
Since this is an API change, it'll need to be run through the gamut of downstream tests.

albeanth:
Converting to draft. This is looking like a fairly intrusive PR.

albeanth:
The coverage report doesn't seem right again. I'm seeing the same coverage report results over several PRs. Something is fishy with coveralls. Any ideas?

john-science:
> The coverage report doesn't seem right again. I'm seeing the same coverage report results over several PRs. Something is fishy with coveralls. Any ideas?

Fixed.

albeanth:
See the commit message for https://github.com/terrapower/armi/pull/1425/commits/ebbcdcb6610ae5a25c7fdf3075789ca7ae105a76

TL/DR: Leaving them on the composite left the complication of `self.r.core`. Which didn't work for blocks and blocked this from going through. Since no other `Composite` other than `Block` actually ever use those methods, they can safely get moved over  the `Block` class and `self.r.core` replaced with `self.core`.

armi/reactor/composites.py
```diff
@@ -1277,124 +1276,6 @@ def getNumberDensities(self, expandFissionProducts=False):
             return self._expandLFPs(numberDensities)
         return numberDensities
 
-    def getNeutronEnergyDepositionConstants(self):
```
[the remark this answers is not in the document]
albeanth: See the commit message for https://github.com/terrapower/armi/pull/1425#issuecomment-1758274425.

TL/DR: Those methods called `self.r.core` which blocked the PR from going through. After grepping ARMI and one major downstream project, those methods never get called on any `Composite` other than `Block`. So those methods can safely be moved over to the `Block` class and `self.r.core` can get replaced with `self.core`.

john-science:
To be clear, the reason I haven't tried to merge this is that we are currently in a code freeze. I am not ignoring you.

john-science:
@albeanth  This PR is getting a bit old; there are a bunch of conflicts.

Also, we should move the release note from `0.2.9` to `0.3.1`.

albeanth:
> This PR is getting a bit old; there are a bunch of conflicts.
> 
> Also, we should move the release note from `0.2.9` to `0.3.1`.

@john-science ok, the merge conflicts have been resolved and this feature branch is now up to date with `main`

armi/reactor/blocks.py
```diff
@@ -176,42 +177,6 @@ def core(self):
         c = self.getAncestor(lambda c: isinstance(c, Core))
         return c
```
john-science, line 178: It's still weird that we have `Block.core`, but fine. That's not for this PR.

Also... there is no error handling on `Block.core`? Mark that as sus.

armi/reactor/composites.py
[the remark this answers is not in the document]
john-science: Okay!  As long as we call this out in the "API changes" section of the release notes!

doc/release/0.2.rst
```diff
@@ -1,6 +1,37 @@
 ***********************
-ARMI v0.2 Release Notes
+ARMI v0.3 Release Notes
+=======================
+
+ARMI v0.3.1
+============
+Release Date: TBD
```
john-science, line 7: Whoops!

It looks like you accidentally over-wrote our `0.2` release notes file.  We have a totally different file for the `0.3` release notes.

I don't want to LOSE the old release notes, can you move this to the existing `doc/0.3.rst` files?

Thanks!

armi/reactor/blocks.py
albeanth: What sort of error handling do you have in mind?

john-science: What if the `Block` is not part of a `Core`?

You know, when you're building a `Reactor` from `Blueprints`, you have to create a `Block` and add it to the `Core`.  Or you might be doing some fuel shuffling, and not all `Assemblies`/`Blocks` are in the `Core` at the same time.  Or your `Block` might be in the `SpentFuelPool`.  Lots of situations.

But the question is... should that return `None`, or `raise Error()` or what?  Hmm... What works, but isn't heavy handed?  hmmm

albeanth: Hmmmm. Well `getAncestor` currently returns `None`. I suppose for the case where it's a SFP or something else, I could see a runLog or something being useful. 

Though resolving this feels like worthy of a separate PR. What do you think?

doc/release/0.2.rst
albeanth: Whoops, fixed! https://github.com/terrapower/armi/pull/1425/commits/c2cd59d54051d93b744c0c49e7912eab6b5458c1

armi/reactor/composites.py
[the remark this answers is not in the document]
albeanth: https://github.com/terrapower/armi/pull/1425/commits/c2cd59d54051d93b744c0c49e7912eab6b5458c1

john-science (approved):

PR #1444, opened 2023-10-19
Changing draft PR policy

john-science:
## What is the change?

I am changing our PR policy so opening a "Draft PR" is a fall back and not the norm.

People are abusing "Draft PRs", and it wastes huge amounts of my time.  I look at the ARMI PR list 3-6 times a day, every day, including weekends.  I can no longer support this.

People open a PR before their branch is even remotely finished.  And so, of course, most of their commits break the build and I get 3-8 emails.  In fact, 99.8% of all emails I receive from GitHub are Draft PRs that have been dropped and forgotten. I don't forget about them though, the ARMI PR list is central to all work on the project.

We have on draft PR that has been open for 7 months. There is also one user who currently had 5 PRs open, most of them draft.

There is no benefit to the current system, so I'm changing it.

---

## Checklist

### DIFF
--- a/doc/release/0.2.rst
+++ b/doc/release/0.2.rst
@@ -11,6 +11,7 @@ What's new in ARMI
 #. The ``_copyInputsHelper()`` gives relative path and not absolute after copy. (`PR#1416 <https://github.com/terrapower/armi/pull/1416>`_)
 #. ARMI now mandates ``ruff`` linting. (`PR#1419 <https://github.com/terrapower/armi/pull/1419>`_)
 #. Removed all old ARMI requirements, to start the work fresh. (`PR#1438 <https://github.com/terrapower/armi/pull/1438>`_)
+#. Downgrading Draft PRs as policy. (`PR#1444 <https://github.com/terrapower/armi/pull/1444>`_)
 #. TBD
 
 Bug fixes


opotowsky (commented):
I think this sounds like a fair policy. It is probably being done because of internal policies we have with one of our projects, but those procedures don't need to match. Only open the PR when the code is ready is a valid restriction to have.

@albeanth for visibility.

opotowsky (approved):

albeanth:
> There is also one user who currently had 5 PRs open, most of them draft

I mean, this is me lol. You can call me out. It's cool. 

What should be the guidance for PRs that get switched to draft that are due to schedule and priority shifts? Close and reopen when the time comes?

albeanth:
I ask because it won't be the last time this happens with one of my PRs - and I imagine some other authors will experience this as well, so it would be good to have clear guidance on this moving forward.

albeanth:
What about a policy that if a PR (open or draft) goes without any movement for 1 month (or choose your favorite time) it gets flagged as being inactive and slated for closure? I bet there's a github bot that could even automate this. Is that reasonable?

john-science:
> What about a policy that if a PR (open or draft) goes without any movement for 1 month (or choose your favorite time) it gets flagged as being inactive and slated for closure? I bet there's a github bot that could even automate this. Is that reasonable?
> 
> Such as: https://github.com/actions/stale and it has some docs! https://docs.github.com/en/actions/managing-issues-and-pull-requests/closing-inactive-issues

Whoa.

What an idea.

I mean, I'll have to look at the implementation of it and see if it works for us.  But if it does, that's a great utility.  Good find, man.

albeanth:
> I mean, we'd have to look at the implementation of it and see if it works for us. But if it does, that's a great utility. Good find, man.

I'm happy to try and help stand it up too. Since I'm one of the primary offenders and all 😄

john-science:
> I'm happy to try and help stand it up too. Since I'm one of the primary offenders and all 😄

The blame game is useless.

But I will take any help I can get. :)

PR #1446, opened 2023-10-20
Cleaning up bootstrap warning

john-science:
## What is the change?

The Python version warning in `_bootstrap.py` needed updating.

We really don't WANT people to use Python 3.7 any more.

**NOTE**: I feel like this file's existence is a red flag.  Other than this version thing, this file does two things:

* Hand-configure `tabulate` at a global level. So, I officially hate that library now.
* Build our global `nuclideBases`.

I don't like:

1. That those two things are global.
2. That we have to run so much code when someone does `import armi`. Do you think NumPy runs a ton of code at import time? Hell no.

---

## Checklist

- [X] No [requirements](https://terrapower.github.io/armi/developer/tooling.html#watch-for-requirements) were altered.

### DIFF
--- a/armi/_bootstrap.py
+++ b/armi/_bootstrap.py
@@ -12,25 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""
-Collection of code that needs to be executed before most ARMI components are safe to
-import.
-"""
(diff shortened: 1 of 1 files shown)


opotowsky (approved):
Agreed on the problematic nature of the global crud. 

But also, approved!

PR #1454, opened 2023-10-30
Test Parameter interaction with DB

bdlafleur:
## What is the change?

This PR adds tests for two Parameters related requirements. Specifically it tests the following:
- Allowing a Parameter to not be written with the saveToDB attribute (R_ARMI_RESTRICT_DB_WRITE)
- Allowing a Parameter to define its own serializer for writing to a DB file (R_ARMI_PARAM_SERIALISER)

Additionally, this PR fixes a few stray typos, and includes "armi-venv" as an ignored folder, consistent with the [installation instructions](https://terrapower.github.io/armi/installation.html).

The Parameters requirements must be tested to ensure requirements are met.

---

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ pytestdebug.log
 dump-temp-*
 dump-tests*
 .vim-bookmarks
+armi-venv/*
 venv*/
 .mypy_cache/
 **/__pycache__
--- a/armi/reactor/parameters/parameterDefinitions.py
+++ b/armi/reactor/parameters/parameterDefinitions.py
@@ -147,8 +147,8 @@ class Serializer:
         Defining a Serializer for a Parameter in part defines the underlying
         representation of the data within a database file; the data stored in a database
         are sensitive to the code that wrote them. Changing the method that a Serializer
-        uses to pack or unpack data may break compatibility with old databse files.
-        Therefore, Serializers should be dilligent about signalling changes by updating
+        uses to pack or unpack data may break compatibility with old database files.
+        Therefore, Serializers should be diligent about signalling changes by updating
         their version. It is also good practice, whenever possible, to support reading
         old versions so that database files written by old versions can still be read.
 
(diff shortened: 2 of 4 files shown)


john-science:
@bdlafleur Do you know about `black` and `ruff`?

This PR needs linter tweaks.

Also, it looks like you still need to sign your CLA.  Is that correct? Or is that an error?

.gitignore
```diff
@@ -52,6 +52,7 @@ pytestdebug.log
 dump-temp-*
 dump-tests*
 .vim-bookmarks
+armi-venv/*
```
john-science, line 55: Is this necessary?  It seems awfully specific to what you call your local venv.  Heck, the line below just has "venv", which you could have chosen.

armi/reactor/parameters/parameterDefinitions.py
```diff
@@ -575,6 +575,10 @@ def toWriteToDB(self, assignedMask: Optional[int] = None):
         """
         Get a list of acceptable parameters to store to the database for a level of the data model.
 
+        .. impl:: Implementation for restricting some parameters from being written to the database.
+            :id: I_ARMI_RESTRICT_DB_WRITE
+            :links: R_ARMI_RESTRICT_DB_WRITE
```
john-science, line 580: I think we decided we're not doing `..impl::` tags after all.

.gitignore
bdlafleur: I decided to leave this because this is what name the ARMI installation instructions advices you to name the environment folder. I recommend we change the installation instructions to "venv" if we remove this change.

armi/reactor/parameters/parameterDefinitions.py
albeanth: @john-science as I understand it and last I heard, they were optional. Has this changed recently?

.gitignore
albeanth: @bdlafleur another strategy is to have your venv in a completely separate folder outside of your ARMI clone. This way you won't have to change the gitignore and is flexible is/when you want multiple venvs (e.g., multiple python versions, different packages installed, etc).

armi/reactor/parameters/parameterDefinitions.py
john-science: @albeanth If the team isn't doing these at large, I don't want to be the one team comiting to the extra week(s) worth of work to do the implementation crumbs.

john-science (approved):

PR #1468, opened 2023-11-09
Adding test/impl tags for reactors

bdlafleur:
This PR adds test/impl tags and test for reactor.py requirements.

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_reactors.py
+++ b/armi/reactor/tests/test_reactors.py
@@ -238,7 +238,7 @@ class HexReactorTests(ReactorTests):
 
     def test_factorySortSetting(self):
         """
-        Create a core object from an input yaml
+        Create a core object from an input yaml.
 
         .. test:: Create core object from input yaml
             :id: T_ARMI_R_CORE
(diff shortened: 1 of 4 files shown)


bdlafleur:
There is no new source code added that would reduce coverage. Attributing failed checks to source spray.

john-science:

armi/reactor/tests/test_reactors.py
```diff
@@ -237,6 +237,13 @@ def setUp(self):
         )
 
     def test_factorySortSetting(self):
+        """
+        Create a core object from an input yaml.
+
+        .. test:: Create core object from input yaml
+            :id: T_ARMI_R_CORE
```
john-science, line 244: I love it.

But looking at the req, I think we need this line somewhere in the test:

```python
self.assertTrue(isinstance(r0.core, Composite))
```

which means we need something like this import at the time:

```python
from armi.reactor.composites import Composite
```

john-science (approved):
Great!

Issue #1469, opened 2023-11-09
Properly filter new flags when extending the Flag class


bdlafleur:
Currently, the armi.utils.flags.Flag._registerFields method, which is used when the Flag object is extended with new fields, performs an assert which does not test what I believe it intends to test.

`assert value not in cls._nameToValue`

This line should read:

`assert name not in cls._nameToValue`

PR #1470, opened 2023-11-09
Adding Flag/Flags impl and test tags

bdlafleur:
This PR adds tests and implementation tags for the Flag and Flags classes.
---

## Checklist

### DIFF
--- a/armi/reactor/flags.py
+++ b/armi/reactor/flags.py
@@ -163,12 +163,7 @@ def _fromStringIgnoreErrors(cls, typeSpec):
 
 
 def _fromString(cls, typeSpec):
-    """Make flag from string and fail if any unknown words are encountered.
-
-    .. impl:: Retrieve flag from a string
-        :id: I_ARMI_FLAG_TO_STR0
-        :implements: R_ARMI_FLAG_TO_STR
-    """
+    """Make flag from string and fail if any unknown words are encountered."""
 
     def updateMethod(typeSpec):
         try:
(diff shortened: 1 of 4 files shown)


bdlafleur:
There is no new source code added that would reduce coverage. Attributing failed checks to source spray.

john-science:

armi/reactor/flags.py
```diff
@@ -181,6 +181,10 @@ def _toString(cls, typeSpec):
     """
     Make flag from string and fail if any unknown words are encountered.
 
+    .. impl:: Convert a flag to string
+        :id: I_ARMI_FLAG_TO_STR1
+        :implements: R_ARMI_FLAG_TO_STR
```
john-science, line 186: We should delete this.  You have this req on the `def toString()` method below, that's the correct one.

The REASON we can't have this here is that we can't put requirement stubs on private methods.  (Any method/class/function that starts with a `_` is "private" and won't appear in our API docs.)

Sorry, does that make sense?

armi/reactor/tests/test_flags.py
```diff
@@ -27,8 +27,16 @@ def test_fromString(self):
         self.assertEqual(flags.Flags.fromStringIgnoreErrors("invalid"), flags.Flags(0))
 
     def test_toString(self):
+        """
+        Convert flag to and from string for serialization.
```
john-science, line 31: I like what you're doing here.

Maybe we should change the NAME of this test so it says "to" and "from"?

`test_flagsToAndFromString()` or something like that?

armi/utils/tests/test_flags.py
```diff
@@ -69,14 +69,46 @@ class F(Flag):
         f2 = F.from_bytes(array)
         self.assertEqual(f, f2)
 
-    def test_collision(self):
-        """Make sure that we catch value collisions."""
+    def test_collision_extension(self):
+        """Ensure the set of flags cannot be programmatically extended if duplicate created.
+
+        TODO: This test should fail when extend bug is fixed.
```
john-science, line 75: Okay, this is going to be odd, but it's in our docs.

I understand there are a bunch of `TODO`s in ARMI, I get that.

But our rule is that no NEW `TODO` statement can be added to the repo.  If you are trying to add a new TODO, that means either:

* You need to complete your code before your PR, it's not complete yet.
* There is an existing bug in the repo, in which case we need to open a ticket.

The _goal_ of the rule is to improve the code quality of the repo, which was problematic in the past.  Does that make sense?

Thanks!

bdlafleur: @john-science I've opened a new ticket with this issue noted ([#1469](https://github.com/terrapower/armi/issues/1469)).

john-science (approved):

PR #1471, opened 2023-11-10
Adding tests/impl tags for Utils

bdlafleur:
This PR adds test/impl taags for Utils

---

## Checklist

### DIFF
--- a/armi/utils/densityTools.py
+++ b/armi/utils/densityTools.py
@@ -24,6 +24,10 @@ def getNDensFromMasses(rho, massFracs, normalize=False):
     """
     Convert density (g/cc) and massFracs vector into a number densities vector (#/bn-cm).
 
+    .. impl:: Get number densities
+        :id: I_ARMI_UTIL_MASS2N_DENS0
+        :implements: R_ARMI_UTIL_MASS2N_DENS
+
     Parameters
     ----------
     rho : float
@@ -51,6 +55,10 @@ def getNDensFromMasses(rho, massFracs, normalize=False):
 def getMassFractions(numberDensities):
     """
     Convert number densities (#/bn-cm) into mass fractions.
+
+    .. impl:: Get number densities
+        :id: I_ARMI_UTIL_MASS2N_DENS1
+        :implements: R_ARMI_UTIL_MASS2N_DENS
 
     Parameters
     ----------
@@ -168,6 +176,10 @@ def formatMaterialCard(
     """
     Formats nuclides and densities into a MCNP material card.
 
+    .. impl:: Create MCNP material card
+        :id: I_ARMI_UTIL_MCNP_MAT_CARD
+        :implements: R_ARMI_UTIL_MCNP_MAT_CARD
+
     Parameters
     ----------
     densities : dict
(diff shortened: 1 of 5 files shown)


john-science:

armi/utils/densityTools.py
```diff
@@ -52,6 +56,10 @@ def getMassFractions(numberDensities):
     """
     Convert number densities (#/bn-cm) into mass fractions.
 
+    .. impl:: Get number densities
+        :id: I_ARMI_UTIL_MASS2N_DENS1
+        :implements: R_ARMI_UTIL_MASS2N_DENS
+
```
john-science, line 62: I think we can delete this one.  I believe the `getNDensFromMasses()` function above is the only function this Req is pointing at, and you already have that covered.

armi/utils/tests/test_densityTools.py
```diff
@@ -100,6 +107,13 @@ def test_applyIsotopicsMix(self):
         self.assertAlmostEqual(uo2.massFrac["O"], massFracO)  # non-HM stays unchanged
 
     def test_getMassFractions(self):
+        """
+        Number densities to mass fraction.
+
+        .. test:: Number densities to mass fractions
+            :id: T_ARMI_UTIL_MASS2N_DENS0
+            :tests: R_ARMI_UTIL_MASS2N_DENS
```
john-science, line 115: Much like how above I said this req doesn't go with this function, I think this test (of this function) doesn't go with this req.

We can just remove this `..test::` tag here.

armi/utils/tests/test_densityTools.py
```diff
@@ -108,6 +122,13 @@ def test_getMassFractions(self):
         self.assertAlmostEqual(massFracs["U234"], 0.07937081219437897)
 
     def test_calculateNumberDensity(self):
+        """
+        Mass fraction to number density.
+
+        .. test:: Number densities from mass fraction
+            :id: T_ARMI_UTIL_MASS2N_DENS1
+            :tests: R_ARMI_UTIL_MASS2N_DENS
```
john-science, line 130: We should remove this `..test::` tag.

This req is only implemented by the function `getNDensFromMasses()`, so for this req we only need to worry about tests for that function.

armi/utils/tests/test_hexagon.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2019 TerraPower, LLC
```
john-science, line 1: Can we change this `2019` to `2023`?

Thanks!

armi/utils/tests/test_hexagon.py
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Test hexagon tools."""
+import unittest
+
+from armi.utils import hexagon
+import math
```
john-science, line 18: This whole file is great and I love that you did this.

But, just FYI, if you look at the rest of the files in ARMI you will see we put the imports at the top of the file, in a certain order:

```python
import "standard library things"

import "third party library things"

import "local things"
```

Does that make sense?

So, in this case, the imports here would look like:

```python
import math
import unittest

from armi.utils import hexagon
```

john-science (approved):

PR #1476, opened 2023-11-14
Adding impl/test crumbs for blueprints

bdlafleur:
## What is the change?

This PR adds impl/test crumbs for the blueprints constructs.

---

## Checklist

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -780,6 +780,10 @@ class Case:
         for a parameter sweep or migration) to be written out as input
         for a forthcoming case.
 
+        .. impl:: Write blueprints settings to disk
+            :id: I_ARMI_BP_TO_DB
+            :implements: R_ARMI_BP_TO_DB
+
         Parameters
         ----------
         sourceDir : str (optional)
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -780,10 +780,6 @@ class Case:
         for a parameter sweep or migration) to be written out as input
         for a forthcoming case.
 
-        .. impl:: Write blueprints settings to disk
-            :id: I_ARMI_BP_TO_DB
-            :implements: R_ARMI_BP_TO_DB
-
         Parameters
         ----------
         sourceDir : str (optional)
--- a/armi/cases/tests/test_cases.py
+++ b/armi/cases/tests/test_cases.py
@@ -103,12 +103,7 @@ class TestArmiCase(unittest.TestCase):
             )
 
     def test_independentVariables(self):
-        """Ensure that independentVariables added to a case move with it.
-
-        .. test:: Write blueprints settings to disk
-            :id: T_ARMI_BP_TO_DB0
-            :tests: R_ARMI_BP_TO_DB
-        """
+        """Ensure that independentVariables added to a case move with it."""
         geom = systemLayoutInput.SystemLayoutInput()
         geom.readGeomFromStream(io.StringIO(GEOM_INPUT))
         bp = blueprints.Blueprints.load(BLUEPRINT_INPUT)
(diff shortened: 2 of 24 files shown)


armi/cases/case.py
```diff
@@ -780,6 +780,10 @@ def writeInputs(
         for a parameter sweep or migration) to be written out as input
         for a forthcoming case.
 
+        .. impl:: Write blueprints settings to disk
+            :id: I_ARMI_BP_TO_DB
+            :implements: R_ARMI_BP_TO_DB
+
```
john-science, line 786: I think this should be moved to `saveToSTream()`:

https://github.com/terrapower/armi/blob/main/armi/reactor/blueprints/gridBlueprint.py#L525

And I think the text should read something like:

> .. impl:: Write a blueprint file from a blueprint object.

And we should move the `.. test::` crumb for this, too.

armi/cases/tests/test_cases.py
```diff
@@ -103,7 +103,12 @@ def test_summarizeDesign(self):
             )
 
     def test_independentVariables(self):
-        """Ensure that independentVariables added to a case move with it."""
+        """Ensure that independentVariables added to a case move with it.
+
+        .. test:: Write blueprints settings to disk
+            :id: T_ARMI_BP_TO_DB0
+            :tests: R_ARMI_BP_TO_DB
```
john-science, line 110: The statement of this req is:

> The blueprints package shall allow the user to produce a valid blueprints file from an in-memory blueprint object.

So, I think the `saveToStream()` function is probably what we want, and this test isn't so related.

armi/cases/tests/test_cases.py
```diff
@@ -408,7 +413,12 @@ def test_buildCommand(self):
 
 
 class TestExtraInputWriting(unittest.TestCase):
-    """Make sure extra inputs from interfaces are written."""
+    """Make sure extra inputs from interfaces are written.
+
+    .. test:: Write blueprints settings to disk
+        :id: T_ARMI_BP_TO_DB1
+        :tests: R_ARMI_BP_TO_DB
```
john-science, line 420: Similar to my last to comments, I think this test should be moved to `saveToStream()`.

armi/reactor/blueprints/tests/test_blueprints.py
```diff
@@ -87,7 +87,12 @@ def test_specialIsotopicVectors(self):
         self.assertAlmostEqual(mox["PU239"], 0.00286038)
 
     def test_componentDimensions(self):
-        """Tests that the user can specifiy the dimensions of a component with arbitray fidelity."""
+        """Tests that the user can specify the dimensions of a component with arbitrary fidelity.
+
+        .. test:: Test component dimensions created correctly
```
john-science, line 92: I think we should change this flavor text here to something like:

> .. test:: Test that a component can be correctly created from a blueprint file.

armi/reactor/blueprints/tests/test_gridBlueprints.py
```diff
@@ -381,6 +381,12 @@ def test_simpleRead(self):
             self.assertEqual(gridDesign4.gridContents[-4, -3], "1")
 
     def test_simpleReadLatticeMap(self):
+        """Read lattice map and create a grid.
+
+        .. test:: Define a lattice map in reactor core
+            :id: T_ARMI_BP_GRID
+            :tests: R_ARMI_BP_GRID
```
john-science, line 388: I like it.  I think there should ALSO be a `.. test::` crumb for this req on the `test_tiny_map()` test above.

(Though they will need slightly different test IDs, of course.)

john-science (approved):
Awesome!

PR #1489, opened 2023-11-22
Moving test class crumbs to unit test crumbs

bdlafleur:
## What is the change?
This PR moves any test crumbs from classes into unit tests.

This is required because Sphinx will not render crumbs in Class level comments.
---

## Checklist

### DIFF
--- a/armi/cases/tests/test_cases.py
+++ b/armi/cases/tests/test_cases.py
@@ -202,7 +202,7 @@ class TestArmiCase(unittest.TestCase):
         """
         Test running a case.
 
-        .. test:: Generic mechanism to allow simulation runs
+        .. test:: There is a generic mechanism to allow simulation runs.
             :id: T_ARMI_CASE
             :tests: R_ARMI_CASE
         """
@@ -300,7 +300,7 @@ class TestCaseSuiteDependencies(unittest.TestCase):
         """
         Test the checkInputs() method on a couple of cases.
 
-        .. test:: Test the checkInputs method.
+        .. test:: Check the ARMI inputs for consistency and validity.
             :id: T_ARMI_CASE_CHECK
             :tests: R_ARMI_CASE_CHECK
         """
--- a/armi/cases/tests/test_cases.py
+++ b/armi/cases/tests/test_cases.py
@@ -411,6 +411,13 @@ class TestCaseSuiteDependencies(unittest.TestCase):
         self.assertIn(self.c1, self.c2.dependencies)
 
     def test_explicitDependency(self):
+        """
+        Test dependencies for case suites.
+
+        .. test:: Dependence allows for one case to start after the completion of another.
+            :id: T_ARMI_CASE_SUITE
+            :tests: R_ARMI_CASE_SUITE
+        """
         self.c1.addExplicitDependency(self.c2)
 
         self.assertIn(self.c2, self.c1.dependencies)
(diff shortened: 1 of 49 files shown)


armi/cases/tests/test_suiteBuilder.py
```diff
@@ -46,19 +46,20 @@ def __call__(self, cs, bp, geom):
...
         assert builder.modifierSets == []
 
     def test_buildSuite(self):
+        """
+        Initialize an LHC suite.
+
+        .. test:: A generic mechanism to allow users to modify user inputs in cases.
+            :id: T_ARMI_CASE_MOD0
+            :tests: R_ARMI_CASE_MOD
+        """
```
keckler, line 62: I have no issue with this test tag being moved here, but I wanted to comment that I think we're not testing the right thing in general here. 

Nobody uses this latin hypercube sampler, to my knowledge. We are more typically using the `FullFactorialSuiteBuilder` or `SeparateEffectsSuiteBuilder` classes. It would seem that we should instead be tagging tests for those, though I don't seem to see any tests for those classes.

So maybe we need to add some tests, and then also tag those new tests to satisfy this R_ARMI_CASE_MOD requirement.

Let me know if that's unclear.

armi/materials/tests/test_materials.py
```diff
@@ -935,16 +935,16 @@ def test_propertyValidTemperature(self):
...
 
     MAT_CLASS = materials.Lead
 
     def test_volumetricExpansion(self):
+        """Unit tests for lead materials.
+
+        .. test:: There is a base class for fluid materials.
+            :id: T_ARMI_MAT_FLUID2
+            :tests: R_ARMI_MAT_FLUID
+        """
```
keckler, line 947: I think the tag needs to be moved into the `test_linearExpansion()` method, where it is proved that the linear expansion coefficient is zero (which is the defining characteristic of a fluid within the context or ARMI).

armi/cases/tests/test_cases.py
```diff
@@ -269,12 +269,7 @@ def test_clone(self):
 
 
 class TestCaseSuiteDependencies(unittest.TestCase):
-    """CaseSuite tests.
-
-    .. test:: Dependence allows for one case to start after the completion of another.
-        :id: T_ARMI_CASE_SUITE
-        :tests: R_ARMI_CASE_SUITE
```
keckler, line 276: Did this test tag just disappear?

I think you can move it down to `test_explicitDepdendency()`.

keckler (approved):
Awesome, thanks for the new unit tests!

armi/cases/tests/test_suiteBuilder.py
```diff
@@ -78,3 +96,78 @@ def test_addDegreeOfFreedom(self):
...
+        assert builder.modifierSets[3][0].value == 2
+        assert builder.modifierSets[3][1].value == 4
+
+        assert builder.modifierSets[4][0].value == 1
+        assert builder.modifierSets[4][1].value == 5
+
+        assert builder.modifierSets[5][0].value == 2
+        assert builder.modifierSets[5][1].value == 5
+
+        assert len(builder.modifierSets) == 6
```
john-science, line 137: Can you change these to use `self.assertEquals()`?

All of our unit tests use that logic; it reports better through PyTest.

Thanks!

bdlafleur: @john-science Converted asserts to self.asserts.

PR #1503, opened 2023-11-28
Adding tests and crumbs for high-level framework concepts

john-science:
## What is the change?

Adding tests and crumbs for high-level framework concepts.

This is part of on-going requirement work.

---

## Checklist

### DIFF
--- a/armi/operators/operator.py
+++ b/armi/operators/operator.py
@@ -402,10 +402,6 @@ class Operator:
     def _performTightCoupling(self, cycle: int, timeNode: int, writeDB: bool = True):
         """If requested, perform tight coupling and write out database.
 
-        .. impl:: The operator shall allow tight coupling between physics systems.
-            :id: I_ARMI_OPERATOR_PHYSICS
-            :implements: R_ARMI_OPERATOR_PHYSICS
-
         Notes
         -----
         writeDB is False for OperatorSnapshots as the DB gets written at EOL.
--- a/armi/operators/tests/test_operators.py
+++ b/armi/operators/tests/test_operators.py
@@ -56,11 +56,23 @@ class InterfaceC(Interface):
     name = "Third"
 
 
-# TODO: Add a test that shows time evolution of Reactor (R_EVOLVING_STATE)
 class OperatorTests(unittest.TestCase):
     def setUp(self):
         self.o, self.r = test_reactors.loadTestReactor()
         self.activeInterfaces = [ii for ii in self.o.interfaces if ii.enabled()]
+
+    def test_operatorData(self):
+        """Test that the operator has input data, a reactor model, and a list of interfaces.
+
+        .. test:: The Operator includes input data and the reactor data model.
+            :id: T_ARMI_OPERATOR_COMM
+            :tests: R_ARMI_OPERATOR_COMM
+        """
+        self.assertEqual(self.o.r, self.r)
+        self.assertEqual(type(self.o.cs), settings.Settings)
+        self.assertGreater(len(self.o.interfaces), 0)
+        for i in self.o.interfaces:
+            self.assertTrue(isinstance(i, Interface))
 
     def test_addInterfaceSubclassCollision(self):
         cs = settings.Settings()
@@ -185,7 +197,12 @@ class TestTightCoupling(unittest.TestCase):
             self.assertEqual(self.o.r.core.p.coupledIteration, 0)
 
     def test_performTightCoupling_notConverged(self):
-        """Ensure that the appropriate ``runLog.warning`` is addressed in tight coupling reaches max num of iters."""
+        """Ensure that the appropriate ``runLog.warning`` is addressed in tight coupling reaches max num of iters.
+
+        .. test:: The tight coupling logic can fail if there is no convergence.
+            :id: T_ARMI_OPERATOR_PHYSICS0
+            :tests: R_ARMI_OPERATOR_PHYSICS
+        """
 
         class NoConverge(TightCoupler):
             def isConverged(self, _val: TightCoupler._SUPPORTED_TYPES) -> bool:
--- a/armi/operators/tests/test_operators.py
+++ b/armi/operators/tests/test_operators.py
@@ -21,24 +21,24 @@ import sys
 import unittest
 
 from armi import settings
+from armi.bookkeeping.db.databaseInterface import DatabaseInterface
 from armi.interfaces import Interface, TightCoupler
 from armi.operators.operator import Operator
-from armi.reactor.tests import test_reactors
-from armi.settings.caseSettings import Settings
-from armi.utils.directoryChangers import TemporaryDirectoryChanger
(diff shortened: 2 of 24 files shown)


armi/operators/operator.py
```diff
@@ -402,10 +402,6 @@ def _timeNodeLoop(self, cycle, timeNode):
     def _performTightCoupling(self, cycle: int, timeNode: int, writeDB: bool = True):
         """If requested, perform tight coupling and write out database.
 
-        .. impl:: The operator shall allow tight coupling between physics systems.
-            :id: I_ARMI_OPERATOR_PHYSICS
-            :implements: R_ARMI_OPERATOR_PHYSICS
-
```
keckler, line 408: I assume this was removed because it was on a private method, but did you replace it somewhere?

armi/operators/tests/test_operators.py
```diff
@@ -56,12 +58,75 @@ class InterfaceC(Interface):
     name = "Third"
 
 
-# TODO: Add a test that shows time evolution of Reactor (R_EVOLVING_STATE)
```
keckler, line 59: What is this req? I don't see it mentioned anywhere...

armi/operators/tests/test_operators.py
```diff
@@ -156,6 +221,20 @@ def test_snapshotRequest(self, fakeDirList, fakeCopy):
...
+        """Test that an operator can be created from settings.
+
+        .. test:: Create an operator from settings.
+            :id: T_ARMI_OPERATOR_SETTINGS
+            :tests: R_ARMI_OPERATOR_SETTINGS
+        """
+        cs = settings.Settings()
+        o = Operator(cs)
+        self.assertTrue(isinstance(o, Operator))
+        self.assertTrue(isinstance(o.cs, settings.Settings))
```
keckler, line 235: It might also be easy to test here that the operator _depends_ on the settings, which is called out in the acceptance criteria.

armi/operators/tests/test_operators.py
```diff
@@ -164,6 +243,20 @@ def setUp(self):
         self.o.r = Reactor("empty", None)
         self.o.r.core = Core("empty")
 
+    def test_getStepLengths(self):
```
keckler, line 246: Why is this test in the `TestTightCoupling` class?

keckler: FYI there are many similar tests here: https://github.com/terrapower/armi/blob/f2909b5588682ffc3592ad3d7ac6802e59611611/armi/operators/tests/test_operators.py#L274

armi/tests/test_plugins.py
```diff
@@ -18,9 +18,67 @@
...
+
+        # Check the default values of parameters defined by the neutronics plugin
+        self.assertIsNone(b.p.axMesh)
+        self.assertEqual(b.p.flux, 0)
+        self.assertEqual(b.p.power, 0)
+        self.assertEqual(b.p.pdens, 0)
+
+        # Check the default values of parameters defined by the fuel peformance plugin
+        self.assertEqual(b.p.gasPorosity, 0)
+        self.assertEqual(b.p.gasPorosity, 0)
```
keckler, line 56: This is duplicate with the line above.

armi/tests/test_plugins.py
```diff
@@ -18,9 +18,67 @@
...
+            :tests: R_ARMI_PLUGIN_INTERFACES
+        """
+        plugin = NeutronicsPlugin()
+
+        cs = settings.Settings()
+        results = plugin.exposeInterfaces(cs)
+
+        # each plugin should return a list
+        self.assertIsInstance(results, list)
+        for result in results:
```
keckler, line 72:
```suggestion
        self.assertGreater(len(results))
        for result in results:
```
This will ensure that the plugin can register multiple interfaces, as the requirement stipulates.

armi/settings/tests/test_settings.py
```diff
@@ -58,7 +58,7 @@ def defineSettings():
         ]
 
 
-class DummyPlugin2(plugins.ArmiPlugin):
+class DummyParamPlugin2(plugins.ArmiPlugin):
```
keckler, line 61: Not sure why the names of these classes changed. At the least, I would have expected them to change into `DummySettingPluginX`, since they define a setting not a parameter.

armi/settings/tests/test_settings.py
```diff
@@ -41,7 +41,7 @@
 THIS_DIR = os.path.dirname(__file__)
 
 
-class DummyPlugin1(plugins.ArmiPlugin):
+class DummyParamPlugin1(plugins.ArmiPlugin):
```
keckler, line 44: Since the associated req stipulates adding "multiple settings", it might be a good idea to add a second new setting from within this class and test that.

keckler (changes requested):

armi/operators/operator.py
john-science: Yup!

It is on the `TightCoupler` class in `armi/interfaces.py`, and it is on `Operator.interactAllCoupled()` in `armi/operators/operator.py`.

armi/settings/tests/test_settings.py
john-science: Ah, yeah, that's a funny typo as a result of some other testing I was doing. I'll make that change. 1 sec.

armi/tests/test_plugins.py
john-science: Fixed. It was meant to be liquidPorosity.

armi/settings/tests/test_settings.py
john-science: Fixed.

john-science:
I see the failing test. I fixed it.

armi/settings/tests/test_settings.py
john-science: Sure, easy. I added a second setting to this class: "avocados".

And added a couple checks to `test_pluginSettings()` below that also verify that registered correctly.

armi/operators/tests/test_operators.py
john-science: Actually, I put ANOTHER test for this same requirement below in the test class `CyclesSettingsTests`.

I wanted one test in both classes, because these two test classes have very different operators:

* The `TestTightCoupling` class has a simple operator where step lengths are defined by the settings: nCycles, cycleLength, and burnSteps.
* The `CyclesSettingsTests` class has an operator where people are using the "manual" definition of detailed time steps.

I wanted to test both of these, and it just so happens, they are both happening in the same file.  I thought my crumbs described the difference well enough.  But see what you think.

armi/operators/tests/test_operators.py
john-science: Frankly, there isn't much to an operator, except a bunch of properties that directly call settings.

But I added some testing to this test to validate `burnSteps` and `powerFractions` values in the operator match what came from the setings.

keckler (approved):
Looks great, thanks for the changes.

PR #1506, opened 2023-11-29
Adding impl/test crumbs

john-science:
## What is the change?

Adding impl/test crumbs for some random requirements.

This is part of on-going requirement work.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -157,7 +157,7 @@ class Block(composites.Composite):
 
         return b
 
-    def _createHomogenizedCopy(self, pinSpatialLocators=False):
+    def createHomogenizedCopy(self, pinSpatialLocators=False):
         """
         Create a copy of a block.
 
@@ -1629,7 +1629,7 @@ class HexBlock(Block):
             round(y, units.FLOAT_DIMENSION_DECIMALS),
         )
 
-    def _createHomogenizedCopy(self, pinSpatialLocators=False):
+    def createHomogenizedCopy(self, pinSpatialLocators=False):
         """
         Create a new homogenized copy of a block that is less expensive than a full deepcopy.
 
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -1081,7 +1081,12 @@ class Block(composites.Composite):
         return sortedComponents
 
     def getNumPins(self):
-        """Return the number of pins in this block."""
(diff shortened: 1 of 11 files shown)


armi/reactor/blocks.py
```diff
@@ -157,7 +157,7 @@ def __deepcopy__(self, memo):
 
         return b
 
-    def _createHomogenizedCopy(self, pinSpatialLocators=False):
+    def createHomogenizedCopy(self, pinSpatialLocators=False):
```
keckler, line 160: This (private) method is referenced in a docstring of one of our internal plugins. Not a huge deal, just FYI.

armi/reactor/tests/test_components.py
```diff
@@ -217,13 +217,21 @@ def test_solid_material(self):
...
+        self.assertIn("HT9", str(self.component.getProperties()))
+
         self.component.material = air.Air()
         self.assertFalse(self.component.containsSolidMaterial())
 
         self.component.material = alloy200.Alloy200()
         self.assertTrue(self.component.containsSolidMaterial())
 
+        self.assertIn("Alloy200", str(self.component.getProperties()))
+
```
keckler, line 234: I think we probably need to take this a little further and show that there are actually properties here. I don't immediately see how testing for the material name is proving this.

armi/reactor/blocks.py
john-science: Ah, okay, I'll have to fix that too.

Can you point me at it?  I don't see it.

john-science: Handled.  Thanks!

armi/reactor/tests/test_components.py
keckler:
```suggestion

        .. test:: Components have material properties.
            :id: T_ARMI_COMP_MAT
            :tests: R_ARMI_COMP_MAT
        """
        self.assertTrue(hasattr(self.component.material, "density"))

        self.component.material = air.Air()
        self.assertFalse(self.component.containsSolidMaterial())

        self.component.material = alloy200.Alloy200()
        self.assertTrue(self.component.containsSolidMaterial())

        self.assertTrue(hasattr(self.component.material, "density"))

```

What do you think of that instead? I deleted the old asserts, but we could leave those in as well.

john-science: Okay, I added those. And kept the old ones.

PR #1508, opened 2023-11-29
Adding some impl crumbs to docstrings

john-science:
## What is the change?

Adding some impl crumbs to docstrings.

---

## Checklist

### DIFF
--- a/armi/nuclearDataIO/xsCollections.py
+++ b/armi/nuclearDataIO/xsCollections.py
@@ -788,6 +788,10 @@ def computeMacroscopicGroupConstants(
     """
     Compute any macroscopic group constants given number densities and a microscopic library.
 
+    .. impl:: Compute macroscopic cross sections from microscopic cross sections and number densities.
+        :id: I_ARMI_NUCDATA_MACRO
+        :implements: R_ARMI_NUCDATA_MACRO
+
     Parameters
     ----------
     constantName : str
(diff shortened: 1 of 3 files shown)


armi/nuclearDataIO/xsCollections.py
```diff
@@ -381,6 +381,12 @@ def __init__(
     def createMacrosOnBlocklist(
         self, microLibrary, blockList, nucNames=None, libType="micros"
     ):
+        """Create macroscopic cross sections for a list of blocks.
+
+        .. impl:: Build macroscopic cross sections for blocks.
+            :id: I_ARMI_MACRO_XS
+            :implements: R_ARMI_MACRO_XS
```
albeanth, line 388: So this requirement looks like it's supposed to live in the macroXSGenerationInterface. Based on that, it looks like this is probably the intended implementation:
https://github.com/terrapower/armi/blob/0c4d5f4c4353f670ce6096b239af360addcc6fde/armi/physics/neutronics/macroXSGenerationInterface.py#L131-L140

This method in xsCollections.py looks relatively unused, actually.

john-science: it is not used in ARMI, but it _is_ used downstream, but some internal projects.

So, there's that. Hmmm

john-science: Actually, that class `MacroXSGenerationInterface` is used  heavily, though only by one downstream project.

Which is interesting, and maybe we could put a not on that  project to see if that should be moved out of ARMI. 

But that file isn't part of this PR.

john-science: Okay, I moved the `impl` crumb from xsCollections to `MacroXSGenerationInterface.buildMacros())`.

albeanth (approved):

PR #1514, opened 2023-12-01
Reving unit tests for parallel process param updating

john-science:
## What is the change?

This PR does two things to do with parallel processing in ARMI:

* Revives some ARMI unit tests that were (apparently) deactivated years ago.
* Add some unit test crumbs for a couple requirements.

This is part of the on-going requirement work.

---

## Checklist


armi/reactor/tests/test_parameters.py
```diff
@@ -509,228 +523,80 @@ class MockSyncPC(parameters.ParameterCollection):
...
-    def mpitest_noConflicts_setByString(self):
-        """Make sure params set by string also work with sync."""
-        for ci, comp in enumerate(self.comps):
-            if ci % context.MPI_SIZE == context.MPI_RANK:
-                comp.p.param2 = (context.MPI_RANK + 1) * 30.0
-            else:
-                self.assertNotEqual((context.MPI_RANK + 1) * 30.0, comp.p.param2)
+    @unittest.skipIf(context.MPI_SIZE <= 1 or MPI_EXE is None, "Parallel test only")
+    def test_withConflicts(self):
+        """Test conflicts arrise correctly if we force a conflict.
```
keckler, line 572:
```suggestion
        """Test conflicts arise correctly if we force a conflict.
```

armi/reactor/tests/test_parameters.py
```diff
@@ -509,228 +523,80 @@ class MockSyncPC(parameters.ParameterCollection):
 def makeComp(name):
     c = composites.Composite(name)
     c.p = MockSyncPC()
+    for pd in c.p.paramDefs:
+        pd.assigned = parameters.NEVER
```
keckler, line 527: Did these need to be added? It appears that the same thing is done in the `SynchronizationTests.setUp()` method, which appears to be the only place that `makeComp()` is used.

Maybe you meant to remove the similar lines from the `setUp()` method?

armi/reactor/tests/test_parameters.py
```diff
@@ -509,228 +523,80 @@ class MockSyncPC(parameters.ParameterCollection):
...
-    def assertNotEqual(self, expected, actual):
-        if expected == actual:
-            raise AssertionError(
-                "(expected) {} == {} (actual)".format(expected, actual)
-            )
+        .. test:: Synchronize a reactor's state across processes.
+            :id: T_ARMI_CMP_MPI0
+            :tests: R_ARMI_CMP_MPI
+        """
+        syncCount = self.r.syncMpiState()
```
keckler, line 556: I think this is unneeded? Or at least the returned `syncCount` is unused, right? If so, either remove this line or replace `syncCount` with `_`, perhaps.

keckler (commented):
Thanks for migrating over these tests which were not being triggered before. This is an important fix.

Can you explain why you decided not to migrate over the tests that were looking at the things like `SINCE_ANYTHING` or `SINCE_DISTRIBUTE_STATE`, specifically in the old tests called `mpitest_noConflictsMaintainWithStateRetainer` and `mpitest_rxCoeffsProcess`?

While I understand that those "tests" were not actually being executed in the past, this part of ARMI is something that has caused us issues in the past with our plugins, and having tests hitting that part of the code would be very valuable.

john-science:
> Can you explain why you decided not to migrate over the tests that were looking at the things like `SINCE_ANYTHING` or `SINCE_DISTRIBUTE_STATE`, specifically in the old tests called `mpitest_noConflictsMaintainWithStateRetainer` and `mpitest_rxCoeffsProcess`?

I don't think those tests every "worked", and it might take me a week to build the kind of infrastructure needed to test them. I wasted a day trying to get them working and it was just hard to figure out how to do inside PyTest.

My guess is those would be testable inside some sort of cluster-based integration test. But, either way, I don't have a week free time right now to experiment with building new testing tools. Especially since there are no guarantees I would succeed, and it might just be pure R&D time.

armi/reactor/tests/test_parameters.py
john-science: Not sure why spellcheck is failing me. But good catch. 

Fixed.

armi/reactor/tests/test_parameters.py
john-science: Fair point.  This is just something I didn't cleanup while trying various versions of these tests.

I removed them from the `setUp()` method.

armi/reactor/tests/test_parameters.py
john-science: I think something must have changed in practice in the last 4 or 5 years since these tests were written.  It appears we sync the MPI state more often, so now the only way to flip all the parameter sync flags to their correct starting positions is to call a `syncMpiState()` while everything is still clean.  (Or replicate a ton of work with things like `ps.assigned = parameters.NEVER`, as you commented on above.

keckler (approved):
Okay, I made a ticket to keep track of the fact that some (previously not working) tests were deleted: https://github.com/terrapower/armi/issues/1515

With that, I'm good to approve. Thanks!

PR #1559, opened 2023-12-18
Add sys info printout for all nodes

opotowsky:
## What is the change?

This adds a little extra info to the runLog for cases. In addition to the machine information, there's a `systeminfo` printout included for each node to show the processor type and OS info.

### For the reviewer:

I do not think this is an important enough update to include in the release notes, but will add it if the you feel otherwise. Additionally, I am not able to add a test for this unless I can add this to an MPI test file.

This helps prove that a case was run on a particular system. This is motivated by downstream analysis needs.

---

## Checklist

- [ ] [Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify that the new/changed code works.

### DIFF
--- a/armi/bookkeeping/report/reportingUtils.py
+++ b/armi/bookkeeping/report/reportingUtils.py
@@ -21,6 +21,7 @@ import collections
 import os
 import pathlib
 import re
+import subprocess
 import sys
 import tabulate
 import textwrap
@@ -170,6 +171,7 @@ def writeWelcomeHeaders(o, cs):
             processorNames = context.MPI_NODENAMES
             uniqueNames = set(processorNames)
             nodeMappingData = []
+            sysInfo = ""
             for uniqueName in uniqueNames:
                 matchingProcs = [
                     str(rank)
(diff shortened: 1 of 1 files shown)


armi/bookkeeping/report/reportingUtils.py
```diff
@@ -188,6 +195,8 @@ def _writeMachineInformation():
                     tablefmt="armi",
                 )
             )
+            runLog.header("=========== System Information ===========")
+            runLog.info(sysInfo)
```
john-science, line 199: Perhaps we can wrap these two lines in a check?

```python
if sysInfo:
    runLog.header("=========== System Information ===========")
    runLog.info(sysInfo)
``

File: armi/bookkeeping/report/reportingUtils.py [hunk 2]
```diff
@@ -180,6 +182,11 @@ def _writeMachineInformation():
                 nodeMappingData.append(
                     (uniqueName, numProcessors, ", ".join(matchingProcs))
                 )
+                # Run sys info on each unique node too
```
john-science, line 185: Can we change this comment to something like:

```python
# If this is on Windows: run sys info on each unique node too
```

armi/bookkeeping/report/reportingUtils.py
opotowsky: done!

armi/bookkeeping/report/reportingUtils.py
opotowsky: done!

john-science (approved):

john-science (approved):

PR #1560, opened 2023-12-19
Create method to retrieve active interfaces

albeanth:
## What is the change?
Adds a method on the Operator to retrieve the active interfaces. Aims to reduce copy and pasted code and improve testing for what interfaces should be active for a given interact state.

Being able to check which interfaces are active for a given interact state is particularly useful for ARMI Applications.

---

## Checklist

### DIFF
--- a/armi/operators/operator.py
+++ b/armi/operators/operator.py
@@ -946,46 +946,35 @@ class Operator:
         activeInterfaces: List[Interfaces]
             The interfaces deemed active for the given interactState.
         """
-        activeInterfaces = []
+        # Validate the inputs
+        if interactState not in ("BOL", "BOC", "EveryNode", "EOC", "EOL", "Coupled"):
+            raise ValueError(f"{interactState} is an unknown interaction state!")
+
+        # Ensure the interface is enabled.
+        enabled = lambda i: i.enabled()
         if interactState == "BOL":
-            activeInterfaces = [
-                ii
-                for ii in self.interfaces
-                if (ii.enabled() or ii.bolForce())
-                and ii.name not in excludedInterfaceNames
-            ]
-            activeInterfaces = [
-                ii
-                for ii in activeInterfaces
-                if ii.name not in self.cs[CONF_DEFERRED_INTERFACE_NAMES]
-            ]
-        elif interactState == "BOC":
-            activeInterfaces = [ii for ii in self.interfaces if ii.enabled()]
-            if cycle < self.cs[CONF_DEFERRED_INTERFACES_CYCLE]:
-                activeInterfaces = [
-                    ii
-                    for ii in activeInterfaces
-                    if ii.name not in self.cs[CONF_DEFERRED_INTERFACE_NAMES]
-                ]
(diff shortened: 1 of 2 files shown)


armi/operators/operator.py
```diff
@@ -621,22 +608,12 @@ def interactAllEveryNode(self, cycle, tn, excludedInterfaceNames=None):
...
 
     def interactAllEOC(self, cycle, excludedInterfaceNames=None):
         """Interact end of cycle for all enabled interfaces."""
-        excludedInterfaceNames = excludedInterfaceNames or ()
-        activeInterfaces = [
-            ii
-            for ii in self.interfaces
-            if ii.enabled() and ii.name not in excludedInterfaceNames
-        ]
+        activeInterfaces = self.getActiveInterfaces("EOC")
```
keckler, line 616: It appears that you've lost the `excludedInterfaceNames` in a number of these calls.

john-science (approved):

armi/operators/operator.py
john-science: Okay, let me do a tally of the `excludedInterfaceNames` logic, just to be sure:

* `interactAllBOL` - uses `excludedInterfaceNames` and `self.cs["deferredInterfaceNames"]`
* `interactAllBOC` - uses `self.cs["deferredInterfacesCycle"]`, if the cycle is early enough
* `interactAllEveryNode` - uses `excludedInterfaceNames`
* `interactAllEOC` - uses `excludedInterfaceNames`
* `interactAllEOL` - uses neither
* `interactAllCoupled` - uses neither

john-science: @keckler and @albeanth  It looks to me like the new logic (in the new `getActiveInterfaces()` method matches the old logic.

Feel free to fact-check me on that.

john-science: Okay, I see it.

1 minute

john-science: Fixed

armi/operators/operator.py
```diff
@@ -608,12 +608,12 @@ def interactAllEveryNode(self, cycle, tn, excludedInterfaceNames=None):
         excludedInterfaceNames : list, optional
             Names of interface names that will not be interacted with.
         """
-        activeInterfaces = self.getActiveInterfaces("EveryNode")
+        activeInterfaces = self.getActiveInterfaces("EveryNode", excludedInterfaceNames)
         self._interactAll("EveryNode", activeInterfaces, cycle, tn)
 
     def interactAllEOC(self, cycle, excludedInterfaceNames=None):
```
albeanth, line 614: If you change this default to be an empty tuple like above, you can get rid of the extra error handling in https://github.com/terrapower/armi/pull/1560/commits/8fe16b2f1a8a8d975b08439d4c19345e660c531b

john-science: I can fix this, but sadly, some of our internal cases DO explicitly pass `None` in here, so that error-handling will have to remain.

Fixed.

PR #1565, opened 2023-12-22
Remove Parent tests/impls/reqs

Ashlita6:
## What is the change?

This PR is to combine the children and parent requirements/implementations and tests into one. 

Parent requirements are presenting downstream challenges for document review and they aren't needed because they are satisfied by the children. 

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_components.py
+++ b/armi/reactor/tests/test_components.py
@@ -75,6 +75,12 @@ class TestComponentFactory(unittest.TestCase):
         )
 
     def test_factory(self):
+        """Creating and verifying void and fuel components.
+
+        .. test:: Example void and fuel components are initialized.
+            :id: T_ARMI_COMP_DEF0
+            :tests: R_ARMI_COMP_DEF
+        """
         voidAttrs = self.getCircleVoidDict()
         voidComp = components.factory(voidAttrs.pop("shape"), [], voidAttrs)
         fuelAttrs = self.getCircleFuelDict()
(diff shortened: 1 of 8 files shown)


armi/reactor/tests/test_components.py
```diff
@@ -85,6 +91,12 @@ def test_factory(self):
         self.assertIsInstance(fuelComp.material, materials.UZr)
 
     def test_componentInitializationAndDuplication(self):
+        """Initialize and duplicate a component, veifying the parameters.
```
keckler, line 94: I don't really understand what this test is trying to prove.

It just creates a bunch of components and then verifies that they can be `deepcopy`-ed? Maybe you can add some clarification to the docstring about what this is supposed to prove?

keckler (approved):
These impls/tests look good with the corresponding PR on our internal app.

Thanks @Ashlita6 and @opotowsky !

armi/reactor/tests/test_components.py
opotowsky: You're right....I think the only thing it proves is that a component has parameters...

john-science: Correct. The requirement says that a component must have parameters, so this test covers that.  (So do many others, but I picked one.)

The test above proves that components have attributes. That's from the same req.

Do we need to add to the docstring here?

opotowsky: I think the docstring repeats what you just said, so it's probably fine?

john-science (approved):

PR #1567, opened 2023-12-29
Cleanup on Aisle 5! Dirty Testing

opotowsky:
## What is the change?

This PR cleans up a number of tests that leave dust behind.

I noticed some tests were a little bit dirty, so I added `TemporaryDirectoryChanger` where possible and manually delete files from tests that call code that "get around" the dir changer. (The code itself, in some places, has file paths that are not relative.)

---

## Checklist

### DIFF
--- a/armi/nuclearDataIO/tests/test_xsLibraries.py
+++ b/armi/nuclearDataIO/tests/test_xsLibraries.py
@@ -311,7 +311,7 @@ class TestXSlibraryMerging(TempFileMixin):
         del cls.libLumped
 
     def setUp(self):
-        super().setUp()
+        TempFileMixin.setUp(self)
         # load a library that is in the ARMI tree. This should
         # be a small library with LFPs, Actinides, structure, and coolant
         for attrName, path in [
--- a/armi/nuclearDataIO/tests/test_xsLibraries.py
+++ b/armi/nuclearDataIO/tests/test_xsLibraries.py
@@ -54,7 +54,7 @@ DLAYXS_MCC3 = os.path.join(FIXTURE_DIR_CCCC, "mc2v3.dlayxs")
 UFG_FLUX_EDIT = os.path.join(FIXTURE_DIR, "mc2v3-AA.flux_ufg")
 
 
-class TempFileMixin:
+class TempFileMixin(unittest.TestCase):
     """really a test case."""
 
     def setUp(self):
@@ -67,12 +67,12 @@ class TempFileMixin:
     @property
     def testFileName(self):
         return os.path.join(
-            THIS_DIR,
+            self.td.destination,
             "{}-{}.nucdata".format(self.__class__.__name__, self._testMethodName),
         )
 
 
-class TestXSLibrary(unittest.TestCase, TempFileMixin):
+class TestXSLibrary(TempFileMixin):
     @classmethod
     def setUpClass(cls):
         cls.isotxsAA = isotxs.readBinary(ISOTXS_AA)
@@ -233,9 +233,6 @@ class TestXSLibrary(unittest.TestCase, TempFileMixin):
         # check to make sure they labels overlap... or are actually the same
         writer.writeBinary(self.xsLib, self.testFileName)
         self.assertTrue(filecmp.cmp(refFile, self.testFileName))
-        # These files get around the TempDirChanger for some reason,
-        # and the bug is not obvious
-        os.remove(self.testFileName)
 
 
 class TestGetISOTXSFilesInWorkingDirectory(unittest.TestCase):
@@ -292,7 +289,7 @@ class TestGetISOTXSFilesInWorkingDirectory(unittest.TestCase):
 
 
 # NOTE: This is just a base class, so it isn't run directly.
-class TestXSlibraryMerging(unittest.TestCase, TempFileMixin):
+class TestXSlibraryMerging(TempFileMixin):
     """A shared class that defines tests that should be true for all IsotxsLibrary merging."""
 
     @classmethod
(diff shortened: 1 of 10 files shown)


keckler (approved):

jakehader:
@opotowsky - I don't think `TestXSlibraryMerging` is using the temporary directory changer since `setUp` is being overwritten. Could you check that for the XS library tests?

armi/tests/test_notebooks.py
```diff
@@ -36,6 +36,9 @@ def test_runParamSweep(self):
 
     def test_runDataModel(self):
         runNotebook(os.path.join(TUTORIALS, "data_model.ipynb"))
+        # Do some cleanup because some code run in the notebook doesn't honor the
```
jakehader, line 39: Can you wrap this in a temporary directory changer before calling the jupyter notebook? Maybe within the `runNotebook` function?

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -490,6 +491,13 @@ def test_repeatShuffles(self):
         for a in self.r.sfp.getChildren():
             self.assertEqual(a.getLocation(), "SFP")
 
+        # Do some cleanup, since the fuelHandler Interface has code that gets
```
jakehader, line 494: This code is that?

jakehader (changes requested):

armi/tests/test_notebooks.py
opotowsky: I tried that first. I think some places in the code use absolute file paths and I don't want to change the code unless there's a good reason to

opotowsky:
> @opotowsky - I don't think `TestXSlibraryMerging` is using the temporary directory changer since `setUp` is being overwritten. Could you check that for the XS library tests?

The `TempFileMixin` class defines `setUp` and `tearDown` methods, so I didn't think the `setUpClass` methods would override those. I will look into this though.

jakehader:
> > @opotowsky - I don't think `TestXSlibraryMerging` is using the temporary directory changer since `setUp` is being overwritten. Could you check that for the XS library tests?
> 
> The `TempFileMixin` class defines `setUp` and `tearDown` methods, so I didn't think the `setUpClass` methods would override those. I will look into this though.

I think there is a overwritten `setUp` - https://github.com/terrapower/armi/blob/4b4c52eb36235c103f86618614ad61dcc66abeae/armi/nuclearDataIO/tests/test_xsLibraries.py#L316

opotowsky:
> > @opotowsky - I don't think `TestXSlibraryMerging` is using the temporary directory changer since `setUp` is being overwritten. Could you check that for the XS library tests?
> 
> The `TempFileMixin` class defines `setUp` and `tearDown` methods, so I didn't think the `setUpClass` methods would override those. I will look into this though.

@jakehader Ah, I didn't see at first the `TestXSlibraryMerging` has both `setUpClass` and `setUp`. I added some supers, even though it doesn't appear needed for the tests. 

I did fine the absolute file path plaguing these tests though, fix incoming!

armi/physics/fuelCycle/tests/test_fuelHandlers.py
opotowsky: not sure what you're asking here.

opotowsky:
@jakehader the inheritance was messed up AND there was an absolute file path creating the files outside of the temp dir. Both are fixed in the xs tests. Let me know if you had other concerns...I reviewed the manual file removals and don't think any are fixable without changing the code.

opotowsky:
@john-science 

This update broke CI:
image

It's a dependency of `jupyter-contrib-nbextensions`, which is a deprecated project (no release since 2022). I tested removing it in this PR to see if CI would run, and it did. Does this have any documentation building ramifications?

armi/physics/fuelCycle/tests/test_fuelHandlers.py
opotowsky: `runShuffling` calls `manageFuel` in the fuel handler interface. This makes the shuffles png's, as well as calls `makeShuffleReport`. Both of these use an `fName` variable that functions as absolute. Putting `test_repeatShuffles` inside a TempDirChanger does not alleviate the files left behind that are manually removed here. 

We could change `manageFuel` and `makeShuffleReport` to take the `fName` arg so the test can give it a full file path, but that's a code change that would require a lot of testing and may have unintended consequences.

armi/nuclearDataIO/tests/test_xsLibraries.py
```diff
@@ -311,6 +311,7 @@ def tearDownClass(cls):
         del cls.libLumped
 
     def setUp(self):
+        super().setUp()
```
jakehader, line 314: I don't like super but it makes sense if it works.

jakehader (approved):

armi/nuclearDataIO/tests/test_xsLibraries.py
opotowsky: https://github.com/terrapower/armi/pull/1567/commits/6aff1d6e846bde953d7e66960f26f3f30ef37082 🙈

pyproject.toml
```diff
@@ -81,7 +81,6 @@ test = [
     "black==22.6.0",
     "docutils",
     "ipykernel",
-    "jupyter-contrib-nbextensions",
```
john-science, line 84: @opotowsky I thought this was needed to build the docs.

IF so, then we should move it down to the "docs" dependencies list, not delete it.

Can you check on that?

john-science: I just want to check that our jupyter notebooks in the docs still build and run, after this change you made.

opotowsky: Done! https://github.com/terrapower/armi/pull/1567/commits/eaf635557f9ab3dc5a50dacb3f5e8c4c6440cfb3

john-science (approved):

PR #1584, opened 2024-01-16
Adding implementation description to runLog impl crumbs

john-science:
## What is the change?

Adding information to some `impl` crumbs in `runLog`.

Adding more information about the runLog system.

---

## Checklist

### DIFF
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -327,17 +327,15 @@ def concatenateLogs(logDir=None):
         :id: I_ARMI_LOG_MPI
         :implements: R_ARMI_LOG_MPI
 
-        The log files are plain text files. Since ARMI is
-        frequently run in parallel, the situation arises where
-        each ARMI process generates its own plain text log file.
-        This function combines the separate log files, per
-        process, into one log file.
+        The log files are plain text files. Since ARMI is frequently run in parallel,
+        the situation arises where each ARMI process generates its own plain text log
+        file. This function combines the separate log files, per process, into one log
+        file.
 
-        The files are written in numerical order, with the lead
-        process stdout first then the lead process stderr. Then
-        each other process is written to the combined file, in
-        order, stdout then stderr. Finally, the original stdout
-        and stderr files are deleted.
+        The files are written in numerical order, with the lead process stdout first
+        then the lead process stderr. Then each other process is written to the
+        combined file, in order, stdout then stderr. Finally, the original stdout and
+        stderr files are deleted.
     """
     if logDir is None:
         logDir = LOG_DIR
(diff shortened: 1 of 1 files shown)


ntouran (changes requested):
One public guide that it'd be awesome to shoot for, as it's based on NQA-1 and carries a lot of clout, is [NUREG-1737, section 3.3.1, Software Design and Implementation Document (SDID)](https://www.nrc.gov/docs/ML0101/ML010170081.pdf). If we write our implementation docs to that, there's a high liklihood it will mesh well with individual org's QA specific requirements and the underlying regulations. 

Therefore, I think we need some more elements to these documents

- [x] Descriptions of major design components related to specified requirements

This is mostly here in the various docstrings. I think some additional narrative at the top of this module in the module docstring describing how and why we decided to make a wrapper on top of python's logging module would be interesting. (side note: I think it was because we were ignorant of logging module's full capabilities!)

- [x] Technical description of a program (i.e., control flow, data flow, control logic,
data structures, the routine and calling hierarchy, variable definitions,
identification of inputs and outputs etc.)

This could be expanded. For example, can we get a graphviz image of how processes send data to this module and then where the data goes on its way to text files/console? Variable definitions are included in type hints and parameter definitions in most cases, but should be expanded.  Inputs/outputs could be expanded slightly more (e.g. with more type hints). 

- [x] Allowable / prescribed input and output ranges

This is missing and needs to be added. What are the limitations of the data that can be logged, if any?

- [ ] Design Implementation - Model implementation or integration brings together
the source code with individual models to form an operational package to obtain desired results

This one is maybe not as applicable/generic as the others above. I think top-level module docstring description of the overall design approach of this module will cover this fine.

armi/runLog.py
```diff
@@ -503,9 +515,41 @@ class RunLogger(logging.Logger):
...
+        And at the end of the ARMI-based simulation, the analyst doing
+        the simulation will have a full record of potentially intersting
+        information they can use to understand their run.
+
     .. impl:: Logging is done to the screen and to file.
         :id: I_ARMI_LOG_IO
         :implements: R_ARMI_LOG_IO
+
+        By default, this logger prints all log messages during an ARMI
+        simulation to screen and to a plain text file. That is, at any
```
opotowsky, line 546: Can you explain this a little better? I see if rank is 0 the `StreamHandler` is used and otherwise the `FileHandler` is used. -- I THINK that's the ticket but am not sure.

armi/runLog.py
```diff
@@ -503,9 +515,41 @@ class RunLogger(logging.Logger):
...
+        place in the ARMI application, developers can interject a plain
+        text logging message, and when that code is hit during an ARMI
+        simulation, the text will be piped to screen and to a log file.
+        This helpful for analysts who want to watch their simulation in
+        real time, and for everyone that want a record of anything
+        interesting that happened during the run.
```
opotowsky, line 552: I'd say this could get added to the basis or to additional info in the requirement, but isn't needed for the implementation. But it's fine if it stays too...I won't enforce this comment!

armi/runLog.py
```diff
@@ -503,9 +515,41 @@ class RunLogger(logging.Logger):
         :id: I_ARMI_LOG
         :implements: R_ARMI_LOG
 
+        This logger makes it easy for users to add log statements to
+        and ARMI application, and ARMI will control the flow of
+        those log statements. In particular, ARMI overrides the normal
+        Python logging tooling, to allow developers to pipe their log
+        statements to both screen and file. This works for stdout and
+        stderr.
```
opotowsky, line 523: I think these two sentences should be moved to the LOG_IO impl, since it's directly relevant to that.

armi/runLog.py
```diff
@@ -503,9 +515,41 @@ class RunLogger(logging.Logger):
...
+        full errors can be recorded with this tool. These different
+        types of log statements are easily identified by specifying the
+        "verbosity" of the log statement, in the code where you do the
+        logging. The most important verbosities are: error, warning,
+        info (which is just to add information to the log), and debug
+        (which is meant to be extremely detailed and not always
+        desired).
+
+        And at the end of the ARMI-based simulation, the analyst doing
+        the simulation will have a full record of potentially intersting
```
opotowsky, line 538:
```suggestion
        At the end of the ARMI-based simulation, the analyst doing
        the simulation will have a full record of potentially interesting
```

armi/runLog.py
```diff
@@ -503,9 +515,41 @@ class RunLogger(logging.Logger):
...
+        record during a run. For instance, basic notifications of what
+        is happening in the run, or basic warnings about something that
+        could be an issue, so you want people to know about. Likewise,
+        full errors can be recorded with this tool. These different
+        types of log statements are easily identified by specifying the
+        "verbosity" of the log statement, in the code where you do the
+        logging. The most important verbosities are: error, warning,
+        info (which is just to add information to the log), and debug
+        (which is meant to be extremely detailed and not always
+        desired).
```
opotowsky, line 535: I would think calling out `_RunLog.setVerbosity` would be needed here? The verbosity is set that way. 

And maybe the full list of available log levels/types are in `_RunLog.getLogLevels`.

Hard to call those out since they are in a private class, but would textually doing it be OK even if the API-linking wouldn't happen?

john-science: I will add some text that shows how end-users will call the logger:

```python
runLog.error("This sort of error might usually terminate the run.")
runLog.warning("The run will not terminate, but users might want to know.")
runLog.info("This is the usual verbosity.")
runLog.debug("This is only logged during a debug run.")
```

armi/runLog.py
john-science: Hmmm... That's an interesting point.

armi/runLog.py
john-science: When you use the `runLog()`:

```python
runLog.info("whatever")
```

This gets printed to the screen AND recorded to the log file. 

That is what I was hoping to explain.

opotowsky: I understand the outcome but I don't see how that is implemented in the code. All I see is what I mentioned above, which doesn't align with what's in the description. (I trust the description is correct, I just need help finding that in the code)

armi/runLog.py
```diff
@@ -326,6 +326,18 @@ def concatenateLogs(logDir=None):
...
+        frequently run in parallel, the situation arises where
+        each ARMI process generates its own plain text log file.
+        This function combines the separate log files, per
+        process into one log file.
+
+        The files are written in numerical order, with the lead
+        process stdout first then the lead process stderr. Then
+        each other process is written to the combined file, in
+        order, stdout then stderr. Finally, the original stdout
+        and stderr files are deleted.
```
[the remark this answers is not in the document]
john-science: I rewrote both, based on your comments, see what you think.

armi/runLog.py
john-science: Removed

armi/runLog.py
```diff
@@ -503,9 +513,37 @@ class RunLogger(logging.Logger):
...
+        This logger makes it easy for users to add log statements to and ARMI
+        application, and ARMI will control the flow of those log statements. In
+        particular, ARMI overrides the normal Python logging tooling, to allow
+        developers to pipe their log statements to both screen and file. This works for
+        stdout and stderr.
+
+        By default, this logger prints all log messages during an ARMI simulation to
+        screen and to a plain text file. That is, at any place in the ARMI application,
+        developers can interject a plain text logging message, and when that code is
+        hit during an ARMI simulation, the text will be piped to screen and a log file.
```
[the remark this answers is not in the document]
john-science: Done

opotowsky (approved):

PR #1587, opened 2024-01-19
Adding implementation information to the materials package

john-science:
## What is the change?

Adding implementation information to the materials package

This is part of on-going QA work.

---

## Checklist

### DIFF
--- a/armi/materials/__init__.py
+++ b/armi/materials/__init__.py
@@ -48,9 +48,10 @@ def setMaterialNamespaceOrder(order):
     """
     Set the material namespace order at the Python interpreter, global level.
 
-    .. impl:: Materials can be searched across packages in a defined namespace.
-        :id: I_ARMI_MAT_NAMESPACE
-        :implements: R_ARMI_MAT_NAMESPACE
+    .. impl:: Material collections are defined with an order of precedence in the case
+        of duplicates.
+        :id: I_ARMI_MAT_ORDER
+        :implements: R_ARMI_MAT_ORDER
 
         An ARMI application will need materials. Materials can be imported from
         any code the application has access to, like plugin packages. This leads to
@@ -140,10 +141,9 @@ def resolveMaterialClassByName(name: str, namespaceOrder: List[str] = None):
     gets used (Framework's UO2 vs. a user plugins UO2 vs. the Kentucky Transportation
     Cabinet's UO2) is up to the user at runtime.
 
-    .. impl:: Material collections are defined with an order of precedence in the case
-        of duplicates.
-        :id: I_ARMI_MAT_ORDER
-        :implements: R_ARMI_MAT_ORDER
+    .. impl:: Materials can be searched across packages in a defined namespace.
+        :id: I_ARMI_MAT_NAMESPACE
+        :implements: R_ARMI_MAT_NAMESPACE
 
         During the runtime of an ARMI application, but particularly during the
         construction of the reactor in memory, materials will be requested by name. At
--- a/armi/materials/__init__.py
+++ b/armi/materials/__init__.py
@@ -23,25 +23,24 @@ alone (like :py:mod:`thermal scattering laws <armi.nucDirectory.thermalScatterin
 As the fundamental macroscopic building blocks of any physical object,
 these are highly important to reactor analysis.
 
-This module handles the dynamic importing of all the materials defined here at the framework
-level as well as in all the attached plugins. It is expected that most teams will
-have special material definitions that they will want to define.
+This module handles the dynamic importing of all the materials defined here at the
+framework level as well as in all the attached plugins. It is expected that most teams
+will have special material definitions that they will want to define.
 
 It may also make sense in the future to support user-input materials that are not
 hard-coded into the app.
 
 The base class for all materials is in :py:mod:`armi.materials.material`.
 """
-import pkgutil
-import importlib
 from typing import List
+import importlib
 import inspect
+import pkgutil
 
 from armi.materials.material import Material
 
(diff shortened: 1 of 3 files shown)


armi/materials/__init__.py
```diff
@@ -52,6 +51,20 @@ def setMaterialNamespaceOrder(order):
...
+        import multiple collections of materials. To handle this, ARMI keeps a list of
+        material namespaces. This is an ordered list of importable packages that ARMI
+        can search for a particular material by name.
+
+        This automatic exploration of an importable package saves the user the
+        tedium have having to import or include hundreds of materials manually, or
+        individually in a settings file. But it comes with a caveat; the list is
+        ordered. If two different namespaces in the list include a material with the
+        same name, the first one found in the list is chosen, e.g. earlier namespace in
+        the list have precedence.
```
bsculac, line 67:
```suggestion
        An ARMI application will need materials. Materials can be imported from
        any code the application has access to, like plugin packages. This leads to 
        the situation where one ARMI application will want to import multiple 
        collections of materials. To handle this, ARMI keeps a list of material 
        namespaces. This is an ordered list of importable packages that ARMI
        can search for a particular material by name.

        This automatic exploration of an importable package saves the user the
        tedium have having to import or include hundreds of materials manually, or
        individually in a settings file. But it comes with a caveat; the list is
        ordered. If two different namespaces in the list include a material with the
        same name, the first one found in the list is chosen, i.e. earlier namespaces in
        the list have precedence.
```

armi/materials/__init__.py
```diff
@@ -120,30 +133,42 @@ def resolveMaterialClassByName(name: str, namespaceOrder: List[str] = None):
...
+        During the runtime of an ARMI application, but particularly during the
+        construction of the reactor in memory, materials will be identified by name. At
+        that point, this code is called to search for that material name. The search
+        goes through the ordered list of Python namespaces provided to search for the
+        name. The first time an instance of that material is found, it is returned. In
+        this way, the first items in the material namespace list take precedence.
+
+        When a meterial name is pased to this function, it might be something simple
+        like the string ``"UO2"``. But it could be much more specific, like
+        ``armi.materials.uraniumOxide:UO2``.
```
bsculac, line 158:
```suggestion
        During the runtime of an ARMI application, but particularly during the
        construction of the reactor in memory, materials will be requested from the 
        materials plugin by name. At that point, this code is called to search for that 
        material name. The search goes through the ordered list of Python 
        namespaces provided to search for the name. The first time an instance of 
        that material is found, it is returned. In this way, the first items in the material 
        namespace list take precedence.
        
        When a material name is passed to this function, it may be either a simple 
        name like the string ``"UO2"`` or it may be much more specific, like
        ``armi.materials.uraniumOxide:UO2``.
```
Rephrase second paragraph as needed, I think it would be good to provide a more complete description of the options available. Based on my reading of the other portions of the docstring, these are the two options.

armi/materials/material.py
```diff
@@ -34,16 +34,29 @@
...
 
     .. impl:: The abstract material class.
         :id: I_ARMI_MAT_PROPERTIES
         :implements: R_ARMI_MAT_PROPERTIES
 
+        The ARMI Materials library is based on the Object-Oriented Programming design
+        approach, and uses this generic ``Material`` base class. In this class we
+        define a large number of material properties like density, heat capacity, or
+        linear expansion. But this one class can't implement all of those mathematical
+        curves itself, so ARMI subclasses this base class to fill in the information.
```
bsculac, line 47:
```suggestion
        The ARMI Materials library is based on the Object-Oriented Programming design
        approach, and uses this generic ``Material`` base class. In this class we
        define a large number of material properties like density, heat capacity, or
        linear expansion coefficient. Specific materials then subclass this base class to
        assign particular values to those properties.
```

armi/materials/material.py
```diff
@@ -108,6 +121,12 @@ def name(self):
         .. impl:: The name of a material is accessible.
             :id: I_ARMI_MAT_NAME
             :implements: R_ARMI_MAT_NAME
+
+            Every instance of an ARMI material must have a simple, human-readable
+            string name. And, if possible, we want this string to match the class
+            name. (This, of course, puts some limits on both the string and the
+            class name.) These names must easily be retrievable as a class
+            property.
```
bsculac, line 129:
```suggestion
            Every instance of an ARMI material must have a simple, human-readable
            string name. And, if possible, we want this string to match the class
            name. (This, of course, puts some limits on both the string and the
            class name.) These names are easily retrievable as a class property.
```
Not sure if the impl description should say something like "These names are easily retrievable as a class property, **as shown here**." since the intent is for it to go into the documentation. That's where my mind wanted to go with the text, feel free to add if you think that is good or ignore if you don't.

armi/materials/material.py
```diff
@@ -725,6 +744,11 @@ def linearExpansion(self, Tk=None, Tc=None):
         .. impl:: Fluid materials are not thermally expandable.
             :id: I_ARMI_MAT_FLUID
             :implements: R_ARMI_MAT_FLUID
+
+            ARMI defines fluid materials as those that do not thermally expand. To
+            ensure this, any fluid materials can subclass this class, and they will
+            inherit this method with sets the linear expansion to zero at all
+            temperatures.
```
bsculac, line 751:
```suggestion
            ARMI does not model thermal expansion of fluids. The ``Fluid`` superclass
            therefore sets the thermal expansion coefficient to zero. All fluids 
            subclassing  the ``Fluid`` material will inherit this method which sets the
            linear expansion coefficient to zero at all temperatures.
```

bsculac (changes requested):
Editorial comments made, ready for originator review.

bsculac (approved):
Groovy

PR #1613, opened 2024-01-24
FluxFiles Deserve Hashes Too!

opotowsky:
## What is the change?

Adding fluxFiles to one of the crossSectionControl items that is gifted with a SHA. 

We want to be able to trace inputs such as these internally.

---

## Checklist

- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html). 😈

### DIFF
--- a/armi/bookkeeping/report/reportingUtils.py
+++ b/armi/bookkeeping/report/reportingUtils.py
@@ -138,20 +138,21 @@ def writeWelcomeHeaders(o, cs):
             inputInfo.append((label, fName, shaHash))
 
         # bonus: grab the files stored in the crossSectionControl section
-        for fluxSection, fluxData in cs["crossSectionControl"].items():
+        for xsID, xsSetting in cs["crossSectionControl"].items():
             fNames = []
             # Users shouldn't ever have both of these defined, but this is not the place
             # for code to fail if they do. Allow for both to not be None.
-            if fluxData.xsFileLocation is not None:
-                # list of files
-                fNames.extend(fluxData.xsFileLocation)
-            if fluxData.fluxFileLocation is not None:
+            if xsSetting.xsFileLocation is not None:
+                # possibly a list of files
+                if isinstance(xsSetting.xsFileLocation, list):
+                    fNames.extend(xsSetting.xsFileLocation)
+                else:
+                    fNames.append(xsSetting.xsFileLocation)
+            if xsSetting.fluxFileLocation is not None:
                 # single file
-                fNames.append(fluxData.fluxFileLocation)
+                fNames.append(xsSetting.fluxFileLocation)
             for fName in fNames:
-                label = f"crossSectionControl-{fluxSection}"
-                if isinstance(fName, list):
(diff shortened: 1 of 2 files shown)


armi/bookkeeping/report/reportingUtils.py
```diff
@@ -139,13 +139,21 @@ def _listInputFiles(cs):
...
         for fluxSection, fluxData in cs["crossSectionControl"].items():
+            fNames = []
+            # Users shouldn't ever have both of these defined, but this is not the place
+            # for code to fail if they do. Allow for both to not be None.
             if fluxData.xsFileLocation is not None:
+                # list of files
+                fNames.extend(fluxData.xsFileLocation)
+            if fluxData.fluxFileLocation is not None:
+                # single file
+                fNames.append(fluxData.fluxFileLocation)
```
opotowsky, line 150: I need confirmation that this is accurate in all cases. Don't want to mess things up if single xsFiles are allowed (single item list will work here)

mgjarrett: I think it will always be a `list`. Even if a user just wants to provide one file, it should still be a single item `list`.

armi/bookkeeping/report/reportingUtils.py
```diff
@@ -139,13 +139,21 @@ def _listInputFiles(cs):
 
         # bonus: grab the files stored in the crossSectionControl section
         for fluxSection, fluxData in cs["crossSectionControl"].items():
```
mgjarrett, line 141: we don't *have* to fix it in this PR, but the pre-existing variable names of `fluxSection` and `fluxData` here are confusing to me. I would propose we replace them with:

`fluxSection` -> `xsID`
`fluxData` -> `xsSettings`  (open to suggestions for this one)

mgjarrett:
@opotowsky it looks good to me. I don't think we need any changes to the code you added, but I did leave a comment about potentially changing some of the pre-existing variable names.

armi/bookkeeping/report/reportingUtils.py
john-science: Agreed. Not sure if now is the time. But agreed.

armi/bookkeeping/report/reportingUtils.py
opotowsky: good that's what I thought! thanks for confirming

mgjarrett (approved):
I'm approving as a reviewer, but I also contributed, so maybe we should get one more approval from someone else?

armi/bookkeeping/report/reportingUtils.py
opotowsky: oh you alrady did it, even better!

opotowsky:
Ok I just reran the case submission internally and our hashes are lookin good! @john-science this can go in whenever you want it to (fine to push it to after the impl PRs are in)

john-science:
I retroactively approve this PR.  :)

PR #1617, opened 2024-01-24
Removing the non-Python term class method

john-science:
## What is the change?

Removing the non-Python term class method

In Python, a "method" is a "function" that is part of a class.  That is how those two terms are defined in Python. In other languages, those words mean slightly different things, sometimes.

In Python, the only time you say "class method", is if you use the [@classmethod](https://docs.python.org/3/library/functions.html#classmethod) decorator. So, this PR cleans up the language in a few places.

---

## Checklist

### DIFF
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -617,7 +617,7 @@ class ArmiObject(metaclass=CompositeModelType):
         Return a new instance of the specific ParameterCollection type associated with this object.
 
         This has the same effect as ``obj.paramCollectionType()``. Getting a new
-        instance through a method like this is useful in situations where the
+        instance through a class method like this is useful in situations where the
         ``paramCollectionType`` is not a top-level object and therefore cannot be
         trivially pickled. Since we know that by the time we want to make any instances
         of/unpickle a given ``ArmiObject``, such a class attribute will have been
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -617,11 +617,11 @@ class ArmiObject(metaclass=CompositeModelType):
         Return a new instance of the specific ParameterCollection type associated with this object.
 
         This has the same effect as ``obj.paramCollectionType()``. Getting a new
-        instance through a class method like this is useful in situations where the
+        instance through a method like this is useful in situations where the
         ``paramCollectionType`` is not a top-level object and therefore cannot be
         trivially pickled. Since we know that by the time we want to make any instances
         of/unpickle a given ``ArmiObject``, such a class attribute will have been
-        created and associated. So, we use this top-level class method to dig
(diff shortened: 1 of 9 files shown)


mgjarrett:
I think "class method" is a python term, although it is used incorrectly in some of the instances in this PR. The term "class method" appears many times in the python docs. The meaning as I understand it at a high-level is a method that is bound to a class rather than an instance of a class.

https://docs.python.org/3/reference/datamodel.html#special-method-names

image

armi/utils/flags.py
```diff
@@ -233,7 +233,7 @@ def extend(cls, fields: Dict[str, Union[int, auto]]):
             :id: I_ARMI_FLAG_EXTEND0
             :implements: R_ARMI_FLAG_EXTEND
 
-            A class method to extend a ``Flag`` with a vector of provided additional ``fields``,
+            A method to extend a ``Flag`` with a vector of provided additional ``fields``,
```
mgjarrett, line 236: This is one of the instances where I think "class method" is the proper term. Just "method" isn't technically incorrect, it's just lest specific.

armi/reactor/composites.py
```diff
@@ -617,11 +617,11 @@ def getParameterCollection(cls):
         Return a new instance of the specific ParameterCollection type associated with this object.
 
         This has the same effect as ``obj.paramCollectionType()``. Getting a new
-        instance through a class method like this is useful in situations where the
+        instance through a method like this is useful in situations where the
```
mgjarrett, line 620: This is one of the instances where I think "class method" is the proper term. This is maybe the most significant one, because the whole point of this paragraph is to talk about why the `@classmethod` decorator is useful. This is probably the one case where removing the word "class" changes the meaning of the docstring.

mgjarrett (changes requested):
There are one or two cases where I think we should keep the term "class method"

john-science:
> I think "class method" is a python term, although it is used incorrectly in some of the instances in this PR. The term "class method" appears many times in the python docs. The meaning as I understand it at a high-level is a method that is bound to a class rather than an instance of a class.

Well, in Python there is [classmethod](https://docs.python.org/3/library/functions.html#classmethod), of course. but I don't think any of these are that.

armi/reactor/composites.py
john-science: Oh, it actually does use `@classmethod`. My bad.  I didn't even check. I didn't think ARMI had too many of those. I'll fix it.

john-science, line 620:
```suggestion
        instance through a class method like this is useful in situations where the
```

armi/utils/flags.py
john-science: Yup. Fixing.

opotowsky (approved):

Issue #1624, opened 2024-01-25
There is a 109-page "List of Nuclides" table in the API


john-science:
In the ARMI docs, there is a full [list of Nuclides](https://terrapower.github.io/releases/armi/v0.3.0/.apidocs/armi.nucDirectory.nuclideBases.html#id1) table. When rendered to PDF it is about 109 pages long:

https://github.com/terrapower/armi/blob/c07fe4f3059eb5527d9d61055e5ab940c153b873/armi/nucDirectory/nuclideBases.py#L94-L99

I think there are two possibilities:

1. This isn't necessary in the docs: we can just remove the table.
2. This is important data: we need to pull it out of the API and put it somewhere more findable in the docs.

Which of these is true?

albeanth:
I vote 2 with an idea for the somewhere part:
- Replacing it with a script that can generate it into a csv for Excel. If someone is really interested in it, doing something like will be more useful than a super long table. 

That said, if option 1 is popular, I won't argue.

jakehader:
I think option 2 is correct, especially with half life data. I think the table is good in document - if someone wants it in Excel or csv they can import ARMI and get it from the nuclide objects directly.

albeanth:
@jakehader how about a compromise where we show a select number of nuclides and then reference the rest? E.g., show U238/225, some Zr, B10, C, H, Cs, Xe, etc, so the table is reasonably sized for the docs and then say something like "to see the remained X thousand nuclides, install ARMI and run this script."

Thoughts?

keckler:
We don't explicitly list out cross section data in our docs, I don't see any reason why half life data should be any different.

My vote is for (1), this isn't necessary in the docs.

john-science:
> We don't explicitly list out cross section data in our docs, I don't see any reason why half life data should be any different.
> 
> My vote is for (1), this isn't necessary in the docs.

Perhaps the real question is: Has anyone EVER used this table as a reference?  No one has ever asked me about it, so I'm inclined to guess "no".

jakehader:
> @jakehader how about a compromise where we show a select number of nuclides and then reference the rest? E.g., show U238/225, some Zr, B10, C, H, Cs, Xe, etc, so the table is reasonably sized for the docs and then say something like "to see the remained X thousand nuclides, install ARMI and run this script."
> 
> Thoughts?

[...]

PR #1634, opened 2024-01-26
Catchall PR for minor formatting/typo updates to docstrings

mgjarrett:
## What is the change?

Typographical and formatting updates to docstrings and implementation detail that is going into the SQA docs for ARMI.

Edits are being made as part of a review of the SQA documentation for ARMI.

---

## Checklist

- [ ] ~[Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify that the new/changed code works.~

### DIFF
--- a/armi/nucDirectory/elements.py
+++ b/armi/nucDirectory/elements.py
@@ -90,41 +90,41 @@ Retrieve elements that are classified as actinides:
      <Element  LR (Z=103), Lawrencium, ChemicalGroup.ACTINIDE, ChemicalPhase.SOLID>]
 
 
-For specific data on nuclides within each element, refer to the 
-:ref:`nuclide bases summary table <nuclide-bases-table>`.
+.. only:: html
 
+    For specific data on nuclides within each element, refer to the
+    :ref:`nuclide bases summary table <nuclide-bases-table>`.
 
-.. exec::
-    from tabulate import tabulate
-    from armi.nucDirectory import elements
+    .. exec::
+        from tabulate import tabulate
+        from armi.nucDirectory import elements
 
-    attributes = ['z',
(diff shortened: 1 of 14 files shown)


armi/nucDirectory/elements.py
```diff
@@ -90,41 +90,41 @@
...
-            f'``{len(element.nuclides)}``',
-        ]
-
-    sortedElements = sorted(elements.byZ.values())
-    return create_table(tabulate(tabular_data=[getAttributes(elem) for elem in sortedElements],
-                                 headers=attributes,
-                                 tablefmt='rst'),
-                        caption='List of elements')
-
+.. only:: html
```
john-science, line 93: You know we have a ticket open for this, right?

https://github.com/terrapower/armi/issues/1624

jakehader (approved):

armi/nucDirectory/elements.py
john-science: @albeanth @mgjarrett This is certainly one of the main options in the ticket I linked.  But I thought maybe we would discuss it first.  I certainly think it doesn't belong in the API. But it could be somewhere else in the docs.

Thoughts?

mgjarrett:
@opotowsky @john-science I don't see any other fixes coming in. Should we merge this now?

armi/nucDirectory/elements.py
albeanth: I think this is a fine band-aid solution but shouldn't be the long term solution. 

Having a 109-page long table in the HTML is also not very useful. I think we should either 1) get rid of it, or 2) have a small table with ~10 nuclides as an example and reference the rest in a script that can be run.

PR #1648, opened 2024-02-15
Fix rotation of hexagons in block diagrams

### LINKED ISSUE #1421: `plotBlockDiagram()` doesn't respect block orientation
The [`plotBlockDiagram()`](https://github.com/terrapower/armi/blob/8041ad889a0203da4602f0be321b54ebef5995b6/armi/utils/plotting.py#L1375) function appears to always assume that the lattice within the block is "flats-up". This becomes an issue if one has manually defined a "points-up" 
pin grid. In that case you'll get a print like the following:
image

The problem lines appear to be in this chunk:
https://github.com/terrapower/armi/blob/8041ad889a0203da4602f0be321b54ebef5995b6/armi/utils/plotting.py#L1283-L1290

Instead of always rotating the hex patch by 30 degrees, it needs to instead be able to interrogate the underlying grid to know the rotation. Unfortunately I think this is a challenge right now, as the block itself doesn't know the type of grid that is in it (that info is stored on the associated block _blueprint_). Even still, many block blueprints don't actually have a grid attribute even if the block later gets an auto-generated grid. This is related to https://github.com/terrapower/armi/pull/730#discussion_r913942220

This type of problem could be solved once https://github.com/terrapower/armi/issues/1284 is implemented, I would guess.

mgjarrett:
## What is the change?

Rotate hex patches based on the `cornersUp` property before plotting for a block diagram.

The rotation was previously hard-coded to assume a "flats up" hexagon block. This is not always the case. A new `cornersUp` property for the `HexGrid` class implemented in #1649 allows us to check for the correct orientation of the hexagons.

- Fix #1421
---

## Checklist

- [ ] [Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify that the new/changed code works.


mgjarrett:
Note: others may be working on parallel solutions to the issues linked in the PR description. I'm just trying to share the bare-bones outline of a path we could take to fix the problem; I'm not necessarily saying _this_ PR needs to be merged.

john-science:
As I said in the other ticket (https://github.com/terrapower/armi/issues/1278), I have am working on a PR that fixes two different `HexGrid` bugs.  I "assigned" that ticket to me to reduce code duplication.  Sorry about that.

I like your property, and haven't looked at your plotting fix yet.

armi/utils/plotting.py
```diff
@@ -1283,10 +1290,10 @@ def _makeComponentPatch(component, position, cold):
     elif isinstance(component, Hexagon):
         if component.getDimension("ip", cold=cold) != 0:
```
keckler, line 1291: The changes that you made are only inside of this `if` clause. This means the rotation will only be applied to the hexes that have an `ip`. 

The other major hexagon that doesn't have an `ip` is the `DerivedShape` that often represents the coolant. This one doesn't get rotated by the logic that you've implemented. Here is a diagram that I've produced using your changes:
image

keckler (changes requested):
Thanks @mgjarrett for trying these changes! I think there are a couple things that need to be fixed before we can maybe merge it.

Ultimately I think we can make it work, but _man_ this `plotBlockDiagram` method is frustratingly specific and fragile. In the longer term, it might make sense to split the method into two methods to handle the cases of hexagonal vs cartesian assemblies independently. I think that would make things a lot cleaner. Not asking for you to do this now, just maybe something to think about in a later ticket.

keckler:
I might also suggest changing the name of the PR, since this PR is _mostly_ about fixing the block diagrams, IMO. #1649 includes the same changes related to the new `cornersUp` property.

mgjarrett:
@keckler thanks for the review.

My goal here was just to show in code what the outline of my idea for fixing the problem was; I thought it would be easier than describing it in words. I knew @john-science was working on this issue, so I wanted to wait and see what the code looked like after he finished his work.

Now that he's fixed most of the confusion between flats/corners up, I think it makes sense to change the scope of this to just fixing `plotBlockDiagram`, as you suggest. I'll take a stab at it when I get a chance today.

mgjarrett:
@keckler I incorporated your comments; the plots look good now (at least for a flats up block; I don't have a corners up block handy for testing).

image

I merged #1649 into this branch locally and tested it, and the plotting still works. So once #1649 is merged into main, this PR can be merged with main and it should all work out.

mgjarrett:
I thought about adding unit tests, but this is mostly plotting stuff that is difficult to test directly.

armi/utils/plotting.py
```diff
@@ -1240,14 +1250,17 @@ def _makeComponentPatch(component, position, cold):
         cold: boolean
             True if looking for dimension at cold temps
 
+        hexRotation: float, optional
+            Amount of counterclockwise rotation (in degrees) for a hexagon component patch
```
keckler, line 1254:
```suggestion
            Amount of counterclockwise rotation (in degrees) for a hexagon component patch. 0 degrees
            corresponds to a hexagon with its corner pointing up.
```

keckler: Please correct the suggest wording if I got it backwards.

armi/utils/plotting.py
keckler: The pictures appear to be fixed for me as well.

keckler (approved):
Thank you for working on this!

mgjarrett:
Hmm still needs some work for corners up. I think the plotting is right, but the generation of the pin grid is wrong.

image

mgjarrett:
> Hmm still needs some work for corners up. I think the plotting is right, but the generation of the pin grid is wrong.
> 
> image

I looked into this problem a little more. There's nothing wrong with the plotting code necessarily, it's just that `block.autoCreateSpatialGrids()` only works if the block is "flats up". I tried to figure out a way to make it work with `cornersUp` blocks, but the block doesn't know what kind of grid it is going into when the local pin coordinates are auto-created.

Maybe we could check the orientation of the pin coordinates once the block has been placed in a `core` grid, and then rotate the pins 30 degrees counterclockwise if necessary? It's not a great solution, but it's the only thing I can think of.

mgjarrett:
this will be superseded by #1926 -- assuming that is going to be merged, I will close this one

john-science:
> this will be superseded by #1926 -- assuming that is going to be merged, I will close this one

Oh, I'm sorry Michael.  We made a list of bug tickets yesterday, and I was just going through them, trying to solve the low-hanging fruit.

I just didn't see this PR, and forgot. Sorry!

PR #1653, opened 2024-02-22
Inherit XS Settings when creating a new XS ID

mgjarrett:
## What is the change?

When the cross section group manager creates a new XS ID that is based on a previous XS ID within `CrossSectionGroupManager._getModifiedReprBlocks()`, the new settings will now be inherited from the settings for the original XS group.

Previously, a new XS ID would just be initialized with default cross section settings. Typically, when a user is creating a modified representative block with a new XS ID, they would want to keep the cross section settings from the original XS group.

---

## Checklist


jakehader:
Hi @mgjarrett - I might be misremembering but I think that we need to handle the case where the original XS settings are a link to an external cross section file or files. In that case we might need to default to 0D. Maybe I am misrememberinf why I implemented it this way.

mgjarrett:
> Hi @mgjarrett - I might be misremembering but I think that we need to handle the case where the original XS settings are a link to an external cross section file or files. In that case we might need to default to 0D. Maybe I am misrememberinf why I implemented it this way.

We shouldn't be trying to make a modified representative block based on a XS group that uses pregenerated cross sections. There is a warning logged in that situation:

https://github.com/terrapower/armi/blob/6163e942abe873129533b84a6f2efa632f260b47/armi/physics/neutronics/crossSectionGroupManager.py#L1255-L1261

No error is thrown, and you could even end up not getting the warning, depending on the arguments passed into `_getModifiedReprBlocks`. Maybe we need stronger controls around this. I can't tell if the intention is actually to prevent the user from passing in pregenerated XS to this method, or just to warn and allow it anyway.

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -849,6 +868,36 @@ def test_createRepresentativeBlocksUsingExistingBlocks(self):
...
+        self.assertEqual(baSettings.numExternalRings, aaSettings.numExternalRings)
+        self.assertEqual(baSettings.mergeIntoClad, aaSettings.mergeIntoClad)
+        self.assertEqual(
+            baSettings.meshSubdivisionsPerCm, aaSettings.meshSubdivisionsPerCm
+        )
+        self.assertEqual(baSettings.xsMaxAtomNumber, aaSettings.xsMaxAtomNumber)
+        self.assertEqual(baSettings.minDriverDensity, aaSettings.minDriverDensity)
+        self.assertEqual(baSettings.averageByComponent, aaSettings.averageByComponent)
+        self.assertEqual(baSettings.xsExecuteExclusive, aaSettings.xsExecuteExclusive)
+        self.assertEqual(baSettings.xsPriority, aaSettings.xsPriority)
```
keckler, line 899: Might be cleaner and easier to extend if this were done programmatically in a loop. Then it could be made so that it extends automatically when new sub-settings are added to `crossSectionControl`. That might only be practical if [`XSModelingOptions`](https://github.com/terrapower/armi/blob/f9903fe5a5d30ceda389d6494020f7a0c0f01b92/armi/physics/neutronics/crossSectionSettings.py#L328) were converted to a [`dataclass`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) though.

keckler: Not important!

mgjarrett: Good idea, I was able to use `xsSettings.__dict__` to simplify it significantly

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -1229,7 +1229,8 @@ def createRepresentativeBlocksUsingExistingBlocks(
             oldXSID = origXSIDsFromNew[newXSID]
             oldBlockCollection = blockCollectionByXsGroup[oldXSID]
             newBlockCollection = oldBlockCollection.__class__(
-                oldBlockCollection.allNuclidesInProblem
+                oldBlockCollection.allNuclidesInProblem,
+                averageByComponent=oldBlockCollection.averageByComponent,
             )
```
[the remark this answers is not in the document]
mgjarrett: I did the `Flag` -> `str` conversion by taking `str(flag).split(".")[1]`. I assume this is the correct way to do it; there isn't a good `toStr` method that exists for the flags.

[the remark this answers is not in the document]
keckler: Check out this method: https://github.com/terrapower/armi/blob/f9903fe5a5d30ceda389d6494020f7a0c0f01b92/armi/reactor/flags.py#L180

The docstring is wrong, but I think it does what you want.

keckler (approved):

armi/physics/neutronics/crossSectionGroupManager.py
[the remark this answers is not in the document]
mgjarrett: oh, I guess I didn't look hard enough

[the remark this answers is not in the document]
mgjarrett: Updated it to use the built-in function

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -1228,8 +1227,17 @@ def createRepresentativeBlocksUsingExistingBlocks(
         for newXSID in modifiedReprBlocks:
             oldXSID = origXSIDsFromNew[newXSID]
             oldBlockCollection = blockCollectionByXsGroup[oldXSID]
+            if len(oldBlockCollection._validRepresentativeBlockTypes) > 0:
+                validBlockTypes = []
+                for flag in oldBlockCollection._validRepresentativeBlockTypes:
+                    # validBlockTypes.append(str(flag).split(".")[1])
```
keckler, line 1233:
```suggestion
```

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -1233,8 +1232,16 @@ def createRepresentativeBlocksUsingExistingBlocks(
...
+                validBlockTypes = []
+                for flag in oldBlockCollection._validRepresentativeBlockTypes:
+                    validBlockTypes.append(flags._toString(Flags, flag))
+            else:
+                validBlockTypes = None
             newBlockCollection = oldBlockCollection.__class__(
-                oldBlockCollection.allNuclidesInProblem
+                oldBlockCollection.allNuclidesInProblem,
+                validBlockTypes=validBlockTypes,
+                averageByComponent=oldBlockCollection.averageByComponent,
```
john-science, line 1244: This is the business logic you are trying to implement with this ticket. And, on first glance, I think it looks good.

But I had to read it a couple times to figure out what you are doing. And, of course, I am looking at a PR with a good description.

What that tells me is this complicated logic you added probably needs at least a one-line comment telling people what they are about to read.

Does that make sense? It's an opinion, but I think the logic is complicated enough it should have a comment. (Alternatively, I suppose, some people would say it could be moved to another function with a helpful name. Meh.)

mgjarrett: Good point, I added a comment to describe what's happening in the new block of code

john-science (approved):

PR #1649, opened 2024-02-20
Fix some bugs in "corners up" Hex Grids

john-science:
## What is the change?

Four bugs were found, and fixed:

* [The hex grid/ascii map was rotated 60 degrees from the blueprints](https://github.com/terrapower/armi/pull/1649/commits/4261b82221d438cd47a8898d4d28ac7ad9ef3601)
* [HexGrid._makeOffsets()](https://github.com/terrapower/armi/commit/467a613505a081d31da3f73548dcaca5e2d94f6a) didn't reset the offsets before adding to them.
* [HexGrid.changePitch()](https://github.com/terrapower/armi/commit/15be52bfacc54005b0ec8bab601951186af5dfa3) didn't differentiate between "corners up" and "flats up".
* [HexGrid.pitch](https://github.com/terrapower/armi/pull/1649/commits/79a68cfa0f4ca4c348bfa920f5aef779f216d4e0) calculation was fixed.

Four "potential" bugs were mentioned in the ticket, but after much deliberation, these are just design choices and do not require any fixes:

* `HexGrid._getSymmetricIdenticalsThird`
* `HexGrid.overlapsWhichSymmetryLine`
* `HexGrid._indicesAndEdgeFromRingAndPos`

Please note, while @drewj-usnctech suggested we fix the above bugs by making subclasses for `HexGrid` for "flats up" and "corners up", this caused too many large diffs downstream. So the solution I took was to make a class property for "corners up" True/False.

## Why is the change being made?

This PR will close #1278

---

## Checklist

- [X] This PR has only [one purpose or idea](https://terrapower.github.io/armi/developer/tooling.html#one-idea-one-pr).
- [X] [Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify that the new/changed code works.

- [X] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [X] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html).

- [x] The [release notes](https://terrapower.github.io/armi/release/index.html) (location `doc/release/0.X.rst`) are up-to-date with any important changes.
- [X] The [documentation](https://terrapower.github.io/armi/developer/tooling.html#document-it) is still up-to-date in the `doc` folder.
- [x] If any [requirements](https://terrapower.github.io/armi/developer/tooling.html#watch-for-requirements) were affected, mention it in the [release notes](https://terrapower.github.io/armi/release/index.html).
[...]

PR #1654, opened 2024-02-27
Use clearer input syntax for hexagonal lattice pitch

### LINKED ISSUE #252: Finding hexagonal pin pitch for wire-less assemblies
(Came up working on a custom plugin for the HTGR model in #251 but felt it merited a different discussion)

`HexBlock`s have a `getPinPitch` method which was my first guess for finding the pin pitch for a hexagonal pin lattice. But the [docs](https://terrapower.github.io/armi/.apidocs/armi.reactor.blocks.html#armi.reactor.blocks.HexBlock.getPinPitch) indicate that the pin pitch is defined by, and thus requires, wire and cladding. For an HTGR, where the fuel pellets might be placed in channels, these components might not exist in the reactor.

For a workaround, I found that adding a dummy component
```yaml
        pitch:
            shape: Hexagon
            material: Void
            Tinput: 600
            Thot: 600
            ip: 3
            op: 3
```
worked well. But this would imply that each block would require this component, even if the same underlying lattice map was used.

Could ARMI be extended to support a pitch attribute in the `lattice map` section for hexagonal lattices. The C5G7 example gives something like
```yaml
    UO2 grid:
        symmetry: full
        geom: cartesian
        lattice pitch:
            x: 1.26
            y: 1.26
```

### LINKED ISSUE #1284: Incorrect pitch for tips up sub-block pin grid
In the same vein of #1278, the pin pitch seems to be off. The following grid definitions
```yaml
  pins:
    lattice pitch:
      x: 1.2
      y: 1.2
    geom: hex_corners_up
    lattice map: |
      - - O F F
       - F F F F
        F F F F F
         F F F F
          F F F
    # geom: hex
    # lattice map: |
    #   - F
    #    F F
    #   F F F
    #    F F
    #   F F F
    #    F F
    #   F F F
    #    F F
    #     F 
```
produce two different pitches. For the normal "flats up" hexagon, the pitch is correct: `b.spatialGrid.pitch=1.2`. But for the "tips up" grid, the pitch is off: `b.spatialGrid.pitch=1.0392304845413265` using
```python
import armi
armi.configure()
o = armi.init(fName="sample.yaml")
b = o.r.core[0][0]
g = b.spatialGrid
print(f"{b.spatialGrid.pitch=}")
```

My current workaround is to look at the `_unitSteps` attribute, which feels not ideal (using private attribute) 
```python
>>> b.spatialGrid._unitSteps
array([[ 0.6       , -0.6       ,  0.        ],
       [ 1.03923048,  1.03923048,  0.        ],
       [ 0.        ,  0.        ,  0.        ]])
>>> realPitch = b.spatialGrid._unitSteps[0,0] * 2
```
If I understand these right, the `_unitSteps[0, 0]` is the change in x for changing ring. So each new ring nets us a full new x-site, since they're all neighbors.

mgjarrett:
## What is the change?

Added an option for a user to specify a hexagonal pitch directly:

```yaml
latticePitch: {hex: 1.5}
```

Previously, a user had to specify the triangular pitch using the `x` attribute of `latticePitch`, which is confusing because the pitch is not necessarily equal to the `x` distance between two neighbors in a hex grid.

```yaml
latticePitch: {x: 1.5}
```

https://github.com/terrapower/armi/blob/main/armi/reactor/grids/hexagonal.py#L152-L161

- ~Fix 1284~ (fix for this issue pulled into a different PR, #1649)
- Close #252

---

## Checklist


john-science:
I'm just thinking aloud here, but what something "flat-to-flat" be better for hex grids?

Isn't the goal to call out that we are going to be explicitly picking one over the other?

mgjarrett:
> I'm just thinking aloud here, but what something "flat-to-flat" be better for hex grids?
> 
> Isn't the goal to call out that we are going to be explicitly picking one over the other?

I'm not sure I understand what you are proposing. Do you mean specifying the pitch/flat-to-flat distance as `flat-to-flat` in the blueprints instead of `latticePitch`? I think `latticePitch` is still a fine descriptor of that parameter for a hexagonal lattice. And this allows us to not break any backwards compatibility with existing blueprints files, which would be an enormous pain.

mgjarrett:
The bug fix was pulled out of this PR and into #1649 

The remaining part of this PR is an optional/nice-to-have feature. It's just allowing users to define the hex pitch in blueprints in a slightly different way than before. No rush to push this through now, really.

john-science:
The unit tests broke after my PR merged. Sorry!

mgjarrett:
> The unit tests broke after my PR merged. Sorry!

No worries! I just have to pull out what was moved over to #1649 and I think it will work.

john-science:
@mgjarrett Forgive me, I have totally lost where this PR was at. When you think it is ready for review, take it out of Draft mode.

mgjarrett:
I think the PR is ready. For context, I was trying to address the issue that Drew and Mitch were discussing 3 years ago over on this issue #252:

image

armi/bookkeeping/db/debug_db.py
```diff
@@ -0,0 +1,22 @@
+import nala
+
+nala.configure()
```
john-science: Nala?

Noopo!

armi/bookkeeping/db/debug_db.py
```diff
@@ -0,0 +1,22 @@
...
+
+import os
+
+from armi.bookkeeping.db import loadOperator
+
+cycle = 0
+node = 0
+
+fileName = "natrium-combined-c0n0-v1.1.0.h5"
```
john-science: Natrium? Please remove.

armi/bookkeeping/db/v110_to_v120.py
```diff
@@ -0,0 +1,54 @@
+"""
+Nala Database Migration Script.
```
john-science: Please remove this file from the PR

mgjarrett:
I had unintentionally added some files that were sitting around in my git directory to this PR. These files have been removed, so this PR is now ready.

armi/reactor/blueprints/tests/test_gridBlueprints.py
```diff
@@ -360,7 +361,8 @@ def tearDown(self):
 
     def test_simpleRead(self):
         gridDesign = self.grids["control"]
-        _ = gridDesign.construct()
+        grid = gridDesign.construct()
+        self.assertAlmostEqual(grid.pitch, 1.2)
```
john-science, line 365: I was thinking it would be nice to show that this works for corners up and flats up.

I was think you could copy/paste this test, and just search/replace `"geom: hex_corners_up"` with `"geom: hex`.

Thoughts?

keckler:
I reopened this, because it is a change that many of us would like to see implemented.

armi/reactor/blueprints/tests/test_gridBlueprints.py
john-science: bump

john-science:
> I reopened this, because it is a change that many of us would like to see implemented.

What's the status of the feature in this PR?

(Also, I want to merge `main` into this PR, but I can't because it is in a forked repo I can't touch. So, that would be nice.)

armi/reactor/blueprints/tests/test_gridBlueprints.py
mgjarrett: I added a flats up grid to the unit test.

armi/reactor/blueprints/gridBlueprint.py
```diff
@@ -130,6 +130,46 @@ def __init__(self, x=0.0, y=0.0, z=0.0):
         self.z = z
 
 
+class Pitch(yamlize.Object):
+    """A x, y, z triplet or triangular hex pitch for coordinates or lattice pitch."""
```
john-science:
```suggestion
    """A x, y, z triplet or triangular hex pitch for coordinates or lattice pitch for hexagonal grids."""
```

armi/reactor/blueprints/tests/test_gridBlueprints.py
```diff
@@ -30,6 +30,7 @@
 control:
     geom: hex_corners_up
     symmetry: full
+    lattice pitch: {{hex: 1.2}}
```
john-science: Out of curiosity, could this also be written like this?

```yaml
lattice pitch:
    hex: 1.2
```

mgjarrett: Yes, that is simpler and it does work. Updating it now.

armi/reactor/blueprints/gridBlueprint.py
```diff
@@ -130,6 +130,46 @@ def __init__(self, x=0.0, y=0.0, z=0.0):
...
+        Raises
+        ------
+        InputError
+            * If a `hex` pitch and `x` or `y` pitch are provided simultaneously.
+            * If no non-zero value is provided for any parameter.
+        """
+        if hex and (x or y):
+            raise InputError("Cannot mix `hex` with `x` and `y` attributes of `latticePitch`.")
+
+        if not any(hex, x, y, z):
```
keckler, line 164: This is not the correct call signature for `any()`. It takes one arg, which is an iterable type.

keckler: @mgjarrett @john-science

john-science: On it.

PR #1665, opened 2024-03-12
Conserve mass at component level for a.setBlockMesh()

mgjarrett:
## What is the change?

In the execution of `a.setBlockMesh()`, conserve masses of specific components instead of specific nuclides within the block. This prevents unwanted scaling of nuclide densities for nuclides that exist in multiple different components.

`a.setBlockMesh()` has an option, `conserveMassFlag="auto"`, which will try to automatically conserve mass of the fuel while allowing other components to change mass with changing block heights. This conservation is based upon getting the nuclides associated with the fuel component, and then scaling the density of those nuclides based on the change in block height using `b.setNumberDensities()`.

This method does not work when the fuel has nuclides that are also in other components of the block, like cladding or structure. In this case, the mass of those components will also be conserved through the height scaling, which is not the desired behavior.

Instead, `a.setBlockMesh()` needs to conserve mass by setting number densities directly on the components where mass needs to be conserved, instead of setting the densities on the full block.

The undesired behavior that this PR attempts to prevent is related to the same root cause as other outstanding GitHub issues about the "dehomogenization" underpinning the `Composite` method `setNumberDensities()` (#1374, in addition to several downstream plugin issues)

---

## Checklist


armi/reactor/assemblies.py
```diff
@@ -703,30 +703,31 @@ def _shouldMassBeConserved(self, belowFuelColumn, b):
...
                 # grid-plate dose calcs.
                 conserveMass = True
-                adjustList = b.getNuclides()
                 # conserve mass of everything except coolant.
-                coolant = b.getComponent(Flags.COOLANT)
-                coolantList = coolant.getNuclides() if coolant else []
-                for nuc in coolantList:
-                    if nuc in adjustList:
-                        adjustList.remove(nuc)
+                conserveComponents = [
```
albeanth, line 714: Do you think it would be safer to make this check on the material type instead of the flag? E.g., if the material is a `Fluid` then skip it? 
https://github.com/terrapower/armi/blob/827d223cd6fc085c039390a2c3a0391f4e611fe5/armi/materials/material.py#L728

mgjarrett: @albeanth I implemented your suggestion

armi/reactor/tests/test_assemblies.py
```diff
@@ -1151,6 +1151,10 @@ def test_snapAxialMeshToReferenceConservingMassBasedOnBlockIgniter(self):
         ################################
         igniterFuel = self.r.core.childrenByLocator[grid[0, 0, 0]]
         # gridplate, fuel, fuel, fuel, plenum
+        for b in igniterFuel.getBlocks(Flags.FUEL):
+            fuelComp = b.getComponent(Flags.FUEL)
+            fuelComp.setNumberDensity("FE56", 1e-10)
```
albeanth, line 1156: Can you add a comment here for why you are adding Fe56 and Na23 to the fuel?

PR #1667, opened 2024-03-21
Cleaning out the overly-specific methods in ArmiObject

### LINKED ISSUE #1375: Audit Composite/ArmiObject methods and their applicability to all objects
This is a general request to go through the implementation of each composite and to understand if an implementation should be applied generically to all objects in the composite tree. Examples include:

- [ArmiObject.changeNDensByFactor()](https://github.com/terrapower/armi/blob/64351a0a5538e6a150d2c4f9a1183fa6558f5269/armi/reactor/composites.py#L1705-L1710)
- [ArmiObject.getDominantMaterial()](https://github.com/terrapower/armi/blob/64351a0a5538e6a150d2c4f9a1183fa6558f5269/armi/reactor/composites.py#L2656)
- [ArmiObject.getMgFlux()](https://github.com/terrapower/armi/blob/64351a0a5538e6a150d2c4f9a1183fa6558f5269/armi/reactor/composites.py#L2285) - suspect
- [ArmiObject.getMaxVolume()](https://github.com/terrapower/armi/blob/64351a0a5538e6a150d2c4f9a1183fa6558f5269/armi/reactor/composites.py#L901) - abstract method
- [ArmiObject.getMicroSuffix()](https://github.com/terrapower/armi/blob/64351a0a5538e6a150d2c4f9a1183fa6558f5269/armi/reactor/composites.py#L950) - abstract method
- [ArmiObject.getUraniumNumEnrich()](https://github.com/terrapower/armi/blob/64351a0a5538e6a150d2c4f9a1183fa6558f5269/armi/reactor/composites.py#L1785) - suspect
- [ArmiObject.getUraniumMassEnrich()](https://github.com/terrapower/armi/blob/64351a0a5538e6a150d2c4f9a1183fa6558f5269/armi/reactor/composites.py#L1777) - suspect
- etc.

We might have went a bit too far on the data model in terms of inheritance of methods across the system and it leads to quirks like having material-specific implementations, using a `LumpedFissionProduct` within the system (physics related), assuming that an implementation could be made generically on a `Core`, `Assembly`, `Block`, or `Component`. 

Some things to think about are:

- Using functions that can consume an `ArmiObject`, check its type, and then perform the work requested
- Forcing implementations so that it's clearer how inheritance is happening (e.g., `Blocks` jump to `Composites` and then to `Components` in some implementations and this can be confusing to sort through)

Associated Tickets/Issues
---------------------------
- [ ] https://github.com/terrapower/armi/issues/1374

john-science:
## What is the change?

Removing extraneous `ArmiOjbect` methods:

* Moving `ArmiObject.getBoronMassEnrich()` to `Block`
* Moving `ArmiObject.getPuMoles()` to `Block`
* Moving `ArmiObject.getUraniumMassEnrich()` to `Block`
* Removing `ArmiObject.getMaxUraniumMassEnrich.()`
* Removing `ArmiObject.getMaxVolume()` & `Block.getMaxVolume()`
* Removing `ArmiObject.getPuFrac()`
* Removing `ArmiObject.getPuMass()`
* Removing `ArmiObject.getPuN()`
* Removing `ArmiObject.getZrFrac()`
* Removing `ArmiObject.printDensities()`
* Moving `Composite.isOnWhichSymmetryLine()` to `Assembly`
* Removing `Block.isOnWhichSymmetryLine()`

There are several methods in `ArmiObject` (and `Composite`) that seem a little too specific to be included everywhere in the ARMI object tree.  And after doing some research, some of them are entirely unused anyway.

Close #1375

---

## Checklist

### DIFF
--- a/armi/reactor/components/__init__.py
+++ b/armi/reactor/components/__init__.py
@@ -344,7 +344,7 @@ class DerivedShape(UnshapedComponent):
     @staticmethod
     def getMaxVolume(block):
         """
-        The maximum volume of the given Block if it were totally full.
+        The maximum volume of the parent Block.
 
         Parameters
         ----------
--- a/armi/reactor/components/__init__.py
+++ b/armi/reactor/components/__init__.py
@@ -341,22 +341,16 @@ class DerivedShape(UnshapedComponent):
         """
         return self._deriveVolumeAndArea()
 
-    @staticmethod
-    def getMaxVolume(block):
+    def getMaxVolume(self):
         """
         The maximum volume of the parent Block.
-
-        Parameters
-        ----------
-        block : Block
-            Block of interest.
 
         Returns
         -------
         vol : float
             volume in cm^3.
         """
-        return block.getMaxArea() * block.getHeight()
+        return self.parent.getMaxArea() * self.parent.getHeight()
 
     def _deriveVolumeAndArea(self):
         """
@@ -400,7 +394,7 @@ class DerivedShape(UnshapedComponent):
             except:  # noqa: bare-except
                 siblingArea = None
 
-        remainingVolume = DerivedShape.getMaxVolume(self.parent) - siblingVolume
+        remainingVolume = self.getMaxVolume() - siblingVolume
         if siblingArea:
             remainingArea = self.parent.getMaxArea() - siblingArea
 
@@ -410,7 +404,7 @@ class DerivedShape(UnshapedComponent):
                 f"The component areas in {self.parent} exceed the maximum "
                 "allowable volume based on the geometry. Check that the "
                 "geometry is defined correctly.\n"
-                f"Maximum allowable volume: {DerivedShape.getMaxVolume(self.parent)} "
+                f"Maximum allowable volume: {self.getMaxVolume()} "
                 f"cm^3\nVolume of all non-derived shape components: {siblingVolume} cm^3\n"
             )
             runLog.error(msg)
--- a/armi/reactor/components/__init__.py
+++ b/armi/reactor/components/__init__.py
@@ -341,6 +341,23 @@ class DerivedShape(UnshapedComponent):
         """
         return self._deriveVolumeAndArea()
 
+    @staticmethod
+    def getMaxVolume(block):
+        """
+        The maximum volume of the given Block if it were totally full.
+
+        Parameters
+        ----------
+        block : Block
+            Block of interest.
+
+        Returns
+        -------
+        vol : float
+            volume in cm^3.
+        """
+        return block.getMaxArea() * block.getHeight()
+
     def _deriveVolumeAndArea(self):
         """
         Derive the volume and area of a ``DerivedShape``.
@@ -383,7 +400,7 @@ class DerivedShape(UnshapedComponent):
             except:  # noqa: bare-except
                 siblingArea = None
 
-        remainingVolume = self.parent.getMaxVolume() - siblingVolume
+        remainingVolume = DerivedShape.getMaxVolume(self.parent) - siblingVolume
         if siblingArea:
             remainingArea = self.parent.getMaxArea() - siblingArea
 
@@ -393,8 +410,8 @@ class DerivedShape(UnshapedComponent):
                 f"The component areas in {self.parent} exceed the maximum "
                 "allowable volume based on the geometry. Check that the "
                 "geometry is defined correctly.\n"
-                f"Maximum allowable volume: {self.parent.getMaxVolume()} cm^3\n"
-                f"Volume of all non-derived shape components: {siblingVolume} cm^3\n"
+                f"Maximum allowable volume: {DerivedShape.getMaxVolume(self.parent)} "
+                f"cm^3\nVolume of all non-derived shape components: {siblingVolume} cm^3\n"
             )
             runLog.error(msg)
             raise ValueError(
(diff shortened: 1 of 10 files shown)


john-science:
Bump

armi/reactor/components/__init__.py
```diff
@@ -341,6 +341,23 @@ def computeVolume(self):
         """
         return self._deriveVolumeAndArea()
 
+    @staticmethod
+    def getMaxVolume(block):
+        """
+        The maximum volume of the given Block if it were totally full.
```
albeanth, line 347:
```suggestion
        The maximum volume of the parent Block.
```

armi/reactor/components/__init__.py
```diff
@@ -341,6 +341,23 @@ def computeVolume(self):
...
+        ----------
+        block : Block
+            Block of interest.
+
+        Returns
+        -------
+        vol : float
+            volume in cm^3.
+        """
+        return block.getMaxArea() * block.getHeight()
```
albeanth, line 359: Thoughts? Clearly you could need to make this not a `staticmethod`, but I think it makes sense since this method is only called in the context when a DerivedShape component has a parent block. 
```suggestion
        return self.parent.getMaxArea() * self.parent.getHeight()
```

armi/reactor/tests/test_blocks.py
```diff
@@ -1098,12 +1079,11 @@ def test_adjustDensity(self):
 
         cur = self.block.getNumberDensity("U235")
         ref = densAdj * u235Dens
-        places = 6
-        self.assertAlmostEqual(cur, ref, places=places)
+        self.assertAlmostEqual(cur, ref, places=6)
```
albeanth, line 1082: I think the precision can be increased here.

armi/reactor/tests/test_blocks.py
```diff
@@ -1098,12 +1079,11 @@ def test_adjustDensity(self):
...
 
         cur = self.block.getNumberDensity("U238")
         ref = densAdj * u238Dens
-        self.assertAlmostEqual(cur, ref, places=places)
+        self.assertAlmostEqual(cur, ref, places=6)
```
albeanth, line 1086: I think the precision can be increased here.

albeanth (commented):
Hmmmm. 

> Moving ArmiObject.getBoronMassEnrich() to Block
Moving ArmiObject.getPuMoles() to Block
Moving ArmiObject.getUraniumMassEnrich() to Block

I could see utility in having this at the composite level. I don't think it's unreasonable to have the ability to query these things at the Assembly, Core, or Component level (specifically on the Component as well, for flexibility with future applications)

john-science:
@albeanth 

> Moving ArmiObject.getBoronMassEnrich() to Block
> Moving ArmiObject.getPuMoles() to Block
> Moving ArmiObject.getUraniumMassEnrich() to Block
> 
> I could see utility in having this at the composite level....

If it helps, I tested this PR downstream in all repos, including benchmark repos, it passes.  So that is firm confirmation that these methods are NOT used at the `Composite` level, if that helps.

Cards on the table...

None of these methods should have been added to the API in the first place.  That was the cause of Jake/Mark's complaint.  They are fair too specific.  An entire method for just one nuclide?  If we have one for "boron mass' why not "boron moles" and "uranium mass" and "uranium moles" and OMG that paradigm leads us to hundreds and hundreds of nearly identical stupid methods.

Jake and Mark make a compelling argument: (1) remove any of these too-specific methods we can, and (2) reduce the surface area of the ones we can't remove.

Moving these methods from `ArmiObject` / `Composite` to something less specific means these methods aren't usable everywhere.  (Which, they aren't now.)  And that's a good thing. Because we want to _limit_ how much these methods are used.

albeanth:
> They are fair too specific. An entire method for just one nuclide? If we have one for "boron mass' why not "boron moles" and "uranium mass" and "uranium moles" and OMG that paradigm leads us to hundreds and hundreds of nearly identical stupid methods.

Oh yeah, totally agree here. I wonder if they could be generalized? Hmmm, likely not in a useful/elegant way. The underlying functionality of these is retained, so any downstream user could recreate them as needed. 

Ok, I will retract. Remove!

john-science:
> Oh yeah, totally agree here. I wonder if they could be generalized?

Yeah, I was thinking something like:

```python
# class ArmiObject
    def getNuclide(self, nuclide, units="mass"):
        """Note: units needs to be in ("mass", "frac", "moles")"""
```

Or, you know, something like that.  Some sort of more _general_ method that replaces all this stuff.

#just-thinking-aloud

albeanth:
> > Oh yeah, totally agree here. I wonder if they could be generalized?
> 
> Yeah, I was thinking something like:
> 
> ```python
> # class ArmiObject
>     def getNuclide(self, nuclide, units="mass"):
>         """Note: units needs to be in ("mass", "frac", "moles")"""
> ```
> 
> Or, you know, something like that. Some sort of more _general_ method that replaces all this stuff.
> 
> Also, I note that we already have `ArmiObject.getNumberDensity()`, which takes a nuclide name and does _most_ of the heavy lifting all these other methods do; it's just a difference of units.
> 
> #just-thinking-aloud

You could always do a poll in the discussion on what would be useful!

armi/reactor/tests/test_blocks.py
john-science, line 1082:
```suggestion
        self.assertAlmostEqual(cur, ref, places=9)
```

armi/reactor/tests/test_blocks.py
john-science, line 1086:
```suggestion
        self.assertAlmostEqual(cur, ref, places=9)
```

armi/reactor/components/__init__.py
john-science: There is one place downsteram where this is used as a static method, so it'll be less work to just keep it that way.

john-science: done

PR #1670, opened 2024-03-26
Adding a little code coverage to key Assembly methods

john-science:
## What is the change?

Here I add a little code coverage to three `Assembly` methods:

* `Assembly.add()`
* `Assembly.getArea()`
* `Assembly.getDim()`

Also, `Assembly.add()` had a couple lines of broken code that I removed.

This was a request. And these are important methods.

## Extra Note

I would just like to document somewhere that I think the `Assembly.getArea()` method is properly insane. When an `Assembly` has no children/`Blocks`, this method returns a "default area" of `1.0`.  Surely that should be an error? Or it should return `None`. Or it should return `0.0`. Returning a valid numerical value just seems crazy.

Thoughts?

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_assemblies.py
+++ b/armi/reactor/tests/test_assemblies.py
@@ -368,10 +368,14 @@ class Assembly_TestCase(unittest.TestCase):
             :id: T_ARMI_ASSEM_DIMS0
             :tests: R_ARMI_ASSEM_DIMS
         """
+        # Default case: for assemblies with no blocks
+        a = HexAssembly("TestAssem", assemNum=10)
+        self.assertEqual(a.getArea(), 1)
+
+        # more realistic case: a hex block/assembly
         cur = self.assembly.getArea()
         ref = math.sqrt(3) / 2.0 * self.hexDims["op"] ** 2
-        places = 6
-        self.assertAlmostEqual(cur, ref, places=places)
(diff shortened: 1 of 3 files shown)


armi/reactor/tests/test_assemblies.py
```diff
@@ -363,10 +368,14 @@ def test_getArea(self):
...
+        # Default case: for assemblies with no blocks
+        a = HexAssembly("TestAssem", assemNum=10)
+        self.assertEqual(a.getArea(), 1)
+
+        # more realistic case: a hex block/assembly
         cur = self.assembly.getArea()
         ref = math.sqrt(3) / 2.0 * self.hexDims["op"] ** 2
-        places = 6
-        self.assertAlmostEqual(cur, ref, places=places)
+        self.assertAlmostEqual(cur, ref, places=6)
```
albeanth, line 378: the precision of this can be increased. in my testing you can even swap it with a `self.assertEqual(cur, ref)`

albeanth (approved):

albeanth:
> When an Assembly has no children/Blocks, this method returns a "default area" of 1.0. Surely that should be an error? Or it should return None. Or it should return 0.0. Returning a valid numerical value just seems crazy.

Yeah I agree. It's odd. If you make the change, do a lot of other things break? 

@ntouran @jakehader any historical significance here?

PR #1671, opened 2024-03-28
Removing stub from isotopic depletion init file

### LINKED ISSUE #1458: Remove unnecessary `isotopicDepletion` stub
This file is not useful. If a user wants to use the `STACK_ORDER.DEPLETION` constant, they can just use it directly from the `interfaces` module. None of the other imports in this file are used.

https://github.com/terrapower/armi/blob/9da65879499bda3b9f1e9829e3d9f935275eeaa7/armi/physics/neutronics/isotopicDepletion/__init__.py#L15-L23

Unfortunately we have an internal plugin that directly imports `ORDER` from here. So we can't just delete this. But it should be slated for removal.

john-science:
## What is the change?

Removing some unnecessary stub code from the isotopic depletion `__init__.py` module-identification file.

To close #1458

---

## Checklist


armi/physics/neutronics/isotopicDepletion/__init__.py
```diff
@@ -13,11 +13,3 @@
 # limitations under the License.
 
 """The depletion physics package contains utility/framework code related to the physics of transmutation and decay."""
-# ruff: noqa: F401
```
keckler, line 16: Is there a good reason we don't just remove this file entirely? `__init__.py` files are not required beyond Python 3.3, and I don't see any imports breaking if we remove this.

Side point is that we don't even mention this `isotopicDepletion` namespace in this discussion:
https://github.com/terrapower/armi/blob/827d223cd6fc085c039390a2c3a0391f4e611fe5/armi/physics/__init__.py#L15-L46

john-science: > Is there a good reason we don't just remove this file entirely? __init__.py

@keckler @albeanth Sorry, was this the confusion originally?

In Python, you MUST have a file called `__init__.py` in every single folder that you want to put Python files in.  This is part of the language, and how the Python interpreter knows to include "folders of Python files" in importable libraries.  (This requirement has softened a bit in recent years, but is still firm in some scenarios.)

As long as there are other Python files in this folder, the `__init__.py` files exists.

keckler: This is not true. See, for instance: https://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-init-py-trap

By including that `__init__.py` file, we are preventing the use of namespace packages at that level. Perhaps that IS something we want to do, but that would come as a surprise to me, and it is also a bit counter to some text we have in the codebase here: https://github.com/terrapower/armi/blob/827d223cd6fc085c039390a2c3a0391f4e611fe5/armi/physics/__init__.py#L48

john-science: Okay, I tested it. Deleted.

john-science:
I have tested this thoroughly downstream, and found no issues of any kind.

Issue #1690, opened 2024-04-18
Remove silent overwriting of `shutil.copy` in ARMI utils


opotowsky:
`safeCopy` silently overwrites `shutil.copy` in ARMI:

https://github.com/terrapower/armi/blob/aa758f3a49d9e71f0ba4d616fcaf49df4091bfa5/armi/utils/__init__.py#L845-L847

I'm making this ticket so that when `shutil.copy` is replaced in all downstream plugins with `safeCopy`, I want to remove this menace from the code base. 

I also need to check in with other ARMI users so they have a heads up on a potentially breaking change (so, if shutil is being leveraged, they can do nothing and revert to the real `shutil.copy`, or they can also swap to explicitly using `safeCopy`)

john-science:
I agree, silently over-writing a standard library tool whenever you `import armi` is a bad solution to whatever problem this was trying to solve.

I would say "let's do it right now!", but since it's an API-breaking change, we will be slow and deliberate.

@drewj-usnctech Just FYI.

onufer:
this seems super low priority. I agree it is confusing and surprising but its going to be a lot of work to change and very little benifit.

opotowsky:
Thanks for reopening @onufer 

I grepped around and it won't be a huge lift. I agree it's low priority, but I'd like to get in the habit of paying back some tech debt and this is an easy one.

john-science:
> this seems super low priority.

I can see that from a "fixing this doesn't help us model a reactor" stand point. I really can.

But, from a quality perspective, this is a nasty piece of ugly spaghetti code that makes us look unprofessional and sloppy.

Thus my concern.

Also, I could fix this problem in like... 2 minutes in each downstream repo that is affected. So, it's just the easiest thing in the world to fix.

opotowsky:
Agreed. It actually can be done piecemeal, and doesn't need massive coordination.

1. replace `shutil.copy` with `safeCopy` in ARMI
2. repeat one by one for each plugin
3. last, remove the `shutil.copy` overwrite in ARMI.

PR #1691, opened 2024-04-19
Fix `safeCopy` to work on Linux

opotowsky:
## What is the change?

Replace `shutil.copyfile` and `shutil.copymode` with a `copy` (windows) or `cp` (linux) command. This accomplishes the same thing as before (file copy with metadata). 

We are NOT taking away the fact that `safeCopy` overrides `shutil.copy` in this PR, because there are significant downstream impacts and we need to check in with other ARMI users first. That is being tracked in the ticket #1690.

Depending on how permissions are set up, this doesn't work on all linux systems.

---

## Checklist


opotowsky:
More internal testing/processes are needed before this can be merged, but it's ready for a review, at least

armi/utils/__init__.py
```diff
@@ -810,8 +810,14 @@ def safeCopy(src: str, dst: str) -> None:
     if os.path.isdir(dst):
         dst = os.path.join(dst, os.path.basename(src))
     srcSize = os.path.getsize(src)
-    shutil.copyfile(src, dst)
-    shutil.copymode(src, dst)
+    # Historical note: this used to run shutil.copyfile then shutil.copymode,
+    # but the latter clashed with permissions on a linux system. Both copy and
+    # cp -a accomplish the same thing.
```
john-science, line 815: Personal preference. I don't like historical notes in the code. That's what commits and PRs are for.

When I'm looking at the code, I am trying to understand what it is doing, and I would rather have a comment that describes THAT.

opotowsky: Will fix!

armi/utils/__init__.py
```diff
@@ -810,8 +810,14 @@ def safeCopy(src: str, dst: str) -> None:
...
+    if os.name == "nt":
+        cmd = f'copy "{src}" "{dst}"'
+    else:
+        cmd = f'cp -a "{src}" "{dst}"'
+    os.system(cmd)
     while True:
```
opotowsky, line 821: note that we need an exit strategy here. Decide on a max time then throw an exception?

armi/utils/__init__.py
opotowsky: done

armi/utils/__init__.py
opotowsky: done

armi/utils/__init__.py
```diff
@@ -810,8 +810,14 @@ def safeCopy(src: str, dst: str) -> None:
...
         dst = os.path.join(dst, os.path.basename(src))
     srcSize = os.path.getsize(src)
-    shutil.copyfile(src, dst)
-    shutil.copymode(src, dst)
+    # Historical note: this used to run shutil.copyfile then shutil.copymode,
+    # but the latter clashed with permissions on a linux system. Both copy and
+    # cp -a accomplish the same thing.
+    if os.name == "nt":
+        cmd = f'copy "{src}" "{dst}"'
+    else:
```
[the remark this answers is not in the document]
opotowsky: done

armi/utils/__init__.py
```diff
@@ -806,17 +806,39 @@ def merge(self, *otherDictionaries) -> None:
 
 def safeCopy(src: str, dst: str) -> None:
     """This copy overwrites ``shutil.copy`` and checks that copy operation is truly completed before continuing."""
+    # Convert files to OS-independence
+    src = os.path.abspath(src)
+    dst = os.path.abspath(dst)
```
opotowsky, line 811: Hey @john-science I wanted to check with you on this choice. It's needed since the file paths were used to being handled via python and so going straight to the command line broke some file copies.

john-science: Changing everything to absolute paths is generally a good idea and I like it.

armi/utils/__init__.py
```diff
@@ -806,17 +806,39 @@ def merge(self, *otherDictionaries) -> None:
...
     while True:
         dstSize = os.path.getsize(dst)
         if srcSize == dstSize:
             break
         time.sleep(waitTime)
+        totalWaitTime += waitTime
+        if totalWaitTime > maxWaitTime:
+            runLog.warning(
+                f"File copy from {dst} to {src} has failed due to exceeding "
+                + f"``maxWaitTime`` of {maxWaitTime/60} min."
```
john-science, line 838:
```suggestion
                + f"a maximum wait time of {maxWaitTime/60} minutes."
```

john-science: Just minor phrasing.  I like log statements to be more "language-ish" than "code-ish".  But I won't die on this hill.

armi/utils/__init__.py
```diff
@@ -806,17 +806,39 @@ def merge(self, *otherDictionaries) -> None:
...
+    if "win" in sys.platform:
+        cmd = f'copy "{src}" "{dst}"'
+    elif "linux" in sys.platform:
+        cmd = f'cp -a "{src}" "{dst}"'
+    else:
+        runLog.warning(
+            "Cannot perform ``safeCopy`` on files because ARMI only supports "
+            + "Linux and Windows."
+        )
+    os.system(cmd)
```
john-science, line 827: The "else" case here will raise an error because the `cmd` variable doesn't exist yet.

opotowsky: Oh, oops. 

This begs the question:
Do we want a ton of "if windows elif linux else kill the run" all throughout the code base? Why not just kill the run from the getgo?

john-science: Well, we need something here, otherwise this is just invalid Python.  Hmmm... I will make a recommend.

armi/utils/__init__.py
```diff
@@ -806,17 +806,38 @@ def merge(self, *otherDictionaries) -> None:
...
         dst = os.path.join(dst, os.path.basename(src))
     srcSize = os.path.getsize(src)
-    shutil.copyfile(src, dst)
-    shutil.copymode(src, dst)
+    if "win" in sys.platform:
+        cmd = f'copy "{src}" "{dst}"'
+    elif "linux" in sys.platform:
+        cmd = f'cp "{src}" "{dst}"'
+    else:
+        runLog.warning(
```
john-science, line 820:
```suggestion
        raise OSError(
```

armi/utils/__init__.py
```diff
@@ -806,17 +806,38 @@ def merge(self, *otherDictionaries) -> None:
...
 def safeCopy(src: str, dst: str) -> None:
     """This copy overwrites ``shutil.copy`` and checks that copy operation is truly completed before continuing."""
-    waitTime = 0.01  # 10 ms
+    # Convert files to OS-independence
+    src = os.path.abspath(src)
+    dst = os.path.abspath(dst)
     if os.path.isdir(dst):
         dst = os.path.join(dst, os.path.basename(src))
     srcSize = os.path.getsize(src)
-    shutil.copyfile(src, dst)
```
jakehader, line 813: Can we wrap this in a try/except and then proceed with the Windows / Linux check? Additionally, the detection of Windows / Linux and any language specific utility code requirements would be helpful at the start on ARMI. Is it possible to search the code base for these (maybe in a couple places where we use if "Win..." or if "Linux" and then pull those checks to ARMI __init__ or bootstrap? Just thinking about a potential speed up for failure if something is missing

opotowsky: I don't like keeping code in that is that fragile. I think the change I made is a little more robust

As far as the OS-check, I agree we shouldn't be adding a gazillion checks throughout the code. We should catch the OS failure during init or bootstrap like you suggest, then we can safely do "if win this else that" throughout the code.

But @john-science gets the final call there.

opotowsky: Actually @jakehader: I can have the windows `if` do the old/shutil approach, and the linux `if` do the `cp`/command line stuff. I'm fine with that. We know it works on windows and has for years.

opotowsky: @jakehader for the OS check, we are going to handle that in a different ticket. The code for now should stay until that ticket is addressed. It's out of scope for this PR.

And for windows, @john-science prefers the `copy` approach over the `shutil` approach. I don't care which we go with, but it seems cleaner to me to have parallel approaches implemented, so I lean towards the way I have it coded now. 

So as of right now John is just awaiting me to complete some internal testing to prove this PR is ready.

john-science: @opotowsky @jakehader That's a nice idea, I like it.  But I decided to make that it's own [ticket](https://github.com/terrapower/armi/issues/1693).

This is for Arrielle's expediency right now.  But more so, I like to keep "one feature = one PR" when I can.

opotowsky: HAHA @john-science jinx
https://github.com/terrapower/armi/issues/1694

opotowsky: feel free to close my duplicate ticket

john-science (approved):

PR #1705, opened 2024-05-16
Fix safeCopy on Windows

opotowsky:
## What is the change?

1. Reverting the `os.system(cmd)` for windows to `shutil.copyfile`/`shutil.copymode`
2. Reducing the wait time to 5 minutes 

1. `copy` via `os.system` on Windows was taking an obscenely long time. And really we just needed the update from #1691 for linux functionality.
2. 30 min for a wait time is bonkers, so given some tests run we think we can safely reduce it to 5 min.

---

## Checklist

- [x] ~[Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify any new/changed code.~ No updates needed

### DIFF
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -813,17 +813,18 @@ def safeCopy(src: str, dst: str) -> None:
         dst = os.path.join(dst, os.path.basename(src))
     srcSize = os.path.getsize(src)
     if "win" in sys.platform:
-        cmd = f'copy "{src}" "{dst}"'
+        shutil.copyfile(src, dst)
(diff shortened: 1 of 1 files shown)


john-science (approved):

john-science:
Approved.

## Notes

1. We discussed a sliding scale (bigger files get a longer time out).  But.. there is no meaningful way to choose that sliding scale. So I think the "single number timeout for all files" solution that Arrielle implemented here.
2. This function exists for a bonkers reason.  Long term, I think we should move this function out of ARMI and into the cluster that is causing the problem.

PR #1726, opened 2024-06-11
Isolating jagged arrays for numpy compatibility

mgjarrett:
## What is the change?

Implement a `JaggedArray` class to interface between our jagged data structures (e.g., pin-level data within a block) and HDF5 database without using numpy to flatten the jagged array.

The current code feeds jagged arrays to numpy and then uses numpy to flatten them for storing in the database and reconstitute them when reading data back from the database. Numpy no longer supports jagged arrays, so we are currently pinned to version 1.23.5 (the latest release is 1.26.4).

The current version of numpy we are using gives a `VisibleDeprecationWarning` about this:

```
.\armi\bookkeeping\db\database3.py:1361: VisibleDeprecationWarning: Creating an ndarray from ragged nested
sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If
you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  data = numpy.array(layout.location)[layoutIndicesForType][
```

This change introduces a `JaggedArray` class that handles the packing into and unpacking out of 1D data structures for the HDF5 database. This allows us to use a current version of numpy.

There is a discussion about this topic: https://github.com/terrapower/armi/discussions/1669

Also, I found this open ticket (https://github.com/terrapower/armi/issues/246) _after_ having done all of the work in this PR. I didn't realize how much others had already thought about how to work around this. Well, I guess I'm throwing my potential solution in the ring. As was noted many times in that ticket, matching the performance of numpy might be the main challenge. Hopefully, the amount of jagged data we have is small enough that the performance hit isn't too significant.

Notes about numpy compatibility
-------------------------------------

Using a current version of numpy, if you try to pass jagged data into `numpy.array()`, you get an error: 

```python
>>> a = [1, 2, 3]
>>> b = [4, 5, 6, 7]
>>> c = [0, 0, 1]
>>> import numpy
>>> d = numpy.array((a, c))
>>> e = numpy.array((b, d))
  File "<stdin>", line 1, in <module>
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.
```

You can get numpy to accept it if you pass in `dtype=object`, but this data type doesn't have the features needed to facilitate flattening the array to store in HDF5, i.e., `array.flatten()` doesn't do anything. So I don't know whether this would help us at all. I'm not 100% positive, but I think we'd still be in the same position of needing to implement the packing and unpacking routines ourselves.

```python
>>> e = numpy.array((b, d), dtype=object)
>>> print(e)
[list([4, 5, 6, 7]) array([[1, 2, 3],
                           [0, 0, 1]])]
>>> e.flatten()
array([list([4, 5, 6, 7]), array([[1, 2, 3],
                                  [0, 0, 1]])], dtype=object)
```

---

## Checklist


armi/bookkeeping/db/database3.py
```diff
@@ -912,7 +1077,16 @@ def _writeParams(self, h5group, comps):
                     attrs[_SERIALIZER_NAME] = paramDef.serializer.__name__
                     attrs[_SERIALIZER_VERSION] = paramDef.serializer.version
                 else:
```
mgjarrett, line 1079: Note: we are almost always ending up in this `else` block. Looking through all of the parameter definitions in the framework, the only one that specifies a `Serializer` is `flags`:

https://github.com/terrapower/armi/blob/8cfc244ced1767bf71376d41796c618165c61bb7/armi/reactor/composites.py#L60

armi/bookkeeping/db/database3.py
```diff
@@ -912,7 +1077,16 @@ def _writeParams(self, h5group, comps):
...
+                    else:
+                        jagged = False
+                    data = (
+                        JaggedArray(temp, paramDef.name)
+                        if jagged
+                        else numpy.array(temp)
+                    )
                     del temp
 
             # Convert Unicode to byte-string
```
mgjarrett, line 1092: The next 3 `if` statements should all be mutually exclusive. Might want to make it an `if`/`elif` block.

armi/bookkeeping/db/database3.py
```diff
@@ -1585,6 +1771,15 @@ def packSpecialData(
...
+    if isinstance(arrayData, JaggedArray):
+        attrs["jagged"] = True
+        attrs["offsets"] = arrayData.offsets
+        attrs["shapes"] = arrayData.shapes
+        attrs["noneLocations"] = arrayData.nones
+        if len(arrayData.nones > 0):
+            data = replaceNonesWithNonsense(data, paramName, arrayData.nones)
+        return data, attrs
+
     # conform non-numpy arrays to numpy
```
mgjarrett, line 1783: We might want to consider `assert`ing that the data passed into `packSpecialData` is either a `numpy.ndarray` or `JaggedArray` instance. If some kind of `list`-of-`tuple`, `tuple`-of-`list`s, etc. is passed in, it should probably be converted to either a `JaggedArray` (if jagged) or `numpy.ndarray` (if not jagged) and then passed in.

armi/physics/neutronics/__init__.py
```diff
@@ -252,18 +250,20 @@ def applyEffectiveDelayedNeutronFractionToCore(core, cs):
         core.p.betaDecayConstants = numpy.array(decayConstants)
 
         reportTableData.append(("Total Delayed Neutron Fraction", core.p.beta))
-        reportTableData.append(
-            ("Group-wise Delayed Neutron Fractions", core.p.betaComponents)
-        )
-        reportTableData.append(
-            ("Group-wise Precursor Decay Constants", core.p.betaDecayConstants)
-        )
+        for i, betaComponent in enumerate(core.p.betaComponents):
```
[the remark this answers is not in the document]
john-science: I'm not a big fan of `tabulate` in the first place. If you need to change the version of `tabulate` we use, I am 100% okay with that.

armi/bookkeeping/db/database3.py
john-science: Yup, that sounds right.

Though we might be adding some more serializers for this ticket:  https://github.com/terrapower/armi/discussions/1688

john-science:
I love this! Thanks for getting started!

Can we close this until the PR until you are ready for review?  https://terrapower.github.io/armi/developer/tooling.html#don-t-open-until-it-is-ready

Thanks!

armi/bookkeeping/db/database3.py
```diff
@@ -100,6 +94,168 @@ def getH5GroupName(cycle: int, timeNode: int, statePointName: str = None) -> str
     return "c{:0>2}n{:0>2}{}".format(cycle, timeNode, statePointName or "")
 
 
+class JaggedArray:
```
john-science, line 97: By preference, I would add this new class to it's own file. (This is safe, since `Database3` will import it, but it doesn't import `Database3`.)

The idea is simple: the file `Database3.py` is already 2000 lines long, and has several classes in it. It's not _always_ necessary, of course, but "one file = one class" (when it works) does make it easier to read a code base.

Thoughts?

armi/bookkeeping/db/database3.py
```diff
@@ -100,6 +94,168 @@ def getH5GroupName(cycle: int, timeNode: int, statePointName: str = None) -> str
...
+    def __iter__(self):
+        """Iterate over the unpacked list."""
+        return iter(self.unpack())
+
+    def __contains__(self, other):
+        return other in self.flattenedArray
+
+    @staticmethod
+    def flatten(x):
+        if isinstance(x, list):
```
john-science, line 163: I'm a little confused about this. Surely you could also flatting a tuple as well as a list?

If this is something to do with how you are treating tuples and lists differently, maybe:

1. We need a docstring explaining the limits of this method.
2. This should be a private method, not static, so no one outside this class can see it.

Thoughts?

armi/bookkeeping/db/database3.py
```diff
@@ -100,6 +94,168 @@ def getH5GroupName(cycle: int, timeNode: int, statePointName: str = None) -> str
     return "c{:0>2}n{:0>2}{}".format(cycle, timeNode, statePointName or "")
 
 
+class JaggedArray:
+    def __init__(self, jaggedData, paramName):
+        """
+        Take a list of numpy arrays or lists and flatten them into a single 1D array.
+
```
john-science, line 101: It is more _typical_ to have a long docstring on the class level, not the `__init__` method. Can we move this docstring up, so the class has a docstring?

armi/bookkeeping/db/database3.py
```diff
@@ -100,6 +94,168 @@ def getH5GroupName(cycle: int, timeNode: int, statePointName: str = None) -> str
...
+        jaggedData: list of numpy.ndarray
+            A list of numpy arrays (or lists) to be flattened into a single array
+        paramName: str
+            The name of the parameter represented by this data
+        """
+        offset = 0
+        flattenedArray = []
+        offsets = []
+        shapes = []
+        for array in jaggedData:
```
john-science, line 122: MINOR thing: "array" is a bad variable name in Python because it conflicts with the Python standard library "import array".

You're not the first person to hit this annoyance. Sorry!

john-science:
@mgjarrett  I'm really excited about this PR.  Holler if you need anything.

armi/bookkeeping/db/database3.py
```diff
@@ -869,6 +1025,15 @@ def _compose(self, comps, cs, parent=None):
         return comp
 
     def _writeParams(self, h5group, comps):
+        def _getShape(array):
+            """Get the shape of a numpy.ndarray or list."""
+            if isinstance(array, numpy.ndarray):
+                return array.shape
+            elif isinstance(array, list):
+                return (len(array),)
```
john-science, line 1033:
```suggestion
        def _getShape(arr):
            """Get the shape of a numpy.ndarray or list."""
            if isinstance(arr, numpy.ndarray):
                return arr.shape
            elif isinstance(arr, list):
                return (len(arr),)
```

john-science:
@mgjarrett I hope you don't mind, I fixed the merge conflicts in this branch, so the CI would run again.

And while I was at it, I moved `JaggedArray` to it's own file. I can't help myself, sorry!

armi/bookkeeping/db/database3.py
mgjarrett: > So, your implied thesis here is that the place where ARMI (and its ecosystem) mostly ends up `JaggedArray`s is when writing data to the DB?

Sorry, I was out for a couple weeks and just now circling back to this.

Yes, I don't know of anywhere ARMI creates a jagged array and tries to pass it into `numpy` other than when it's writing to a database. It was a shortcut for packing data; we would just do `a = numpy.array(jaggedData)`, and then `a.flatten()` to get a 1D representation of it.

The jagged data mainly comes from pin-level data in blocks -- for example, the parameter `b.p.linPowByPin`. During the course of a simulation, each block with pins has this parameter stored as a 1D array. When it is time to store this data in the database, ARMI puts all of the `linPowByPin` data into a single array. Since different blocks can have different number of pins, this is a jagged array.

armi/bookkeeping/db/database3.py
mgjarrett: Yeah, it should include `tuple` and also probably `numpy.ndarray`, just in case.

I should add a docstring either way. The only reason I wouldn't want to make it private is so that the `JaggedArray` "quacks" like a numpy array with the `flatten()` method.

mgjarrett:
> @mgjarrett I hope you don't mind, I fixed the merge conflicts in this branch, so the CI would run again.
> 
> And while I was at it, I moved `JaggedArray` to it's own file. I can't help myself, sorry!

Thank you! I'm hoping to get back to this today, look at your comments and changes and clean up whatever needs to be fixed.

mgjarrett:
I did some simple profiling to make sure there's not an unreasonable increase in run time between the old numpy version and the new implementation. (A small increase is probably acceptable because this will enable us to move to Python 3.11 or 3.12, which is reportedly significantly faster than Python 3.9).

The test case I created for profiling is quite simple, consisting of two separate runs. For the first run:

1. Load a reactor operator from inputs
2. Populate `b.p.linPowByPin` with random numbers on each block (this is a jagged dataset)
3. Write the reactor to a database

For the second run:

1. Load from the database saved in the previous run

The time to write the database is about  11-12 seconds, compared to an overall run time of about 100 s. Most of the database writing is related to writing the layout. The `_writeParams` portion is less than 5 seconds, and the `packSpecialData` portion of the database write is less than 1 second in both cases. The database write is slightly faster in the base case (current nala main, numpy 1.23.5), than the new implementation.

Profiling results for writing to the database
-----------------------------------------------

  | writeToDB (s) | _writeParams (s) | packSpecialData (s)
-- | -- | -- | --
Nala Main | 11.5 | 4.39 | 0.336
Jagged Arrays (PR#589) | 11.8 | 4.72 | 0.503
Jagged Arrays (PR#589) (numpy 1.23.5) | 11.9 | 4.97 | 0.559

Profiling results for reading from the database
-----------------------------------------------

Again, the new implementation is slightly slower than the old implementation, although not by much (1.4 s vs. 1.5 s). PR#589 is slower across the board, even for parts of the code that are not touched by the `JaggedArray` update. This difference persisted after multiple runs. It is possibly due to the newer version of numpy (1.26.4) is slower than the older version (1.23.5) on Python 3.9, or something else that has changed in the code. We can't compare on Python 3.11 or 3.12 because numpy 1.23.5 is not compatible. Running the new code with numpy 1.23.5 did slightly improve the run time for `unpackSpecialData` and for the `_readParams`.

  | load (s) | _readParams (s) | unpackSpecialData (s)
-- | -- | -- | --
Nala Main Run 1 | 36.3 | 3.78 | 1.45
Nala Main Run 2 | 40.9 | 3.84 | 1.35
Nala Main Run 3 | 41.6 | 4.28 | 1.35
Jagged Arrays (PR#589) Run 1 | 40.9 | 4.04 | 1.53
Jagged Arrays (PR#589) Run 2 | 40.3 | 4.16 | 1.66
Jagged Arrays (PR#589) Run 3 | 41.6 | 3.75 | 1.40
Jagged Arrays (PR#589) (numpy 1.23.5) | 39.2 | 3.62 | 1.31

john-science:
@mgjarrett Your timing results are really interesting. I suppose Python 3.9 is the target version now, but 3.11 will be soon.

On the one hand, people might not like the speed decrease, on the other hand we need to be able to unlock modern versions of NumPy and Python. And both of those are blocked by this NumPy issue.

opotowsky:
> @mgjarrett Your timing results are really interesting. I suppose Python 3.9 is the target version now, but 3.11 will be soon.
> 
> On the one hand, people might not like the speed decrease, on the other hand we need to be able to unlock modern versions of NumPy and Python. And both of those are blocked by this NumPy issue.

The speed decrease is not even noticeable (considering the normal variation in run times) for an actual case. I'm quite pumped about this!

john-science:
@mgjarrett When this PR is ready for review, please move it from Draft to normal. Right now, I can't tell when you think this PR is ready for prime time.

(And again, ARMI discourages opening Draft PRs.)

armi/bookkeeping/db/jaggedArray.py
```diff
@@ -0,0 +1,191 @@
...
+"""
+
+from typing import List, Optional
+
+import numpy as np
+
+from armi import runLog
+
+
+class JaggedArray:
```
john-science, line 27: I suppose I wouldn't _hate_ having unit tests directly for this class.

mgjarrett: Fair enough, I'll work on that

john-science: Okay, thanks.

Other than that, I think this PR is ready for prime time. 

Holler when you're ready, and I'm happy to get this merge party started.

mgjarrett: Okay, I added some unit tests. Mostly testing the ability to round-trip data to the HDF5 file and back. I think it may have improved the coverage a bit.

john-science (approved):
Great work!

Thanks so much! I have really been wanting this feature, but couldn't find time on the schedule.

You rock.

PR #1729, opened 2024-06-12
Clearing worker processor instances of reactor state after every interaction

zachmprince:
## What is the change?

The PR creates a method in `Operator` that is called `_finalizeInteract` which is called inside interface loop of `_interactAll`. This method is overridden by `OperatorMPI` to "reset" worker processors by broadcasting a command to force execution of `_resetWorker`.

The purpose of this change to force worker resets after each interaction is complete, instead of relying on plugins to return the correct `mpiActionRequiresReset` during `MpiAction` invocations. The goal was to reduce memory when going to subsequent interfaces that don't require a distributed state. However, the addition of this reset call is not reducing memory on worker processors, which means the "reset" implementation may be flawed. However, the change proposed in this PR is still appropriate in principle and improvements to the reset functionality can be improved in the future.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/memoryProfiler.py
+++ b/armi/bookkeeping/memoryProfiler.py
@@ -342,7 +342,7 @@ class SystemAndProcessMemoryUsage:
         self.processMemoryInMB: Optional[float] = None
         if _havePsutil:
             self.percentNodeRamUsed = psutil.virtual_memory().percent
-            self.processMemoryInMB = psutil.Process().memory_info().rss / (1012.0**2)
+            self.processMemoryInMB = psutil.Process().memory_info().rss / (1024.0**2)
 
     def __isub__(self, other):
         if self.percentNodeRamUsed is not None and other.percentNodeRamUsed is not None:
--- a/armi/operators/operatorMPI.py
+++ b/armi/operators/operatorMPI.py
@@ -163,8 +163,8 @@ class OperatorMPI(Operator):
                 note = context.MPI_COMM.bcast("wait", root=0)
                 if note != "wait":
                     raise RuntimeError('did not get "wait". Got {0}'.format(note))
-            elif cmd == "clear":
-                runLog.info("Workers are being cleared.")
+            elif cmd == "reset":
+                runLog.info("Workers are being reset.")
             else:
                 # we don't understand the command on our own. check the interfaces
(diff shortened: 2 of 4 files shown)


opotowsky (approved):
🚀

armi/operators/operatorMPI.py
```diff
@@ -163,6 +163,8 @@ def workerOperate(self):
                 note = context.MPI_COMM.bcast("wait", root=0)
                 if note != "wait":
                     raise RuntimeError('did not get "wait". Got {0}'.format(note))
+            elif cmd == "reset":
```
john-science, line 166: I actually think ARMI has _way_ too many log statements as it is.  People add them, forget they exist, and never look at them again. Happens all the time.

Why is this one necessary for people to see during ever run?

zachmprince: Definitely helped with confirming that this implementation was working. This specific one is only outputted on worker nodes, but there is one in `_finalizeInteract` that is logged on root. I can get rid of both, one of them, or none of them. I'll leave that to your decision.

john-science:
Quick Question:

You added new code, why are there no tests of this?  (Did you try, and it was too hard to test?)

armi/operators/operatorMPI.py
zachmprince: Indeed. I definitely want to avoid what's in the `else` block with this new command.

zachmprince:
> Quick Question:
> 
> You added new code, why are there no tests of this? (Did you try, and it was too hard to test?)

I just didn't think of it. I can add a test.

armi/bookkeeping/memoryProfiler.py
```diff
@@ -342,7 +342,7 @@ def __init__(self):
         self.processMemoryInMB: Optional[float] = None
         if _havePsutil:
             self.percentNodeRamUsed = psutil.virtual_memory().percent
-            self.processMemoryInMB = psutil.Process().memory_info().rss / (1012.0**2)
+            self.processMemoryInMB = psutil.Process().memory_info().rss / (1024.0**2)
```
john-science, line 345: lawls

zachmprince:
@john-science @opotowsky Changed log message to `extra` and added a test in `test_mpiFeatures`.

john-science (approved):

Issue #1348, opened 2023-07-17
Force Parameters to have descriptions


john-science:
When constructing a `Parameter` object, we already have multiple checks to ensure the `Parameter` is valid:

https://github.com/terrapower/armi/blob/59388d29f8929178e0e1aa4d8f0f160095a2b739/armi/reactor/parameters/parameterDefinitions.py#L250-L253

Why not add some to ensure there is a valid `description` field?  I want to say the same for `units`, but many parameters are unitless, so we would have to change "unitless" to be a non-empty string.

```python
assert description is not None and len(description)
```

> After all, what good is a `Parameter` if no one can tell what it is?

john-science:
@keckler FYI

@sombrereau @drewj-usnctech  Would this be a problem for either of you?

jakehader:
I like the idea of enforcing that these are set. Could be a breaking change, but probably a good one? Maybe we can start implementing warnings to fix before we enforce so that folks have a chance to change? I also like the idea of enforcing locations would be good too.

drewj-usnctech:
On board for this. Parameters should have descriptions. I think setting `units=None` should still be supported as there are _some_ cases where parameters really are unitless. Like number of things? It would be a small sample because the vast majority of things have units.

Maybe a non-None field like `units=UNITLESS`? `None` seems simpler tbh

john-science:
I am changing this ticket to only FORCE Parameter descriptions, not units.

And I already have a `DeprecationWarning` in place:

https://github.com/terrapower/armi/blob/59388d29f8929178e0e1aa4d8f0f160095a2b739/armi/reactor/parameters/parameterDefinitions.py#L255-L259

The plan is:

* This version of ARMI will be released with the `DeprecationWarning`.
* The very next version of ARMI will change that to a hard `Error`.
* Then I will close this ticket.

Estimated time to close this ticket: 2 months (of waiting).

john-science:
I have put this on the ARMI `0.4.0` [Roadmap](https://github.com/terrapower/armi/wiki/Roadmap).

Really, the change in ARMI is no work at all. The time here will be spent on testing downstream projects to see if they will break.

PR #1733, opened 2024-06-16
Adding descriptions to two settings

### LINKED ISSUE #1727: Add descriptions for `copyFilesFrom` and `copyFilesTo` case settings
We have two case settings with no descriptions:
image

These settings need descriptions. From what I can tell, `copyFilesFrom` can be a list of any length, but then `copyFilesTo` needs to be either of zero-length, length-1, or the same length as `copyFilesFrom`. This should be described for the user. It would also be good to know when the file copying takes place so the user knows when the new files will be available.

john-science:
## What is the change?

Adding descriptions to two settings: `CONF_COPY_FILES_FROM` and `CONF_COPY_FILES_TO`.

Also, while I was at it, this work let me to:

* Improve the docstring on the method `MainInterface._moveFiles()`
* Add a `DeprecationWarning` that soon settings will _require_ descriptions (same as with [parameters](https://github.com/terrapower/armi/issues/1348)).
* I fixed some formatting and docstrings as I found them (sorry!)

To close #1727 

These two settings are slightly complicated, and interact heavily, so they should have descriptions that explain their use.

---

## Checklist

- [X] This PR has only [one purpose or idea](https://terrapower.github.io/armi/developer/tooling.html#one-idea-one-pr). (Except the random cleanup I did. Sorry!)

### DIFF
--- a/armi/bookkeeping/mainInterface.py
+++ b/armi/bookkeeping/mainInterface.py
@@ -28,6 +28,8 @@ from armi import operators
 from armi import runLog
 from armi import utils
 from armi.bookkeeping.db.database3 import Database3
+from armi.settings.fwSettings.globalSettings import CONF_COPY_FILES_FROM
+from armi.settings.fwSettings.globalSettings import CONF_COPY_FILES_TO
 from armi.utils import pathTools
 from armi.utils.customExceptions import InputError
 
@@ -46,9 +48,8 @@ class MainInterface(interfaces.Interface):
 
     Notes
     -----
-    Interacts early so that the database is accessible as soon as possible in the run.
-    The database interfaces interacts near the end of the interface stack, but the main
-    interface interacts first.
+    Interacts early so that the database is accessible as soon as possible in the run. The database
+    interfaces runs near the end of the interface stack, but the main interface interacts first.
     """
 
     name = "main"
@@ -64,10 +65,9 @@ class MainInterface(interfaces.Interface):
 
         Notes
         -----
-        This happens here rather than on the database interface, as the database
-        interacts near the end of the stack. Some interactBOL methods may be
-        dependent on having data in the database, such as calls to history tracker
-        during a restart run.
+        This happens here rather than on the database interface, as the database interacts near the
+        end of the stack. Some interactBOL methods may be dependent on having data in the database,
+        such as calls to history tracker during a restart run.
         """
         dbi = self.o.getInterface("database")
         if not dbi.enabled():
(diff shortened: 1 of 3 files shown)


armi/settings/setting.py
```diff
@@ -115,6 +115,10 @@ def __init__(
         """
         self.name = name
         self.description = description or name
+        runLog.warning(
+            f"DeprecationWarning: Setting {name} defined without description.",
+            single=True,
+        )
```
keckler, line 121: Isn't this deprecation warning going to be printed all the time...? Like, shouldn't it be wrapped under an if-statement?

armi/bookkeeping/mainInterface.py
```diff
@@ -80,30 +80,49 @@ def _activateDB(self):
...
+        - Can be of length zero (that just means no files will be copied).
+        - The file names listed can use the ``*`` glob syntax, to reference multiple files.
+
+
+        ``CONF_COPY_FILES_TO`` :
+
+        - List of directories to copy the files into.
+        - Can be of length zero; all files will be copied to the local dir.
+        - Can be of length one; all files will be copied to that dir.
+        - The only other valid length for this list _must_ be the same length as the "from" list.
```
keckler, line 112: This is all really useful context for a user. Unfortunately users won't typically see this info because it is in the docstring of a private method. Is there a better place to put this that will make it user-facing?

armi/bookkeeping/mainInterface.py
```diff
@@ -80,30 +80,49 @@ def _activateDB(self):
...
 
     def _moveFiles(self):
-        # check for orificed flow bounds files. These will often be named based on the
-        # case that this one is dependent upon, but not always. For example, testSassys
-        # is dependent on the safety case but requires physics bounds files.  now copy
-        # the files over
+        """
+        At the start of each run, arbitrary lists of user-defined files can be copied around.
+
+        This logic is controlled by the settings ``CONF_COPY_FILES_FROM`` & ``CONF_COPY_FILES_TO``.
```
keckler, line 98: I'm pretty certain that the capital snake case `CONF` variables will not get rendered as the actual string names, right? In situations like this, I never know if it would be better to use the variable name or the string name. One of the drawbacks of these `CONF` variables, IMO. 

I am not partial to either way, but perhaps consider if it would be better to just use the strings in this docstring instead of the variable names. Up to you.

armi/settings/setting.py
john-science: No, this is why I have `single=True`. That means that this warning will only ever be printed ONCE during a run.

keckler: Sorry, let me clarify. This warning will be printed indiscriminately. There is no check of if the setting actually has a description or not. The runLog statement is hit even if the setting has a description.

john-science, line 121:
```suggestion
        if not description or description in ("None", "none"):
            runLog.warning(
                f"DeprecationWarning: Setting {name} defined without description.",
                single=True,
            )
```

armi/bookkeeping/mainInterface.py
john-science: > I'm pretty certain that the capital snake case CONF variables will not get rendered as the actual string names, right?

Nope, in this situation, it will just literally say "CONF_XYZ". I thought that was better, but it's a close call, for me.

I can change it, if you like, sure.

armi/bookkeeping/mainInterface.py
john-science: Hmm... I can' think of where else it would belong.

armi/bookkeeping/mainInterface.py
```diff
@@ -80,30 +80,49 @@ def _activateDB(self):
...
+
+        ``CONF_COPY_FILES_FROM`` :
+
+        - List of files to copy (cannot be directories).
+        - Can be of length zero (that just means no files will be copied).
+        - The file names listed can use the ``*`` glob syntax, to reference multiple files.
+
+
+        ``CONF_COPY_FILES_TO`` :
```
john-science, line 107:
```suggestion
        This logic is controlled by the settings ``copyFilesFrom`` & ``copyFilesTo``.

        ``copyFilesFrom`` :

        - List of files to copy (cannot be directories).
        - Can be of length zero (that just means no files will be copied).
        - The file names listed can use the ``*`` glob syntax, to reference multiple files.


        ``copyFilesTo`` :
```

armi/settings/setting.py
john-science: Fixed

armi/bookkeeping/mainInterface.py
john-science: Done

keckler (approved):

PR #1743, opened 2024-06-20
We shouldn't be wiping the global elements durint tests

john-science:
## What is the change?

We have been calling `elements.factory()` during the unit tests, this PR removes that.

In some of our downstream unit testing, which is significantly more parallel, wiping the global `elements` object from memory, even temporarily, can cause other unit tests to fail.

This has always been quite hard to track down, since it is an async bug that only happens very rarely.  Happily, these calls were not necessary in the unit tests in the first place, so we can just remove them.

---

## Checklist


opotowsky (approved):
Easy!! Nice work!

PR #1744, opened 2024-06-21
Making MacOS unit tests parallel on GH Actions

john-science:
## What is the change?

1. Making MacOS unit tests parallel on GH Actions.
2. Again, our Python 3.7 tests are breaking for some numpy reason, so we're dropping 3.7. (Not worth our time to fix / support.)

> This reduces the GitHub run time for our macos tests from 14.5 minutes to 4 minutes!

I recently found an update to `runLog.py` that would fix various test parallelization issues. I am using it here to allow our MacOS unit tests on GitHub Actions to run in parallel. 

---

## Checklist


albeanth:
If we're dropping 3.7, the docs need to be updated too
https://github.com/terrapower/armi/blob/189075860cf2db6936c93181ce06fc5427520176/doc/user/user_install.rst?plain=1#L15

john-science:
> If we're dropping 3.7, the docs need to be updated too

Well, I was only planning on dropping CI for Python 3.7.  You should I should drop support as well?

Thoughts?

albeanth:
Hmmmm. Well, if we aren't testing for it in CI, then we aren't really maintaining it. And I don't think our docs should start off with using a version of Python that we don't test for. At least that's my thinking. What do you think?

john-science:
> We could update the docs to say install with Py 3.9+ but also have a footnote saying that 3.7 may work, it's just no longer supported.

Well, I guess we're doing this.

albeanth (approved):

Issue #1272, opened 2023-05-19
densities are ignored on custom isotopics when material is provided


onufer:
https://github.com/terrapower/armi/blob/76031431c1d96e17c1135874926bd24ac92a3e65/armi/reactor/blueprints/isotopicOptions.py#L359-L367

We should be able to support this. it should probably be elevated to more than a warning since it is completely ignoring using input.

A code fix where the density was accounted for would be better

jakehader:
@keckler - you may have some opinions on this with your benchmarking custom classes for materials. Have anything to add here with respect to your implementation?

keckler:
Thanks for checking! 

@onufer and I talked about this offline, and we are in agreement that this should be supported.

I think the biggest thing that needs to be considered is how the mass density that a user specifies should scale with thermal expansion. 

I think the most obvious or expected action would be that the component applies the thermal expansion relative to "nominal" based on the material class's thermal expansion routines. This sounds straightforward, but in my experience is difficult to generalize based on the wide variety of ways in which the thermal expansion is implemented under the hood on different material classes. It would also require the temperature associated with the input number densities to be specified, although we could force this to be equal to `Tinput` for the component.

A related ticket that keeps getting brought up and shut down is https://github.com/terrapower/armi/issues/535

PR #1745, opened 2024-06-21
Custom densities

bsculac:
## What is the change?

The ability to specify custom densities for non-custom materials is added in this PR.

Previously, only materials designated "Custom" in the blueprints file were allowed to have custom densities. This change allows perturbation of the component density as the component is generated. This is specifically done to prevent modification of the base material properties, which could have unintended downstream effects.

This PR attempts to resolve https://github.com/terrapower/armi/issues/1272. The use case is for adjustment of library materials to account for changes in density during fabrication (e.g. porosity) without having to use workarounds like smear density adjustments.

---

## Checklist


armi/reactor/blueprints/isotopicOptions.py
```diff
@@ -393,15 +393,11 @@ def apply(self, material):
...
-                    "(which implies a mass density) on `{}` with custom isotopics `{}`. "
-                    "This has no effect on this Material class; you can only "
-                    "override mass density on `Custom` "
-                    "materials. Consider switching to number fraction input. "
-                    "Continuing to use {} mass density.".format(
-                        material, self.name, material
-                    )
+                runLog.important(
+                    "A custom density or number densities has been specified for a non-custom "
+                    "material. The material density will not be updated to prevent unintentional "
```
onufer, line 398: `material object's density`
or
"The component's density has been updated to reflect the change, but the material it points to preserves the actual material density (just like for 2d expansion)"

armi/reactor/blueprints/tests/test_customIsotopics.py
```diff
@@ -216,18 +220,23 @@ def test_massFractionsAreApplied(self):
...
 
-        # density only works with a Custom material type.
-        self.assertAlmostEqual(fuel0.density(), fuel2.density())
         self.assertEqual(
             set(fuel2.p.numberDensities.keys()), set(fuel1.p.numberDensities.keys())
         )  # keys are same
 
+        self.assertAlmostEqual(19.1, fuel2.density())
+        # original material density should not be changed after setting a custom density component
+        self.assertNotAlmostEqual(fuel2.density(), fuel6.density(), places=2)
```
onufer, line 235: change the fuel2 density to something besides 19.1 so that we can confirm it actually happened. these tests seem to indicate the baseline density is 19.1, so the test hasn't confirmed a change

bsculac: The density of 19.1 is what is set in the blueprints above. Fuel blocks 0, 2, and 6 are all using the non-custom material. Fuel blocks 0 and 6 do not have a custom density set, while fuel block 2 has a custom density. Line 235 verifies that the custom density of fuel block 2 is not equal to fuel block 6, therefore the density of fuel block 2 has been changed from the baseline material density.

This test originally tested that fuel block 2 did _not_ change density because the material was not set to custom. Let me know if additional clarification is needed.

onufer: ooooOOOOoo. its really easy for my brain to skip the  word "not"

could you please assert it is equal to whatever its equal to so its more black and white what the actual density is?

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -225,8 +225,33 @@ class can then use to modify the isotopics as necessary.
...
 
+    def setCustomDensity(self, constructedComponent, blueprint):
+        """Apply a custom density to a material with custom isotopics but not a 'custom material'."""
+        if self.isotopics is not None:
+            # must nest because one can't check for attribues on a NoneType
+            if blueprint.customIsotopics[self.isotopics].density is not None:
+                mat = materials.resolveMaterialClassByName(self.material)()
+                if not isinstance(mat, materials.Custom):
+                    densityRatio = (
+                        blueprint.customIsotopics[self.isotopics].density
```
onufer, line 242: do we always have a density? i thought some custom isotopic input options maybe didnt specify density. maybe they all eventually resolve it though

keckler:
At what temperature is the entered density supposed to correspond to?

keckler:
This PR should probably update this section of the docs: 
https://terrapower.github.io/armi/user/inputs.html#custom-isotopics

bsculac:
> At what temperature is the entered density supposed to correspond to?

The density is set at component construction so it should be the cold dimension unless the blueprints was specified to be at hot dimensions.

armi/reactor/blueprints/componentBlueprint.py
bsculac: No, the custom isotopics entry can be specified without a density, which is why line 238 checks for it. But if `customIsotopics` exists then it's attributes are initialized to `None`

armi/reactor/blueprints/isotopicOptions.py
bsculac:
```suggestion
                    "material. The material object's density will not be updated to prevent unintentional "
```

keckler:
> > At what temperature is the entered density supposed to correspond to?
> 
> The density is set at component construction so it should be the cold dimension unless the blueprint was specified to be at hot dimensions.

I suppose I am asking for this to be outlined somewhere, for instance in the blueprints documentation.

john-science:
I would love a unit test that actually tests the new feature you are building here. Show me the _goal_ in your unit test, and show me the edge cases.

And example edge case might be proving that this PR fixes the bug in #1272.

As always, if that's too much of a lift I can help.

armi/reactor/blueprints/tests/test_customIsotopics.py
```diff
@@ -179,10 +192,168 @@ class TestCustomIsotopics(unittest.TestCase):
...
+
+    # >>> from armi.nucDirectory import elements, nuclideBases
+    # >>> import numpy
+    # >>> u = elements.bySymbol['U']
+    # >>> Mi = numpy.array([n.weight for n in u.getNaturalIsotopics()])
+    # >>> w_i = numpy.array([n.abundance for n in u.getNaturalIsotopics()])
+    # >>> Ni = 19.1 * w_i * 6.0221e23 / Mi
+    # array([  2.65398007e+18,   3.52549755e+20,   4.79692055e+22])
+    # >>> for n, ni in zip(u.getNaturalIsotopics(), Ni):
+    # >>>     print '        {}: {:.7e}'.format(n.name, ni) # requires 7 decimal places!
```
john-science, line 279: Is this bit an error? It looks like random Python code inside a YAML file.

armi/tests/tutorials/c5g7-blueprints.yaml
```diff
@@ -162,7 +162,6 @@ blocks:
             # has same od as fuel, so there needs to be something in the gap.
             shape: Circle
             material: Void
-            isotopics: moderator
```
john-science, line 165: Sorry, honest question, why did we need to make this change?

(Just trying to wrap my head around the PR.  I am not worried either way.)

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -225,8 +225,52 @@ class can then use to modify the isotopics as necessary.
...
+                        / constructedComponent.density()
+                    )
+                    constructedComponent.changeNDensByFactor(densityRatio)
+
+                    runLog.important(
+                        "A custom material density was specified in the custom isotopics for non-custom "
+                        "material {} in component {}. The component density has been altered to "
+                        "{}.".format(
+                            mat, constructedComponent, constructedComponent.density()
+                        )
```
john-science, line 271:
```suggestion
                        ), single=True
```

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -225,8 +225,52 @@ class can then use to modify the isotopics as necessary.
...
+                    )
```
john-science, line 272:
```suggestion
                    runLog.important(
                        "A custom material density was specified in the custom isotopics for non-custom "
                        "material {} in component {}. The component density has been altered to "
                        "{}.".format(
                            mat, constructedComponent, constructedComponent.density()
                        ),
                        single=True
                    )
```

john-science: What I am suggesting here is that we make sure this log message is only ever printed once. Imagine if this message were printed hundreds of times for a single reactor, during construction. Boring, right?  This `single=True` should make sure no logging duplication happens.

armi/tests/tutorials/c5g7-blueprints.yaml
bsculac: This PR does not allow density to be assigned to Void and throws an error if a user attempts to do so. Previously, this would weakly fail, emitting a warning that nothing was done.

The reason for throwing an error (aside from it being silly to assign properties to Void) was to avoid dividing by zero.

armi/reactor/blueprints/componentBlueprint.py
bsculac:
```suggestion
                    runLog.important(
                        "A custom material density was specified in the custom isotopics for non-custom "
                        "material {}. The component density has been altered to "
                        "{}.".format(
                            mat, constructedComponent.density()
                        ),
                        single=True
                    )
```
Good point, edited the suggestion to remove the component print to make the messages non-unique.

armi/reactor/blueprints/isotopicOptions.py
```diff
@@ -393,15 +393,11 @@ def apply(self, material):
...
-                    "materials. Consider switching to number fraction input. "
-                    "Continuing to use {} mass density.".format(
-                        material, self.name, material
-                    )
+                runLog.important(
+                    "A custom density or number densities has been specified for non-custom "
+                    "material {}. The material object's density will not be updated to prevent unintentional "
+                    "density changes across the model. Only custom materials may have a density "
+                    "specified.".format(material)
                 )
```
bsculac, line 401:
```suggestion
                runLog.important(
                    "A custom density or number densities has been specified for non-custom "
                    "material {}. The material object's density will not be updated to prevent unintentional "
                    "density changes across the model. Only custom materials may have a density "
                    "specified.".format(material),
                    single=True
                )
```
Incorporate `single=True` suggestion from other comment.

armi/reactor/blueprints/tests/test_customIsotopics.py
john-science: Score, thanks!

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -225,8 +225,51 @@ class can then use to modify the isotopics as necessary.
...
+        # set the custom density for non-custom material components after construction
+        self.setCustomDensity(constructedObject, blueprint, matMods)
+
         return constructedObject
 
+    def setCustomDensity(self, constructedComponent, blueprint, matMods):
+        """Apply a custom density to a material with custom isotopics but not a 'custom material'."""
+        if self.isotopics is not None:
+            # must nest because one can't check for attribues on a NoneType
+            if blueprint.customIsotopics[self.isotopics].density > 0:
```
john-science, line 238: These two `if` blocks make the entire rest of the method double-indented. Another option would be to do:

```python
        if self.isotopics is None:
            return

    if blueprint.customIsotopics[self.isotopics].density <= 0:
            return
```

This might make the rest of the method easier to read. But I'm not going to die on any hills today, I'm just pointing it out for the future.

john-science, line 238: If I go into the Python interpreter and do this:

```bash
>>> None > 0
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

If the `.density` is `None`, this line doesn't return `False`, it raises a `TypeError`.

bsculac: This is a good suggestion but it makes me realize that a user specifying a zero or negative density is not handled. I will be pushing something up that handles said case.

bsculac: I've altered the logic in this section, let me know what you think.

john-science (approved):

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -225,8 +225,67 @@ class can then use to modify the isotopics as necessary.
...
 
+    def setCustomDensity(self, constructedComponent, blueprint, matMods):
+        """Apply a custom density to a material with custom isotopics but not a 'custom material'."""
+        if self.isotopics is None:
+            # No custom isotopics specified
+            return
+        density = blueprint.customIsotopics[self.isotopics].density
+        if density is None:
+            # Nothing to do
+            return
```
john-science, line 242:
```suggestion
            return

        density = blueprint.customIsotopics[self.isotopics].density
        if density is None:
            # Nothing to do
            return

```

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -225,8 +225,69 @@ class can then use to modify the isotopics as necessary.
...
+                    )
+                )
+                raise ValueError(
+                    "Cannot apply custom densities to materials without density."
+                )
+
+            densityRatio = (
+                blueprint.customIsotopics[self.isotopics].density
+                / constructedComponent.density()
+            )
```
john-science, line 281:
```suggestion
            densityRatio = density / constructedComponent.density()
```

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -225,8 +225,69 @@ class can then use to modify the isotopics as necessary.
...
+        if self.isotopics is None:
+            # No custom isotopics specified
+            return
+
+        density = blueprint.customIsotopics[self.isotopics].density
+        if density is None:
+            # Nothing to do
+            return
+
+        if blueprint.customIsotopics[self.isotopics].density <= 0:
```
john-science, line 245:
```suggestion
        if density <= 0:
```

john-science: Just seems easier to read. NBD. I won't die on this hill.

armi/reactor/blueprints/componentBlueprint.py
john-science: Just seems easier to read. NBD. I won't die on this hill.

armi/reactor/blueprints/tests/test_customIsotopics.py
```diff
@@ -216,18 +332,89 @@ def test_massFractionsAreApplied(self):
...
+        # The block with custom density but not the 'Custom' material
+        fuel2 = self.a[2].getComponent(Flags.FUEL)
+        # A block like the template block, but made after the custom block
+        fuel6 = self.a[6].getComponent(Flags.FUEL)
+        # A block with custom density set via number density
+        fuel8 = self.a[8].getComponent(Flags.FUEL)
+
+        # Check that the density is set correctly on the custom density block,
+        # and that it is not the same as the original
+        self.assertAlmostEqual(19.1, fuel2.density())
```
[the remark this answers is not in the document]
bsculac: If you are asking if the custom density logic should be adjusting to the working temperature, probably. When I wrote this PR, I had misunderstood what Tin in the blueprints was used for, see https://github.com/terrapower/armi/issues/1818#issuecomment-2289766135 for more details. In short, there is an assumption of hot heights in the blueprints unless the option `inputHeightsConsideredHot` is `False`, so in the default case of `inputHeightsConsideredHot`=True the feature in this PR will assign incorrect densities for components with custom isotopics because this feature assumes the blueprints are for cold dimensions.

PR #1750, opened 2024-06-26
Deleting objects that are still trying to hold on to old reactor after reset

zachmprince:
## What is the change?

This change is a follow on to #1729 after it was observed that the garbage collector was not clearing `Reactor` data from a state distribution on worker processors after a reset. This is due to objects retaining references to the distributed `Reactor`. Upon investigation, these references came from a cyclic reference between `Core`, `Reactor`, and `spatialGrid` objects, shown by this diagram created by `objgraph`:

image

The solution was quite simple: setting `spatialGrid.armiObject` attribute to `None`, as well as the `parent` attribute in the `Reactor` children.

Retaining the distributed state of the `Reactor` on workers is unnecessary memory overhead for interactions that don't require a distributed state.

---

## Checklist

### DIFF
--- a/armi/operators/operator.py
+++ b/armi/operators/operator.py
@@ -1077,6 +1077,8 @@ class Operator:
         """
         if self.r:
             self.r.o = None
+            for comp in self.r:
+                comp.parent = None
         self.r = None
         for i in self.interfaces:
             i.o = None
--- a/armi/operators/operatorMPI.py
+++ b/armi/operators/operatorMPI.py
@@ -238,6 +238,7 @@ class OperatorMPI(Operator):
         cs = self.cs
         bp = self.r.blueprints
         spatialGrid = self.r.core.spatialGrid
+        spatialGrid.armiObject = None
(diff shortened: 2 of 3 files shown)


opotowsky (approved):
Approved from my end! Not sure this can be tested, so I'll let John S comment on that. 

Could you add this and the previous PR to the release notes? I think they are pretty important-to-mention updates!

zachmprince:
> Approved from my end! Not sure this can be tested, so I'll let John S comment on that.

Coming up with the test for this is not trivial to me. I'm not even sure what would be tested. Plus, the effectiveness of the change is largely dependent on the plugins and what interfaces are active.

> Could you add this and the previous PR to the release notes? I think they are pretty important-to-mention updates!

Added an item to the release notes. I wasn't sure whether to put it at the top or bottom of the list lol.

john-science (approved):
The actual changes themselves are simple and I like them.

You are right, these would be annoying to test, as you / we would need to build some testing tooling to test the rest of the `detach` code around it.

PR #1769, opened 2024-07-08
Reducing the number of warnings when building docs

zachmprince:
## What is the change?

This PR mainly includes minor fixes to reduce errors/warnings during the documentation build. Here are a list of things included:

- Escaping RST special characters
- Formatting pieces of documentation that should be rendered as code
- Making various type definitions more explicit to avoid warnings there are multiple possible interpretations
- Removing duplicate references
- Making some links "anonymous"
- Removing API headings that aren't allowed by numpy code standards
- Formatting multiline lists properly
- Various black/ruff formatting

A more significant change includes modifications to the `dochelpers` package. This includes:

- A new function that escapes special characters in a inputted string. This is used in the parameter and settings report table functions, which were including special characters from the descriptions.
- The ability to a label to created tables

This somewhat of a "basic" maintenance PR: making docs render properly. On a more practical level, reducing the number of warnings when building docs helps pick out future bugs written in the documentation.

---

## Checklist


armi/materials/thU.py
```diff
@@ -16,9 +16,6 @@
 Thorium Uranium metal.
 
 Data is from [IAEA-TECDOCT-1450]_.
-
-.. [IAEA-TECDOCT-1450] Thorium fuel cycle -- Potential benefits and challenges, IAEA-TECDOC-1450 (2005).
-    https://www-pub.iaea.org/mtcd/publications/pdf/te_1450_web.pdf
```
john-science, line 21: I don't understand this change.

This PDF link is still valid. And it is referenced as a data source in the line above.

We can't just delete this and lose the reference.

armi/materials/thoriumOxide.py
```diff
@@ -16,9 +16,6 @@
 Thorium Oxide solid ceramic.
 
 Data is from [IAEA-TECDOCT-1450]_.
-
-.. [IAEA-TECDOCT-1450] Thorium fuel cycle -- Potential benefits and challenges, IAEA-TECDOC-1450 (2005).
-    https://www-pub.iaea.org/mtcd/publications/pdf/te_1450_web.pdf
```
john-science, line 21: I don't understand this change.

This PDF link is still valid. And it is referenced as a data source in the line above.

We can't just delete this and lose the reference.

armi/nuclearDataIO/cccc/__init__.py
```diff
@@ -58,7 +58,7 @@
 
 .. [CCCC-IV] R. Douglas O'Dell, "Standard Interface Files and Procedures for Reactor Physics
              Codes, Version IV," LA-6941-MS, Los Alamos National Laboratory (September 1977).
-             Web. doi:10.2172/5369298. (`OSTI <https://www.osti.gov/biblio/5369298>`_)
+             Web. doi:10.2172/5369298. (`OSTI <https://www.osti.gov/biblio/5369298>`__)
```
john-science, line 61: Can you explain this one to me?

I just don't get the change, that's all.

The link is currently working on the docs:  https://terrapower.github.io/armi/.apidocs/armi.nuclearDataIO.cccc.html#module-armi.nuclearDataIO.cccc

armi/nuclearDataIO/cccc/gamiso.py
```diff
@@ -33,7 +33,7 @@
 
 .. [GAMSOR] Smith, M. A., Lee, C. H., and Hill, R. N. GAMSOR: Gamma Source Preparation and DIF3D
             Flux Solution. United States: N. p., 2016. Web. doi:10.2172/1343095. `On OSTI
-            <https://www.osti.gov/biblio/1343095-gamsor-gamma-source-preparation-dif3d-flux-solution>`_
+            <https://www.osti.gov/biblio/1343095-gamsor-gamma-source-preparation-dif3d-flux-solution>`__
```
john-science, line 36: ditto above

armi/nuclearDataIO/cccc/pmatrx.py
```diff
@@ -20,7 +20,7 @@
 .. [MC23] Lee, Changho, Jung, Yeon Sang, and Yang, Won Sik. MC2-3: Multigroup Cross Section
           Generation Code for Fast Reactor Analysis Nuclear. United States: N. p., 2018. Web.
           doi:10.2172/1483949. (`OSTI
-          <https://www.osti.gov/biblio/1483949-mc2-multigroup-cross-section-generation-code-fast-reactor-analysis-nuclear>`_)
+          <https://www.osti.gov/biblio/1483949-mc2-multigroup-cross-section-generation-code-fast-reactor-analysis-nuclear>`__)
```
john-science, line 23: ditto above

armi/plugins.py
```diff
@@ -309,10 +309,12 @@ def defineBlockTypes() -> List:
         corresponding ``Component`` type that should activate it. For instance a
         ``HexBlock`` would be created when the largest component is a ``Hexagon``::
 
+            [(Hexagon, HexBlock)]
+
         Returns
         -------
         list
-            [(Hexagon, HexBlock)]
+            ``[(compType, BlockType), ...]``
```
john-science, line 317: I understand this fixed the RST, but is it still correct?

armi/materials/thU.py
zachmprince: I think I might revert this change. Sphinx was throwing a duplicate reference warning due to this reference already being defined in thorium.py. However, this is endemic to many plugins so it is impossible to avoid all the warnings. So this change isn't improving things much.

My hope in the future is to switch to using the sphinx-bibtex extension, so all references are defined in a single place.

armi/nuclearDataIO/cccc/__init__.py
zachmprince: This is switching the link to be "anonymous", so the `OSTI` label isn't an explicit target. Without this, `OSTI` is created as a reference target twice, throwing a `WARNING: Duplicate explicit target name`. See this for a detailed explanation: https://stackoverflow.com/questions/74252545/differences-between-anonymous-and-non-anonymous-hyperlinks-in-sphinx

armi/plugins.py
zachmprince: According to the docstring above this, it is correct. But I can switch to what was originally there if I'm mistaken.

armi/nuclearDataIO/cccc/gamiso.py
zachmprince: This one isn't necessary since there is only one `On OSTI`, but I figured it would be good practice.

PR #1771, opened 2024-07-10
Fixing a bug where getReactionRates didn't work on Core

### LINKED ISSUE #1754: `composite.getReactionRates()` doesn't work on the `core` level
Something appears to be wrong with this method:
https://github.com/terrapower/armi/blob/9d6e2b0097ac2e203c7726fdaa44a1a854648938/armi/reactor/composites.py#L3060-L3092

I recently loaded up a database and tried to examine some reaction rates as follows:
```python
>  r.core.getReactionRates("U235")
Loading microscopic cross section library `ISOTXS`
 {'nG': 0, 'nF': 0, 'n2n': 0, 'nA': 0, 'nP': 0, 'n3n': 0}

> r.core.getAssemblyWithStringLocation('002-001').getReactionRates("U235")
{'nG': 4.219163274625927e+16,
 'nF': 1.5933660772471552e+17,
 'n2n': 86100493155576.52,
 'nA': 0.0,
 'nP': 0.0,
 'n3n': 0}

>  r.core.getAssemblyWithStringLocation('002-001').getReactionRates("U238")
{'nG': 1.3747239159395538e+17,
 'nF': 2.0985393232144132e+16,
 'n2n': 706063666454843.1,
 'nA': 0.0,
 'nP': 0.0,
 'n3n': 0}
```

It doesn't make sense that the returned reaction rates for the core as a whole are zero, but non-zero values are returned for individual assemblies from the same core.

john-science:
## What is the change?

Fixing a bug where `getReactionRates()` didn't work on a `Core`.  Also, I added unit tests to ensure it is working at various other levels: `Block`, `Assembly`, `Core`, and `Reactor`.

To close #1754, which was found by @keckler.

---

## Checklist

### DIFF
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -3031,6 +3031,7 @@ class Composite(ArmiObject):
         -----
         If you set nDensity to 1/CM2_PER_BARN this makes 1 group cross section generation easier
         """
+        from armi.reactor.assemblies import Assembly
         from armi.reactor.blocks import Block
         from armi.reactor.reactors import Core
 
@@ -3038,18 +3039,16 @@ class Composite(ArmiObject):
             nDensity = self.getNumberDensity(nucName)
 
         try:
-            core = self.getAncestor(lambda c: isinstance(c, Core))
-
-            try:
+            if isinstance(self, Assembly):
                 block = self.getChildren(
                     deep=True, predicate=lambda o: isinstance(o, Block)
                 )[0]
-            except Exception:
+            else:
                 block = self.getAncestor(lambda x: isinstance(x, Block))
 
             return getReactionRateDict(
                 nucName,
-                core.lib,
+                self.getAncestor(lambda c: isinstance(c, Core)).lib,
                 block.getMicroSuffix(),
                 self.getIntegratedMgFlux(),
                 nDensity,
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -2997,7 +2997,6 @@ class Composite(ArmiObject):
         ----------
         adjoint : bool, optional
             Return adjoint flux instead of real
-
         gamma : bool, optional
             Whether to return the neutron flux or the gamma flux.
 
@@ -3007,11 +3006,11 @@ class Composite(ArmiObject):
             multigroup neutron tracklength in [n-cm/s]
         """
         integratedMgFlux = numpy.zeros(1)
-
         for c in self:
-            integratedMgFlux = integratedMgFlux + c.getIntegratedMgFlux(
-                adjoint=adjoint, gamma=gamma
-            )
+            mgFlux = c.getIntegratedMgFlux(adjoint=adjoint, gamma=gamma)
+            if mgFlux is not None:
+                integratedMgFlux = integratedMgFlux + mgFlux
+
         return integratedMgFlux
 
     def _getReactionRates(self, nucName, nDensity=None):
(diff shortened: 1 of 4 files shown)


armi/reactor/composites.py
```diff
@@ -3032,15 +3031,25 @@ def _getReactionRates(self, nucName, nDensity=None):
...
+                )[0]
+            else:
+                block = self.getAncestor(lambda x: isinstance(x, Block))
+
             return getReactionRateDict(
                 nucName,
-                self.getAncestorWithFlags(Flags.CORE).lib,
-                self.getAncestor(lambda x: isinstance(x, Block)).getMicroSuffix(),
+                self.getAncestor(lambda c: isinstance(c, Core)).lib,
+                block.getMicroSuffix(),
```
keckler, line 3052: I think that there is a problem here. If the object passed into `_getReactionRates()` is an `Assembly` or `Core`, then the `block` used here will just be some random block, and the micro suffix will be blanket applied to all of the reaction rate calculations. This is not correct, because in reality each block in a core can have a different suffix (and thus different XS set).

Let me know if you disagree or don't follow.

keckler (changes requested):

armi/reactor/composites.py
john-science: Ah, maybe you're right.  Okay, I got a little fast-and-loose.

But we don't define micro-suffixes for the Core or Reactor level.

keckler: > But we don't define micro-suffixes for the Core or Reactor level

Agreed, but when looping over the blocks in the core, the reaction rate calculation for each block should use the suffix for that specific block.

I was trying to think of how to implement it correctly based on your changes, and I anticipate you'll have to do a little refactoring of the method calls.

john-science: Okay, okay, my solution is bonk.

Should the reaction rates in the Core be the sum of the Reaction rates of all the Blocks in the Core?  If that's the case, I guess I can figure something out.

I suppose I could also just have `Core.getReactionRates()` throw an error, but I'd like to think we're not there yet.

keckler: Yes, the reaction rates returned by `getReactionRates()` are said to be volume integrated, so just summing them all up is what you're looking to accomplish. 

I don't think its necessary to throw an error on a Core object. This seems to be doable, just with a little reformulation. 

In particular, I think that if the call to `getChildren()` [here](https://github.com/terrapower/armi/blob/36a5bdfde7c4ef32d079225b1938e5db86618276/armi/reactor/composites.py#L3086) were replaced with `getChildren(deep=True)`, then we could guarantee that the objects passed into `_getReactionRates()` are always `Component`s, which always are associated with a given micro suffix of their parent blocks.

Do you think that would work?

john-science: Okay, take a look now.  Tell me what you think.

Sorry for the futz, perhaps I have too many balls in the air today.

armi/reactor/composites.py
```diff
@@ -3057,9 +3063,37 @@ def _getReactionRates(self, nucName, nDensity=None):
...
+
+        # Try to just get the children.
+        if not len(objects):
+            objects in self.getChildren()
+
+        # Not all composite objects are iterable (i.e. components)
+        if not len(objects):
+            objects = [self]
+
+        return objects
```
keckler, line 3092: I believe we can replace all of this logic with this suggestion, which (IMO) looks simpler.
```suggestion
        from armi.reactor.components import Component

        return self.getChildren(
            deep=True, predicate=lambda x: isinstance(x, Component)
        )
```

keckler: I tried that on my local machine, and all your tests pass, and I believe it is still accomplishing the same thing.

armi/reactor/composites.py
keckler: Yep, I think this works. I made another suggestion to simplify, but in general I think this is looking correct now.

https://github.com/terrapower/armi/pull/1771#discussion_r1674190476

armi/reactor/composites.py
keckler: Let me know what you think here. After we decide, I can approve the PR.

john-science: I will try it!

That doesn't cover the situation where there are no children, right?

So, the final `objects = [self]` case is missing?

PR #1774, opened 2024-07-11
Forcing Parameters and Settings to have Descriptions

### LINKED ISSUE #1348: Force Parameters to have descriptions

### LINKED ISSUE #1734: Force Settings to have descriptions
I have [often](https://github.com/terrapower/armi/issues/1727) had to field questions about settings, because they did not have proper descriptions. This is a sign that settings without descriptions are problematic for our user base.

I propose we add an assertion to the settings constructor, to demand that a setting has a valid description, as is being done with [parameters](https://github.com/terrapower/armi/issues/1348).

This is an API-breaking change for our downstream repos, so we will have to take that into account with our timeline.

The following ARMI settings do not currently have descriptions:

* [x] [debug](https://github.com/terrapower/armi/blob/3b822bc34c88706f629a7d6b4f53b96aca64d30c/armi/settings/fwSettings/globalSettings.py#L449-L451) - Is this unused
* [x] [burnupPeakingFactor](https://github.com/terrapower/armi/blob/3b822bc34c88706f629a7d6b4f53b96aca64d30c/armi/settings/fwSettings/globalSettings.py#L423-L429)
* [x] [freshFeedType](https://github.com/terrapower/armi/blob/3b822bc34c88706f629a7d6b4f53b96aca64d30c/armi/settings/fwSettings/globalSettings.py#L523-L529)
* [x] [jumpRingNum](https://github.com/terrapower/armi/blob/3b822bc34c88706f629a7d6b4f53b96aca64d30c/armi/physics/fuelCycle/settings.py#L103-L105) 
* [x] [levelsPerCascade](https://github.com/terrapower/armi/blob/3b822bc34c88706f629a7d6b4f53b96aca64d30c/armi/physics/fuelCycle/settings.py#L106-L112)
* [x] [removePerCycle](https://github.com/terrapower/armi/blob/3b822bc34c88706f629a7d6b4f53b96aca64d30c/armi/settings/fwSettings/globalSettings.py#L665-L667)

> Update: All settings in ARMI now have descriptions.

john-science:
## What is the change?

This PR removes to `DeprecationWarning`s in ARMI, that allow for empty description fields in a `Setting` or `Parameter`.

A strong point of ARMI is that users / developers can create new settings to configure a run and parameters to model a reactor quite easily. However, people also tend to abandon settings and parameters silently, and we never know. OR a setting/parameter will be used in different ways by different people, thus leading to problems.

In the past, I have seen these problems only occur when the setting/parameter doesn't have a description, so we have had a long push to enforce this field.

This will close #1348 and close #1734

---

## Checklist

### DIFF
--- a/armi/reactor/parameters/parameterDefinitions.py
+++ b/armi/reactor/parameters/parameterDefinitions.py
@@ -32,7 +32,6 @@ import re
 
 import numpy
 
-from armi import runLog
 from armi.reactor.flags import Flags
 from armi.reactor.parameters.exceptions import ParameterError, ParameterDefinitionError
 
@@ -265,16 +264,12 @@ class Parameter:
         categories,
         serializer: Optional[Type[Serializer]] = None,
     ):
-        assert self._validName.match(name), "{} is not a valid param name".format(name)
         # nonsensical to have a serializer with no intention of saving to DB
         assert not (serializer is not None and not saveToDB)
         assert serializer is None or saveToDB
-        # TODO: This warning is temporary. At some point, it will become an AssertionError.
-        if not len(description):
-            runLog.warning(
-                f"DeprecationWarning: Parameter {name} defined without description.",
-                single=True,
-            )
+        assert self._validName.match(name), "{} is not a valid param name".format(name)
+        assert len(description), f"Parameter {name} defined without description."
+
         self.collectionType = _Undefined
         self.name = name
         self.fieldName = "_p_" + name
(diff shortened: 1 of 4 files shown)


opotowsky (approved):
I think we need to wait on one downstream pre-requisite to be merged first, but approved for when the time is right!

PR #1785, opened 2024-07-18
Moving rxSwing core parameter to final time node

onufer:
## What is the change?

After PR #1090, we don't write data calculated during EOC to database. RX swing is the main parameter to suffer as a consequence, so this PR moves its calculation to the last node rather than the EOC interaction.

Also adds a minor `runlog` statement, and bumped the example app submodule.

So that rx swing is stored in the database (it is zeroed out in interact BOL since its last cycle.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -122,7 +122,7 @@ class GlobalFluxInterface(interfaces.Interface):
             eocKeff = self.r.core.p.keffUnc or self.r.core.p.keff
             swing = (eocKeff - self._bocKeff) / (eocKeff * self._bocKeff)
             self.r.core.p.rxSwing = swing * units.ABS_REACTIVITY_TO_PCM
-            runLog.important(
+            runLog.info(
                 f"BOC Uncontrolled keff: {self._bocKeff},  "
                 f"EOC Uncontrolled keff: {self.r.core.p.keffUnc}, "
                 f"Cycle Reactivity Swing: {self.r.core.p.rxSwing} pcm"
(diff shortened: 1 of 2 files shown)


opotowsky (commented):
you need to run black or rufff (linting failed) but you're good to run internal testing on this!

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -102,23 +102,30 @@ def interactEveryNode(self, cycle, node):
...
+            self._bocKeff = self.r.core.p.keffUnc or self.r.core.p.keff
+
+        # A 1 burnstep cycle would have 2 nodes, and the last node would be node index 1 (first is zero)
+        lastNodeInCycle = getBurnSteps(self.cs)[self.r.p.cycle]
+        if self.r.p.timeNode == lastNodeInCycle and self._bocKeff is not None:
+
+            eocKeff = self.r.core.p.keffUnc or self.r.core.p.keff
+            swing = (eocKeff - self._bocKeff) / (eocKeff * self._bocKeff)
+            self.r.core.p.rxSwing = swing * units.ABS_REACTIVITY_TO_PCM
+            runLog.important(
```
opotowsky, line 125: Does this need to be `important` or could it be something like `extra` that you toggle the verbosity to when needed? (I don't have a sense of what analysts want to see)

john-science: I agree. I'm not thrilled that every PR people add more and more log statements.

But `important` is usually reserved for things that may-or-may-not be bugs, and people better pay attention to.  If we are keeping this log statement, I would say `extra`.

onufer: info is OK. its defiantly more important than `extra`.

opotowsky: offline, decided to go with info!

opotowsky (approved):
nice work!

PR #1788, opened 2024-07-19
Fix MultipleComponentMerger with DerivedShape solvent

AidanMcDonald:
## What is the change?

Simple bug fix for `MultipleComponentMerger`.

Add a check to avoid attempting to run `_verifyExpansion` with a `DerivedShape` solvent. `ComponentMerger` bypasses this check (which doesn't make sense for a `DerivedShape` solvent) here:

https://github.com/terrapower/armi/blob/ed35e229b2b2f5fef1a4b5dd0801d811cae31e31/armi/reactor/converters/blockConverters.py#L97-L98

`MultipleComponentMerger` waits to verify expansion until all components are merged, but the direct call to `_verifyExpansion` avoids the `DerivedShape` component check. I manually added the check to the `_verifyExpansion` call in `MultipleComponentMerger`.

Currently, `MultipleComponentMerger` fails when solvent is a `DerivedShape` `Component`. Reproducible example of the bug:

```python
import armi
armi.configure()

# Using the FFTF case
o = armi.init(fName='FFTF.yaml')
# Get a block with coolant and wire
b = o.r.core.getBlocks()[200]
# ComponentMerger works fine
newb = armi.reactor.converters.blockConverters.ComponentMerger(b, "wire", "coolant").convert()
# MultipleComponentMerger breaks
newb = armi.reactor.converters.blockConverters.MultipleComponentMerger(b, ["wire"], "coolant").convert()
```

---

## Checklist

### DIFF
--- a/armi/reactor/converters/blockConverters.py
+++ b/armi/reactor/converters/blockConverters.py
@@ -311,7 +311,7 @@ class MultipleComponentMerger(BlockConverter):
                 soluteName, self.solventName, minID=self.specifiedMinID
             )
         solvent = self._sourceBlock.getComponentByName(self.solventName)
-        if not solvent.__class__ is components.DerivedShape:
+        if solvent.__class__ is not components.DerivedShape:
             BlockConverter._verifyExpansion(self, self.soluteNames, solvent)
         return self._sourceBlock
 
--- a/armi/reactor/converters/blockConverters.py
+++ b/armi/reactor/converters/blockConverters.py
@@ -311,7 +311,8 @@ class MultipleComponentMerger(BlockConverter):
                 soluteName, self.solventName, minID=self.specifiedMinID
             )
         solvent = self._sourceBlock.getComponentByName(self.solventName)
-        BlockConverter._verifyExpansion(self, self.soluteNames, solvent)
+        if not solvent.__class__ is components.DerivedShape:
+            BlockConverter._verifyExpansion(self, self.soluteNames, solvent)
         return self._sourceBlock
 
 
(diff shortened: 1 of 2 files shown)


AidanMcDonald:
@john-science , This is equivalent to the bug fix that I submitted before and then said was unnecessary. Due to some poor git management on my part, I still had the fix when I said it was unnecessary, sorry for the confusion.

armi/reactor/converters/blockConverters.py
```diff
@@ -311,7 +311,8 @@ def convert(self):
                 soluteName, self.solventName, minID=self.specifiedMinID
             )
         solvent = self._sourceBlock.getComponentByName(self.solventName)
-        BlockConverter._verifyExpansion(self, self.soluteNames, solvent)
+        if solvent.__class__ is not components.DerivedShape:
```
jakehader, line 314: Can we add a test case for the scenario you ran into and make sure that the else condition is checked or tested? For example, if the solvent is not a derived shape should the block converter even have defined behavior? I'm not sure.

AidanMcDonald: Test added. `MultipleComponentMerger` had no tests. I test dissolving two components into a derived shape solvent (the case that was broken) and dissolving two components into a circle solvent (the only non-derived shape solvent with defined behavior). The tests are similar to existing tests `test_dissolveWireIntoCoolant` and `test_dissolveLinerIntoClad` for `ComponentMerger`.

armi/reactor/converters/tests/test_blockConverter.py
```diff
@@ -81,6 +81,22 @@ def _test_dissolve(self, block, soluteName, solventName):
         self.assertNotIn(soluteName, convertedBlock.getComponentNames())
         self._checkAreaAndComposition(block, convertedBlock)
 
+    def test_dissolveMultiple(self):
+        """Test dissolving multiple components into another."""
```
jakehader, line 85: @john-science this test may need a requirement and implementation tag.

john-science (approved):

john-science:
I was able to show this doesn't break any downstream unit tests.

PR #1790, opened 2024-07-19
Add virt memory attribute to memory profiler

opotowsky:
## What is the change?

Add `processVirtualMemoryInMB` attribute to `SystemAndProcessMemoryUsage`, but do nothing else (not adding or changing what prints out with ARMI's memory profiler).

This isn't able to be tested at all via GH actions. I do not think this warrants an entry in the release notes but let me know if I'm wrong.

I want to measure virtual memory usage for a downstream application, since there are sometimes large discrepancies between physical and virtual and I've had some OOM errors as a result. this enables me to take a more conservative approach to memory usage.

---

## Checklist

- [x] ~[Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify any new/changed code.~

### DIFF
--- a/armi/bookkeeping/memoryProfiler.py
+++ b/armi/bookkeeping/memoryProfiler.py
@@ -340,14 +340,19 @@ class SystemAndProcessMemoryUsage:
         # directly by the standard operator and reports, so easier said than done.
         self.percentNodeRamUsed: Optional[float] = None
         self.processMemoryInMB: Optional[float] = None
+        self.processVirtualMemoryInMB: Optional[float] = None
         if _havePsutil:
             self.percentNodeRamUsed = psutil.virtual_memory().percent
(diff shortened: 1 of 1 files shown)


john-science:
@opotowsky I would leave this out of the release notes.  It is not API breaking, and is a very small new feature.

opotowsky:
> @opotowsky Is this something we need to add to ARMI? It _sounds_ like you just want this information for debugging? If so, you could make this some one-line call to `psutil.Process().memory_info().vms` in your downstream project.
> 
> Or are you writing code that will need to use this regularly?
> 
> Just trying to understand the situation.

For sure. Right now it's just me, but I could see extending the memory Profiler to use this info. Also, the broadcast/gather is really useful and is why I'm using the memory profiler instead of writing my own code (which would just duplciate what's here)

john-science (approved):

PR #1791, opened 2024-07-22
Adding a couple little unit tests for context.py

john-science:
## What is the change?

* Adding a couple little unit tests for context.py
* Also, I went through and cleaned a lot of rambling out of the `context.py` comments and docstrings.

GitHub will run all the unit tests in ARMI in series, and in parallel under MPI.  We also have GitHub run all the unit tests under Mac, Windows, and Linux.

So the tests here really aren't much, just a starting place. But they WILL be run in a full 2x3 grid of variations, which is useful for something as critical as `context.py`.

---

## Checklist

### DIFF
--- a/armi/context.py
+++ b/armi/context.py
@@ -15,7 +15,8 @@
 """
 Module containing global constants that reflect the executing context of ARMI.
 
-Things like the MPI environment, executing user, etc. live here.
+ARMI's global state information: operating system information, environment data, user data, memory
+parallelism, temporary storage locations, and if operational mode (interactive, gui, or batch).
 """
 from logging import DEBUG
 import datetime
--- a/armi/context.py
+++ b/armi/context.py
@@ -79,7 +79,7 @@ USER = getpass.getuser()
 START_TIME = time.ctime()
 
 # Set batch mode if not a TTY, which means you're on a cluster writing to a stdout file. In this
-# mode you cannot respond to prompts. (This does not work replabliy for both Windows and Linux so an
+# mode you cannot respond to prompts. (This does not work reliably for both Windows and Linux so an
 # os-specific solution is applied.)
 isatty = sys.stdout.isatty() if "win" in sys.platform else sys.stdin.isatty()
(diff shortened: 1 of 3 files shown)


armi/context.py
```diff
@@ -93,33 +78,30 @@ def setMode(cls, mode):
 USER = getpass.getuser()
 START_TIME = time.ctime()
 
-# Set batch mode if not a TTY, which means you're on a cluster writing to a stdout file
-# In this mode you cannot respond to prompts or anything
-# This does not work replabliy for both windows and linux so an os-specific solution is
-# applied
+# Set batch mode if not a TTY, which means you're on a cluster writing to a stdout file. In this
+# mode you cannot respond to prompts. (This does not work replabliy for both Windows and Linux so an
```
opotowsky, line 82:
```suggestion
# mode you cannot respond to prompts. (This does not work reliably for both Windows and Linux so an
```
catching my own typo!

armi/tests/test_context.py
```diff
@@ -0,0 +1,39 @@
...
+
+from armi import context
+
+
+class TestContext(unittest.TestCase):
+    def setUp(self):
+        pass
+
+    def test_mpiSizeAndRank(self):
+        if context.MPI_SIZE > 1:
```
opotowsky, line 26: this would need to be added to the mpi unit tests to be hit, right?

armi/context.py
```diff
@@ -15,10 +15,7 @@
 """
 Module containing global constants that reflect the executing context of ARMI.
 
-This contains information about the circumstatces under which an ARMI application is
-running. Things like the MPI environment, executing user, etc. live here. These are
-re-exported by the `armi` package, but live here so that import loops won't lead to
-as many issues.
+Things like the MPI environment, executing user, etc. live here.
```
jakehader, line 18: Better name than `things`? Maybe "Implements global state attributes that are dependent on the environment, including serial and distributed memory parallelism, Operating System flags, modes of running the framework (e.g., interactive, batch), temporary storage of application files, etc."?

john-science: Done!

PR #1792, opened 2024-07-22
Add OpenMC label to ThermalScattering

AidanMcDonald:
## What is the change?

Added support for thermal scattering data labels in OpenMC's syntax (usually something like `"c_H_in_H2O"`) analogous to current support for ENDFB8 and ACE labels. Also added tests similar to the existing ENDFB8 and ACE label tests. The label generator does not currently work for all thermal scattering data available to OpenMC (for example `"c_O_in_H2O_solid"` or `"c_ortho_H"`), but it works for all thermal scattering laws used in ARMI in `thermalScattering.byNbAndCompound`.

An OpenMC label generator is necessary to use thermal scattering data in the ARMI [OpenMC plugin](https://github.com/terrapower/armicontrib-openmc), and it is simpler to include it in ARMI's thermal scattering module rather than in the plugin itself.

---

## Checklist

### DIFF
--- a/armi/nucDirectory/tests/test_thermalScattering.py
+++ b/armi/nucDirectory/tests/test_thermalScattering.py
@@ -92,6 +92,22 @@ class TestThermalScattering(unittest.TestCase):
         fe56tsl = ts.ThermalScattering(fe56)
         self.assertEqual(fe56tsl._genACELabel(), "fe-56")
 
+    def test_OpenMCCompound(self):
+        si = nb.byName["SI"]
+        o = nb.byName["O"]
+        sio2 = ts.ThermalScattering((si, o), "SiO2_alpha")
+        self.assertEqual(sio2._genOpenMCLabel(), "c_SiO2_alpha")
+
+    def test_OpenMCElementInCompound(self):
+        hyd = nb.byName["H"]
+        hInH2O = ts.ThermalScattering(hyd, "H2O")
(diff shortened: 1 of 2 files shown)


jakehader:
Hi @AidanMcDonald - would it be possible to implement this in the OpenMC plugin directly or can we think about ways to abstract the thermal scattering laws implementation so that specific code-based names and attributes aren't added to the code base? I think this would be a more robust way of handling the framework for "n" codes and so that if OpenMC changes the naming it doesn't impact the framework design.

jakehader (changes requested):
See general comment

AidanMcDonald:
@jakehader , thanks for the feedback. Here's my reasoning for implementing it this way:

My intention is to implement the OpenMC label generator in the same way as the current ENDFB8 and ACE label generators since the generation and use is almost identical. I agree that adding specific names isn't ideal, but just like with the ENDF and ACE label generators, some of the labels don't follow logical rules and must be hard-coded. Since that is the case, I think it makes sense to keep the hard-coded ENDF, ACE, and OpenMC labels together like the hard-coded ENDF and ACE labels are now.

With that said, I have made a function in the OpenMC plugin that will work for now (featuring a lot of if statements for the "hard-coded" labels) if that's the direction you want to go.

AidanMcDonald:
Moving discussion to #1804 and closing for now. I will keep the branch in my fork just in case.

PR #1806, opened 2024-08-01
Don't make a runLog named -mpi.log

mgjarrett:
## What is the change?

Ensure that the `caseTitle` used to named the concatenated worker log file is not an empty string.

In a parallel run, each worker writes logging output to a separate file in the `logs/` directory. At the end of the run, these files are all concatenated into a single file. The logic for picking the name of the file is to deduce the `caseTitle` from the name of one of these log files, e.g. `ARMI.caseTitle.0001.stdout`. However, sometimes a file with the name `ARMI..0001.stdout` may exist in the `logs/` directory in addition to the `ARMI.caseTitle.0001.stdout` file, which results in a concatenated file named `-mpi.log`.

Instead of grabbing the first string we find between the first and second period characters in any log file name, we take the first string that isn't simply `""`. Then the name of the log file will be `<something>-mpi.log`.

---

## Checklist

### DIFF
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -44,16 +44,15 @@ Or change the log level the same way:
 
     runLog.setVerbosity('debug')
 """
-from glob import glob
 import collections
 import logging
 import operator
 import os
 import sys
 import time
+from glob import glob
 
 from armi import context
-
 
 # global constants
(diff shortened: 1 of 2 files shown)


john-science (approved):
Thanks for the test!

PR #1815, opened 2024-08-08
Adding two-letter XS codes to the input docs

### LINKED ISSUE #1709: Document the possibility of 2-letter XS types
There is an undocumented feature in ARMI:

https://github.com/terrapower/armi/blob/cfc23356334253cb714171349f51acceb2db0225/armi/reactor/blocks.py#L406-L411

It would be good to write this in the user documentation, maybe in [this section about assembly blueprints](https://terrapower.github.io/armi/user/inputs.html#assemblies).

Other related docstrings:
https://github.com/terrapower/armi/blob/cfc23356334253cb714171349f51acceb2db0225/armi/physics/neutronics/crossSectionGroupManager.py#L1009-L1013
https://github.com/terrapower/armi/blob/cfc23356334253cb714171349f51acceb2db0225/armi/physics/neutronics/crossSectionGroupManager.py#L86-L104

john-science:
> DO NOT MERGE YET! LET JOHN MERGE THIS ONE!

## What is the change?

Adding to the docs an explanation that we can have two-letter XS group names.

Better documentation is always better.

Also, to close #1709 

---

## Checklist

### DIFF
--- a/doc/user/inputs.rst
+++ b/doc/user/inputs.rst
@@ -787,9 +787,8 @@ xs types
   best to set blocks that have much different compositions to have separate cross section types. The
   tradeoff is that the more XS types you define, the more CPU time the case will take to run.
 
-  Representing xsType by a single letter limits users to 26 groups. So ARMI will allow 2-letter
-  xsType designations if and only if the ``buGroups`` setting has length 1 (i.e. no burnup groups
-  are defined). This is useful for high-fidelity XS modeling.
+  Representing xsType by a single letter (A-Z) or number (0-9) limits users to 36 groups. So ARMI
+  will allow 2-letter xsType designations if and only if the ``buGroups`` setting has length 1 (i.e. no burnup groups are defined). This is useful for high-fidelity XS modeling.
 
 axial mesh points
   Blocks will be broken up into this many uniform mesh points in the deterministic neutronics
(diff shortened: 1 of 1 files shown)


doc/user/inputs.rst
```diff
@@ -779,24 +779,26 @@ specifier
...
-  the more XS types you define, the more CPU time the case will take to run.
+  The **cross-section type** is usually a single capital letter that identifies which cross section
+  (XS) set will be applied to the block. Each cross section set must be defined for at least one
+  block with fissile fuel. When the lattice physics code executes in ARMI, it determines the
+  representative blocks from each cross section type and burnup group and runs it to create the
+  cross section set for all blocks of the same type and in the same burnup group. Generally, it is
+  best to set blocks that have much different compositions to have separate cross section types. The
+  tradeoff is that the more XS types you define, the more CPU time the case will take to run.
+
+  Representing xsType by a single letter limits users to 26 groups. So ARMI will allow 2-letter
```
keckler, line 790: You may or may not know this, but I am 99% certain that we allow for xsType to be numbers as well as letters. So with a-z and then 0-9, there are actually 36 possibilities.

We might want to update this type of language here as well as in the docstrings that I pointed out in the original issue (#1709 )

keckler (approved):
The new language looks good. Thanks for that! I put one comment that could be addressed, but it is not strictly necessary. I think it would be helpful for users, though, so I would advocate that we address it.

doc/user/inputs.rst
john-science: Done!

PR #1819, opened 2024-08-13
pu frac storage update

onufer:
## What is the change?

Update materials to defer to the block's plutonium fraction 

material.puFrac is not stored in the database and materials are supposed to be stateless:
https://github.com/terrapower/armi/pull/1062

---

## Checklist

### DIFF
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -25,6 +25,7 @@ import numpy
 from armi import runLog
 from armi.nucDirectory import nuclideBases
 from armi.reactor.flags import TypeSpec
+from armi.reactor import blocks
 from armi.utils import densityTools
 from armi.utils.units import getTk, getTc
 
@@ -155,7 +156,11 @@ class Material:
         """Return the grandparent (typically block) associated with this material"""
         # would be nice to use composite.getAncestor, but material is not a composite
         component = self.parent
-        return None if component is None else component.parent
+        return (
+            None
+            if component is None
+            else component.getAncestor(lambda x: isinstance(x, blocks.Block))
+        )
 
     def getChildrenWithFlags(self, typeSpec: TypeSpec, exactMatch=True):
         """Return empty list, representing that this object has no children."""
@@ -874,7 +879,7 @@ class FuelMaterial(Material):
         return self_puFrac
 
     @puFrac.setter
-    def x(self, puFrac):
+    def puFrac(self, puFrac):
         runLog.warning(
             "Materials should not store state, so this component will look at its (grand)parent for Pu fraction unless it us detached. "
             "Storing properties on the material objected will be deprecated in the near future.",
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -151,6 +151,12 @@ class Material:
         """Return empty list, representing that materials have no children."""
         return []
 
+    def getGrandparent(self):
+        """Return the grandparent (typically block) associated with this material"""
+        # would be nice to use composite.getAncestor, but material is not a composite
+        component = self.parent
+        return None if component is None else component.parent
+
     def getChildrenWithFlags(self, typeSpec: TypeSpec, exactMatch=True):
         """Return empty list, representing that this object has no children."""
         return []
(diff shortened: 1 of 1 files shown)


armi/materials/material.py
```diff
@@ -151,6 +151,12 @@ def getChildren(
         """Return empty list, representing that materials have no children."""
         return []
 
+    def getGrandparent(self):
+        """Return the grandparent (typically block) associated with this material"""
+        # would be nice to use composite.getAncestor, but material is not a composite
+        component = self.parent
+        return None if component is None else component.parent
```
jakehader, line 158: I think you can replace this with `component.getAncestor(lambda x: isinstance(x, Block))`

armi/materials/material.py
```diff
@@ -845,9 +851,36 @@ class FuelMaterial(Material):
...
+            return myBlock.p.puFrac
+        runLog.warning(
+            "{self} has no block attached, deferring to internally stored Pu Frac. "
+            "Storing properties on the material objected will be deprecated in the near future.",
+            single=true,
+        )
+        return self_puFrac
+
+    @puFrac.setter
+    def x(self, puFrac):
```
jakehader, line 877: This can be `def puFrac(self, value)`

armi/materials/material.py
```diff
@@ -845,9 +851,36 @@ class FuelMaterial(Material):
...
+        runLog.warning(
+            "Materials should not store state, so this component will look at its (grand)parent for Pu fraction unless it us detached. "
+            "Storing properties on the material objected will be deprecated in the near future.",
+            single=true,
+        )
+        self._puFrac = puFrac
```
jakehader, line 883: When we set the `puFrac` directly, should this assume a reduction in `uFrac` so that the total fraction is 1.0? I wonder if a 1.0 check should be added.

armi/materials/material.py
```diff
@@ -151,6 +151,12 @@ def getChildren(
         """Return empty list, representing that materials have no children."""
         return []
 
+    def getGrandparent(self):
```
jakehader, line 154: maybe something like, `def getAssociatedBlock`?

armi/materials/material.py
```diff
@@ -845,9 +851,36 @@ class FuelMaterial(Material):
     class1_wt_frac = None
     class1_custom_isotopics = None
     class2_custom_isotopics = None
-    puFrac = 0.0
-    uFrac = 0.0
-    zrFrac = 0.0
+    puFracDefault = 0.0
+    uFracDefault = 0.0
+    zrFracDefault = 0.0
```
jakehader, line 856: Adding plutonium, uranium, and zirconium to a generic `FuelMaterial` seems odd. We should discuss.

jakehader: Probably outside of the scope of this change, just a comment. Similar comment for @john-science in the `duplicate` method.

armi/materials/material.py
```diff
@@ -845,9 +856,36 @@ class FuelMaterial(Material):
...
+
+    def __init__(self):
+        self._puFrac = 0
```
jakehader, line 864: We will still need attributes for `uFrac` and `zrFrac` -- see `duplicate` method, so we can have `setters` and `getters` for accessing to keep `self._uFrac`, `self._zrFrac`, and `self._puFrac` private.

onufer: code works as is, so we dont need them immediately without clear expectations on behavior of u and zr frac with depletion

armi/materials/material.py
```diff
@@ -845,9 +856,36 @@ class FuelMaterial(Material):
     class1_wt_frac = None
     class1_custom_isotopics = None
     class2_custom_isotopics = None
-    puFrac = 0.0
-    uFrac = 0.0
-    zrFrac = 0.0
+    puFracDefault = 0.0
+    uFracDefault = 0.0
+    zrFracDefault = 0.0
```
sammiller11235, line 861: we need to be clear whether this is weight fraction or atomic fraction. Prefer `puWtFracDefault`. The `puFrac` parameter in ARMI is not weight fraction of total mass, but instead weight fraction of heavy metal.

armi/materials/material.py
```diff
@@ -845,9 +856,36 @@ class FuelMaterial(Material):
...
+        self._puFrac = 0
+        Material._init_()
+
+    @property
+    def puFrac(self):
+        """Get the Pu Frac."""
+        # ideally the parameters would be stored on component, not Block (parent, not grandparent)
+        myBlock = self.getAssociatedBlock()
+        if myBlock is not None:
+            return myBlock.p.puFrac
```
sammiller11235, line 873: maybe `myBlock.getComponent[Flags.FUEL].getMassFrac("PU")`?

albeanth:
This change is getting moved to a downstream ARMI application.

Issue #1818, opened 2024-08-13
Use of custom isotopics with thermal expansion turned on does not correctly preserve mass


keckler:
#1745 added the ability to combine a custom isotopic density vector with the material properties of a material class. I reviewed that PR and thought that everything looked good, but now I've been trying to use it on a real case and I've found a problem with mass conservation.

It appears that the component properly thermally expands. But during the thermal expansion, the component's number density is not adjusted away from the value specified in `custom isotopics` definition. This means that the density and the volume of the component become out of sync with each other, and as the component expands it gains mass.

I've attached a simple example input that you can use to reproduce the problem. It has only one single real component which uses a custom isotopic as well as a material class. Follow these steps to see the problem:
1. Note that `Tinput` == `Thot` in the component
2. Run the case
3. Open iPython
4. `import armi; armi.configure()`
5. `from armi.bookkeeping.db import databaseFactory`
6. `db = databaseFactory("anl-afci-177.h5", 'r')`
7. `with db:     r = db.load(0,0)`
8. `r.core[0][0].getMass()`
9. Note the mass as **26.567467**
10. Edit the blueprints with `Thot: 250`
11. Rerun the case
12. `db = databaseFactory("anl-afci-177.h5", 'r')`
13. `with db:     r = db.load(0,0)`
14. `r.core[0][0].getMass()`
15. Note the mass as **26.7636707**

These masses should be exactly equal, but they are not. I have tried the same thing _without_ using a custom isotopic in the component definition to confirm that this is not a problem with the axial expansion routine itself, and the masses were the same in that case.

@bsculac @onufer 

[example.zip](https://github.com/user-attachments/files/16601995/example.zip)

bsculac:
I replicated the problem, the output from the test case shows that the component with custom isotopics is initialized to the same mass, specifically the info message for the application of custom isotopics is the same:
`[impt] A custom material density was specified in the custom isotopics for non-custom material <Material: MOX>. The component density has been altered to 26.567467529113692.`

However, the BOL mass report indicates the difference:

25 C: 
![image](https://github.com/user-attachments/assets/d0ee70a0-aa32-459f-8e08-c53fbfd6953c)

250 C:
![image](https://github.com/user-attachments/assets/3729d222-63bf-4d39-943f-b7be8c55bb66)

So the error is occurring somewhere between component construction and the BOL mass report.

Outputs attached.

[25.txt](https://github.com/user-attachments/files/16606170/25.txt)
[...]

PR #1822, opened 2024-08-14
Fixing the affect of custom isotopics on material density

jakehader:
## What is the change?

This fixes some problems that were found with #1745. Those issues are described in #1818 .

We are fixing a bug that made it into the codebase.

---

## Checklist


john-science:
@jakehader   Just a reminder of the ARMI PR policy:

https://github.com/terrapower/armi/blob/bd0b43adadf6e9c0063e27e490353f4a16e1b623/doc/developer/tooling.rst?plain=1#L45-L49

We are all extremely busy, and if a PR gets opened before it is ready I will get an email:

* every single time you commit, and the CI passes
* for each and every time a single CI job fails

I love the help, and the interesting in fixing the feature. But in the future, please wait until the feature is ready for final review before opening a PR.

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -275,13 +287,33 @@ def setCustomDensity(self, constructedComponent, blueprint, matMods):
...
+                f = 1.0 / (1 + dLL) ** 3
+            
+            print(comp.density())
+            print(f)
+            print(comp.inputTemperatureInC)
+            print(comp.temperatureInC)
+            scaledDensity = comp.density() / f
+            densityRatio = density / scaledDensity
+            comp.changeNDensByFactor(densityRatio)
+            print(comp.density())
```
keckler, line 311: Reminder to remove print statements.

jakehader:
> @jakehader Just a reminder of the ARMI PR policy:
> 
> https://github.com/terrapower/armi/blob/bd0b43adadf6e9c0063e27e490353f4a16e1b623/doc/developer/tooling.rst?plain=1#L45-L49
> 
> We are all extremely busy, and if a PR gets opened before it is ready I will get an email:
> 
> * every single time you commit, and the CI passes
> * for each and every time a single CI job fails
> 
> I love the help, and the interesting in fixing the feature. But in the future, please wait until the feature is ready for final review before opening a PR.

Thanks for the reminder @john-science - I put the PR here but forgot to make it a draft. This was just intended to provide some ideas for how to resolve the ticket and to put something out there to be stress tested by @keckler and others. 

There are failing unit tests because I changed the construct arguments so that needs to be resolved still (probably a poor design to require case settings to build a component). Also, we will need to add more unit tests around custom Isotopics (same comment on the ticket).

keckler:
I ran some tests on my FFTF model to see if this branch works. A couple points:
1. The keff in my model increased by nearly 5000 pcm to a value that is very incorrect.
2. The predicted isothermal temperature reactivity coefficient looks correct.
3. The fuel density when the model is initialized to Thot is way too high (~10%).
4. The _change_ in density between different Thot values appears to be roughly correct, and the number of atoms is being correctly conserved.

So it seems like this change is half working. Somehow the initialized model is getting the densities very incorrect, though.

jakehader:
Was this fixed elsewhere @keckler?

keckler:
Nope, still broken

armi/reactor/blueprints/tests/test_customIsotopics.py
```diff
@@ -356,14 +361,19 @@ def test_densitiesAppliedToNonCustomMaterials(self):
...
 
+        dLL = fuel2.material.linearExpansionFactor(Tc=600, T0=25)
+        # the exponent here is 3 because inputHeightsConsideredHot = False.
+        # if inputHeightsConsideredHot were True, then we would use a factor of 2 instead
+        f = 1 / ((1 + dLL) ** 3)
+
         # Check that the density is set correctly on the custom density block,
         # and that it is not the same as the original
-        self.assertAlmostEqual(19.1, fuel2.density())
+        self.assertAlmostEqual(19.1 * f, fuel2.density())
```
keckler, line 371: I am pretty sure that this test was incorrect from the start. See https://github.com/terrapower/armi/pull/1745/files#r1865040194

armi/reactor/blueprints/tests/test_customIsotopics.py
```diff
@@ -356,14 +361,19 @@ def test_densitiesAppliedToNonCustomMaterials(self):
...
         self.assertNotAlmostEqual(fuel0.density(), fuel2.density(), places=2)
         # Check that the custom density block has the correct material
         self.assertEqual("UZr", fuel2.material.name)
         # Check that the block with only number densities set has a new density
-        self.assertAlmostEqual(19.1, fuel8.density())
+        self.assertAlmostEqual(19.1 * f, fuel8.density())
```
keckler, line 376: I am pretty sure that this test was incorrect from the start. See https://github.com/terrapower/armi/pull/1745/files#r1865040194

armi/reactor/blueprints/isotopicOptions.py
```diff
@@ -394,6 +394,7 @@ def apply(self, material):
         material.massFrac = dict(self.massFracs)
         if self.density is not None:
             if not isinstance(material, materials.Custom):
+                # TODO I think this warning should go away?
```
keckler, line 397: I would like a second set of eyes on the runLog statement below. I am 99% sure that that print should be removed completely, because we are now allowing for non-`Custom` materials to have density specified.

keckler: @bsculac

bsculac: The runlog appears to still be correct and applicable, as the density is not applied to `material` in this function if  `material` is not a custom material per line 396. I don't recognize anywhere in the changes that the density of a library material is being set, am I missing it?

Related: I believe this was intended to explicitly call out that setting a custom density for a library material using custom isotopics is not allowed. Presumably a library material would have known, vetted properties that should not be overwritten by users in a passive way. From recollection, using custom isotopics on a library material allows the density to be set on a component specifically using the custom isotopics, but using custom isotopics to set the density for the reference material is no bueno.

keckler: I think you're right. 

The thing that keeps tripping me up is the last sentence: "Only custom materials may have a density specified."

I feel like that can be reworded to make the intention more clear. Right now I read that and think that its saying "the density that you specified is doing nothing".

I'll try to rejigger a bit.

bsculac: Yeah, I vaguely remember struggling how to phrase this without being very wordy. Some ideas: 

"Only custom materials may have a density specified using custom isotopics, though components using custom materials may have their densities changed." 

"The reference density of materials in the materials library will not be changed by custom isotopics, but components with custom isotopics using a library material will use the density implied by the custom isotopics."

keckler:
Running my FFTF tests, I now believe that this PR is working correctly.

armi/reactor/blueprints/isotopicOptions.py
keckler: Alright, I updated the wording. Thanks for helping here.

keckler:
@john-science I believe that this is ready for your eyes. 

In addition to the updated unit tests, I did quite a bit of offline testing with my FFTF model to convince myself that this is working properly.

john-science:
Random thought: I think this PR is big enough for a release note in `doc/release/0.5.rst`.

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -186,7 +186,7 @@ def shape(self, shape):
     mergeWith = yamlize.Attribute(type=str, default=None)
     area = yamlize.Attribute(type=float, default=None)
 
-    def construct(self, blueprint, matMods):
+    def construct(self, blueprint, matMods, inputHeightsConsideredHot):
```
john-science, line 189: Okay, so in the `Block` class above this makes more sense to me. But the variable name here is strange. Because "input heights" makes sense if this Component is part of an `Assembly` in some way, sure.

But `Components` are used by the Mech team and we want to support non-pin-type reactors. (I understand we don't do great at this yet, but I'm working on it.)

What do you think?

john-science: Also, maybe this new parameter should have a default value?  What do you think?

keckler: This parameter and its name are directly reflective of the case setting of the same name. I tried to reflect that intention in the docstring, but maybe I could be more clear there.

When Jake put up the draft PR, he was originally passing an entire case settings object down to this level. I did away with that because I find it frustrating when I have to whip up an entire `cs` just for a single boolean. 

So I made the name for this parameter exactly the same as the case setting whose value should be passed down here. I thought that made things simpler, but if you think it'd be more clear either (1) with a different name, or (2) if we passed the entire `cs`, then I am open to those options.

keckler: With respect to a default value...

I'm not sure how I feel there. 

On the one hand, we could potentially make the default be the same as the default for the `inputHeightsConsideredHot` case setting (`True`). On the other hand, I really don't like that because the default on that case setting is the opposite of our typical workflow at this point in time (i.e. having axial expansion turned on).

I think my ideal solution would be if there were a way to not have to pass the value via method calls. But the only way I could conceive to do that is to store a `cs` on the `ComponentBlueprint`, and that is something I really don't want to do.

Maybe you see a clean way to get around all this?

john-science: It is definitely better to pass one value instead of the entire settings object. For sure. Nice.

I suppose "input dimensions considered hot" would have been a more accurate name for that setting. But the people who created it were definitely thinking of pin-type reactors.

doc/release/0.5.rst
```diff
@@ -97,6 +97,7 @@ Bug Fixes
 #. Fixed hex block rotation in ``plotBlockDiagram``. (`PR#1926 <https://github.com/terrapower/armi/pull/1926>`_)
 #. Fixed edge case in ``assemblyBlueprint._checkParamConsistency()``. (`PR#1928 <https://github.com/terrapower/armi/pull/1928>`_)
 #. Fixed wetted perimeter for hex inner ducts. (`PR#1985 <https://github.com/terrapower/armi/pull/1985>`_)
+#. Fixed number densities when custom isotopics and material classes are combined on a component (`PR#1822 <https://github.com/terrapower/armi/pull/1822>`_)
```
john-science, line 100:
```suggestion
#. Fixing number densities when custom isotopics and material properties are combined. (`PR#1822 <https://github.com/terrapower/armi/pull/1822>`_)
```

john-science: Minor changes:

* First word ends with "ing" (don't ask)
* shortened the language
* added a period, needed by RST-to-PDF

john-science (approved):

PR #1825, opened 2024-08-16
Fixing bug parsing ReportsEntryPoint CLI

### LINKED ISSUE #1816: `report` entry point input arguments are broken
The `report` entry point is supposed to take specific `(cycle, node)` combinations as an input argument, or a `max_node` or `min_node`. However, all of these inputs are parsed as strings, and are then later an attempt is made to compare `(cycle, node)` as a `tuple` to these strings. The check of the form `(int, int) not in str` fails because the two types cannot be compared.

https://github.com/terrapower/armi/blob/d7028672d345aadada35815ecbce8f2fdb0761aa/armi/cli/reportsEntryPoint.py#L120-L134

This code is not unit-tested at all, and it has presumably been broken for a long time.

john-science:
## What is the change?

When users use the `ReportsEntryPoint` CLI, they can pass in various integers. The CLI parser here never converts the CLI strings to integers, leading to an error if you use them.

To close #1816

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_reactors.py
+++ b/armi/reactor/tests/test_reactors.py
@@ -834,8 +834,6 @@ class HexReactorTests(ReactorTests):
         self.assertEqual(dominantFuel.getName(), "UZr")
         self.assertEqual(dominantCool.getName(), "Sodium")
 
-        self.assertEqual(list(dominantCool.getNuclides()), ["NA23"])
-
     def test_getSymmetryFactor(self):
         """
         Test getSymmetryFactor().
--- a/armi/reactor/tests/test_reactors.py
+++ b/armi/reactor/tests/test_reactors.py
@@ -834,6 +834,8 @@ class HexReactorTests(ReactorTests):
         self.assertEqual(dominantFuel.getName(), "UZr")
         self.assertEqual(dominantCool.getName(), "Sodium")
 
+        self.assertEqual(list(dominantCool.getNuclides()), ["NA23"])
+
     def test_getSymmetryFactor(self):
         """
         Test getSymmetryFactor().
--- a/armi/utils/mathematics.py
+++ b/armi/utils/mathematics.py
@@ -26,7 +26,7 @@ SCIPAT_SPECIAL = re.compile(r"([+-]?\d*\.\d+)[eEdD]?([+-]\d+)")
 
 def average1DWithinTolerance(vals, tolerance=0.2):
     """
-    Compute the average of a series of arrays with a tolerance.
+    Compute the average of a series of 1D arrays with a tolerance.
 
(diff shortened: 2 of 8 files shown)


armi/reactor/tests/test_reactors.py
```diff
@@ -834,8 +834,6 @@ def test_getDominantMaterial(self):
         self.assertEqual(dominantFuel.getName(), "UZr")
         self.assertEqual(dominantCool.getName(), "Sodium")
 
-        self.assertEqual(list(dominantCool.getNuclides()), ["NA23"])
```
mgjarrett, line 837: Why was this line deleted? Is it related to the main purpose of this PR, or is it in here by accident?

armi/utils/mathematics.py
```diff
@@ -57,7 +64,7 @@ def average1DWithinTolerance(vals, tolerance=0.2):
     return avg
 
 
-def convertToSlice(x, increment=False):
+def convertToSlice(x, increment=0):
```
mgjarrett, line 67: With this change, you could consider removing lines 121-122 below:

```
if increment is False:
    increment = 0
```

This would effectively change the API so we might want to avoid that, but I don't see any usage in our codebase where `increment=False` is passed explicitly.

mgjarrett:
The bug fix and unit tests for the `report` entry point looks good to me. Some of the changes appear unrelated to this bug fix, at least on the surface. The specific files that seem possibly unrelated are:

- test_reactors.py
- mathematics.py
- doc/conf.py

Were these changes intended to be included? And if so, how do they relate to the reports entry point?

armi/utils/mathematics.py
john-science: Yeah, I was just trying to preserve the old API.

But... I have no idea _why_ anyone would want to pass in `False` here. This is a math function, after all.

Hmmm

john-science:
> Were these changes intended to be included? And if so, how do they relate to the reports entry point?

Whoops! The last commit was to the wrong branch. One sec, I'll revert that.

mgjarrett (approved):

armi/reactor/tests/test_reactors.py
john-science: That is part of another PR now, I removed that line from this one.

Ignore it.

(It was because of a `DeprecationWarning`.)

armi/utils/mathematics.py
john-science: Okay, I removed that strange `False` from the API, and I am running a big suite of downstream unit tests.

john-science: Success!  Okay, done.

mgjarrett (approved):

PR #1835, opened 2024-08-26
Removing defunct flags CORE and REACTOR

john-science:
## What is the change?

Here I removed two flags that are barely used and are definitely unnecessary.

To close #1833

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database3.py
+++ b/armi/bookkeeping/db/database3.py
@@ -72,9 +72,8 @@ from armi.reactor.assemblies import Assembly
 from armi.reactor.blocks import Block
 from armi.reactor.components import Component
 from armi.reactor.composites import ArmiObject
-from armi.reactor.flags import Flags
 from armi.reactor.parameters import parameterCollections
-from armi.reactor.reactors import Core
+from armi.reactor.reactors import Core, Reactor
 from armi.settings.fwSettings.globalSettings import CONF_SORT_REACTOR
 from armi.utils import getNodesPerCycle
 from armi.utils.textProcessors import resolveMarkupInclusions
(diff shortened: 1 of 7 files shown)


armi/bookkeeping/db/database3.py
```diff
@@ -854,11 +853,7 @@ def _compose(self, comps, cs, parent=None):
             comp.add(child)
 
         if isinstance(comp, Core):
-            # TODO: This is also an issue related to geoms and which core is "The Core".
-            # We only have a good geom for the main core, so can't do process loading on
-            # the SFP, etc.
-            if comp.hasFlags(Flags.CORE):
-                comp.processLoading(cs, dbLoad=True)
+            comp.processLoading(cs, dbLoad=True)
```
keckler, line 856: The TODO comment that you removed is interesting.

When I look at the docstring for `Core`, it indicates that `Core` instances might be SPFs or other:
https://github.com/terrapower/armi/blob/35a0c87c4ccd97c5a04ba8eba330c0bfc90983f6/armi/reactor/reactors.py#L243-L247

That kind of worries me that we might have SFPs sitting around that might secretly be built from `Core`s. 

Looking further, though, we now have a dedicated `SpentFuelPool` class since #1349 . I wonder if the docstring that I pasted above might be in need of updating?

john-science: Yes, that old `TODO` is wildly out-of-date nonsense.  The `SFP` has been it's own class, and not any form of `CORE` in... at least 4 years.

keckler: Great, that is my understanding as well. 

Can we possibly update the docstring on `Core` as part of this PR then? Perhaps we can include some text in there that says like "If you want to build other groups of assemblies, see the `assemblyLists` module."

What do you think?

john-science: @keckler Agreed. Done. You will also find an SCRE on that change.

keckler (approved):

PR #1839, opened 2024-08-27
Documenting TightCoupling in the GlobalFluxInterface

### LINKED ISSUE #1570: GlobalFluxInterface has a silent, undocumented, default tight coupling
I'm working on an internal tight coupling workflow that should have converged, but didn't. After some digging, I found the `GlobalFluxInterface` sets an undocumented default tight coupler, even if you don't request to converge on keff

https://github.com/terrapower/armi/blob/24e36fe447914b1824f7bca01fdf0a648602b357/armi/physics/neutronics/globalFlux/globalFluxInterface.py#L65-L77

I understand why a default tight coupler for global flux makes sense. In most cases, this is probably what you want to do. But the behavior isn't documented, at least not in the [tight coupling docs](https://terrapower.github.io/armi/user/physics_coupling.html#tight-coupling), which led to my confusion. And it's not included in any logging so the log file wasn't _super_ helpful.

The thing that tipped me off was the addition of the `keff` convergence table in addition to the field I cared about. So there was _something_ about converging on keff, but I had no idea where it came from

john-science:
## What is the change?

The `GlobalFluxInterface` has a default `TightCoupling`. That was documented, but only deep inside the class, in the private method `_setTightCouplingDefaults()` where it was implemented.

So we are adding a note to this in the ARMI docs, to make this fact more visible.

Drew ran into this problem, and lost some time until he found this method. So, that's no good.  We are adding this to the docs, for better visibility.

To close #1570

---

## Checklist

### DIFF
--- a/doc/user/physics_coupling.rst
+++ b/doc/user/physics_coupling.rst
@@ -76,3 +76,12 @@ The total convergence of the power distribution is finally measured through the 
 
 .. math::
     \epsilon = \| \xi \|_{\inf} = \max \xi.
+
+
+The Global Flux Interface
+-------------------------
+The :py:class:`Global Flux Interface <armi.physics.neutronics.globalFlux.globalFluxInterface.GlobalFluxInterface>`
+class will attempt to set its own ``TightCoupler`` based on ``keff``. To see the specifics, see:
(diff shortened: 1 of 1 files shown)


drewj-tp (approved):
Thanks!

PR #1841, opened 2024-08-28
Allow SpentFuelPool types

mgjarrett:
## What is the change?

ARMI only allows there to be one child composite of a reactor that is the reactor core itself. This was previously checked with the flag `Flags.CORE`.

A recent PR (https://github.com/terrapower/armi/pull/1835) switched to checking the type of the composite, i.e. `if isinstance(c, Core)`. This broke downstream unit tests because some models have a `SpentFuelPool` which is a subclass of `Core`. The check needs to be updated to accept a `SpentFuelPool` in addition to a `Core`.

Downstream unit tests broke as a result of a recent PR changing the logic of what is allowed to be defined in blueprints.

---

## Checklist


mgjarrett:
Closing this, there is a better solution through updating the blueprints for downstream testing

Issue #1533, opened 2023-12-08
Assembly Rotation Cleanup


albeanth:
There several rotation methods:
1. https://github.com/terrapower/armi/blob/7e1e16736b7082c2ce4892bb36243fb080c63a48/armi/reactor/assemblies.py#L1215
2. https://github.com/terrapower/armi/blob/7e1e16736b7082c2ce4892bb36243fb080c63a48/armi/reactor/assemblies.py#L312
3. https://github.com/terrapower/armi/blob/7e1e16736b7082c2ce4892bb36243fb080c63a48/armi/reactor/blocks.py#L1827
4. https://github.com/terrapower/armi/blob/7e1e16736b7082c2ce4892bb36243fb080c63a48/armi/reactor/blocks.py#L1902
5. https://github.com/terrapower/armi/blob/7e1e16736b7082c2ce4892bb36243fb080c63a48/armi/physics/fuelCycle/assemblyRotationAlgorithms.py#L76

It looks like they all boil down to the two methods on `blocks.py`, `rotate` and `rotatePins`. The docstring for `rotatePins` suggests that calling it on it's own does _not_ constitute rotation.
https://github.com/terrapower/armi/blob/7e1e16736b7082c2ce4892bb36243fb080c63a48/armi/reactor/blocks.py#L1928-L1932

That said, I think we should consider making `rotatePins` private and/or putting warnings in docstrings (on that method, assembly.rotatePins, and assemblyRotationAlgorithms.py::simpleAssemblyRotation) saying that it's an incomplete rotation.

albeanth:
Additionally, `rotatePins` is fixed to increments of 60 degree rotations, while `rotate` is expected to be a flexible rotation. However, once `assembly.rotate` boils down to `blocks.rotate`, that flexible rotation angle needs to be in 60 degree increments
https://github.com/terrapower/armi/blob/7e1e16736b7082c2ce4892bb36243fb080c63a48/armi/reactor/blocks.py#L1839-L1844

drewj-tp:
Should we remove `Assembly.rotatePins`? If you call `Assembly.rotate`, it will call `Block.rotate` for all contained blocks, which in turn (at least for hex blocks) calls `Block.rotatePins`, which could double rotate your pins.

john-science:
[...]

PR #1846, opened 2024-08-30
Rotation cleanup: Remove Assembly.rotatePins, make Block.rotatePins private

drewj-tp:
## What is the change?

> 🚨 This change does *not* handle the rotation of all parameters that are tracked on pins and need to be rotated. That will come next after this cleanup because that will be a lot harder to manage and test and handle correctly 🚨 

As described in #1533, assembly rotation needed some love.

`Block.rotatePins` didn't do as much as people maybe would like (e.g., rotate parameters) so it is now private. `Block.rotate` should be preferred and will do (hopefully) everything people expect it to do

`Assembly.rotatePins` suffered from the same problem, and now is removed. `Assembly.rotate` should be preferred.

The `justCompute` flag in `Block._rotatePins` caught my eye and I found it was used when determining the new location of a pin with max burnup post-rotation. This means that in order to figure out where one pin would go, we would compute new locations for *every* pin in the block. And then maybe not use that as we rotate the assembly in all the possible orientations. This felt weird so I the "pin location post-rotate" functionality from that method and made it a new utility in `armi.utils.hexagon`. Now, when we are determining the optimal location of a pin during rotation, we just look at the proposed location of that pin.

`Block.rotate` did a lot (and it should!) but the method was very long. So I broke it up into a few private methods that are a bit more contained.

A large body of work is in progress to handle the parameter rotation. That work will be made easier after this cleanup.

## Other things to (maybe) do

- Remove `justCompute` switch in the now private `Block._rotatePins`? I can't find any usage of it now.
- Move the `Block.setOrientation` calls from `Block._rotatePins` to `Block.rotate`

---

## Checklist

### DIFF
--- a/armi/physics/fuelCycle/assemblyRotationAlgorithms.py
+++ b/armi/physics/fuelCycle/assemblyRotationAlgorithms.py
@@ -31,6 +31,7 @@ from armi.physics.fuelCycle.settings import CONF_ASSEM_ROTATION_STATIONARY
 
 
 def _rotationNumberToRadians(rot: int) -> float:
+    """Convert a rotation number to radians, assuming a HexAssembly."""
     return rot * math.pi / 3
 
 
--- a/armi/physics/fuelCycle/assemblyRotationAlgorithms.py
+++ b/armi/physics/fuelCycle/assemblyRotationAlgorithms.py
@@ -76,7 +76,6 @@ def _rotateByComparingLocations(aNow, aPrev):
     aPrev : Assembly
         Assembly that previously occupied the location of this assembly.
         If ``aNow`` has not been moved, this should be ``aNow``
-
     """
     rot = getOptimalAssemblyOrientation(aNow, aPrev)
     radians = _rotationNumberToRadians(rot)
--- a/armi/physics/fuelCycle/assemblyRotationAlgorithms.py
+++ b/armi/physics/fuelCycle/assemblyRotationAlgorithms.py
@@ -19,8 +19,10 @@ Notes
 These algorithms are defined in assemblyRotationAlgorithms.py, but they are used in:
 ``FuelHandler.outage()``.
 
-.. warning:: Nothing should do in this file, but rotation algorithms.
+.. warning:: Nothing should go in this file, but rotation algorithms.
 """
+import math
+
 from armi import runLog
 from armi.physics.fuelCycle.hexAssemblyFuelMgmtUtils import (
     getOptimalAssemblyOrientation,
@@ -28,8 +30,12 @@ from armi.physics.fuelCycle.hexAssemblyFuelMgmtUtils import (
 from armi.physics.fuelCycle.settings import CONF_ASSEM_ROTATION_STATIONARY
 
 
+def _rotationNumberToRadians(rot: int) -> float:
+    return rot * math.pi / 3
+
+
 def buReducingAssemblyRotation(fh):
-    r"""
+    """
     Rotates all detail assemblies to put the highest bu pin in the lowest power orientation.
 
     Parameters
@@ -48,29 +54,37 @@ def buReducingAssemblyRotation(fh):
         aNow = fh.r.core.getAssemblyWithStringLocation(aPrev.lastLocationLabel)
         # no point in rotation if there's no pin detail
         if aNow in hist.getDetailAssemblies():
-            rot = getOptimalAssemblyOrientation(aNow, aPrev)
-            aNow.rotatePins(rot)  # rot = integer between 0 and 5
+            _rotateByComparingLocations(aNow, aPrev)
             numRotated += 1
-            # Print out rotation operation (mainly for testing)
-            # hex indices (i,j) = (ring,pos)
-            (i, j) = aNow.spatialLocator.getRingPos()
-            runLog.important(
-                "Rotating Assembly ({0},{1}) to Orientation {2}".format(i, j, rot)
-            )
 
-    # rotate NON-MOVING assemblies (stationary)
     if fh.cs[CONF_ASSEM_ROTATION_STATIONARY]:
         for a in hist.getDetailAssemblies():
             if a not in fh.moved:
-                rot = getOptimalAssemblyOrientation(a, a)
-                a.rotatePins(rot)  # rot = integer between 0 and 6
+                _rotateByComparingLocations(a, a)
                 numRotated += 1
-                (i, j) = a.spatialLocator.getRingPos()
-                runLog.important(
-                    "Rotating Assembly ({0},{1}) to Orientation {2}".format(i, j, rot)
-                )
 
     runLog.info("Rotated {0} assemblies".format(numRotated))
+
+
+def _rotateByComparingLocations(aNow, aPrev):
+    """Rotate an assembly based on its previous location.
+
+    Parameters
+    ----------
+    aNow : Assembly
+        Assembly to be rotated
+    aPrev : Assembly
+        Assembly that previously occupied the location of this assembly.
+        If ``aNow`` has not been moved, this should be ``aNow``
+
+    """
+    rot = getOptimalAssemblyOrientation(aNow, aPrev)
+    radians = _rotationNumberToRadians(rot)
+    aNow.rotate(radians)
+    (ring, pos) = aNow.spatialLocator.getRingPos()
+    runLog.important(
+        "Rotating Assembly ({0},{1}) to Orientation {2}".format(ring, pos, rot)
+    )
 
 
 def simpleAssemblyRotation(fh):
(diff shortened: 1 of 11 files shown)


jakehader:
Maybe this exists already, but it would be good to know if something has already been rotated relative to its as-designed (e.g., by construction) and by how much. Maybe this is a good use of flags or some type of flag on the composite?

drewj-tp:
> Maybe this exists already, but it would be good to know if something has already been rotated relative to its as-designed layout (e.g., by construction) and by how much. Maybe this is a good use of flags or some type of flag on the composite?

We have `Block.p.orientation` and `Assembly.p.orientation` so assuming a value of zero indicates the "as designed" orientation, that may suffice. If I'm missing something, can you make a ticket so I can track that as part of the upcoming assembly rotation work?

drewj-tp:
@john-science I'm not sure where to add this to the release notes. We have a `0.4.1` section that's TBD, but this is a breaking change and feels like it deserves a larger release number bump. That said, there are other removals in the current `0.4.1` section

drewj-tp:
@john-science or @albeanth this is ready to go and has some interesting connections to something @mgjarrett put up - #1859

mgjarrett:
I have one of the only internal downstream uses of the public `rotatePins` method with `justCompute=True` (at least I know of). The `justCompute=True` is really just about doing hex lattice logic / math and doesn't need to be bound to a specific instance of a block. I would be fine with removing the `justCompute` argument entirely. I can refactor the downstream application to use the new `armi.utils.hexagon.getIndexOfRotatedCell`

drewj-tp:
> @drewj-tp Does this PR close the ticket #1533 ?

Yes, I just didn't add the special language for github to link the two. They should be linked now

armi/reactor/tests/test_blocks.py
```diff
@@ -1454,33 +1454,33 @@ def test_106_getAreaFractions(self):
...
         self.assertEqual(b.p.pinLocation[5], 2)
 
-        self.assertRaises(ValueError, b.rotatePins, -1)
-        self.assertRaises(ValueError, b.rotatePins, 10)
-        self.assertRaises((ValueError, TypeError), b.rotatePins, None)
-        self.assertRaises((ValueError, TypeError), b.rotatePins, "a")
+        self.assertRaises(ValueError, b._rotatePins, -1)
+        self.assertRaises(ValueError, b._rotatePins, 10)
+        self.assertRaises((ValueError, TypeError), b._rotatePins, None)
+        self.assertRaises((ValueError, TypeError), b._rotatePins, "a")
```
john-science, line 1483: Can we get a unit test showing that the new `rotate()` and `_rotatePins()` work for blocks/assemblies that don't have pins at all?  I was looking to figure it out and then realized it would be an informative test to have.

Looking at your code, that would have to still be a `HexBlock`, right?

armi/utils/hexagon.py
```diff
@@ -145,3 +145,50 @@ def numPositionsInRing(ring):
         rings is indexed to 1, i.e. the centermost position in the lattice is ``ring=1``.
     """
     return (ring - 1) * 6 if ring != 1 else 1
+
+
+def totalPositionUpToRing(ring: int) -> int:
```
john-science, line 150: This could use some unit tests.

For instance, something like:

```python
def test_totalPositionUpToRing(self):
    self.assertEqual(totalPositionUpToRing(1), 1)
    self.assertEqual(totalPositionUpToRing(3), 19)

    for r in range(1, 10):
        tot = sum([numPositionsInRing(rr) for rr in range(1, r + 1)])
        self.assertEqual(totalPositionUpToRing(r), tot)
```

Just thinking aloud.

armi/utils/hexagon.py
```diff
@@ -145,3 +145,50 @@ def numPositionsInRing(ring):
...
+    """Return the number of positions in a hexagon with a given number of rings."""
+    return 1 + 3 * ring * (ring - 1)
+
+
+def getIndexOfRotatedCell(initialCellIndex: int, orientationNumber: int) -> int:
```
john-science, line 155: This could also use some direct unit testing.  For instance:

* Test a `self.assertRaises(ValueError)` for invalid orientation number.
* Test a couple known results for when `initialCellIndex > 1`.
* Test a couple known results for when `initialCellIndex== 1`.
* Test a `self.assertRaises(ValueError)` for negative cell number.

armi/physics/fuelCycle/assemblyRotationAlgorithms.py
```diff
@@ -19,17 +19,23 @@
...
+
 from armi import runLog
 from armi.physics.fuelCycle.hexAssemblyFuelMgmtUtils import (
     getOptimalAssemblyOrientation,
 )
 from armi.physics.fuelCycle.settings import CONF_ASSEM_ROTATION_STATIONARY
 
 
+def _rotationNumberToRadians(rot: int) -> float:
+    return rot * math.pi / 3
```
john-science, line 34:
```suggestion
def _rotationNumberToRadians(rot: int) -> float:
    """Convert a rotation number to radians, assuming a HexAssembly."""
    return rot * math.pi / 3
```

john-science: We have a lot of implicit assumptions of hex-based assemblies around ARMI. They are frequently traps. I just want to call them out.

You can certainly fix/change the wording, to your heart's content.

armi/physics/fuelCycle/assemblyRotationAlgorithms.py
```diff
@@ -48,31 +54,39 @@ def buReducingAssemblyRotation(fh):
...
+    """Rotate an assembly based on its previous location.
+
+    Parameters
+    ----------
+    aNow : Assembly
+        Assembly to be rotated
+    aPrev : Assembly
+        Assembly that previously occupied the location of this assembly.
+        If ``aNow`` has not been moved, this should be ``aNow``
+
```
john-science, line 79:
```suggestion
```

armi/reactor/assemblies.py
```diff
@@ -1272,7 +1263,27 @@ def isOnWhichSymmetryLine(self):
...
+        ----------
+        rad : float
+            Counter clockwise rotation in radians. **MUST** be in increments of
+            60 degrees (PI / 3)
+
+        Raises
+        ------
+        ValueError
+            If rotation is not divisible by pi / 3.
+
```
john-science, line 1279:
```suggestion
```

armi/utils/tests/test_hexagon.py
```diff
@@ -43,3 +44,43 @@ def test_numPositionsInRing(self):
...
         self.assertEqual(hexagon.numPositionsInRing(3), 12)
         self.assertEqual(hexagon.numPositionsInRing(4), 18)
+
+    def test_rotatedCellCenter(self):
+        # Cell one is always the center of the hexagon and its position
+        # is rotation invariant
+        for rot in range(6):
+            self.assertTrue(hexagon.getIndexOfRotatedCell(1, rot), 1)
+
+    def test_rotatedFirstRing(self):
```
drewj-tp: @john-science here are two direct tests for `hexagon.getIndexOfRotatedCell` like you called out. I'm adding tests for the error conditions and changing the comment to a docstring on `test_rotatedCellCenter`. Is there anything you'd like to see beyond that?

armi/utils/hexagon.py
drewj-tp: Good call

armi/reactor/tests/test_blocks.py
drewj-tp: `Block.rotate` is not implemented, only `HexBlock.rotate`

https://github.com/terrapower/armi/blob/8feea21034480954bc793ffd8901ffee36c96b3f/armi/reactor/blocks.py#L1538-L1546

Similarly, there is no `Block.rotatePins`, it's only defined on `HexBlock`.

Are you looking for some tests on an empty `HexBlock` with no children? So it doesn't have any pins?

john-science: Yeah, I just want a test that shows (now and 3 years from now, when people have monkied with the feature a lot) that the rotation still works in the case there are no pins.

Or show that it fails in the way you want it to.

john-science (approved):
Okay, you can decide if you want to add that rotation test for a hex block with no pins. 

Other than that, I think this looks good.

john-science (approved):

keckler:
This PR breaks my FFTF model, FYI. I was using `assembly.rotatePins()` (whether that was bugged or not, I don't know).

jakehader:
> This PR breaks my FFTF model, FYI. I was using `assembly.rotatePins()` (whether that was bugged or not, I don't know).

@drewj-tp - FYI

PR #1849, opened 2024-09-02
Adding GitHub content protection rule to CI

john-science:
## What is the change?

I have set our CI workflows to only have "read"-level access to the repository by default. That is, when CI runs, it can clone the repo and build a venv or run tests, but the CI does not have the power to write to the GitHub repository (which it could have with the GitHub auth token you can inject into a GitHub Action).

GitHub added [this security feature](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#permissions) in 2021. It is not a complete solution onto itself. Anyone on our GitHub "team" could commit code directly to our repo, of course. But we have our repo set such that when someone who isn't on our "team" opens a PR, the CI will not run until the repo Owner approves the workflows. This gives us a chance to review the changes they made.

So it would be a Big Red Flag if someone we don't know opened a PR into our repo, and they had to visibly change this repo setting from "read" to "write", or they just removed the permissions entirely. 

Thus, human intervention is still part of our security system. But this PR provides us another safeguard and a stopgap against security concerns.

For more information [this blog post](https://dev.to/azu/how-to-set-github-actions-s-permissions-hln) was quick and helpful.

---

## Checklist

### DIFF
--- a/.github/workflows/black.yaml
+++ b/.github/workflows/black.yaml
@@ -2,17 +2,20 @@ name: black
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 # use workaround due to: https://github.com/psf/black/issues/2079#issuecomment-812359146
 jobs:
-    check-formatting:
-        runs-on: ubuntu-24.04
-        steps:
-            - uses: actions/checkout@v2
-            - name: Set up Python 3.11
-              uses: actions/setup-python@v2
-              with:
-                python-version: '3.11'
-            - name: Install Black
-              run: pip install 'black==22.6.0'
-            - name: Run black --check .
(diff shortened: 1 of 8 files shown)


drewj-tp (approved):
This is a good change. 

Poking through the blog and the docs (which were helpful!) I may be missing something, but it looks like by default, the auth token used in CI **from forks** has read permissions by default - https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

image

so maybe that is new? anyways, I think this is still a good change that we should merge.

opotowsky (approved):

Issue #1440, opened 2023-10-18
Potential for incorrect material properties upon restart or db load


keckler:
Whenever a material property is modified by a `material modification` and that modifier doesn't alter the material in such a way that the change is recorded in the database, the property will likely be wrong when the reactor is loaded from the database. This includes instances of just playing around with the reactor in iPython or in restarts/snapshots.

A concrete example of this is the `TD_frac` material modifier in either `UraniumOxide` or `B4C`. Because `TD_frac` (and its associated instance variable `theoreticalDensityFrac`) is not written to the database directly, it is left at its default when the reactor is loaded back up from the database. So if you do a restart run with a case that has either of those materials in it, your material densities will be wrong if your `TD_frac` entries don't exactly equal the default values in the material class. For `UraniumOxide` that is 1.0 and for `B4C` that is 0.90.

There are other materials, including `ThoriumOxide` and some lesser-used materials that will have a similar issue. 

This same type of error is possible for other material modifiers as well. For instance if your thermal properties like thermal conductivity or heat capacity are dependent on composition as specified in the `material modifications`. I don't see any examples of this in the public framework.

I believe this has the potential to be causing lots of analysis problems, depending on how one is using restart cases in their workflow.

john-science:
Yes, I believe you correctly understand the current situation.

The Database treats materials as static things.

But people can inject whatever code they _want_ into the materials, of course. Including adding a lot of state to a material object. But all that arbitrary state information will be lost when you translate your run to the database.

This shouldn't affect most ARMI runs, but it would affect restarted runs. Mostly, it could affect post-run analysis.

I think it is infeasible to try to store any arbitrary materials state in the database. But if there is one important variable, it should be stored as a parameter on the component, and not on the material.

keckler:
To clarify, I do not believe this bug is impacting the number densities stored as a component parameter, at least in the way that we are currently using the material's `density()` method. 

In the current usage, `c.material.density()` is just used to instantiate the number densities. From that point on (at least in neutronics-type calculations), if we want to know the density ARMI actually looks to the number density parameter, since that is updated with burnup. This means that the number densities going into our neutronics calculations are not adversely impacted by this.

Other use cases may use `c.material.density()` in broader ways, though, which could be dangerous during restarts/snapshots.

drewj-tp:
Taking the tact of attempting to make the theoretical density a component parameter, defaulting to one

[...]

Issue #1436, opened 2023-10-16
Density and thermal expansion data of `UraniumOxide` class are not completely aligned


keckler:
Long story short, if you build an assembly with `UraniumOxide` in it and then expand components of that assembly, you will see that the total mass is different depending on how you calculate it. My belief is that this is caused by the `UraniumOxide.density()` and `UraniumOxide.linearExpansion()` and/or `UraniumOxide.linearExpansionPercent()` methods not being completely aligned with each other. Here is an example of the problem.

Let's build a very simple assembly with just one component, made of `UraniumOxide`.
```python
import copy

import armi; armi.configure()

from armi.reactor.converters.axialExpansionChanger import AxialExpansionChanger
from armi.reactor import grids, components, assemblies, blocks
from armi.reactor.flags import Flags

bFuel = blocks.HexBlock("fuel", height=1)
bDummy = blocks.HexBlock("dummy", height=1)

bFuel.add(components.Circle("fuel", "UraniumOxide", Tinput=25, Thot=25, od=2, id=0, mult=1))
bDummy.add(components.Hexagon("dummy", "Void", Tinput=25, Thot=25, op=20, ip=0, mult=1))
bDummy.p.flags = Flags.DUMMY
bFuel.setAxialExpTargetComp(bFuel[0])
bDummy.setAxialExpTargetComp(bDummy[0])

a = assemblies.HexAssembly("fuel")
a.spatialGrid = grids.axialUnitGrid(2)
a.spatialGrid.armiObject = a

a.add(copy.deepcopy(bFuel))
a.add(copy.deepcopy(bDummy))

a.calculateZCoords()
a.reestablishBlockOrder()
```

We can now change the temperature of the component and perform full expansion:
```python
axc = AxialExpansionChanger(detailedAxialExpansion=True)
axc.setAssembly(a)
dT = 500
axc.expansionData.updateComponentTemp(a[0][0], a[0][0].temperatureInC + dT)
axc.expansionData.computeThermalExpansionFactors()
axc.axiallyExpandAssembly()
```

If we now query the mass of that component in two separate ways, we get different answers:
```python
a[0][0].getMass()
34.319448492499255

a[0][0].getVolume() * a[0][0].material.density(Tc=25+dT)
34.476302875694955
```

That clearly is not correct. This difference is small, but it presents a problem when one is trying to show that mass is conserved between their blueprints and the actual instantiated ARMI model.

[...]

PR #1852, opened 2024-09-03
Write and read material theoretical density to/from database

### LINKED ISSUE #1863: Material theoretical density is not written to database and can be lost in restart runs
#1440 is a large ticket that points to a broader problem that if materials have some special attributes that determine properties, those attributes are not written to database and lost in restarts. That's a large architectural problem.

A smaller, specific, and urgent problem is

> A concrete example of this is the TD_frac material modifier in either UraniumOxide or B4C. Because TD_frac (and its associated instance variable theoreticalDensityFrac) is not written to the database directly, it is left at its default when the reactor is loaded back up from the database. So if you do a restart run with a case that has either of those materials in it, your material densities will be wrong if your TD_frac entries don't exactly equal the default values in the material class. For UraniumOxide that is 1.0 and for B4C that is 0.90.

drewj-tp:
## What is the change?

Added a new component parameter `theoreticalDensityFrac`. When loading from blueprints, the material's theoretical density is assigned to this parameter. Since it's a parameter, it will be written to the database and then loaded onto the component when reading from db. But, we need to reassign the theoretical density to the material because a lot of `Material.density` calls use this value.

This PR adds logic to the database loading sequence to apply the theoretical density component parameter to the material's theoretical density.

This PR also wraps `Material.density` in a decorator that will prefer to call `self.parent.density` when applicable.

Closes #1863

## Checklist


keckler:
I have not done a detailed review of this yet, but a few comments:
1. I have added @john-science as a reviewer, because we really need his eyes here
2. In some ways, this PR is very related to https://github.com/terrapower/armi/pull/1062. If you haven't seen that one, I encourage you to take a look.
3. I also do not particularly like this solution. Based on my first look at it, this whole situation is very ugly and messy, and it seems to me like a symptom of a deeper problem with how we are handling materials and their relationship with components. Perhaps it is worth thinking about this some more before we continue down this path.
4. Is there an issue that we need to link this PR against?

Anyways, thank you for the PR and for the thought that you put into it. Clearly this is not the simplest problem to solve!

drewj-tp:
Thanks @keckler. Lots of good points here.

Definitely feel the connection to #1062, in addition to #1140 and #1819: materials are not as stateless as we want them to be. And I think that's the deeper problem you're referencing. And that's a larger effort than this symptom (td_frac) but I'm open to discussing with you, @john-science @opotowsky and @albeanth.

> Is there an issue that we need to link this PR against

Yesish? #1440 touches on TD_frac as a symptom of the statefulness. This PR doesn't close that entire issue but does address a part of it. 

> There are some internal materials that we will probably have to migrate.

This _should_ be transparent to materials. It just reinforces / builds the connection between BP -> material modifications -> TD frac attribute <-> component property <-> database. I'll double check

drewj-tp:
Chatted w/ @john-science and a maybe better solution is to have `Material.getTD` fall back to `self.parent.p.theoreticalDensityFrac` because we have reasonable confidence these materials will have a component parameters (yes they have state still but this is an okay(ish) state to have). Similarly, `Material.adjustTD` would be an alternate setter for the parents td frac param.

This could also coincide w/ some deprecation warnings about removing the material td frac related things. But we'd need to make sure we shift the scaling by TD from the material side to the component side. e.g., from
```python
# usage by components
dens = self.material.density()
# materials.py
def density(...):
    return actualDensity * self.getTD()
```
to
```python
# usage
dens = self.getTD() * self.material.density()
# materials.py
def density(...):
    return actualDensity
```
Which wouldn't be the worst change. Just would need to make sure the change got propogated well.

drewj-tp:
@keckler @john-science I updated the approach to have `Material` theoretical density fallback to `self.parent.p.theoreticalDensityFrac` which kind of breaks statelessness but at least lets the data be written to / loaded from database correctly.

Should be good to go!

drewj-tp:
Had a chat with @keckler this morning and we both agree the recent changes here are better but still have some yuck. His main critique (correct me if I'm wrong) is that we have multiple sources of truth for the density. And `Component.density()` will diverge from `Component.material.density() for reasons beyond theoretical density (depletion updates composition, thermal expansion). And that `Component.density`, being derived from the compositions, is correctly written to / read from the database.

Therefore, if you were to examine a "from blueprints" reactor with a modified theoretical density compared to a "from database" reactor at the same timestep,
1. `Component.density` would agree for the same component, but
2. `Component.material.density` would not agree (due to the TD frac)

A solution we are circling around involves discouraging calls to `Material.density` except for when freshly creating a component. Or if there is a parent attached to the material. If we can find a way to enforce this (docs, warnings, maybe exceptions (later?)) then users doing restart analysis would see something like
```python
>>> b = r.core[0][0]
>>> c = b[0]
>>> dens = c.material.density(...)
[warn] The density of a component is correct only from that component, not its material
```
Where the warning is logged if the density of a material with a parent is called.

We'd probably want to do something to consume this warning when we are initializing the component and/or for the first time. Somewhere around

https://github.com/terrapower/armi/blob/9ea428bbbc196e80729e6153b2dc7e249d744aad/armi/reactor/components/component.py#L1263-L1271

But we also have a non-zero number of tests that look at the agreement between `Component.density(...)` and `Component.material.density(...)`

And, what do we do about methods like

https://github.com/terrapower/armi/blob/9ea428bbbc196e80729e6153b2dc7e249d744aad/armi/materials/material.py#L653-L674

that are convenience methods but still subject to this problem

We'd also need to make sure internal projects (and external users!) are avoiding calls to `Component.material.density` when appropriate.

drewj-tp:
Related to https://github.com/terrapower/armi/issues/1436

drewj-tp:
PR builds are passing but push PRs are failing?

- mac failed on push : https://github.com/terrapower/armi/actions/runs/10778501995/job/29889968100?pr=1852
- mac passed on pr - https://github.com/terrapower/armi/actions/runs/10778502638/job/29889970229?pr=1852
- windows failed on push - https://github.com/terrapower/armi/actions/runs/10778501988/job/29889968010?pr=1852
- windows passed on pr - https://github.com/terrapower/armi/actions/runs/10778502629/job/29889969948?pr=1852

both [failures ](https://github.com/terrapower/armi/actions/runs/10778501995/job/29889968100#step:5:566)are due to xs library read/write?
```python
FAILED armi/nuclearDataIO/tests/test_xsLibraries.py::Isotxs_merge_Tests::test_cannotMergeXSLibWithSameNuclideNames - OSError: Failed to read/write <IsotxsIO /Users/runner/work/armi/armi/armi/nuclearDataIO/tests/fixtures/ISOAB> 


  File "/Users/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 329, in readWrite
    nuclideIO.rwNuclide()
  File "/Users/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 483, in rwNuclide
    self._rw7DRecord(blockNumIndex, subBlock)
  File "/Users/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 686, in _rw7DRecord
    dataVals.append(record.rwFloat(0.0))
                    ^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/armi/armi/armi/nuclearDataIO/cccc/cccc.py", line 381, in rwFloat
    (f,) = struct.unpack("f", self._stream.read(self._floatSize))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
struct.error: unpack requires a buffer of 4 bytes
```
🤷 
Going to re-roll tests

armi/reactor/components/component.py
```diff
@@ -428,6 +432,24 @@ def getProperties(self):
         """
         return self.material
 
+    def _resolveMaterialParams(self):
+        """Pull any important parameters off the material."""
+        # Two cases we need to handle.
```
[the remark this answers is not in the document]
drewj-tp: This functionality was removed

john-science:
> PR builds are passing but push PRs are failing?
> 🤷 Going to re-roll tests

Well, we can't have tests that only pass _some_ of the time. That's no good. Maybe we'll run them 10 more times and see if they pass.

armi/materials/material.py
```diff
@@ -32,6 +33,36 @@
...
+
+    If a Material is linked to a Component, ``Material.density`` may produce
+    different results from ``Component.density``. The component's density
+    is considered the source of truth because it incorporates changes in volume,
+    composition, and temperature in concert with the state of the reactor.
+
+    See Also
+    --------
+    - https://github.com/terrapower/armi/issues/1440
+
```
john-science, line 47:
```suggestion
```

john-science: We don't link to PRs or Issues in the code.

The idea is that either that information is important enough to write in a docstring (or the docs), or it's not.

armi/reactor/components/component.py
```diff
@@ -1309,6 +1315,21 @@ def getFuelMass(self) -> float:
...
+        some properties could be doubly applied.
+        """
+        # This exists because the theoretical density is initially defined as a material
+        # modification, and then stored as a Material attribute. When reading from blueprints,
+        # the blueprint loader sets the theoretical density parameter from the Material
+        # attribute. Component parameters are also set when reading from the database.
+        # But, we need to set the Material attribute so routines that fetch a material's
+        # density property account for the theoretical density.
+        # See https://github.com/terrapower/armi/issues/1440
+        self.material.adjustTD(self.p.theoreticalDensityFrac)
```
john-science, line 1331:
```suggestion
        some properties could be doubly applied.

        This exists because the theoretical density is initially defined as a material
        modification, and then stored as a Material attribute. When reading from blueprints,
        the blueprint loader sets the theoretical density parameter from the Material
        attribute. Component parameters are also set when reading from the database.
        But, we need to set the Material attribute so routines that fetch a material's
        density property account for the theoretical density.
        """
        self.material.adjustTD(self.p.theoreticalDensityFrac)
```

john-science: Two things:

1. We don't link tickets or PRs in the docstrings. Either we document completely in the code, or it's not worth documenting. (At least in ARMI.)
2. I figure if we have 15 lines of comments, that could go in the docstring. Thoughts?

armi/materials/material.py
drewj-tp: Fair enough

armi/reactor/components/component.py
drewj-tp: > I figure if we have 15 lines of comments, that could go in the docstring. Thoughts?

🤷 I could go either way. It felt a bit too implementation-specific / in-the-weeds for a docstring. But the comment-to-code ratio is high enough that it probably deserves a bit more significance

john-science:
A Linux unit test just failed:

```bash
FAILED armi/nuclearDataIO/tests/test_xsLibraries.py::Isotxs_merge_Tests::test_canRemoveIsotopes - OSError: Failed to read/write <IsotxsIO /home/runner/work/armi/armi/armi/nuclearDataIO/tests/fixtures/ISOAB> 


  File "/home/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 329, in readWrite
    nuclideIO.rwNuclide()
  File "/home/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 483, in rwNuclide
    self._rw7DRecord(blockNumIndex, subBlock)
  File "/home/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 686, in _rw7DRecord
    dataVals.append(record.rwFloat(0.0))
  File "/home/runner/work/armi/armi/armi/nuclearDataIO/cccc/cccc.py", line 381, in rwFloat
    (f,) = struct.unpack("f", self._stream.read(self._floatSize))
struct.error: unpack requires a buffer of 4 bytes
===== 1 failed, 2273 passed, 19 skipped, 26 warnings in 264.15s (0:04:24) ======
```

Interesting.

drewj-tp:
> A Linux unit test just failed:
> 
> ```shell
> FAILED armi/nuclearDataIO/tests/test_xsLibraries.py::Isotxs_merge_Tests::test_canRemoveIsotopes - OSError: Failed to read/write <IsotxsIO /home/runner/work/armi/armi/armi/nuclearDataIO/tests/fixtures/ISOAB> 
> 
> 
> Traceback (most recent call last):
>   File "/home/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 329, in readWrite
>     nuclideIO.rwNuclide()
>   File "/home/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 483, in rwNuclide
>     self._rw7DRecord(blockNumIndex, subBlock)
>   File "/home/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 686, in _rw7DRecord
>     dataVals.append(record.rwFloat(0.0))
>   File "/home/runner/work/armi/armi/armi/nuclearDataIO/cccc/cccc.py", line 381, in rwFloat
>     (f,) = struct.unpack("f", self._stream.read(self._floatSize))
> struct.error: unpack requires a buffer of 4 bytes
> ===== 1 failed, 2273 passed, 19 skipped, 26 warnings in 264.15s (0:04:24) ======
> ```
> 
> Interesting.

Very interesting. Same errors I saw in https://github.com/terrapower/armi/pull/1852#issuecomment-2338748072

Maybe it's a race condition w/ the parallel tests? https://github.com/terrapower/armi/actions/runs/10797243037/job/29948075116?pr=1852#step:6:409

```python
OSError: Failed to read/write <IsotxsIO /home/runner/work/armi/armi/armi/nuclearDataIO/tests/fixtures/ISOAB>
```

armi/reactor/components/component.py
```diff
@@ -226,6 +226,8 @@ class Component(composites.Composite, metaclass=ComponentType):
 
     pDefs = componentParameters.getComponentParameterDefinitions()
 
+    material: materials.Material
```
john-science, line 229: I feel like this will confuse people with the import at the top of this file:

> from armi.materials import material

At the very least, this variable needs to be called `materialClass`.

But, probably it would be cleaner not to save class types as class attributes. Just do `isinstance(Material)` directly, it will be easier to read.

armi/reactor/components/component.py
```diff
@@ -1309,6 +1315,21 @@ def getFuelMass(self) -> float:
...
+        some properties could be doubly applied.
+
+        This exists because the theoretical density is initially defined as a material
+        modification, and then stored as a Material attribute. When reading from blueprints,
+        the blueprint loader sets the theoretical density parameter from the Material
+        attribute. Component parameters are also set when reading from the database.
+        But, we need to set the Material attribute so routines that fetch a material's
+        density property account for the theoretical density.
+        """
+        self.material.adjustTD(self.p.theoreticalDensityFrac)
```
john-science, line 1331: So this is a little confusing here, `self.material` is just the class (not the instance/object) `Material`, it is not _A_ `Material()` instance.  So this line:

```python
Material.adjustTD(self.p.theoreticalDensityFrac)
```

How does that work?

armi/reactor/components/component.py
drewj-tp: There is no assignment nor variable creation; there's no `=` sign. It's just a type hint that every attribute *should* be of type `materials.Material`.
```python
In [1]: class D:
   ...:     x: int
   ...:

In [2]: d = D()

In [3]: d.x
AttributeError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 d.x

AttributeError: 'D' object has no attribute 'x'
```
https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#classes

john-science: Yeah, I see that is a colon, I _read_ it as an equal sign.

And then I probably got excited and just slapped a "Mission Accomplished" banner on that aircraft carrier and moved on.

I see it now.

drewj-tp:
In the failed instances of this test, the following have all failed
- `FAILED armi/nuclearDataIO/tests/test_xsLibraries.py::Isotxs_merge_Tests::test_canRemoveIsotopes - OSError: Failed to read/write <IsotxsIO /home/runner/work/armi/armi/armi/nuclearDataIO/tests/fixtures/ISOAA>`
- `FAILED armi/nuclearDataIO/tests/test_xsLibraries.py::Isotxs_merge_Tests::test_canRemoveIsotopes - OSError: Failed to read/write <IsotxsIO D:\a\armi\armi\armi\nuclearDataIO\tests\fixtures\ISOAB>`
- `FAILED armi/nuclearDataIO/tests/test_xsLibraries.py::Isotxs_merge_Tests::test_cannotMergeXSLibWithSameNuclideNames - OSError: Failed to read/write <IsotxsIO /Users/runner/work/armi/armi/armi/nuclearDataIO/tests/fixtures/ISOAB>`
- `FAILED armi/nuclearDataIO/tests/test_xsLibraries.py::Isotxs_merge_Tests::test_canRemoveIsotopes - OSError: Failed to read/write <IsotxsIO D:\a\armi\armi\armi\nuclearDataIO\tests\fixtures\ISOAA>`

What can we learn from this?

1. Confirming operating system agnostic - `D:\a\armi\...` and `/home/runner/...` and `/Users/runner/...`
2. Two fixture files `nuclearDataIO/tests/fixtures/ISOAA` and `nuclearDataIO/tests/fixtures/ISOAB`
3. Occurs in `Isotxs_merge_Tests::test_canRemoveIsotopes` and `Isotxs_merge_Tests::test_cannotMergeXSLibWithSameNuclideNames`

It's possible that separate threads are trying to read from the same fixture resource at the same time. But that _should_ be okay. I've launched a bunch of separate threads all reading from the same file on linux and there are no errors with multiple processes reading the same file.

I updated the script to have some processes open the file for writing and some for reading. That also seems to be working, at least in terms of "the operating system does not yell at me." 

So if multiple parallel workers are trying to read from these fixtures at the same time, we _should_ be good.

If, and this is an unfounded **if**, some worker is **writing** to the fixture file while another process is reading it, **and** the fixture file is not fully complete by the time the reader process tries to read, then we could see these errors about not getting all the right bytes for a float that we see.

**But** these fixtures are tracked in the repository and _should not be edited during tests_. The tests in question create separate empty library objects and do merge actions on those empty libraries with fixture libraries.

The error traceback sure looks like we're trying to **write** a value no matter what in 

```
  File "/Users/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 329, in readWrite
    nuclideIO.rwNuclide()
  File "/Users/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 483, in rwNuclide
    self._rw7DRecord(blockNumIndex, subBlock)
  File "/Users/runner/work/armi/armi/armi/nuclearDataIO/cccc/isotxs.py", line 686, in _rw7DRecord
    dataVals.append(record.rwFloat(0.0))
                    ^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/armi/armi/armi/nuclearDataIO/cccc/cccc.py", line 381, in rwFloat
    (f,) = struct.unpack("f", self._stream.read(self._floatSize))
```

https://github.com/terrapower/armi/blob/0b3c7c1f8e88731c284fc80b7fadd27dcc6d49bb/armi/nuclearDataIO/cccc/isotxs.py#L685-L686

despite the docstring saying it's only trying to read the record

https://github.com/terrapower/armi/blob/0b3c7c1f8e88731c284fc80b7fadd27dcc6d49bb/armi/nuclearDataIO/cccc/isotxs.py#L603-L605

The files are also opened for reading in binary only during the well named `isotxs.readBinary`

## Test class set up

This might be a red herring. But the test setup for the library merging tests are strange?

https://github.com/terrapower/armi/blob/0b3c7c1f8e88731c284fc80b7fadd27dcc6d49bb/armi/nuclearDataIO/tests/test_xsLibraries.py#L299-L328

There's a `setUpClass` that unsets class attributes. So only called once per process. That's good.

But then the `setUp` method for each *instance* on the test tries to set class attributes?

armi/materials/material.py
```diff
@@ -32,6 +33,31 @@
...
+    @functools.wraps(f)
+    def inner(self: "Material", *args, **kwargs) -> float:
+        if self.parent is not None:
+            runLog.warning(
+                "Calling Material.density when attached to a Component is "
+                "undesirable and can introduce subtle differences. Prefer "
+                "directly calling the parent component's density method. "
+                f"Found on {self=} attached to {self.parent=}",
+                single=True,
```
john-science, line 53: Okay, so we have `single=True` here, so I think you want to reduce the noise on this warning. Great.

But then we are explicitly printing the ID of the parent component.

So this will print for every component someone calls `component.material.density()` on.

I think, if you're going to all the trouble to append the exact object ID, we should remove the `single=True`.  Or you could remove the ID and just do the names of the material and component class?

I don't know. But I'm not entirely convinced by the combination here.

armi/materials/material.py
```diff
@@ -32,6 +33,31 @@
...
+    """Wrap Material.density to warn people about potential problems.
+
+    If a Material is linked to a Component, ``Material.density`` may produce
+    different results from ``Component.density``. The component's density
+    is considered the source of truth because it incorporates changes in volume,
+    composition, and temperature in concert with the state of the reactor.
+    """
+
+    @functools.wraps(f)
+    def inner(self: "Material", *args, **kwargs) -> float:
```
john-science, line 46: I've never seen this before.  You can type hint a variable as a string representing the class?

I guess I expected `self: Material`.

drewj-tp: Yeah it's a forward declaration. At this point in the file, `Material` is not defined. Wrapping type hints in strings is a way to type hint something that isn't known to the runtime interpreter at that time. Same approach can be used for wrapping things under the `typing.TYPE_CHECKING` import guard

armi/materials/material.py
drewj-tp: Hmm good points. We probably want to help discover all cases where this pattern is being used, so removing `single=True` and making a lot more noise helps that. And, at least for `Component.density` that may need to call the actual `Material.density` method, we sidestep the decorator by calling `Material.density.__wrapped__` so we shouldn't flood the logs in that case.

Removing `single=True` feels like the simplest solution. Will remove

armi/nuclearDataIO/tests/test_xsLibraries.py
drewj-tp, line 1: Changes due to #1889

john-science:
Okay, we believe the strange test errors were not related to this branch. And we have fixed them in `main`.

doc/release/0.4.rst
```diff
@@ -24,7 +24,7 @@ API Changes
 #. Removing unused function ``SkippingXsGen_BuChangedLessThanTolerance``. (`PR#1845 <https://github.com/terrapower/armi/pull/1845>`_)
 #. Allow for unknown Flags when opening a DB. (`PR#1844 <https://github.com/terrapower/armi/pull/1835>`_)
 #. ``Assembly.rotatePins`` and ``Block.rotatePins`` have been removed. Prefer to use
-   :meth:`armi.reactor.assemblies.Assembly.rotate` and meth:`armi.reactor.block.Block.rotate`
+   :meth:`armi.reactor.assemblies.Assembly.rotate` and :meth:`armi.reactor.block.Block.rotate`
```
john-science, line 27:
```suggestion
  ``Assembly.rotate`` and ``Block.rotate``
```

john-science: I know this is an RST file, but we use this file in many and varied ways, and not all of those ways will be built with Sphinx and the ARMI API docs handy for auto-linking.

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -225,6 +225,9 @@ class can then use to modify the isotopics as necessary.
             constructedObject = components.factory(shape, [], kwargs)
             _setComponentFlags(constructedObject, self.flags, blueprint)
             insertDepletableNuclideKeys(constructedObject, blueprint)
+            constructedObject.p.theoreticalDensityFrac = (
```
john-science, line 228: Out of curiosity, why isn't this necessary after line 217-220 above?

drewj-tp: That logic is for a component group. It includes a recursive call to `componentDesign.construct` to build the nested component

https://github.com/terrapower/armi/blob/f0b1accb58243001b83d7f9f4c619c9dfcc84c15/armi/reactor/blueprints/componentBlueprint.py#L212-L222

So, in theory, that inner component and material could have its theoretical density modified. I'm also not sure if component groups are used in the wild. It also looks like material modifications do not work for component groups because they pass `matMods=dict()`

https://github.com/terrapower/armi/blob/f0b1accb58243001b83d7f9f4c619c9dfcc84c15/armi/reactor/blueprints/componentBlueprint.py#L217

drewj-tp:
Test cases found the warning message produced a quarter of a million log entries. Draft while I 1) reduce that, and 2) find a more helpful way to track down calls to `Material.denstiy`

drewj-tp:
@john-science @keckler https://github.com/terrapower/armi/pull/1852/commits/8d6f62464ad65577da90fdc35df55ee9038aec07 introduced a mechanism to both reduce the number of log entries and provide a way to track down offending calls
```
[warn] Found call to Material.density in /path/to/a.py at line 126. Calls to Material.density when attached to a component have the potential to induce subtle differences as Component.density and Material.density can diverge.
[warn] Found call to Material.density in /path/to/b.py at line 237. Calls to Material.density when attached to a component have the potential to induce subtle differences as Component.density and Material.density can diverge.
[warn] Found call to Material.density in /path/to/b.py at line 238. Calls to Material.density when attached to a component have the potential to induce subtle differences as Component.density and Material.density can diverge.
[warn] Found call to Material.density in /path/to/c.py at line 189. Calls to Material.density when attached to a component have the potential to induce subtle differences as Component.density and Material.density can diverge.
[warn] Found call to Material.density in /path/to/c.py at line 207. Calls to Material.density when attached to a component have the potential to induce subtle differences as Component.density and Material.density can diverge.
[warn] Found call to Material.density in /path/to/c.py at line 215. Calls to Material.density when attached to a component have the potential to induce subtle differences as Component.density and Material.density can diverge.
[warn] Found call to Material.density in /path/to/c.py at line 74. Calls to Material.density when attached to a component have the potential to induce subtle differences as Component.density and Material.density can diverge.

[info] ----- Final Warning Count --------
[info]     COUNT                LABEL

[info]      222       Found call to Material.density in /path/to/a.py at line 126
[info]      1830      Found call to Material.density in /path/to/b.py at line 237
[info]      1830      Found call to Material.density in /path/to/b.py at line 238
[info]      3498      Found call to Material.density in /path/to/b.py at line 134
[info]      3498      Found call to Material.density in /path/to/b.py at line 135
[info]     13992      Found call to Material.density in /path/to/c.py at line 74
[info]     474765     Found call to Material.density in /path/to/c.py at line 189
[info]     474765     Found call to Material.density in /path/to/c.py at line 207
[info]     474765     Found call to Material.density in /path/to/c.py at line 215

```

armi/reactor/blueprints/tests/test_materialModifications.py
```diff
@@ -54,13 +54,41 @@ class TestMaterialModifications(unittest.TestCase):
...
+
+    @staticmethod
+    def _loadAssembly(bpBase: str, materialModifications: str, assem: str):
+        yamlString = bpBase + "\n" + materialModifications
         design = blueprints.Blueprints.load(yamlString)
         design._prepConstruction(settings.Settings())
-        return design.assemblies["fuel a"]
+        return design.assemblies[assem]
+
+    def loadB4CAssembly(self, materialModifications: str):
```
john-science, line 90: I'm not convinced this method should exist. It's a one-line method, that just calls another method, and it is only used once.

john-science (approved):
Great work!

PR #1853, opened 2024-09-04
Fix the `MpiDirectoryChanger`

### LINKED ISSUE #1851: Workers do not return to original directory when using `MpiDirectoryChanger`
The [`MpiDirectoryChanger`](https://github.com/terrapower/armi/blob/9ea428bbbc196e80729e6153b2dc7e249d744aad/armi/utils/directoryChangersMpi.py#L26) is supposed to be the MPI version of the [`DirectoryChanger`](https://github.com/terrapower/armi/blob/9ea428bbbc196e80729e6153b2dc7e249d744aad/armi/utils/directoryChangers.py#L36). Using a `with` statement, all processors are supposed to enter some user-specified directory, execute code, and then exit back to the original directory.

It appears that this code is bugged. It looks like all of the workers go in to the directory, but they do not go back to the original directory afterwards. The main processor _does_ behave as expected, though.

I think that the issue is in the `close()` method:
https://github.com/terrapower/armi/blob/9ea428bbbc196e80729e6153b2dc7e249d744aad/armi/utils/directoryChangersMpi.py#L49-L58

The `open()` method has a call to `broadcast()`. I believe that the `close()` method needs to have a similar `broadcast()` statement so that the workers know where to go.

keckler:
## What is the change?

This PR fixes the `MpiDirectoryChanger` and adds a test to prove the fix. See the description in #1851 for a description of the problem. Since this new unit test is parallel, it may be necessary for you to run it locally to confirm its working (I think our GH actions do not do parallel tests).

Closes #1851 .

---

## Checklist


armi/utils/directoryChangersMpi.py
```diff
@@ -54,6 +54,7 @@ def open(self):
 
     def close(self):
         cdma = _ChangeDirectoryMpiAction(self.initial)
+        cdma = cdma.broadcast(cdma)
```
keckler, line 57: If you want to prove to yourself that this was bugged before, remove this line and run the new unit test that I put up with this PR.

john-science:
@keckler It looks really cool!

I would like to improve the testing though. Right now, the testing you added won't run in parallel by default, right? As you say in the module-level docstring, you need to run `mpiexec` on this file by hand. BUT did you know we already have a couple files in ARMI where we do that?

I think you should move your test file to `armi/tests/` and add your test file to these two GitHub Workflows:

https://github.com/terrapower/armi/blob/f9f4f693e076152529a7869bdbd007f32630c794/.github/workflows/unittests.yaml#L32-L37

https://github.com/terrapower/armi/blob/f9f4f693e076152529a7869bdbd007f32630c794/.github/workflows/coverage.yaml#L37-L41

I can also do that for you, if you need.

keckler:
> BUT did you know we already have a couple files in ARMI where we do that?

I guess I did know about the existing parallel tests, but it totally slipped my mind. I will migrate!

keckler:
@john-science I made the new test trigger during github CI. Looks like it ran successfully.

Let me know if you have anything else!

john-science (approved):
Nice!

I am happy with it, if you think this is sufficiently tested downstream that it is ready to merge.

keckler:
I have not done Nala tests using this branch. I guess I'd like to do that before we merge it into ARMI/main, because it is possible that this might lead to some changes in those results. I don't know yet. TBD.

john-science:
> I have not done Nala tests using this branch. I guess I'd like to do that before we merge it into ARMI/main, because it is possible that this might lead to some changes in those results. I don't know yet. TBD.

Holler when that's done.

keckler:
@john-science Okay, doesn't look like this impacted the internal integration tests. Please merge when it makes sense for you. Thanks!

john-science:
Oh!

Can we add your new test line to the `coverage.yaml` Workflow as well?

https://github.com/terrapower/armi/blob/f9f4f693e076152529a7869bdbd007f32630c794/.github/workflows/coverage.yaml#L37-L41

Or should I add that to your PR myself? Is that easier?

Up to you!

keckler:
Done!

PR #1859, opened 2024-09-06
Use a more robust method for determining number of pins in `HexBlock._rotatePins()`

mgjarrett:
## What is the change?

Use `block.getNumPins()` instead of counting `Flags.CLAD` components. Round this number up to a number of pins that fits exactly into a regular hex lattice.

1. Using block.getNumPins(): because counting the multiplicity of `Flags.CLAD` components on a block will not always be equal to the number of pins (e.g., a solid steel reflector slug)
2. Rounding up the number of pins: If a block is missing pin locations from a full hex lattice, we should round up the number of pins. The math for mapping the indices only works if the number of pins fits exactly into a regular hex lattice (19, 37, 61, 91, etc.)

---

## Checklist


mgjarrett:
Note that I just realized #1846 will make `rotatePins` private, so if that is merged first then I would have to update the release notes, which currently mark this as being an API change.

drewj-tp:
The `justCompute=True` behavior is something that I think could be moved off of `HexBlock.rotatePins`. The place it's used in ARMI at least is just to find post-rotation location of pins to minimize burnup

https://github.com/terrapower/armi/blob/f5267e0cc332ac9c97e8fe1c4e49e28076455399/armi/physics/fuelCycle/hexAssemblyFuelMgmtUtils.py#L105

Which feels like overkill to emulate rotating every single pin in a block just to see where one pin would end up. So I put some of the code in `armi.utils.hexagon` in https://github.com/terrapower/armi/pull/1846/commits/afd452893febe9f46f7c007ddf5a27e0aa381af9.

If there are other needs for `justCompute=True`, let me know and I can try to fold them in to #1846 or we can put them in a new ticket

mgjarrett:
Marking this as draft until #1846 is merged; this PR will be updated after that

mgjarrett:
after #1846 made `_rotatePins` private, it no longer makes sense to add the optional `numPins` argument to this function, because users won't be calling this function directly anyway.

However, I think it could still be worthwhile to make the determination of the number of pins more robust. `Block` has a `getNumPins` method that is more robust than the one-liner used.

https://github.com/terrapower/armi/blob/abd1db4adfa0c536f9bf9f08f0f1e8b74b1cd0d7/armi/reactor/blocks.py#L1047-L1079

 Is there any reason not to use `getNumPins` here?

https://github.com/terrapower/armi/blob/abd1db4adfa0c536f9bf9f08f0f1e8b74b1cd0d7/armi/reactor/blocks.py#L2141

doc/release/0.4.rst
```diff
@@ -15,6 +15,7 @@ New Features
    :func:`armi.utils.hexagon.getIndexOfRotatedCell`
    (`PR#1846 <https://github.com/terrapower/armi/1846`)
 #. Allow merging a component with zero area into another component (`PR#1858 <https://github.com/terrapower/armi/pull/1858>`_)
+#. Use ``Block.getNumPins()`` in HexBlock._rotatePins() (`PR#1859 <https://github.com/terrapower/armi/pull/1859>`_)
```
john-science, line 18:
```suggestion
#. Use ``Block.getNumPins()`` in ``HexBlock._rotatePins()``. (`PR#1859 <https://github.com/terrapower/armi/pull/1859>`_)
```

armi/reactor/blocks.py
```diff
@@ -2136,9 +2136,13 @@ def _rotatePins(self, rotNum, justCompute=False):
...
 
-        # Pin numbers start at 1. Number of pins in the block is assumed to be based on
-        # cladding count.
-        numPins = self.getNumComponents(Flags.CLAD)
+        numPins = self.getNumPins()
+        hexRings = [3 * r * (r - 1) + 1 for r in range(1, 11)]
+        for i in range(len(hexRings) - 1):
+            if numPins > hexRings[i] and numPins < hexRings[i + 1]:
+                # round numPins up
+                numPins = hexRings[i + 1]
```
drewj-tp, line 2144: So this is trying to find the number of pins in a full hex lattice in the event we have fewer pins? So if I have two rings of pins with the center spot missing, `numPins` would be six, but this logic would round up to seven for two full rings?

If that's the case, could these two functions be helpful?

https://github.com/terrapower/armi/blob/f1f3dbea05c2564404a03d8e053f72c2af15daa9/armi/utils/hexagon.py#L107-L111

https://github.com/terrapower/armi/blob/f1f3dbea05c2564404a03d8e053f72c2af15daa9/armi/utils/hexagon.py#L150-L151

like
```python
numPins = self.getNumPins()
numRings = numRingsToHoldNumCells(numPins)
fullNumPins = totalPositionsUpToRing(numRings)
```

drewj-tp (changes requested):
Can you update the PR description accordingly? The `Block.rotatePins` method has been removed, but the PR description discusses changes to it

john-science (approved):

drewj-tp (approved):

PR #1861, opened 2024-09-09
Refactor axial expansion

john-science:
## What is the change?

This PR does a refactor of the axial expansion functionality in ARMI. A new directory is structure is created to enable better traceability with testing and general maintainability. 

### Original: 

```
armi/reactor/converters/
|-- axialExpansionChanger.py
```

### New:

```
armi/reactor/converters/axialExpansionChanger
|-- __init__.py
|-- assemblyAxialLinkage.py
|-- axialExpansionChanger.py
|-- expansionData.py
```

The `axialExpansionChanger.py` file was pretty long, and there is potentially a lot of work coming up on this file. So we are trying to make the future work cleaner.

---

## Checklist


armi/reactor/converters/tests/test_axialExpansionChanger.py
albeanth, line 1: Thoughts on splitting up the tests too? Kind of like is done in #1435?

john-science: The tests are imported in a few repos downstream.

I think they make more sense being under the new module/folder structure too. But that would break downstream tests. 

Right now, downstream, the following things are imported from `armi.reactor.converters.tests.test_axialExpansionChanger`:

* `AxialExpansionTestBase`
* `buildTestAssemblyWithFakeMaterial`
* `AxialExpansionChanger`
* `getSolidComponents`
* `ExpansionData`
* `_determineLinked`

About the only high-level thing people _aren't_ importing from this test file is `AssemblyAxialLinkage`. BUT, I note with some interest, that class does not have any direct unit tests yet.

john-science: To be clear, it's not my Favorite thing that people can-and-do import ARMI's tests and testing utilities downstream, it makes things awkward. 

But I have been shot down several times when I have mentioned changing this:

1. At this point, it would require a HUGE amount of work downstream, because ARMI's API includes many widely-used testing tools.
2. I am usually the only one who thinks this "API includes testing" situation is ugly or bad.

My preference would be to have a module for ARMI testing utilities at `armi/utils/testing/`. That would be a more normal design pattern, and would allow for more freedom with testing changes.

#shrug

albeanth: > BUT that class does not have any direct unit tests yet.

au contraire mon frère! The testing class name could be a bit clearer though, admittedly. 

https://github.com/terrapower/armi/blob/f318392fd0804adc3a579c5edf6fca1df0e40d0f/armi/reactor/converters/tests/test_axialExpansionChanger.py#L947

albeanth: > My preference would be to have a module for ARMI testing utilities at armi/utils/testing/. That would be a more normal design pattern, and would allow for more freedom with testing changes.

Big Bad Breaking Change for ARMI 1.0?

albeanth: > I think they make more sense being under the new module/folder structure too. But that would break downstream tests.

Ah, ok. No worries then. We can get it in a subsequent PR.

john-science: > Big Bad Breaking Change for ARMI 1.0?

This is a big change. But if you say "ARMI 1.0"... gosh, maybe. If I had a big dev team, this would be part of our "technical debt" work. We would create `armi/utils/testing/` or whatever, and just move one thing over at a time.

It is awkward that my userbase disagrees with me on this design point. Ignoring your users needs is also a bad idea.

HMMMMMMMMMMMMMMMMMMMMMMMMMM

armi/reactor/blocks.py
```diff
@@ -1568,7 +1568,7 @@ class during axial expansion.
 
         See Also
         --------
-        armi.reactor.converters.axialExpansionChanger.py::ExpansionData::_setTargetComponents
+        armi.reactor.converters.axialExpansionChanger.expansionData.ExpansionData::_setTargetComponents
```
drewj-tp, line 1571: (Soap boxy / not your fault / pedantic nit)

Should we remove references to private methods in docs for public methods?

I understand this was here before and isn't the scope of this PR and we don't need to make this change. More starting to ring the bell of "stop using private methods or make them public and properly tested"

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -0,0 +1,310 @@
...
+            raise RuntimeError
+        if 0.0 in expFrac:
+            msg = (
+                "An expansion fraction, L1/L0, equal to 0.0, is not physical. Expansion fractions "
+                "should be greater than 0.0."
+            )
+            runLog.error(msg)
+            raise RuntimeError(msg)
+        for exp in expFrac:
+            if exp < 0.0:
```
drewj-tp, line 99: Chance for a performance bump (in a refactor so out of scope, I'm not going to make this a blocker)

Can we put the check for a zeroth valued expansion fraction in the part where we're checking for negatives? List containment already iterates over the entire list. And the point about non-physical expansion factors still applies for zero and negatively valued fractions.
Something like
```python
for exp in expFrac:
    if exp <= 0:
        msg = f"Expansion fractions must be positive. An expansion fraction of {exp} is non-physical"
        raise RuntimeError(msg)
```

doc/user/inputs.rst
```diff
@@ -765,7 +765,7 @@ A complete definition of an inner-core assembly may be seen below::
         either cold or hot dimensions. In older versions of ARMI, it was required to enter heights
         in the hot dimension (this behavior is preserved by setting `inputHeightsConsideredHot: True`).
         However, with the
-        :py:class:`axial expansion changer <armi.reactor.converters.axialExpansionChanger.AxialExpansionChanger>`,
+        :py:class:`axial expansion changer <armi.reactor.converters.axialExpansionChanger.axialExpansionChanger.AxialExpansionChanger>`,
```
drewj-tp, line 768: Similar suggestion
```suggestion
        :py:class:`axial expansion changer <armi.reactor.converters.axialExpansionChanger.AxialExpansionChanger>`,
```

armi/reactor/blocks.py
```diff
@@ -1554,7 +1554,7 @@ def setAxialExpTargetComp(self, targetComponent):
 
             Sets the ``axialExpTargetComponent`` parameter on the block to the name of the Component
             which is passed in. This is then used by the
-            :py:class:`~armi.reactor.converters.axialExpansionChanger.AxialExpansionChanger`
+            :py:class:`~armi.reactor.converters.axialExpansionChanger.axialExpansionChanger.AxialExpansionChanger`
```
drewj-tp, line 1557: Since these classes are imported into `armi.reactor.converters.axialExpansionChanger`, does this doc work?
```suggestion
            :py:class:`~armi.reactor.converters.axialExpansionChanger.AxialExpansionChanger`
```
The triplet `axialExpansionChanger.axialExpansionChanger.AxialExpansionChanger` is throwing me off

drewj-tp (approved):
Thanks @john-science I don't have any major suggestions here. Some nits that I'm not going to hold as blockers for this PR

armi/reactor/blocks.py
john-science: That's an interesting point. Yeah, I'm going to do that.

john-science: Done

armi/reactor/converters/axialExpansionChanger/expansionData.py
john-science: @albeanth He just wants to add an equal sign. And that seems fine to me. But is there some situation in which we have zero expansion input here?  Does that ever happen as a matter of book-keeping, such that this change would break things?

drewj-tp: Not so much *just* adding an equal sign, but combining the
```python
if 0.0 in expFrac:
    # error with non-physical zero expansion
for exp in expFrac:
    if exp < 0:
        # error with non-physical negative expansion
```
with the snippet above. We want to error for both zero and negative expansion factors, and the error message in both cases indicates the non-physicality of a non-positive expansion factor.

armi/reactor/blocks.py
albeanth: > The triplet axialExpansionChanger.axialExpansionChanger.AxialExpansionChanger is throwing me off

🤢

PR #1870, opened 2024-09-12
Providing plugin hook getAxialExpansionChanger

### LINKED ISSUE #1843: Plugin hook for initial axial expansion
There isn't an easy way to override the initial axial expansion performed on assemblies at construction. Providing a mechanism like this would be useful.

I propose a plugin hook like
```python
@staticmethod
@HOOKSPEC
def initialAssemblyAxialExpansion(assems: list[Assembly], cs: Settings) -> bool:
    pass
```
where
- `assems` is an iterable of all assemblies in the `core`, or rather all assemblies being constructed at this invocation, and
- `cs` is the case settings for this run

We don't want to perform multiple expansions, so when iterating over the hooks, we would cease iteration after the first call returned true. An example of this would be if the inputs heights considered hot flag is true, the default implementation would not perform expansion at construction and return false.

The default implementation would call this section of code

https://github.com/terrapower/armi/blob/6fd89e1b095c6290195967674758df0c64434caa/armi/reactor/blueprints/__init__.py#L314-L332

and be marked with the [`trylast`](https://pluggy.readthedocs.io/en/stable/index.html#call-time-order) argument. This would allow other plugins to potentially perform their expansion earlier, and then return `True` to signify no other expansion should take place.

drewj-tp:
## What is the change?

Plugins are now able to implement the `getAxialExpansionChanger` hook to provide their own class responsible for
performing the initial thermal axial expansion of assemblies. The default implemented hook, brought by the `ReactorPlugin`, exposes the `armi.reactor.converters.axialExpansion.AxialExpansionChanger` class.

This allows plugins to bring their own machinery for thermal axial expansion during the blueprints construction stage. Plugins can even opt to make a subclass that does not perform axial expansion at all.

Closes #1843 

## Todo

- [x] Release notes (need pr number to edit release notes)
- [x] Testing (need to prove only the first such class is returned)
  - Local dev experience indicates this is the case, but tests will be needed to prove this and prevent future problems 

---

## Checklist


armi/reactor/blueprints/__init__.py
```diff
@@ -326,7 +326,8 @@ def _prepConstruction(self, cs):
                     for a in list(self.assemblies.values())
                     if not any(a.hasFlags(f) for f in assemsToSkip)
                 )
-                axialExpansionChanger.axialExpansionChanger.expandColdDimsToHot(
+                chngrT = getPluginManager().hook.getAxialExpansionChanger()
```
albeanth, line 329: This will likely expose my lack of understanding about the inner workings of the plugins systems, but what happens if an Application (a "Plugin" in ARMI terms) has Plugins (also "Plugin" in ARMI terms) that have multiple axialExpansionChanger subclasses? How would those get resolved? 

E.g.: 
ARMI_application:
- Plugin A
- Plugin B (has axialExpansionChangerForPluginB)
- Plugin C
- Plugin D (has axialExpansionChangerForPluginD)

drewj-tp: Good question. The call order for plugins is first in first out. So PluginB would have its class returned first.

My though process was that ARMI would provide a sensible default expander, and then a single plugin for an application would provide that application's expander. You could also envision a world where an application brings in an `AxialExpansionInterface` (e.g., for things like `interactEveryNode` through an `AxialExpansionPlugin` and that plugin would provide this hook as well.

drewj-tp: @albeanth added this to the hook docstring

        This hook should only be provided by one additional plugin in your application. Otherwise
        the `order of hook execution <https://pluggy.readthedocs.io/en/stable/index.html#call-time-order>`_
        may not provide the behavior you expect.

https://github.com/terrapower/armi/pull/1870/commits/9dbbca288fa024d87da69bd5731b58045ad9d174

armi/plugins.py
```diff
@@ -665,6 +666,48 @@ def defineSystemBuilders() -> Dict[str, Callable[[str], "Composite"]]:
...
+        Will be used during initial construction of the core and assemblies, and
+        can be a standard class for performing all the thermally driven axial
+        expansion.
+
+        The first object returned that is not ``None`` will be used.
+        Plugins are encouraged to add the ``tryfirst=True`` arguments to their
+        ``HOOKIMPL`` invocations to make sure their specific are earlier in the
+        hook call sequence. The default reactor plugin
+        :class:`armi.reactor.ReactorPlugin` is marked ``trylast=True`` to help
+        put other implementations earlier in the sequence.
```
albeanth, line 684: Does this need to be updated given the App guidance you added in 9dbbca288fa024d87da69bd5731b58045ad9d174?

armi/tests/test_plugins.py
```diff
@@ -42,13 +43,25 @@ def defineFlags():
...
+class SillyAxialExpansionChanger(AxialExpansionChanger):
+    pass
+
+
+class SillyAxialPlugin(plugins.ArmiPlugin):
+    @staticmethod
+    @plugins.HOOKIMPL
+    def getAxialExpansionChanger() -> type[SillyAxialExpansionChanger]:
+        return SillyAxialExpansionChanger
+
```
albeanth, line 55: Can you add some comments to the above for posterity? I could totally see myself coming back here in 6+ months and being like "what the heck is this again....?" 

E.g., within `SillyAxialExpansionChanger` something like "a test subclass that an ARMI Application could create"

and then within `SillyAxialPlugin` something like "the plugin hook that an ARMI Application would have to register the `SillyAxialExpansionChanger` subclass"

armi/tests/test_plugins.py
```diff
@@ -42,13 +43,25 @@ def defineFlags():
...
 class TestPluginRegistration(unittest.TestCase):
     def setUp(self):
         """
         Manipulate the standard App. We can't just configure our own, since the
         pytest environment bleeds between tests.
         """
-        self._backupApp = deepcopy(getApp())
+        self.app = getApp()
+        self._backupApp = deepcopy(self.app)
```
albeanth, line 64: what's `self._backupApp` for?

armi/tests/test_plugins.py
```diff
@@ -92,6 +105,15 @@ def test_defineFlags(self):
         # show the flag exists now
         self.assertEqual(type(Flags.SUPER_FLAG._value), int)
 
+    def test_axialExpansionHook(self):
+        """Test that plugins can override the axial expansion of assemblies via a hook."""
+        pm = self.app.pluginManager
+        first = pm.hook.getAxialExpansionChanger()
+        self.assertIs(first, AxialExpansionChanger)
+        pm.register(SillyAxialPlugin)
```
albeanth, line 113: same posterity comment here: can you add an inline comment at line 113 saying something like "now register the Plugin (ARMI App) with the subclass. the axial expansion changer class pulled from the plugin system is the `SillyAxialExpansionChanger` instead of the default in ARMI"

albeanth (commented):
This looks great to me. Just some feedback on the comments for clarity and posterity/long-term maintenance.

armi/tests/test_plugins.py
drewj-tp: Yeah this was a weird one. `getApp` returns the global armi application. The intent of these tests is to modify the app so we can register plugins and look for behavior changes. But that global app, and any changes to it, would carry over from test to test. So these tests attempt to restore the global app state when they are done

armi/plugins.py
drewj-tp: I'm inclined to say no because those notes are a few lines lower in the `Notes` section of this docstring. If you think more docs are needed, or that paragraph needs to be higher up in the docs, let me know or make an in-line suggestion

armi/tests/test_plugins.py
drewj-tp: Sure!

armi/tests/test_plugins.py
drewj-tp: Sure thing!

armi/tests/test_plugins.py
drewj-tp: b1270c93

armi/tests/test_plugins.py
drewj-tp: b1270c93

armi/tests/test_plugins.py
albeanth: Tagging in @john-science as this is getting out of my scope of practice

john-science: This was just someone's attempt to find a way to test plugins.  The problem is that to TEST plugin tooling, you are going to want to register a bunch of plugins.  But if you do that, it will stomp all over every unit test, because they all share a default `App()`.

So this was an attempt to have the plugin infrastructure tests be isolated from stomping all over every other unit tests in ARMI.

You should absolutely not make the change you are making here. Leave it as it is, please.

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
```diff
@@ -148,6 +83,71 @@ def __init__(self, detailedAxialExpansion: bool = False):
...
+        Calling this method will result in an increase in mass via applyColdHeightMassIncrease!
+
+        See Also
+        --------
+        :py:meth:`armi.reactor.converters.axialExpansionChanger.axialExpansionChanger.AxialExpansionChanger.applyColdHeightMassIncrease`
+        """
+        assems = list(assems)
+        if not referenceAssembly:
+            referenceAssembly = getDefaultReferenceAssem(assems)
+        axialExpChanger = cls(isDetailedAxialExpansion)
```
john-science, line 136: Why is this line necessary?

Is there some internal code that will be making use of this?

armi/tests/test_plugins.py
john-science: Okay, I think your change doesn't break anything.

But it certainly makes understanding the already complicated test logic harder.  I don't love that.

doc/release/0.4.rst
```diff
@@ -12,6 +12,8 @@ New Features
 #. Removing the ``tabulate`` dependency by ingesting it to ``armi.utils.tabulate``. (`PR#1811 <https://github.com/terrapower/armi/pull/1811>`_)
 #. Adding ``--skip-inspection`` flag to ``CompareCases`` CLI. (`PR#1842 <https://github.com/terrapower/armi/pull/1842>`_)
 #. Allow merging a component with zero area into another component (`PR#1858 <https://github.com/terrapower/armi/pull/1858>`_)
+#. Plugins can provide the ``getAxialExpansionChanger`` hook to customize axial expansion
```
john-science, line 15:
```suggestion
#. Plugins can provide the ``getAxialExpansionChanger`` hook to customize axial expansion.
```

armi/plugins.py
albeanth: resolved with https://github.com/terrapower/armi/pull/1870/commits/bb01240d1324a01f94df41a2acff9c15b730bb2f

john-science:
I only have two small things on this PR.  Take a look.

So... the purpose of this design is two-fold?

1. Developers can now drop in their own `AxialExpansionChanger` subclasses.
2. Developers can now opt to not have an `AxialExpansionChanger` changer at all, by just not having this plugin hook in any of their plugins?

Do I have the goals right?

IF I do, I guess I have some questions:

1. What happens if you use this plugin hook in more than one plugin?
    - I assume this has to fail, right?
    - Have you tested that?
    - How does it fail?
2. Does ARMI run if zero plugins implement this hook?
    - I want to make sure this change doesn't break for reactors with no assemblies.

armi/tests/test_plugins.py
albeanth: @john-science I'm going to put the resolution of this in your court, if you don't mind!

drewj-tp: > Okay, I think your change doesn't break anything.
> 
> But it certainly makes understanding the already complicated test logic harder. I don't love that.

Happy to revert. Tried to remove the number of `deepcopy` and `getApp` calls. But I agree mucking around with these global app states is not ideal. There's probably a way we can test against a generic application without the global state. That should probably be a separate ticket though, but it would make these tests cleaner and more maintainable

drewj-tp (commented):
> I only have two small things on this PR. Take a look.
> 
> So... the purpose of this design is two-fold?
> 
> 1. Developers can now drop in their own `AxialExpansionChanger` subclasses.
> 2. Developers can now opt to not have an `AxialExpansionChanger` changer at all, by just not having this plugin hook in any of their plugins?
> 
> Do I have the goals right?

Goal 1 is correct. Any application can drop in their own `AxialExpansionChanger` subclass.

Goal 2 is slightly incorrect. The default `ReactorPlugin` implements this hook that brings the current `AxialExpansionChanger`.  If an application does not register any plugin that implement this hook, the "default" state is to bring the existing expansion changer into the mix. To put it another way, this plugin hook is always implemented, but applications can circumvent the default expander by implementing the hook.

And, if somehow you do not register the `ReactorPlugin` and find yourself in a situation where *no* plugin registered implements this hook, I have no idea what pluggy will do. It may return `None` and then get some attribute errors trying to call `None.expandColdDimsToHot` as if it were an expander class?

> IF I do, I guess I have some questions:
> 
> 1. What happens if you use this plugin hook in more than one plugin?
>    
>    * I assume this has to fail, right?
>    * Have you tested that?
>    * How does it fail?

@albeanth had this [same question](https://github.com/terrapower/armi/pull/1870#discussion_r1757392947). If multiple plugins implement this hook, the first non-`None` result will be "selected" and used. The goal is there should be at most two plugins that implement this hook and registered with the application

1. Default `ReactorPlugin.getAxialExpanderClass`
2. Some other plugin deemed responsible of owning axial expansion.

The `ReactorPlugin.getAxialExpanderClass` hook implementation is marked with `trylast=True` so, regardless of the plugin registration ordering, it should be the last such hook to be called.

The test I added covers two plugins implementing this hook and showing the behavior of the `trylast` marker.

1. No new plugin -> call plugin hooks -> default `AxialExpansionChanger` fetched from `ReactorPlugin`
2. Register a different plugin that implements the hook -> call plugin hooks hooks -> expansion changer from second plugin is registered

https://github.com/terrapower/armi/blob/b1270c93cb4417f64b313e1678d294a6884539dd/armi/tests/test_plugins.py#L110-L120

> 2. Does ARMI run if zero plugins implement this hook?
>    
>    * I want to make sure this change doesn't break for reactors with no assemblies.

One plugin already implements and registers this hook - `ReactorPlugin.getAxialExpanderClass`. And that performs the current behavior. So reactors with no assemblies should not see any changes if they do not implement this hook.

However, such applications could implement this hook on a plugin that produces a do-nothing expander class. That is (potentially, untested) way to skip axial expansion beyond setting `Tinput=Thot`

armi/reactor/blueprints/__init__.py
john-science, line 329:
```suggestion
                chngrT = getPluginManagerOrFail().hook.getAxialExpansionChanger()
```

armi/reactor/blueprints/__init__.py
```diff
@@ -326,7 +326,8 @@ def _prepConstruction(self, cs):
...
+                chngrT.expandColdDimsToHot(
```
john-science, line 330: Okay, so the `pluggy` docs don't provide any particular tools for failing if no plugin implements a hook. So we are clearly supposed to handle that ourselves.

I guess I would say if `chngrT` (strange variable name) is None, the code shouldn't fail, or throw a warning. I would prefer it if this line was wrapped in an `if`, and the model was allowed to continue without failure.

Or would this cause a failure elsewhere in the run?

drewj-tp: Yeah the docs don't have anything on that case. But a test file shows if nothing implements the hook, in this case we just get `None` returned

<details>

<summary>Demo no hook impl</summary>

```python
"""Play around with plugins"""

import pluggy

spec = pluggy.HookspecMarker("demo")
impl = pluggy.HookimplMarker("demo")

class MySpec:
    @staticmethod
    @spec(firstresult=True)
    def get_number() -> float:
        """Give me a number, any number."""

pm = pluggy.PluginManager("demo")
pm.add_hookspecs(MySpec)
result = pm.hook.get_number()
assert result is None, result
```
</details>

I'll add a condition if we get `None` in the event an application that inherits from `armi.apps.App` un-registers the `ReactorPlugin` and also does not provide their own implementation of this hook.

I do not expect this to cause any later failures. This would just result in not doing any axial expansion when construction assemblies from blueprints.

drewj-tp: 86be86c7

john-science: Sorry, I may not be clear.

I don't want people who have reactors without assemblies to HAVE to implement this hook.  That is a waste of their time.

So, I would prefer if people who were modeling liquid-fuel reactors (for instance) to not have to understand this took, and they can just skip it.

So, all we have to do here is something like:

```python
                chngrT = getPluginManager().hook.getAxialExpansionChanger()
                if chngrT is not None:
                    chngrT.expandColdDimsToHot(
```

john-science: I'm sorry if I haven't been clear.  But since not all reactors _have_ assemblies that would apply to this hook, I expect to in the near future the default `ReactorPlugin` probably won't implement this hook. That will be something we move elsewhere, to make this feature invisibly optional.

This is another case of ARMI assuming that everyone is modeling "pin-type reactors with hex-shaped assemblies".  The default functionality of ARMI should not be so specific, though it was in many places, and still is. So we have been making a concerted effort to remove such assumptions.

Imagine in the future that ARMI has a default `App`, but also a `HexAssemlyApp`. And the `HexAssemblyApp` pulls in a ton of details like this one. It might also pull in a new `HexAssemblyCore` base class for people to subclass.

Maybe the real solution won't look like that exactly, but I have heard that ARMI's many assumptions about reactor type causes non-TerraPower users a lot of wasted effort. So we are working on that. (Actually, I heard that from _you_ Drew. You are one of the people who told me that. I would love to pick your brain about what ARMI's assumptions wasted your team's time in the past.)

drewj-tp: The changes you requested were added in 86be86c7c01f272ff78d02f334e22f18a6b59f40

https://github.com/terrapower/armi/blob/86be86c7c01f272ff78d02f334e22f18a6b59f40/armi/reactor/blueprints/__init__.py#L329-L334

I support these changes for the reasons you laid out: fewer assumptions about the reactor someone is modeling the better.

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
drewj-tp: I don't think I follow the question. I can try to explain the intent and maybe that will help.

The axial expansion specific behavior that existed in blueprints construction was a call to the `expandColdDimsToHot` function. Inside that function, it created an `AxialExpansionChanger` instance to perform the initial expansion of assemblies. The drawback is there is no way to alter _what_ performs the axial expansion inside that `expandColdDimsToHot` let alone modify the function call there.

By making this a classmethod on the base axial expander class and allowing plugins to provide their own expander subclass **if they wish**, we can provide more control to plugins and applications about how the expansion on blueprints is performed. If at all.

I could do something like
```python
class SillyAxialExpander(AxialExpansionChanger):
    def axiallyExpandAssembly(self):
        # do my own expansion on an assembly
```
or
```python
class NoOpExpander(AxialExpansionChanger):
    @classmethod
    def expandColdDimsToHot(cls, ...):
        # actually don't do anything
        return
```
Now if my plugin implements this hook and returns one of these classes, I have exercised my control over how to perform the initial expansion of assemblies. And potentially not at all in the case of the more explicit `NoOpExpander`

Note: Changes in https://github.com/terrapower/armi/commit/86be86c7c01f272ff78d02f334e22f18a6b59f40 mean allow the case where no plugin implements this hook to also not do any expansion.

john-science: Okay, sure, I get that.

Thanks!

armi/tests/test_plugins.py
```diff
@@ -68,7 +83,7 @@ def test_defineFlags(self):
             :id: T_ARMI_PLUGIN_REGISTER
             :tests: R_ARMI_PLUGIN
         """
-        app = getApp()
+        app = self.app
```
john-science, line 86:
```suggestion
        app = getApp()
```

john-science (approved):
After CI passes, I think this is ready for show time, if you guys do.

albeanth:
> After CI passes, I think this is ready for show time, if you guys do.

This is apart of a larger downstream slough of changes, so let's hold tight till those are ready?

albeanth (approved):

john-science (approved):

PR #1877, opened 2024-09-13
Rotating more parameters in HexBlock.rotate

### LINKED ISSUE #1860: Rotate more parameters during HexBlock.rotate
A follow on from #1533 

when `Block.rotate` is called, the only thing it really does are
- [ ] Updates the `orientation` parameter according to the rotation
- [ ] Updates the `pinLocation` parameter providing "1-d spiral-based" location of each pin
- [ ] Updates the `displacementX` and `displacementY` parameters
  - (Aside: why does rotation change block/assembly displacement?)

Any parameters stored on blocks or their pin-like components that should be rotated are not. We should rotate *some* parameters so the data layout matches the orientation of the block

## What to rotate

The motivation for a block being rotated is its parent assembly is being rotated. Presently, ARMI rotates assemblies via a fuel handler module, e.g., `armi.physics.fuelCycle.assemblyRotationAlgorithms.buReducingAssemblyRotation`

https://github.com/terrapower/armi/blob/f5267e0cc332ac9c97e8fe1c4e49e28076455399/armi/physics/fuelCycle/assemblyRotationAlgorithms.py#L50-L52

In nuclear reactors, assemblies are moved or shuffled during reloading outages. The logic above emulates the rotation of assemblies, potentially during a shuffling operation, but still during an outage. One would expect a physics solve to occur **after** rotation, and before someone does something with the assemblies (e.g., plugins accessing the data model). Data like pin fluxes, linear pin power, etc. would be overwritten prior to being called after rotation. However, data like composition, displacements per atom (DPA), are cumulative and would need to be handled such that the data for the rotated pins rotates correctly.

I think we will need some logic to identify parameters on blocks and components that are rotatable and *should* be rotated. My current plan is to leverage `Category.pinQuantities` and `Categories.cumulative` and develop a way to rotate those data arrays.

drewj-tp:
## What is the change?

Calls to `HexBlock.rotate` will additionally rotate parameters on the block with the `ParamLocation.CHILDREN` tag.

Rotating a block previously did not update all of the relevant parameters.

Closes #1860 

---

## Checklist


drewj-tp:
cc @albeanth @onufer
who else should peek at this?

drewj-tp:
Tests are failing because we have a test that does `HexAssembly.rotate` and then does extensive testing on parameters on the hex block. These are redundant if we also have extensive testing on `HexBlock.rotate`.

I'm of the opinion those should be moved to the block tests. Then we just need to make sure that `HexAssembly.rotate` calls `HexBlock.rotate`

albeanth:
> cc @albeanth @onufer who else should peek at this?

Maybe @alexhjames since he wrote some of the rotation stuff, I believe

albeanth:
> Tests are failing because we have a test that does `HexAssembly.rotate` and then does extensive testing on parameters on the hex block. These are redundant if we also have extensive testing on `HexBlock.rotate`.
> 
> I'm of the opinion those should be moved to the block tests. Then we just need to make sure that `HexAssembly.rotate` calls `HexBlock.rotate`

I haven't look at the tests in question, but if they are repetitive and can be moved, then I think it's probably fair to do.

onufer:
i agree with this. it makes no sense to rotate rates (flux/fast flux/dpa rate). When you rotate an assembly the solution changes so the rates need to be recalculated (and we do that). Integrated params (fluence, fast fluence, dpa) should be rotated and we intend to. Thanks!

alexhjames:
> i agree with this. it makes no sense to rotate rates (flux/fast flux/dpa rate). When you rotate an assembly the solution changes so the rates need to be recalculated (and we do that). Integrated params (fluence, fast fluence, dpa) should be rotated and we intend to. Thanks!

Not sure I agree with this. The rotate method gets called both during shuffling, in which case your comment makes sense, but also during `growToFullCore` where it doesn't. The assemblies that are out of the first third need ALL spatially varying parameters rotated as well. 

maybe we should have two methods, one for growing to full core and one for shuffling? Like a
`rotate` vs a `reflect`?

armi/reactor/blocks.py
```diff
@@ -2022,6 +2025,22 @@ def rotate(self, rad):
             in 60-degree increments (i.e., PI/6, PI/3, PI, 2 * PI/3, 5 * PI/6,
             and 2 * PI)
 
+        Notes
+        -----
+        Parameters marked with the ``rotatable``
+        :class:`~armi.reactor.parameters.parameterDefinitions.Category` will be rotated.
+        These parameters must adhere to one of the following rules to be rotated:
```
alexhjames, line 2032: Missing vector valued quantities like `displacementX` and `displacementY` where two block parameters form a vector. Alternatively, you could restructure `displacementX` and `displacmentY` to be a single parameter that follows these rules.

drewj-tp:
Summarizing a separate chat.

During rotation from shuffling and moving assemblies, some parameters will be wrong because they don't match the state of the reactor. And that's acceptable for now.

I'll make a separate ticket to capture that

drewj-tp:
> maybe we should have two methods, one for growing to full core and one for shuffling? Like a rotate vs a reflect?

If we were only rotating certain parameters during shuffling-based `HexBlock.rotate` and doing a lot more in `growToFullCore`, this could be a good strategy. My intention now is to rotate everything that looks rotatable in `HexBlock.rotate` so the reflect method may not be needed now.

armi/reactor/tests/test_blocks.py
```diff
@@ -1742,21 +1742,46 @@ def test_getReactionRates(self):
...
+    *. ``2`` (ring, pos) = (2, 2)
+    *. ...
+    *. ``7`` (ring, pos) = (2, 6)
+    *. ``8`` (ring, pos) = (3, 1)
+    *. ...
+
+    When we rotate a block a single 60 degree rotation counter clock wise,
+    the data that was originally at (r, p) = (2, 1) is now at (r, p) = (2, 2)
+    and should be at index ``2`` in the post-rotated array.
+
```
john-science, line 1776:
```suggestion
    When we rotate a HexBlock a single 60 degree rotation counter clock wise,
    the data that was originally at (r, p) = (2, 1) is now at (r, p) = (2, 2)
    and should be at index ``2`` in the post-rotated array.
```

armi/reactor/tests/test_blocks.py
```diff
@@ -1804,14 +1829,31 @@ def test_rotateBoundaryParameters(self):
...
+        Pre-rotate pin layout -> Post-rotate layout
+
+              2  1      1  6
+            3  0  6 -> 2  0  5
+              4  5      3  4
+
+        - Pre-rotate data: ``[0, 1, 2, 3, 4, 5, 6, ...]``
+        - Post-rotate data: ``[0, 6, 1, 2, 3, 4, 5, ...]``
+
```
john-science, line 1848:
```suggestion
        Pre-rotate pin layout -> Post-rotate layout::

              2  1      1  6
            3  0  6 -> 2  0  5
              4  5      3  4

        - Pre-rotate data: ``[0, 1, 2, 3, 4, 5, 6, ...]``
        - Post-rotate data: ``[0, 6, 1, 2, 3, 4, 5, ...]``
```

john-science: I don't think your diagram will render in RST without the `::`.  Tell me if I'm wrong.

doc/release/0.4.rst
```diff
@@ -15,8 +15,6 @@ New Features
    :func:`armi.utils.hexagon.getIndexOfRotatedCell`
    (`PR#1846 <https://github.com/terrapower/armi/1846`)
 #. Allow merging a component with zero area into another component (`PR#1858 <https://github.com/terrapower/armi/pull/1858>`_)
-#. :meth:`armi.reactor.blocks.HexBlock.rotate` only rotates parameters that are marked with
```
john-science: I can see the logic in removing this release note. Good catch!

Are you going to add your own release note?

armi/reactor/blocks.py
```diff
@@ -2192,6 +2175,23 @@ def _rotatePins(self, rotNum, justCompute=False):
...
+            name = param.name
+            original = self.p[name]
+            if isinstance(original, (list, np.ndarray)):
+                pass
+            elif isinstance(original, (int, float)):
+                pass
+            elif original is None:
+                pass
+            else:
+                raise TypeError(f"{name=} :: {original=}")
```
john-science, line 2193:
```suggestion
            if isinstance(original, (list, np.ndarray, int, float, None)):
                pass
            elif original is not None:
                raise TypeError(f"{name=} :: {original=}")
```

john-science: Not a demand. Just see what you think.

drewj-tp: Sorry this is very much draft still, not ready to go. Yes that's a better way to do this. But I need to add the special logic inside the list/array check to do the rotation

doc/release/0.4.rst
drewj-tp: Yes

armi/reactor/tests/test_blocks.py
drewj-tp: Yeah you're right. I wasn't sure we included test docstrings in the docs but we do so I'll fix it

armi/reactor/blocks.py
drewj-tp: Yes that was not called out in the docs. I've revamped the PR and kept the existing rotation routines (displacement, orientation) so those should be unimpacted

doc/release/0.4.rst
drewj-tp: Added

drewj-tp:
okay @HunterPSmith @john-science @alexhjames @onufer this feels much better now.

Existing rotation is not touched. Duct params, orientation, and displacement should be rotated as they were before.

Any block parameter with the `ParamLocation.CHILDREN` will be rotated, assuming it has enough entries to fill a full hexagonal lattice. So if you have three hex rings of things, you should provide an array with 19 items. 

A couple cases where this could be a problem:
1. You have empty sites so there are no pins at certain locations.
2. You have intermixed pin types (e.g., fuel and coolant) on different lattice sites

A concrete example of where this could be a problem would be the MHTGR 350 benchmark (#224). The assemblies have fuel compacts and coolant channels on the same hexagonal lattice. Users would have to make sure the things that are populating `Block.p["linPowByPin"]` and `Block.p["percentBuByPin"]` provide empty data the coolant sites.

If we want to remove that assumption, we need to find a way to pass through the pin positions. Maybe with the spatial grid locations? But that's not clear to me right now.

doc/release/0.4.rst
```diff
@@ -11,8 +11,12 @@ New Features
 #. ARMI now supports Python 3.12. (`PR#1813 <https://github.com/terrapower/armi/pull/1813>`_)
 #. Removing the ``tabulate`` dependency by ingesting it to ``armi.utils.tabulate``. (`PR#1811 <https://github.com/terrapower/armi/pull/1811>`_)
 #. Adding ``--skip-inspection`` flag to ``CompareCases`` CLI. (`PR#1842 <https://github.com/terrapower/armi/pull/1842>`_)
-#. Provide utilities for determining location of a rotated object in a hexagonal lattice (``getIndexOfRotatedCell``). (`PR#1846 <https://github.com/terrapower/armi/1846`)
-#. Allow merging a component with zero area into another component. (`PR#1858 <https://github.com/terrapower/armi/pull/1858>`_)
+#. Provide utilities for determining location of a rotated object in a hexagonal lattice:
+   :func:`armi.utils.hexagon.getIndexOfRotatedCell`
+   (`PR#1846 <https://github.com/terrapower/armi/1846`)
```
john-science, line 16:
```suggestion
#. Provide utilities for determining location of a rotated object in a hexagonal lattice: ``getIndexOfRotatedCell``. (`PR#1846 <https://github.com/terrapower/armi/1846`_)
```

john-science: Minor, and unimportant cleanup:

1. This text is taken and rendered downstream without access to the API (for QA docs), so the RST links to API docs won't work.
2. It needs that `_` at the end of the URL for RST to rend the URL correctly.
3. Added a period to the end.

armi/reactor/blocks.py
```diff
@@ -2027,6 +2027,11 @@ def rotate(self, rad):
         self._rotatePins(rotNum)
         self._rotateBoundaryParameters(rotNum)
         self._rotateDisplacement(rad)
+        self._rotatePinParameters(rotNum)
+
+    def _getParamsWhere(self, f: Callable[[parameters.Parameter], bool]):
```
john-science, line 2032: Two questions:

1. Do you want this here, or could this be a public method on `Composite` on `Parameter`?
    - I just notice there is already `Composite.getParamValuesAtZ()`.
    - I also notice that there is already a `Parameter.atLocation()`.
    - My THOUGHT is this looks generally useful, beyond `Block`s.
    - I'm okay being wrong, just wondering.
2. Should we have a unit test or two for this new method?

armi/reactor/blocks.py
```diff
@@ -2167,6 +2175,25 @@ def _rotatePins(self, rotNum, justCompute=False):
 
         return rotateIndexLookup
 
+    def _rotatePinParameters(self, rotNum: int):
```
john-science, line 2178: Could use a one-line comment. WE had to talk about this a lot, so there must be something interesting to say about it.

Up to you though; not a deal breaker.

john-science:
Love the tests, man!

armi/reactor/tests/test_blocks.py
```diff
@@ -1766,6 +1735,126 @@ def test_getReactionRates(self):
...
+    """Tests for the ability for a block to be rotated."""
+
+    BOUNDARY_PARAMS = [
+        "pointsCornerDpa",
+        "cornerFastFlux",
+        "pointsCornerFastFluxFr",
+        "pointsCornerDpaRate",
+        "pointsEdgeFastFluxFr",
+        "pointsEdgeDpa",
+        "pointsEdgeDpaRate",
```
john-science, line 1748:
```suggestion
        "cornerFastFlux",
        "pointsCornerDpa",
        "pointsCornerDpaRate",
        "pointsCornerFastFluxFr",
        "pointsEdgeDpa",
        "pointsEdgeDpaRate",
        "pointsEdgeFastFluxFr",
```

john-science: I just alphabetized because I couldn't tell if there were any duplicates, and I was too lazy to check the slow way.

doc/release/0.4.rst
drewj-tp: Yeah you had mentioned that elsewhere and I think it got lost in resolving merge conflicts

armi/reactor/blocks.py
drewj-tp: All very good thoughts. I slapped this here because I needed it on `HexBlock` and couldn't find a good spot elsewhere.

It might make the most sense to be placed onto the `ParameterCollection` so it's auto assigned to everything that has parameters. And that makes sense: the thing that collects parameters allows you to do iteration over them.

I'll pull something into a separate PR because know I know we have a need for it, let's pull the discussion there, and then more "users who need things rotated" discussions can happen in this PR.

And yes, I will add tests now that it's being escalated to a public thing

Cool?

armi/reactor/blocks.py
drewj-tp: Good with me

armi/reactor/parameters/parameterDefinitions.py
```diff
@@ -55,38 +55,33 @@
...
+    """Parameters are accumulated over many time steps"""
     cumulativeOverCycle = "cumulative over cycle"
+    """Parameters that are reset at beginning of cycle and accumulated over each cycle."""
     assignInBlueprints = "assign in blueprints"
+    """Parameters that should be assigned in blueprints (e.g., control rod elevation)"""
     retainOnReplacement = "retain on replacement"
     pinQuantities = "pinQuantities"
+    """Parameters are defined on the pin level within a block"""
     fluxQuantities = "fluxQuantities"
+    """Parameters are related to neutron or gamma flux"""
```
drewj-tp, line 72: (This came up in a chat so wanted to put my motivation here)

The change here is to make these docstrings attached to the class constants. They were already documented in the `Category` docstring (some of them at least) but I found two flaws

1. They were documented in the `Notes` section but would make more sense to be properly documented as [class attributes](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring)
2. Attaching the docstring to the object itself means tools like python, ipython, vscode, etc. can provide context where you are about the thing you're looking at. You don't need to separately pull up the ARMI docs.

I'm not going to die on this hill if people have stronger opinions. This is just my rationale for this change.

armi/reactor/blocks.py
john-science: That sounds great, man. I see [your ticket](https://github.com/terrapower/armi/issues/1898). Thanks for self-assigning it.

armi/reactor/tests/test_blocks.py
```diff
@@ -1766,6 +1735,126 @@ def test_getReactionRates(self):
         )
 
 
+class BlockRotateTests(unittest.TestCase):
+    """Tests for the ability for a block to be rotated."""
+
+    BOUNDARY_PARAMS = [
+        "cornerFastFlux",
```
alexhjames, line 1742: missing other parameters like the `THcornTemp` and `THedgeTemp`

armi/reactor/tests/test_blocks.py
```diff
@@ -1766,6 +1735,126 @@ def test_getReactionRates(self):
         )
 
 
+class BlockRotateTests(unittest.TestCase):
+    """Tests for the ability for a block to be rotated."""
+
+    BOUNDARY_PARAMS = [
```
alexhjames, line 1741: Would it be better to put all this class attributes in a `setupClass` method?

armi/reactor/tests/test_blocks.py
```diff
@@ -1766,6 +1735,126 @@ def test_getReactionRates(self):
...
+        "cornerFastFlux",
+        "pointsCornerDpa",
+        "pointsCornerDpaRate",
+        "pointsCornerFastFluxFr",
+        "pointsEdgeDpa",
+        "pointsEdgeDpaRate",
+        "pointsEdgeFastFluxFr",
+    ]
+    BOUNDARY_DATA = np.arange(6, dtype=float) * 10
```
alexhjames, line 1750: Does the rotate method work on things that aren't numpy arrays? Like a standard python list? We probably should work on both data types and test that. We had an error early on where Mongoose was adding the data to the TH parameters as numpy objects and that caused errors.

alexhjames (changes requested):

armi/reactor/blocks.py
drewj-tp: `HexBlock.rotate` docs heavily updated in https://github.com/terrapower/armi/pull/1877/commits/28d1a353f425f8f9e6b32c42d69e5d19afade001

This specific method doc'ed in 2ddc312de9eacafed79224af42c5ef3651e0e115

armi/reactor/tests/test_blocks.py
alexhjames: I don't see how that test is related to the boundary data rotation.

drewj-tp: My fault. While the comment is on the boundary data attribute, the call out to "rotate method" focused me elsewhere. I'll modify this test to rotate some boundary params as lists

drewj-tp: How does 8b81b3e2da7d4920eb52844bcf91a5c854a0d8be look?

alexhjames: I'm fine with the content of the test. Honestly, it took me a minute to parse what you were doing in `_assignParamData`. It might be worth elaborating on the comment to be like even index parameters in the the names list get an array and odd index parameters in the names list get a list. Also it would help to rename `referenceData` to `referenceArrayData` and `listData` to `referenceListData`.

alexhjames (approved):

armi/utils/hexagon.py
```diff
@@ -191,3 +194,76 @@ def getIndexOfRotatedCell(initialCellIndex: int, orientationNumber: int) -> int:
...
+    """
+    if not isinstance(data, (list, np.ndarray)):
+        raise TypeError(f"{data=}")
+    cells = np.size(data, axis=0)
+    if cells == 0:
+        # Nothing to rotate
+        return data
+    nRings = numRingsToHoldNumCells(cells)
+    nCellsInFullRings = totalPositionsUpToRing(nRings)
+    if cells != nCellsInFullRings:
```
mgjarrett, line 248: Is there a way we can support this? We will have some blocks where the pins don't fill a full hex ring, but the rings do have an explicit position defined by a grid in the blueprints.

Would the onus be on the user/developer to cast this data into a full hexagon before trying to rotate, and then map it back after rotating? Maybe there is a way to automate that mapping.

armi/reactor/blocks.py
```diff
@@ -2167,6 +2197,26 @@ def _rotatePins(self, rotNum, justCompute=False):
...
+        """Rotate parameters assigned on pins."""
+        params = self.p.where(lambda pd: pd.atLocation(ParamLocation.CHILDREN))
```
mgjarrett, line 2202: It seems like we are fundamentally redefining the way pin-level data is stored in ARMI here.

Up to this point, parameter data that exists at `ParamLocation.CHILDREN` has always been associated with pin indices. For some parameter defined on pins, e.g., `linPowByPin`, the data never gets reshuffled after a rotation; only `pinLocation` gets rotated. `linyPowByPin[n]` will always contain the data for the `n`th pin. If you want to know where the `n`th pin is currently located, you would use `_rotatePins` and/or the `pinLocation` parameter to find out.

For example, see the method `setPinPowers`:

https://github.com/terrapower/armi/blob/96a5c88d6167cee989378599b6d341c351c291bd/armi/reactor/blocks.py#L1983-L1990

An array of data is passed in that represents a distribution of pin powers indexed by pin location. The `pinLocation` is mapped to a `pinNum` before being stored in the `linPowByPin*` parameter, using `pinNum` as an index.

If we start rotating `pinLocation` and also `linPowByPin*` and other pin-level parameters, then there's a risk of double-counting rotations.

Fundamentally, I think any parameter that is `ParamLocation.CHILDREN` should not have any information about how much the block has been rotated baked into it (other than `pinLocation`). The rotations should be tracked by the `pinLocation` parameter, while the other parameters remain invariant to rotation.

drewj-tp: Okay, I've been digging through the code base and internal projects to better understand rotation. What @mgjarrett brings up in

> For some parameter defined on pins, e.g., linPowByPin, the data never gets reshuffled after a rotation; only pinLocation gets rotated. linyPowByPin[n] will always contain the data for the nth pin. If you want to know where the nth pin is currently located, you would use _rotatePins and/or the pinLocation parameter to find out

matches what I'm seeing. I would probably lean towards calling `pinLocation` something like `pinIndex` because it is _really_ a mapping from values in an (potentially unrotated) data array where that pin's data should actually live in parameters but that's not the point.

This brings up a weird point/implication. Things calling `setPinPowers` and, similarly, [`Block.setPinMgFluxes`](https://github.com/terrapower/armi/blob/f1f3dbea05c2564404a03d8e053f72c2af15daa9/armi/reactor/blocks.py#L377-L383) don't use any of the spatial information from the block. Data are rotated in the `setPin*` calls. By further extension, any pin-like parameter would need a similar setter, e.g., `setPercentBuByPin`. 

If they attempt to pull spatial information from [`Block.getPinCoords`](https://github.com/terrapower/armi/blob/f1f3dbea05c2564404a03d8e053f72c2af15daa9/armi/reactor/blocks.py#L1589-L1595), that reflects the unrotated state. Which is probably something that should be rotated so the location of `pin[n]` found in `Block.getPinCoords()[n]` reflects where that pin exists in the block before and after rotation.

But, doing that would likely mean we have another double rotation problem that @mgjarrett pointed out
1. Pin locations from `getPinCoords` are rotated.
2. Data obtained using the current pin locations.
3. Data are "re-rotated" mapping through `b.p.pinLocation`

## Decisions
1. Do we want to update the spatial location of pins through rotation as reported by `Block.getPinCoords`?
   - I'm leaning towards yes because you'd expect XY coordinates to change through rotation
 2. Do we want block spatial parameters to reflect the current orientation of the block (like this PR proposed), or the un-rotated state of the block (like we currently do)
   - I'm leaning towards yes partially because sunk cost fallacy (I've written code to do that already) but it also feels odd that you could rotate a block, look at some parameter data, and not _really_ know where that data exists in the block
   - Docs would help
   - We'd also need logic to handle parameter rotation when pin-like data are set for every pin-like parameter

PR #1881, opened 2024-09-16
Renaming Reactor.moveList to Reactor.moves

john-science:
## What is the change?

I am renaming `Reactor.moveList` to `Reactor.moves`. 

(There is only one downstream project that I can find that might be affected by this change.)

The attribute is actually a `dict` and not a `list`. And this is one of the main reasons we don't inject the variable type into the variable name when we write code.

---

## Checklist


opotowsky (approved):
Easy PR to approve. Since the downstream updates are queued up, go ahead and merge when you feel this is appropriate

PR #1882, opened 2024-09-16
Implementing more detailed `__lt__` for components.

### LINKED ISSUE #1875: Implement more detailed `__lt__` for component sorting
Components on a block are sorted by their `getBoundingCircleOuterDiameter`.

https://github.com/terrapower/armi/blob/abd1db4adfa0c536f9bf9f08f0f1e8b74b1cd0d7/armi/reactor/components/component.py#L272-L288

Sometimes, two components have the same outer diameter, in which case the sorting done by `__lt__` will depend on the order of comparison. i.e., 

`a < b` evaluates to `False`
`b < a` also evaluates to `False`

As a result, 

`sorted([a, b]) = [a, b]`
while
`sorted([b, a]) = [b, a]`

Ideally, the outcome of `sorted` will not be dependent on the order of the input.

We should fall back to `getCircleInnerDiameter` for comparison when sorting components when `getBoundingCircleOuterDiameter` is equivalent. I pushed an example of what I think the sorting should do here:

https://github.com/terrapower/armi/compare/main...component_sort

mgjarrett:
## What is the change?

Fall back to `getCircleInnerDiameter` for comparison when sorting components when `getBoundingCircleOuterDiameter` is equivalent.

Components on a block are sorted by their `getBoundingCircleOuterDiameter`. Sometimes, two components have the same outer diameter, in which case the sorting done by `__lt__` will depend on the order of comparison. i.e., 

`a < b` evaluates to `False`
`b < a` also evaluates to `False`
`b == a` evaluates to `True`, even though the two components are not necessarily identical, they only have one identical dimension.

`sorted([a, b]) != sorted([b, a])` even though `b` and `a` are not identical.

Close #1875

---

## Checklist


armi/reactor/blocks.py
```diff
@@ -18,29 +18,31 @@
...
+    composites,
+    geometry,
+    grids,
+    parameters,
+)
+from armi.reactor.components.basicShapes import Circle, Hexagon, Rectangle
+from armi.reactor.components.complexShapes import Helix, HoledHexagon
+
+# this unused import is required so that the RadialSegment component type is registered
+from armi.reactor.components.volumetricShapes import RadialSegment
```
drewj-tp: Does this mean radial segments are not nominally registered for use in blueprints?

armi/reactor/__init__.py
```diff
@@ -51,7 +51,7 @@
 See :doc:`/developer/index`.
 """
 
-from typing import Dict, Callable, Union, TYPE_CHECKING
+from typing import TYPE_CHECKING, Callable, Dict, Union
```
drewj-tp, line 54: I have no strong opinions on the sorting of imports. Should we standardize the practice across armi (e.g., a ruff plugin?) so that we don't have a lot of these unrelated changes in some PRs?

Could be a useful addition to our dev guide if there are vs code settings that will do this automatically too

cc @john-science @opotowsky @albeanth

armi/reactor/components/__init__.py
```diff
@@ -32,10 +32,11 @@
 import numpy as np
 
 from armi import runLog
-from armi.reactor.components.component import *  # noqa: F403
-from armi.reactor.components.basicShapes import *  # noqa: F403
-from armi.reactor.components.complexShapes import *  # noqa: F403
-from armi.reactor.components.volumetricShapes import *  # noqa: F403
```
drewj-tp, line 38: Love the removal of a `*` import 👍

armi/reactor/components/__init__.py
```diff
@@ -186,6 +187,23 @@ def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
...
+        Tc : float
+            Ignored for this component
+        cold : bool, optional
+            If True, compute the area with as-input dimensions, instead of thermally-expanded.
+
+        Notes
+        -----
+        Tc is not used in this method for this particular component.
+        """
+        return 0.0
```
drewj-tp, line 205: This seems like an okay thing for a shaped component to have to define on their own. Maybe a `DerivedShape` is the only exception.

Saying this because we `try`/`catch` a `NotImplementedError` below so if  `Component.getCircleInnerDiameter` is required for components, we'd be okay in that context.

But that could be a larger change that is likely better served outside this PR.

Non-blocking comment. If there's traction I can make a ticket

armi/reactor/components/volumetricShapes.py
```diff
@@ -250,7 +254,11 @@ def getComponentVolume(self):
         return vol
 
     def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
-        return self.getDimension("outer_radius", Tc, cold)
+        return 2.0 * self.getDimension("outer_radius", Tc, cold)
```
drewj-tp, line 257: Oops... is this method only used for sorting?

armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -42,7 +41,7 @@ def __init__(self, value):
         # note: yamlizable does not call an __init__ method, instead it uses __new__ and setattr
         self.value = value
         if isinstance(value, str):
-            if not components.COMPONENT_LINK_REGEX.search(value):
+            if not component.COMPONENT_LINK_REGEX.search(value):
```
drewj-tp, line 44: Is this regex defined in `armi.reactor.components.component` and then imported into `armi.reactor.components`?

This is not a change I would have expected to see in this PR

armi/reactor/components/component.py
```diff
@@ -278,14 +273,26 @@ def __lt__(self, other):
...
-            raise ValueError(
-                "Components 1 ({} with OD {}) and 2 ({} and OD {}) cannot be ordered because their "
-                "bounding circle outer diameters are not comparable.".format(
-                    self, thisOD, other, thatOD
+            if thisOD == thatOD:
+                thisID = self.getCircleInnerDiameter(cold=True)
+                thatID = other.getCircleInnerDiameter(cold=True)
+                return thisID < thatID
+            else:
+                return thisOD < thatOD
```
drewj-tp, line 281: Can you add this behavior to the `__lt__` docstring?

armi/reactor/tests/test_components.py
```diff
@@ -508,6 +513,50 @@ def test_getAreaColdTrue(self):
         self.assertAlmostEqual(totalAreaCold, totalAreaHot, delta=1e-10)
 
 
+class TestComponentList(unittest.TestCase):
+    def setUp(self):
+        self.componentList = []
```
drewj-tp, line 518: Generally not advisable to embed variable types in their names. See #1881 

This `TestCase` is responsible for testing the sorting of various components
```suggestion
class TestComponentSort(unittest.TestCase):
    def setUp(self):
        self.components = []
```

drewj-tp (commented):
Of the 21 files changed here, maybe three of them are actually related to the heart of this PR and touch the `__lt__` changes.

They're not bad changes per se, but they are nearly entirely import related changes: sorting of imports, breaking multiple imports, importing just the module not the things needed in the module. They just added extraneous diffs to this PR. 

What can we collectively do to avoid this? There's probably some linter / formatting config we can add, maybe some "blessed" vs code related settings we can advertise. But, these changes make it hard to agree that "this PR has one idea or purpose" with the extra, unrelated changes

mgjarrett:
> Of the 21 files changed here, maybe three of them are actually related to the heart of this PR and touch the `__lt__` changes.
> 
> They're not bad changes per se, but they are nearly entirely import related changes: sorting of imports, breaking multiple imports, importing just the module not the things needed in the module. They just added extraneous diffs to this PR.
> 
> What can we collectively do to avoid this? There's probably some linter / formatting config we can add, maybe some "blessed" vs code related settings we can advertise. But, these changes make it hard to agree that "this PR has one idea or purpose" with the extra, unrelated changes

Sorry, I'm in the middle of working on this. The root of the problem is that I ran into a circular import, and all of the import changes are related to fixing the circular import. It started small and then I went down a rabbit hole -- I'll probably break those changes out into a separate PR now.

armi/reactor/components/__init__.py
mgjarrett: I moved this to a separate PR.

armi/reactor/__init__.py
mgjarrett: Yeah VSCode is automatically sorting imports alphabetically for me, I haven't figured out how to turn it off yet. According to VSCode, it's doing this according to `ruff` rules, but I don't see it happening for anybody else so I must be doing something wrong.

armi/reactor/tests/test_components.py
```diff
@@ -508,6 +513,50 @@ def test_getAreaColdTrue(self):
...
+        hexDims = {"Tinput": 273.0, "Thot": 273.0, "op": 2.6, "ip": 2.0, "mult": 1.0}
+        pinComp = Circle("pin", "UZr", Tinput=273.0, Thot=273.0, od=0.08, mult=169.0)
+        gapComp = Circle(
+            "gap", "Sodium", Tinput=273.0, Thot=273.0, id=0.08, od=0.08, mult=169.0
+        )
+        ductComp = Hexagon("duct", "HT9", **hexDims)
```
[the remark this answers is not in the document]
mgjarrett: I just copy/pasted it from test_blocks.py, but I agree there doesn't seem to be any reason to use kwargs

armi/reactor/components/__init__.py
mgjarrett: The assumption for the unshaped component is to return the smallest possible bounding circle outer diameter, i.e., assume the component has a disk shape. It seems consistent with that assumption to assume that the inner diameter is 0.0.

The exception is caught in `__lt__`, but only for the purpose of producing a clearer error message. The error is still raised and the code crashes. So we still need to be able to sort `UnshapedComponent`s.

armi/reactor/components/__init__.py
```diff
@@ -186,6 +186,23 @@ def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
...
+    def getCircleInnerDiameter(self, Tc=None, cold=False):
+        """
+        Component is unshaped; assume it is circular and there is no ID (return 0.0).
+
+        Parameters
+        ----------
+        Tc : float
+            Ignored for this component
+        cold : bool, optional
+            If True, compute the area with as-input dimensions, instead of thermally-expanded.
```
john-science, line 198:
```suggestion
        Tc : float, optional
            Ignored for this component
        cold : bool, optional
            Ignored for this component
```

armi/reactor/components/basicShapes.py
```diff
@@ -20,8 +20,7 @@
 """
 import math
 
-from armi.reactor.components import ShapedComponent
-from armi.reactor.components import componentParameters
+from armi.reactor.components.component import ShapedComponent, componentParameters
```
john-science, line 23: Mike, can we fix your IDE?

I don't particularly like these changes, and they make your PRs more muddled to read. Like... you didn't actually touch this file.

#sorry

doc/release/0.4.rst
```diff
@@ -15,6 +15,7 @@ New Features
    :func:`armi.utils.hexagon.getIndexOfRotatedCell`
    (`PR#1846 <https://github.com/terrapower/armi/1846`)
 #. Allow merging a component with zero area into another component (`PR#1858 <https://github.com/terrapower/armi/pull/1858>`_)
+#. Use inner diameter for sorting components when outer diameter is identical (`PR#1882 <https://github.com/terrapower/armi/pull/1882>`_)
```
john-science, line 18:
```suggestion
#. Allow merging a component with zero area into another component. (`PR#1858 <https://github.com/terrapower/armi/pull/1858>`_)
#. Use inner diameter for sorting components when outer diameter is identical. (`PR#1882 <https://github.com/terrapower/armi/pull/1882>`_)
```

john-science: Arrielle pointed out to me last release that these periods greatly improve how the release docs render.

So I'm just adding them back in. NBD

armi/reactor/__init__.py
mgjarrett: Okay I was able to turn it off, so no more rampant import sorting

armi/reactor/components/basicShapes.py
mgjarrett: Sure thing. I was able to turn it off, so I'll go through and revert these import changes.

armi/reactor/components/volumetricShapes.py
mgjarrett: I think only for sorting, yes

doc/release/0.5.rst
```diff
@@ -46,6 +46,7 @@ New Features
 #. New plugin hook ``beforeReactorConstruction`` to process settings before reactor init. (`PR#1945 <https://github.com/terrapower/armi/pull/1945>`_)
 #. Improving performance in the lattice physics interface by not updating cross sections at ``everyNode`` during coupled calculations. (`PR#1963 <https://github.com/terrapower/armi/pull/1963>`_)
 #. Allow merging a component with zero area into another component. (`PR#1858 <https://github.com/terrapower/armi/pull/1858>`_)
+#. Use inner diameter for sorting components when outer diameter is identical. (`PR#1882 <https://github.com/terrapower/armi/pull/1882>`_)
```
john-science, line 49: ARMI 0.5.0 was released.

Can we move this release note up to the `0.5.1` release?

Thanks!

john-science:
@mgjarrett I believe I see a merge conflict.

This PR removes the file: `doc/release/0.4.rst`.  We don't want to lose that file.

Thanks!

armi/reactor/components/__init__.py
```diff
@@ -186,6 +186,23 @@ def getBoundingCircleOuterDiameter(self, Tc=None, cold=False):
...
+        Parameters
+        ----------
+        Tc : float, optional
+            Ignored for this component
+        cold : bool, optional
+            Ignored for this component
+
+        Notes
+        -----
+        Tc is not used in this method for this particular component.
```
john-science, line 202:
```suggestion
```

john-science: That's an optional argument in this method, we don't need to call it out a second time.

armi/reactor/components/component.py
```diff
@@ -271,19 +271,34 @@ def __lt__(self, other):
...
+        except (NotImplementedError, Exception) as e:
+            if isinstance(e, NotImplementedError):
+                raise NotImplementedError(
+                    "getCircleInnerDiameter not implemented for at least one of {}, {}".format(
+                        self, other
+                    )
```
john-science, line 293:
```suggestion
                    f"getCircleInnerDiameter not implemented for at least one of {self}, {other}"
```

john-science: I won't die on this hill.  I am just interested in moving toward more f-strings.

john-science:
@mgjarrett Just wondering why this is "Draft". It looks close to done to me.

armi/reactor/components/component.py
opotowsky: (I'm team f-string too)

mgjarrett:
> @mgjarrett Just wondering why this is "Draft". It looks close to done to me.

This is essentially done, but it was paired with a downstream plugin update that was in a holding pattern. This can be merged now, I will take it out of draft.

PR #1887, opened 2024-09-16
Improving efficiency of reaction rate calcs

mgjarrett:
## What is the change?

Improve the efficiency of reaction rate calculations.

A common post-processing step for a `GlobalFluxInterface` is to compute reaction rates from the flux solution and the microscopic cross sections that exist on the reactor model. The implementation of this calculation is done in a straightforward way, but it is naive with-respect-to the typical data structure in an ARMI simulation. The efficiency for typical problems can be greatly improved by refactoring the loops so that the outermost loop is over cross section type.

Often, there are thousands of blocks in a reactor model but only a handful of unique microscopic cross section types; each set of microscopic cross sections might be shared across several hundred blocks. An outer iteration over cross section types with an inner iteration over blocks can significantly reduce the number of times a microscopic cross section set needs to be retrieved.

---

## Checklist


john-science:
Just a head's up, I know you're not done. But this seems like an SCR situation.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -1313,6 +1313,82 @@ def computeDpaRate(mgFlux, dpaXs):
     return dpaPerSecond
 
 
+def calcReactionRatesBlockList(objList, keff, xsNucDict):
+    r"""
+    Compute 1-group reaction rates for the objcects in objList (usually a block).
```
keckler, line 1318:
```suggestion
    Compute 1-group reaction rates for the objects in objList (usually a block).
```

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -1313,6 +1313,82 @@ def computeDpaRate(mgFlux, dpaXs):
...
+
+    .. impl:: Return the reaction rates for a given ArmiObject
+        :id: I_ARMI_FLUX_RX_RATES_BY_XS_ID
+        :implements: R_ARMI_FLUX_RX_RATES
+
+        This is an alternative implementation of :need:`I_ARMI_FLUX_RX_RATES` that
+        is more efficient when computing reaction rates for a large set of blocks
+        that share a common set of microscopic cross sections.
+
+        Fore more detail on the reation rate calculations, see :need:`I_ARMI_FLUX_RX_RATES`.
```
keckler, line 1342:
```suggestion
        For more detail on the reation rate calculations, see :need:`I_ARMI_FLUX_RX_RATES`.
```

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -1313,6 +1313,82 @@ def computeDpaRate(mgFlux, dpaXs):
...
+        The list of objects to compute reaction rates on. Notionally this could be upgraded to be
+        any kind of ArmiObject but with params defined as they are it currently is only
+        implemented for a block.
+
+    keff : float
+        The keff of the core. This is required to get the neutron production rate correct
+        via the neutron balance statement (since nuSigF has a 1/keff term).
+
+    xsNucDict: Dict[str, XSNuclide]
+        Microscopic cross sections to use in computing the reaction rates.
```
keckler, line 1332: Can you add context to this parameter description of what the keys should be (i.e. what they _mean_ as compared to just their type)?

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -1313,6 +1313,82 @@ def computeDpaRate(mgFlux, dpaXs):
...
+
+        This is an alternative implementation of :need:`I_ARMI_FLUX_RX_RATES` that
+        is more efficient when computing reaction rates for a large set of blocks
+        that share a common set of microscopic cross sections.
+
+        Fore more detail on the reation rate calculations, see :need:`I_ARMI_FLUX_RX_RATES`.
+    """
+    for obj in objList:
+        rate = {}
+        for simple in RX_PARAM_NAMES:
```
keckler, line 1346: `simple`?

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -1313,6 +1313,82 @@ def computeDpaRate(mgFlux, dpaXs):
...
+            rate[simple] = 0.0
+
+        numberDensities = obj.getNumberDensities()
+        try:
+            mgFlux = np.array(obj.getMgFlux())
+        except TypeError:
+            continue
```
keckler, line 1353: Should we allow for this function to silently continue if an object without flux is passed in? The explicit purpose of this function is to calculate reaction rates, so passing an object without flux is just a fundamentally wrong premise.

A hard error feels appropriate to me here, unless there is a common case that I am not aware of.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -1313,6 +1313,82 @@ def computeDpaRate(mgFlux, dpaXs):
...
+
+        for nucName, numberDensity in numberDensities.items():
+            if numberDensity == 0.0:
+                continue
+            nucrate = {}
```
keckler, line 1358: Suggest changing to `nucRate`

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -1313,6 +1313,82 @@ def computeDpaRate(mgFlux, dpaXs):
...
+            nucrate["rateProdN2n"] += 2.0 * numberDensity * mgFlux.dot(micros.n2n)
+
+            for simple in RX_PARAM_NAMES:
+                if nucrate[simple]:
+                    rate[simple] += nucrate[simple]
+
+        for paramName, val in rate.items():
+            obj.p[paramName] = val  # put in #/cm^3/s
+
+        vFuel = obj.getComponentAreaFrac(Flags.FUEL) if rate["rateFis"] > 0.0 else 1.0
```
keckler, line 1387: I understand that you copied this line from the `calcReactionRates()` function, but what the heck is going on here?

I assume `vFuel` is meant to mean "volume of fuel", but instead of calculating volume it gets the area fraction. In the case of axially-extruded blocks, _maybe_ ...sure...

But if the fission rate is 0 then we somehow want the volume of fuel to be 1.0? Huh? I do not understand what is happening here.

keckler: I guess it is just so that we don't do a `0/0` on the next line?

It seems like this little chunk of code could be more clear.

armi/reactor/converters/uniformMesh.py
```diff
@@ -1100,13 +1101,45 @@ def _mapStateFromReactorToOther(
...
+    @staticmethod
+    def _calculateReactionRatesEfficient(core, keff):
+        """
+        First, sort blocks into groups by XS type. Then, we just need to grab micros for each XS type once.
+
+        Iterate over list of blocks with the given XS type; calculate reaction rates for these blocks
+        """
+        from armi.physics.neutronics.globalFlux.globalFluxInterface import (
+            calcReactionRatesBlockList,
+        )
```
keckler, line 1127: My major question here is if it is worthwhile to have the `calcReactionRatesBlockList()` method as a part of the API. It feels quite specific to me, and I'm not sure if it is worth being public.

If we instead acknowledged the limited use cases of `calcReactionRatesBlockList()` by maybe making it private, we could clean things up in a few ways:
1. You can remove this import within a method
2. You could remove a number of `if` statements that are there to catch edge cases, which would speed things up even more
3. You could stop passing around so many arguments in function calls

My feeling is that it could just be a lot cleaner overall, and that adding `calcReactionRatesBlockList()` to the API won't benefit hardly anybody.

keckler (approved):
I think the PR probably does what you want it to do, but just added some suggestions to maybe make things more clear.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
mgjarrett: This is copy/paste from `calcReactionRates`. It's very old, I have no idea where this variable name originated.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
mgjarrett: This is more copy/paste:

image

Based on the comment, it seems like the thinking goes that if an object without flux is passed in, you can just skip calculating the reaction rates on that object. You don't have to crash the whole run.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
mgjarrett: My guess is that somebody was just going for the minimum number of lines of code to get the algebraically correct result.

I could make it more clear with more lines. Something like this:

image

armi/reactor/converters/uniformMesh.py
mgjarrett: I can move `calcReactionRatesBlockList` to uniformMesh.py (the only place it is used) and then make it private, but I don't immediately see how that's going to help with (2) and (3). I'll push up what I have and maybe you can point out the opportunities for simplification.

armi/reactor/converters/uniformMesh.py
```diff
@@ -1100,13 +1101,45 @@ def _mapStateFromReactorToOther(
...
+        for b in core.getBlocks():
+            xsTypeGroups[b.getMicroSuffix()].append(b)
+
+        for xsID, blockList in xsTypeGroups.items():
+            nucSet = set()
+            for b in blockList:
+                nucSet.update(
+                    nuc for nuc, ndens in b.getNumberDensities().items() if ndens > 0.0
+                )
+            nucList = sorted(nucSet)
```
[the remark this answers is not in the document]
mgjarrett: Not necessary

armi/physics/neutronics/globalFlux/globalFluxInterface.py
keckler: Looks like you already did it. Great!

armi/reactor/converters/uniformMesh.py
```diff
@@ -1130,6 +1160,89 @@ def _calculateReactionRates(lib, keff, assem):
...
+
+            numberDensities = obj.getNumberDensities()
+            try:
+                mgFlux = np.array(obj.getMgFlux())
+            except TypeError:
+                continue
+
+            for nucName, numberDensity in numberDensities.items():
+                if numberDensity == 0.0:
+                    continue
```
keckler, line 1210: Instead of looping over `numberDensities` and then checking that the number density is non-zero, would it make more sense to loop over the nuclides in `xsNucDict`? You've already done work to screen out isotopes with zero number density from there, so the number of isotopes in there is less and you don't have to do any further checking for zero values.

I might be misinterpreting this, I'm finding the logic somewhat complex.

armi/reactor/converters/uniformMesh.py
```diff
@@ -1130,6 +1160,89 @@ def _calculateReactionRates(lib, keff, assem):
...
+
+        .. impl:: Return the reaction rates for a given ArmiObject
+            :id: I_ARMI_FLUX_RX_RATES_BY_XS_ID
+            :implements: R_ARMI_FLUX_RX_RATES
+
+            This is an alternative implementation of :need:`I_ARMI_FLUX_RX_RATES` that
+            is more efficient when computing reaction rates for a large set of blocks
+            that share a common set of microscopic cross sections.
+
+            For more detail on the reation rate calculations, see :need:`I_ARMI_FLUX_RX_RATES`.
```
keckler, line 1192: This `impl` tag will not be seen because this method is private... That's my fault.

I think you can use the :meta public: directive like this: https://github.com/terrapower/armi/blob/119842ef7f959b5a5e50dcd7ac85a19d693ac708/armi/materials/mixture.py#L24

armi/reactor/converters/uniformMesh.py
keckler: I made some suggestions in comments. I might be off base, but hopefully they make sense.

I also wonder if the abstraction from blocks to `objList` is worth anything. I guess maybe, I don't know. Not a big deal.

armi/reactor/converters/uniformMesh.py
mgjarrett: `xsNucDict` will have an entry for any nuclide that exists at a non-zero number density in any block` in the `blockList`.

This line is checking whether the nuclide has a nonzero density in a particular block or other object.

armi/reactor/converters/uniformMesh.py
```diff
@@ -1130,6 +1160,89 @@ def _calculateReactionRates(lib, keff, assem):
...
+        """
+        from armi.physics.neutronics.globalFlux.globalFluxInterface import (
+            RX_PARAM_NAMES,
+            RX_ABS_MICRO_LABELS,
+        )
```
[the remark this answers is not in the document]
mgjarrett: Yeah, the import isn't great here but I can't import at the top of the file because of a circular import.

We could explicitly redefine `RX_PARAM_NAMES` and `RX_ABS_MICRO_LABELS` here, but then there's a risk that those strings get changed over in `globalFluxInterface` but not here.

armi/reactor/converters/uniformMesh.py
keckler: True, maybe the check cannot be removed. But I still think that looping over the other list might make more sense, no? `xsNucDict` is shorter without loss of needed information.

keckler (approved):
With those last couple comments, make any changes if you so desire. But I am good to approve this as-is. Thank you for this work, it will be a big improvement!

armi/physics/neutronics/globalFlux/globalFluxInterface.py
mgjarrett: I don't know whether we're ever actually hitting this exception. It might be good to remove it if possible, but that could probably be a separate PR where it is also removed from the original function I copied it out of.

armi/reactor/converters/uniformMesh.py
mgjarrett: `xsNucDict` is not necessarily shorter than `obj.getNumberDensities()`. You could have a list of 500 blocks where 495 of them have 3 nuclides and 5 of them have 400 nuclides, and `xsNucDict` would have 400 nuclides. This is obviously an extreme situation that I've devised here that isn't realistic, but it's just hard to say that one of these lists is always going to be shorter than the other one.

If you refactored, it would be something like:

```
for nucName in xsNucDict:
    numberDensity = numberDensities.get(nucName, 0.0)
    if numberDensity == 0.0:
        continue
```

I think since you always have to get the dict value from `numberDensities`, it makes sense to iterate over `numberDensities.items()` instead of iterating over the keys and then doing `numberDensities.get(key, 0.0)`. In general, in a group of blocks, if _any_ of the blocks have a non-zero number density for a given `nucName`, then you'd have to iterate through that nuclide for every block and run `numberDensities.get(nucName, 0.0)` to check if that nuclide is there.

armi/reactor/converters/uniformMesh.py
```diff
@@ -1130,6 +1161,86 @@ def _calculateReactionRates(lib, keff, assem):
...
+            nuclide names (e.g., "U235") and values are the associated XSNuclide objects
+            from the cross section library, which contain the microscopic cross section
+            data for a given nuclide in the current cross section group.
+
+        .. impl:: Return the reaction rates for a given ArmiObject
+            :id: I_ARMI_FLUX_RX_RATES_BY_XS_ID
+            :implements: R_ARMI_FLUX_RX_RATES
+
+            This is an alternative implementation of :need:`I_ARMI_FLUX_RX_RATES` that
+            is more efficient when computing reaction rates for a large set of blocks
```
john-science, line 1192: This won't work.

If you put the `.. impl` tag under the `Parameters` header, the docs build won't find the `impl` tag. 

In RST, this will just look like a "tool tip" under the `xsNucDict` parameter.

The fix is to move this ABOVE the Parameters section.

john-science (approved):
Approved, barring one change.

PR #1889, opened 2024-09-17
Making the xsLibrary tests more thread safe

john-science:
## What is the change?

I have changed some tests to read in some data files by thread, rather than by class.

We believe this will make these tests more thread safe.

---

## Checklist

### DIFF
--- a/armi/nuclearDataIO/tests/test_xsLibraries.py
+++ b/armi/nuclearDataIO/tests/test_xsLibraries.py
@@ -53,8 +53,8 @@ DLAYXS_MCC3 = os.path.join(FIXTURE_DIR_CCCC, "mc2v3.dlayxs")
 UFG_FLUX_EDIT = os.path.join(FIXTURE_DIR, "mc2v3-AA.flux_ufg")
 
 
-class TempFileMixin(unittest.TestCase):
-    """really a test case."""
+class TempFileMixin:
+    """Not a test; just helpful test tooling."""
 
     def setUp(self):
         self.td = TemporaryDirectoryChanger()
@@ -71,7 +71,7 @@ class TempFileMixin(unittest.TestCase):
         )
 
 
-class TestXSLibrary(TempFileMixin):
+class TestXSLibrary(TempFileMixin, unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.isotxsAA = isotxs.readBinary(ISOTXS_AA)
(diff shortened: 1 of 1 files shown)


armi/nuclearDataIO/tests/test_xsLibraries.py
```diff
@@ -546,4 +493,4 @@ def test_mergeAllXSLibFiles(self):
 
 
 # Remove the abstract class, so that it does not run (all tests would fail)
-del TestXSlibraryMerging
```
ntouran, line 549: now that you renamed it, isn't there's no reason to delete it down here anymore so you can remove these lines. The test discovery finds all classes that start with `Test` in their name. Hmm maybe not b/c other test classes in here don't start with `Test`...

ntouran (commented):
Loving it. I was thinking of adding the xdist `_testMethodName` thing (see https://github.com/terrapower/armi/commit/ba309d0f83ab901a4b4594f750f47eef0b86b143) but I think your solution is simpler and cleaner and should also solve the same problem. By loading libs directly onto the test case instance before each individual test, there should be no opportunity for different threads to clobber each other, even though they're reading from the same file. They'll all have independent file handles and store results in independent memory. should work!

drewj-tp (approved):
This is a worthwhile change, regardless of if/how it impacts #1852

Thanks @john-science and @ntouran!

armi/nuclearDataIO/tests/test_xsLibraries.py
john-science: Unfortunately, I tried deleting this line, and I get test failures:

```bash
  File "C:\Users\jstilley\codes\repos\nala\framework\armi\nuclearDataIO\tests\test_xsLibraries.py", line 303, in setUp
    self.libAA = self.getReadFunc()(self.getLibAAPath())
  File "C:\Users\jstilley\codes\repos\nala\framework\armi\nuclearDataIO\tests\test_xsLibraries.py", line 312, in getReadFunc
    raise NotImplementedError()
NotImplementedError
```

I guess `pytest` finds all classes that subclass `unittest.TestCase`, it doesn't just go by the name of the class.

Which is strange, because I'm pretty sure it _does_ find unit test methods by any name starting with `test`.  That's a strange asymmetry.

drewj-tp: > pytest will automatically collect unittest.TestCase subclasses and their test methods in test_*.py or *_test.py files

https://docs.pytest.org/en/7.1.x/how-to/unittest.html

ntouran: ah ok thanks for the reminder. so its the module name not the class name. I stand corrected.

drewj-tp:
I've re-ran #1852 tests 5+ times all morning and no failure! 🎉

Issue #1871, opened 2024-09-12
ARMI should not forcibly add a SFP to the Reactor


john-science:
Right now, ARMI forcibly adds a Spent Fuel Pool (SFP) to every `Reactor` object:

https://github.com/terrapower/armi/blob/5a41b25fe9f262acb03347923443c4a2808837db/armi/reactor/reactors.py#L213-L214

I think this is rather a lazy idea and default, meant to temporarily help out a small number of users downstream who had blueprints that didn't have SFPs in them.

I think it's time to pull the Band-Aid™ off, and make people define their own SFP in the blueprints.

And, not for nothing, I just hate the default values that are in our default SFP:

https://github.com/terrapower/armi/blob/5a41b25fe9f262acb03347923443c4a2808837db/armi/reactor/blueprints/__init__.py#L558

https://github.com/terrapower/armi/blob/5a41b25fe9f262acb03347923443c4a2808837db/armi/reactor/assemblyLists.py#L108-L110

john-science:
Okay, in the couple of years since this ticket was opened, the underlying problem was solved. Yes, we still have a default SFP in ARMI, but it is no longer a hyper-specific situation. (It had been a Cartesian grid with a specific spacing, with 10 assemblies per column and an infinite number of columns, with no water.)

Now this is just a simple empty grid, without a lot of crazy details in it:

https://github.com/terrapower/armi/blob/bfd0394a48681b30282eb74de13d4d3e10794bd4/armi/reactor/blueprints/__init__.py#L528-L534

So while there IS a default SFP, it's just a placeholder now. Not a strange non-physical textbook idea.

PR #1891, opened 2024-09-17
Adding support for ex-core structures

john-science:
## What is the change?

Added initial support for ex-core structures in ARMI.

* Blueprints can now easily define ex-core structures on their own grids.
* The old `SpentFuelPool` logic was rolled into the new `ExcoreStructure` class.
* The old `AssemblyList` class was removed.
* Reporting tools were updated to match.
* Docs were updated to match.

The ARMI `Reactor` class used to have two children that were also attributes: `r.core` and `r.sfp`.  Now it has two children that are also attributes: `r.core` and `r.excore`.  And `r.excore` is currently a dictionary where keys map to an arbitrary set of ex-core objects:

* `r.excore["sfp"]`
* `r.excore["ivs"]`
* `r.excore["evst1"]`

But, for convivence, you can also access these like attributes:

* `r.excore.sfp`
* `r.excore.ivs`
* `r.excore.evst1`

Thus far, ARMI has really only provided a data object model for the Reactor core (and also a default spent fuel pool). The large teams that model and analyze nuclear reactors will also want to model ex-core structures for fuel handling and mechanical analysis reasons.

---

## Checklist


armi/reactor/blueprints/__init__.py
```diff
@@ -572,20 +572,19 @@ def migrate(bp: Blueprints, cs):
...
+
+     * creates a grid blueprint to represent the core layout from the old ``geomFile`` setting, and
+       applies that grid to a ``core`` system.
+     * moves the radial and azimuthal submesh values from the ``geomFile`` to the assembly designs,
+       but only if they are uniform (this is limiting, but could be made more sophisticated in the
+       future, if there is need)
+
+    This allows settings-driven core map to still be used for backwards compatibility. At some point
+    once the input stabilizes, we may wish to move this out to the dedicated migration portion of
+    the code, and not perform the migration so implicitly.
```
keckler, line 587: Is there any actual change here, or did you just change the length of all the lines?

keckler (commented):
There is a lot to review here. I just did a very quick scan, but will need some time to digest.

armi/reactor/blueprints/__init__.py
john-science: I reverted this change back to `main`. 

It's gone. I was doing a lot of work in this file, but then reverted that work and left the comment change.  It's back to normal now.

armi/reactor/blueprints/reactorBlueprint.py
```diff
@@ -171,34 +170,51 @@ def construct(self, cs, bp, reactor, geom=None, loadAssems=True):
...
-            # because they will be populated with DistributeState.
+            # Non-primary nodes get the assemblies via DistributeState.
             return None
 
-        # TODO: This is also pretty specific to Core-like things. We envision systems
-        # with non-Core-like structure. Again, probably only doable with subclassing of
-        # Blueprints
+        system = self._constructAssemblies(cs, bp, loadAssems, system, gridDesign)
+
+        # TODO: We should allow for non-Assembly objects/geometries to be loaded into the grid. For
```
john-science, line 178: NOTE: I am _not_ adding a new today. This section of this file had **2** TODOs, I rephrased and reduced it to **1**.

armi/reactor/blueprints/reactorBlueprint.py
```diff
@@ -118,89 +116,107 @@ def _resolveSystemType(typ: str):
...
-    def _loadAssemblies(self, cs, container, gridContents, bp):
-        runLog.header(
-            "=========== Adding Assemblies to {} ===========".format(container)
-        )
+    def _loadComposites(self, cs, container, gridContents, bp):
+        runLog.header(f"=========== Adding Composites to {container} ===========")
         badLocations = set()
         for locationInfo, aTypeID in gridContents.items():
+
+            # TODO: We should allow for non-Assembly objects/geometries to be loaded into the grid.
```
john-science, line 218: NOTE: I am not adding a new today. This section of this file had 2 TODOs, I rephrased and reduced it to 1.

alexhjames:
Out of curiosity, why is `excore` it's own object/dictionary? Why not just allow apps to register the excore structures they want directly to reactor? So like `r.core`, `r.sfp`, `r.evst` etc?

drewj-tp:
> Out of curiosity, why is `excore` it's own object/dictionary? Why not just allow apps to register the excore structures they want directly to reactor? So like `r.core`, `r.sfp`, `r.evst` etc?

This feels very natural and something that could make a user's life easier. `r.evst` vs `r.excore["evst"]` is fewer characters.

But to support the level of dynamism needed for `r.evst` to resolve when that attribute is not defined on the class would likely require some new logic in [`Reactor.__getattr__`](https://docs.python.org/3/reference/datamodel.html#object.__getattr__) to be able to dynamically resolve new children. But this also means you can't do children like `"primary pump 3"` which may also be more natural

<details>

<summary> Prototype </summary>

```python
In [1]: class R:
   ...:     def __init__(self):
   ...:         self.core = None
   ...:         self.__excore = {}
   ...:     def add(self, name, obj):
   ...:         self.__excore[name] = obj
   ...:     def __getattr__(self, name):
   ...:         try:
   ...:             return self.__excore[name]
   ...:         except KeyError as ke:
   ...:             raise AttributeError(name) from ke
   ...:

In [2]: r = R()

In [3]: r.core = "hello"

In [4]: r.core
Out[4]: 'hello'

In [5]: r.add("evst", [])

In [6]: r.evst
Out[6]: []

In [7]: r.sfp
KeyError                                  Traceback (most recent call last)
Cell In[1], line 9, in R.__getattr__(self, name)
      8 try:
----> 9     return self.__excore[name]
     10 except KeyError as ke:

KeyError: 'sfp'

The above exception was the direct cause of the following exception:

AttributeError                            Traceback (most recent call last)
Cell In[7], line 1
----> 1 r.sfp

Cell In[1], line 11, in R.__getattr__(self, name)
      9     return self.__excore[name]
     10 except KeyError as ke:
---> 11     raise AttributeError(name) from ke

AttributeError: sfp
```

</details>

Food for thought.

john-science:
> Out of curiosity, why is `excore` it's own object/dictionary? Why not just allow apps to register the excore structures they want directly to reactor? So like `r.core`, `r.sfp`, `r.evst` etc?

Well, that's possible. But we are going to have dozens or hundreds of excore structures, and they area all just `r.whatever`, and there's no organization left.

It's certainly a thought I had early on, but I don't know.

I think if we were to do THAT, I would still prefer: `r.excore.sfp`, `r.excore.ivs`, `r.excore.evst3`.  At least that way, it's clear what these things are.

hmm

alexhjames:
I think I was going along the lines of more general would be more flexible. If you have a crazy reactor setup with two cores, then you would be forced into modeling one core as the `core` and one as `excore.core2`. Or similarly, if someone wanted to run analysis on the spent fuel pool like it was a reactor to verify things, then they could... Seems like it's on the plugins to know what container to look at and run analysis on.

john-science:
Sure, we could do something more like this:

```python
class Recctor:
    # ...

    def __getattr__(self, key):                                                                                                                                                                                    
        try:                                                                                                                                                                                                       
            return self.__dict__[key]                                                                                                                                                                              
        except KeyError:                                                                                                                                                                                           
            if key in self.excore:                                                                                                                                                                                      
                return self.excore[key]                                                                                                                                                                                 
            else:                                                                                                                                                                                                  
                raise                                                                                                                                                                                              
                                                                                                                                                                                                                   
    def __setattr__(self, key, value):                                                                                                                                                                             
        try:                                                                                                                                                                                                       
            self.__dict__[key] = value                                                                                                                                                                             
        except KeyError:                                                                                                                                                                                           
            if key in self.excore:                                                                                                                                                                                      
                self.excore[key] = value                                                                                                                                                                                
            else:                                                                                                                                                                                                  
                raise   
```

This allows `r.sfp` and `r.evst1`, if that's just super important to everyone.

Though I still think `r.excore.sfp` and `r.excore.evst` is more clear.

drewj-tp:
Want to clarify that I am not likely going to be an immediate user nor stakeholder of this feature so take my suggestion w/ a grain of salt. Not a requirement

john-science:
@drewj-tp @jakehader @alexhjames 

I guess encapsulation is my goal here:

```
r.core - core physical object
r.excore - ex-core physical object
r.o - operator
r.id - number
r.p - parameter collection
r.cached - tooling
r.spatialGrid - tooling
r.factory - method
r.add - method
r.isFuel - method
r.backUp - method
```

I mean, I can make `r.sfp` a thing, and `r.ivs`, but it will mean a metric TON of error checking to make sure people don't name their excore structures `id` or `p` or `o` or `cached` or `factory`, because that would break the `Reactor` class.
 
And I'll have to check that no one tries to over-write the method `r.factory()` by doing `r.factory = ExcoreStructure()`.
 
Or, in code you could set `r.ivs = ExcoreStructure()` , but I need to make sure you don't do `r.o = ExcoreStructure()`.
 
There would just be so much error checking where I need to list every single time anyone adds or touches any method or attribute on a `Reactor` object. I would have to keep a full list of all attributes and methods in `Reactor` and it's parents `Composite` and `ArmiObject`.

keckler:
There is at least one place in our benchmarks that will need to be migrated due to this change. We used to allow `r.sfp`, but now it is `r.excore["sfp"]`.

armi/reactor/reactors.py
```diff
@@ -210,6 +216,7 @@ def factory(cs, bp, geom: Optional[SystemLayoutInput] = None) -> Reactor:
     if cs[CONF_GEOM_FILE]:
         blueprints.migrate(bp, cs)
 
+    # For now, ARMI will create a default Spent Fuel Pool and add it to every reactor.
```
keckler, line 219: I do not think this is something that should be necessary any more. As long as all the `excore` things are being written to the database, then building a SFP by default should not be needed. 

A default SFP should be added if the user has specified `trackAssems: True` in their blueprints. Otherwise, I do not think it needs to be or should be added.

armi/reactor/reactors.py
```diff
@@ -639,8 +646,8 @@ def removeAllAssemblies(self, discharge=True):
         assems = set(self)
         for a in assems:
             self.removeAssembly(a, discharge)
-        if hasattr(self.parent, "sfp"):
-            self.parent.sfp.removeAll()
+        if hasattr(self.parent, "excore") and self.parent.excore.get("sfp"):
+            self.parent.excore.sfp.removeAll()
```
keckler, line 650: This check feels a little suspect...

If we need to check the SFP here, why do we not also need to check all the other ex-core structures for assemblies as well?

As a side note, I cannot find a single place that `removeAllAssemblies()` is used, and I can't really imagine what its for...

armi/reactor/excoreStructure.py
```diff
@@ -0,0 +1,147 @@
...
+        self.spatialGrid = None
+
+    def __repr__(self):
+        return "<{}: {} id:{}>".format(self.__class__.__name__, self.name, id(self))
+
+    @property
+    def r(self):
+        return self.getAncestor(fn=lambda x: x.__class__.__name__ == "Reactor")
+
+    def add(self, obj, loc=None):
```
keckler, line 47: The `add()` method on `AssemblyList` had some functionality to allow for things to be added without specifying a location. 

https://github.com/terrapower/armi/blob/2fac7a8e2f105a29174ac33bf0e28318fc9272a3/armi/reactor/assemblyLists.py#L151-L154

It looks like you have decided to get rid of that feature. I think that was nice to have, and removing it will possibly (probably?) break some things downstream.

armi/bookkeeping/report/reportInterface.py
```diff
@@ -166,11 +167,66 @@ def writeReports(self):
...
+        runLog.important("-" * len(title))
+        totFis = 0.0
+        for a in sfp.getChildren():
+            runLog.important(
+                "{assembly:15s} discharged at t={dTime:10f} after {residence:10f} yrs. It entered at cycle: {cycle}. "
+                "It has {fiss:10f} kg (x {mult}) fissile and peak BU={bu:.2f} %.".format(
+                    assembly=a,
+                    dTime=a.p.dischargeTime,
+                    residence=(a.p.dischargeTime - a.p.chargeTime),
+                    cycle=a.p.chargeCycle,
```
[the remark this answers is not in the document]
keckler: That being said, I see that you just migrated this code from one place to another, so it wasn't you who introduced these particular parameters here.

Regardless, might be worth a second look.

keckler (commented):
First off, thank you for all of this effort. This is definitely a step in the right direction. A few comments beyond those left in the code:
1. This implementation seems to assume that any ex-core _thing_ will be grid-like. But probably there are only so many things that might actually be like that, and something more flexible might be needed in the future. This is not a problem for this PR, more so a seed for future work.
2. There are a lot of places where we do a check to make sure that an SFP exists. But as noted by one of the comments that you added, ARMI right now is adding an SFP by default. So maybe those checks are unnecessary?
3. I support the decision to stash everything under `r.excore` instead of making them attributes on `r` directly. I think this makes sense. One comment that Alex brought up is about the possibility of having two `Core`s, and then one of them would have to live under `excore` which is weird. While I agree with this, I also think that storing two `Core` objects on a single `Reactor` doesn't make much sense from a traditional viewpoint. If `Reactor` were instead `Plant`, then maybe I'd agree more with Alex.
4. There is a lot going on here. While everything appears to look good, I will not be at all surprised if this breaks some stuff in my workflow. Oh well.

armi/bookkeeping/report/reportInterface.py
```diff
@@ -166,11 +167,66 @@ def writeReports(self):
...
         # spent fuel pool report
-        if self.r.sfp is not None:
-            self.r.sfp.report()
-            self.r.sfp.count()
+        if self.r.excore.get("sfp") is not None:
+            self.reportSFP(self.r.excore["sfp"])
+            self.countAssembliesSFP(self.r.excore["sfp"])
+
+    @staticmethod
+    def reportSFP(sfp):
```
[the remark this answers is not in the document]
john-science: These are supposed to "reporting tools".

They are a heck of a lot more useful if people can mix-and-match and grab on the go then if you have to build a full `Operator` and grab the `Interface`.

And it was a trivially easy change to make. Since they only take the ONE object, and nothing else from the `self` space.

(I won't die on this hill.)

armi/bookkeeping/report/reportInterface.py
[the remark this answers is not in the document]
john-science: Yeah, this was just me moving the reporting code out of the `armi/reactor/` space and into `armi/bookkeeping/report/`. 

I didn't change anything.

In answer to your question, the lines here will report the "dischargeTime" as the most recent discharge time. That parameter looks correctly set for that, but that's all it really does. It doesn't understand the history at all.

armi/bookkeeping/report/reportInterface.py
```diff
@@ -166,11 +167,66 @@ def writeReports(self):
...
+        for a in sfp.getChildren():
+            runLog.important(
+                "{assembly:15s} discharged at t={dTime:10f} after {residence:10f} yrs. It entered at cycle: {cycle}. "
+                "It has {fiss:10f} kg (x {mult}) fissile and peak BU={bu:.2f} %.".format(
+                    assembly=a,
+                    dTime=a.p.dischargeTime,
+                    residence=(a.p.dischargeTime - a.p.chargeTime),
+                    cycle=a.p.chargeCycle,
+                    fiss=a.getFissileMass() * a.p.multiplicity,
+                    bu=a.getMaxParam("percentBu"),
```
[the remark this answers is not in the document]
john-science: Again, I didn't write any of this, I just migrated it from a bad file.

I can look at it, but I don't know either.

armi/cases/case.py
```diff
@@ -302,24 +302,21 @@ def __eq__(self, that):
...
         """
         Set the task dependence based on the :code:`dependencies`.
 
         This accounts for whether or not the dependency is enabled.
-
-        TODO
-        ----
-        This is a leftover from before the release of the ARMI framework. The API of the
-        proprietary cluster communication library is being used here. This should either
-        be moved out into the cluster plugin, or the library should be made available.
```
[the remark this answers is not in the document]
john-science: I recently helped re-write our internal cluster plugin for our recent release.

This comment is so out of date and defunct it just doesn't apply any more.

(Just FYI, the "cluster communication library" repo has been archived.)

armi/reactor/excoreStructure.py
john-science: You have the choice of having the `obj` contain a `obj.spatialLocator` _OR_ passing in a `loc` directly.

You can do either.

armi/reactor/excoreStructure.py
```diff
@@ -0,0 +1,147 @@
...
+        if "assemNum" in obj.p and obj.p.assemNum < 0:
+            # update the assembly count in the Reactor
+            newNum = self.r.incrementAssemNum()
+            obj.renumber(newNum)
+
+        obj.spatialLocator = loc
+        super().add(obj)
+
+
+class ExcoreCollection(dict):
```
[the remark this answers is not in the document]
john-science: Done

armi/bookkeeping/report/reportInterface.py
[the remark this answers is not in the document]
keckler: True enough, I agree

armi/reactor/reactors.py
john-science: @keckler I have opened a separate ticket for this:  https://github.com/terrapower/armi/issues/1871

This is a separate ticket because that change will be bigger than this one. That change will require changes in downstream TP projects, and will generally more lines of code touched in ARMI than this one.  Now, it will also be a much easier change for me to make, but I'm just trying to keep this PR as small as I can.

TLDR; I agree!

armi/reactor/reactors.py
```diff
@@ -127,12 +128,16 @@ def add(self, container):
...
                 )
             self.core = cores[0]
 
-        if isinstance(container, SpentFuelPool):
-            self.sfp = container
+        if isinstance(container, ExcoreStructure):
+            nomen = container.name.replace(" ", "").lower()
+            if nomen == "spentfuelpool":
+                # special case
+                nomen = "sfp"
```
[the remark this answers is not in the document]
john-science: I can add that, sure. But this will go away when I do the ticket: https://github.com/terrapower/armi/issues/1871

This only exists to backwards support a couple in-house TP projects.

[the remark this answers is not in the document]
john-science: This will go away when ARMI no longer forces a default SFP on people.

armi/reactor/reactors.py
john-science: Agreed. I'll just remove this...

armi/bookkeeping/report/reportInterface.py
```diff
@@ -166,11 +167,66 @@ def writeReports(self):
...
+                    mult=a.p.multiplicity,
+                )
+            )
+            totFis += a.getFissileMass() * a.p.multiplicity / 1000  # convert to kg
+
+        runLog.important(
+            "Total full-core fissile inventory of {0} is {1:.4E} MT".format(
```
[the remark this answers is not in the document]
john-science: fixed

armi/bookkeeping/report/reportInterface.py
```diff
@@ -166,11 +167,66 @@ def writeReports(self):
...
+            runLog.important(
+                "{assembly:15s} discharged at t={dTime:10f} after {residence:10f} yrs. It entered at cycle: {cycle}. "
+                "It has {fiss:10f} kg (x {mult}) fissile and peak BU={bu:.2f} %.".format(
+                    assembly=a,
+                    dTime=a.p.dischargeTime,
+                    residence=(a.p.dischargeTime - a.p.chargeTime),
+                    cycle=a.p.chargeCycle,
+                    fiss=a.getFissileMass() * a.p.multiplicity,
+                    bu=a.getMaxParam("percentBu"),
+                    mult=a.p.multiplicity,
```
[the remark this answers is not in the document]
john-science: fixed

armi/reactor/excoreStructure.py
keckler: Yes, I understand that the location can be specified in two ways. But my understanding is that the location did not used to have to be specified at all, and the assembly would then just get whatever "next position" that is open. 

That ability seemed useful, and it has now gone away.

Am I misunderstanding the new implementation?

john-science: That logic still exists for the Spent Fuel Pool, over in `spentFuelPool.py`.

But the `ExcoreStructure` class is extremely general. These are meant to support things that people have asked me for, like the ability to define ducts and pipes on a grid.

In this, most general possible situation, all you have are "objects in space", and there is no obvious "next place in the grid".

Thoughts?

keckler: Ah, I see. You're right, looks good to me.

keckler (approved):

john-science:
@john-science Test this downstream before merging.

PR #1899, opened 2024-09-20
Adding ParameterCollection.where for conditional parameter iteration

### LINKED ISSUE #1898: Allow a way to efficiently iterate over some parameters
(Pulling a discussion out of https://github.com/terrapower/armi/pull/1877#discussion_r1769233599)

We have a lot of parameters. Not all of them are used in every location. For example, in working on `HexBlock.rotate`, lots of things needs to be rotated. Two cases are
1. Parameters defined on the midpoints and vertices of a hexagon (e.g., with `ParamLocation.CORNERS`)
2. Parameters defined on the children of the block (e.g., with `ParamLocation.CHILDREN`)

For parameters meeting condition 1, we have

https://github.com/terrapower/armi/blob/3dd2763535f0a5694215a0f1f876db49eea3a11c/armi/reactor/blocks.py#L2041-L2042

`ParameterDefinitions.atLocation` involves iterating over all parameters, making an intermediate `ParameterDefinitionCollection`, [extending](https://github.com/terrapower/armi/blob/3dd2763535f0a5694215a0f1f876db49eea3a11c/armi/reactor/parameters/parameterDefinitions.py#L492-L506) to it, twice (once for corners, once for edges).
And then iterate again over the combined names.

In #1877, I added a helper onto `HexBlock` that allowed a pattern like
```python
for param in self._getParamsWhere(lambda pd: pd.atLocation(ParamLocation.CORNERS)):
    pass
```
This allowed for one iteration over all the parameters and used [`filter`](https://docs.python.org/3/library/functions.html#filter) to only produce parameters that met the condition.

This seems more generally useful so let's add it.

## Where?

I don't think it should be added to a composite. Those already have too many methods.

I'm thinking a `ParameterCollection` so you can do something like
```python
assem = r.core[0]
assem.p.where(lambda pd: pd.hasCategory("neutronics"))
```
There could be some confusion about `where` needing a condition for the **values** of the parameter collection, not so much their definitions. The same way [`numpy.where`](https://numpy.org/doc/stable/reference/generated/numpy.where.html#numpy.where) and [`pandas.where`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.where.html#pandas.DataFrame.where) behave. 

So maybe a `ParameterDefinitionCollection`? Then you'd need to do
```python
assem.p.pDefs.where(pd: pd.atLocation(ParamLocation.EDGES))
```
that feels a bit better? Grabbing parameters where their *definition* meets a certain category. Yeah, I think I like that better

drewj-tp:
## What is the change?

Provide `ParameterCollection.where()` for better iteration of parameters

Allows selective iteration of parameters that meet, by their definition
not their value, a certain condition. This would be useful to
iterate over all parameters on a ``Block`` that are relevant for
neutronics calculations with:

```python
for p in block.p.where(lambda pd: pd.hasCategory("neutronics")):
    ...
```

The argument is a function that should return true for a given parameter
and can be complicated:

```python
block.p.where(
    lambda pd: (
        pd.atLocation(ParamLocation.EDGES) or
        pd.atLocation(ParamLocation.CORNERS)
    )
)
```

Inspired by `HexBlock` rotation where we need to do stuff for parameters that meet certain criteria.

Closes #1898 

---

## Checklist


armi/reactor/parameters/parameterCollections.py
```diff
@@ -493,6 +494,32 @@ def restoreBackup(self, paramsToApply):
...
+            Iterator, **not** list or tuple, that produces each parameter that
+            meets ``f(parameter) == True``.
+
+        Examples
+        --------
+        >>> block = r.core[0][0]
+        >>> pdef = block.p.paramDefs
+        >>> for param in pdef.where(lambda pd: pd.atLocation(ParamLocation.EDGES)):
+        ...     print(param.name, block.p[param.name])
+
```
john-science, line 519: UNIMPORTANT

I see you put these empty lines at the end of a about half your docstrings. I just checked, the RST renders fine without this.

armi/reactor/tests/test_parameters.py
```diff
@@ -502,3 +511,86 @@ class MockPCChild(MockPC):
...
+            pb.defParam(
+                name="edgeTemperature",
+                description="edge temperature",
+                location=parameters.ParamLocation.EDGES,
+                categories=[parameters.Category.thermalHydraulics],
+                units="",
+            )
+
+    pc: typing.ClassVar[parameters.ParameterCollection]
+
```
john-science, line 554:
```suggestion
```

john-science: Up to you, but I believe this line is unneeded. The `setupClass()` method handles this for you.

armi/reactor/parameters/parameterCollections.py
drewj-tp: Old habits not consistency implemented I guess 😅How would you like to proceed?

<details>

<summary> Rabbit hole research on conflicting and inconsistent docstring standards </summary>

The trailing empty line in a `numpydoc` is an implied but not explicit part of the standard. All of the examples provided in [the standard](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) have the trailing line included. But it's not stated as a part of the spec and things work well with / without it.

But also, the standard includes this line [towards the end](https://numpydoc.readthedocs.io/en/latest/format.html#common-rest-concepts):

> Line spacing and indentation are significant and should be carefully followed

The [Python docstring conventions PEP](https://peps.python.org/pep-0257/#multi-line-docstrings) also includes

> Insert a blank line after all docstrings (one-line or multi-line) that document a class

</details>

doc/release/0.4.rst
```diff
@@ -14,6 +14,10 @@ New Features
 #. Allow merging a component with zero area into another component (`PR#1858 <https://github.com/terrapower/armi/pull/1858>`_)
 #. Provide utilities for determining location of a rotated object in a hexagonal lattice (``getIndexOfRotatedCell``). (`PR#1846 <https://github.com/terrapower/armi/1846`)
 #. Allow merging a component with zero area into another component. (`PR#1858 <https://github.com/terrapower/armi/pull/1858>`_)
+#. Provide ``Parameter.hasCategory`` for quickly checking if a parameter is defined with a given category.
+   (`PR#1899 <https://github.com/terrapower/armi/pull/1899>`_)
+#. Provide ``ParameterCollection.where`` for efficient iteration over parameters who's definition
```
john-science, line 19:
```suggestion
#. Provide ``ParameterCollection.where`` for efficient iteration over parameters who's definition.
```

armi/reactor/parameters/parameterCollections.py
john-science: Eww... Well, without that knowledge, we have manually scrubbed all such blank lines from ARMI, because the RST renders fine without them and empty lines don't seem helpful.

I can write a script to add them all back in, if this becomes necessary.

Thanks for the info!

armi/reactor/parameters/parameterCollections.py
```diff
@@ -367,7 +367,7 @@ def __contains__(self, name):
         else:
             return name in self._hist
 
-    def __eq__(self, other):
+    def __eq__(self, other: "ParameterCollection"):
```
john-science, line 370: Why is this a string?  `ParameterCollection` is a class defined in this file, wouldn't it be safe to just use the class name?

If I'm missing something here, just holler.

drewj-tp: Because it's not yet defined. We're in the definition of that class so python hasn't finished defining `ParameterCollection`.
```python
In [1]: class D:
   ...:     def __eq__(self, other: D):
   ...:         pass
   ...:
NameError                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 class D:
      2     def __eq__(self, other: D):
      3         pass

Cell In[1], line 2, in D()
      1 class D:
----> 2     def __eq__(self, other: D):
      3         pass

NameError: name 'D' is not defined
```

armi/reactor/tests/test_parameters.py
```diff
@@ -502,3 +510,84 @@ class MockPCChild(MockPC):
...
+        ):
+            self.assertTrue(p.atLocation(parameters.ParamLocation.EDGES), msg=p)
+            names.remove(p.name)
+        self.assertFalse(names, msg=f"{names=} should be empty!")
+
+    def test_complicated(self):
+        """Test a multi-condition filter."""
+        names = {
+            "cornerFlux",
+        }
```
john-science, line 583:
```suggestion
        names = {"cornerFlux"}
```

armi/reactor/tests/test_parameters.py
```diff
@@ -502,3 +510,84 @@ class MockPCChild(MockPC):
...
+        ):
+            self.assertTrue(p.hasCategory(parameters.Category.neutronics), msg=p)
+            names.remove(p.name)
+        self.assertFalse(names, msg=f"{names=} should be empty!")
+
+    def test_onLocation(self):
+        """Test the use of Parameter.atLocation in filtering."""
+        names = {
+            "edgeTemperature",
+        }
```
john-science, line 571:
```suggestion
        names = {"edgeTemperature"}
```

john-science (approved):

PR #1914, opened 2024-09-27
Organizing test settings files by category

john-science:
## What is the change?

Organized the settings in the test settings files by category.

> Take a look and see what you think. I generated these files via scripting, so I can make tweaks if need be.

@albeanth [recently suggested](https://github.com/terrapower/armi/discussions/1083) that ARMI settings files could have categories to help us organize them. Now, that is a fun idea, but will be a very large API-breaking change. This PR gives us a lot of the benefits TODAY, as opposed to waiting a long time for ALL of the benefits.

---

## Checklist

### DIFF
--- a/armi/tests/armiRun.yaml
+++ b/armi/tests/armiRun.yaml
@@ -1,4 +1,5 @@
 settings:
+# global
   availabilityFactor: 1
   beta: 0.003454
   branchVerbosity: debug
@@ -6,6 +7,47 @@ settings:
     - 100
   burnSteps: 2
   comment: Simple test input.
+  cycleLength: 2000.0
+  detailAssemLocationsBOL:
+    - 002-001
+  freshFeedType: igniter fuel
+  loadingFile: refSmallReactor.yaml
+  moduleVerbosity:
+    armi.reactor.reactors: info
+  nCycles: 6
+  outputFileExtension: png
+  power: 100000000.0
+  smallRun: true
+  startCycle: 1
+  startNode: 2
+  targetK: 1.002
+  verbosity: extra
+  versions:
+    armi: uncontrolled
+
+# database
+  db: false
+
+# equilibrium
+  eqRingSchedule:
+    - 13
+    - 1
+
+# fuel cycle
+  fuelHandlerName: EquilibriumShuffler
+  jumpRingNum: 9
+  shuffleLogic: refSmallReactorShuffleLogic.py
+
+# neutronics
+  epsFSAvg: 1e-06
+  epsFSPoint: 1e-06
+  loadPadElevation: 200.0
+
+# report
+  genReports: false
+  summarizeAssemDesign: false
+
+# unknown
   crossSectionControl:
     DA:
       geometry: 0D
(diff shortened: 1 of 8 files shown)


john-science:
This little exercise has already born some fruit.

@keckler Notice the setting here `epsBurnTime` is in an "unknown" category? That's because this IS NOT an ARMI setting, but comes from an internal neutronics plugin. Now I want to remove those here, but will have to test if it breaks downstream unit tests.

@drewejohnson @albeanth When you organize settings like this, it makes it _really_ look like there should be a group of settings explicitly tied to the "axial expansion" plugin.

albeanth:
> Now I want to remove those here, but will have to test if it breaks downstream unit tests.

I'd be in favor of removing them if it's not too big of a deal! 

> When you organize settings like this, it makes it really look like there should be a group of settings explicitly tied to the "axial expansion" plugin.

Yeah, that's not entirely surprising, tbh.

armi/tests/armiRun.yaml
```diff
@@ -31,30 +51,23 @@ settings:
...
-  shuffleLogic: refSmallReactorShuffleLogic.py
-  smallRun: true
   summarizeAssemDesign: false
-  targetK: 1.002
-  verbosity: extra
-  versions:
-    armi: uncontrolled
+
+# unknown
+  epsBurnTime: 0.001
```
john-science, line 73:
```suggestion
```

armi/tests/refTestCartesian.yaml
```diff
@@ -1,21 +1,30 @@
...
+# neutronics
+  epsFSAvg: 1e-06
+  epsFSPoint: 1e-06
+  loadPadElevation: 200.0
+
+# report
+  summarizeAssemDesign: false
+
+# unknown
+  epsBurnTime: 0.001
```
john-science, line 30:
```suggestion
```

armi/tests/tutorials/c5g7-settings.yaml
```diff
@@ -1,16 +1,23 @@
...
     armi: uncontrolled
+
+# neutronics
+  genXS: Neutron
+
+# report
+  genReports: false
+
+# unknown
+  fakeBad: 123456
```
albeanth, line 23: Can we remove this?

albeanth (approved):
Just the one minor comment to try and remove that other unknown setting.

armi/tests/tutorials/c5g7-settings.yaml
john-science: Hmm... I mean, this is a unit tests file, and that is in the unit test.

I can look at re-writing that test.

john-science: Okay, I changed the test. We're good.

armi/tests/test_lwrInputs.py
```diff
@@ -63,7 +63,6 @@ def test_loadC5G7(self):
             streamVal = mock.getStdout()
             self.assertIn("UraniumOxide", streamVal, msg=streamVal)
             self.assertIn("SaturatedWater", streamVal, msg=streamVal)
-            self.assertIn("invalid settings: fakeBad", streamVal, msg=streamVal)
```
albeanth, line 66: Ah I see, I didn't even think that it was there on purpose. This sort of behavior (checking that invalid settings are caught in settings validation) feels like it should be covered elsewhere, yeah? 

Just want to make sure we aren't losing something by ripping this out.

john-science: We have this in several other places, not to worry.

PR #1917, opened 2024-09-30
Loading some ARMI DBs without the App that created them

john-science:
## What is the change?

This PR add supports for loading of some ARMI Databases without the App that created them. This is not meant to be a "complete solution" to load ANY ARMI database with that app that created it. But it should help ease the burden a bit.

Also, since reading a data file without understanding the program that created it is a potentially hazardous idea, I have placed some training wheels on the resultant reactor data model. The parameters on this data model are set to read-only mode. This is to make it CLEAR that you do not have enough information to continue to time-evolve or alter this data model.

This was a user request.

---

## Checklist


armi/bookkeeping/db/passiveDBLoadPlugin.py
```diff
@@ -0,0 +1,104 @@
...
+            PassThroughYamlize, loader, node, round_trip_data
+        )
+
+
+# TODO: JOHN needs testing!
+class PassiveDBLoadPlugin(plugins.ArmiPlugin):
+    """Provides the ability to ignore parameters sections of blueprint files.
+
+    This plugin allows you to more easily open a database, because you can ignore sections of the
+    blueprint files and as many of the parameters as you want.
```
alexhjames, line 46: Just to clarify, this isn't skipping the parameters right? You can still look at that parameter's value, you just don't necessarily know what it means or what its units are without outside knowledge.

john-science: The parameters are still loading, but with defaulted / dummy definitions and units. And they won't be attached to any range validation.

john-science:
@alexhjames @jeffbaylor I just want to keep this on everyone's radar. I'm still hoping for a review.

drewj-tp:
Passing comment that came to me this morning: this only works if all the materials and shapes in the database are loadable, right? This change seems to
- skip blueprints if we don't know how to process them
- set all parameters to read only

Which is good.

But if my application defines a new component that only exists in my application, we will have no way to load that, correct? Same for a material.

If we really want to make it possible so any application can open any database, as the title here suggests, those are two additional things we'd need to handle

john-science:
> But if my application defines a new component that only exists in my application, we will have no way to load that, correct? Same for a material.
> 
> If we really want to make it possible so any application can open any database, as the title here suggests, those are two additional things we'd need to handle

Correct.  There are actually many more things that people can do in their code that will make it hard to open an ARMI DB without the Application that created it.

Please note that the code in this PR does not suggest a complete solution to that problem. I provide a tool to ignore blueprint sections, and another to ignore parameters. That's really all you get here.

The only "complete solution" to load a DB without the App that created it would mean storing huge portions of that code in the DB.  And I don't think that's a rabbit hole we want to go down, it's been tried before.

TLDR; the title of this PR and its ticket are perhaps the problem here. I grant that. Here I provide 2 or 3 tools that downstream users can try to use to solve the problem, but only in a limited fashion.

armi/bookkeeping/db/passiveDBLoadPlugin.py
```diff
@@ -0,0 +1,104 @@
...
+        for nomen in ["flags", "serialNum"]:
+            if nomen in names:
+                names.remove(nomen)
+
+        # build a collection of defaulted parameters to passively ignore
+        desc = "This is just a silly placeholder Parameter. It's meaning is unknown."
+        pDefs = parameters.ParameterDefinitionCollection()
+        with pDefs.createBuilder(location=ParamLocation.AVERAGE) as pb:
+            for param in names:
+                pb.defParam(param, units=units.UNITLESS, description=desc)
```
alexhjames, line 102: Would we want to make `saveToDB` false for these parameters? I can see both sides. You may not want to write them since you don't know what they are, but it may be good to write them so that you don't lose information from the database file.

alexhjames (changes requested):
I completed reviewing the PR contents and have a few comments. I have not done any testing with Oxbow yet. I'll work on that next.

armi/reactor/reactorParameters.py
```diff
@@ -770,3 +770,38 @@ def defineCoreParameters():
...
+
+    Notes
+    -----
+    Once you make one Reactor read-only, it will lead to side effects in all Reactors you open in
+    ARMI. So if you run this function once with `readOnly=True`, and you want to move on to other
+    things, run this method on your reactor again with `readOnly=False` to unclock ARMI.
+    """
+    for pdef0 in r.p.paramDefs:
+        pdef0.readOnly = readOnly
+        for system in r.getChildren():
```
alexhjames, line 796: One more comment I missed. This loop and it's children shouldn't be nested in the outer for loop.

armi/bookkeeping/db/database3.py
```diff
@@ -798,6 +796,42 @@ def load(
...
+        statePointName : str, optional
+            Statepoint name (e.g., "special" for "c00n00-special/")
+
+        Returns
+        -------
+        Reactor
+            The top-level object stored in the database; a Reactor.
+        """
+        r = self.load(cycle, node, cs, bp, statePointName, allowMissing=True)
+        makeParametersReadOnly(r)
```
alexhjames, line 832: One last thing is that volume needs to either be set for each component (by calling `componet.getVolume()` or the `getVolume()` call needs to be modified to not modify the volume parameter. This needs to happen before the parameters are made read only.

armi/reactor/reactorParameters.py
```diff
@@ -770,3 +770,38 @@ def defineCoreParameters():
...
+            for pdef1 in system.p.paramDefs:
```
jakehader, line 797: Is there a get children recursive functionality on the reactor? Not sure, but it might be better to have one in case the reactor or systems are not only 4 deep

armi/bookkeeping/db/passiveDBLoadPlugin.py
```diff
@@ -0,0 +1,104 @@
...
+            if len(paramNames):
+                params[dataClass] = PassiveDBLoadPlugin.buildParamColl(paramNames)
+
+        return params
+
+    @staticmethod
+    def buildParamColl(names):
+        """Try replacing any missing parameters with unitless nonsense."""
+        # handle the special cases
+        for nomen in ["flags", "serialNum"]:
```
jakehader, line 93: Can these special cases we exposed in some way more broadly so they are classified? What makes these special cases?

armi/bookkeeping/db/passiveDBLoadPlugin.py
```diff
@@ -0,0 +1,104 @@
...
+            if nomen in names:
+                names.remove(nomen)
+
+        # build a collection of defaulted parameters to passively ignore
+        desc = "This is just a silly placeholder Parameter. It's meaning is unknown."
```
jakehader, line 98: The case that @alexhjames brings up with `saveToDB` set to True brings up an interesting point about saving state. Imagine we have app1 that makes a database that is loaded by app2 and app2 adds new parameters and saves the database. Now opening this back into app1 would not only drop the descriptions that app2 added, but wouldn't it also drop the descriptions of the parameters that app1 should know about? 

In the most extreme case I think you could end up with unit less and no descriptions on all parameters in the database? Could be totally wrong here but his comment sparked the thought.

armi/bookkeeping/db/database3.py
jakehader: I think volume might be special, but I'm curious about linked components and areas as well. There might be several edge cases here that need to be checked explicitly on component modifications. The tests seem to demonstrate that setting a parameter directly that is in read only won't work, but we should check mesh related things as well.

armi/bookkeeping/db/passiveDBLoadPlugin.py
```diff
@@ -0,0 +1,104 @@
...
+    This was designed to allow loading an ARMI database without the application that created it.
+
+    Notes
+    -----
+    Obviously, if you are ignoring huge groups of parameters or whole sections of the blueprints,
+    you are losing information. There is no way to use this plugin and still claim full fidelity of
+    your understanding of the reactor. ARMI does not support any such claims.
+    """
+
+    SKIP_BP_SECTIONS = []
```
jakehader, line 56: Any special things for other inputs, like settings? Are settings stored in the database and used in the loading?

armi/bookkeeping/db/passiveDBLoadPlugin.py
```diff
@@ -0,0 +1,104 @@
...
+    SKIP_PARAMS = {}
```
jakehader, line 57: How is this class intended to be used? Is the intention that an app would subclass this and then apply specific parameters here to skip? I'm probably missing something

armi/bookkeeping/db/passiveDBLoadPlugin.py
jakehader, line 98: Also, probably need a better description.

armi/bookkeeping/db/tests/test_databaseInterface.py
```diff
@@ -456,6 +457,21 @@ def _fullCoreSizeChecker(self, r):
...
+            # now show we can no longer edit those parameters
+            with self.assertRaises(RuntimeError):
+                r.core.p.keff = 0.99
+
+            with self.assertRaises(RuntimeError):
+                b = r.core.getFirstBlock()
+                b.p.power = 432.1
+
+            # needed so that futher tests can run
+            makeParametersReadOnly(r, readOnly=False)
```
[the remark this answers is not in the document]
john-science: Done!

armi/bookkeeping/db/database3.py
```diff
@@ -798,6 +796,42 @@ def load(
...
+        self,
+        cycle,
+        node,
+        cs=None,
+        bp=None,
+        statePointName=None,
+    ):
+        """Load a new reactor, in read-only mode from a DB at (cycle, node).
+
+        Case settings and blueprints can be provided, or read from the database. Providing these
```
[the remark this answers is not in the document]
john-science: done

armi/bookkeeping/db/passiveDBLoadPlugin.py
john-science: Unknown settings are just ignored in this setup because we use `DB.load(allowMissing=True)`.

armi/bookkeeping/db/passiveDBLoadPlugin.py
john-science: I added a usage guide to the docstring.

armi/bookkeeping/db/passiveDBLoadPlugin.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+
+
+class PassiveDBLoadPlugin(plugins.ArmiPlugin):
+    """Provides the ability to ignore parameters sections of blueprint files.
+
+    This plugin allows you to more easily open a database, because you can ignore sections of the
+    blueprint files and as many of the parameters as you want.
+
+    This was designed to allow loading an ARMI database without the application that created it.
```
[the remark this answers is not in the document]
john-science: done

armi/bookkeeping/db/passiveDBLoadPlugin.py
john-science: done

armi/bookkeeping/db/passiveDBLoadPlugin.py
john-science: That's defunct code. Removed.

armi/bookkeeping/db/passiveDBLoadPlugin.py
```diff
@@ -0,0 +1,104 @@
...
+        """Try replacing any missing parameters with unitless nonsense."""
+        # handle the special cases
+        for nomen in ["flags", "serialNum"]:
+            if nomen in names:
+                names.remove(nomen)
+
+        # build a collection of defaulted parameters to passively ignore
+        desc = "This is just a silly placeholder Parameter. It's meaning is unknown."
+        pDefs = parameters.ParameterDefinitionCollection()
+        with pDefs.createBuilder(location=ParamLocation.AVERAGE) as pb:
```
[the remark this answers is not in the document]
john-science: Yeah, there is not `ParamLocation.UNDEFINED`. And we have a lot of logic in ARMI and downstream that depends on param location. I don't want to inject a "undefined" into the system because... you know... people will start _using_ it.  Can't have that.

alexhjames (approved):
My only question would be if there should be any accompanying SQA documentation for the read only load (like req/implementation tags)? If you want to discuss that at a later date, I'm fine with this PR as is.

armi/reactor/reactorParameters.py
```diff
@@ -770,3 +770,29 @@ def defineCoreParameters():
...
+    Parameters
+    ----------
+    r : Reactor
+        Full reactor object, to be modified.
+
+    Notes
+    -----
+    Once you make one Reactor read-only, you cannot make it writeable again.
+    """
+    r.p.readOnly = True
```
jakehader, line 790: Hey @john-science I had a comment on this related to assuming this is the structure of the rector. Is there a better way to do a recursive get all children from the top level? Right now this could be fragile

armi/bookkeeping/db/database3.py
```diff
@@ -798,6 +796,37 @@ def load(
...
+        """
+        r = self.load(cycle, node, statePointName=statePointName, allowMissing=True)
+        self._setParamsBeforeFreezing(r)
+        makeParametersReadOnly(r)
+        return r
+
+    @staticmethod
+    def _setParamsBeforeFreezing(r):
+        """Set some special case parameters before they are made read-only."""
+        for a in r.core:
```
jakehader, line 824: See comment below

armi/bookkeeping/db/database3.py
jakehader: @john-science or @alexhjames did we test any edge cases?

armi/reactor/reactorParameters.py
john-science: Wait, one sec.

john-science: done

armi/bookkeeping/db/database3.py
john-science: ?

jakehader: Sorry, the in meant the comment about getChildren(deep=True). I wonder if we can recurse to only the last layer of the tree to call getVolume(). Similar reasoning

jakehader: Maybe this could be 

```
for obj in r.getChildren(deep=True):
    if not isinstance(obj, Component):
        continue
    obj.getVolume()  
```

or it might work to just call `r.getVolume()` since the recursion should happen all the way down the tree automatically?

john-science: Done

jakehader: Thanks! Just out of curiosity, would `r.getVolume()` work?

jakehader (approved):
Nice job! Thanks @john-science!

PR #1920, opened 2024-09-30
Extensibility improvements to axial expansion classes

### LINKED ISSUE #1453: Remove unnecessary spec for dummy block
https://github.com/terrapower/armi/blob/753660d71f3777680d541ff5d042d2d71100f4a3/armi/reactor/converters/axialExpansionChanger.py#L752-L753

The axial expansion target component for the dummy block is never used. We don't need to enforce that a coolant component be the axial expansion target component for the dummy block. `_setTargetComponent` can also just loop over the non-dummy blocks.

### LINKED ISSUE #1918: Allow axial expansion related classes to be more extensible
I'm working on some internal axial expansion work and finding myself needing to modify some of the functionality on the base class. In some cases I can call `super` and be good, but in other places I need a bit more control. I've got some partially developed code that I wanted to present in a ticket before fully submitting the PR. Estimating 10/6 for draft PR.

## `ExpanionData.computeThermalExpansionFactors`

One example is modifications to how `ExpansionData` computes expansion factors for a given component. The current logic for this method is fully encased in a single method

https://github.com/terrapower/armi/blob/34619dfa7991454d45d5d61e9002bea766a353a3/armi/reactor/converters/axialExpansionChanger/expansionData.py#L174-L190

but there isn't a simple way to drop in more bespoke expansion into this method without rewriting the entire routine. So, I propose cracking this method specifically into one primary method and two helper methods, more or less like
```python
def  computeThermalExpansionFactors(self):
    for b in self._a:
        self._blockThermalExpansionFactors(b)  # name TBD
def _blockThermalExpansionFactors(self, b):
    for c in getSolidComponents(b):
        self._componentThermalExpansionFactors(c)
def _componentThermalExpansionFactors(self, c):
    # more or less the internals of the current implementation
```

## Assembly axial link check
The logic for determining if two components are axially linked is vital for determining how the tops and bottoms of blocks move through expansion. This logic is performed in `_determinedLinked` and isn't really overridable from within

https://github.com/terrapower/armi/blob/34619dfa7991454d45d5d61e9002bea766a353a3/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py#L186-L200

I propose a staticmethod that points back to that function. Subclasses can use or override this method if they need additional logic for determining component linking.

## Small data class for axial links

The data structure for the linking between blocks and between components is `dict[T, list[T | None, T | None]]` so there's lines like

https://github.com/terrapower/armi/blob/34619dfa7991454d45d5d61e9002bea766a353a3/armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py#L325

where it may not be immediately clear (outside of some neighboring comments) what position `[0]` in this returned value represents. 

I would like to add a small dataclass like
```python
class Link:
    lower: T | None
    upper: T | None
```
(type `T` standin) that would make that above code snippet (and others) look like
```python
b.p.zbottom = self.linked.linkedBlocks[b].lower.p.ztop
```
that's more readable in my opinion.

## Other considerations

1. Provide an iterator form of `getSolidComponents` since we mostly use it for iteration - #1915 
2. Combine check for zero-valued and negative expansion factors - https://github.com/terrapower/armi/pull/1861#discussion_r1751058614
3. https://github.com/terrapower/armi/issues/1453

drewj-tp:
## What is the change?

A handful of improvements to allow subclasses of axial expansion things more control. A general theme was to take large methods and break them up so subclasses can override at their discretion.

Some documentation has been added to axial expansion related classes. It's by no means complete, nor to the level of #935. But as I fumbled through the code and got helped from @albeanth, I added what I could to the docstrings.
 
I am working on features internally that need more control over performing axial expansion that require these changes.

Closes #1453 

Closes #1918 
---

## Checklist

### DIFF
--- a/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
+++ b/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
@@ -75,9 +75,7 @@ def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
         odB = componentB.getBoundingCircleOuterDiameter(cold=True)
         biggerID = max(idA, idB)
         smallerOD = min(odA, odB)
-        if biggerID >= smallerOD:
-            return False
-        return True
+        return biggerID < smallerOD
     return False
 
 
--- a/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
+++ b/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
@@ -268,7 +268,7 @@ class AssemblyAxialLinkage:
         for ib, linkdBlk in enumerate(self.linkedBlocks[b]):
             if linkdBlk is not None:
                 for otherC in iterSolidComponents(linkdBlk.getChildren()):
-                    if areAxiallyLinked(c, otherC):
+                    if self.areAxiallyLinked(c, otherC):
                         if lstLinkedC[ib] is not None:
                             errMsg = (
                                 "Multiple component axial linkages have been found for "
@@ -292,3 +292,34 @@ class AssemblyAxialLinkage:
                 f"Assembly {self.a}, Block {b}, Component {c} has nothing linked above it!",
                 single=True,
             )
+
+    @staticmethod
+    def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
+        """Check if two components are axially linked.
+
+        Components are considered linked if the following are found to be true:
+
+        1. Both contain solid materials.
+        2. They have compatible types (e.g., ``Circle`` and ``Circle``).
+        3. Their multiplicities are the same.
+        4. The smallest inner bounding diameter of the two is less than the largest outer
+        bounding diameter of the two.
+
+        Parameters
+        ----------
+        componentA : :py:class:`Component <armi.reactor.components.component.Component>`
+            component of interest
+        componentB : :py:class:`Component <armi.reactor.components.component.Component>`
+            component to compare and see if is linked to componentA
+
+        Returns
+        -------
+        bool
+            Status of linkage check
+
+        See Also
+        --------
+        :func:`areAxiallyLinked` for more details. This method is provided to allow
+        subclasses the ability to override the linkage check.
+        """
+        return areAxiallyLinked(componentA, componentB)
--- a/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
+++ b/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
@@ -284,12 +284,12 @@ class AssemblyAxialLinkage:
 
         self.linkedComponents[c] = lstLinkedC
 
-        if lstLinkedC[0] is None:
+        if self.linkedBlocks[b].lower is None and lstLinkedC.lower is None:
             runLog.debug(
                 f"Assembly {self.a}, Block {b}, Component {c} has nothing linked below it!",
                 single=True,
             )
-        if lstLinkedC[1] is None:
+        if self.linkedBlocks[b].upper is None and lstLinkedC.upper is None:
             runLog.debug(
                 f"Assembly {self.a}, Block {b}, Component {c} has nothing linked above it!",
                 single=True,
--- a/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
+++ b/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
@@ -19,7 +19,7 @@ from armi import runLog
 from armi.reactor.blocks import Block
 from armi.reactor.components import Component, UnshapedComponent
 from armi.reactor.converters.axialExpansionChanger.expansionData import (
-    getSolidComponents,
+    iterSolidComponents,
 )
 
 
@@ -191,7 +191,7 @@ class AssemblyAxialLinkage:
         """Gets the block and component based linkage."""
         for b in self.a:
             self._getLinkedBlocks(b)
-            for c in getSolidComponents(b):
+            for c in iterSolidComponents(b):
                 self._getLinkedComponents(b, c)
 
     def _getLinkedBlocks(self, b: Block):
@@ -267,7 +267,7 @@ class AssemblyAxialLinkage:
         lstLinkedC = AxialLink(None, None)
         for ib, linkdBlk in enumerate(self.linkedBlocks[b]):
             if linkdBlk is not None:
-                for otherC in getSolidComponents(linkdBlk.getChildren()):
+                for otherC in iterSolidComponents(linkdBlk.getChildren()):
                     if areAxiallyLinked(c, otherC):
                         if lstLinkedC[ib] is not None:
                             errMsg = (
--- a/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
+++ b/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
@@ -13,14 +13,23 @@
 # limitations under the License.
 
 from armi import runLog
-from armi.reactor.components import UnshapedComponent
+from armi.reactor.blocks import Block
+from armi.reactor.components import Component, UnshapedComponent
 from armi.reactor.converters.axialExpansionChanger.expansionData import (
     getSolidComponents,
 )
 
 
-def _determineLinked(componentA, componentB):
+def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
     """Determine axial component linkage for two components.
+
+    Components are considered linked if the following are found to be true:
+
+    1. Both contain solid materials.
+    2. They have compatible types (e.g., ``Circle`` and ``Circle``).
+    3. Their multiplicities are the same.
+    4. The smallest inner bounding diameter of the two is less than the largest outer
+       bounding diameter of the two.
 
     Parameters
     ----------
@@ -33,8 +42,6 @@ def _determineLinked(componentA, componentB):
     -----
     - Requires that shapes have the getCircleInnerDiameter and getBoundingCircleOuterDiameter
       defined
-    - For axial linkage to be True, components MUST be solids, the same Component Class,
-      multiplicity, and meet inner and outer diameter requirements.
     - When component dimensions are retrieved, cold=True to ensure that dimensions are evaluated
       at cold/input temperatures. At temperature, solid-solid interfaces in ARMI may produce
       slight overlaps due to thermal expansion. Handling these potential overlaps are out of scope.
@@ -57,29 +64,18 @@ def _determineLinked(componentA, componentB):
                 "they are going to be assumed to not be linked.",
                 single=True,
             )
-            linked = False
-        else:
-            idA, odA = (
-                componentA.getCircleInnerDiameter(cold=True),
-                componentA.getBoundingCircleOuterDiameter(cold=True),
-            )
-            idB, odB = (
-                componentB.getCircleInnerDiameter(cold=True),
-                componentB.getBoundingCircleOuterDiameter(cold=True),
-            )
-
-            biggerID = max(idA, idB)
-            smallerOD = min(odA, odB)
-            if biggerID >= smallerOD:
-                # one object fits inside the other
-                linked = False
-            else:
-                linked = True
-
-    else:
-        linked = False
-
-    return linked
+            return False
+        # Check if one component could fit within the other
+        idA = componentA.getCircleInnerDiameter(cold=True)
+        odA = componentA.getBoundingCircleOuterDiameter(cold=True)
+        idB = componentB.getCircleInnerDiameter(cold=True)
+        odB = componentB.getBoundingCircleOuterDiameter(cold=True)
+        biggerID = max(idA, idB)
+        smallerOD = min(odA, odB)
+        if biggerID >= smallerOD:
+            return False
+        return True
+    return False
 
 
 class AssemblyAxialLinkage:
@@ -112,7 +108,7 @@ class AssemblyAxialLinkage:
             for c in getSolidComponents(b):
                 self._getLinkedComponents(b, c)
 
-    def _getLinkedBlocks(self, b):
+    def _getLinkedBlocks(self, b: Block):
         """Retrieve the axial linkage for block b.
 
         Parameters
@@ -168,7 +164,7 @@ class AssemblyAxialLinkage:
                 single=True,
             )
 
-    def _getLinkedComponents(self, b, c):
+    def _getLinkedComponents(self, b: Block, c: Component):
         """Retrieve the axial linkage for component c.
 
         Parameters
@@ -187,7 +183,7 @@ class AssemblyAxialLinkage:
         for ib, linkdBlk in enumerate(self.linkedBlocks[b]):
             if linkdBlk is not None:
                 for otherC in getSolidComponents(linkdBlk.getChildren()):
-                    if _determineLinked(c, otherC):
+                    if areAxiallyLinked(c, otherC):
                         if lstLinkedC[ib] is not None:
                             errMsg = (
                                 "Multiple component axial linkages have been found for "
--- a/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
+++ b/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import typing
+import dataclasses
+
 from armi import runLog
 from armi.reactor.blocks import Block
 from armi.reactor.components import Component, UnshapedComponent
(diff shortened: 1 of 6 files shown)


armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -201,13 +282,44 @@ def _getLinkedComponents(self, b, c):
...
-        if lstLinkedC[1] is None:
+        if self.linkedBlocks[b].upper is None and lstLinkedC.upper is None:
             runLog.debug(
                 f"Assembly {self.a}, Block {b}, Component {c} has nothing linked above it!",
                 single=True,
             )
+
+    @staticmethod
+    def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
+        """Check if two components are axially linked.
```
drewj-tp, line 298: @keckler This change here means it's possible to have a subclass of the `AssemblyAxialLinkage` provide its own way to check if components should be linked. It's not _super_ easy, but possible. Rough steps:

1. Application brings a custom axial expansion changer subclass via hooks - #1870 
2. Special axial expansion changer subclass uses a subclass of `AssemblyAxialLinkage`
3. Special `AssemblyAxialLinkage` overrides this method and adds custom logic as necessary.

It would work in a pinch to help with https://github.com/terrapower/armi/issues/769

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -57,118 +69,191 @@ def _determineLinked(componentA, componentB):
...
+# Make a generic type so we can "template" the axial link class based on what could be above/below a thing
+Comp = typing.TypeVar("Comp", Block, Component)
+
+
+@dataclasses.dataclass
+class AxialLink(typing.Generic[Comp]):
+    """Small class for named references to objects above and below a specific object.
+
+    Axial expansion in ARMI works by identifying what objects occupy the same space axially.
+    For components, what components in the blocks above and below axially align? This is used
```
john-science, line 93:
```suggestion
    Axial expansion in ARMI works by identifying what objects occupy the same axial space.
    For components in blocks, identify which above and below axially align. This is used
```

john-science: Something struck me as off about the phrasing, but I'm not married to my suggestion. Take it or leave it.

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -57,118 +69,191 @@ def _determineLinked(componentA, componentB):
...
-            else:
-                linked = True
+    def __iter__(self):
+        return iter([self.lower, self.upper])
 
-    else:
-        linked = False
 
-    return linked
+if typing.TYPE_CHECKING:
```
john-science, line 157: Can we move this up to just after the regular imports? It's strange to see importing this far down the file.

#wont-die-on-that-hill

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -28,82 +28,103 @@
...
     """
-    return [c for c in b if not isinstance(c.material, material.Fluid)]
+    return list(iterSolidComponents(b))
 
 
 class ExpansionData:
-    """Data container for axial expansion."""
+    r"""Data container for axial expansion.
+
+    The primary responsibility of this class is to determine the axial expansion factors
```
john-science, line 62: Thanks for adding all this documentation!

john-science (approved):
As code cleanup goes, I think this is an improvement. Thanks!

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -183,31 +268,50 @@ def _getLinkedComponents(self, b, c):
...
             runLog.debug(
                 f"Assembly {self.a}, Block {b}, Component {c} has nothing linked above it!",
                 single=True,
             )
+
+    @staticmethod
+    def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
+        """Check if two components are axially linked.
+
+        Components are considered linked if the following are found to be true:
```
albeanth, line 292: This and the following list feel like they can be deleted since they're restated up in in the `areAxiallyLinked` function.

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -12,16 +12,34 @@
...
+    from armi.reactor.assemblies import Assembly
+
+
+def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
     """Determine axial component linkage for two components.
 
+    Components are considered linked if the following are found to be true:
+
+    1. Both contain solid materials.
+    2. They have compatible types (e.g., ``Circle`` and ``Circle``).
```
albeanth, line 38:
```suggestion
    2. They are the same types (e.g., ``Circle`` and ``Circle``).
```
Is just a bit more explicit/concrete imo

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -12,16 +12,34 @@
...
+    3. Their multiplicities are the same.
+    4. The smallest inner bounding diameter of the two is less than the largest outer
```
albeanth, line 40:
```suggestion
    4. The biggest inner bounding diameter of the two is less than the smallest outer
```
For the change, consider a pin above an annulus where the pin would fit into the annulus (so not overlapping). 
- The smallest inner bounding diameter is the pin (it is zero)
- zero is less than the OD of the annulus; so by that logic, they would overlap, which they don't. 

But if we flip it, 
- The biggest ID is that of the annulus. 
- If the ID of the annulus is less than the smaller of the ODs (the pin), then we're good.

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -57,118 +73,187 @@ def _determineLinked(componentA, componentB):
...
-            if biggerID >= smallerOD:
-                # one object fits inside the other
-                linked = False
-            else:
-                linked = True
 
-    else:
-        linked = False
+# Make a generic type so we can "template" the axial link class based on what could be above/below a thing
+Comp = typing.TypeVar("Comp", Block, Component)
```
albeanth, line 89: You'll have to teach this to me too

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -57,118 +73,187 @@ def _determineLinked(componentA, componentB):
...
-                single=True,
-            )
+    def _findComponentLinkedTo(
+        self, c: Component, otherBlock: typing.Optional[Block]
+    ) -> typing.Optional[Component]:
+        if otherBlock is None:
+            return None
+        candidate = None
+        # Iterate over all solid components in the other block that are linked to this one
+        areLinked = functools.partial(self.areAxiallyLinked, c)
```
albeanth, line 241: I trust this works, but I don't know how it does 😅

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
```diff
@@ -325,7 +330,7 @@ def axiallyExpandAssembly(self):
                 b.p.zbottom = self.linked.linkedBlocks[b][0].p.ztop
             isDummyBlock = ib == (numOfBlocks - 1)
             if not isDummyBlock:
-                for c in getSolidComponents(b):
+                for c in iterSolidComponents(b):
                     growFrac = self.expansionData.getExpansionFactor(c)
                     runLog.debug(msg=f"      Component {c}, growFrac = {growFrac:.4e}")
                     c.height = growFrac * blockHeight
```
albeanth, line 336: a few lines below here have `self.linked.linkedComponents[c][0]`, shouldn't they get replaced with `self.linked.linkedComponents[c].lower`? 

same with `self.linked.linkedBlocks[b][0]`

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -28,82 +28,103 @@
...
+if TYPE_CHECKING:
+    from armi.reactor.components import Component
+    from armi.reactor.blocks import Block
+    from armi.reactor.assemblies import Assembly
 
-def getSolidComponents(b):
+
+def iterSolidComponents(b: "Block") -> Iterable["Component"]:
+    """Iterate over all solid components in the block."""
+    return filter(lambda c: not isinstance(c.material, material.Fluid), b)
```
albeanth, line 39: Tis fancier, yes. But is this better than looping approach that existed before? Maybe it's so it can be used in the filter function in `_findComponentLinkedTo`?

EDIT: I see your `see also` statement... more teaching!

albeanth (commented):
Some comments for you! Looking into the tests now, but wanted to get these comments off to you.

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -28,82 +28,103 @@
...
+
+
+def getSolidComponents(b: "Block") -> list["Component"]:
```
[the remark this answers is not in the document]
albeanth: Just pulled down your branch. I see they aren't actually strings. Or at least not rendering as such in vscode. I suppose this is what you're going for? But also odd since `Block` and such have been defined already. Maybe I'm off? 
image

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
albeanth: I see why this works now... You included `__getitem__` and `__setitem__` for backwards compatability. Which is cool. But what's stopping us from just changing it? 

Not saying you should remove the `__getitem__` and `__setitem__` implementation (because who knows what's out there), but rather just making everything consistent within this file.

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -857,9 +855,18 @@ def setUp(self):
         self.a = buildTestAssemblyWithFakeMaterial(name="HT9")
 
     def test_getSolidComponents(self):
+        """Show that getSolidComponents produces a list of solids, and is consistent with iterSolidComponents."""
         for b in self.a:
-            for c in getSolidComponents(b):
+            solids = getSolidComponents(b)
+            ids = set(map(id, solids))
```
albeanth, line 861: oh man... `map`s.... to the google lol

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -1266,3 +1273,109 @@ def linearExpansionPercent(self, Tk=None, Tc=None):
...
+        self.assertIs(self.link.lower, self.LOWER_BLOCK)
+        self.link[1] = None
+        self.assertIsNone(self.link.upper)
+        self.link[1] = self.UPPER_BLOCK
+        self.assertIs(self.link.upper, self.UPPER_BLOCK)
+        with self.assertRaises(AttributeError):
+            self.link[-1] = None
+
+    def test_iteration(self):
+        """Test the ability to iterate over the items in the link."""
```
albeanth, line 1310: just curious, why do we need to have the ability to iterate here? We only ever access AxialLink.lower or AxialLink.upper, right?

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -217,15 +246,21 @@ def _setTargetComponents(self, setFuel):
             elif b.hasFlags(Flags.PLENUM) or b.hasFlags(Flags.ACLP):
```
albeanth, line 246: I can't comment above, but the content in the `if b.p.axialExpTargetComponent` block above should be replaced by a call to `_setExpansionTarget`.

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
drewj-tp: Yes but that's not what we have implemented

https://github.com/terrapower/armi/blob/15e313850c6f71d47f3fde5ec1d047a9fadb7c43/armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py#L49

so you could have A be a subclass of the type of B and be okay.

I'm okay if we want to make that change to `type(A) is type(B)`

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
drewj-tp: Okay that tracks and matches the code

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
drewj-tp: This is like saying `Comp` can be used in place of `Block` or `Component`. But we can't use `typing.Union[Block, Component]` because we want `lower` and `upper` to be the same type.

If we did `Comp = typing.Union[Block, Component]`, this would be valid (from a type checker)
```python
AxialLink(lower=HexBlock(...), upper=Hexagon(...))
```
which is not what we want

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -57,118 +73,187 @@ def _determineLinked(componentA, componentB):
...
+
+        Parameters
+        ----------
+        blocks : sequence of armi.reactor.blocks.Block
+            Ordered sequence of blocks from bottom to top. Could just as easily be an
+            :class:`armi.reactor.assemblies.Assembly`.
+
+        Returns
+        -------
+        dict of Block -> AxialLink
```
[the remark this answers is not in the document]
drewj-tp: Nope!

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -57,118 +73,187 @@ def _determineLinked(componentA, componentB):
...
+            return cls._getLinkedBlocks(blocks, nBlocks)
+        raise ValueError("No blocks passed. Cannot determine links")
+
+    @staticmethod
+    def _getLinkedBlocks(
+        blocks: typing.Sequence[Block], nBlocks: int
+    ) -> dict[Block, AxialLink[Block]]:
+        # Use islice to avoid making intermediate lists of subsequences of blocks
+        lower = itertools.chain((None,), itertools.islice(blocks, 0, nBlocks - 1))
+        upper = itertools.chain(itertools.islice(blocks, 1, None), (None,))
```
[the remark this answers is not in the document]
drewj-tp: Short answer here, long answer in the next comment.
- iterating over lower produces `None`, `blocks[0]`, `blocks[1]`, ..., `blocks[-2]`
   - nothing is below the first block
- iterating over upper produces `blocks[1]`, `blocks[2]`, ..., `blocks[-1]`, `None`
   - nothing is above the last block
- the things from [`islice`](https://docs.python.org/3/library/itertools.html#itertools.islice) can be seen as a normal list slice `blocks[1:None]` to produce the second item through the last item
- [`chain`](https://docs.python.org/3/library/itertools.html#itertools.chain) strings together iterators

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -57,118 +73,187 @@ def _determineLinked(componentA, componentB):
...
+        links = {}
+        for low, mid, high in zip(lower, blocks, upper):
+            links[mid] = AxialLink(lower=low, upper=high)
```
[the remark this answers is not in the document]
drewj-tp: No problem! 

Disregarding the first and last block, we could write something like
```python
for i in range(1, len(blocks) - 1)
    below= blocks[i - 1]
    middle = blocks[i]
    above = blocks[i + 1]
    links[middle] = AxialLink(below, above)
```
And then the first and last block need special care for indicating nothing is below or above them. The `lower` and `upper` iterators take care of that. If they were lists, `lower[i] = links[blocks[i]].lower`, regardless of the location of `blocks[i]`.

So now we have three "iterables of blocks" between `lower`, `blocks`, and `upper`. `lower` and `upper` are not lists, but we probably could have made them with some list-slicing and the use of `chain`. I chose not to do list-slices because we just need to iterate over blocks, not all the functionality (and memory usage and construction time) of a list of blocks that slicing would produce

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
drewj-tp: [`functools.partial`](https://docs.python.org/3/library/functools.html#functools.partial) is a way to do a lot of things, but I usually use it to make a function that I'm going to call a lot with similar arguments. This could have been replaced in the `filter` call with a lambda expression
```python
for otherComp in filter(lambda other: self.areAxiallyLinked(c, other), iterSolidComponents(otherBlock)):
...
```
The benefit (or at least my perceived benefit) is we don't need to "re-lookup" via `.`-access `self.areAxiallyLinked` at each step through the iteration.

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
drewj-tp: Yeah I can remove the checklist here since we reference it in the other function.

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
drewj-tp: > a few lines below here have self.linked.linkedComponents[c][0], shouldn't they get replaced with self.linked.linkedComponents[c].lower?

> same with self.linked.linkedBlocks[b][0]

Those are good spots to update. I'll make use of the new link dataclass.

> You included __getitem__ and __setitem__ for backwards compatability. Which is cool. But what's stopping us from just changing it?

We should definitely use the `.lower` and `.upper` attributes. Just missed this block here

armi/reactor/converters/axialExpansionChanger/expansionData.py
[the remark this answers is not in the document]
drewj-tp: > I see they aren't actually strings. Or at least not rendering as such in vscode

> But also odd since Block and such have been defined already

They're type hints for things not included in the **runtime** environment. The `Component` and `Block` definitions are imported under this [`typing.TYPE_CHECKING` guard](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING)
```python
if TYPE_CHECKING:
    from armi.reactor.blocks import Block
    from armi.reactor.assemblies import Assembly
```
so importing this file is not going to import `Component` nor `Block` into your environment. But type checkers, like pylance from vscode, **do** set `TYPE_CHECKING` to be truthy, so they import those objects so we developers can do type hinting on helpful things.

I try to put things that I only need to import for type checking under this guard to avoid the (sometimes small, sometimes not small) hit of importing modules at run time just for type checking. We don't _really_ need to dig into `armi.reactor.blocks` and bring out `Block` at run time for this module to work.

Happy to go either way, just let me know

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -28,82 +28,103 @@
...
+    r"""Data container for axial expansion.
+
+    The primary responsibility of this class is to determine the axial expansion factors
+    for each component in the assembly. Expansion factors can be compute from the component
+    temperatures in :meth:`computeThermalExpansionFactors` or provided directly to the class
+    via :meth:`setExpansionFactors`.
+
+    This class relies on the concept of a "target" expansion component for each block. While
+    components will expand at different rates, the final height of the block must be determined.
+    The target component, determined by :meth:`determineTargetComponents`\, will drive the total
```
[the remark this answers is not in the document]
drewj-tp: I've definitely had weird things happen sometimes with sphinx not liking non-space characters after a backtick. We'll take it!

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -217,15 +246,21 @@ def _setTargetComponents(self, setFuel):
...
             else:
                 self.determineTargetComponent(b)
 
-    def determineTargetComponent(self, b, flagOfInterest=None):
-        """Determines target component, stores it on the block, and appends it to
-        self._componentDeterminesBlockHeight.
+    def determineTargetComponent(
+        self, b: "Block", flagOfInterest: Optional[Flags] = None
+    ) -> "Component":
+        """Determines target component, or the component who's expansion will determine block height.
```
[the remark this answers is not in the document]
drewj-tp: Hmm maybe something in between. I'll push a new change and you let me know what you think

armi/reactor/converters/tests/test_axialExpansionChanger.py
drewj-tp: read this as `ids = {id(s) for s in solids}`

armi/reactor/converters/tests/test_axialExpansionChanger.py
drewj-tp: Back compat? I think it was needed in earlier versions but let me see about removing `AxialLink.__iter__`

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
albeanth: After discussion - go with `type(A) is type(B)`

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
albeanth: Let's not worry about backwards compatibility here. This is a big improvement and I _highly_ doubt that anyone outside us is leveraging this in some unknown way. It's a safe change to make and if I'm wrong, I'll take heat!

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
albeanth: https://github.com/terrapower/armi/pull/1920/commits/74fcd300def5109ae2a91f91743424904683ca88

armi/reactor/converters/tests/test_axialExpansionChanger.py
albeanth: https://github.com/terrapower/armi/pull/1920/commits/1cc156a9cd0e6bf0c8c6fa33003461bd48636911

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
albeanth: https://github.com/terrapower/armi/pull/1920/commits/1cc156a9cd0e6bf0c8c6fa33003461bd48636911

albeanth:
@john-science this PR is apart of a larger downstream internal change, so please hold off on merging this. thank you!

albeanth:
> Holler when you guys want this merged. I see it is doubly approved.

This is getting run on additional downstream testing. If all goes well, it should be good to go today. I'll ping you when we're ready.

john-science:
@drewj-tp I see this as approved, and we have a release coming up. Any status on when this might get merged?

albeanth:
> @drewj-tp I see this as approved, and we have a release coming up. Any status on when this might get merged?

Same status as yesterday. Blocked by downstream testing going awry. @drewj-tp, @mgjarrett, and myself need to troubleshoot today.[

john-science:
bump

drewj-tp:
> bump

I think today's the day!

PR #1925, opened 2024-10-03
Removing unnecessary column in print-out

### LINKED ISSUE #1192: "Property Data was Modified from the Source?"
During a run, ARMI spits out a table into the stdout that looks like this:
image

I don't usually think about this, but it caught my eye because all of the columns in the table say `False`, but I am FOR SURE modifying the source data in my current run.

So I dug into this a bit deeper. This is where that table is generated:
https://github.com/terrapower/armi/blob/3e47ec6e6359870866502c9356747cf1da4f9515/armi/reactor/blueprints/reactorBlueprint.py#L256-L272

It looks like the value is _always_ set to `False`. So really the output there is a lie.

So do we want to fix this, or just get rid of it completely?

john-science:
## What is the change?

There is an incorrect and unused column in a summary table in the reactor blueprints, and this PR fixes it.

I could have fixed the column to be correct, but since no one but Chris has EVER noticed it was broken, I infer the column is unused.

close #1192

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/reactorBlueprint.py
+++ b/armi/reactor/blueprints/reactorBlueprint.py
@@ -144,7 +144,7 @@ class SystemBlueprint(yamlize.Object):
         """
         from armi.reactor import reactors  # avoid circular import
 
-        runLog.info("Constructing the `{}`".format(self.name))
+        runLog.info(f"Constructing the `{self.name}`")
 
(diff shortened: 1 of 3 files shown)


keckler (approved):
I think this is the right way to go. Thanks!

PR #1926, opened 2024-10-04
Fixing HexBlock rotation in plotBlockDiagram

### LINKED ISSUE #1421: `plotBlockDiagram()` doesn't respect block orientation

john-science:
## What is the change?

Fixed a bug in `plotBlockDiagram()`.

The method `plotBlockDiagram()` incorrectly plots `HexBlock`s if they are rotated to the "corners up" rotation.

close #1421

---

## Checklist


keckler:
First question without even looking at the code -- Can we get a picture of the correct plots?

armi/reactor/tests/test_blocks.py
```diff
@@ -1876,30 +1872,33 @@ def test_negativeVolume(self):
...
         r = tests.getEmptyHexReactor()
         a = makeTestAssembly(1, 1)
-        a.add(self.HexBlock)
+        a.add(self.hexBlock)
         loc1 = r.core.spatialGrid[0, 1, 0]
         r.core.add(a, loc1)
 
+    def test_cornersUp(self):
+        self.assertTrue(self.hexBlock.cornersUp())
+
```
keckler, line 1901: Can we add another test for the case where this would return `False`? Otherwise it is not clear that the new `Block.cornersUp` method is working.

armi/utils/plotting.py
```diff
@@ -1296,11 +1295,10 @@ def _makeComponentPatch(component, position, cold):
                 patch = matplotlib.patches.Polygon(corners, fill=True)
                 blockPatch.append(patch)
         else:
-            # Just make it a hexagon...
+            # Just make it a hexagon
             blockPatch = matplotlib.patches.RegularPolygon(
                 (x, y), 6, radius=component.getDimension("op", cold=cold) / math.sqrt(3)
```
keckler, line 1300: I think this one needs to be rotated as well.

armi/utils/plotting.py
```diff
@@ -1224,27 +1223,27 @@ def _makeBlockPinPatches(block, cold):
...
     ------
-        blockPatch: List
-            A list of Patch objects that together represent a component in the diagram.
+    blockPatch: list
+        A list of Patch objects that together represent a component in the diagram.
 
     Notes
     -----
-    Currently accepts components of shape DerivedShape, Helix, Circle, or Square
+    Currently accepts components of shape DerivedShape, Circle, Helix, Hexagon, or Square
```
keckler, line 1246: Not related to your PR, but this line says that the method accepts `DerivedShape`s. It does not appear to me that this is true.

keckler (commented):
I think also this chunk might need to be made aware of the orientation:
https://github.com/terrapower/armi/pull/1926/files#diff-0c52725b9d66a2bcefb306c31e4a9479fe8c206453c22b3e4cd268a1efef3c16R1181-R1185

armi/utils/plotting.py
john-science: done

armi/utils/plotting.py
john-science: eone

armi/reactor/tests/test_blocks.py
john-science: done

keckler:
@john-science Thank you for fixing those few comments. Please also note that there are a couple more things that would be good before approving this PR:
1. https://github.com/terrapower/armi/pull/1926#pullrequestreview-2348839742
2. https://github.com/terrapower/armi/pull/1926#issuecomment-2394104416

armi/reactor/blocks.py
```diff
@@ -1785,6 +1785,13 @@ def coords(self):
             round(y, units.FLOAT_DIMENSION_DECIMALS),
         )
 
+    def cornersUp(self):
+        """Determine if the hex shape of is corners up or flats up, in relation to the Y axis."""
+        if self.spatialGrid is None:
+            return None
+
+        return self.spatialGrid._unitSteps[0][1] != 0.0
```
drewj-tp, line 1793: What about passing off to `HexGrid.cornersUp`?

https://github.com/terrapower/armi/blob/9dbe15c621ecac1060bdfde9715cf87697c24208/armi/reactor/grids/hexagonal.py#L107-L115

```suggestion
        return self.spatialGrid.cornersUp
```

armi/reactor/blocks.py
```diff
@@ -1785,6 +1785,13 @@ def coords(self):
             round(y, units.FLOAT_DIMENSION_DECIMALS),
         )
 
+    def cornersUp(self):
+        """Determine if the hex shape of is corners up or flats up, in relation to the Y axis."""
+        if self.spatialGrid is None:
+            return None
```
drewj-tp, line 1791: I guess this is false-like? But returning `None` from a bool-like method feels weird. 

Would it make more sense to return False or error here?

drewj-tp (commented):
I'm currently poking around grid like things and some things here overlapped

armi/reactor/blocks.py
john-science: A Python `None` is a null value. It is just as valid to use `None` for a null value when a return type is an `int`, `numpy.ndarray` or `bool`.

Seems standard enough to return an empty value from a method.

What would you rather do?  I don't want to return an error, because this does not seem like a scenario in which I want to kill a simulation.

john-science: hmmm

john-science:
@drewj-tp Well, in defense of your idea, I am still getting this:

blockDiagram27

It appears the bug for the hexagonal shaped composites was quite easy to fix. But the coordinates for the pins are just wrong from the start. (There's nothing I can do in the plotting routine to fix that.)

john-science:
This PR is temporarily on hold for another PR: https://github.com/terrapower/armi/pull/1947

I believe my solution in this was over-engineered to work AROUND the problem present in the other PR. Once that other PR merges, I should be able to simplify this PR.

john-science:
@keckler and @drewj-tp 

Here I plot `armiRunSmallest.yaml` with the default `hex_corners_up`:

hex_corners_up

and here I do the same thing, but alter the reactor to be "flats up":

hex_flats_up

So, after the recent bug fix, I think we finally have success!

Whew.

keckler (approved):
Approved, thank you.

Please update the description in the top comment of the PR. In particular, that `HexBlock.cornersUp()` method is no longer added.

PR #1927, opened 2024-10-04
Adding Block.getInputHeight

### LINKED ISSUE #1924: Provide a method for a block to report its input height
> Our team opinion here is we need a method that can go up to the blueprints and get the T=0 heights from a block.

This method would have other benefits outside this plot, e.g., axial thermal expansion between hot and as fabricated / neutral stress heights of blocks.

_Originally posted by @drewj-tp in https://github.com/terrapower/armi/issues/1158#issuecomment-2392102394_

drewj-tp:
## What is the change?

The new method `Block.getInputHeight()` digs up the composite tree to find the Reactor, or something with blueprints (usually the reactor). Then finds the parent Assembly's blueprint design and extracts the height at the index of this block.

If we don't find any blueprints, throw an error.

There are a couple use cases where you want to know the "as fabricated" height of a `Block`. We can't always rely on `Block.p.heightBOL` parameter because that may or may not get expanded shortly after construction via blueprints.

Closes #1924 

---

## Checklist


armi/reactor/blocks.py
```diff
@@ -1732,6 +1732,35 @@ def getUraniumMassEnrich(self):
...
+            usually stored on the reactor object, which is typically an ancestor of
+            the block (block -> assembly -> core -> reactor). However, this may be the case
+            when creating blocks from scratch in testing where the entire composite
+            tree may not exist.
+        """
+        ancestorWithBp = self.getAncestor(
+            lambda o: getattr(o, "blueprints", None) is not None
+        )
+        if ancestorWithBp is not None:
+            parent: composites.Composite = self.parent
```
john-science, line 1756: seems like we could leave this line out. `self.parent` is only called twice.

#not-blocking

armi/reactor/tests/test_blocks.py
```diff
@@ -1797,6 +1798,45 @@ def test_getReactionRates(self):
...
+            return assem
+        raise KeyError(nameInBP)
+
+    def test_foundReactor(self):
+        """Test the input height is pullable from blueprints."""
+        r = loadTestReactor()[1]
+        msg = "Input height from blueprints differs. Did a blueprint get updated and not this test?"
+
+        # Grab a block from an assembly, so long as we have the height
+        assem = self._getAssemblyWithType(r.core, "igniter fuel")
```
john-science, line 1814: ARMI already provides:

```python
r.core.getFirstAssembly(Flags.IGNITER)
```
or perhaps:

```python
r.core.getFirstAssembly(Flags.IGNITER | Flags.FUEL)
```

It _feels_ to me like that would be simpler than added a whole helper method to this class that does the same thing.

Unless I'm missing something, and that didn't work in this case?

john-science (approved):

armi/reactor/tests/test_blocks.py
drewj-tp: I think I added this helper because some of the other methods didn't quite work but I don't think I tried this one. It works so I'll roll it in

john-science (approved):

PR #1929, opened 2024-10-04
Fixing edge case in assemblyBlueprint._checkParamConsistency

### LINKED ISSUE #1223: Some material modifications are allowed to be the wrong length
We currently have a check that tries to ensure that material modifications in the blueprints are of the same length as the number of blocks in that assembly:

https://github.com/terrapower/armi/blob/db8dc8c28fbea8614cac6c2d9b39a8c624838719/armi/reactor/blueprints/assemblyBlueprint.py#L221-L243

There is apparently a corner case that isn't caught. 

If you have a `by block` and `by component` material modification with the same name on a given block, only one of those will be checked. This is because the method above builds a dict of which things to check, and the same dict keys get reused for `by block` and `by component` mods. So the first one gets overwritten by the second.

john-science:
## What is the change?

Fixing edge case in `assemblyBlueprint._checkParamConsistency()`.

I don't know how often this edge case is caught in the real world, but it happened at least once, and I hate that.

close #1223

## Further Testing

I believe we should run downstream testing before merging this PR. I am a touch worried this happens in the wild.

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/assemblyBlueprint.py
+++ b/armi/reactor/blueprints/assemblyBlueprint.py
@@ -25,10 +25,10 @@ import yamlize
 from armi import getPluginManagerOrFail
 from armi import runLog
 from armi.reactor import assemblies
-from armi.reactor.flags import Flags
+from armi.reactor import grids
 from armi.reactor import parameters
 from armi.reactor.blueprints import blockBlueprint
-from armi.reactor import grids
+from armi.reactor.flags import Flags
 from armi.settings.fwSettings.globalSettings import CONF_INPUT_HEIGHTS_HOT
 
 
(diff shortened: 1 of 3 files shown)


keckler (approved):
No comments. This looks good to me. I think it accurately captures the edge case that was reported. Thanks!

armi/reactor/blueprints/assemblyBlueprint.py
```diff
@@ -280,29 +279,35 @@ def _createBlock(self, cs, blueprint, bDesign, axialIndex):
...
             "heights": self.height,
             "xs types": self.xsTypes,
         }
 
-        for mod in [self.materialModifications] + list(
-            self.materialModifications.byComponent.values()
-        ):
-            for modName, modList in mod.items():
-                paramName = "material modifications for {}".format(modName)
+        # check high-level mat mods
```
john-science, line 291:
```suggestion
        # check by-block mat mods
```

PR #1933, opened 2024-10-06
Ensuring we only calculate smear density for pinned blocks

### LINKED ISSUE #318: No component matched warning for non-pin type fuel
When using non-pin type fuel, the user is warned on "No component matched" for Flags.WIRE and Flags.CLAD. It is not clear where the warning is originating from or what can be done to fix it. 

This is occurring in a non-standard fast reactor benchmark (BFS-73-1), where there are no pins/clad defined and the core uses a stack up of "disks" or "plates".

```
[info] Will expand CR, FE, H, NI, SI, TI, O, NA, AL, MN elementals to have natural isotopics
[info] Constructing assembly `Inner Fuel 0 Rods`
[warn] No component matched Flags.WIRE in <Homogenized UO2 Aluminum Can Reflector Pellet: 0 SS Bars block-bol-000 at ExCore XS: B BU GP: A>. Returning None
[warn] No component matched Flags.CLAD in <Homogenized UO2 Aluminum Can Reflector Pellet: 0 SS Bars block-bol-000 at ExCore XS: B BU GP: A>. Returning None
[info] Constructing assembly `Inner Fuel 1 Rods`
[info] Constructing assembly `Inner Fuel 2 Rods`
```

john-science:
## What is the change?

Checking that the `Block` has pins before calculating smear density.

The `Block.getSmearDensity()` method only works for pinned blocks, but it does not check for pins, so users with un-pinned blocks will get mysterious and confusing warnings. This PR removes those warnings, until such a time as we have something like a `PinnedBlock` class.

close #318 

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -205,9 +205,10 @@ class Block(composites.Composite):
         all the area is fuel, it has 100% smear density. Lower smear density allows more room for
         swelling.
 
-        .. warning:: This requires circular fuel and circular cladding. Designs that vary
-            from this will be wrong. It may make sense in the future to put this somewhere a
-            bit more design specific.
+        Warning
+        -------
+        This requires circular fuel and circular cladding. Designs that vary from this will be
+        wrong. It may make sense in the future to put this somewhere a bit more design specific.
 
         Notes
         -----
@@ -225,15 +226,15 @@ class Block(composites.Composite):
 
         Returns
         -------
-        smearDensity : float
-            The smear density as a fraction
+        float
+            The smear density as a fraction.
         """
         fuels = self.getComponents(Flags.FUEL)
         if not fuels:
-            # Smear density is not computed for non-fuel blocks
+            # smear density is not computed for non-fuel blocks
             return 0.0
         elif not self.getNumPins():
-            # Smear density is only defined for pinned blocks
+            # smear density is only defined for pinned blocks
             return 0.0
 
         circles = self.getComponentsOfShape(components.Circle)
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -230,7 +230,11 @@ class Block(composites.Composite):
         """
         fuels = self.getComponents(Flags.FUEL)
         if not fuels:
-            return 0.0  # Smear density is not computed for non-fuel blocks
+            # Smear density is not computed for non-fuel blocks
+            return 0.0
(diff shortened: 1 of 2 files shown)


armi/reactor/blocks.py
```diff
@@ -230,7 +230,11 @@ def getSmearDensity(self, cold=True):
         """
         fuels = self.getComponents(Flags.FUEL)
         if not fuels:
-            return 0.0  # Smear density is not computed for non-fuel blocks
+            # Smear density is not computed for non-fuel blocks
+            return 0.0
+        elif not self.getNumPins():
+            # Smear density is only defined for pinned blocks
+            return 0.0
```
drewj-tp, line 237: Tests? Maybe we already have tests for that touch these conditionals 🤷

drewj-tp (commented):
Easy enough. Changes look good. Would appreciate a test or evidence we already test this condition but probably good regardless

armi/reactor/blocks.py
john-science: Totally fair.  Done.

armi/reactor/tests/test_blocks.py
```diff
@@ -395,6 +395,21 @@ def test_getSmearDensityMultipleLiner(self):
...
+
+        # show smear density is only defined for pinned fuel blocks
+        b1 = blocks.HexBlock("TestFuelHexBlock")
+        b1.setType("fuel")
+        b1.p.nPins = 0
+        fuel = components.Circle(
+            "fuel", "UZr", Tinput=25.0, Thot=25.0, od=0.84, id=0.6, mult=0
+        )
+        b1.add(fuel)
+        self.assertEqual(b1.getSmearDensity(), 0.0)
```
drewj-tp, line 411: This test is confusing until you catch the `nPins = 0` line. I see a circle of UZr, that looks like a pin and we should have a non-zero smear density.

What about making it a hex of sodium? Just to make it even more clear there should be no fuel nor fuel pins in the block?

john-science: Oh, wait.

Actually, the second "if" statement I am testing is that the it IS a fuel block, but it ISN'T pinned. So, the block needs to be defined as "fuel", but not "pinned", that's the edge case this PR is about.

drewj-tp: Oooooo gotcha gotcha. Yep, that makes sense.

PR #1934, opened 2024-10-06
Relaxing copyInterfaceInputs to not require a valid Setting

### LINKED ISSUE #772: copyInterfaceInputs should not require a valid settings object

### LINKED ISSUE #825: Either move valid settings check to specifyInputs, or address and close Issue #772
In `case.py`'s `copyInterfaceInputs`, there is a check for a valid setting:

https://github.com/terrapower/armi/blob/9fcc89393d585d99cd8c99fbfed841844284bfe9/armi/cases/case.py#L803-L810

This check being in `copyInterfaceInputs` makes it untestable. It should either get removed (which addresses Issue #772) or be moved to `specifyInputs`.

john-science:
## What is the change?

Relaxing the case method `copyInterfaceInputs()` to not require a valid Setting.

There are multiple open requests to allow a case to copy inputs for things that are not mapped to "valid" settings.

close #772
close #825

---

## Checklist

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -882,6 +882,7 @@ def copyInterfaceInputs(
     enables developers to add new inputs in a plugin-dependent/ modular way.
 
     This function should now be able to handle the updating of:
+
       - a single file (relative or absolute)
       - a list of files (relative or absolute), and
       - a file entry that has a wildcard processing into multiple files.
@@ -900,12 +901,12 @@ def copyInterfaceInputs(
         The source case settings to find input files
     destination : str
         The target directory to copy input files to
-    sourceDir : str (optional)
+    sourceDir : str, optional
         The directory from which to copy files. Defaults to cs.inputDirectory
 
     Returns
     -------
-    newSettings : dict
+    dict
         A new settings object that contains settings for the keys and values that are
         either an absolute file path, a list of absolute file paths, or the original
         file path if absolute paths could not be resolved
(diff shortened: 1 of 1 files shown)


armi/cases/case.py
```diff
@@ -936,6 +936,8 @@ def copyInterfaceInputs(
                 except NonexistentSetting(key):
                     runLog.debug(f"{key} is not a valid setting; continuing on anyway.")
                     label = key
+            else:
```
opotowsky, line 939: I'm not sure about this approach. I'll try to find time this week to test and discuss this.

john-science: Okay, I added the check below, like we discussed.

opotowsky: nice!

PR #1935, opened 2024-10-07
Adding logic so HexAssemblies are hexagonal

### LINKED ISSUE #1528: Add some logic so HexAssemblies are hexagonal
Right now, the classes `HexAssembly` and `CartesianAssembly` are just empty slates:

https://github.com/terrapower/armi/blob/8b9873cfe9df8dac6a26b28a7520c45b65fa3a50/armi/reactor/assemblies.py#L1229-L1237

https://github.com/terrapower/armi/blob/8b9873cfe9df8dac6a26b28a7520c45b65fa3a50/armi/reactor/assemblies.py#L1240-L1241

But they give people the assumption they can trust these assemblies have a shape, or the blocks inside them do.  But obviously no.

I recommend that we override the superclass method `Assembly.add()` here to check the shape of the Blocks added to assemblies of these types. That will give users some trust, anyway, that instances of these classes are what you'd think at first glance.

john-science:
## What is the change?

Added some detail to classes `HexAssembly` and `CartesianAssembly` to ensure only the correct blocks are added to them.

There was a complaint that these classes were just placeholders.

close #1528

---

## Checklist


john-science:
@jeffbaylor @alexhjames Just FYI.

jakehader:
Hi @john-science - how we feel about removing the Hex, Cartesian, and THRZ classes? We used to have a bunch of logic in these subclasses but they were removed in favor of minimizing. I don't think an assembly "shape" truly matters. I think that THRZ is the odd one here and it might be better to clean that one up to remove shape dependencies altogether.

Really, an assembly should just be a stacked list of 1D blocks, where most of the magic happens.

drewj-tp:
> Hi @john-science - how we feel about removing the Hex, Cartesian, and THRZ classes? We used to have a bunch of logic in these subclasses but they were removed in favor of minimizing. I don't think an assembly "shape" truly matters. I think that THRZ is the odd one here and it might be better to clean that one up to remove shape dependencies altogether.

@jakehader I think this is good fodder for https://github.com/terrapower/armi/discussions/1896. I also think it makes sense to have shapiness on assemblies too. Even if to just check that you don't add an incompatible block-type to an assembly. Finding a cartesian block amongst a bunch of hex blocks feels like something that could be desirable to avoid

> Really, an assembly should just be a stacked list of 1D blocks, where most of the magic happens.

Agree to an extent. If we re-named the _base_ assembly class to be something like `BlockStack`, that would allow reactors that don't have assemblies (PBRs, MSRs) to still use _some_ of the API even though they don't have assemblies in the SFR/LWR/HTGR sense.

See maybe #1030?

john-science:
Okay, perhaps I will re-work this PR so that we ONLY have `Assembly`, which has an attribute `a.blockType`. And if `.blockType` is set (optional), then that `Assembly` will only accept `Block`s of that type.

My _guess_ is that is a small change that can happen only in ARMI and not involve any downstream changes. And then we can leave the bigger, high-level changes to the "abstraction" Discussion.

If it turns out such a change will require many downstream changes, I will revisit the idea.

john-science:
@jakehader @drewj-tp Okay, per our discussion here and over on [here](https://github.com/terrapower/armi/discussions/1896), I modified my solution here. The generic `Assembly` class now just has a class attribute `.blockType` that can be `HexBlock` for `HexAssembly`.

This is closer to your end game.

However, the full end game of, say, removing `HexAssembly` , for instance, is way beyond the scope of this ticket. We are going to have a higher level discussion about ARMI's abstract classes, and hopefully that will go in the next release. (But not this one.)

armi/reactor/assemblies.py
```diff
@@ -43,6 +43,7 @@ class Assembly(composites.Composite):
     """
 
     pDefs = assemblyParameters.getAssemblyParameterDefinitions()
+    blockType = None
```
drewj-tp, line 46: Two suggestions:

1. Make this more clearly a non-public, class-specific variable that people should not try to edit / access. Something like `_BLOCK_TYPE`. We can add a [`ClassVar`](https://docs.python.org/3/library/typing.html#typing.ClassVar) type hint to further discourage instances modifying this type.
2. If it's public, can we add a docstring?

armi/reactor/assemblies.py
```diff
@@ -180,6 +180,11 @@ def add(self, obj: blocks.Block):
             are updated. The axial mesh and other Block geometry parameters are
             updated in ``calculateZCoords``.
         """
+        if self.blockType is not None and not isinstance(obj, self.blockType):
+            msg = f"Cannot add {obj} to this Assembly, it is not a {self.blockType}."
+            runLog.error(msg)
+            raise TypeError(msg)
```
drewj-tp, line 186: Combine this logic with the logic in `self.insert` into a method? `self._checkPotentialChild`?

Non-blocking request

armi/reactor/assemblies.py
```diff
@@ -199,6 +204,11 @@ def moveTo(self, locator):
 
     def insert(self, index, obj):
         """Insert an object at a given index position with the assembly."""
+        if self.blockType is not None and not isinstance(obj, self.blockType):
+            msg = f"Cannot insert {obj} to this Assembly, it is not a {self.blockType}."
+            runLog.error(msg)
+            raise TypeError(msg)
```
drewj-tp, line 210: Similar non-blocking request for `self._checkPotentialChild`

drewj-tp (changes requested):
I'm not sold on the composition vs. inheritance argument here. I think having a hex assembly and a Cartesian assembly makes sense (which I know is not the point but that seems to be part of the inspiration)

But that said, we need some way to delegate the "type of thing that can be added to an assembly" work to be unique for hex and cartesian assemblies. To that end, this is probably the minimal set of changes to accomplish that goal.

armi/reactor/assemblies.py
john-science: Done

armi/reactor/assemblies.py
john-science: Done.

armi/reactor/assemblies.py
john-science: I did (1), good call.

armi/reactor/assemblies.py
```diff
@@ -15,25 +15,26 @@
...
 Assemblies are collections of Blocks.
 
-Generally, blocks are stacked from bottom to top.
+Generally, Blocks are stacked from bottom to top.
 """
 import copy
 import math
 import pickle
 from random import randint
+from typing import ClassVar, Type
```
drewj-tp, line 24: Adding to make the `Assembly._BLOCK_TYPE` hint complete
```suggestion
from typing import ClassVar, Optional, Type
```

armi/reactor/assemblies.py
```diff
@@ -42,15 +43,15 @@ class Assembly(composites.Composite):
     Append blocks to add them up. Index blocks with 0 being the bottom.
     """
 
+    _BLOCK_TYPE: ClassVar[Type[blocks.Block]] = None
```
drewj-tp, line 46: If we're allowing `self._BLOCK_TYPE` to be `None`, this helps codify that
```suggestion
    _BLOCK_TYPE: ClassVar[Optional[Type[blocks.Block]]] = None
```

armi/reactor/assemblies.py
```diff
@@ -87,14 +88,12 @@ def __lt__(self, other):
...
-        the operator returns ``False``.  This behavior may lead to some strange sorting
-        behavior when two or more Assemblies are being compared that do not live in the
-        same grid. It may be beneficial in the future to maintain the more strict behavior
-        of ArmiObject's ``__lt__`` implementation once the SFP situation is cleared up.
+        As with other ArmiObjects, Assemblies are sorted based on location. Assemblies are more
+        permissive in the grid consistency checks to accomodate situations where assemblies might be
+        children of the same Core, but not in the same grid as each other (like in the spent fuel
+        pool). In these situations, the operator returns ``False``.  This behavior may lead to some
+        strange sorting behavior when two or more Assemblies are being compared that do not live in
+        the same grid.
```
drewj-tp, line 96: 👍 🧹

drewj-tp (commented):
Thanks for incorporating my comments! Minor request and then this is good to go

drewj-tp (approved):

PR #1937, opened 2024-10-07
Transposing pinMgFluxes so that the leading dimension represents pin index

zachmprince:
## What is the change?

- Updating `blocks.Block.setPinMgFluxes` so the parameters get set assuming the leading dimension is pin index
- Modifying parameter descriptions to reflect this change.

This is in support of #1860 where it would be easier to rotate pin parameters if they were all consistent on which dimension represents the pin index.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -367,7 +367,7 @@ class Block(composites.Composite):
         if self.hasFlags(Flags.FUEL):
             pinFluxes = fluxes[(np.array(self.p.pinLocation) - 1)]
         else:
-            pinFluxes = fluxes
+            pinFluxes = fluxes[:]
 
         if gamma:
             if adjoint:
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -346,13 +346,10 @@ class Block(composites.Composite):
         """
         Store the pin-detailed multi-group neutron flux.
 
-        The [g][i] indexing is transposed to be a list of lists, one for each pin. This makes it
-        simple to do depletion for each pin, etc.
-
         Parameters
         ----------
-        fluxes : 2-D list of floats
-            The block-level pin multigroup fluxes. fluxes[g][i] represents the flux in group g for
+        fluxes : np.ndarray
+            The block-level pin multigroup fluxes. fluxes[i, g] represents the flux in group g for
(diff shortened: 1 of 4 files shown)


armi/reactor/blocks.py
```diff
@@ -362,28 +359,16 @@ def setPinMgFluxes(self, fluxes, adjoint=False, gamma=False):
...
-            if self.hasFlags(Flags.FUEL):
-                pinLoc = self.p.pinLocation[pinNum - 1]
-            else:
-                pinLoc = pinNum
-
-            for g in range(G):
-                thisPinFlux.append(fluxes[g][pinLoc - 1])
-            pinFluxes.append(thisPinFlux)
+        if self.hasFlags(Flags.FUEL):
+            pinFluxes = fluxes[(np.array(self.p.pinLocation) - 1)]
```
drewj-tp, line 368: Clever! We don't need to go over all pins, find their pin location, etc. if we use `pinLocation` for array indexing!

drewj-tp (changes requested):
Can you add a test for this method? We don't have any tests for it here, or else they would also likely need to be updated.

Understanding you were likely not the original author of this method. But it would be great to demonstrate that `Block.setPinMgFluxes` does what we want

armi/reactor/blocks.py
zachmprince: Thank you! Numpy can be beautiful when used to its potential.

zachmprince:
> Can you add a test for this method? We don't have any tests for it here, or else they would also likely need to be updated.
> 
> Understanding you were likely not the original author of this method. But it would be great to demonstrate that `Block.setPinMgFluxes` does what we want

Good suggestion. I modified the original test to be a bit more robust and actually show that the indexing is correct.

drewj-tp:
I still think this is a good change. But now that #1860 is closed and we're not rotating the parameter data vectors anymore, this is less vital. 

Good, but it does not impact the assembly rotation logic. So long as the structure of the data array is documented

zachmprince:
I would like to keep this change since it makes sense to me, regardless of its impact.

drewj-tp (approved):
I'll be sure to come back to this in light of the `pinLocation` related changes in #1939. But the transposition looks good!

Thanks for the tests too

PR #1940, opened 2024-10-08
Raising Error when loading inconsistent data from DB

mgjarrett:
## What is the change?

Raise a `ValueError` when loading data with special packing from a database and the size of the data does not match the length of the list of composites to which the data is supposed to be assigned.

If the length of `comps` and `data` did not match, there were two possibilities:

1. `if len(comps) < len(data)` the database load would fail and ARMI would crash. This is okay, since there is no risk of loading data incorrectly.
2. `if len(comps) > len(data)`, then the missing parameter values from `data` would be filled in with an empty string, `""`, because of the use of `itertools.zip_longest` with `fillvalue=""`. This is bad, because ARMI would silently fill in missing data with `""` without telling you, and the data that was assigned would likely be assigned to the wrong composite.

The length of the special `data` being unpacked should always match the number of composites to which it is being assigned. If these two things do not match, ARMI should raise a `ValueError` with a warning message and abort.

---

## Checklist


armi/bookkeeping/db/tests/test_database3.py
```diff
@@ -841,3 +841,23 @@ def test_readWriteRoundTrip(self):
...
+        # modify the HDF5 file to corrupt a dataset
+        with h5py.File(self.db.fileName, "r+") as hf:
+            circleGroup = hf["c00n01"]["Circle"]
+            circleMass = np.array(circleGroup["massHmBOL"][()])
+            badData = circleMass[:-1]
+            del circleGroup["massHmBOL"]
+            circleGroup.create_dataset("massHmBOL", data=badData)
+
+        with self.assertRaises(ValueError):
+            with database3.Database3(self._testMethodName + ".h5", "r") as db:
```
john-science, line 862:
```suggestion
            with database3.Database3(self.db.fileName, "r") as db:
```

john-science: Just trying to be consistent with line 854 above.

john-science (approved):

PR #1943, opened 2024-10-10
HexBlock.rotate updates child spatial locators

### LINKED ISSUE #1939: Update spatial locations after HexBlock.rotate
Replaces #1860 because a [decision](https://github.com/terrapower/armi/pull/1877#issuecomment-2400277764) has been made to not rotate data parameters. Instead

1. update the spatial location of components through rotation.
2. enforce that, for pin-like data parameters like `Block.p.linPowByPin`, the data at index `i` is defined at `Block.getPinCoords()[i]`,
3. setters like `Block.setPinPowers` are no longer needed and just set the data parameter directly.
4. discourage the usage of `Block.p.pinLocation`

drewj-tp:
## What is the change?

Calls to `HexBlock.rotate` update the spatial locator on all child components. This is necessary because you'd expect rotation to move things in space.

Tests are added showing
1. pin coordinates, as reported by `Block.getPinCoordinates` reflect the rotation
2. location objects, reported by `Component.spatialLocator` reflect the rotation

Other changes were made in support of this functionality. 

### What about `pinLocation`

It still exists but is not modified at all during calls to `HexBlock.rotate`. Routines that were using `pinLocation` for a pre/post mapping should be okay, as evidenced by the lack of changes to any `pinLocation` related tests (e.g., `setPinPowers`). The mapping is transparent because it remains a list of one-indexed "pin locations" such that `data[pinLocations[i] - 1]` is still `data[i]` since `pinLocations[i] == i + 1`.

### `HexGrid.rotateIndex`

To facilitate the rotation, I added `HexGrid.rotateIndex` to take in an index location, rotate it some number of 60-degree rotations, and return the new location. I drew a lot of information and support from https://www.redblobgames.com/grids/hexagons/ (hereafter RBG) that I'll try to synthesize here.

ARMI's hex grid indexing is very similar to the "axial" or "trapezoidal" coordinate system described in RBG. Some key differences

1. We use `i,j` to define location whereas RBG uses `q,r`. 
4. In the RBG examples, their global Y direction is opposite ours, i.e., our grid is flipped along the x-axis. This means a clockwise rotation in our coordinate system is a counterclockwise rotation in theirs, and vice-versa.

Rotation of items on a hex grid is almost trivial using what RBG refers to as a cubic coordinate system, or where a point on the 2-D grid can be defined by three coordinates: `q, r, s`. The conversion between axial and cube just means computing `s = -q - r`.

60-degree rotation in this system is created by shuffling and negating the points. Following their examples, starting at location `(q, r, s)`, increasing CW rotations would place a cell at

1. `(-r, -s, -q)`
2. `(s, q, r)`
5. `(-q, -r, -s)`
6. and so on.

After three rotations, we have gone 180 degrees so it tracks that our locations is essentially the same coordinates but opposite sign.

This is tested by rotating a point some amount, and comparing the XY coordinates in the cartesian plane according to a rotation matrix
```
[[ cos(theta), -sin(theta)]
 [ sin(theta),   cos(theta)]]
```

### `Block.getPinLocations` for obtaining index locations

Might want to change the name to not use "location" because that feels confusing w/ `getPinCoordinates`. Maybe `getPinIndices`? `getPinIndexLocations`?

Anyway.  Some talk with internal stakeholders implied, to me, that mapping between ARMI's pin ordering and ordering in other applications would be more suited with integer indices rather than Cartesian XY coordinates. So `Block.getPinLocations` will produce all the `IndexLocation` objects for pins.

### `Block.getPinCoordinates` returns numpy array

It felt odd to get a list of arrays? This should be back compatible for things expecting list of arrays, as the `getPinCoordinates` test was not updated to use `(N, 3)` array features until later
- Returns array in 1c603a2d
- Test updated in 0a599b27

I guess not fully back compatible if you're doing pop/append/index operations on the coordinates... I can revert this change if we want to save it for later. Just made sense as a code-cleanup operation.

Closes #1939 

## Personal Todo

- [x] Docs on rotation, and how to find where data like `linPowByPin` exists in space
- ~~[ ] Remove usage of `pinLocation` in pin data setters like `setPinPowers` (could be follow on since it still works, it's just unnecessary)~~ Not necessary at this time

---

## Checklist


drewj-tp:
There is at least one project that is slightly broken b/c `Block.getPinCoordinates` returns an array and not a list. That's a trivial fix for me. Otherwise, this is ready for review (but not merge just yet)

john-science:
@drewj-tp Do you expect your solution here to need changing after the sub-block grid fix? https://github.com/terrapower/armi/pull/1947

drewj-tp:
> @drewj-tp Do you expect your solution here to need changing after the sub-block grid fix? #1947

I'll check but I don't think so. The rotation logic shoves most of the hard work onto the grid, so we should be okay?

zachmprince (changes requested):
Implementation is awesome! It seems to include everything we talked about for block rotation functionality. I just had few minor suggestions related to documentation.

armi/reactor/blocks.py
```diff
@@ -2009,10 +2049,36 @@ def rotate(self, rad):
 
         """
         rotNum = round((rad % (2 * math.pi)) / math.radians(60))
-        self._rotatePins(rotNum)
+        self._rotateChildLocations(rad, rotNum)
+        self.p.orientation[2] += rotNum * 60
         self._rotateBoundaryParameters(rotNum)
         self._rotateDisplacement(rad)
 
+    def _rotateChildLocations(self, radians: float, rotNum: int):
```
john-science, line 2057: Could use a one-sentence docstring. Just imagine someone is seeing this method for the first time 4 years from now, and they don't know about this PR. Broad strokes.

armi/reactor/blocks.py
```diff
@@ -2009,10 +2049,36 @@ def rotate(self, rad):
...
+                newLocations = list(map(locationRotator, c.spatialLocator))
+                c.spatialLocator = grids.MultiIndexLocation(self.spatialGrid)
+                c.spatialLocator.extend(newLocations)
+            elif isinstance(c.spatialLocator, grids.CoordinateLocation):
+                oldCoords = c.spatialLocator.getLocalCoordinates()
+                newXY = rotationMatrix.dot(oldCoords[:2])
+                newLocation = grids.CoordinateLocation(
+                    newXY[0], newXY[1], oldCoords[2], self.spatialGrid
+                )
+                c.spatialLocator = newLocation
```
john-science, line 2080: There's no "else" clause here. Hmmm... Usually, it's good practice to catch those. Raise a warning or an error.

I guess in this location the question is what if the component has multiplicity 1, so it just has the default `IndexLocation`?  We just don't  want to rotate that because we assume it is at the center of the block? Is that your idea?

armi/reactor/blocks.py
```diff
@@ -2344,36 +2318,40 @@ def autoCreateSpatialGrids(self):
...
-        )
-        spatialLocators = grids.MultiIndexLocation(grid=self.spatialGrid)
-        numLocations = 0
-        for ring in range(ringNumber):
-            numLocations = numLocations + hexagon.numPositionsInRing(ring + 1)
-        if numLocations != self.getNumPins():
+        if self.spatialGrid is not None:
+            grid = self.spatialGrid
+        else:
+            grid = grids.HexGrid.fromPitch(
```
john-science, line 2329: Do you want to tie this to the block by using the `armiObject` attribute? Just curious.

armi/reactor/blocks.py
```diff
@@ -2344,36 +2318,40 @@ def autoCreateSpatialGrids(self):
...
+                self.getPinPitch(cold=True), numRings=nRings, cornersUp=True
+            )
+        spatialLocators = grids.MultiIndexLocation(grid=grid)
+        numLocations = hexagon.totalPositionsUpToRing(nRings)
+        if numLocations != nPins:
             raise ValueError(
-                "Cannot create spatialGrid, number of locations in rings{} not equal to pin number{}".format(
-                    numLocations, self.getNumPins()
-                )
+                f"Cannot create spatialGrid, number of locations in rings {nRings} "
```
john-science, line 2336: I like this.

But if this happens on a different node than the primary node, all the primary node log will see is this process was killed, not why.  Can we do something like this?

```python
msg = (
    f"Cannot create spatialGrid, number of locations in rings {nRings} "
    f"not equal to {nPins=}"
)
runLog.error(msg)
ValueError(msg)
```

armi/reactor/tests/test_hexBlockRotate.py
```diff
@@ -0,0 +1,245 @@
...
+import math
+import unittest
+
+import numpy as np
+
+from armi.reactor.blocks import HexBlock
+from armi.reactor.components import Component
+from armi.reactor.grids import MultiIndexLocation, CoordinateLocation
+from armi.utils import iterables
+from armi.reactor.tests.test_blocks import loadTestBlock, NUM_PINS_IN_TEST_BLOCK
```
john-science, line 26:
```suggestion
from armi.reactor.blocks import HexBlock
from armi.reactor.components import Component
from armi.reactor.grids import MultiIndexLocation, CoordinateLocation
from armi.reactor.tests.test_blocks import loadTestBlock, NUM_PINS_IN_TEST_BLOCK
from armi.utils import iterables
```

john-science: Just alphabetizing imports.

armi/reactor/tests/test_hexBlockRotate.py
```diff
@@ -0,0 +1,245 @@
...
+            msg = f"{name=} :: {degrees=} :: {data=}"
+            np.testing.assert_array_equal(data, expected, err_msg=msg)
+
+    def test_pinRotationLocations(self):
+        """Test that pin locations are updated through rotation.
+
+        .. test:: HexBlock.getPinLocations is consistent with rotation.
+            :id: T_ARMI_ROTATE_HEX_PIN_LOCS
+            :tests: R_ARMI_ROTATE_HEX
+
```
john-science, line 124:
```suggestion
```

armi/reactor/tests/test_hexBlockRotate.py
```diff
@@ -0,0 +1,245 @@
...
+        for paramName in self.PIN_PARAMS:
+            actual = fresh.p[paramName]
+            np.testing.assert_equal(actual, self.PIN_DATA, err_msg=paramName)
+
+
+class EmptyBlockRotateTest(unittest.TestCase):
+    """Rotation tests on an empty hexagonal block.
+
+    Useful for enforcing rotation works on blocks without pins.
+
```
john-science, line 232:
```suggestion
```

doc/user/spatial_block_parameters.rst
```diff
@@ -0,0 +1,339 @@
...
+are assigned to locations on that grid according to their ``specifcier`` blueprint attribute. Below is an example
+of a "flats up" hexagonal core grid of fuel assemblies with 1/3 symmetric.
+
+.. code:: yaml
+
+    grids:
+      core:
+        geom: hex
+        symmetry: third periodic
+        lattice map: |
```
john-science, line 31: This is correct, but perhaps incomplete.

You can also define a lattice map as an explicit list of X/Y coordinates, and what lives there (if drawing pictures isn't your style):

https://github.com/terrapower/armi/blob/ffd6400a4902f7b5fb3188472d37e75e25f4a168/armi/tests/godiva/godiva-blueprints.yaml#L476-L484

I am fully aware this is not popular, but I'm just mentioning it.

armi/reactor/blocks.py
drewj-tp: Yeah I don't love this either. It looks if a component is placed on a grid, even if it has a single location, it **always** gets a `MultiIndexLocator`

https://github.com/terrapower/armi/blob/a3ef1a39325e2f33de966f8d11a7e7c4d380f3a3/armi/reactor/blueprints/blockBlueprint.py#L163-L182

https://github.com/terrapower/armi/blob/a3ef1a39325e2f33de966f8d11a7e7c4d380f3a3/armi/reactor/blueprints/gridBlueprint.py#L473-L477

The `CoordinateLocation` gets set during `HexBlock.autoCreateSpatialGrids`

https://github.com/terrapower/armi/blob/a3ef1a39325e2f33de966f8d11a7e7c4d380f3a3/armi/reactor/blocks.py#L2372-L2376

So we should have a `MultiIndexLocation` with one or more entries, or a `CoordinateLocation`. I'll make sure this raises the appropriate error.

armi/reactor/blocks.py
drewj-tp: Will update. I've seen that pattern a few places, but not enough to know it was our preferred way to do logging / errors

armi/reactor/blocks.py
drewj-tp: That's a good idea

armi/reactor/tests/test_hexBlockRotate.py
```diff
@@ -0,0 +1,245 @@
...
+    def test_orientationVector(self):
+        """Test the z-value in the orientation vector matches rotation.
+
+        .. test:: Rotate a hex block in 60 degree increments.
+            :id: T_ARMI_ROTATE_HEX
+            :tests: R_ARMI_ROTATE_HEX
+
+        .. test:: Update block orientation.
+            :id: T_ARMI_ROTATE_HEX_ORIENTATION
+            :tests: R_ARMI_ROTATE_HEX_PARAMS
```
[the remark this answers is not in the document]
drewj-tp: Updated. Let me know what you think. If you think it needs more, please leave an in-line suggestion

armi/reactor/blocks.py
drewj-tp: Updated. Let me know what you think. If you think it needs more, please leave an in-line suggestion

armi/reactor/blocks.py
drewj-tp: Some [testing ](https://github.com/terrapower/armi/actions/runs/11333020227/job/31516341360#step:6:1126)uncovered an `IndexLocation`
```
TypeError: <DerivedShape: coolant> on <fuel B-2307749255035-001 at ExCore XS: A BU GP: A> has an invalid spatial locator for rotation: <IndexLocation @ (0,0,0)>
```
Added logic to handle that case as well

armi/reactor/blocks.py
john-science: Well, long term, I would prefer to make it so `runLog.error()` did this by default.  After all, "errors" should kill a run.

And Mark and Nick agree with me.

But... it'll take time and a bunch of things will have to be changed/fixed downstream. So... not yet.

doc/release/0.4.rst
```diff
@@ -36,7 +37,8 @@ API Changes
 #. Removing ``Assembly.doubleResolution()``. (`PR#1951 <https://github.com/terrapower/armi/pull/1951>`_)
 #. Removing ``assemblyLists.py`` and the ``AssemblyList`` class. (`PR#1891 <https://github.com/terrapower/armi/pull/1891>`_)
 #. Removing ``Assembly.rotatePins`` and ``Block.rotatePins``. Prefer ``Assembly.rotate`` and ``Block.rotate``. (`PR#1846 <https://github.com/terrapower/armi/1846`_)
-#. Transposing ``pinMgFluxes`` parameters so that leading dimension is pin index (`PR#1937 <https://github.com/terrapower/armi/pull/1937>`)
+#. Transposing ``pinMgFluxes`` parameters so that leading dimension is pin index (`PR#1937 <https://github.com/terrapower/armi/pull/1937>`_)
```
john-science, line 40:
```suggestion
#. Transposing ``pinMgFluxes`` parameters so that leading dimension is pin index. (`PR#1937 <https://github.com/terrapower/armi/pull/1937>`_)
```

doc/user/spatial_block_parameters.rst
```diff
@@ -0,0 +1,339 @@
...
+are chosen for brevity but we could have also done ``fuel`` and ``guide`` or ``F`` and ``G``. Do what makes sense
+for you.
+
+.. note::
+
+    Like with assembly specifiers, keeping the lattice IDs to have the same number of characters will
+    help the grid render nicer in text editors. This is not a requirement, but it may make life easier
+    for you and your team.
+
+Our complete block definition would start like
```
john-science, line 93:
```suggestion
Our complete block definition would start like::
```

john-science: Unless I'm wrong, then just close this.

doc/user/spatial_block_parameters.rst
```diff
@@ -0,0 +1,339 @@
...
+-------------------
+
+Components that live on a spatial grid have a ``spatialLocator`` attribute to help indicate where that
+component exists in space. If we grab the fuel component from the UO2 block in the
+:ref:`ANL AFCI 177 example <walkthrough-inputs>` we can see where it exists in the block::
+
+    >>> import armi
+    >>> armi.configure()
+    >>> from armi.reactor.flags import Flags
+    >>> r = armi.init(fName="c5g7-settings.yaml").r
```
john-science, line 176: You say you want to use the example reactor `anl-afci-177`, but then you point at the `c5g7` reactor. These are two completely different sets of settings files and blueprints in ARMI's test area.

doc/user/spatial_block_parameters.rst
```diff
@@ -0,0 +1,339 @@
...
+           [ 0.4444    ,  0.76972338,  0.        ],
+           [-0.4444    ,  0.76972338,  0.        ],
+           [-0.8888    ,  0.        ,  0.        ],
+    ...
+           [ 5.7772    ,  3.84861689,  0.        ],
+           [ 5.3328    ,  4.61834027,  0.        ],
+           [ 4.8884    ,  5.38806365,  0.        ],
+           [ 4.444     ,  6.15778703,  0.        ]])
+
+We get a ``(271, 3)`` array because we have 271 of these fuel components in the block, and each row contains one
```
john-science, line 219: I think this would be cleaner and easier to read if you just did `coordsFromFuel.shape` or something. We don't think the users are going to be spot checking these values here. What we want is to show them the dimensionality of this example.

Just my two cents.

armi/reactor/grids/hexagonal.py
```diff
@@ -572,3 +575,45 @@ def generateSortedHexLocationList(self, nLocs: int):
...
+        -------
+        IndexLocation
+            Index in the grid after rotation
+
+        Notes
+        -----
+        Rotation uses a three-dimensional index in what can be known elsewhere
+        by the confusing name of "cubic" coordinate system for a hexagon. Cubic stems
+        from the notion of using three dimensions, ``(q, r, s)`` to describe a point
+        in the hexagonal grid. The conversion from ARMI's indexing follows::
```
john-science, line 599: So, I need to go through ARMI docs and docstrings and fix this globally, but just FYI:

We can not use "ARMI" as a noun, any more. The lawyers argue that this can lead to "genericification" and can cause us to lose our trademark.

So, we are compelled to say things like "the ARMI framework" or "the ARMI data model" or "the ARMI software" or "the ARMI database", etc.

Sorry!  Not my call.

armi/reactor/blocks.py
```diff
@@ -1559,26 +1560,49 @@ class during axial expansion.
...
+        """Produce all the index locations for pins in the block.
+
+        Returns
+        -------
+        list[grids.IndexLocation]
+            Integer locations where pins can be found in the block.
+
+        Notes
+        -----
+        Only components with ``Flags.CLAD`` are considered to define a pin's location.
```
john-science, line 1573: I know this isn't part of your design, but maybe this method could take an optional `flag` parameter that defaults to `Flags.CLAD`. That way, if in the future someone wants to use some other flag to identify what they want to define as a "pin".

This request is scope creep, but it seems like a one-line change and a single unit test to make this more extensible in ways that you have convinced me are important.

Thoughts?

drewj-tp: Are you okay if I make a ticket and/or attach this to a discussion? While `getPinLocations(Flags.FUEL)` tracks, more generally trying to find locations of children seems useful. But, calling `getPinLocations(Flags.POISON)` feels weird.

john-science: This can be a separate ticket, NBD.  It's a very minor idea, and won't hold up this PR either way.

doc/user/spatial_block_parameters.rst
```diff
@@ -0,0 +1,329 @@
...
+    >>> <IndexLocation @ (1,0,0)>
+    >>> fuel.spatialLocator[1].getLocalCoordinates()
+    array([0.4444    , 0.76972338, 0.        ]))
+    >>> fuelBlock.rotate(math.radians(60))
+    >>> fuel.spatialLocator[1]
+    <IndexLocation @ (0,1,0)>
+    >>> fuel.spatialLocator[1].getLocalCoordinates()
+    array([-0.4444    ,  0.76972338,  0.        ])
+
+Because the sub-block on this grid is a corners up grid, to tightly fit inside the flats up hex block,
```
john-science, line 293:
```suggestion
Because this sub-block grid is a corners up hex grid, to tightly fit inside the flats up hex block,
```

doc/user/spatial_block_parameters.rst
```diff
@@ -0,0 +1,329 @@
+************************
+Spatial block parameters
+************************
+
+Many parameters assigned on a ``Block`` are scalar quantities that are useful for visualization
+and simple queries (e.g., block with the maximum burnup in an assembly). Spatial parameters in
+the a block, such as power produced by each pin, is also of interest. Especially when communicating
```
john-science, line 7:
```suggestion
a block, such as power produced by each pin, is also of interest. Especially when communicating
```

doc/user/spatial_block_parameters.rst
```diff
@@ -0,0 +1,329 @@
...
+data to physics codes that support sub-block geometric modeling. This page will talk about
+how spatial information is assigned to components on a block, how spatial data can be assigned
+and accessed, and how those data may or may not be updated by the framework.
+
+Sub-block spatial grid
+======================
+
+There are two ways to create the block grid: explicitly via blueprints, or via an automated builder. The former
```
john-science, line 15:
```suggestion
There are two ways to create the block grid: explicitly via blueprints or via an automated builder. The former
```

john-science: Commas usually only separate items if the list is longer than 2 items long.

I won't die on this hill.

doc/user/spatial_block_parameters.rst
```diff
@@ -0,0 +1,329 @@
...
+is recommended, but the later can work in some specific circumstances.
+
+Blueprints
+----------
+
+In your blueprints file, you likely have a core grid that defines where assemblies reside in the reactor. Assemblies
+are assigned to locations on that grid according to their ``specifcier`` blueprint attribute. Below is an example
+of a "flats up" hexagonal core grid of fuel assemblies with 1/3 symmetric.
```
john-science, line 23:
```suggestion
of a "flats up" hexagonal core grid of fuel assemblies with 1/3 symmetry.
```

john-science (approved):

john-science (approved):

drewj-tp:
Had some downstream differences that I couldn't quite square until I removed

1. Changes to `HexBlock.autoCreateSpatialGrids`
   - Also being modified in #1947 
2. Implementation of `MultiIndexLocation.getLocalCoordinates`
   - Maybe the "bug" I thought I found was a product of a bad test configuration? For whatever reason, I'm not hitting a potentially related problem that caused me to look at this method

armi/reactor/tests/test_hexBlockRotate.py
```diff
@@ -0,0 +1,266 @@
...
+    PIN_PARAMS = [
+        "percentBuByPin",
+        "linPowByPin",
+    ]
+
+    PIN_DATA = np.arange(NUM_PINS_IN_TEST_BLOCK, dtype=float)
+
+    @classmethod
+    def setUpClass(cls):
+        cls.BASE_BLOCK = loadTestBlock()
```
john-science, line 59: So... if this is in `setUpClass(cls)` and not `setUp(self)`, then you think ALL tests should share the same, class-level block?

But... what if I'm running the tests in parallel, and each test rotates the block? Then... depending on the order the tests are run in, the block might get rotated 4 times during one test?

I understand your tests are passing on GitHub, I can see that. 

But I think this should be a `setUp(self)` declaration, because if each test rotates a block once, I think we want to conceptually keep that block in question unique to each test.

IMO

armi/reactor/tests/test_hexBlockRotate.py
```diff
@@ -0,0 +1,266 @@
...
+        """Test the z-value in the orientation vector matches rotation.
+
+        .. test:: Demonstrate that a HexBlock can be rotated in 60 degree increments.
+            :id: T_ARMI_ROTATE_HEX
+            :tests: R_ARMI_ROTATE_HEX
+
+        .. test:: After rotating a block, the orientation parameter reflects the current rotation.
+            :id: T_ARMI_ROTATE_HEX_ORIENTATION
+            :tests: R_ARMI_ROTATE_HEX_PARAMS
+            :acceptance_criteria: Through a unit test, compare the z-rotation entry against the rotation amount.
```
john-science, line 84:
```suggestion
            :acceptance_criteria: Compare the z-rotation entry against the rotation amount.
```

john-science: We can't have the `acceptance_criteria` for every single ARMI req start off with "in a unit test", it's redundant.

IMHO

john-science, line 84:
```suggestion
```

john-science (approved):

zachmprince (approved):

john-science (approved):

PR #1945, opened 2024-10-10
New plugin hook for before reactor construction

opotowsky:
## What is the change?

This adds a new plugin hook `beforeReactorConstruction`. Then some code (material namespace order setting) is pulled out of the  reactor factory and into the hook call for the reactor plugin.  

This cleans up the reactor construction and sets the stage for plugins to fancify the `beforeReactorConstruction` hook should they be interested in fancification. 

---

## Checklist

### DIFF
--- a/armi/reactor/__init__.py
+++ b/armi/reactor/__init__.py
@@ -66,12 +66,13 @@ class ReactorPlugin(plugins.ArmiPlugin):
     @staticmethod
     @plugins.HOOKIMPL
     def beforeReactorConstruction(cs) -> None:
+        """Just before reactor construction, update the material "registry" with user settings,
+        if it is set. Often it is set by the application.
+        """
         from armi.settings.fwSettings.globalSettings import (
             CONF_MATERIAL_NAMESPACE_ORDER,
         )
 
-        # just before reactor construction, update the material "registry" with user settings,
-        # if it is set. Often it is set by the application.
         if cs[CONF_MATERIAL_NAMESPACE_ORDER]:
             materials.setMaterialNamespaceOrder(cs[CONF_MATERIAL_NAMESPACE_ORDER])
 
--- a/armi/reactor/reactors.py
+++ b/armi/reactor/reactors.py
@@ -179,7 +179,6 @@ def factory(cs, bp, geom: Optional[SystemLayoutInput] = None) -> Reactor:
     from armi.reactor import blueprints
 
     runLog.header("=========== Constructing Reactor and Verifying Inputs ===========")
-
     getPluginManagerOrFail().hook.beforeReactorConstruction(cs=cs)
 
     r = Reactor(cs.caseTitle, bp)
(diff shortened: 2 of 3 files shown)


armi/plugins.py
```diff
@@ -267,6 +267,15 @@ def onProcessCoreLoading(core, cs, dbLoad) -> None:
         constructing a Core from Blueprints, or after loading it from a database.
         """
 
+    @staticmethod
+    @HOOKSPEC
+    def beforeReactorConstruction(cs) -> None:
+        """
+        Function to call before the reactor is constructed.
+
+        TODO fill out. Requirement for this?
```
opotowsky, line 276: Need to talk offline about this.

ntouran:
Here's a patch that will get at least some test coverage. It's not easy to confirm positively that the reactor doesn't exist because the plugin hook has no knowledge of that. I think just showing that it's run during init is probably sufficient. 

```patch
diff --git a/armi/tests/test_user_plugins.py b/armi/tests/test_user_plugins.py
index 0840afd5..0cc19f0d 100644
--- a/armi/tests/test_user_plugins.py
+++ b/armi/tests/test_user_plugins.py
@@ -12,6 +12,7 @@
 """Tests for the UserPlugin class."""
+
 import copy
 import os
 import unittest
@@ -106,6 +107,11 @@ class UserPluginOnProcessCoreLoading(plugins.UserPlugin):
         for b in blocks:
             b.p.height += 1.0

+    @staticmethod
+    @plugins.HOOKIMPL
+    def beforeReactorConstruction(cs) -> None:
+        cs._beforeReactorConstructionFlag = True
+

 class UpInterface(interfaces.Interface):
     """
@@ -248,6 +254,7 @@ class TestUserPlugins(unittest.TestCase):
             TEST_ROOT, inputFileName="smallestTestReactor/armiRunSmallest.yaml"
         )
         fuels = r.core.getBlocks(Flags.FUEL)
+        self.assertTrue(o.cs._beforeReactorConstructionFlag)

         # prove that our plugin affects the core in the desired way
         heights = [float(f.p.height) for f in fuels]
```

armi/reactor/__init__.py
```diff
@@ -62,6 +63,18 @@
...
 
+    @staticmethod
+    @plugins.HOOKIMPL
+    def beforeReactorConstruction(cs) -> None:
+        from armi.settings.fwSettings.globalSettings import (
+            CONF_MATERIAL_NAMESPACE_ORDER,
+        )
+
+        # just before reactor construction, update the material "registry" with user settings,
+        # if it is set. Often it is set by the application.
```
john-science, line 74:
```suggestion
        """Just before reactor construction, update the material "registry" with user settings,
        if it is set. Often it is set by the application.
        """
        from armi.settings.fwSettings.globalSettings import (
            CONF_MATERIAL_NAMESPACE_ORDER,
        )

```

john-science: My thought is this could be a useful docstring.  

#not-a-deal-breaker

armi/reactor/reactors.py
```diff
@@ -180,10 +179,9 @@ def factory(cs, bp, geom: Optional[SystemLayoutInput] = None) -> Reactor:
     from armi.reactor import blueprints
 
     runLog.header("=========== Constructing Reactor and Verifying Inputs ===========")
-    # just before reactor construction, update the material "registry" with user settings,
-    # if it is set. Often it is set by the application.
-    if cs[CONF_MATERIAL_NAMESPACE_ORDER]:
-        materials.setMaterialNamespaceOrder(cs[CONF_MATERIAL_NAMESPACE_ORDER])
+
```
john-science, line 182:
```suggestion
```

john-science (approved):

PR #1947, opened 2024-10-10
Fixing sub-block grids

john-science:
## What is the change?

It appears there was a bug where Components inside a Block (like fuel pins) would have the wrong location if they were on a Hex grid that was corners-side up. The pins would be incorrectly rotated to the "corners up" grid rotation, even they should be flats-up.

The change I made to fix the issue:

* Move adding a spatial grid to Components in a Block until the Assembly containing those Blocks is added to the Core or SFP
* calculate "Is this a corners up grid?" and pass that into the method that adds the spatial grid to the Blocks (`autoCreateSpatialGrid()`)

Because fixing bugs is good.

---

## Checklist


john-science:
I have proof that this WAS a bug, but this PR fixes it.  I added a bunch of unit tests in this PR, but they all pass here, however, if you copy one of the new ones to ARMI `main`, it fails:

```bash
__________________________________ TestHexBlockOrientation.test_validateFlatsUp __________________________________ 

  File "Python3.9.7\lib\unittest\case.py", line 59, in testPartExecutor
    yield
  File "Python3.9.7\lib\unittest\case.py", line 592, in run
    self._callTestMethod(testMethod)
  File "Python3.9.7\lib\unittest\case.py", line 550, in _callTestMethod
    method()
  File "armi\armi\reactor\tests\test_blocks.py", line 2281, in test_validateFlatsUp
    self.assertFalse(b.spatialGrid.cornersUp)
  File "Python3.9.7\lib\unittest\case.py", line 674, in assertFalse
    raise self.failureException(msg)
AssertionError: True is not false
```

mgjarrett:
I haven't reviewed this, but I skimmed it and I think this solution is great! I had stopped working on #1648 because I ran into the exact problem that this PR addresses and I didn't know how we wanted to resolve it. This looks good to me 👍

armi/reactor/assemblies.py
```diff
@@ -1250,6 +1249,15 @@ def isOnWhichSymmetryLine(self):
...
 
+    def orientBlocks(self, cornersUp):
+        """Add special grids to the blocks inside this Assembly, respecting their orientation."""
+        for b in self:
+            if b.spatialGrid is None:
+                try:
+                    b.autoCreateSpatialGrids(cornersUp)
+                except (ValueError, NotImplementedError) as e:
+                    runLog.warning(str(e), single=True)
+
```
drewj-tp, line 1260: Should this live on the `HexAssembly` because we have a `cornerUp` flag that is hex specific?

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -240,11 +239,6 @@ def construct(
         b.verifyBlockDims()
         b.spatialGrid = spatialGrid
 
-        if b.spatialGrid is None and cs[globalSettings.CONF_BLOCK_AUTO_GRID]:
```
drewj-tp, line 243: This is the single place this is used inside the framework. And likely anywhere based on shared experience and discussion with some other devs. My understanding is very few if any people set this setting to be false.

We should do one of these things

1. Include the `CONF_BLOCK_AUTO_GRID` setting check inside `HexBlock.autoCreateSpatialGrids`.
2. Include the `CON_BLOCK_AUTO_GRID` setting check inside `HexAssembly.orientBlocks` when prior to calling the auto grid creation
3. Document in release notes and in the `CONF_BLOCK_AUTO_GRID` setting definition that this is removed and should not be used anymore

I'm inclined for option 3 thought it is, technically, API breaking. Someone out there who relies on `CONF_BLOCK_AUTO_GRID = False` will have a problem. My gut is it's a small group, but I've been wrong before and will be wrong again

armi/reactor/blocks.py
```diff
@@ -2349,29 +2354,31 @@ def autoCreateSpatialGrids(self):
...
+        # set the spatial position of the sub-block components
+        spatialLocators = grids.MultiIndexLocation(grid=grid)
         for ring in range(ringNumber):
             for pos in range(grid.getPositionsInRing(ring + 1)):
                 i, j = grid.getIndicesFromRingAndPos(ring + 1, pos + 1)
                 spatialLocators.append(grid[i, j, 0])
+
+        # finally, fill the spatial grid, and put the sub-block components on it
         if self.spatialGrid is None:
             self.spatialGrid = grid
```
drewj-tp, line 2383: The caller for this method in `Assembly.orientBlocks` only calls this if `HexBlock.spatialGrid is None`. Do we need to have this check here?

armi/reactor/blocks.py
```diff
@@ -2349,29 +2354,31 @@ def autoCreateSpatialGrids(self):
...
 
-        ringNumber = hexagon.numRingsToHoldNumCells(self.getNumPins())
-        # For the below to work, there must not be multiple wire or multiple clad types.
-        # note that it's the pointed end of the cell hexes that are up (but the
-        # macro shape of the pins forms a hex with a flat top fitting in the assembly)
+        # build the grid, from pitch and orientation
         grid = grids.HexGrid.fromPitch(
-            self.getPinPitch(cold=True), numRings=0, cornersUp=True
+            self.getPinPitch(cold=True), numRings=0, cornersUp=cornersUp
         )
```
drewj-tp, line 2360: Since we are more likely to be calling this on a block w/o a grid, we should "anchor" this grid to the block, e.g.,
`HexGrid.fromPitch(..., armiObject=self)`

This `armiObject` is used to dig up the composite tree on the object attached to the grid

https://github.com/terrapower/armi/blob/d3cc8e5fcb1f83206e372baa657b3d51a7f6496c/armi/reactor/grids/structuredGrid.py#L468-L484

And that is some convoluted logic... I guess it's helpful trying to get global coordinates on e.g., a component in a block so you'd need to dig up to the global coordinate system component -> block -> assembly -> core and find where on the "parent" grid your grid is located?

armi/reactor/grids/tests/test_grids.py
```diff
@@ -261,6 +261,82 @@ def test_positions(self):
...
+            self.assertAlmostEqual(
+                (sum(val**2 for val in directionVector)) ** 0.5,
+                1.0,
+                msg=f"Direction vector {directionVector} should have "
+                "magnitude 1 for pitch 1.",
+            )
+        assert_allclose(grid.getCoordinates((1, 0, 0)), iDirection)
+        assert_allclose(grid.getCoordinates((0, 1, 0)), jDirection)
+
+    def test_getLocalCoordinates(self):
```
drewj-tp, line 287: A docstring would be helpful because we're not really checking the validity of `getLocalCoordinates` like I assumed by reading the test name. We're making sure that the `getLocalCoordinates` differ for corners and flats up hexes

armi/reactor/grids/tests/test_grids.py
```diff
@@ -261,6 +261,82 @@ def test_positions(self):
...
+        self.assertGreater(minY, -1)
+
+        # long in X
+        maxX = max(v[0] for v in vals)
+        minX = min(v[0] for v in vals)
+        self.assertAlmostEqual(maxX, 1)
+        self.assertAlmostEqual(minX, -1)
+
+    def test_getLocalCoordinatesFlatsUp(self):
+        # validate the first ring of a flats-up hex gric
```
drewj-tp, line 321:
```suggestion
        # validate the first ring of a flats-up hex grid
```

armi/reactor/grids/tests/test_grids.py
```diff
@@ -261,6 +261,82 @@ def test_positions(self):
...
 
+    def test_getCoordinatesCornersUp(self):
+        grid = grids.HexGrid.fromPitch(1.0, cornersUp=True)
+        self.assertAlmostEqual(grid.pitch, 1.0)
+        side = 1.0 / math.sqrt(3)
+        assert_allclose(grid.getCoordinates((0, 0, 0)), (0.0, 0.0, 0.0))
+        assert_allclose(grid.getCoordinates((1, 0, 0)), (0.5, 1.5 * side, 0.0))
+        assert_allclose(grid.getCoordinates((-1, 0, 0)), (-0.5, -1.5 * side, 0.0))
+        assert_allclose(grid.getCoordinates((0, 1, 0)), (-0.5, 1.5 * side, 0.0))
+        assert_allclose(grid.getCoordinates((1, -1, 0)), (1, 0.0, 0.0))
```
drewj-tp, line 272: Math checks out 👍

drewj-tp (changes requested):
Thanks for digging into this @john-science. I think this is the most feasible way for us to solve the problem. 

I'm going to provide some context as to why we need to loop over every block in every assembly during `Core.add` to solve this problem.

1. When we make a block from blueprints, we have not fully created the assembly, let alone the grid for that assembly.
2. It is possible to use the same assembly design in two different grids, so we can't set the orientation in the "cached" assembly stored on [the blueprint object](https://github.com/terrapower/armi/blob/d3cc8e5fcb1f83206e372baa657b3d51a7f6496c/armi/reactor/blueprints/__init__.py#L264-L269) during constructing assemblies
3. We don't really know the orientation of a block in an assembly until that assembly is fully realized and added to a thing (core, sfp, etc.)

armi/reactor/assemblies.py
john-science: All assemblies have a `.autoCreateSpatialGrids()` method though, right?

armi/reactor/blocks.py
john-science: This is a public method. People can call it multiple times, any time they want, downstream in their plugins.

So, this method would cause havoc if we let people call it multiple times in a row.

(I am considering moving this check (`if self.spatialGrid is None`) up to the top of the method in a different PR, by the way.  But we have like 6 unit tests designed specifically for this check to work this way that will break. So I didn't want to muddy the waters and make this PR less clear for you to read. So I'll do it as it's own stand-alone PR.)

armi/reactor/blueprints/blockBlueprint.py
john-science: I'm afraid (1) and (2) aren't possible without making changes to like 12 full methods to pass this information around.

My plan was to do (3) and remove this unused setting after this PR, as I was trying to reduce the phase space of this PR, to make it easier to review.

BUT, since you asked for it, sure I can remove the setting in this PR. I honestly don't understand what `False` ever meant in ARMI anyway. As far as I can tell, `False` has not been used in the last 8 years, at least inside TP.

john-science: done

armi/reactor/tests/test_blocks.py
```diff
@@ -2244,6 +2246,80 @@ def test_gridNotCreatedMultipleMultiplicities(self):
...
+                x, y, _ = loc.getLocalCoordinates()
+                if x > maxX:
+                    maxX = x
+                elif x < minX:
+                    minX = x
+
+                if y > maxY:
+                    maxY = y
+                elif y < minY:
+                    minY = y
```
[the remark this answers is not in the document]
john-science: done

drewj-tp:
Talked with @john-science and I wanted to clarify something that I missed in the initial review

> The pins would be incorrectly rotated to the "corners up" grid rotation, even for flats-up grids

A hexagonal block on a flats-up core grid **should** have a corners up pin grid if it intends to tightly pack pins into the block. If you look at the problem plot from #1421 



it looks like the pins are in a flats up grid because they form the shape of a flats up hexagon. But! The individual pins are on a corners up grid if you draw a tiny hexagon around each lattice site. Or the ability to walk from neighbor to neighbor in the x-direction does not require moving in the y-direction.

armi/reactor/blocks.py
```diff
@@ -2379,7 +2381,18 @@ def autoCreateSpatialGrids(self, cornersUp=True):
...
+            armiObject=self,
+            cornersUp=cornersUp,
+        )
+
+        # build the sub-grid, with opposite orientation to the block
+        subGrid = grids.HexGrid.fromPitch(
+            self.getPinPitch(cold=True),
+            numRings=0,
+            armiObject=self,
+            cornersUp=not cornersUp,
```
drewj-tp, line 2395: This is a deviation from how ARMI has previously reported the pin locations. On main, the following takes place
```python
>>> r = armi.init(fName="armi/tests/tutorials/anl-afci-177.yaml").r
>>> a = r.core.getFirstAssembly(Flags.FUEL)
>>> b = a.getFirstBlock(Flags.FUEL)
>>> fuel = b.getChildrenWithFlags(Flags.FUEL)[0]
>>> fuel.spatialLocator.grid is b.spatialGrid
True
```
The docs for `Composite.spatialGrid` and `Composite.spatialLocator` could be better, but they allude to this fact

https://github.com/terrapower/armi/blob/01b0adee861c3d9ffe9430b709ea1f1db39ca790/armi/reactor/composites.py#L322-L325

So `Composite.spatialGrid` is the sub-grid, or the grid on which the children live. `Core.spatialGrid` is the grid for all the assemblies, `Assembly.spatialGrid` for blocks, `Block.spatialGrid` for components.

I think the negation needs to go up a level higher:
1. `HexBlock.autoCreateSpatialGrid` makes the `HexBlock.spatialGrid` in the orientation prescribed by `cornersUp`
2. All the child components live on the grid defined by `HexBlock.spatialGrid`
3. When we add the `HexAssembly` to a `Core.spatialGrid` (or SFP) we need to invert the orientation with e.g.,
```python
def add(self, hexAssem):
    blockGridCornersUp = not self.spatialGrid.cornersUp
    for b in hexAssem:
        b.autoCreateSpatialGrid(..., cornersUp=blockGridCornersUp)
```

armi/reactor/grids/grid.py
```diff
@@ -158,6 +158,10 @@ def __setstate__(self, state: Dict):
     def isAxialOnly(self) -> bool:
         """Indicate to parts of ARMI if this Grid handles only axial cells."""
 
+    @property
+    def cornersUp(self) -> bool:
+        return False
+
```
drewj-tp, line 164: `CartesianGrid.cornersUp` doesn't make sense to me. This seems like putting hex specific logic onto the base class just to unify the API.

armi/reactor/assemblies.py
drewj-tp: Blocks yes, but the `cornersUp` switch is hex specific. What does this mean for a cartesian grid? I get that it's good OOP practice to have the same signatures for overridden methods, but this feels in the vein of "everything is hexagonal in ARMI"

armi/reactor/blocks.py
```diff
@@ -292,7 +292,7 @@ def getSmearDensity(self, cold=True):
 
         return smearDensity
 
-    def autoCreateSpatialGrids(self):
+    def autoCreateSpatialGrids(self, cornersUp=True):
```
drewj-tp, line 295: Same reservations about putting `cornersUp` switch in the generic `Block.autoCreateSpatialGrids`

armi/reactor/grids/tests/test_grids.py
```diff
@@ -261,6 +262,86 @@ def test_positions(self):
...
+        vals = []
+        for pos in range(grid.getPositionsInRing(2)):
+            i, j = grid.getIndicesFromRingAndPos(2, pos + 1)
+            vals.append(grid[i, j, 0].getLocalCoordinates())
+
+        # short in Y
+        maxY = max(v[1] for v in vals)
+        minY = min(v[1] for v in vals)
+        self.assertLess(maxY, 1)
+        self.assertGreater(minY, -1)
```
drewj-tp, line 316: Recommend a stronger check for asserting max and min y locations are +/- `pitch * cos(60)`

armi/reactor/grids/tests/test_grids.py
```diff
@@ -261,6 +262,86 @@ def test_positions(self):
...
+        maxY = max(v[1] for v in vals)
+        minY = min(v[1] for v in vals)
+        self.assertAlmostEqual(maxY, 1)
+        self.assertAlmostEqual(minY, -1)
+
+        # short in X
+        maxX = max(v[0] for v in vals)
+        minX = min(v[0] for v in vals)
+        self.assertLess(maxX, 1)
+        self.assertGreater(minX, -1)
```
drewj-tp, line 343: Same suggestion to check max and min x positions are +/- `pitch * cos(60)`

armi/reactor/spentFuelPool.py
```diff
@@ -70,6 +70,10 @@ def add(self, assem, loc=None):
         else:
             loc = self._getNextLocation()
 
+        # orient the blocks to match this grid
+        cornerUp = self.spatialGrid.cornersUp
+        assem.orientBlocks(cornerUp)
```
drewj-tp, line 75:
```suggestion
        assem.orientBlocks(not cornerUp)
```

armi/reactor/cores.py
```diff
@@ -563,11 +563,15 @@ def add(self, a, spatialLocator=None):
...
             )
             raise RuntimeError("Core already contains an assembly with the same name.")
+
         self.assembliesByName[aName] = a
         for b in a:
             self.blocksByName[b.getName()] = b
 
         if self.geomType == geometry.GeomType.HEX:
+            cornerUp = self.spatialGrid.cornersUp
+            a.orientBlocks(cornersUp=cornerUp)
```
drewj-tp, line 573: `Block.spatialGrid` should be the opposite orientation of the grid on which its parent assembly lies
```suggestion
            a.orientBlocks(cornersUp=not cornerUp)
```

armi/reactor/tests/test_blocks.py
```diff
@@ -2278,6 +2283,84 @@ def test_gridNotCreatedMultipleMultiplicities(self):
...
+        """Validate the spatial grid for a corners up HexBlock and its children."""
+        # load a corners up reactor
+        _o, r = loadTestReactor(
+            os.path.join(TEST_ROOT, "smallestTestReactor"),
+            inputFileName="armiRunSmallest.yaml",
+        )
+
+        # grab a pinned fuel block, and verify it is corners up
+        b = r.core.getFirstBlock(Flags.FUEL)
+        self.assertTrue(b.spatialGrid.cornersUp)
```
drewj-tp, line 2330: This test reactor has a corners up core grid so `Block.spatialGrid` should be flats up
```suggestion
        self.assertFalse(b.spatialGrid.cornersUp)
```

armi/reactor/tests/test_blocks.py
```diff
@@ -2278,6 +2283,84 @@ def test_gridNotCreatedMultipleMultiplicities(self):
...
+
+        # for a flats up sub-grid, the hex centroids should stretch more in Y than X
```
drewj-tp, line 2332:
```suggestion
        # for a flats up block-grid, the hex centroids should stretch more in Y than X
```

armi/reactor/tests/test_blocks.py
```diff
@@ -2278,6 +2283,84 @@ def test_gridNotCreatedMultipleMultiplicities(self):
...
+        open(testFile, "w").write(txt)
+
+        # load a flats up reactor
+        _o, r = loadTestReactor(
+            self.td.destination, inputFileName="armiRunSmallest.yaml"
+        )
+
+        # grab a pinned fuel block, and verify it is flats up
+        b = r.core.getFirstBlock(Flags.FUEL)
+        self.assertFalse(b.spatialGrid.cornersUp)
```
drewj-tp, line 2357: `Block.spatialGrid` in a flats up core should be corners up
```suggestion
        self.assertTrue(b.spatialGrid.cornersUp)
```

armi/reactor/tests/test_blocks.py
```diff
@@ -2278,6 +2283,84 @@ def test_gridNotCreatedMultipleMultiplicities(self):
...
+
+        # for a corners up sub-grid, the hex centroids should stretch more in X than Y
```
drewj-tp, line 2359:
```suggestion
        # for a corners up block-grid, the hex centroids should stretch more in X than Y
```

drewj-tp (changes requested):
My main concern is that `Block.spatialGrid` should be the grid on which its children live. I added some suggestions to help communicate this and I hope I caught enough locations so accepting the changes will still let unit tests pass. 

But I believe our desired outcome is
```python
b.spatialGrid.cornersUp !=  b.parent.parent.spatialGrid.cornersUp
```

john-science:
@drewj-tp I need to disambiguate. You say:

> My main concern is that `Block.spatialGrid` should be the grid on which its children live.

But below that you have 4 concerns I see:

1. You want `b.spatialGrid.cornersUp != b.parent.parent.spatialGrid.cornersUp`
2. You want `b.spatialGrid.cornersUp == list(b.getChildren())[0].spatialGrid.cornersUp`
3. You don't want the base `Grid` class to have a `.cornersUp()` method.
4. You don't want to pass `cornersUp` into `autoCreateSpatialGrid`.

Do I have that right?

drewj-tp:
> @drewj-tp I need to disambiguate. You say:
> 
> > My main concern is that `Block.spatialGrid` should be the grid on which its children live.
> 
> But below that you have 4 concerns I see:

Correct. I should be clearer on concern level and preference.

> 1. You want `b.spatialGrid.cornersUp != b.parent.parent.spatialGrid.cornersUp`

If we are doing `HexBlock.autoCreateSpatialGrid` yes, this is a must have.

Under the assumption the user wants to pack some number of pin-like objects into a hexagon, the grid of the hexagonal lattice should be 30-degrees rotated from the orientation of the block.

image

A user should still be allowed to explicitly create a flats up pin grid in a flats up core grid via blueprints.

> 2. You want `b.spatialGrid.cornersUp == list(b.getChildren())[0].spatialGrid.cornersUp`

Almost. `Composite.spatialGrid` indicates this attribute is the grid for children of the composite. `Core.spatialGrid` is the grid on which assemblies live. The `Composite.spatialLocator` tells us where this object lives in it's parents grid. So
```python
all(child.spatialLocator.grid is parent.spatialGrid for child in parent) == True
```
That is how I believe ARMI does it already and it makes sense to me. This is also a must have in that I think deviations are a backwards incompatible change that I don't believe is warranted.

> 3. You don't want the base `Grid` class to have a `.cornersUp()` method.

Want yes. Must have, no. We already have a lot of hex specific stuff on base classes. Why stop now.

> 4. You don't want to pass `cornersUp` into `autoCreateSpatialGrid`.

Want, yes. Must have, no. We have a lot of hex specific stuff in methods that should be generic. Why stop now.

armi/reactor/tests/test_blocks.py
```diff
@@ -2270,12 +2275,96 @@ def test_gridNotCreatedMultipleMultiplicities(self):
...
+
+                if y > maxY:
+                    maxY = y
+                elif y < minY:
+                    minY = y
+
+        return minX, maxX, minY, maxY
+
+    def test_validateReactorCornersUp(self):
+        """Validate the spatial grid for a corners up HexBlock and its children."""
```
drewj-tp, line 2321: To indicate a point of confusion here: 

1. The reactor is corners up as evidenced by the test name and `self.assertTrue(r.core.spatialGrid.cornersUp)`
2. The comments and docstrings allege a hex block in this reactor will be "corners up"
3. But, we `self.assertFalse(b.spatialGrid.cornersUp)` (which is the correct approach, in my understanding)
4. Later assertions on pin locations reference "flats up block grid"

So there is some ambiguity in what a "flats up" or "corners up" hex block means. Is it the orientation of the bounding hexagonal, derived from the core grid? Or the orientation of the grid attached to the block? I would recommend the former.

armi/reactor/tests/test_blocks.py
```diff
@@ -2270,12 +2275,96 @@ def test_gridNotCreatedMultipleMultiplicities(self):
...
+        # grab a pinned fuel block, and verify it is corners up
+        b = r.core.getFirstBlock(Flags.FUEL)
+        self.assertTrue(r.core.spatialGrid.cornersUp)
+        self.assertFalse(b.spatialGrid.cornersUp)
+        self.assertNotEqual(r.core.spatialGrid.cornersUp, b.spatialGrid.cornersUp)
+
+        # for a flats up block-grid, the hex centroids should stretch more in X than Y
+        minX, maxX, minY, maxY = self.getLocalCoordinatesBlockBounds(b)
+        ratio = (maxY - minY) / (maxX - minX)
+        self.assertAlmostEqual(ratio, math.sqrt(3) / 2, delta=0.0001)
```
drewj-tp, line 2337: I think shows we have a pin grid that is of the opposite orientation we want. The pins on a block in a corners up reactor should be on a flats up lattice. However, this test shows that while `b.spatialGrid` indicates a flats up grid, the actual pins are located on a corners up lattice.

1. If the block resides in a corners up grid, it's pin should be in a flats up orientation.
2. A hexagonal lattice in the flats up orientation stretches more in Y than in X.
   - The maximum Y distance between adjacent lattice sites is the pitch. Sites stack in the y-direction.
   - The X distance between adjacent lattice sites is `pitch * cos(30) = p * sqrt(3) / 2 < pitch`
   - For a single ring, `maxY - minY = 2 * pitch`
   - For a single ring, `maxX - minX = 2 * pitch * sqrt(3) / 2 = pitch * sqrt(3)`
3. The ratio should be `2 * pitch / (pitch * sqrt(3)) = 2 / sqrt(3)`
4. A corners up lattice would be and stretch longer in X than in Y, and have a ratio of `sqrt(3) / 2` like is tested here.

drewj-tp (changes requested):
The pins on the block are in the incorrect orientation still, due to the sub-block flipped grid.

hex_pins

<details>

<summary> Quick plotter </summary>

```python
"""Plot pins in an assembly in a reactor"""

import math
import pathlib

from matplotlib import pyplot, patches

import armi
from armi.reactor.components import Circle, Hexagon
from armi.reactor.tests.test_reactors import TEST_ROOT, loadTestReactor


def _plotPins(c: Circle, ax: pyplot.Axes):
    loc = c.spatialLocator
    radius = 0.5 * c.getDimension("od")
    for l in loc:
        xy = l.getLocalCoordinates()[:2]
        patch = patches.Circle(xy, radius=radius)
        ax.add_patch(patch)


def _plotDuct(c: Hexagon, ax: pyplot.Axes, cornersUp: bool):
    patch = patches.RegularPolygon(
        xy=(0, 0),
        numVertices=6,
        radius=c.getDimension("op") / 2,
        facecolor="w",
        edgecolor="k",
        orientation=0 if cornersUp else math.radians(30),
    )
    ax.add_patch(patch)


def plotGeometry(b: HexBlock, ax: pyplot.Axes, cornersUpBlock: bool):
    d = b.getChildrenWithFlags(Flags.DUCT)[0]
    _plotDuct(d, ax, cornersUpBlock)
    f = b.getChildrenWithFlags(Flags.FUEL)[0]
    _plotPins(f, ax)
    window = round(d.getDimension("op") * 0.525)
    ax.set(
        xlim=(-window, window),
        ylim=(-window, window),
    )


if __name__ == "__main__":
    armi.configure()
    _, r = loadTestReactor(
        pathlib.Path(TEST_ROOT) / "smallestTestReactor",
        inputFileName="armiRunSmallest.yaml",
    )

    b = r.core.getFirstBlock(Flags.FUEL)


    plotGeometry(
        b, pyplot.gca(), cornersUpBlock=r.core.spatialGrid.cornersUp,
    )
    pyplot.savefig("hex_pins.png")
    pyplot.show()
```

</details>

armi/reactor/blocks.py
```diff
@@ -2372,29 +2379,42 @@ def autoCreateSpatialGrids(self):
...
+            armiObject=self,
+            cornersUp=cornersUp,
+        )
+
+        # build the sub-grid, with opposite orientation to the block
+        subGrid = grids.HexGrid.fromPitch(
+            self.getPinPitch(cold=True),
+            numRings=0,
+            armiObject=self,
+            cornersUp=not cornersUp,
```
john-science, line 2395: @drewj-tp Whoops, I forgot to ditch this here. This is the cause of the problems you saw.  

Sorry about that.

drewj-tp (approved):
Thank you @john-science! Nothing more from me 🖖 🚀

drewj-tp (approved):

armi/reactor/blocks.py
```diff
@@ -292,22 +292,28 @@ def getSmearDensity(self, cold=True):
...
 
         Raises
         ------
         ValueError
             If the multiplicities of the block are not only 1 or N or if generated ringNumber leads
             to more positions than necessary.
         """
-        raise NotImplementedError()
+        if self.spatialGrid is None:
+            self.spatialGrid = systemSpatialGrid
```
drewj-tp, line 316: I think we shouldn't set the system spatial grid to the block's spatial grid on the base `Block.autoCreateSpatialGrid` implementation.

armi/reactor/blocks.py
```diff
@@ -293,22 +293,28 @@ def getSmearDensity(self, cold=True):
...
```
keckler, line 317: @john-science Am I mistaken, or is the docstring for this method pretty wrong? A couple points seem incorrect to me:
1. The method does not raise any errors, so the `Raises` section of the docstring seems wrong
2. This method does not appear to have the "ability to create a spatialGrid" as the docstring states

Maybe this docstring was intended for the child class?

PR #1970, opened 2024-10-24
Add new memory runLog info to memoryProfiler.py 

albeanth:
## What is the change?
Adds some potentially useful `runLog` statements to the memory profiler interface on how much memory is is use , how much is available, and the total available for a simulation. The first two are dynamic (i.e., change with (cycle, node)) while the third is static. 

Some memory intensive downstream analysis found this useful, so we thought we'd push it up to ARMI. 

---

## Checklist


armi/bookkeeping/memoryProfiler.py
```diff
@@ -78,6 +114,11 @@ def __init__(self, r, cs):
 
     def interactBOL(self):
         interfaces.Interface.interactBOL(self)
+
+        # prints the bottom line memory state...
+        # the code after this is somewhat confusing to interpret
+        printCurrentMemoryState(self.cs["mpiTasksPerNode"])
```
albeanth, line 120: the new equivalent of this setting isn't available in ARMI anymore.... shoot.

albeanth: after chatting with @opotowsky we put a bandaid `try/except` on this

armi/bookkeeping/memoryProfiler.py
```diff
@@ -36,16 +36,19 @@
 https://pythonhosted.org/psutil/
 https://docs.python.org/3/library/gc.html#gc.garbage
 """
+from os import cpu_count
 from typing import Optional
 import gc
 import sys
+from math import floor
```
john-science, line 43:
```suggestion
from math import floor
from os import cpu_count
from typing import Optional
import gc
import sys
```

john-science: just alphabetizing the imports

armi/bookkeeping/memoryProfiler.py
```diff
@@ -101,11 +129,30 @@ def interactEveryNode(self, cycle, node):
             mpiAction.broadcast().invoke(self.o, self.r, self.cs)
 
     def interactEOL(self):
-        r"""End of life hook. Good place to wrap up or print out summary outputs."""
+        """End of life hook. Good place to wrap up or print out summary outputs."""
         if self.cs["debugMem"]:
```
john-science, line 133: I'm just wondering... should this same settings check be done at `interactBOL` and `interactEveryNode`?

I could be totally off-base here, I just can't tell why this is only done at the one time node.

john-science (approved):
Approved! I have a couple of minor thoughts for you, but no hills I'm prepared to die on.

armi/bookkeeping/memoryProfiler.py
albeanth: I'm not following. Can you elaborate?

john-science: Well, this memory profiler does nothing in `interactEOL` unless the setting `debugMem` is set to `True`.

So.... I would EXPECT that setting would control all the time points, not just the one. 

If that's not the case cool, I'm just wondering what the difference is.

albeanth: That setting does get used in `interactBOL` and `interactEveryNode`. These two just also do other stuff too. 

Granted I don't know the original intent. The git blame for the vast majority of this module go well beyond my time with TerraPower.

opotowsky (approved):
I'm a fan! 

Note that we realize this code probably shouldn't be in ARMI but to push towards a deadline we have [a separate cleanup ticket](https://github.com/terrapower/armi/issues/1975) to bring parts of the memory profiler into internal tooling, where it belongs.

PR #1984, opened 2024-10-28
Update `copyOrWarn` and `getFileSHA1Hash` to account for directories

opotowsky:
## What is the change?

`copyOrWarn` can now handle directories. This also includes an edit to `getFileSHA1Hash` so that it can handle directories too.

Sometimes a directory needs to be copied too. Let's give the directories the attention they deserve!

This supports a case setting downstream that is a directory, not a file path.

---

## Checklist

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -880,9 +880,10 @@ def copyInterfaceInputs(
     This function should now be able to handle the updating of:
 
       - a single file (relative or absolute)
-      - a list of files (relative or absolute), and
+      - a list of files (relative or absolute)
       - a file entry that has a wildcard processing into multiple files.
         Glob is used to offer support for wildcards.
+      - a directory and its contents
 
     If the file paths are absolute, do nothing. The case will be able to find the file.
 
@@ -950,7 +951,7 @@ def copyInterfaceInputs(
                 path = pathlib.Path(f)
                 if not WILDCARD and not RELATIVE:
                     try:
-                        if path.is_absolute() and path.exists() and path.is_file():
+                        if path.is_absolute() and path.exists():
                             # Path is absolute, no settings modification or filecopy needed
                             newFiles.append(path)
                             continue
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -855,7 +855,7 @@ def _copyInputsHelper(
     -------
     destFilePath (or origFile) : str
     """
-    sourceName = os.path.basename(sourcePath)
+    sourceName = pathlib.Path(sourcePath).name
     destFilePath = os.path.join(destPath, sourceName)
     try:
         pathTools.copyOrWarn(fileDescription, sourcePath, destFilePath)
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -24,7 +24,6 @@ import re
 import shutil
 import sys
 import tempfile
-import threading
 import time
 
 from armi import runLog
@@ -65,19 +64,6 @@ def getFileSHA1Hash(filePath, digits=40):
                 sha1.update(data)
 
     return sha1.hexdigest()[:digits]
-
-
-def copyWithoutBlocking(src, dest):
-    """
-    Copy a file in a separate thread to avoid blocking while IO completes.
-
-    Useful for copying large files while ARMI moves along.
-    """
-    files = "{} to {}".format(src, dest)
-    runLog.extra("Copying (without blocking) {}".format(files))
-    t = threading.Thread(target=shutil.copy, args=(src, dest))
-    t.start()
-    return t
 
 
 def getPowerFractions(cs):
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -38,22 +38,31 @@ _HASH_BUFFER_SIZE = 1024 * 1024
 
 def getFileSHA1Hash(filePath, digits=40):
     """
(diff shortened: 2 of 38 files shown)


john-science:
Hey Arrielle! Can you re-run the linters?

Thanks!

opotowsky:
(working on some tests now + fixing some code, so this isn't quite ready yet)

armi/utils/__init__.py
```diff
@@ -38,37 +37,33 @@
...
 
+    for file in filesToHash:
+        with open(file, "rb") as f:
+            while True:
+                data = f.read(_HASH_BUFFER_SIZE)
+                if not data:
+                    break
+                sha1.update(data)
 
-def copyWithoutBlocking(src, dest):
```
opotowsky, line 61: this is used NOWHERE as far as I can tell. So I'm deleting it

armi/utils/__init__.py
```diff
@@ -38,37 +37,33 @@
 
 def getFileSHA1Hash(filePath, digits=40):
```
opotowsky, line 38: @john-science this was deceptively easy to do but I want your sanity check on this. Here's my understanding. The single file approach behaves the same as before. The dir approach does this:

1. loop through each file in dir
2. update sha1

The unit test does this with 2 files. To demonstrate how this was working I printed things out. Here's a simplification of that:
```
File1 sha: aabbcc
File2 sha: xxyyzz
Dir sha with update: axbycz
```

Is this acceptable behavior? No one but a downstream plugin will be using this dir option for now.

armi/utils/pathTools.py
```diff
@@ -54,29 +54,34 @@ def armiAbsPath(*pathParts):
...
+        else:
+            shutil.copy(sourcePath, destinationPath)
         runLog.debug(
-            "Copied {}: {} -> {}".format(fileDescription, sourcePath, destinationPath)
+            "Copied {}: {} -> {}".format(
+                filepathDescription, sourcePath, destinationPath
+            )
         )
     except shutil.SameFileError:
         pass
```
opotowsky, line 80: this exception is untestable. it WORKS properly but if I try to `assertRaises(shutil.SameFileError)` then I get this failure:

image

armi/utils/__init__.py
john-science: Agreed,

I can find no usage of this in ARMI or the broader ARMI ecosystem.

Good find!

armi/cases/case.py
```diff
@@ -855,7 +855,7 @@ def _copyInputsHelper(
     -------
     destFilePath (or origFile) : str
     """
-    sourceName = os.path.basename(sourcePath)
+    sourceName = pathlib.Path(sourcePath).name
```
opotowsky, line 858: this does the same thing, but `pathlib` is robust to a trailing `/` whereas `os.path` is not.

armi/utils/__init__.py
john-science: The approach makes sense to me.  Let me keep looking.

armi/utils/pathTools.py
john-science: Yeah, it would be hard to test the `pass` line. They usually are.

In our code coverage calculation, ARMI actually ignores `pass` lines, and does not count them.

Thanks for trying!  But you're good.

doc/release/0.4.rst
```diff
@@ -25,6 +25,7 @@ New Features
 #. Provide ``Block.getInputHeight`` for determining the height of a block from blueprints. (`PR#1927 <https://github.com/terrapower/armi/pull/1927>`_)
 #. Improve performance by changing the lattice physics interface so that cross sections are not updated on ``everyNode`` calls during coupled calculations (`PR#1963 <https://github.com/terrapower/armi/pull/1963>`_)
 #. Improve efficiency of reaction rate calculations. (`PR#1887 <https://github.com/terrapower/armi/pull/1887>`_)
+#. Update ``copyOrWarn`` and ``getFileSHA1Hash`` to account for directories. (`PR#1984 <https://github.com/terrapower/armi/pull/1984>`_)
```
john-science, line 28:
```suggestion
#. Updating ``copyOrWarn`` and ``getFileSHA1Hash`` to account for directories. (`PR#1984 <https://github.com/terrapower/armi/pull/1984>`_)
```

armi/utils/tests/test_utils.py
```diff
@@ -24,26 +24,47 @@
...
 
 
 class TestGeneralUtils(unittest.TestCase):
+    def test_getFileSHA1Hash(self):
+        with directoryChangers.TemporaryDirectoryChanger():
+            path = "test.txt"
+            with open(path, "w") as f1:
+                f1.write("test")
+            sha = getFileSHA1Hash(path)
+            self.assertIn("a94a8", sha)
```
john-science, line 54: I'm trying to decide if this test is fully stable.

I mean, it passed GitHub CI once, so maybe. Probably?

Different operating systems, different implementations of SHA1, different timestamps.  

I guess this works. Huh.

john-science: I guess, as long as the contents of the file are always utf-8, the hash will always be the same.

Okay, sure.

opotowsky: Yea I was actually trying to figure out how to test this and TOTALLY thought it would be unstable. So I was nervous about this test.

But it passed in 2 different places and then also the CI here so I guess it must be!

PR #1994, opened 2024-10-31
Removing broken plot buVsTime

### LINKED ISSUE #1902: BU vs time plot prints useless/incorrect results
One of the plots generated by default in ARMI is burnup vs time. The same plot also includes DPA vs time. In all of my runs, they contain all zeros:
image

The plotting is done by this function:
https://github.com/terrapower/armi/blob/59c392c1a6aa98e19c9955c88ca1a6106d1bcc35/armi/utils/reportPlotting.py#L216-L265

That function is quite fickle and incorporates a number of assumptions that are not often met by any old reactor that somebody might build. In those cases, you end up with a useless plot like the one that I showed above. Besides being useless, the plot is just wrong/misleading.

I would advocate moving highly-specific plots like this into the application that actually might use it (if any).

john-science:
## What is the change?

Removing broken plot `buVsTime`.

This plot has been broken for a few years, so I know it has been unused for that time.

I spoke with analysts that you might expect to want it, and they do not.

close #1902

---

## Checklist

### DIFF
--- a/armi/utils/reportPlotting.py
+++ b/armi/utils/reportPlotting.py
@@ -24,7 +24,6 @@ for instance, plot some sequence of objects in a loop at every time node. If you
 to see your memory usage grow inexplicably, you should question any plots that you are
 generating.
 """
-
 import itertools
 import math
 import os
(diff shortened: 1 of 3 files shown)


keckler (approved):
Easy, thanks!

PR #2011, opened 2024-11-19
Building docs during PRs

john-science:
## What is the change?

I am adding a new workflow that will ensure that docs are build and archived during PRs.

Now, if someone makes a change to the docs, they can download a zip file of the entire HTML doc directory, and look at the docs changes themselves.

A couple times in the past, @drewj-tp and I found this would be helpful.

---

## Checklist


drewj-tp (approved):
Looks good. Downloaded the zip file from the workflow and it looked similar enough to the website 👍 

The workflow is very similar to our main workflow. I wonder how we could combine them. Then if need to change our doc building pipeline, we only have one place to change. They only differ in that last "step": pushed to main result in a deployment to the website, while PRs produce the artifact.

Non-blocking but I'd like to hear your thoughts on how feasible that'd be. Could be a follow up action / ticket as well

PR #2013, opened 2024-11-20
Better diagnostics on ISOTXS compare 

onufer:
## What is the change?

This makes the isotxs diffs useable. we had this in the past but at some point it got lost. what you get now is unclear (right side). What you get after this PR is headers that tell you what is not equal
image

because at some point the useful messages were deleted and the compare is not currently useful.

---

## Checklist


john-science (approved):

armi/nuclearDataIO/cccc/isotxs.py
```diff
@@ -123,7 +124,10 @@ def compare(lib1, lib2, tolerance=0.0, verbose=False):
                 runLog.warning(warning.format(nuc2, 2, 1))
             equal = False
             continue
-        equal &= compareNuclideXS(nuc1, nuc2, tolerance, verbose)
+        nucEqual = compareNuclideXS(nuc1, nuc2, tolerance, verbose)
```
jakehader, line 127: Could consider expanding in this compareNuclideXS function to write out all differences or plot them for further assessments.

aaronjamesreynolds (approved):
This looks good to me. This slight changes greatly improves the usefulness of this comparison tool.

onufer:
i pushed up a change that will not print anything for nuclides that are equal i think this is good since it will print less lines
image

armi/nuclearDataIO/cccc/isotxs.py
onufer: verbose prints the matricies. there are often hundreds of differences and hundreds of plots im not sure are that benifical.

aaronjamesreynolds (approved):
The new additions make it so that ARMI only identifies XSs that differ between two ISOTXS libraries. This makes the diff easier to read as it isn't crowded with information about XSs that agree.

A confirmation message is still printed out if all XSs in each library are identical, which is a good explicit signal for the user to have.

john-science:
@onufer  Are you hoping this PR will make it into the (imminent) ARMI 0.5.0 release?

opotowsky:
@john-science I think this is safe to merge. I do believe it's meant to get into the 0.5 release

PR #2017, opened 2024-11-21
account for symmetry in volume integrated parameters in assemblies on symmetry lines

bsculac:
## What is the change?

The block parameters `initialB10ComponentVolume`, `massHmBOL`, and `molesHmBOL` did not have a location tag that would cause them to be multiplied by the symmetry factor when expanding to full core, so the VOLUME_INTEGRATED tag was added. This should cause these values to be correctly accounted for when assemblies are placed on symmetry boundaries. A survey of other parameters was made, and the VOLUME_INTEGRATED tag was added as needed to cause all block parameters that should change with the volume of the block to have the tag.

The assembly method `moveTo` is modified to have a check for the symmetry factor of the source and destination locations. The ratio of the symmetry factors is used to adjust the volume integrated parameters in the blocks within the assembly to account for the blocks being split between different sections of the core. 

The `moveTo` method is used for placing the assembly in the core and for movement within the core, so this change should account for all the modifications needed to correctly represent block parameters that change based on the volume of the block. Note that this does not change values that are calculated on the fly, such as the volume with the `getVolume` method.

Changes that are made based on discovered knock on effects from this change:
1. The `getParamMax` assembly method did not account for the symmetry factor when making comparisons between assemblies. It would therefore return incorrect values for parameters which were stored in a way that accounted for the symmetry factor. See https://github.com/terrapower/armi/pull/2017/files#diff-447b13bcb80d46b6c7901b9dae9ec8bcdf6f4f5f176a61b7beae558fd368dba2R216-R224

Incorrect values were calculated for `initialB10ComponentVolume` and downstream values dependent on it for assemblies located in the center of 1/3rd symmetric cores.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/tests/test_comparedb3.py
+++ b/armi/bookkeeping/db/tests/test_comparedb3.py
@@ -181,7 +181,7 @@ class TestCompareDB3(unittest.TestCase):
             dbs[1]._fullPath,
             timestepCompare=[(0, 0), (0, 1)],
         )
-        self.assertEqual(len(diffs.diffs), 576)
+        self.assertEqual(len(diffs.diffs), 504)
         # Cycle length is only diff (x3)
         self.assertEqual(diffs.nDiffs(), 3)
 
--- a/armi/bookkeeping/db/tests/test_comparedb3.py
+++ b/armi/bookkeeping/db/tests/test_comparedb3.py
@@ -181,7 +181,7 @@ class TestCompareDB3(unittest.TestCase):
             dbs[1]._fullPath,
             timestepCompare=[(0, 0), (0, 1)],
         )
-        self.assertEqual(len(diffs.diffs), 552)
+        self.assertEqual(len(diffs.diffs), 576)
         # Cycle length is only diff (x3)
         self.assertEqual(diffs.nDiffs(), 3)
 
--- a/armi/bookkeeping/db/tests/test_comparedb3.py
+++ b/armi/bookkeeping/db/tests/test_comparedb3.py
@@ -181,7 +181,7 @@ class TestCompareDB3(unittest.TestCase):
             dbs[1]._fullPath,
             timestepCompare=[(0, 0), (0, 1)],
         )
-        self.assertEqual(len(diffs.diffs), 480)
+        self.assertEqual(len(diffs.diffs), 552)
         # Cycle length is only diff (x3)
         self.assertEqual(diffs.nDiffs(), 3)
 
--- a/armi/bookkeeping/tests/test_historyTracker.py
+++ b/armi/bookkeeping/tests/test_historyTracker.py
@@ -168,6 +168,12 @@ class TestHistoryTracker(ArmiTestHelper):
         )
         b.getVolume()
         bName = b.name
+        print(
+            "XXXXXXXXXXXXXXXXXXXXXXXXXX",
+            b.getLocation(),
+            b.getSymmetryFactor(),
+            b.p.power,
+        )
 
         # duration is None in this DB
         hti = o.getInterface("history")
@@ -184,7 +190,12 @@ class TestHistoryTracker(ArmiTestHelper):
             params[param] = []
             for ts, years in enumerate(timesInYears):
                 cycle, node = utils.getCycleNodeFromCumulativeNode(ts, self.o.cs)
-
+                print(
+                    "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+                    hti.getBlockHistoryVal(bName, param, (cycle, node)),
+                    bName,
+                )
+                print(cycle, node)
                 params[param].append(
                     hti.getBlockHistoryVal(bName, param, (cycle, node))
                 )
@@ -196,7 +207,9 @@ class TestHistoryTracker(ArmiTestHelper):
         # verify the power parameter is retrievable from the history
         self.assertEqual(o.cs["power"], 1000000000.0)
         self.assertAlmostEqual(params["power"][0], 360, delta=0.1)
-        self.assertEqual(params["power"][0], params["power"][1])
+        self.assertEqual(
+            params["power"][0] / 3, params["power"][1]
+        )  # assembly was moved to the central location with 1/3rd symmetry
 
         # verify the power density parameter is retrievable from the history
         self.assertAlmostEqual(params["pdens"][0], 0.0785, delta=0.001)
--- a/armi/bookkeeping/tests/test_historyTracker.py
+++ b/armi/bookkeeping/tests/test_historyTracker.py
@@ -168,12 +168,6 @@ class TestHistoryTracker(ArmiTestHelper):
         )
         b.getVolume()
         bName = b.name
-        print(
-            "XXXXXXXXXXXXXXXXXXXXXXXXXX",
-            b.getLocation(),
-            b.getSymmetryFactor(),
-            b.p.power,
-        )
 
         # duration is None in this DB
         hti = o.getInterface("history")
@@ -190,12 +184,6 @@ class TestHistoryTracker(ArmiTestHelper):
             params[param] = []
             for ts, years in enumerate(timesInYears):
                 cycle, node = utils.getCycleNodeFromCumulativeNode(ts, self.o.cs)
-                print(
-                    "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
-                    hti.getBlockHistoryVal(bName, param, (cycle, node)),
-                    bName,
-                )
-                print(cycle, node)
                 params[param].append(
                     hti.getBlockHistoryVal(bName, param, (cycle, node))
                 )
--- a/armi/bookkeeping/tests/test_historyTracker.py
+++ b/armi/bookkeeping/tests/test_historyTracker.py
@@ -195,9 +195,8 @@ class TestHistoryTracker(ArmiTestHelper):
         # verify the power parameter is retrievable from the history
         self.assertEqual(o.cs["power"], 1000000000.0)
         self.assertAlmostEqual(params["power"][0], 360, delta=0.1)
-        self.assertEqual(
-            params["power"][0] / 3, params["power"][1]
-        )  # assembly was moved to the central location with 1/3rd symmetry
+        # assembly was moved to the central location with 1/3rd symmetry
+        self.assertEqual(params["power"][0] / 3, params["power"][1])
 
         # verify the power density parameter is retrievable from the history
         self.assertAlmostEqual(params["pdens"][0], 0.0785, delta=0.001)
--- a/armi/bookkeeping/tests/test_historyTracker.py
+++ b/armi/bookkeeping/tests/test_historyTracker.py
@@ -231,8 +231,8 @@ class TestHistoryTracker(ArmiTestHelper):
 
         # test that detailAssemblyNames() is working
         self.assertEqual(len(history.detailAssemblyNames), 1)
-        history.addAllFuelAssems()
-        self.assertEqual(len(history.detailAssemblyNames), 51)
+        history.addAllDetailedAssems()
+        self.assertEqual(len(history.detailAssemblyNames), 54)
 
     def test_getBlockInAssembly(self):
         history = self.o.getInterface("history")
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -225,7 +225,7 @@ class FuelHandler:
         if blockLevelMax:
             return a.getChildParamValues(paramName).max() * multiplier
 
-        return a.p[paramName]
+        return a.p[paramName] * multiplier
 
     def findAssembly(
         self,
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -38,6 +38,7 @@ from armi.physics.fuelCycle.fuelHandlerInterface import FuelHandlerInterface
 from armi.physics.fuelCycle.settings import CONF_ASSEMBLY_ROTATION_ALG
 from armi.reactor.flags import Flags
 from armi.utils.customExceptions import InputError
+from armi.reactor.parameters import ParamLocation
 
 
 class FuelHandler:
@@ -212,8 +213,21 @@ class FuelHandler:
     @staticmethod
     def _getParamMax(a, paramName, blockLevelMax=True):
         """Get parameter with Block-level maximum."""
+        print(
+            "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+            a.getBlocks()[0].p.paramDefs.atLocation(ParamLocation.VOLUME_INTEGRATED),
+        )
+        volIntegratedParamNames = [
+            p.name
+            for p in a.getBlocks()[0].p.paramDefs.atLocation(
+                ParamLocation.VOLUME_INTEGRATED
+            )
+        ]
+        multiplier = (
+            a.getSymmetryFactor() if paramName in volIntegratedParamNames else 1.0
+        )
         if blockLevelMax:
-            return a.getChildParamValues(paramName).max()
+            return a.getChildParamValues(paramName).max() * multiplier
 
         return a.p[paramName]
 
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -213,15 +213,11 @@ class FuelHandler:
     @staticmethod
     def _getParamMax(a, paramName, blockLevelMax=True):
         """Get parameter with Block-level maximum."""
-        volIntegratedParamNames = [
-            p.name
-            for p in a.getBlocks()[0].p.paramDefs.atLocation(
-                ParamLocation.VOLUME_INTEGRATED
-            )
-        ]
-        multiplier = (
-            a.getSymmetryFactor() if paramName in volIntegratedParamNames else 1.0
+        isVolumeIntegrated = (
+            a.getBlocks()[0].p.paramDefs[paramName].location
+            == ParamLocation.VOLUME_INTEGRATED
         )
+        multiplier = a.getSymmetryFactor() if isVolumeIntegrated else 1.0
         if blockLevelMax:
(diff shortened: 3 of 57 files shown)


bsculac:
Working on running a test case to demonstrate that the parameters are correctly accounted for in assemblies that are split between symmetric portions of the core.

john-science:
@bsculac Are you hoping this PR will make it into the (imminent) ARMI 0.5.0 release?

bsculac:
> @bsculac Are you hoping this PR will make it into the (imminent) ARMI 0.5.0 release?

I don't think it needs to hold up the release. I'm not sure I'll have enough time to wrap it up before code freeze, I'm out until after Thanksgiving.

armi/reactor/assemblies.py
```diff
@@ -206,13 +206,28 @@ def insert(self, index, obj):
...
         self.clearCache()
+        self.scaleParamsToNewSymmetryFactor(oldSymmetryFactor)
+
+    def scaleParamsToNewSymmetryFactor(self, oldSymmetryFactor=1.0):
+        volIntegratedParamsToScale = self.getBlocks()[0].p.paramDefs.atLocation(
+            ParamLocation.VOLUME_INTEGRATED
+        )
+        scalingFactor = oldSymmetryFactor / self.getSymmetryFactor()
+        for b in self.getBlocks():
+            for param in volIntegratedParamsToScale:
```
john-science, line 225: Okay, so this seems suspect.  You are looping over EVERY parameter in a block, and doing a volume integration.  This is obviously done elsewhere in the code.

So, for instance, you have [a failing test](https://github.com/terrapower/armi/actions/runs/12147041096/job/33872380305?pr=2017#step:6:382), and the error is:

```python
    self.assertEqual(params["power"][0], params["power"][1])
    # AssertionError: 360.0940445113585 != 120.03134817045283
```

So, you are applying this scaling factor to the `power` parameter for this block. Thus tripling the power of every block in the reactor, thus tripling the power of the entire reactor.

That is a HUGE change, and from the description on this PR, my guess is that you don't think ARMI miscalculates the power (and every other parameter) by a factor of 3. RIght? Please say yes.

armi/bookkeeping/db/tests/test_comparedb3.py
```diff
@@ -181,7 +181,7 @@ def test_compareDatabaseSim(self):
             dbs[1]._fullPath,
             timestepCompare=[(0, 0), (0, 1)],
         )
-        self.assertEqual(len(diffs.diffs), 504)
+        self.assertEqual(len(diffs.diffs), 576)
```
john-science, line 184: This is a very big change. I am surprised.

bsculac: We chatted about this, most of these diffs are `None` diffs. I'm not sure I understand the point of that line in particular, it seems to be checking whether the PR is updating the number in accordance with the (new number of parameters on the block) * 3 since the diffs report 3 separate values for each diff (mean, median, max or something like that).

armi/reactor/assemblies.py
bsculac: Resolved not seeing the error on local, this is apparently the sort of change that requires a reinstall of ARMI.

bsculac: I've located one reason for the failing test, the core model used in the test is from the example notebook, and has a dummy power distribution that does not account for the center assembly only having 1/3rd of itself in the core. In progressing the simulation from c0n0 to c0n1 a fuel shuffle occurs. This shuffle exchanges the center assembly (assembly A) with some other assembly in the core (assembly B). So assembly A would triple in power with this feature, moving from the center location with symmetry factor 3 to somewhere with symmetry factor 1. The opposite would happen to assembly B.

bsculac: The example reactor in https://github.com/terrapower/armi/blob/main/armi/tests/tutorials/data_model.ipynb reports **incorrect** power densities for blocks in the central location. This is because the volume that is generated by the `b.getVolume()` call in the screenshot is reduced by the symmetry factor but the power is not. 
image

bsculac: Additionally, the unit test 
`self.assertEqual(params["power"][0], params["power"][1])`
is incorrect to assert this. From what I can tell, ARMI is built to assume assemblies that are split by boundaries do not contribute their entire power values to the modeled section of the core. The 0 index here is when the assembly is wholly within the modeled core, index 1 is after the assembly is moved to the central location and is only 1/3rd in the modeled core.

armi/bookkeeping/db/tests/test_comparedb3.py
john-science: I have rarely seen anyone's PR change this number by more than 1 or 2. 

So, I am suspicious that there is a problem here somewhere. My guess is this number will drop down a lot when the CI starts passing.

armi/reactor/assemblies.py
john-science: @keckler @drewj-tp Brian believe the power in the center assembly is off by 1/3. If he's right, this is a one-line fix in ARMI (and it's already done in this PR). But if he's wrong, I certainly don't want to make the change.

bsculac: This is not true, I am not asserting the power is wrong in the center assembly for all reactors. I haven't checked yet, but I would guess that the `power` parameter is handled specially to account for 1/3rd symmetry, and/or during a core solve the power should be calculated correctly. What I am saying is some other parameters are not adjusted to account for the symmetry factor.

That said, the power used in the data model example notebook is wrong (aside from how it was constructed) because when the low power assembly is shuffled into the center it has the same power as it had before the shuffle. This property is then used in a unit test (the one that was discussed as failing above) to assert something that should not be true.

armi/reactor/assemblies.py
```diff
@@ -206,13 +206,28 @@ def insert(self, index, obj):
...
         # symmetry may have changed (either moving on or off of symmetry line)
         self.clearCache()
+        self.scaleParamsToNewSymmetryFactor(oldSymmetryFactor)
+
+    def scaleParamsToNewSymmetryFactor(self, oldSymmetryFactor=1.0):
+        volIntegratedParamsToScale = self.getBlocks()[0].p.paramDefs.atLocation(
+            ParamLocation.VOLUME_INTEGRATED
+        )
+        scalingFactor = oldSymmetryFactor / self.getSymmetryFactor()
+        for b in self.getBlocks():
```
keckler, line 224: Suggestion: I think you can wrap this in an `if scalingFactor != 1` clause, and avoid needlessly looping over a ton of assemblies that were shuffled but not to/from the central assembly.

bsculac:
John and I talked about this for a bit. I think we arrived at two conclusions:

1. The core idea of adjusting some parameters as an assembly is placed to/from a position like the center position in a 1/3rd symmetric core (i.e. somewhere it would be in multiple symmetric sections of the core) is fine. The specific means may need different implementations, e.g. selecting all volume integrated parameters for adjustment may not be ideal. Also, the knock on effects of this change would have to be thoroughly evaluated.

2. However, the problem this PR is attempting to fix may need a different solution than adjustment of parameters upon an assembly's placement in the core. John was of the opinion that when an assembly is constructed and initially placed in the core it should have no valid parameters that would need adjustment for the symmetry factor, those parameters should be assigned in `interactBOL`. I don't necessarily disagree, but this is a decision the project needs to make (or reaffirm).

@john-science please correct me if I've misrepresented our discussion, or if you'd like to add anything.

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -261,8 +261,9 @@ def test_width(self):
         for ring, power in zip(range(1, 8), range(10, 80, 10)):
             aList = assemsByRing[ring]
             for a in aList:
+                sf = a.getSymmetryFactor()  # center assembly is only 1/3rd in the core
                 for b in a:
-                    b.p.power = power
+                    b.p.power = power / sf
```
onufer, line 266: are there not other params like power. cant we handle all volume integrated params the same??

armi/reactor/assemblies.py
keckler: I am taking a look at this now...

I think that @bsculac 's assessment is correct in that `VOLUME_INTEGRATED` parameters _should_ be adjusted when an assembly is moved to/from a position that has non-unity symmetry factor (most of the time the central assembly). I also think that he is correct in that this is not currently being done by ARMI. 

I think there are a few reasons why this has escaped notice until now:
* Not all models use 1/3 symmetry, in which case this is not relevant
* The major model that we deal with which _does_ use 1/3 symmetry also has an abnormal assembly in that position, which means that it is not easy to spot abnormal behavior based on trends in other assemblies
* Even for models in which everything is right to trigger this bug, it only has a practical impact on a small number of parameters. Most parameters are recalculated in subsequent interfaces where the symmetry factor is accurately accounted for. For those parameters, the "bug" only persists in the time between BOC and when the relevant physics interface is run. But then once the physics interface runs, the parameter is updated and eventually what is written to the database is blissfully correct. This is not the case for all parameters, though. In particular, this PR points out a number of parameters that are really only ever calculated at BOL, and they are not further updated. In these cases, those values WILL BE WRONG when they are written to the database. I believe that Brian's fix will remedy this situation.

So my assessment is that this should be fixed, and the way that Brian has proposed to fix it seems adequate to me.

I do have some small worry that other places in the code might be doing this already in a one-off sense for specific parameters. I think it would be reasonable to rely on our internal integration tests to identify any of these cases.

Just my $0.02.

armi/reactor/assemblies.py
```diff
@@ -206,13 +206,30 @@ def insert(self, index, obj):
...
+        volIntegratedParamsToScale = self.getBlocks()[0].p.paramDefs.atLocation(
+            ParamLocation.VOLUME_INTEGRATED
+        )
+        for b in self.getBlocks():
+            for param in volIntegratedParamsToScale:
+                name = param.name
+                if b.p[name] is None:
+                    continue
+                else:
+                    b.p[name] = b.p[name] * scalingFactor
```
john-science, line 232: What if the volume-integrated parameter is a list or a 2D array of values?

john-science:
@bsculac I see a lot of work was done on this PR yesterday. Great!

Last I heard, you thought this PR wasn't ready. But I see the tests are passing now.

I'm not sure if this PR is ready for final review or not.... could you move this PR out of Draft when you think I should give it a final/real review?

Thanks!

armi/reactor/assemblies.py
bsculac: I was actually working on that when you posted :). Should be in hand now.

armi/reactor/blockParameters.py
```diff
@@ -84,19 +84,25 @@ def getBlockParameterDefinitions():
         pb.defParam(
             "molesHmBOL",
             units=f"{units.MOLES}",
-            description="Total number of atoms of heavy metal at BOL assuming a full assembly",
+            description="Total number of atoms of heavy metal at BOL.",
+            location=ParamLocation.VOLUME_INTEGRATED,
```
mgjarrett, line 88: This works, but it's a little bit confusing to have these parameters defined within a context manager with `location=ParamLocation.AVERAGE`.

bsculac: Moved!

armi/reactor/assemblies.py
```diff
@@ -206,13 +207,34 @@ def insert(self, index, obj):
...
+            for param in volIntegratedParamsToScale:
+                name = param.name
+                if b.p[name] is None or isinstance(b.p[name], str):
+                    continue
+                elif isinstance(b.p[name], np.ndarray):
+                    b.p[name] = b.p[name] * scalingFactor
+                elif isinstance(b.p[name], Iterable):
+                    b.p[name] = [value * scalingFactor for value in b.p[name]]
+                else:
+                    b.p[name] = b.p[name] * scalingFactor
```
john-science, line 237:
```suggestion
                elif isinstance(b.p[name], Iterable):
                    b.p[name] = [value * scalingFactor for value in b.p[name]]
                else:
                    # numpy array or other
                    b.p[name] = b.p[name] * scalingFactor
```

armi/bookkeeping/tests/test_historyTracker.py
```diff
@@ -196,7 +195,9 @@ def test_historyParameters(self):
         # verify the power parameter is retrievable from the history
         self.assertEqual(o.cs["power"], 1000000000.0)
         self.assertAlmostEqual(params["power"][0], 360, delta=0.1)
-        self.assertEqual(params["power"][0], params["power"][1])
+        self.assertEqual(
+            params["power"][0] / 3, params["power"][1]
+        )  # assembly was moved to the central location with 1/3rd symmetry
```
john-science, line 200:
```suggestion
        # assembly was moved to the central location with 1/3rd symmetry
        self.assertEqual(params["power"][0] / 3, params["power"][1])
```

armi/reactor/assemblies.py
john-science: Okay, thanks Chris.

And as long as this logic all works for a non-Natrium situation (where symmetry factor will usually be `1`), I'm happy.

armi/reactor/blockParameters.py
```diff
@@ -161,6 +137,37 @@ def getBlockParameterDefinitions():
             categories=["cumulative"],
         )
 
+    with pDefs.createBuilder(
+        default=0.0, location=ParamLocation.VOLUME_INTEGRATED, categories=["depletion"]
+    ) as pb:
```
john-science, line 142: Just to check. The default value for "save to database" here is "True".

That's what you want, right?

(I got no dog in that fight, I'm just making sure the implied default is what you want.)

bsculac: yup, I copied the context manager from the original context manager that was being used to define these parameters before I broke them out into their own thing. So this should be no change to the functionality as originally specified.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -212,10 +213,19 @@ def _compareAssem(candidate, current):
     @staticmethod
     def _getParamMax(a, paramName, blockLevelMax=True):
         """Get parameter with Block-level maximum."""
+        volIntegratedParamNames = [
+            p.name
+            for p in a.getBlocks()[0].p.paramDefs.atLocation(
+                ParamLocation.VOLUME_INTEGRATED
+            )
+        ]
```
john-science, line 221: Okay, so I see what you're doing here. You are getting _all_ the volume-integrated parameters and then checking to see if the `"paramName"` passed into this method is in that list.

But why not just check if that one parameter is volume-integrated?

```python
a.getBlocks()[0].p.paramDefs[paramName].location == ParamLocation.VOLUME_INTEGRATED
```

Does my way not work? Am I missing something?

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -212,10 +213,19 @@ def _compareAssem(candidate, current):
     @staticmethod
     def _getParamMax(a, paramName, blockLevelMax=True):
         """Get parameter with Block-level maximum."""
+        volIntegratedParamNames = [
```
john-science, line 216:
```suggestion
        # handle special case: volume-integrated parameters where symmetry factor is 1.0
        volIntegratedParamNames = [
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -212,10 +213,19 @@ def _compareAssem(candidate, current):
...
+            p.name
+            for p in a.getBlocks()[0].p.paramDefs.atLocation(
+                ParamLocation.VOLUME_INTEGRATED
+            )
+        ]
+        multiplier = (
+            a.getSymmetryFactor() if paramName in volIntegratedParamNames else 1.0
```
john-science, line 223: Perhaps we could further simplify this logic by first checking if `a.getSymmetryFactor()` is 1?

```python
multiplier = a.getSymmetryFactor()
if multiplier != 1.0:
    # your logic above to calculate volIntegratedParamNames
    multiplier = multiplier if paramName in volIntegratedParamNames else 1.0
```

Does that make sense?  I just mean, if the symmetry factor is 1.0, we don't need to do any of this logic, and we can skip a lot of math by not inspecting parameters.

armi/reactor/assemblies.py
```diff
@@ -206,13 +207,33 @@ def insert(self, index, obj):
...
             self.p.daysSinceLastMove = 0.0
         self.parent.childrenByLocator[locator] = self
         # symmetry may have changed (either moving on or off of symmetry line)
         self.clearCache()
+        self.scaleParamsToNewSymmetryFactor(oldSymmetryFactor)
+
+    def scaleParamsToNewSymmetryFactor(self, oldSymmetryFactor):
+        scalingFactor = oldSymmetryFactor / self.getSymmetryFactor()
+        if scalingFactor == 1:
+            return  # noop
```
john-science, line 223:
```suggestion
            return
```

armi/reactor/assemblies.py
```diff
@@ -206,13 +207,33 @@ def insert(self, index, obj):
...
+        volIntegratedParamsToScale = self.getBlocks()[0].p.paramDefs.atLocation(
```
john-science, line 224:
```suggestion
            return

        volIntegratedParamsToScale = self.getBlocks()[0].p.paramDefs.atLocation(
```

armi/physics/fuelCycle/fuelHandlers.py
bsculac: I was scrounging the logic from elsewhere, probably the logic for growing to full core. I've updated the logic here.

armi/physics/fuelCycle/fuelHandlers.py
bsculac: added!

john-science (approved):

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -212,10 +213,18 @@ def _compareAssem(candidate, current):
     @staticmethod
     def _getParamMax(a, paramName, blockLevelMax=True):
         """Get parameter with Block-level maximum."""
+        multiplier = a.getSymmetryFactor()
+        # handle special case: volume-integrated parameters where symmetry factor is 1.0
+        if multiplier != 1:
```
john-science, line 218:
```suggestion
        if multiplier != 1:
            # handle special case: volume-integrated parameters where symmetry factor is not 1
```

PR #2018, opened 2024-11-21
Improving calculation for total job memory

opotowsky:
## What is the change?

This swaps the calculation for total job memory to use a more straightforward cluster resource request parameter. Some edits and additions to the tests flow from the update.

This makes the math simpler!

---

## Checklist

- [x] ~The [release notes](https://terrapower.github.io/armi/developer/tooling.html#add-release-notes) have been updated if necessary.~ too small to matter

### DIFF
--- a/armi/bookkeeping/tests/test_memoryProfiler.py
+++ b/armi/bookkeeping/tests/test_memoryProfiler.py
@@ -181,7 +181,7 @@ class TestMemoryProfiler(unittest.TestCase):
             self.assertIn("Currently using 6.0 GB of memory.", stdOut)
             self.assertIn("There is 44.0 GB of memory left.", stdOut)
             self.assertIn("There is a total allocation of 50.0 GB", stdOut)
-            # Try another for funzies
+            # Try another for funzies where we only use half the available resources on the node
             mockLogs.emptyStdout()
             self.memPro.cs = {"cpusPerTask": 5, "nTasks": 1}
             self.memPro.printCurrentMemoryState()
--- a/armi/bookkeeping/tests/test_memoryProfiler.py
+++ b/armi/bookkeeping/tests/test_memoryProfiler.py
@@ -139,9 +139,18 @@ class TestMemoryProfiler(unittest.TestCase):
         vMem.total = (1024**3) * 50
         mockVMem.return_value = vMem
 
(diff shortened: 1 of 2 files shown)


john-science:
Wait, neither of these settings exist in ARMI:

```python
self.cs["nTasksPerNode"]  # OLD
self.cs["cpusPerTask"]  # NEW
```

If we are modifying this code, I would prefer to move it out of ARMI.

john-science (approved):

albeanth (commented):
Can you update the docstring for test_getTotalJobMemory?

armi/bookkeeping/tests/test_memoryProfiler.py
```diff
@@ -166,20 +175,26 @@ def test_printCurrentMemoryState(
...
-            csMock = MagicMock()
-            csMock.__getitem__.return_value = 2
-            self.memPro.cs = csMock
+            self.memPro.cs = {"cpusPerTask": 1, "nTasks": 10}
             self.memPro.printCurrentMemoryState()
             stdOut = mockLogs.getStdout()
             self.assertIn("Currently using 6.0 GB of memory.", stdOut)
             self.assertIn("There is 44.0 GB of memory left.", stdOut)
             self.assertIn("There is a total allocation of 50.0 GB", stdOut)
+            # Try another for funzies
```
albeanth, line 184:
```suggestion
            # Try another for funzies where we only use half the available resources on the node
```

opotowsky:
> Can you update the docstring for test_getTotalJobMemory?

I'm not sure how it's inaccurate?

opotowsky:
> Wait, neither of these settings exist in ARMI:
> 
> ```python
> self.cs["nTasksPerNode"]  # OLD
> self.cs["cpusPerTask"]  # NEW
> ```
> 
> If we are modifying this code, I would prefer to move it out of ARMI.

Discussed offline. Removal will happen! Just not in this PR.

albeanth (approved):

albeanth:
> > Can you update the docstring for test_getTotalJobMemory?
> 
> I'm not sure how it's inaccurate? It's the same test as before, just with the new inputs. (and I added a combo or two)

you're right! I was wrong. All good!

PR #2022, opened 2024-11-26
Move files from temporary directory changer

mgjarrett:
## What is the change?

Use `shutil.move` / `mv` instead of `shutil.copy` / `cp` to move files from temporary directory back to the original working directory on `__exit__`.

Copying files from one location to another keeps the original file intact, but the original files in a temporary directory do not need to be kept around after the `__exit__` from the `TemporaryDirectoryChanger`. These files are immediately deleted after the copy operation is completed.

Using `shutil.move()` is significantly faster than `shutil.copy()` when the source and destination locations are on the same drive, since the data doesn't have to physically be copied to a new place in memory.

---

## Checklist

### DIFF
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -836,6 +836,7 @@ def safeCopy(src: str, dst: str) -> None:
 
     runLog.extra("Copied {} -> {}".format(src, dst))
 
+
 def safeMove(src: str, dst: str) -> None:
     """This copy overwrites ``shutil.move`` and checks that move operation is truly completed before continuing."""
     # Convert files to OS-independence
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -844,6 +844,7 @@ def safeMove(src: str, dst: str) -> None:
     dst = os.path.abspath(dst)
     if os.path.isdir(dst):
         dst = os.path.join(dst, os.path.basename(src))
+
     srcSize = os.path.getsize(src)
     if "win" in sys.platform:
         shutil.move(src, dst)
@@ -853,7 +854,7 @@ def safeMove(src: str, dst: str) -> None:
     else:
         raise OSError(
             "Cannot perform ``safeMove`` on files because ARMI only supports "
-            + "Linux and Windows."
+            + "Linux, MacOS, and Windows."
         )
     waitTime = 0.01  # 10 ms
     maxWaitTime = 6000  # 1 min
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -846,8 +846,7 @@ def safeMove(src: str, dst: str) -> None:
         dst = os.path.join(dst, os.path.basename(src))
     srcSize = os.path.getsize(src)
     if "win" in sys.platform:
-        shutil.copyfile(src, dst)
-        shutil.copymode(src, dst)
+        os.replace(src, dst)
     elif "linux" in sys.platform:
         cmd = f'mv "{src}" "{dst}"'
         os.system(cmd)
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -872,6 +872,7 @@ def safeMove(src: str, dst: str) -> None:
             break
 
     runLog.extra("Moved {} -> {}".format(src, dst))
+    return dst
 
 
 # Allow us to check the copy operation is complete before continuing
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -848,7 +848,7 @@ def safeMove(src: str, dst: str) -> None:
     if "win" in sys.platform:
         shutil.move(src, dst)
     elif "linux" in sys.platform:
-        cmd = f'cp "{src}" "{dst}"'
+        cmd = f'mv "{src}" "{dst}"'
         os.system(cmd)
     else:
         raise OSError(
@@ -856,7 +856,7 @@ def safeMove(src: str, dst: str) -> None:
             + "Linux and Windows."
         )
     waitTime = 0.01  # 10 ms
-    maxWaitTime = 60  # 1 min
+    maxWaitTime = 6000  # 1 min
     totalWaitTime = 0
     while True:
         try:
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -859,9 +859,12 @@ def safeMove(src: str, dst: str) -> None:
     maxWaitTime = 60  # 1 min
     totalWaitTime = 0
     while True:
-        dstSize = os.path.getsize(dst)
-        if srcSize == dstSize:
-            break
+        try:
+            dstSize = os.path.getsize(dst)
+            if srcSize == dstSize:
+                break
+        except FileNotFoundError:
+            pass
         time.sleep(waitTime)
         totalWaitTime += waitTime
         if totalWaitTime > maxWaitTime:
--- a/armi/utils/__init__.py
+++ b/armi/utils/__init__.py
@@ -846,7 +846,8 @@ def safeMove(src: str, dst: str) -> None:
         dst = os.path.join(dst, os.path.basename(src))
     srcSize = os.path.getsize(src)
     if "win" in sys.platform:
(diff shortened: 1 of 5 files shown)


armi/utils/__init__.py
```diff
@@ -837,6 +837,49 @@ def safeCopy(src: str, dst: str) -> None:
     runLog.extra("Copied {} -> {}".format(src, dst))
 
 
+def safeMove(src: str, dst: str) -> None:
```
john-science, line 840: Can we get a new unit test for this new function?

If it helps I see the test go something like this:

* Use the `TemporaryDirectoryChanger` to create a temp dir for your test
* create two dirs, one called `src` and one called `dst`
* add a blank fie or two inside `src` (or just add 1 character of text inside them)
* test 1: use this function to move the files from `src/` into `dst`
* test 2: use this function to move the entire `src/` dir into `dst`

Something like that, anyway.

opotowsky (commented):
Looks like the move doesn't work with windows/mac. 

But I'd like to make a higher level comment. The secret overwriting of shutil.copy is something I want to remove from ARMI (i.e., I want to go change all the instances of `shutil.copy` to explicitly use `safeCopy` instead, then remove the overwrite. See #1690.

Maybe for this update it makes sense to keep the same scheme and we just tack this onto that tech debt payoff since it won't really be causing any extra work.

But I'd also be happy if the `shutil.move` calls were just `safeMove` instead. :-)

mgjarrett:
I didn't touch `safeCopy` at all, but I added some `assert` statements to verify that the `safeCopy` call is actually copying the files. The unit test for `safeCopy` passes locally for me, but here it is now failing on those assertions, which is strange.

It's also weird that the existing unit test of `safeCopy` doesn't actually assert that the files are copied successfully. Makes me wonder if there is a known issue here.

armi/utils/__init__.py
mgjarrett: I added some unit tests

armi/utils/__init__.py
```diff
@@ -837,6 +837,47 @@ def safeCopy(src: str, dst: str) -> None:
...
+    if "win" in sys.platform:
+        shutil.move(src, dst)
+    elif "linux" in sys.platform:
+        cmd = f'mv "{src}" "{dst}"'
+        os.system(cmd)
+    else:
+        raise OSError(
+            "Cannot perform ``safeMove`` on files because ARMI only supports "
+            + "Linux and Windows."
+        )
```
opotowsky, line 857: We run unit tests on mac OS now. So is this OSError (here and in other places) accurate? @john-science

opotowsky: (keep in mind Mike is just following the "template" created in `safeCopy`)

opotowsky (approved):
This looks great!

armi/utils/__init__.py
john-science: @mgjarrett ARMI CI runs on Mac, Linux, and Windows. Yet I don't see this error being raised when the unit tests run on MacOS.  (Your CI passes in this PR.)

What's going on?

john-science: Also, I note, every single other feature in ARMI supports the MacOS. I'm not sure why this would be different.

opotowsky: @john-science before macos was added, we added this exact if/elif/else configuration to several places in ARMI. You'll note it is in `safeCopy` just above this in this file.

I'm not sure why the test passes on macos, but note that `test_safeCopy` also passes and it's the same structure.

opotowsky:
> > The secret overwriting of shutil.copy is something I want to remove from ARMI
> 
> I note we already have a ticket for this: #1690
> 
> But I think Michael's PR doesn't touch that, because he didn't alter `safeCopy()`, but added a new `saveMove()`. Am I wrong?

Yep! the first iteration in this PR had a similar setup, (where shutil.move was overwritten with `safeMove`) but that was undone in favor of not overwriting `shutil.move`.

armi/utils/__init__.py
```diff
@@ -837,6 +837,47 @@ def safeCopy(src: str, dst: str) -> None:
...
+    srcSize = os.path.getsize(src)
+    if "win" in sys.platform:
+        shutil.move(src, dst)
+    elif "linux" in sys.platform:
+        cmd = f'mv "{src}" "{dst}"'
+        os.system(cmd)
+    else:
+        raise OSError(
+            "Cannot perform ``safeMove`` on files because ARMI only supports "
+            + "Linux and Windows."
```
john-science, line 856:
```suggestion
            + "Linux, MacOS, and Windows."
```

armi/utils/__init__.py
```diff
@@ -837,6 +837,47 @@ def safeCopy(src: str, dst: str) -> None:
...
+def safeMove(src: str, dst: str) -> None:
+    """Check that a file has been successfully moved before continuing."""
+    # Convert files to OS-independence
+    src = os.path.abspath(src)
+    dst = os.path.abspath(dst)
+    if os.path.isdir(dst):
+        dst = os.path.join(dst, os.path.basename(src))
+    srcSize = os.path.getsize(src)
+    if "win" in sys.platform:
+        shutil.move(src, dst)
```
john-science, line 849:
```suggestion
    if os.path.isdir(dst):
        dst = os.path.join(dst, os.path.basename(src))

    srcSize = os.path.getsize(src)
    if "win" in sys.platform:
        shutil.move(src, dst)
```

armi/utils/__init__.py
```diff
@@ -837,6 +837,48 @@ def safeCopy(src: str, dst: str) -> None:
...
+    """Check that a file has been successfully moved before continuing."""
+    # Convert files to OS-independence
+    src = os.path.abspath(src)
+    dst = os.path.abspath(dst)
+    if os.path.isdir(dst):
+        dst = os.path.join(dst, os.path.basename(src))
+
+    srcSize = os.path.getsize(src)
+    if "win" in sys.platform:
+        shutil.move(src, dst)
```
john-science, line 850:
```suggestion
    if "win" in sys.platform:
        # this covers Windows ("win32") and MacOS ("darwin")
        shutil.move(src, dst)
```

armi/utils/__init__.py
```diff
@@ -837,6 +837,48 @@ def safeCopy(src: str, dst: str) -> None:
...
+    elif "linux" in sys.platform:
+        cmd = f'mv "{src}" "{dst}"'
+        os.system(cmd)
+    else:
+        raise OSError(
+            "Cannot perform ``safeMove`` on files because ARMI only supports "
+            + "Linux, MacOS, and Windows."
+        )
+    waitTime = 0.01  # 10 ms
```
john-science, line 859:
```suggestion
        )

    waitTime = 0.01  # 10 ms
```

armi/utils/__init__.py
```diff
@@ -837,6 +837,48 @@ def safeCopy(src: str, dst: str) -> None:
...
+        time.sleep(waitTime)
+        totalWaitTime += waitTime
+        if totalWaitTime > maxWaitTime:
+            runLog.warning(
+                f"File move from {dst} to {src} has failed due to exceeding "
+                + f"a maximum wait time of {maxWaitTime/60} minutes."
+            )
+            break
+
+    runLog.extra("Moved {} -> {}".format(src, dst))
```
john-science, line 878: Okay, this print-out that says the move was successful will happen even if the move times out.

Let me fix that.

armi/utils/__init__.py
```diff
@@ -837,6 +837,48 @@ def safeCopy(src: str, dst: str) -> None:
...
+        except FileNotFoundError:
+            pass
+        time.sleep(waitTime)
+        totalWaitTime += waitTime
+        if totalWaitTime > maxWaitTime:
+            runLog.warning(
+                f"File move from {dst} to {src} has failed due to exceeding "
+                + f"a maximum wait time of {maxWaitTime/60} minutes."
+            )
+            break
```
john-science, line 876:
```suggestion
            return
```

john-science (approved):

PR #1998, opened 2024-11-01
Fixing a couple of plots to use initial block height

john-science:
## What is the change?

Moving certain plots to use the BOL block height.

## Why is the change being made?

This is to improve consistency in our block plots. 

close #1158

## SCR Information

Change Type: fixes

One-Sentence Description: Fixing a couple of plots to use initial block heightt.

One-line Impact on Requirements: NA: None of these plots of requirements.

---

## Checklist

- [x] This PR has only [one purpose or idea](https://terrapower.github.io/armi/developer/tooling.html#one-idea-one-pr).
- [x] [Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify any new/changed code.
- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The commit message(s) follow [good practices](https://terrapower.github.io/armi/developer/tooling.html).
- [x] The [documentation](https://terrapower.github.io/armi/developer/tooling.html#document-it) is still up-to-date in the `doc` folder.
- [x] The dependencies are still up-to-date in `pyproject.toml`.

Issue #1158, opened 2023-02-06
Clashing prints on blueprints pictures


keckler:
During a run, ARMI prints out pictures of what the various assembly blueprints look like axially. It appears when axial thermal expansion is turned on, the hot block heights are printed in the same picture directly on top of the cold block heights, making it difficult to read either of them.

It's hard for me to put up an example picture of this without exposing proprietary data, so feel free to reach out to me personally if you want an example.

john-science:
@keckler I don't suppose you could provide an example plot for Tony?

@albeanth Do you have any thoughts on this?  (Also, should we do a round-up of all the axial expansion-related tickets that are lying around?)

keckler:
Yeah, I can easily get him an example of the issue. I can send privately.

keckler:
Here is a redacted example
![image](https://github.com/user-attachments/assets/29a6d6c5-524a-4808-bb34-9b8dfb016e1e)

keckler:
It seems to me that the best way to solve this would be for these pictures to just represent the cold heights. That would avoid the clashing prints due to thermal expansion introducing tiny differences in the block heights.

Besides, my assumption is that users use these types of plots to (1) verify their inputs, and (2) to show other people what their assemblies look like. Both of those tasks would be more digestible if the block heights were displayed as cold heights.

albeanth:
> .... If we agree on this, then I can make a PR for this.

Also supports the proposed direction in https://github.com/terrapower/armi/discussions/1824

john-science:
> If we agree on this, then I can make a PR for this.

I like the proposal. Go for it. And thanks, man.

keckler:
The suggestion I made above (to print the plots at cold heights) is not easily to do in practice. The reason for this is that axial thermal expansion takes place before these plots can possibly be generated.

The plots are generated within the `ReportInterface.interactBOL()` hook:
https://github.com/terrapower/armi/blob/96a5c88d6167cee989378599b6d341c351c291bd/armi/bookkeeping/report/reportInterface.py#L61

Oddly, the axial thermal expansion takes place even _before the `interactInit()` hooks of the interface stack_. This means that basically nothing can be configured to happen before axial thermal expansion messes with the defined reactor. This is a real shame, and highlights another motivation for [converting axial thermal expansion into a plugin/interface](https://github.com/terrapower/armi/issues/1843#issuecomment-2322171844).

keckler:
[...]

PR #2030, opened 2024-12-10
Force plotAssemblyTypes to always use instantiated assemblies, and al…

keckler:
This is meant to improve #1998 by allowing the specification of hot versus cold heights in the plots. Additionally, a couple clean ups are implemented:
1. Change `plotAssemblyTypes()` so that it always operates based on instantiated assemblies. Allowing users to pass in blueprints was making things really messy, and provided absolutely no benefit because blueprints already contain `Assembly` objects in the first place through `blueprints.assemblies.values()`.
2. The point above means that the `plotAssemblyTypes()` method signature is changed. That signature is/was pretty ridiculous anyways, so I feel justified in changing it. I updated all calls to that method that I could find.

I would not expect that there are many downstream calls to this `plotAssemblyTypes()` method. Checking all the projects that I have access to, I found it used in exactly one place (one of my own internal repos).


keckler:
I am relatively sure that this is working now. I tested it with a simply dummy case, however I am unable to test it with my big model due to some dependency conflicts. I think it is ready for your eyes @john-science

john-science:
Love it!

1. Can we put something under the 0.5.1 release notes for "bug fixes"?
2. Does this close a ticket we can link to it?
3. Should we close my previous PR on the matter?

john-science (approved):
Love it

keckler:
> Love it!
> 
> 1. Can we put something under the 0.5.1 release notes for "bug fixes"?
> 2. Does this close a ticket we can link to it?
> 3. Should we close my previous PR on the matter?

1. I added the release note
2. See point (3) below
3. My intention was to merge these changes into your branch (PR #1998 ). You'll note that I currently have this PR pointed that way and was working off of your branch when I developed this code. If we do that, then I think your PR is already linked to the appropriate ticket (#1158 ).

Let me know what you think of that plan.

john-science:
> 3\. You'll note that I currently have this PR pointed that way and was working off of your branch

Ah!  No, I didn't notice that. I just assumed this was pointed at `main`.

Thanks!

PR #2031, opened 2024-12-11
Provide better composite iteration methods

### LINKED ISSUE #1915: Provide more Composite.iter* methods
ARMI does a lot of iteration over the composite tree. Understandably so, we have lots of methods that support iterating over children of an object, with some filtering. These usually follow the pattern of `parent.getChildren*` like `Composite.getChildren`, `Composite.getChildrenWithFlags`, and `Composite.getChildrenOfType`, further expanded to things like `Assembly.getBlocks`.

A lot of instances of the usage of these `get*` methods are for iteration

https://github.com/terrapower/armi/blob/96a5c88d6167cee989378599b6d341c351c291bd/armi/reactor/components/component.py#L965

https://github.com/terrapower/armi/blob/96a5c88d6167cee989378599b6d341c351c291bd/armi/reactor/converters/axialExpansionChanger/expansionData.py#L253

https://github.com/terrapower/armi/blob/96a5c88d6167cee989378599b6d341c351c291bd/armi/reactor/blocks.py#L633

The tradeoff is these `get*` methods create and populate lists, and then return those lists back to the user. This is great if you need to do multiple iterations or check sizing of the returned object, but it's inefficient for iteration.

I propose we provide `iter*` methods with similar signatures as their `get*` counterparts that do not return lists of things. Instead, they could `yield` back components, produce a `filter` object, etc. something that allows one-pass iteration over things. The `get*` methods could then be distilled to
```python
def getChildrenWithFlags(self, ...):
    return list(self.iterChildrenWithFlags(...))
```
for backwards compatibility.

## Candidate methods
To keep the scope clear, I propose we add, test, and use (where appropriate)
- `Composite.iterChildren`
- `Composite.iterChildrenWithFlags`
- `Composite.iterChildrenOfType`

There are other composite subclasses that I believe could benefit from more iteration methods. But for start, I'll keep the scope limited to `Composite` methods

drewj-tp:
## What is the change?

The high-level idea is to provide ways for users and developers to traverse the composite tree without using lists. For example, calling
```python
for c in r.getChildren(deep=True):
```
first traverses the entire tree and populates a list of all composites in the reactor tree. Then, it returns that list to the caller for iteration. 

This PR adds the following `Composite` methods:

- `iterChildren`
- `iterChildrenWithFlags`
- `iterChildrenOfType`
- `iterChildrenWithMaterials`
  - I hate this one because it's so specific I don't see end users using it often.
  - But it's vital for MPI distribution that we should have a robust mechanism for it.
  - But it's only used in a few places that I don't think it justifies being a `Composite` method, but I can't easily think of a better place for it to live
  - :shrug:

This PR adds the following specific methods

- `Assembly.iterBlocks`

The various `get*` methods still return lists of things for backwards compatibility, but rely on the `iter*` methods. Ideally `Composite.getChildren` would just be `list(self.getChildren(...))` but we need to support `Composite.getChildren(..., includeMaterials=True)` so that's some additional complexity.

In a few places where we use the `get*` methods for once-through iteration, those are replaced with `iter*`. So our example above would instead look lik
```python
for c in r.iterChildren(deep=True):
```

There are still some use cases for making a list, especially if you want to iterate through the thing twice. Something like
```python
objs = r.iterChildren(deep=True)
for o in objs:
    # do something on the first pass
second = list(objs)
```
produces an empty list for `second` because the iterator `objs` is exhausted in the loop.

Efficiency, both time (remove redundant traversal over tree) and space (remove list of children)

Closes #1915 

---

## Checklist

### DIFF
--- a/armi/physics/fuelCycle/tests/test_fuelHandlers.py
+++ b/armi/physics/fuelCycle/tests/test_fuelHandlers.py
@@ -448,7 +448,7 @@ class TestFuelHandler(FuelHandlerTestHelper):
             self.r.p.cycle = cycle
             fh.cycle = cycle
             fh.manageFuel(cycle)
-            for a in self.r.excore["sfp"].getChildren():
+            for a in self.r.excore["sfp"]:
                 self.assertEqual(a.getLocation(), "SFP")
             for b in self.r.core.getBlocks(Flags.FUEL):
                 self.assertGreater(b.p.kgHM, 0.0, "b.p.kgHM not populated!")
@@ -478,7 +478,7 @@ class TestFuelHandler(FuelHandlerTestHelper):
         ensure repeatability.
         """
         # check labels before shuffling:
-        for a in self.r.excore["sfp"].getChildren():
+        for a in self.r.excore["sfp"]:
             self.assertEqual(a.getLocation(), "SFP")
 
         # do some shuffles
@@ -512,7 +512,7 @@ class TestFuelHandler(FuelHandlerTestHelper):
         # make sure the shuffle was repeated perfectly.
         for a in self.r.core.getAssemblies():
             self.assertEqual(a.getName(), firstPassResults[a.getLocation()])
-        for a in self.r.excore["sfp"].getChildren():
+        for a in self.r.excore["sfp"]:
             self.assertEqual(a.getLocation(), "SFP")
 
         # Do some cleanup, since the fuelHandler Interface has code that gets
--- a/armi/physics/fuelCycle/tests/test_fuelHandlers.py
+++ b/armi/physics/fuelCycle/tests/test_fuelHandlers.py
@@ -791,7 +791,7 @@ class TestFuelHandler(FuelHandlerTestHelper):
 
         # grab an arbitrary fuel assembly from the core and from the SFP
         a1 = self.r.core.getAssemblies(Flags.FUEL)[0]
-        a2 = self.r.excore["sfp"].getChildren(Flags.FUEL)[0]
+        a2 = self.r.excore["sfp"].getChildrenWithFlags(Flags.FUEL)[0]
 
         # grab the stationary blocks pre swap
         a1PreSwapStationaryBlocks = [
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -879,9 +879,7 @@ class Assembly(composites.Composite):
         return zip(blocks, zCoords)
 
     def hasContinuousCoolantChannel(self):
-        return all(
-            b.containsAtLeastOneChildWithFlags(Flags.COOLANT) for b in self.getBlocks()
-        )
+        return all(b.containsAtLeastOneChildWithFlags(Flags.COOLANT) for b in self)
 
     def getFirstBlock(self, typeSpec=None, exact=False):
         bs = self.getBlocks(typeSpec, exact=exact)
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -304,10 +304,8 @@ class Assembly(composites.Composite):
         -------
         This is a bit design-specific for pinned assemblies
         """
-        plenumBlocks = self.getBlocks(Flags.PLENUM)
-
         plenumVolume = 0.0
-        for b in plenumBlocks:
+        for b in self.iterBlocks(Flags.PLENUM):
             cladId = b.getComponent(Flags.CLAD).getDimension("id")
             length = b.getHeight()
             plenumVolume += (
@@ -317,7 +315,7 @@ class Assembly(composites.Composite):
 
     def getAveragePlenumTemperature(self):
         """Return the average of the plenum block outlet temperatures."""
-        plenumBlocks = self.getBlocks(Flags.PLENUM)
+        plenumBlocks = self.iterBlocks(Flags.PLENUM)
         plenumTemps = [b.p.THcoolantOutletT for b in plenumBlocks]
 
         # no plenum blocks, use the top block of the assembly for plenum temperature
@@ -792,6 +790,31 @@ class Assembly(composites.Composite):
         with open(fName, "w") as pkl:
             pickle.dump(self, pkl)
 
+    def iterBlocks(self, typeSpec=None, exact=False):
+        """Produce an iterator over all blocks in this assembly from bottom to top.
+
+        Parameters
+        ----------
+        typeSpec : Flags or list of Flags, optional
+            Restrict returned blocks to have these flags.
+        exact : bool, optional
+            If true, only produce blocks that have those exact flags.
+
+        Returns
+        -------
+        iterable of Block
+
+        See also
+        --------
+        * :meth:`__iter__` - if no type spec provided, assemblies can be
+          naturally iterated upon.
+        * :meth:`iterChildrenWithFlags` - alternative if you know you have
+           a type spec that isn't ``None``.
+        """
+        if typeSpec is None:
+            return iter(self)
+        return self.iterChildrenWithFlags(typeSpec, exact)
+
     def getBlocks(self, typeSpec=None, exact=False):
         """
         Get blocks in an assembly from bottom to top.
@@ -808,10 +831,7 @@ class Assembly(composites.Composite):
         blocks : list
             List of blocks.
         """
-        if typeSpec is None:
-            return self.getChildren()
-        else:
-            return self.getChildrenWithFlags(typeSpec, exactMatch=exact)
+        return list(self.iterBlocks(typeSpec, exact))
 
     def getBlocksAndZ(self, typeSpec=None, returnBottomZ=False, returnTopZ=False):
         """
@@ -1169,7 +1189,7 @@ class Assembly(composites.Composite):
         blockCounter : int
             number of blocks of this type
         """
-        return len(self.getBlocks(blockTypeSpec))
+        return sum(1 for _ in self.iterBlocks(blockTypeSpec))
 
     def getDim(self, typeSpec, dimName):
         """
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -904,14 +904,11 @@ class Assembly(composites.Composite):
             return None
 
     def getFirstBlockByType(self, typeName):
-        bs = [
-            b
-            for b in self.getChildren(deep=False)
-            if isinstance(b, blocks.Block) and b.getType() == typeName
-        ]
-        if bs:
-            return bs[0]
-        return None
+        blocks = filter(lambda b: b.getType() == typeName, self)
+        try:
+            return next(blocks)
+        except StopIteration:
+            return None
 
     def getBlockAtElevation(self, elevation):
         """
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -882,10 +882,25 @@ class Assembly(composites.Composite):
         return all(b.containsAtLeastOneChildWithFlags(Flags.COOLANT) for b in self)
 
     def getFirstBlock(self, typeSpec=None, exact=False):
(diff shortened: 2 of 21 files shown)


john-science:
This feature is definitely big enough for a release note.  Since we are planning this PR for just post-release, it should go under `0.5.1` "API Changes":

https://github.com/terrapower/armi/blob/68268c153ec9e9b73dcc0dbe5fa1cdc1a1298731/doc/release/0.5.rst?plain=1#L5-L15

john-science:
Okay, on first pass I like this. It seems straight forward, and I don't see any problems.

I'm going to hold off until Monday to give it a "final review", and hopefully merge it. #fingers-crossed

Nice work!

jakehader:
@drewj-tp - are there any immediate perform gains from this change that you can share in memory and/or speed? I like the idea of iterators, but I'm curious from a macroscopic perspective of managing the entire reactor how this propagates.

Thanks!

drewj-tp:
> @drewj-tp - are there any immediate performance gains from this change that you can share in memory and/or speed? I like the idea of iterators, but I'm curious from a macroscopic perspective of managing the entire reactor how this propagates.
> 
> Thanks!

Good questions. I ran some tests on the FFTF model and, when traversing the entire reactor tree:
- 30x memory reduction
- 35% speed up

We'll still see some benefit if users don't migrate to the new `iterChildren` and still have large loops over `r.getChildren()`. With the new implementation, where `getChildren(...)` is basically `list(c.iterChildren(...))`, we get 18x memory reduction. The speed up is negligible and/or in the noise.

<details><summary>Memory</summary>

```python
import pathlib
import pickle

import armi

import tracemalloc

armi.configure()

pickleJar = pathlib.Path("fftf.pkl")
if pickleJar.exists():
    REACTOR = pickle.loads(pickleJar.read_bytes())
else:
    REACTOR = armi.init(fName="fftf/FFTF.yaml").r
    pickleJar.write_bytes(pickle.dumps(REACTOR))

tracemalloc.start()

for _ in REACTOR.getChildren(deep=True):
    pass

firstSize, firstPeak = tracemalloc.get_traced_memory()

tracemalloc.reset_peak()

for _ in REACTOR.iterChildren(deep=True):
    pass

secondSize, secondPeak = tracemalloc.get_traced_memory()


print(f"{firstSize=} :: {firstPeak=}")
print(f"{secondSize=} :: {secondPeak=}")
print(f"{firstSize/secondSize=} :: {firstPeak/secondPeak=}")
```
</details>

<details><summary>Time</summary>

```python
import pathlib
import pickle

import armi

import timeit

armi.configure()

pickleJar = pathlib.Path("fftf.pkl")
if pickleJar.exists():
    REACTOR = pickle.loads(pickleJar.read_bytes())
else:
    REACTOR = armi.init(fName="fftf/FFTF.yaml").r
    pickleJar.write_bytes(pickle.dumps(REACTOR))


def getChildren(r):
    list(r.getChildren(deep=True))

def iterChildren(r):
    list(r.iterChildren(deep=True))

N = 500

FAKE_GLOBALS = locals()

tGet = timeit.timeit("getChildren(REACTOR)", number=N, globals=FAKE_GLOBALS) / N

tIter = timeit.timeit("iterChildren(REACTOR)", number=N, globals=FAKE_GLOBALS) / N

print(f"{tGet=}")
print(f"{tIter=}")
print(f"{tIter/tGet=}")
```
</details>

john-science:
> * 30x memory reduction
> * 35% speed up

That's awesome!

I would always argue strongly that you will get more performance improvements if/when we elbow the downstream repositories to start using these new tools.

Hell, we could grep around and open the PRs ourselves.

Just sayin.

drewj-tp:
>I would always argue strongly that you will get more performance improvements if/when we elbow the downstream repositories to start using these new tools.

> Hell, we could grep around and open the PRs ourselves

I have a follow on PR here that will help and then I'll start pinging projects on improvements

doc/release/0.5.rst
```diff
@@ -9,6 +9,9 @@ Release Date: TBD
 New Features
 ------------
 #. Move instead of copy files from TemporaryDirectoryChanger. (`PR#2022 <https://github.com/terrapower/armi/pull/2022>`_)
+#. Provide ``Composite.iterChildren``, ``Composite.iterChildrenWithFlags``, ``Composite.iterChildrenOfType``,
```
john-science, line 12: ARMI 0.5.0 was released.

Can we move this up in this file to the 0.5.1 release notes? 

Thanks!

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -498,7 +498,7 @@ def test_repeatShuffles(self):
         ensure repeatability.
         """
         # check labels before shuffling:
-        for a in self.r.excore["sfp"].getChildren():
+        for a in self.r.excore["sfp"]:
```
john-science, line 501: This called `iterChildren()` under the hood now, right?

armi/reactor/assemblies.py
```diff
@@ -814,6 +812,31 @@ def dump(self, fName=None):
         with open(fName, "w") as pkl:
             pickle.dump(self, pkl)
 
+    def iterBlocks(self, typeSpec=None, exact=False):
```
john-science, line 815: So, why is `iterBlocks()` necessary? For an `Assembly` this is the same as iterating through all the children (shallow, not deep), right?

doc/release/0.5.rst
drewj-tp: This is in the 0.5.1 section

armi/reactor/composites.py
```diff
@@ -525,6 +525,10 @@ def getChildren(self, deep=False, generationNum=1, includeMaterials=False):
         """Return the children of this object."""
         raise NotImplementedError
 
+    def iterChildrenWithFlags(self, typeSpec: TypeSpec, exactMatch=True):
+        """Produce an iterator of children that have given flags."""
+        raise NotImplementedError()
```
john-science, line 530: Not sure I follow.  Why is this not implemented? It's `iterChildren` with a predicate, right?  Can that not be this high up the hierarchy for some reason?

armi/physics/fuelCycle/tests/test_fuelHandlers.py
drewj-tp: Yes / no. `getChildren` uses `iterChildren` but if you're just trying to iterate over the immediate children of composite, using `Composite.__iter__` works.

john-science:
@drewj-tp Love it!

I have a few small questions for you, at your leisure.

But I love the feature. Assuming your downstream testing suite passes, I'm convinced.

armi/reactor/assemblies.py
drewj-tp: You're totally right. I added this because I wanted to do a simple find/replace `Assembly.getBlocks` -> `Assembly.iterBlocks` in cases where it made sense. But I could just as easily replace that with `iterChildren` or `iterChildrenWithFlags` as appropriate. Then we have one less method floating around

armi/reactor/composites.py
drewj-tp: Oh excellent point. I'll move the `Composite.iterChildrenWithFlags` up to the `ArmiObject`

armi/reactor/assemblies.py
john-science: I was just thinking of "The Zen of Python":

> There should be one-- and preferably only one --obvious way to do it.

It is easier on users if the API is smaller.

BUT I won't die on this hill.

armi/reactor/composites.py
drewj-tp: Okay so `ArmiObject.iterChildren` is not defined and it maybe could be because `ArmiObject.getChildren` is defined

https://github.com/terrapower/armi/blob/c078cde00832c559586cbcc255356bac445a7a30/armi/reactor/composites.py#L524-L526

Since `Composite.iterChildrenWithFlags` relies on `Composite.__iter__` with `filter(lambda o: o.hasFlags(...), self)` (more or less) we could either
1. Implement `ArmiObject.iterChildrenWithFlags` as `ArmiObject.iterChildren(predicate=...)`, or
2. Enforce/assume that *every* subclass of `ArmiObject` is iterable and produces further `ArmiObject`s upon iteration (okay for non-leafs) and then implement `ArmiObject.iterChildrenWithFlags` as a filtered iteration over `self`

Both get us to the same place. My question is more from an object-oriented perspective: is it acceptable to assume/enforce that `ArmiObject`s should have `__iter__` implemented _somehow_?

drewj-tp: Opted to rely on the presence of `ArmiObject.iterChildren -> Iterator[ArmiObject]` which allows us to build a lot of iter/get functionality

drewj-tp:
>Assuming your downstream testing suite passes, I'm convinced

Internal tests are passing!

armi/reactor/assemblies.py
```diff
@@ -1191,7 +1204,11 @@ def countBlocksWithFlags(self, blockTypeSpec=None):
         blockCounter : int
             number of blocks of this type
         """
-        return len(self.getBlocks(blockTypeSpec))
+        if blockTypeSpec is None:
```
john-science, line 1207: Not your fault, but can we fix the docstring in this method?  It says:

> Returns the block at a specified axial dimension elevation (given in cm).

But it looks like it should say:

> Returns the number of blocks at a specified axial dimension elevation (given in cm).

Also, we can just delete this text from the docstring:

> Used as a way to determine what block the control rod will be modifying with a mergeBlocks.

john-science (approved):
I love it!

Great change.

We'll merge when we've looked at downstream testing.

drewj-tp:
> We'll merge when we've looked at downstream testing.

Internal tests are passing 🚀

doc/release/0.5.rst
```diff
@@ -10,6 +10,9 @@ New Features
 ------------
 #. Move instead of copy files from TemporaryDirectoryChanger. (`PR#2022 <https://github.com/terrapower/armi/pull/2022>`_)
 #. Creating the ``armi.testing`` module, to share ARMI testing tools. (`PR#2028 <https://github.com/terrapower/armi/pull/2028>`_)
+#. Provide ``Composite.iterChildren``, ``Composite.iterChildrenWithFlags``, ``Composite.iterChildrenOfType``,
+   ``Composite.iterChildrenWithMaterials``, for efficient composite tree traversal.
+   (`PR#2031 <https://github.com/terrapower/armi/pull/2031>`_)
```
john-science, line 15:
```suggestion
```

armi/reactor/assemblies.py
drewj-tp: > Not your fault, but can we fix the docstring in this method? It says:
> 
>> Returns the block at a specified axial dimension elevation (given in cm).
> 
> But it looks like it should say:
> 
>> Returns the number of blocks at a specified axial dimension elevation (given in cm).

I think github UI smashed two methods together (`getBlockAtElevation` and `countBlocksWithFlags`). 

> Also, we can just delete this text from the docstring:
> 
>> Used as a way to determine what block the control rod will be modifying with a mergeBlocks.

Updated

armi/reactor/composites.py
```diff
@@ -2592,9 +2621,106 @@ def setChildren(self, items):
...
+    def _iterChildren(
+        self, deep: bool, generationNum: int, checker: Callable[["Composite"], bool]
+    ) -> Iterator["Composite"]:
+        if deep or generationNum == 1:
+            yield from filter(checker, self)
+        if deep or generationNum > 1:
+            for c in self:
+                yield from c._iterChildren(deep, generationNum - 1, checker)
+
+    def iterChildrenWithMaterials(self, *args, **kwargs) -> Iterator:
```
john-science, line 2695: I do not see any unit tests for this new method. Thoughts?

Issue #2034, opened 2024-12-13
Snapshots need the same number of cycles and nodes as the original run


onufer:
If you do snapshot and don't have cycles and nodes consistent, you get the following error.

Cycles and nodes don't have meaning for snapshots, so it is probably best to not force this and calculate partial cycle rx swing:

![image](https://github.com/user-attachments/assets/2c6387e0-461f-4f25-b3a6-ae2f1f7d05b7)

PR #2035, opened 2024-12-13
Stop checking for last node in Snapshots

onufer:
## What is the change?

Stop checking if its the last node since this often breaks for snapshots and just calculate reactivity swing at all nodes.

If you do snapshot and don't have cycles and nodes consistent, you get an error.

Cycles and nodes don't have meaning for snapshots, so it is probably best to not force this and calculate partial cycle reactivity swing.

see #2034

## SCR Information

Change Type: features

One-Sentence Description: Stop checking if its the last node since this often breaks for snapshots and just calculate reactivity swing at all nodes.

---

## Checklist

### DIFF
--- a/doc/release/0.5.rst
+++ b/doc/release/0.5.rst
@@ -9,6 +9,7 @@ Release Date: TBD
 New Features
 ------------
 #. Move instead of copy files from TemporaryDirectoryChanger. (`PR#2022 <https://github.com/terrapower/armi/pull/2022>`_)
+#. `rXSwing`` is now calculated for all time nodes making snapshots less failure prone due to non-relevant  inputs. (`PR#2035 <https://github.com/terrapower/armi/pull/2035>`_)
 
 API Changes
 -----------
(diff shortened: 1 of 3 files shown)


doc/release/0.5.rst
```diff
@@ -9,6 +9,7 @@ Release Date: TBD
 New Features
 ------------
 #. Move instead of copy files from TemporaryDirectoryChanger. (`PR#2022 <https://github.com/terrapower/armi/pull/2022>`_)
+#. `rXSwing`` is now calculated for all time nodes making snapshots less failure prone due to non-relevant  inputs. (`PR#2035 <https://github.com/terrapower/armi/pull/2035>`_)
```
john-science, line 12: ARMI 0.5.0 was released.

Can we move this up in this file to the 0.5.1 release notes? 

Thanks!

john-science, line 12:
```suggestion
```

PR #2039, opened 2024-12-19
First order deps

### LINKED ISSUE #2038: ARMI has first-order dependencies that are not declared in our `pyproject.toml`
Okay, ARMI is not declaring these runtime first-order dependencies, because (I am guessing) they are also second-order dependencies. But I consider this a failure of documentation:

* `python-dateutil` - used in `armi/utils/parsing.py`
* ~mpl_toolkits~ - NVM, this is part of mat-props

Similarly, ARMI is not declaring this first-order test dependence:

* `nbformat` - used in `test_historyTracker.py` and `test_notebooks.py`

<br/>

Lastly, ARMI has two dependencies that I find suspect because we are importing an ENTIRE dependency for one function:

* `pyDOE` - only used in [one place](https://github.com/terrapower/armi/blob/c078cde00832c559586cbcc255356bac445a7a30/armi/cases/suiteBuilder.py#L444-L449)
* `python-dateutil` - only used [once](https://github.com/terrapower/armi/blob/c078cde00832c559586cbcc255356bac445a7a30/armi/utils/parsing.py#L97-L99), to read datetime strings _slightly_ more easily.

john-science:
## What is the change?

I added two first-order dependencies to the `pyproject.toml` file.

These are ALSO second-order dependencies. So ARMI builds fine with the `pyproject.toml` file as-is. But I consider this a failure of documentation. And Drew notes that we have less control as things stand.

Generally, I just want to declare all our first-order dependencies as such.

close #2038

(NOTE: I thought we had another first-order dep, mpl_toolkits, but it turns out that is just part of the matprops namespace.)

---

## Checklist

- [x] The [release notes](https://terrapower.github.io/armi/developer/tooling.html#add-release-notes) have been updated if necessary. (I don't believe this warrants a release note.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,6 @@ dependencies = [
     "h5py>=3.9 ; python_version >= '3.11.0'", # Needed because our database files are H5 format
     "htmltree>=0.7.6", # Our reports have HTML output
     "matplotlib>=3.5.3,<3.8.0", # Important plotting library
-    "mpl_toolkits", # For adjusting matplotlib sub-plots
     "numpy>=1.21", # Important math library
     "ordered-set>=3.1.1", # A useful data structure
     "pluggy>=1.2.0", # Central tool behind the ARMI Plugin system
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,17 +31,19 @@ authors = [
 ]
 dependencies = [
     "coverage>=7.2.0", # Code coverage tool. Sadly baked into every Case.
-    "h5py>=3.9 ; python_version >= '3.11.0'", # Needed because our database files are H5 format
+    "dateutil>=2.2", # To read a datetime string more easily.
     "h5py>=3.0,<=3.9 ; python_version < '3.11.0'",
+    "h5py>=3.9 ; python_version >= '3.11.0'", # Needed because our database files are H5 format
     "htmltree>=0.7.6", # Our reports have HTML output
     "matplotlib>=3.5.3,<3.8.0", # Important plotting library
+    "mpl_toolkits", # For adjusting matplotlib sub-plots
     "numpy>=1.21", # Important math library
     "ordered-set>=3.1.1", # A useful data structure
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -31,17 +31,18 @@ authors = [
...
+    "h5py>=3.9 ; python_version >= '3.11.0'", # Needed because our database files are H5 format
     "htmltree>=0.7.6", # Our reports have HTML output
     "matplotlib>=3.5.3,<3.8.0", # Important plotting library
     "numpy>=1.21", # Important math library
     "ordered-set>=3.1.1", # A useful data structure
     "pluggy>=1.2.0", # Central tool behind the ARMI Plugin system
     "pyDOE>=0.3.8", # We import a Latin-hypercube algorithm to explore a phase space
     "pyevtk>=1.2.0", # Handles binary VTK visualization files
-    "ruamel.yaml.clib ; python_version >= '3.11.0'", # C-based core of ruamel below
+    "python-dateutil>=2.2", # To read a datetime string more easily.
```
drewj-tp, line 43: lol one use in a function not used in our codebase. I can see one downstream project that uses this

https://github.com/terrapower/armi/blob/c078cde00832c559586cbcc255356bac445a7a30/armi/utils/parsing.py#L97-L99

Might be worth removing a scarcely used / trivial function that brings in an entire dependency? Not blocking for this. But maybe a follow on

drewj-tp (approved):

pyproject.toml
john-science: Oh, hey, that's a good catch.

Let me test that.  

Yeah, I'd love to rip that noise out.

john-science: That's a good catch.

This function is only used in one project anywhere, that I can see.  It is used in a terrapower internal project called "nuclerdata-uq".  So... that's our limiting factor right now.

At the very least, we could move this method and dependency over to their repo. So I can tear this out of ARMI.

drewj-tp: Either is good for me. We can punt to something later or make the change here and work with that project to update.

PR #2040, opened 2024-12-28
Unpin the max `matplotlib` version, which upgrades numpy

opotowsky:
## What is the change?

tl;dr: fixed some numpy and python deprecation issues, and deleted the max matplotlib pin for py311 and up.

Unpinning the max `matplotlib` pin in ARMI opened some floodgates for other upgrades, specifically `numpy`. I ended up with one test failing in `test_historyTracker.py` because some zeroes were being saved as `np.int64` in the database and that was getting processed into strings when using `getHistory`. So that's where those forced `int()` pieces came from. The rest is self explanatory numpy deprecation warnings. 

`matplotlib` v3.8 being the max version doesn't always work with higher python versions. The pip resolution in github actions  is not behaving the same elsewhere. So let's just fix root cause of the max pin instead. 

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database.py
+++ b/armi/bookkeeping/db/database.py
@@ -1394,8 +1394,9 @@ class Database:
                 # current time step and something has created the group to store aux data
                 continue
 
-            cycle = h5TimeNodeGroup.attrs["cycle"]
-            timeNode = h5TimeNodeGroup.attrs["timeNode"]
+            # might save as int or np.int64, so forcing int keeps things predictable
+            cycle = int(h5TimeNodeGroup.attrs["cycle"])
+            timeNode = int(h5TimeNodeGroup.attrs["timeNode"])
             layout = Layout(
                 (self.versionMajor, self.versionMinor), h5group=h5TimeNodeGroup
             )
--- a/armi/bookkeeping/db/database.py
+++ b/armi/bookkeeping/db/database.py
@@ -1097,7 +1097,7 @@ class Database:
                     )
                 )
 
-            if data.dtype.type is np.string_:
+            if data.dtype.type is np.bytes_:
                 data = np.char.decode(data)
 
             if attrs.get("specialFormatting", False):
@@ -1291,7 +1291,7 @@ class Database:
                         )
                         raise
 
-                    if data.dtype.type is np.string_:
+                    if data.dtype.type is np.bytes_:
                         data = np.char.decode(data)
 
                     if dataSet.attrs.get("specialFormatting", False):
@@ -1450,7 +1450,7 @@ class Database:
                             )
                             raise
 
-                        if data.dtype.type is np.string_:
+                        if data.dtype.type is np.bytes_:
                             data = np.char.decode(data)
 
                         if dataSet.attrs.get("specialFormatting", False):
--- a/armi/bookkeeping/db/databaseInterface.py
+++ b/armi/bookkeeping/db/databaseInterface.py
@@ -390,7 +390,8 @@ class DatabaseInterface(interfaces.Interface):
         if nowRequested:
             for param in params or history.keys():
                 if param == "location":
-                    history[param][now] = tuple(comp.spatialLocator.indices)
+                    # might save as int or np.int64, so forcing int keeps things predictable
+                    history[param][now] = tuple(int(i) for i in comp.spatialLocator.indices)
                 else:
                     history[param][now] = comp.p[param]
 
--- a/armi/bookkeeping/db/databaseInterface.py
+++ b/armi/bookkeeping/db/databaseInterface.py
@@ -391,7 +391,9 @@ class DatabaseInterface(interfaces.Interface):
             for param in params or history.keys():
                 if param == "location":
                     # might save as int or np.int64, so forcing int keeps things predictable
-                    history[param][now] = tuple(int(i) for i in comp.spatialLocator.indices)
+                    history[param][now] = tuple(
+                        int(i) for i in comp.spatialLocator.indices
+                    )
                 else:
                     history[param][now] = comp.p[param]
 
--- a/armi/bookkeeping/tests/armiRun-A0032-aHist-ref.txt
+++ b/armi/bookkeeping/tests/armiRun-A0032-aHist-ref.txt
@@ -1,5 +1,5 @@
-  (np.int64(0),np.int64(0))        (0,1)
-                0.00000E+00  0.00000E+00
+      (0,0)        (0,1)
+0.00000E+00  0.00000E+00
 
 
 key: residence
@@ -33,7 +33,7 @@ key: ztop
 
 
 key: location
-(7,-1,0)  (np.int64(0),np.int64(0),np.int64(0))
+(7,-1,0)  (0,0,0)
 
 
 EOL        bottom         top      center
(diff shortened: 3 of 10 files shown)


armi/bookkeeping/tests/armiRun-A0032-aHist-ref.txt
```diff
@@ -1,5 +1,5 @@
-      (0,0)        (0,1)
-0.00000E+00  0.00000E+00
+  (np.int64(0),np.int64(0))        (0,1)
```
opotowsky, line 1: I will fix this once I get my head wrapped around the other failures!

pyproject.toml
```diff
@@ -31,10 +31,11 @@ authors = [
...
 dependencies = [
     "coverage>=7.2.0", # Code coverage tool. Sadly baked into every Case.
-    "h5py>=3.0,<=3.9 ; python_version < '3.11.0'",
-    "h5py>=3.9 ; python_version >= '3.11.0'", # Needed because our database files are H5 format
+    "h5py>=3.0,<=3.9 ; python_version < '3.10.0'",
+    "h5py>=3.9 ; python_version >= '3.10.0'", # Needed because our database files are H5 format
     "htmltree>=0.7.6", # Our reports have HTML output
-    "matplotlib>=3.5.3,<3.8.0", # Important plotting library
+    "matplotlib>=3.5.3,<=3.8.0 ; python_version < '3.11.0'",
+    "matplotlib>=3.5.3 ; python_version >= '3.11.0'", # Important plotting library
```
opotowsky, line 38: check on removing the max pin entirely once all tests are passing.

pyproject.toml
```diff
@@ -45,7 +46,8 @@ dependencies = [
     "ruamel.yaml.clib ; python_version >= '3.11.0'", # C-based core of ruamel below
     "ruamel.yaml.clib<=0.2.7 ; python_version < '3.11.0'", # C-based core of ruamel below
     "ruamel.yaml<=0.17.21 ; python_version < '3.11.0'", # Our foundational YAML library
-    "scipy>=1.7.0", # Used for curve-fitting and matrix math
+    "scipy<1.14.0 ; python_version <= '3.10.0'",
```
opotowsky, line 49:
```suggestion
    "scipy>=1.7.0,<1.14.0 ; python_version <= '3.10.0'",
```

pyproject.toml
```diff
@@ -31,10 +31,11 @@ authors = [
 ]
 dependencies = [
     "coverage>=7.2.0", # Code coverage tool. Sadly baked into every Case.
-    "h5py>=3.0,<=3.9 ; python_version < '3.11.0'",
-    "h5py>=3.9 ; python_version >= '3.11.0'", # Needed because our database files are H5 format
+    "h5py>=3.0,<=3.9 ; python_version < '3.10.0'",
+    "h5py>=3.9 ; python_version >= '3.10.0'", # Needed because our database files are H5 format
     "htmltree>=0.7.6", # Our reports have HTML output
-    "matplotlib>=3.5.3,<3.8.0", # Important plotting library
+    "matplotlib>=3.5.3,<=3.8.0 ; python_version < '3.11.0'",
```
opotowsky, line 37:
```suggestion
    "matplotlib>=3.5.3,<3.8.0 ; python_version < '3.11.0'",
```

john-science:
Okay, I see broken tests. But I like the IDEA of the PR, and am happy to support it in whatever way you need.

opotowsky:
> Okay, I see broken tests. But I like the IDEA of the PR, and am happy to support it in whatever way you need.

Yep I'm just playing around trying to find a set of deps that works with the unpinned matplotlib for higher python versions. 

I have to find the place in the code responsible for returning the `np.int64` dudes and force an int, otherwise we are close!

opotowsky:
@john-science This is ready now! let me know if it's important enough for the release notes. I guessed "no" but I'm willing to be wrong.

armi/bookkeeping/db/database.py
```diff
@@ -1394,8 +1394,9 @@ def getHistories(
                 # current time step and something has created the group to store aux data
                 continue
 
-            cycle = h5TimeNodeGroup.attrs["cycle"]
-            timeNode = h5TimeNodeGroup.attrs["timeNode"]
+            # might save as int or np.int64, so forcing int keeps things predictable
+            cycle = int(h5TimeNodeGroup.attrs["cycle"])
+            timeNode = int(h5TimeNodeGroup.attrs["timeNode"])
```
opotowsky, line 1399: I couldn't root cause the sometimes-`np.int64` issues with the time steps, so I added a forced int here during `getHistory`

armi/bookkeeping/db/databaseInterface.py
```diff
@@ -390,7 +390,10 @@ def getHistory(
         if nowRequested:
             for param in params or history.keys():
                 if param == "location":
-                    history[param][now] = tuple(comp.spatialLocator.indices)
+                    # might save as int or np.int64, so forcing int keeps things predictable
+                    history[param][now] = tuple(
+                        int(i) for i in comp.spatialLocator.indices
+                    )
```
opotowsky, line 396: Same as the above comment on np.int64

Sometimes the `.indices` returned are int64 and those get processed as such into a string during `writeAssemHistory`. (so you end up qith `np.int64(0)` as a string rather than just `0`). 

Anyway, this is where I found it made a difference to force an int. Again, not a root cause solution.

armi/utils/mathematics.py
```diff
@@ -411,7 +411,8 @@ def goalFunc(guess, func, positiveGuesses):
         tol=tol,
         options={"maxiter": maxIterations},
     )
-    ans = float(X["x"])
+
+    ans = float(X["x"][0])
```
john-science, line 415: Sorry for being obtuse. Why is this change being made?

opotowsky: There was a deprecation warning against treating a list like a float (which I can't believe was even allowed)

These functions return `[float]` instead of `float`, so we have to grab the first/only element in that list

opotowsky: I could add a comment to that effect

opotowsky, line 415:
```suggestion
    # X returns `[num]` instead of `num`, so we have to grab the first/only element in that list
    ans = float(X["x"][0])
```

armi/utils/mathematics.py
```diff
@@ -515,7 +516,7 @@ def parabolaFromPoints(p1, p2, p3):
         print("Error in parabola {} {}".format(A, b))
         raise
 
-    return float(x[0]), float(x[1]), float(x[2])
+    return float(x[0][0]), float(x[1][0]), float(x[2][0])
```
opotowsky, line 519:
```suggestion
    # x returns `[num]` instead of `num`, so we have to grab the first/only element in that list
    return float(x[0][0]), float(x[1][0]), float(x[2][0])
```

armi/utils/mathematics.py
opotowsky: also not all of the code updates are obvious to an error or warning. I was just playing whackamole with test warnings

john-science: Okay, I see:

```python
>>> import numpy as np
>>>
>>> float(np.ones(1))
<stdin>:1: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)

1.0
```

john-science (approved):
Great! Thanks so much!

PR #2045, opened 2025-01-04
Adding a check on the grid/component consistency in the BPs

### LINKED ISSUE #1400: `latticeIDs` and associated grids can be inconsistent
Though less often used, ARMI has the ability to define the radial location of pins within a block. An example of this is here: https://terrapower.github.io/armi/tutorials/walkthrough_lwr_inputs.html

It is possible for the grid defined for a block to have no overlap with the pins defined in that block, and ARMI will not complain. In such a case, the constructed model is very likely wrong. I'll explain...

Below is a working example based on a simplified version of the inputs from the link above.

```yaml
custom isotopics:
    UO2:
        input format: number densities
        U235: 8.65e-4
        U238: 2.225E-2
        O: 4.622E-2
    moderator:
        input format: number densities
        H: 6.70e-2
        O: 3.35E-2
        B: 2.87E-5
    Zr clad:
        input format: number densities
        ZR: 4.30E-2
blocks:
    uo2: &block_uo2
        grid name: UO2 grid
        fuel:
            shape: Circle
            material: UO2
            isotopics: UO2
            Tinput: 25.0
            Thot: 25.0
            od: .8190
            latticeIDs: [U]
        gap 1: &fuel_gap_1
            shape: Circle
            material: Void
            Tinput: 25.0
            Thot: 25.0
            id: fuel.od
            od: zirconium clad.id
            latticeIDs: [U]
        zirconium clad: &clad_Zr
            shape: Circle
            material: Custom
            isotopics: Zr clad
            Tinput: 25.0
            Thot: 25.0
            id: .8360
            od: .9500
            latticeIDs: [U]
        moderator: &moderator
            shape: DerivedShape
            material: SaturatedWater
            isotopics: moderator
            Tinput: 450.0
            Thot: 450.0
        pitch: &pitch
        # dummy component for assembly sizing
            shape: Square
            material: Void
            Tinput: 25.0
            Thot: 25.0
            widthInner: 21.42
            widthOuter: 21.42
            mult: 1.0
            latticeIDs: [FC] 
assemblies:
    heights: &heights
        - 64.26
    axial mesh points: &mesh
        - 3

    UO2:
        flags: fuel
        specifier: UO2
        blocks:
            - *block_uo2
        height: *heights
        axial mesh points: *mesh
        xs types: [A]
systems:
    core:
      grid name: core

      origin:
          x: 0.0
          y: 0.0
          z: 0.0
grids:
    core:
        symmetry: quarter reflective
        geom: cartesian
        lattice pitch:
            x: 21.42
            y: 21.42
        lattice map: |
         UO2
    UO2 grid:
        symmetry: full
        geom: cartesian
        lattice pitch:
            x: 1.26
            y: 1.26
        lattice map: |
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
            U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U  U
nuclide flags:
    H: {burn: false, xs: true}
    O: 
        burn: false
        xs: true
        expandTo: ["O16", "O17"] # O18 is not in many nuclear data sets.
    B: {burn: false, xs: true}
    ZR: {burn: false, xs: true}
    U235: {burn: false, xs: true}
    U238: {burn: false, xs: true}
```

This runs with just the standalone framework.

You can see that there is a lattice map called "UO2 grid" which defines the positions of pins in the block. If you pull up the resulting database and query the `fuel` component for how many of them there are, the amount is consistent with the grid definition:

```python
>> import armi; armi.configure()
>> from armi.bookkeeping.db import databaseFactory
>> db = databaseFactory("c5g7-settings.h5", "r")
>> with db:
>>      r = db.load(0,0)

>> r.core[0][0][0]
<Circle: fuel>

>> r.core[0][0][0].spatialLocator
<MultiIndexLocation with 289 locations>

>>  r.core[0][0][0].p.mult
289
```

Okay, now take the same input and replace the grid with this:
```yaml
    UO2 grid:
        symmetry: full
        geom: cartesian
        lattice pitch:
            x: 1.26
            y: 1.26
        lattice map: |
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
            X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X  X
```

Note that this new grid does not have any locations with a corresponding pin definition from the block.
Okay, now run it and again examine the database:
```python
>> import armi; armi.configure()
>> from armi.bookkeeping.db import databaseFactory
>> db = databaseFactory("c5g7-settings.h5", "r")
>> with db:
>>      r = db.load(0,0)

>> r.core[0][0][0]
<Circle: fuel>

>> r.core[0][0][0].spatialLocator
<IndexLocation @ (0,0,0)>

>>  r.core[0][0][0].p.mult
1.0
```

There is now just 1 fuel component in the block. This is because that is the default `mult` value, and since we did not specify a `mult` on that component and it also isn't found in the grid, we get just 1 fuel pin which is floating (radially) in space. That is logically fine, but the problem is that ARMI apparently doesn't care at all that this grid makes no sense at all for this block. The grid says every position should be filled by a pin named "X", but that pin does not exist!

There _is_ a check that if the user enters a non-unity `mult` for a component, it must show up the same amount of times in the associated grid. So that is good, but the user often doesn't enter the `mult` attribute if they also specify a grid (as in the case of the example in the docs).

I see that there are some difficulties with trying to fully verify the user input of a grid, based on the way that the grid system functions.  But we should at least be catching the issue that I've described in this ticket. Otherwise a simple mistaken `latticeID` entered in the grid would be the difference between only 1 of those components versus whatever amount is intended by the user.

To do this, we should ensure that all IDs entered in a grid appear at least once in the associated block blueprint.

keckler:
## What is the change?

Close #1400 .

This PR adds a check on the grids associated with a block to ensure that the components that are supposed to be in the grid (based on them having the `latticeIDs` attribute) actually are, and in the correct number. 

Specifically, this PR enforces the following conditions:
1. If a block has a grid, all latticeID values defined over all the components in that block must be present in the grid
2. If a block has a grid, all the IDs present in that grid must appear at least once in the latticeIDs of at least one component in the block

If the conditions above are not met, it is very likely that the user has a bug in their blueprints that would otherwise be silently skipped.

#1400 pointed out that there can be inconsistency here by pretty simple user errors. This has come up in my own work, and having ARMI be more robust to this kind of error is desirable.

## SCR Information

Change Type: features

One-Sentence Description: Adds a check on the grids associated with a block to ensure that the components that are supposed to be in the grid actually are, and in the correct number.

One-line Impact on Requirements: This change very slightly touches the unit test T_ARMI_BP_GRID0 for the requirement R_ARMI_BP_GRID.

---

## Checklist


armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -184,6 +184,26 @@ def construct(
...
+                        totalInstancesOfComponentInGrid += nInstances
+
+                    expectedNumberOfInstances = c.getDimension("mult")
+                    if totalInstancesOfComponentInGrid != expectedNumberOfInstances:
+                        raise ValueError(
+                            f"Component {c} in block blueprint '{self.name}' is expected to be present in the "
+                            f"associated block grid {expectedNumberOfInstances} times, but it "
+                            f"is found {totalInstancesOfComponentInGrid} times. "
+                            "Check that the component's latticeIDs align with the block's grid."
+                        )
```
john-science, line 205:
```suggestion
                    numInstances = 0
                    for latticeID in componentDesign.latticeIDs:
                        nInstances = len([i for i in idsInGrid if i == str(latticeID)])
                        numInstances += nInstances

                    mult = c.getDimension("mult")
                    if numInstances != mult:
                        raise ValueError(
                            f"Component {c} in block blueprint '{self.name}' is expected to be present in the "
                            f"associated block grid {mult} times, but it is found {numInstances} times. "
                            "Check that the component's latticeIDs align with the block's grid."
                        )
```

john-science: I won't die on this hill. Just turn it down if you don't like it.

Long variable names SHOULD, in theory, be more clear. But they were so long I had problems reading the code (particularly that error message).

Up to you.

armi/reactor/blueprints/tests/test_blockBlueprints.py
```diff
@@ -297,6 +311,15 @@ def test_blockLattice(self):
...
         self.assertTrue(seen)
 
+    def test_componentsNotInLattice(self):
+        """
+        Ensure that we catch cases when a component is supposed to be in the grid,
+        but is not.
+        """
+        with io.StringIO(FULL_BP_WRONG_LATTICE_MAP) as stream:
+            self.blueprints = blueprints.Blueprints.load(stream)
+            self.blueprints._prepConstruction(self.cs)
```
john-science, line 321: Sorry, are you saying this should fail, but it doesn't?

Perhaps I am confused. Maybe adding an `assert` to this test would make it more clear. That is one of our testing standards, anyway:

https://github.com/terrapower/armi/blob/cf9106d9142a8dda298e6e0c73dc0875233179bf/doc/developer/standards_and_practices.rst?plain=1#L220

keckler:
Running integration tests, will update when they are done.

armi/reactor/blueprints/tests/test_blockBlueprints.py
keckler: Please ignore this, I've updated the PR and that test has been updated to actually _test_ something.

keckler:
The internal integration tests came back clean. I think this is ready for you to look at @john-science .

john-science (approved):
I only had one comment.

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -184,6 +185,30 @@ def construct(
...
+                        allLatticeIds.add(str(latticeID))
+                        # the user has given this component latticeIDs. check that
+                        # each of the ids appears in the grid, otherwise
+                        # their blueprints are probably wrong
+                        if len([i for i in idsInGrid if i == str(latticeID)]) == 0:
+                            raise ValueError(
+                                f"latticeID {latticeID} in block blueprint '{self.name}' is expected "
+                                "to be present in the associated block grid. "
+                                "Check that the component's latticeIDs align with the block's grid."
+                            )
```
john-science, line 200: Just a thought.

This method is suddenly quite long.

Could we move this to a private helper method that just does the error handling? What do you think?

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -184,6 +185,30 @@ def construct(
...
+
+        # for every id in grid, confirm that at least one component had it
+        if gridDesign:
+            idsInGrid = list(gridDesign.gridContents.values())
+            for idInGrid in idsInGrid:
+                if str(idInGrid) not in allLatticeIds:
+                    raise ValueError(
+                        f"ID {idInGrid} in grid {gridDesign.name} is not in any components of block {self.name}. "
+                        "All IDs in the grid must appear in at least one component."
+                    )
```
john-science, line 210: Just a thought.

This method is suddenly quite long.

Could we move this to a private helper method that just does the error handling? What do you think?

PR #2051, opened 2025-01-11
Update the check for a jagged array during `_writeParams`

opotowsky:
## What is the change?

Scenario where `jagged` evaluates to `False` has a loophole (with py313 and numpy 2.2.1): single-element numpy arrays mixed into a list with non-array entries fails in `_writeParams`. (Edit: it wasn't a loophole...it's that I was previously using a very old version of numpy that allowed jagged arrays, and so the parameter that ended up as a jagged array slipped through the cracks.)

This updates the check for `jagged` to account for this scenario by editing `_getShape`.

I was getting this failure sometimes while running cases (with py313 and numpy 2.2.1):

```
    JaggedArray(temp, paramDef.name) if jagged else np.array(temp)
                                                    ~~~~~~~~^^^^^^
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (782,) + inhomogeneous part.
```

This is because the logic for determining whether something is a jagged array misses the scenario where single element numpy arrays are mixed with non array/list entries. The code in `_getShape` evaluates to `(1,)` for all these array entries AS WELL AS for `int`, `float`, or `None`, which means the `jagged` check evaluates to `False`. Then `np.array(temp)` fails with the above error.  So I have the `_getShape` helper function return `1` for the latter scenario. Then the set for a list of elements mixed with single element arrays is `{(1,), 1}`, and causes `jagged` to be `True`.

### Jagged or not jagged, that is the question

We can nix the above update and keep `jagged` as `False` by editing the temp list like so:

``` 
temp = [
    t.item() if isinstance(t, np.ndarray) and len(t) == 1 else t
    for t in temp
]
``` 

I'm just assuming we want `jagged` to be true, so went with the solution in the PR. Also, this edits the list after the fact. I don't care either way.

---

## Checklist


doc/release/0.5.rst
```diff
@@ -18,6 +18,7 @@ API Changes
 
 Bug Fixes
 ---------
+#. Fix database class bug in check for jagged arrays. (`PR#XXXX <https://github.com/terrapower/armi/pull/XXXX>`_)
```
opotowsky, line 21:
```suggestion
#. Fix check for jagged arrays during ``_writeParams``. (`PR#2051 <https://github.com/terrapower/armi/pull/2051>`_)
```

john-science:
Hmmm... it makes sense.

1. Can you add a unit test to show me/everybody what the new expected behavior is in this edge case?
2. Do we need to run downstream CI on this PR?

john-science:
@opotowsky 

I think, perhaps, the method you touch is unused.  It is here:

https://github.com/terrapower/armi/blob/85053e36541b1f32256ecf70cab2adb2982d091a/armi/bookkeeping/db/database.py#L917

But the method is called `_writeParams()`, but it is only used in one place in the method `Database.writeToDB()`:

https://github.com/terrapower/armi/blob/85053e36541b1f32256ecf70cab2adb2982d091a/armi/bookkeeping/db/database.py#L671-L681

But, if you look, the method `_writeParams()` doesn't _actually_ write parameters any more (perhaps it did once).  All it does is return a value that is not used.

Do you agree?

opotowsky:
@john-science it doesn't return anything cause all it does is write things to the DB. Towards the bottom of the method you see this:
image

john-science:
Okay, one option is to move this private function out to being a regular, private method: `Database._getArrayShape`:

https://github.com/terrapower/armi/blob/85053e36541b1f32256ecf70cab2adb2982d091a/armi/bookkeeping/db/database.py#L918-L925

And then add some unit tests.

mgjarrett (approved):

doc/release/0.5.rst
```diff
@@ -18,6 +18,7 @@ API Changes
 
 Bug Fixes
 ---------
+#. Fix check for jagged arrays during ``_writeParams``. (`PR#2051 <https://github.com/terrapower/armi/pull/2051>`_)
```
john-science, line 21:
```suggestion
#. Fixing check for jagged arrays during ``_writeParams``. (`PR#2051 <https://github.com/terrapower/armi/pull/2051>`_)
```

john-science (approved):
Nice catch!

PR #2052, opened 2025-01-13
Removing Database3 from the namespace

### LINKED ISSUE #1673: Change the name Database3 to Database
> This is a big, complex change that will have to be done in phases.

It is an unfortunate fluke of history that ARMI's database class is called `Database3`. This was meant to be a temporary name used during a transition.

But, obviously, this is a big, API-breaking change that will cause a ton of (trivial) code changes in ARMI and downstream. So, I would suggest a long, gradual transition:

* [ ] Move `database3.py` to `database.py`, but create the stub file below in a new `database3.py`.
* [ ] Slowly transition all code in ARMI and downstream to use `Database` (without the "3").
* [ ] Eventually, add a `DeprecationWarning` in `Database3.py`
* [ ] Finally, eventually, remove `Database3` from ARMI.

--------------

Suggested stub file to ease our transition:

```python
# database3.py
from armi.bookkeeping.db.database import *

class Database3(Database):
    pass
```

john-science:
## What is the change?

Removing `Database3` from the API, use `Database`.

The name `Database3` has always been confusing, and it has been a long process to remove it. This moment marks the first time that I have been able to say that no one downstream uses the API endpoint `Database3`.

close #1673

---

## Checklist


opotowsky (approved):
🧼

Finally!

PR #2058, opened 2025-01-20
Use a looser tolerance on pitch for HexGrid roughly equal check

drewj-tp:
## What is the change?
Use `math.isclose` to compare pitches of two hex grids rather than strict equality.

We don't need a super strict check to make sure the grids are roughly equal. The ideal of the method is to check that one index location in one grid would be in approximately the same spatial location in the other grid. Otherwise we can't accurately rotate the position within the grid.

This problem came up in some shuffling work where two assemblies had ever so slightly different pitches, on the order of hundreths of a percent. The strict `other.pitch == self.pitch` failed even though, for the purposes of this method, the two grids are "roughly equal"

---

## Checklist


mgjarrett:
The default tolerance for `math.isclose` is 1e-9; I think we want this to be a little looser (like 1e-3 or 1e-4) to allow for two assemblies that have slightly different pin pitches but otherwise have the same number of pins.

drewj-tp:
> The default tolerance for `math.isclose` is 1e-9; I think we want this to be a little looser (like 1e-3 or 1e-4) to allow for two assemblies that have slightly different pin pitches but otherwise have the same number of pins.

Not sure how that disappeared but I did write out a `rel_tol=1e-4` locally...

Let me re-add that and add a test accordingly

mgjarrett (approved):
Looks good, just need to add release notes

john-science (approved):

PR #2060, opened 2025-01-21
Fix scaling of edge assembly `VOLUME_INTEGRATED` parameters in third core periodic symmetry

mgjarrett:
## What is the change?

Combine volume-integrated parameters for symmetric edge assemblies before removing one of them.

There was previously a condition where the volume-integrated parameters of assemblies on the 0-degree and 120-degree symmetry line in a third-core periodic model were incorrect when the core still had both copies of the symmetric assembly. Typically we remove the assemblies on the 120-degree symmetry line using the `EdgeAssemblyChanger.removeEdgeAssemblies()`, at which point the volume-integrated params in the assemblies on the 0-degree line would become correct again.

A recent updated treatment of the symmetry factor (#2017) applies symmetry factor scaling to volume-integrated block parameters in an assembly immediately before that assembly is moved to a new location in the core (or initially placed in the core). This fixed the condition where the volume-integrated params were technically incorrect while both symmetric identicals were present. However, once the edge assemblies are removed, the volume-integrated params would become incorrect again.

This PR adds a step to scale the volume-integrated params on the 0-degree symmetry line immediately before removing the assemblies along the 120-degree symmetry line, so that the volume-integrated params are always correctly scaled to account for symmetry.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -640,7 +640,7 @@ class GlobalFluxExecuter(executers.DefaultExecuter):
         geomConverter = self.geomConverters.get("edgeAssems")
         if geomConverter:
             geomConverter.scaleParamsRelatedToSymmetry(
-                self.r, paramsToScaleSubset=self.options.paramsToScaleSubset
+                self.r.core, paramsToScaleSubset=self.options.paramsToScaleSubset
             )
 
             # Resets the reactor core model to the correct symmetry and removes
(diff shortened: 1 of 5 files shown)


mgjarrett:
The unit test for the edge assembly changer did not need to be updated because the scaling was already applied in the unit test: 

https://github.com/terrapower/armi/blob/3431c44859a88ead124dfd323781c70c39bae714/armi/reactor/converters/tests/test_geometryConverters.py#L320

It just wasn't being applied when the core was constructed from blueprints.

john-science (approved):
Okay, it's a very lightweight change in terms of number of lines of code. I like it.

And it will only very slightly increase the time to load a Reactor data model from blueprints.

Not a LOT of extra testing, but I see what you mean about the existing unit test.

Thanks!

PR #2067, opened 2025-01-30
Correctly calculate wetted perimeter for depletable components

mgjarrett:
## What is the change?

Update the method `Block.getWettedPerimeter()` to correctly find components whether or not they have `Flags.DEPLETABLE`.

This is kind of an ugly implementation, but that's a result of using `Flags` to mark components as depletable. The `Flags` system probably shouldn't be used for this purpose; it would be preferable if there was a `depletable` property of components separate from the flags system (e.g., like `name` or `spatialGrid`)

Sometimes the same components that define the wetted perimeter of a block will also be marked as depletable. We want to be able to run TH calculations that need to calculate the wetted perimeter at the same time as the depletion calculations.

---

## Checklist


john-science:
@mgjarrett Forgive an honest question, but what does being depletable have to do with wetted perimeters?  It certainly sounds like a question of geometry.

For instance, I happen to know one reactor project that is using ARMI that has liquid fuel. So, in that case, the thing that is wet is itself depletable.

mgjarrett:
> @mgjarrett Forgive an honest question, but what does being depletable have to do with wetted perimeters? It certainly sounds like a question of geometry.
> 
> For instance, I happen to know one reactor project that is using ARMI that has liquid fuel. So, in that case, the thing that is wet is itself depletable.

Yes, the depletable-ness of a component has nothing to do with calculating the wetted perimeter. They are two unrelated things.

The problem here is that the code to calculate the wetted perimeter looks for the components that contribute to the wetted perimeter, like the duct, cladding, and wire. These components are sought using `block.getComponent()` with `exact=True`, so they must have _exactly_ the specified flags. Since a component can either be depletable or not depletable, we need to search both with and without `Flags.DEPLETABLE` included for each component type.

john-science:
> @mgjarrett 
> The problem here is that the code to calculate the wetted perimeter looks for the components that contribute to the wetted perimeter, like the duct, cladding, and wire. These components are sought using `block.getComponent()` with `exact=True`...

Thanks for that.

Last question. Is the problem here we have that `exact=True`? What do you think?

john-science:
@mgjarrett Bump

> Last question. Is the problem here we have that `exact=True`? What do you think?

mgjarrett:
> > @mgjarrett
> > The problem here is that the code to calculate the wetted perimeter looks for the components that contribute to the wetted perimeter, like the duct, cladding, and wire. These components are sought using `block.getComponent()` with `exact=True`...
> 
> Thanks for that.
> 
> Last question. Is the problem here we have that `exact=True`? What do you think?

Yes, that is the problem. But we need `exact=True` to distinguish between `Flags.DUCT` and `Flags.DUCT | Flags.INNER`. For the other components (`Flags.GRID_PLATE`, `Flags.INLET_NOZZLE`, `Flags.HANDLING_SOCKET`) I don't think we typically find those components mixed with other flags.

john-science (approved):

PR #2069, opened 2025-01-31
Expand allowable XS types to include lower case letters

mgjarrett:
## What is the change?

Allow lower case letters to be used as the XS type to increase the number of possible unique XS groups from 26 to 52.

Some models have a large number of different XS types (15-20), and some internal plugins generate additional XS groups while running, which can push the total number of XS groups above 26. The ARMI app will currently crash if a 27th XS type is requested.

Notes
-------

This should work fine on Linux or Mac platforms, which have case-sensitive filesystems. A user might encounter problems if they attempt to exceed 26 XS types on Windows, because the filesystem is not case-sensitive, and a plugin might try to generate a file name that is indistinguishable to a previous filename on Windows (e.g., `latticePhysics-AA.inp` and then `latticePhysics-aA.inp`). While this is unfortunate, the current code would also fail when a user tries to exceed 26 XS types (on any platform), so this PR isn't breaking a currently working capability, it's just changing the manner in which the code would fail.

---

## Checklist


john-science:
Well, the change is certainly easy to understand.

It is unfortunate this wouldn't work on Windows.  (Ridiculous file system.)

@mgjarrett Have you ran into a need for this PR in practice?

john-science (approved):

drewj-tp:
>This should work fine on Linux or Mac platforms, which have case-sensitive filesystems

I would try to find a Mac user and confirm as appropriate. I remember 2022-2023 having some case insensitive file system problems.

Looking for confirmation, it looks like it's possible for a Mac to have a case-insensitive file system

> Choose one of the following APFS formats for Mac computers using macOS 10.13 or later.
> 
> APFS: Uses the APFS format. Choose this option if you don’t need an encrypted or case-sensitive format.
> 
> APFS (Encrypted): Uses the APFS format and encrypts the volume.
> 
> APFS (Case-sensitive): Uses the APFS format and is case-sensitive to file and folder names. For example, folders named “Homework” and “HOMEWORK” are two different folders.
> 
> APFS (Case-sensitive, Encrypted): Uses the APFS format, is case-sensitive to file and folder names, and encrypts the volume. For example, folders named “Homework” and “HOMEWORK” are two different folders.


https://support.apple.com/guide/disk-utility/file-system-formats-dsku19ed921c/mac

mgjarrett:
> Well, the change is certainly easy to understand.
> 
> It is unfortunate this wouldn't work on Windows. (Ridiculous file system.)
> 
> @mgjarrett Have you ran into a need for this PR in practice?

Yes, the motivation for the PR is that I encountered this error in a practical application.

john-science:
> I would try to find a Mac user and confirm as appropriate.

@drewj-tp Is the unit test here sufficient for you? We run all the ARMI unit tests on Mac, Linux, and Windows on GitHub.

drewj-tp:
> > I would try to find a Mac user and confirm as appropriate.
> 
> @drewj-tp Is the unit test here sufficient for you? We run all the ARMI unit tests on Mac, Linux, and Windows on GitHub.

I don't have any issues with it. I just wanted to point out that not all Macs will have case-sensitive file systems. Our GitHub actions runner might so that's okay.

It almost seems like we're putting this as a requirement on the system that uses ARMI. If you don't have a case-sensitive file system, you would either have lots of fatal errors or subtle problems

john-science:
Okay, we think this PR needs a docstring explanation of the OS differences. And more importantly a logging statement based on an >26 check, if you're on Windows.

drewj-tp:
> Okay, we think this PR needs a docstring explanation of the OS differences

I think something in the user manual (non API docs) is appropriate. Like when we introduce the concept of `xs types`

doc/user/inputs.rst
```diff
@@ -788,9 +788,13 @@ xs types
   best to set blocks that have much different compositions to have separate cross section types. The
   tradeoff is that the more XS types you define, the more CPU time the case will take to run.
 
-  Representing xsType by a single letter (A-Z) or number (0-9) limits users to 36 groups. So ARMI
+  Representing xsType by a single capital letter (A-Z) or number (0-9) limits users to 36 groups. So ARMI
```
john-science, line 791:
```suggestion
  Representing xsType by a single capital letter (A-Z) or number (0-9) limits users to 36 groups. So ARMI
```

john-science: Nothing important. This would just cause a Sphinx warning.

PR #2073, opened 2025-02-03
Update number densities for the effect of `puFrac` on axial expansion

mgjarrett:
## What is the change?

Implement a new component function that adjusts the number density of the component to account for the effective change in thermal expansion resulting from change in composition (which affects material properties) at a constant temperature.

We have fuel materials where the thermal expansion coefficient is dependent on the composition of the material. As a UZr metal fuel alloy depletes, it accumulates Pu, which affects material properties. Since the `puFrac` affects the thermal expansion coefficient, a change in the `puFrac` causes a change in dimensions just like a change in temperature would, and the number densities need to be adjusted accordingly to conserve mass.

---

## Checklist

### DIFF
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -515,7 +515,7 @@ class Component(composites.Composite, metaclass=ComponentType):
         Invalidate the volume so that it will be recomputed from current dimensions upon next access.
 
         The updated value will be based on its shape and current dimensions.
-        If there is a parent container and that container contains a DerivedShape, then that must be
+        If there is a parent container and that container contains a erivedShape, then that must be
         updated as well since its volume may be changing.
 
         See Also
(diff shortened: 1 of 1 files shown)


armi/reactor/components/component.py
```diff
@@ -515,7 +515,7 @@ def clearCache(self):
         Invalidate the volume so that it will be recomputed from current dimensions upon next access.
 
         The updated value will be based on its shape and current dimensions.
-        If there is a parent container and that container contains a DerivedShape, then that must be
+        If there is a parent container and that container contains a erivedShape, then that must be
```
albeanth, line 518:
```suggestion
        If there is a parent container and that container contains a DerivedShape, then that must be
```

mgjarrett:
Closing this in favor of a better approach in #2077

PR #2075, opened 2025-02-04
Ignoring Fluid density warning for Component Fluids

drewj-tp:
## What is the change?

Calling `Component.material.density` does not log a warning with a traceback if the material is a subclass of `Fluid`

TH applications need fluid densities and the traceback invocation is kind of a bottleneck.

https://github.com/terrapower/armi/pull/1852 introduced a warning when `Component.material.density(...)` was invoked. The rationale is the `Component` is the source of truth for volume, density, mass, etc. And components are the things that are thermally expanded.

However, fluids are not thermally expanded and so the density of a fluid material is appropriate.

This change undoes the density wrapper applied by the base `Material` when creating a `Fluid` subclass.

A test is added to show that any non-`Fluid` subclass still logs the warning, but a `Fluid` subclass does not.

I attempted to use a "real" material like `Sodium` but lots of other warnings can be logged depending on appropriate bounds of temperatures. So a fake Fluid is used to have more control over what warnings are logged.

I contend that the stack trace should still be used in this message. This is a message meant for application developers to root out problematic parts of their codebase. And it helped us pinpoint very specific call locations with ease.

---

## Checklist

### DIFF
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -759,7 +759,7 @@ class Fluid(Material):
 
     def __init_subclass__(cls):
         # Undo the parent-aware density wrapping. Fluids do not expand in the same way solids, so
-        # Fluid.density(T) is correct. This does not hold for solids because they thermalaly expand.
+        # Fluid.density(T) is correct. This does not hold for solids because they thermally expand.
         if hasattr(cls.density, "__wrapped__"):
             cls.density = cls.density.__wrapped__
 
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -758,11 +758,8 @@ class Fluid(Material):
     """A material that fills its container. Could also be a gas."""
 
     def __init_subclass__(cls):
-        # Undo the parent-aware density wrapping
-        # Justification: fluids do not expand in the same way solids expand so
-        # fluid.density(T) is an appropriate representation of the density of
-        # the fluid. This does not hold for solids because of the thermal expansion
-        # mechanics imposed by the rest of the framework
+        # Undo the parent-aware density wrapping. Fluids do not expand in the same way solids, so
+        # Fluid.density(T) is correct. This does not hold for solids because they thermalaly expand.
         if hasattr(cls.density, "__wrapped__"):
             cls.density = cls.density.__wrapped__
 
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -757,6 +757,15 @@ class Material:
 class Fluid(Material):
     """A material that fills its container. Could also be a gas."""
 
+    def __init_subclass__(cls):
+        # Undo the parent-aware density wrapping
+        # Justification: fluids do not expand in the same way solids expand so
+        # fluid.density(T) is an appropriate representation of the density of
+        # the fluid. This does not hold for solids because of the thermal expansion
+        # mechanics imposed by the rest of the framework
+        if hasattr(cls.density, "__wrapped__"):
+            cls.density = cls.density.__wrapped__
+
     def getThermalExpansionDensityReduction(self, prevTempInC, newTempInC):
         """Return the factor required to update thermal expansion going from one temperature (in
         Celcius) to a new temperature.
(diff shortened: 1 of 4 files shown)


armi/materials/tests/test_fluids.py
```diff
@@ -0,0 +1,64 @@
...
+    def test_fluidDensityWrapperNoWarning(self):
+        """Test that Component.material.density does not raise a warning for fluids.
+
+        The ARMI Framework contains a mechanism to warn users if they as for the density
+        of a material when that material is attached to a component, e.g., ``comp.material.density()``.
+        The problem here is the component is the source of truth for volume and composition. It stores
+        dimensions, which can be thermally expanded according to the material, and number densities, which
+        may be updated through operation. Much of the framework operates on ``component.density`` and
+        other ``Component`` methods for mass accounting. However, ``comp.material.density`` does
+        not know about the new composition or volumes and can diverge from ``component.density``.
```
john-science, line 44:
```suggestion
        The ARMI Framework contains a mechanism to warn users if they ask for the density of a
        material attached to a component, e.g., ``comp.material.density()``. The problem is the
        component is the source of truth for volume and composition. It stores dimensions, which can
        be thermally expanded according to the material and number densities, which may be updated
        during operation. Much of the framework operates on ``Component.density`` and other
        ``Component`` methods for mass accounting. However, ``comp.material.density`` does not know
        about the new composition or volumes and can diverge from ``component.density``.
```

armi/materials/material.py
```diff
@@ -757,6 +757,15 @@ def adjustTD(self, val):
 class Fluid(Material):
     """A material that fills its container. Could also be a gas."""
 
+    def __init_subclass__(cls):
+        # Undo the parent-aware density wrapping
+        # Justification: fluids do not expand in the same way solids expand so
+        # fluid.density(T) is an appropriate representation of the density of
+        # the fluid. This does not hold for solids because of the thermal expansion
+        # mechanics imposed by the rest of the framework
```
john-science, line 765:
```suggestion
        # Undo the parent-aware density wrapping. Fluids do not expand in the same way solids, so
        # Fluid.density(T) is correct. This does not hold for solids because they thermalaly expand.
```

armi/materials/tests/test_fluids.py
```diff
@@ -0,0 +1,64 @@
...
+        material attached to a component, e.g., ``comp.material.density()``. The problem is the
+        component is the source of truth for volume and composition. It stores dimensions, which can
+        be thermally expanded according to the material and number densities, which may be updated
+        during operation. Much of the framework operates on ``Component.density`` and other
+        ``Component`` methods for mass accounting. However, ``comp.material.density`` does not know
+        about the new composition or volumes and can diverge from ``component.density``.
+
+        Additionally, the framework does not do any thermal expansion on fluids. So the above calls to
+        ``component.material.density`` are warranted if the material is a fluid.
+        """
```
john-science, line 48:
```suggestion
        """Test that Component.material.density does not raise a warning for fluids.

        The ARMI Framework contains a mechanism to warn users if they ask for the density of a
        material attached to a component. But the component is the source of truth for volume and
        composition. And can be thermally expanded during operation. Much of the framework operates
        on ``Component.density`` and other ``Component`` methods for mass accounting. However,
        ``comp.material.density`` does not know about the new composition or volumes and can diverge
        from ``component.density``.

        Additionally, the framework does not do any thermal expansion on fluids. So the above calls
        to ``component.material.density`` are warranted for fluids.
        """
```

john-science (approved):
Great!

armi/materials/material.py
```diff
@@ -757,6 +757,12 @@ def adjustTD(self, val):
 class Fluid(Material):
     """A material that fills its container. Could also be a gas."""
 
+    def __init_subclass__(cls):
+        # Undo the parent-aware density wrapping. Fluids do not expand in the same way solids, so
+        # Fluid.density(T) is correct. This does not hold for solids because they thermalaly expand.
```
drewj-tp, line 762:
```suggestion
        # Fluid.density(T) is correct. This does not hold for solids because they thermally expand.
```

PR #2077, opened 2025-02-06
Applying expansion for material composition changes

mgjarrett:
## What is the change?

When a composition is updated, check to see if this change affects the thermal expansion properties of the material. If it does, clear the cache to allow volumes to be recomputed and apply a scaling factor to the number densities to conserve mass.

Material properties can be dependent upon the composition of a material. When the composition of a material changes from depletion (or anything else), the thermal expansion properties can change, which causes a change in volume. This change looks like thermal expansion, but caused by a change in composition rather than a change in temperature. To accurately conserve mass throughout a simulation, it is beneficial to check for this type of expansion any time a composition changes and adjust the number densities accordingly.

---

## Checklist


armi/reactor/components/component.py
```diff
@@ -369,11 +369,21 @@ def applyMaterialMassFracsToNumberDensities(self):
...
         density = self.material.getProperty("pseudoDensity", Tc=self.temperatureInC)
-
         self.p.numberDensities = densityTools.getNDensFromMasses(
             density, self.material.massFrac
         )
 
+        # Sometimes the material thermal expansion depends on its parent's composition (e.g. Pu frac)
+        # so setting number densities can sometimes change thermal expansion behavior.
+        # Call again so that the material has access to its parent's comp when providing the reference
+        # initial density.
```
john-science, line 379:
```suggestion
        # Sometimes material thermal expansion depends on its parent's composition (e.g. Pu frac) so
        # setting number densities can sometimes change thermal expansion behavior. Call again so
        # the material has access to its parent's comp when providing the reference initial density.
```

armi/reactor/components/component.py
```diff
@@ -738,26 +757,62 @@ def setNumberDensities(self, numberDensities):
...
+        -----
+        Sometimes volume/dimensions can change due to the number density change when the material thermal
+        expansion depends on the component's composition (e.g. its plutonium fraction). In this case, changing the
+        density will implicitly change the area/volume. Since it is very difficult to predict the new dimensions
+        ahead of time, and perturbation/depletion calculations are almost exclusively done assuming constant volume,
+        the densities sent are automatically adjusted to conserve mass with the original dimensions. That is, the
+        component's densities are not exactly as passed, but whatever they would need to be to preserve volume
+        integrated number densities (moles) from the pre-perturbed component's volume/dimensions.
+        Note this has no effect if the material thermal expansion has no dependence on component composition.
+        If this is not desired, `self.p.numberDensities` can be set directly.
```
john-science, line 785:
```suggestion
        Sometimes volume/dimensions change due to number density change when the material thermal
        expansion depends on the component's composition (e.g. its plutonium fraction). In this
        case, changing the density will implicitly change the area/volume. Since it is difficult to
        predict the new dimensions, and perturbation/depletion calculations almost exclusively
        assume constant volume, the densities sent are automatically adjusted to conserve mass with
        the original dimensions. That is, the component's densities are not exactly as passed, but
        whatever they would need to be to preserve volume integrated number densities (moles) from
        the pre-perturbed component's volume/dimensions.

        This has no effect if the material thermal expansion has no dependence on component
        composition. If this is not desired, `self.p.numberDensities` can be set directly.
```

armi/reactor/components/component.py
```diff
@@ -738,26 +757,62 @@ def setNumberDensities(self, numberDensities):
...
-        # since we're updating the object the param points to but not the param itself, we have to inform
+
+        # check if thermal expansion changed
+        dLLnew = self.material.linearExpansionPercent(Tc=self.temperatureInC) / 100.0
+        if dLLprev != dLLnew and materialExpansion:
+            # the thermal expansion changed so the volume change is happening at same time as
+            # density change was requested. Attempt to make mass consistent with old dims
+            # (since the density change was for the old volume and otherwise mass wouldn't be conserved)
+
+            # enable recalculation of volume, otherwise it uses stored.
```
john-science, line 810:
```suggestion
            # the thermal expansion changed so the volume change is happening at same time as
            # density change was requested. Attempt to make mass consistent with old dims (since the
            # density change was for the old volume and otherwise mass wouldn't be conserved).

            # enable recalculation of volume, otherwise it uses stored
```

john-science:
@mgjarrett Just curious why this is still draft.

armi/reactor/components/component.py
```diff
@@ -651,6 +660,20 @@ def setTemperature(self, temperatureInC):
         self.changeNDensByFactor(f)
         self.clearLinkedCache()
 
+    def changeNDensByFactor(self, factor):
```
john-science, line 663: Can we get a unit test to cover this new method and its edge cases?

armi/reactor/components/component.py
```diff
@@ -734,22 +752,71 @@ def setNumberDensities(self, numberDensities):
...
+            area = self.getArea()
+
+        # change the densities
+        if wipe:
+            self.p.numberDensities = {}  # clear things not passed
         self.p.numberDensities.update(numberDensities)
-        # since we're updating the object the param points to but not the param itself, we have to inform
-        # the param system to flag it as modified so it properly syncs during ``syncMpiState``.
-        self.p.assigned = parameters.SINCE_ANYTHING
-        self.p.paramDefs["numberDensities"].assigned = parameters.SINCE_ANYTHING
```
john-science, line 752: We can just remove this without hurting parallel runs? That's surprising.

mgjarrett: Nope, this caused a bug. reverting now...

john-science: Ah, okay, that makes more sense.

Thanks!

armi/reactor/components/component.py
```diff
@@ -734,23 +738,93 @@ def setNumberDensities(self, numberDensities):
...
+            else:
+                factor = area / self.getArea()
+            self.changeNDensByFactor(factor)
+
         # since we're updating the object the param points to but not the param itself, we have to inform
         # the param system to flag it as modified so it properly syncs during ``syncMpiState``.
         self.p.assigned = parameters.SINCE_ANYTHING
         self.p.paramDefs["numberDensities"].assigned = parameters.SINCE_ANYTHING
 
+    def changeNDensByFactor(self, factor):
```
john-science, line 812: Can we get a unit test for the new methods we are adding?

Thanks!

drewj-tp: added

armi/reactor/components/component.py
```diff
@@ -747,23 +751,93 @@ def setNumberDensities(self, numberDensities):
...
         self.p.numberDensities.update(numberDensities)
+
+        # check if thermal expansion changed
+        dLLnew = self.material.linearExpansionPercent(Tc=self.temperatureInC) / 100.0
+        if dLLprev != dLLnew and materialExpansion:
+            # the thermal expansion changed so the volume change is happening at same time as
+            # density change was requested. Attempt to make mass consistent with old dims (since the
+            # density change was for the old volume and otherwise mass wouldn't be conserved).
+
+            self.clearLinkedCache()  # enable recalculation of volume, otherwise it uses cached
```
john-science, line 813: I am curious how this effects our performance. It is not immediately obvious to me what will happen. 

Hmmm

armi/reactor/tests/test_components.py
```diff
@@ -49,6 +50,27 @@
     materials,
 )
 from armi.testing import loadTestReactor
+from armi.utils.units import getTc
+
+
+class CompositionDependentExpander(materials.Material):
```
john-science, line 56: I apologize for myself.

But can we rename this something like `MockCompositionDependentExpander`?

I know that seems like a pedantic ask. But, no joke, we have had people import things from out tests classes for non-test uses in the past. And... #sigh. That if words like "mock" or "test" or "dummy" are included, I feel like we've done all we can.

Sorry.

john-science (approved):
I would like that test class rename. But I guess if you all are in a dead rush, I can open up a follow up PR about it.

PR #2084, opened 2025-02-26
Solving test warnings

### LINKED ISSUE #2041: Warnings in tests
I have two types of warnings I see here: 

(1)

image

and (2)

image

I don't want to fix (1) without some advice because that's an issue with the code itself (in multiple locations, but it's the same SORT of problem and will likely be handled in the same way in all locations). 

And (2), I don't know HOW to fix. 

Help?

john-science:
## What is the change?

Here I have made a bunch of random changes to unit tests necessary to remove the warnings we get during a run of our unit tests.

I should note that this solves ALL the warnings, except those from `jupyter` notebooks, because after some research I think we just have to wait for a new version of jupyter notebooks to be released to solve those. Sorry.

To close #2041

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/tests/test_comparedb3.py
+++ b/armi/bookkeeping/db/tests/test_comparedb3.py
@@ -14,6 +14,7 @@
 
 """Tests for the compareDB3 module."""
 import unittest
+import warnings
 
 import h5py
 import numpy as np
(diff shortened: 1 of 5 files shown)


opotowsky (approved):
No comments!

Big happy!

Thanks so much for this!

PR #2085, opened 2025-02-26
Dropping black formatter for ruff

john-science:
## What is the change?

Here I dropped the `black` linter from the project entirely, and switched to using `ruff`. This meant I had to:

* Remove `black` from `pyproject.toml` and the CI
* Update `ruff` to `0.9.7`.
* Changed the `ruff` `target-version` from Python 3.9 to Python 3.13
* Format ARMI with `ruff` (notice this make a TON of small changes for line length).
* Added `ruff format` to CI.
* Updated the docs to remove `black` and add `ruff format`.

1. `ruff format` is ~30 times faster than `black`
2. This will make reduce the number of dependencies we have.

---

## Checklist


keckler:
Question -- Is this going to cause issues for people that currently have the `black` formatter automatically triggered by their IDE? Or are the two formatters supposedly totally compatible in terms of their formatting rules?

ntouran:
> Question -- Is this going to cause issues for people that currently have the black formatter automatically triggered by their IDE? 

yes! they will fight to a degree.

pyproject.toml
```diff
@@ -127,12 +126,12 @@ find = {}
...
-required-version = "0.5.1"
+required-version = "0.9.7"
 
 # Assume Python 3.9
-target-version = "py39"
+target-version = "py313"
 
-# Setting line-length to 120 (though blacks default is 88)
+# Setting line-length to 120
 line-length = 120
```
ntouran, line 135: you sure you don't want to take this opportunity to go to the ruff and black default of 88? 88 is great! https://docs.astral.sh/ruff/settings/#line-length https://stackoverflow.com/a/578318/1279313

otherwise this looks great.

john-science: I tried. Even going down to 100 meant I had over 400 files that required manual changes. (Things like docstrings couldn't be automatically reformatted.)

And I do not have time for manually touching >1,000 lines to make them shorter.

(Also, as a personal preference, I might go 100, but 88 feels a bit too narrow when you have a lot of math on the same line.)

opotowsky:
> Question -- Is this going to cause issues for people that currently have the `black` formatter automatically triggered by their IDE? Or are the two formatters supposedly totally compatible in terms of their formatting rules?

Since ppl are fond of VSCode and this will cause problems for folks while we convert to this, John has graciously agreed to delay this PR. So I marked it as draft. There will be an adjustment period, but we will plan to get everything downstream converted. Only folks who work on ARMI will have to battle the conflicting formatters. `ruff` on the command line will fix that easily though.

doc/release/0.5.rst
```diff
@@ -32,6 +32,7 @@ Bug Fixes
 
 Quality Work
 ------------
+#. Dropping ``black`` formatter for ``ruff format``. (`PR#2085 <https://github.com/terrapower/armi/pull/2085>`_)
```
john-science, line 35:
```suggestion
```

john-science:
@opotowsky  @keckler  I could reproduce this PR in like 1 minute. The real question is: when do we roll out this change to people's workflows?

I expect we need to broach the topic with people, explain, and provide a timeline. But we'll see what y'all think.

keckler:
I am not anti this change. However, I do not see a particularly strong driver for it right now. I also think there is some additional work that is maybe not being considered.

Beyond the PRs that need to be made in various repos to (1) remove the requirement (easy), and (2) actually run `ruff` on the projects (also easy), I think there is possibly some other work that will need to be done.
1. I know that some SQAPs currently call for `black` to be used.
2. If `black` and `ruff` clash formatting, as Nick thinks they might, then this will require people to uninstall their `black` formatter from their IDEs. 

Speaking personally, I like having `black` integrated into my IDE. It seems like there is an integration for `ruff` for my IDE, so I can drop that in instead, but again, that is some more effort for a number of people.

All that being said, I am open to doing this, but it honestly seems like quite a bit of work without much benefit.

As to the question of "when", I am not picky.

opotowsky:
OK this can be reopened and move forward @john-science ....we have the internal concerns handled.

Let me know when you get the merge conflicts handled. We may need some downstream CI adjustments but they are underway already anyway.

.github/workflows/linting.yaml
```diff
@@ -11,11 +11,11 @@ jobs:
...
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Setup Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
-          python-version: '3.9'
+          python-version: '3.13'
```
opotowsky, line 18: Don't you need a `ruff format --check` line added to this in addition to `ruff check .`?

armi/tests/tutorials/param_sweep.ipynb
opotowsky, line 1: The docstring has a small grammatical error here, should be "Modifier that adjust**s**"

.github/workflows/linting.yaml
john-science: Updated!

armi/tests/tutorials/param_sweep.ipynb
john-science: Fixed

Issue #2006, opened 2024-11-05
Should component volume consider symmetryFactor? 


albeanth:
In #1954, we are starting to store more information at the component level. One of those things is the decay power (`powerDecay`). While looking at some downstream test results with these changes, we noticed that the center-most assembly block-wise decay power was not properly accounting for the symmetry factor. After some discussion, we found that it was due to the fact that we were calculating the block-wise decay power from component-level decay power; and the latter was getting calculated without consideration to the symmetry factor. 

E.g., we typically calculate block-wise power by doing something like
```
b.decayPower = b.decayPowerDensity * b.getVolume()
```
where `b.getVolume()` accounts for the symmetry factor. However, a block-wise decay power derived from component-level decay power looks something like
```
for c in b:
    c.decayPower = c.decayPowerDensity * c.getVolume()

b.decayPower = sum(c.decayPower for c in b)
```
where `c.getVolume()` does not respect the symmetry factor. 

So all this said, should c.getVolume() respect the symmetry factor? This could look something like:
![image](https://github.com/user-attachments/assets/e6c030c6-327e-4d55-a2db-459b936365c5)

john-science:
@albeanth "decayPower" is not a parameter in ARMI. That sounds like some MCNP value.

> should `c.getVolume()` respect the symmetry factor?

For your situation, it seems like a clear "yes".  But `Component.computeVolume()` is used in just SO MANY place, so.... can I say for sure, right now, that making the suggested change is correct everywhere? No, I can't say that.

But:

1) We can just test your idea, and see if a million things break.
2) EVEN IF a bunch of other things break, we could add a parameter to that method:

```python
def computeVolume(self, respectSymmetry=False):
```

So, it seems fixable either way.

jakehader:
> > "decayPower" is not a parameter in ARMI.
> 
> It was until #1954. Either way, that's not the important part here.
> 
> I agree, component.getVolume() is used an a gazillion places - and is why I wanted to open this Issue first and see what others thoughts were. Just changing it will likely have _quite_ the ripple effect.
> 
> We could do the optional param, that would be easy. But if it is indeed a bug, then maybe the optional param isn't the "right" option and really it should just be hardcoded in. That's the part I don't know and wanted to get the opinion of others on.

[...]

PR #2086, opened 2025-02-26
Account for symmetry factor in `getMgFlux`

mgjarrett:
## What is the change?

Implement a `getMgFlux` method for `Component` instead of inheriting from the `Composite` class.

The component volume does not account for the parent symmetry factor (see #2006), so we need a special `getMgFlux` that incorporates the symmetry factor here.

This should have been included in #2068 , which addressed #2059 , but this was missed. `getMgFlux` uses `getIntegratedMgFlux`, which was updated to incorporate the symmetry factor. `getMgFlux` needs to also incoporate the `symmetryFactor` to stay consistent.

---

## Checklist

### DIFF
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -1221,6 +1221,46 @@ class Component(composites.Composite, metaclass=ComponentType):
             )
         self.setMassFracs(adjustedMassFracs)
 
+    def getMgFlux(self, adjoint=False, average=False, volume=None, gamma=False):
+        """
+        Return the multigroup neutron flux in [n/cm^2/s].
+
+        The first entry is the first energy group (fastest neutrons). Each additional
+        group is the next energy group, as set in the ISOTXS library.
+
+        On blocks, it is stored integrated over volume on <block>.p.mgFlux
+
+        Parameters
+        ----------
+        adjoint : bool, optional
+            Return adjoint flux instead of real
+
+        average : bool, optional
+            If true, will return average flux between latest and previous. Doesn't work
+            for pin detailed yet
+
+        volume: float, optional
+            If average=True, the volume-integrated flux is divided by volume before
+            being returned. The user may specify a volume here, or the function will
+            obtain the block volume directly.
+
+        gamma : bool, optional
+            Whether to return the neutron flux or the gamma flux.
+
+        Returns
+        -------
+        flux : np.ndarray
+            multigroup neutron flux in [n/cm^2/s]
+        """
+        if average:
+            raise NotImplementedError(
+                "{} class has no method for producing average MG flux -- try"
+                "using blocks".format(self.__class__)
+            )
+
+        volume = volume or self.getVolume() / self.parent.getSymmetryFactor()
+        return self.getIntegratedMgFlux(adjoint=adjoint, gamma=gamma) / volume
+
     def getIntegratedMgFlux(self, adjoint=False, gamma=False):
         """
         Return the multigroup neutron tracklength in [n-cm/s].
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -1228,8 +1228,6 @@ class Component(composites.Composite, metaclass=ComponentType):
         The first entry is the first energy group (fastest neutrons). Each additional
         group is the next energy group, as set in the ISOTXS library.
 
-        On blocks, it is stored integrated over volume on <block>.p.mgFlux
-
         Parameters
         ----------
         adjoint : bool, optional
@@ -1240,9 +1238,9 @@ class Component(composites.Composite, metaclass=ComponentType):
             for pin detailed yet
 
         volume: float, optional
-            If average=True, the volume-integrated flux is divided by volume before
-            being returned. The user may specify a volume here, or the function will
-            obtain the block volume directly.
+            The volume-integrated flux is divided by volume before
+            being returned. The user may specify a volume here, or the function
+            will obtain the block volume directly.
 
         gamma : bool, optional
             Whether to return the neutron flux or the gamma flux.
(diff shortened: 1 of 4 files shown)


armi/reactor/components/component.py
```diff
@@ -1221,6 +1221,44 @@ def adjustMassEnrichment(self, massFraction):
...
+            multigroup neutron flux in [n/cm^2/s]
+        """
+        if average:
+            raise NotImplementedError(
+                "Component has no method for producing average MG flux -- try"
+                "using blocks"
+            )
+
+        volume = volume or self.getVolume() / self.parent.getSymmetryFactor()
+        return self.getIntegratedMgFlux(adjoint=adjoint, gamma=gamma) / volume
```
drewj-tp, line 1260: This is at least consistent with #2086 in that if `Component.getIntegratedMgFlux` accounts for the symmetry factor, we should account for it in `Component.getMgFlux` as we divide by the volume of integration (full or 1/3 symmetry)

And things like `Component.getMass` use `Component.getVolume() * Component.parent.getSymmetryFactor()` so maybe the guidance is 

1. Always use symmetry factor when using component volumes, or
2. `Component.getVolume` should account for its symmetry factor - #2006

armi/reactor/tests/test_blocks.py
```diff
@@ -1268,10 +1268,16 @@ def test_getMgFlux(self, mock_sf):
...
+        neutronFluxInt = fuel.getIntegratedMgFlux()
+        gammaFluxInt = fuel.getIntegratedMgFlux(gamma=True)
+        np.testing.assert_almost_equal(neutronFluxInt, np.ones(5) * volFrac)
+        np.testing.assert_almost_equal(gammaFluxInt, np.ones(4) * volFrac * 2.0)
+
+        neutronFlux = fuel.getMgFlux()
+        gammaFlux = fuel.getMgFlux(gamma=True)
+        normalization = volFrac / fuelVol * self.block.getSymmetryFactor()
+        np.testing.assert_almost_equal(neutronFlux, np.ones(5) * normalization)
+        np.testing.assert_almost_equal(gammaFlux, np.ones(4) * normalization * 2.0)
```
drewj-tp, line 1280: The work going on here is correct, but it's definitely not easy to follow, what with all the symmetry factors and divisions and on-the-fly arrays that don't easily tie back to their source.

Soft non-blocking recommendations to

1. Add comments why we need to divide or multiply out the symmetry factor
2. Store block neutron and gamma flux as variables and use in latter assertions, e.g., `np.testing.assert_almost_equal(neutronFlux, blockNeutronFlux * normalization)`
   - The `* 2.0` in the `gammaFlux` comparison threw me for a loop trying to figure out where the additional doubling happened.

armi/reactor/tests/test_blocks.py
```diff
@@ -1281,10 +1287,16 @@ def test_getMgFlux(self, mock_sf):
...
```
drewj-tp, line 1299: Same readability comments

armi/reactor/components/component.py
```diff
@@ -1221,6 +1221,44 @@ def adjustMassEnrichment(self, massFraction):
             )
         self.setMassFracs(adjustedMassFracs)
 
+    def getMgFlux(self, adjoint=False, average=False, volume=None, gamma=False):
```
drewj-tp, line 1224: FWIW I cannot find a single instance of `o.getMgFlux` with the `volume` being provided to the invokation. I was going to comment about needing to test this case, where `volume is not None` but couldn't find any usage of it. Maybe something for us to consider removing

No actions needed. I'll file a follow up ticket.

drewj-tp (commented):
PR still in draft but the changes make sense compared to #2086. `Component.getVolume` seems to be a footgun

john-science:
Hmm...  I kind of hate that we have `Composite.getMgFlux()` and `Block.getMgFlux()`. That we have to add `Component.getMgFlux()` just compounds it.

Is there no more generic way to solve this problem that adding a `getMyFavoriteParameter()` method for every parameter in the world to every class in the data object model?

I really feel like there should be.

armi/reactor/components/component.py
drewj-tp: https://github.com/terrapower/armi/issues/2087

armi/reactor/components/component.py
drewj-tp: > But it's a relatively big undertaking that we're not ready for just yet. It would ideally come soon

For sure. Wasn't trying to say we should do it now. More trying to understand how people should interact w/ `Component.getVolume`

mgjarrett:
> Hmm... I kind of hate that we have `Composite.getMgFlux()` and `Block.getMgFlux()`. That we have to add `Component.getMgFlux()` just compounds it.
> 
> Is there no more generic way to solve this problem that adding a `getMyFavoriteParameter()` method for every parameter in the world to every class in the data object model?
> 
> I really feel like there should be.

If we resolve #2006 then we would be able to let `Component` inherit `getMgFlux` from `Composite`

drewj-tp (approved):

armi/reactor/components/component.py
```diff
@@ -1221,6 +1221,44 @@ def adjustMassEnrichment(self, massFraction):
...
+        average : bool, optional
+            If true, will return average flux between latest and previous. Doesn't work
+            for pin detailed yet
+
+        volume: float, optional
+            The volume-integrated flux is divided by volume before
+            being returned. The user may specify a volume here, or the function
+            will obtain the block volume directly.
+
+        gamma : bool, optional
```
john-science, line 1245:
```suggestion
        average : bool, optional
            If True, will return average flux between latest and previous. Doesn't work
            for pin detailed.
        volume: float, optional
            The volume-integrated flux is divided by volume before
            being returned. The user may specify a volume here, or the function
            will obtain the block volume directly.
        gamma : bool, optional
```

john-science (approved):

PR #2093, opened 2025-03-19
Speed up axial expansion unit tests

john-science:

This is a very minor change to the axial expansion tests where I make them run a bit faster. Essentially, I reduce the size of the hexagonal core in the input files, and that speeds up the run time of the tests. As it stands, I just removed two rings of assemblies, and one of those rings was only shielding assemblies, that were not relevant to the test.

This cuts the run time for the axial expansion unit tests in half.

NOTE: I also rounded the test data a little bit, because I do not want people thinking the extremely specific and exact numbers in the input test file represent any real world reactors.

## SCR Information

Change Type: trivial

One-Sentence Description: Speeding up unit tests related to axial expansion.

One-line Impact on Requirements: Very minor change to the tests associated with T_ARMI_ASSEM_HEIGHT_PRES and T_ARMI_INP_COLD_HEIGHT, in that I have sped them up.

---

## Checklist


armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -343,22 +342,22 @@ def complexConservationTest(self, a):
             newFE56Mass = a.getMass("FE56")
             newMass = self._getMass(a)
             self.assertAlmostEqual(
-                newFE56Mass / prevFE56Mass, 1.0, places=14, msg=f"{a}"
+                newFE56Mass / prevFE56Mass, 1.0, places=8, msg=f"{a}"
```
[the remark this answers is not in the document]
john-science: fixed

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -351,14 +350,14 @@ def complexConservationTest(self, a):
         newMasses, newNDens = self._getComponentMassAndNDens(a)
         # make sure that the assembly returned to the original state
         for orig, new in zip(origMesh, a.getAxialMesh()):
-            self.assertAlmostEqual(orig, new, places=12, msg=f"{a}")
+            self.assertAlmostEqual(orig, new, places=8, msg=f"{a}")
```
albeanth, line 353: Recovering block heights is a key part of recovering mass. I would prefer to keep this check tight as well.

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -669,7 +667,9 @@ def _checkMass(self, origDetailedNDens, origVolume, newDetailedNDens, newVolume,
                 )
             else:
                 # should not conserve mass here as it is structural material above active fuel
-                self.assertAlmostEqual(newMass / prevMass, 0.99, msg=f"{c}, {c.parent}")
+                self.assertAlmostEqual(
+                    newMass / prevMass, 0.99999999, msg=f"{c}, {c.parent}"
```
albeanth, line 671: Curious why the precision increase here? I don't think it hurts anything, but am just curious.

albeanth (approved):
May I suggest polling ARMI users what they _need_ and/or _want_ from the reactor set up in `armi/tests`? That set up has been around for quite a while I think and may not reflect current needs. And some of the needs that are reflected could (should?) be tested in downstream applications.

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -669,7 +667,9 @@ def _checkMass(self, origDetailedNDens, origVolume, newDetailedNDens, newVolume,
...
+                )
```
john-science, line 672:
```suggestion
                self.assertAlmostEqual(newMass / prevMass, 0.99, msg=f"{c}, {c.parent}")
```

armi/reactor/converters/tests/test_axialExpansionChanger.py
john-science: Reverted

john-science:
> May I suggest polling ARMI users what they _need_ and/or _want_ from the reactor set up in `armi/tests`? That set up has been around for quite a while I think and may not reflect current needs. And some of the needs that are reflected could (should?) be tested in downstream applications.

So, my take is that ARMI can provide more test reactors. For instance, ARMI already offers a "one assembly, one block hex reactor":  https://github.com/terrapower/armi/tree/main/armi/tests/smallestTestReactor

The reason I pose the problem in that way is that... sigh... for better or worse, a LOT of people are using the old/original ARMI test reactor. So, even if people wanted something else, we can't just gut the old test reactor without making changes to hundreds of downstream tests in dozens of repos. Which, yes, is highly doable. But I would expect a lot of "we're too busy for that" and "not a priority" if we tried.

I could be wrong though. Perhaps that is a pessimistic view. I'm happy to hear people out.

albeanth:
> The reason I pose the problem in that way is that... sigh... for better or worse, a LOT of people are using the old/original ARMI test reactor. So, even if people wanted something else, we can't just gut the old test reactor without making changes to hundreds of downstream tests in dozens of repos. Which, yes, is highly doable. But I would expect a lot of "we're too busy for that" and "not a priority" if we tried.
> 
> I could be wrong though. Perhaps that is a pessimistic view. I'm happy to hear people out.

Worse comes to worse, people call pull the old test reactor into their own applications to keep their downstream tests performing the same.

PR #2102, opened 2025-03-25
Fixing issue in copyInterfaceInputs with one file

### LINKED ISSUE #2097: Weird behavior with case clone, copied inputs, and setting expecting a list of files
I saw some weird behavior recently that I'm going to try to explain as best I could.

We have an input file that has a setting that expects list of files. When we have armi clone the case, it
1. removed the list behavior on the setting, and
2. mangled the setting contents when writing the new settings file.

Initially, we had something like

```yaml
specialFiles:
  - first_file.txt
```

And following the clone, we got
```yaml
specialFiles:
  - f
  - i
  - r
  - s
  - t
  - _
  - f
  - i
  - l
  - e
  - .
  - t
  - x
  - t
```

After some inspection, I found this line in `copyInterfaceInputs`, the function responsible for moving ancillary files found in the input to the location of the cloned suite
https://github.com/terrapower/armi/blob/3882d739cc4f13a77e04938299c47ecb16cacc57/armi/cases/case.py#L982-L988

The `first_file.txt` was correctly moved. But this line set the `specialFiles` setting not to be a list of a single element, but a string with the file. Then, logic that writes the new settings file still sees an iterable of strings for `specialFiles` and writes out each character as if it were a new setting.

This caused the run to crash because the code that works with `specialFiles` cannot access the file `f`.

john-science:

If a setting has a list with a single file in it, `copyInterfaceInputs()` will correctly copy the file, but mangle the setting when it writes the settings file back out. This PR fixes that problem.

close #2097

## SCR Information

Change Type: fixes

One-Sentence Description: This fixes an edge case in `copyInterfaceInputs()` where there is a list of files to copy, but the list is only one element long.

---

## Checklist


armi/cases/tests/test_cases.py
```diff
@@ -601,7 +601,7 @@ def test_copyInputsHelper(self):
             self.assertFalse(os.path.exists(destFilePath))
             self.assertEqual(destFilePath, shuffleFile)
 
-    def test_copyInterfaceInputs_singleFile(self):
+    def test_copyInterfaceInputsSingleFile(self):
```
drewj-tp, line 604: Is this similar to `test_copyInterfaceInputsOneFile`

armi/cases/tests/test_cases.py
```diff
@@ -626,8 +626,8 @@ def test_copyInterfaceInputs_nonFilePath(self):
             newSettings = cases.case.copyInterfaceInputs(
                 cs, destination=newDir.destination
             )
-            self.assertFalse(os.path.exists(newSettings[testSetting]))
-            self.assertEqual(newSettings[testSetting], fakeShuffle)
+            self.assertFalse(os.path.exists(newSettings[testSetting][0]))
+            self.assertEqual(newSettings[testSetting][0], fakeShuffle)
```
drewj-tp, line 630: Shouldn't `newSettings[testSetting]` be the file path itself? For shuffle logic, we only provide a file not a list of files

armi/cases/tests/test_cases.py
```diff
@@ -724,7 +753,9 @@ def test_copyInterfaceInputs_absPath(self):
...
             # file exists
-            self.assertTrue(os.path.exists(newSettings[testSetting]))
+            self.assertTrue(os.path.exists(newSettings[testSetting][0]))
             # but not copied to this dir
-            self.assertFalse(os.path.exists(os.path.basename(newSettings[testSetting])))
-            self.assertEqual(str(newSettings[testSetting]), absFile)
+            self.assertFalse(
+                os.path.exists(os.path.basename(newSettings[testSetting][0]))
+            )
+            self.assertEqual(str(newSettings[testSetting][0]), absFile)
```
drewj-tp, line 761: Same concern here about making a setting that was pointing to a single file now point to a list with one file

drewj-tp (changes requested):

armi/cases/tests/test_cases.py
john-science: The difference is the "one file" version just inputs a single string to `copyInterfaceInputs()`. This version does the same thing, but it trips your edge case where someone provides a list of files, but the list is only one element long.

armi/cases/tests/test_cases.py
john-science: `newSettings[testSetting]` in this case returns a list of length one. And that one element is the file path.

armi/cases/tests/test_cases.py
john-science: Well, I could be wrong. But the only way I could reproduce your bug was to do "a list with only one string in it".

armi/cases/tests/test_cases.py
drewj-tp: But the original setting was just a string. Now we've gone from
```yaml
shuffleLogic: fakeFile.py
```
to
```yaml
shuffleLogic:
  - fakeFile.py
```

I would not expect this test to change: we should preserve settings that are just a string

armi/cases/tests/test_cases.py
drewj-tp: I would not expect this test to change. When we provide a setting that points to a string, we should get a string on the other side of the copy

armi/cases/tests/test_cases.py
john-science: Well, I could not reproduce your bug for this case: `shuffleLogic: fakeFile.py`. But I COULD reproduce your bug for this length-one list case:

```
shuffleLogic:
  - fakeFile.py
```

Perhaps I just wan't able to trip the bug for the non-list case when I tried. Do you think you can throw together a unit test that fails for the first case?  `shuffleLogic: fakeFile.py`

drewj-tp: > Well, I could not reproduce your bug for this case: shuffleLogic: fakeFile.py.

This is not the bug. That's one part of our expected behavior: a setting that points to a single string is preserved.

> But I COULD reproduce your bug for this length-one list case:

This is the bug. We take a setting that was a list of a single string and convert it to a string.

The test you added, showing we preserve a setting that has a list of a single file through the copy, would fail if you revert your changes to `case.py` and that demonstrates the problem

drewj-tp (approved):
Thank you @john-science!

PR #2106, opened 2025-03-31
Add Core.iterBlocks and Assembly.iterBlocks

drewj-tp:

Introduce mechanisms for iterating over `Blocks` in the `Reactor` that do not involve making intermediate lists.

`Core.iterBlocks()` assumes you have a `Core` with `Assemblies` with `Blocks`. This is not limited to hex `Reactors` but may not work in all cases. 

It is not as powerful as `Core.getBlocks` because that:

1. Looks to the spent fuel pool,
2. Looks to assemblies that may not exist _right now_

Instead, it just finds all `Blocks` that are grandchildren of the `Core`.

- `Core.iterBlocks` supports filtering by flags and/or filtering with a conditional function. 

- `Assembly.iterBlocks` is really a shortcut for iterating over an `Assembly` with some filtering. You could just as easily do `Assembly.iterChildren` or `Assembly.iterChildrenWithFlags`. Both of those are sensible if we want to keep the API small and composable. But we have `Assembly.getBlocks` so this felt like a good idea. At the very least, some occurrences like `assembly.getBlocks()[0]` were changed to `assembly[0]`

## SCR Information

Change Type: features

One-Sentence Description: Adding Core.iterBlocks and Asssembly.iterBlocks for traversing grandchildren of a Core.

One-line Impact on Requirements: A trivial implementation change to switch from getBlocks to iterBlocks in the tests for T_ARMI_MACRO_XS, T_ARMI_UMC_PARAM_BACKWARD0, T_ARMI_UMC_PARAM_BACKWARD1, T_ARMI_ROTATE_HEX_ASSEM, T_ARMI_ASSEM_BLOCKS, T_ARMI_R_SYMM, and T_ARMI_ADD_EDGE_ASSEMS.

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_cores.py
+++ b/armi/reactor/tests/test_cores.py
@@ -78,3 +78,19 @@ class HexCoreTests(unittest.TestCase):
             actuals = self.core.getAssemblies(zones=fakeZones)
         for a in actuals:
             self.assertIn(a.getLocation(), locations, msg=str(a))
+
+    def test_getBlocks(self):
+        """Test the ability to get all blocks in the core."""
+        blocks = []
+        for a in self.core:
+            blocks.extend(a)
+        actual = self.core.getBlocks()
+        self.assertAllIs(actual, blocks)
+
+    def test_getBlocksWithFlag(self):
+        """Test the ability to get all blocks with a flag in the core."""
+        blocks = []
+        for a in self.core:
+            blocks.extend(filter(lambda b: b.hasFlags(Flags.FUEL), a))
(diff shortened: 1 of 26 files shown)


armi/reactor/tests/test_cores.py
```diff
@@ -0,0 +1,130 @@
+# Copyright 2024 TerraPower, LLC
```
john-science, line 1:
```suggestion
# Copyright 2025 TerraPower, LLC
```

john-science (approved):
I love it!

Seems like a slam dunk.

PR #2107, opened 2025-04-01
Handing empty string defaults better in copyInterfaceInputs

opotowsky:

`copyInterfaceInputs` now handles the possibility that a file setting default could be an empty string. 

This scenario was not covered until this setting default was added downstream. The root of the issue is that `pathlib.Path("") = "."`. If that weren't the case, then I think this code would have just worked. Also, it seemed more robust to account for the absolute path scenario rather than relative path (which can include `..`, `.`, or nothing!) scenario. 

## SCR Information

Change Type: features

One-Sentence Description: Adding logic to copyInterfaceInputs to properly handle a file setting default that is an empty string. 

---

## Checklist

### DIFF
--- a/armi/cases/case.py
+++ b/armi/cases/case.py
@@ -931,13 +931,16 @@ def copyInterfaceInputs(
             for f in files:
                 WILDCARD = False
                 RELATIVE = False
+                EMPTY = False
                 if "*" in f:
                     WILDCARD = True
                 if ".." in f:
                     RELATIVE = True
+                if not f:
+                    EMPTY = True
 
                 path = pathlib.Path(f)
-                if not WILDCARD and not RELATIVE:
+                if not WILDCARD and not RELATIVE and not EMPTY:
                     try:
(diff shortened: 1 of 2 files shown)


john-science (approved):
Okay, so, your PR makes it so that if the default value of a list of files setting is empty, then we just exclude the setting entirely from the output file, rather than writing back out another empty setting?
 
Is that right?
 
And that is only possible if the setting is set to empty in code? Or can this happen from a settings file?

opotowsky:
> Okay, so, your PR makes it so that if the default value of a list of files setting is empty, then we just exclude the setting entirely from the output file, rather than writing back out another empty setting?

This doesn't affect anything regarding lists of files, but rather when the default for a file setting is defined as an empty string `""`.
 
> And that is only possible if the setting is set to empty in code? Or can this happen from a settings file?

This could happen in a settings file but it is unlikely. The scenario here was a default setting for a file being set to an empty stirng.

PR #2109, opened 2025-04-02
Store number densities in numpy arrays instead of dictionary

mgjarrett:

Store component number densities in two numpy arrays instead of a dictionary of `[str: float]`.

`c.p.nuclides`: array of byte strings (`dtype='S6'`) containing nuclide names
`c.p.numberDensities`: array of floats (`dtype='np.float64'`) containing the number densities for the corresponding nuclides in `c.p.nuclides`

There is a tradeoff here between speed (dictionary lookup is very fast) and memory footprint (dictionary is approximately 10x-20x larger in memory than two numpy arrays). This tradeoff is being made because some full-size reactor model objects constructed in downstream applications have become too large, causing runs to either crash or consume a disproportionate amount of computational resources.

## SCR Information

Change Type: features

One-Sentence Description: Change component number density storage from dictionary to numpy arrays.

---

## Checklist


armi/reactor/components/component.py
```diff
@@ -793,15 +816,33 @@ def updateNumberDensities(self, numberDensities, wipe=False):
...
 
         # change the densities
         if wipe:
-            self.p.numberDensities = {}  # clear things not passed
-        self.p.numberDensities.update(numberDensities)
+            self.p.numberDensities = None  # clear things not passed
+            self.p.nuclides = None  # clear things not passed
+            self.p.nuclides = np.array(
+                [nucName.encode() for nucName in numberDensities.keys()]
+            )
```
drewj-tp, line 830: Similar suggestion on `np.asanyarray`

armi/reactor/components/component.py
```diff
@@ -817,17 +858,14 @@ def updateNumberDensities(self, numberDensities, wipe=False):
...
         self.p.assigned = parameters.SINCE_ANYTHING
         self.p.paramDefs["numberDensities"].assigned = parameters.SINCE_ANYTHING
 
     def changeNDensByFactor(self, factor):
         """Change the number density of all nuclides within the object by a multiplicative factor."""
-        newDensities = {
-            nuc: dens * factor for nuc, dens in self.p.numberDensities.items()
-        }
-        self.p.numberDensities = newDensities
+        self.p.numberDensities *= factor
```
drewj-tp, line 868: Now we're talking! :rocket:

armi/reactor/components/component.py
```diff
@@ -695,14 +692,40 @@ def getNumberDensity(self, nucName):
...
+            for i, nuc in enumerate(byteNucs):
+                j = reverseIndex.get(nuc, -1)
+                if j >= 0:
+                    nDens[i] = nDensCopy[j]
+        else:
+            # if it's just a small subset of nuclides, use np.where for direct index lookup
+            for i, nuc in enumerate(byteNucs):
+                j = np.where(nuclideCopy == nuc)[0]
+                if j.size > 0:
+                    nDens[i] = nDensCopy[j[0]]
```
drewj-tp, line 722: I have a suggestion that we can probably tackle after this PR goes through.

If we ensure that the nuclide keys are sorted, we can use `np.searchsorted` which

1. Uses a binary search (on average faster than linear search)
2. Finds the first value of (or near) a match (`nuclideCopy == nuc` produces a `N`-vector of bool)
3. Is vectorized so we can search for multiple nuclides at once and numpy can handle iteration/parallelization for us as appropriate

```python
>>> u = np.array(["B10", "U235"], dtype="|S6")
>>> request = np.array(["B10", "HE4", "U235"], dtype="|S6")
>>> out = np.empty_like(request, dtype=float)
>>> ix = u.searchsorted(request)
>>> ix
array([0, 1, 1])
>>> mask = u[ix] == request
>>> mask
array([ True, False,  True])
```
so now we have a mask where True means we've found the nuclide the user asked for (pull from the density vector)
```python
>>> out[mask] = originalDens[mask]
>>> out[~mask] = 0
```
(we could also make the outgoing array zero at the start and skip the final step. Depends on how often we envision people asking for nuclides that don't exist)

## Challenges

We need to take care to ensure that the keys are always sorted. So calls to
- `Component.setNumberDensity`
- `Component.setNumberDensities`
- `Component.updateNumberDensities`

will need to handle correct insertion of these entries. And np array concatenation / insertion isn't _ideal_ (far less flexible than lists) so we could consider some overallocation trickery: allocate more space than you need so you aren't always allocating elements.

Fun fact: this is what [python lists do too](https://github.com/python/cpython/blob/0dbaeb94a8b39972ebda7782b50bb49488951e3c/Objects/listobject.c#L122-L126)!

armi/reactor/components/component.py
```diff
@@ -793,15 +816,33 @@ def updateNumberDensities(self, numberDensities, wipe=False):
...
+            nDensCopy = np.array(self.p.numberDensities)
+            for nucName, dens in numberDensities.items():
+                i = np.where(nucCopy == nucName.encode())[0]
+                if i.size > 0:
+                    nDensCopy[i[0]] = dens
+                else:
+                    newNucs.append(nucName.encode())
+                    newNumDens.append(dens)
+            self.p.nuclides = np.append(self.p.nuclides, newNucs)
+            self.p.numberDensities = np.append(nDensCopy, newNumDens)
```
drewj-tp, line 845: Trade off here w/ a sorter or unsorted nuclide arrays

1. sorted -> `searchsorted` can be vectorized
2. sorted -> need to then re-insert, in order (!), new nuclide strings and densities

(like my other suggestions on sorted nuclide names, this can be a follow on after we do some more studying and profiling)

armi/reactor/components/component.py
```diff
@@ -906,8 +943,11 @@ def getMass(self, nuclideNames=None):
         )
         nuclideNames = self._getNuclidesFromSpecifier(nuclideNames)
         # densities comes from self.p.numberDensities
-        densities = self.getNuclideNumberDensities(nuclideNames)
-        nDens = {nuc: dens for nuc, dens in zip(nuclideNames, densities)}
+        if len(nuclideNames) > 0:
+            densities = self.getNuclideNumberDensities(nuclideNames)
+            nDens = {nuc: dens for nuc, dens in zip(nuclideNames, densities)}
```
drewj-tp, line 948: `dict` can take in a `zip`
```python
>>> k = ["u", "b"]
>>> d = [1, 2]
>>> dict(zip(k, d))
{'u': 1, 'b': 2}
```
```suggestion
            nDens = dict(zip(nuclideNames, densities))
```

drewj-tp (commented):
Thank you so much for this @mgjarrett! I imagine we'll have a few other people put eyes on this since it's a very impactful low-level change.

I wanted to get some of my thoughts out there for improvements now that we're storing vectors of nuclides and densities. I don't consider the sorting suggestions to be blocking for this feature. And I'm happy to take them on in a follow up PR.

There's a lot of changes unrelated to this PR that hopefully would get cleaned up once #2103 gets merged

armi/bookkeeping/db/tests/test_comparedb3.py
```diff
@@ -184,7 +184,7 @@ def test_compareDatabaseSim(self):
                 dbs[1]._fullPath,
                 timestepCompare=[(0, 0), (0, 1)],
             )
-        self.assertEqual(len(diffs.diffs), 504)
+        self.assertEqual(len(diffs.diffs), 528)
```
john-science, line 187: Do we love this test?  Is it useful to know when we make a PR that changes the number of diffs?

john-science: Does it spark joy?

mgjarrett: I agree that it's pretty silly to just keep updating this number arbitrarily as things are changed in the database.

I think we want to verify that there are some contents in `diffs.diffs`, but maybe we just assert something like `len(diffs.diffs) > 1` and then it wouldn't need to be updated all the time. It's not perfect, but just a thought... open to suggestions.

john-science: Okay, in the `main` branch I have updated this test to not be so fragile.

john-science:
Bump!

This is still interesting to me.

armi/reactor/components/component.py
mgjarrett: This could definitely help, but ensuring that everything is always sorted would add a good bit of complication here, so I'll leave that for future work as you suggested.

armi/bookkeeping/db/database.py
```diff
@@ -1634,3 +1638,19 @@ def collectBlockNumberDensities(blocks) -> Dict[str, np.ndarray]:
         dataDict[nb.getDatabaseName()] = nucDensityMatrix[:, ni]
 
     return dataDict
+
+
+def _applyComponentNumberDensitiesMigration(comps, unpackedData):
```
john-science: It's a little strange this isn't a staticmethod, seeing as it is only used in one class. And it can't be called from other files because it's private.

armi/materials/tests/test_air.py
```diff
@@ -232,16 +232,17 @@ def test_massFrac(self):
         refO = 0.210748
         refAR = 0.004671
 
-        nDens = densityTools.getNDensFromMasses(air.pseudoDensity(Tk=300), air.massFrac)
+        nuclides, nDens = densityTools.getNDensFromMasses(air.pseudoDensity(Tk=300), air.massFrac)
```
john-science, line 235:
```suggestion
        _nuclides, nDens = densityTools.getNDensFromMasses(air.pseudoDensity(Tk=300), air.massFrac)
```

john-science: Either this or, better, test the values of `nuclides`. That is currently an unused variable.

armi/nucDirectory/nuclideBases.py
```diff
@@ -997,20 +981,27 @@ def initReachableActiveNuclidesThroughBurnChain(numberDensityDict, activeNuclide
 
         for interaction in nuclideObj.trans + nuclideObj.decays:
             try:
-                # Interaction nuclides can only be added to the number density
-                # dictionary if they are a part of the user-defined active nuclides
+                # Interaction nuclides can only be added to the number density dictionary if they
```
john-science:
```suggestion
                # Interaction nuclides can only be added to the number density arrays if they
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -997,20 +981,27 @@ def initReachableActiveNuclidesThroughBurnChain(numberDensityDict, activeNuclide
...
-                if productNuclide not in numberDensityDict:
-                    numberDensityDict[productNuclide] = 0.0
+                if productNuclide not in nucNames:
+                    newNucs.add(productNuclide.encode())
             except KeyError:
                 # Keep track of the first production nuclide
                 missingActiveNuclides.add(interaction.productNuclides)
 
-        difference = set(numberDensityDict).difference(memo)
+        newNDens = np.zeros(len(newNucs), dtype=np.float64)
```
john-science, line 996: Can we get a comment on this code block?  I'm not sure I follow. Thanks!

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -482,12 +482,21 @@ def getBlockNuclideTemperatureAvgTerms(block, allNucNames):
...
+
+        allByteNucs = [nucName.encode() for nucName in allNucNames]
+        ndens = []
+        nucCopy = np.array(component.p.nuclides)
+        nDensCopy = np.array(component.p.numberDensities)
+        reverseIndex = {nuc: i for i, nuc in enumerate(nucCopy)}
+        for nuc in allByteNucs:
+            i = reverseIndex.get(nuc, -1)
+            if i >= 0:
+                ndens.append(max(nDensCopy[i], TRACE_NUMBER_DENSITY))
```
john-science, line 496: Sorry for being obtuse. We'll blame Friday.

What's the `-1` logic here for? The default case is you want to get the first element of `component.p.numberDensities`? What's special about the first element?

armi/reactor/components/component.py
```diff
@@ -639,14 +636,49 @@ def getNumberDensity(self, nucName):
...
+        i = np.where(self.p.nuclides == nucName.encode())[0]
+        if i.size > 0:
+            return self.p.numberDensities[i[0]]
+        else:
+            return 0.0
 
     def getNuclideNumberDensities(self, nucNames):
         """Return a list of number densities for the nuc names requested."""
-        return [self.p.numberDensities.get(nucName, 0.0) for nucName in nucNames]
+        if isinstance(nucNames, (list, tuple, np.ndarray)):
```
john-science, line 647: Awww. A simple, one-line method becomes a 31-line method. That's a shame.

armi/reactor/components/component.py
```diff
@@ -639,14 +636,49 @@ def getNumberDensity(self, nucName):
...
+                j = np.where(nuclideCopy == nuc)[0]
+                if j.size > 0:
+                    nDens[i] = nDensCopy[j[0]]
+
+        return nDens
 
     def _getNdensHelper(self):
-        return dict(self.p.numberDensities)
+        nucs = self.getNuclides()
+        return dict(zip(nucs, self.p.numberDensities)) if len(nucs) > 0 else {}
```
john-science, line 691: I can't help but wonder if some of your logic in this file would be simplified if you just let these methods return arrays instead of dictionaries now.

Thoughts?

armi/reactor/components/component.py
```diff
@@ -639,14 +636,49 @@ def getNumberDensity(self, nucName):
...
+            for i, nuc in enumerate(byteNucs):
+                j = reverseIndex.get(nuc, -1)
+                if j >= 0:
+                    nDens[i] = nDensCopy[j]
+        else:
+            # if it's just a small subset of nuclides, use np.where for direct index lookup
+            for i, nuc in enumerate(byteNucs):
+                j = np.where(nuclideCopy == nuc)[0]
+                if j.size > 0:
+                    nDens[i] = nDensCopy[j[0]]
```
john-science: Out of curiosity, did this come about based on performance testing?

Because we COULD just use the `if`-block logic all the time, right? But we think that would be too slow?

armi/physics/neutronics/crossSectionGroupManager.py
mgjarrett: The `-1` is the default case if the `nuc` is not in the `reverseIndex`. If this is the case, `i` is -1, so we skip to the `else` and just append a `0.0` to the `ndens` array.

armi/reactor/components/component.py
mgjarrett: I could write a simple version that is only a few lines, where we just reconstruct the full dictionary and then use the previous one-line function to get the number densities from the dictionary, but I would expect that to be slower.

armi/nucDirectory/nuclideBases.py
mgjarrett: The previous version of this function updated the `numberDensityDict` with new nuclides.

In this version, the `nuclides` and `numberDensities` arrays have replaced the `numberDensityDict`.

This block updates the arrays `nuclides` and `numberDensities` with the new nuclides. I'll add a comment.

armi/reactor/components/component.py
mgjarrett: My goal was to make the smallest possible impact on the API by keeping function signatures the same where possible. I agree that it leads to a lot of back-and-forth between arrays and dictionaries, which makes some of the low-level functions more complicated than they would otherwise need to be.

I like the idea of minimizing the impact on downstream users, but if there are places where we think it's worth changing the API then I'm open to rearranging things.

armi/reactor/components/component.py
mgjarrett: I did some performance testing, although I didn't specifically tune the `len(nDensCopy) / 10` based on any specific results from performance testing.

The use case I'm trying to address here is the situation where `nDens` is on the order of several hundred, but the number densities have only been requested for a small number of nuclides (maybe 1-10). I think this is a somewhat common use case. I wouldn't want to create the whole dictionary of hundreds of number densities just to grab <10 number densities from it.

armi/reactor/components/component.py
mgjarrett: I refactored this down to 16 lines and moved the `else` portion of the last `if/else` block to a separate function, which is 9 lines.

drewj-tp (commented):
I have no lingering comments with this PR. I trust @john-science with the technical review and will follow up elsewhere

john-science (approved):

keckler:
After this PR, I am noticing an oddity that is causing me a lot of trouble. 

If one instantiates a component made of `Void` (i.e. it has zero number density), the value of `nuclides` and `numberDensities` will be an empty numpy array. This is correct and good. 

```
>> c = Circle("myCircle", "Void", 100, 100, 1)
>> c.p.nuclides
array([], dtype=float64)
>>  c.p.numberDensities
array([], dtype=float64)
```

Aside -- should the type for `c.p.nuclides` be `float64`? That seems wrong...

Anyway, if you build a reactor with that has a `Void` component, then save it to a database and later reload it, those values will not be of the same type.

```
>> r.core.getAssemblyWithStringLocation("001-001")[-2][0]
<Circle: plenum_gas>
>> r.core.getAssemblyWithStringLocation("001-001")[-2][0].material
<Material: Void>
>> r.core.getAssemblyWithStringLocation("001-001")[-2][0].p.nuclides
None
>> r.core.getAssemblyWithStringLocation("001-001")[-2][0].p.numberDensities
None
```

The fact that those parameters are `None` after reloading the database is causing issues in places where ARMI expects `nuclides` and `numberDensities` to be iterable.

@mgjarrett @john-science

mgjarrett:
The best solution is probably to update the setter (or implement a new one) for these parameters. They're currently using `isNumpyArray`. I would probably create a new setter, `isNumpyArrayNotNone` (or something like that) to just set it to `np.array(())` when a ` None` value is unpacked from the database.

I would hesitate (not opposed, just some hesitation) to apply this new setter to _all_ parameters that use the `isNumpyArray`, because we could have the reverse condition for some parameters (i.e., ARMI expects a `None` instead of an empty array).

keckler:
> The best solution is probably to update the setter (or implement a new one) for these parameters. They're currently using `isNumpyArray`. I would probably create a new setter, `isNumpyArrayNotNone` (or something like that) to just set it to `np.array(())` when a ` None` value is unpacked from the database.
> 
> I would hesitate (not opposed, just some hesitation) to apply this new setter to _all_ parameters that use the `isNumpyArray`, because we could have the reverse condition for some parameters (i.e., ARMI expects a `None` instead of an empty array).

https://github.com/terrapower/armi/pull/2655

PR #2115, opened 2025-04-08
Adding Reactor construction hook to Database.load()

alexhjames:

Currently the `beforeReactorConstruction` plugin hook is only called during a case settings based load but not during a database load. Without this hook, there is no plugin hook prior to loading a database. In particular, this hook is used by materials plugins to load materials data related to the reactor about to be constructed. Without this hook being called for database loads, the database loads just use whatever material data happened to be loaded prior to that point. 

## SCR Information

Change Type: fixes

One-Sentence Description: Add `beforeReactorConstruction` hook to database loads.

---

## Checklist


alexhjames:
@opotowsky This is likely why Nala has to have a mat_props load_all call in the `configure()`. I think you should be able to remove that after this PR.

opotowsky:
> @opotowsky This is likely why Nala has to have a mat_props load_all call in the `configure()`. I think you should be able to remove that after this PR.

Correct it will remove that one call, but not all the others in every other repo (for unit testing).

armi/bookkeeping/db/database.py
```diff
@@ -762,6 +762,7 @@ def load(
 
         cs = cs or self.loadCS(handleInvalids=handleInvalids)
         bp = bp or self.loadBlueprints()
+        getPluginManagerOrFail().hook.beforeReactorConstruction(cs=cs)
```
john-science, line 765: I don't really want to call the Interface stack randomly throughout the code. That seems like a bad precedent.

I think this will have a lot of indirect effects. Now when people put code in their `beforeReactorConstruction()` hook, they need to consider that the code is called in two scenarios not one.  How will we inform people who are writing a new `beforeReactorConstruction()` that this one hook is unique and not called in one-and-only-one scenario like all the other hooks?

alexhjames: Two thoughts: 

1) If this hook isn't going to be called during a reactor build from a database, then the hook needs to be renamed. "Before Reactor Construction" seems like it would reasonably be expected to be called before any reactor build not just case settings based one. 
2) There needs to be some hook that is inserted into the database load process so that materials can be loaded related to that particular reactor. Looking through the set of hooks available currently, there are no such hooks. This location seems like a reasonable place for that hook because `db.load()` is the equivalent of `reactor.factory()` for the database load process.

https://github.com/terrapower/armi/blob/2fc524aecbf32e7270e6348fd4665447bd47d645/armi/reactor/reactors.py#L179

I'm happy to relocate this hook to somewhere else in the code that has the same effect, create a new hook to be inserted here (or somewhere similar), or leave as is. Let me know what you prefer @john-science.

opotowsky: I think a new db/reactor load hook makes sense. Maybe we want to do more than set materials namespace order in that hook. 

But I defer to John here.

john-science (approved):
Approved pending downstream testing.

PR #2116, opened 2025-04-09
Cleaning up the codeTiming reports

john-science:

This is a cleanup of the code timing reports.

In short, I removed the columns "is active" and "time since" and changed the name of one column from "pauses" to "num iterations".

I also did some basic code cleanup, added some docstrings, and added some unit tests.

### Note

Per Tony's request, this PR has a small API change that I will have to fix downstream.

## SCR Information

Change Type: trivial

One-Sentence Description: Cleaning up the codeTiming reports.

---

## Checklist

### DIFF
--- a/armi/utils/codeTiming.py
+++ b/armi/utils/codeTiming.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Utilities related to profiling code."""
-
 import copy
 import functools
 import os
@@ -25,7 +24,7 @@ def timed(*args):
 
     Examples
     --------
-    ::
+    Here are some examples of using this method::
 
         @timed # your timer will be called the module+method name
         def mymethod(stuff):
@@ -63,24 +62,17 @@ def timed(*args):
         return time_decorator
     else:
         raise ValueError(
-            "The timed decorator has been misused. Input args were {}".format(args)
+            f"The timed decorator has been misused. Input args were {args}"
         )
 
 
-def getMasterTimer():
-    """Duplicate function to the MasterTimer.getMasterTimer method.
-
-    Provided for convenience and developer preference of which to use
-    """
-    return MasterTimer.getMasterTimer()
-
-
 class MasterTimer:
+    """A code timing interface, this class is designed to be a singleton."""
 
     _instance = None
 
     def __init__(self):
-        if MasterTimer._instance:
+        if MasterTimer._instance is not None:
             raise RuntimeError(
                 "{} is a pseudo singleton, do not attempt to make more than one.".format(
                     self.__class__.__name__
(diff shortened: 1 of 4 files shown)


armi/utils/codeTiming.py
```diff
@@ -358,11 +354,12 @@ def __exit__(self, *args, **kwargs):
...
     @property
     def isActive(self):
+        """Return True if the code for this timer still running."""
         return self._active
 
     @property
-    def pauses(self):
-        """If this number seems high remember .start() twice in a row adds a pause."""
+    def numIterations(self):
+        """If this number seems high remember .start() twice in a row adds a numIterations."""
```
albeanth, line 362:
```suggestion
        """If this number seems high, remember .start() twice in a row adds an iteration to numIterations."""
```

armi/utils/tests/test_codeTiming.py
```diff
@@ -28,52 +28,101 @@ def tearDown(self):
...
+
+        time.sleep(0.01)
+        t5 = timer.stop()
+        self.assertGreater(t5, t4)
+        self.assertEqual(timer.overStart, 0)
 
+        time.sleep(0.01)
+        t6 = timer.start()
+        self.assertGreater(t6, t5)
+        self.assertEqual(timer.overStart, 0)
```
albeanth, line 84: lol are all of these necessary? Or maybe wrap them in a for loop?

armi/utils/tests/test_codeTiming.py
```diff
@@ -28,52 +28,101 @@ def tearDown(self):
         codeTiming._Timer._frozen = False
         codeTiming.MasterTimer._instance = None
 
-    def test_method_definitions(self):
+    def test_methodDefinitions(self):
+        """Test that the timer decorators do not interupt the code being decorated."""
```
albeanth, line 32: would be great to have a test that shows that the decorator actually _do_ something too. thoughts?

john-science: That's the rest of the unit tests in the file, right?

albeanth: The rest of the tests cover `MasterTimer` functionality, yes. I suppose I meant more that we should probably show that the if a decorator is used it makes it into the report (or is otherwise useful). I don't see that tested anywhere. I think the args options for the decorator aren't fully tested either - at least not in this test module.

john-science: If it helps, this PR gives the `codeTiming.py` file 98.51% code coverage: https://coveralls.io/jobs/163599512/source_files/2289616344

I can look into adding more testing.

albeanth: Having that high coverage metric is great! If you're confident that the decorator works as intended (adding it to a function does indeed populate in the table with an expected title) with the existing tests, great. I just didn't see one explicitly but I am happy to convinced otherwise

john-science: Okay, I added a test to show that using the decorator works (the report is populated).

albeanth (approved):
Thanks @john-science !

albeanth:
Thoughts on replacing this
https://github.com/terrapower/armi/blob/904f06c2822876ee0b91d0de43ab79a539d5f8b2/armi/operators/operator.py#L506

with
```python
with self.timer.getTimer(f"{interface.name}, {interactionName}, {cycleNodeTag}"):
```
?

john-science:
> ...
> with self.timer.getTimer(f"{interface.name}, {interactionName}, {cycleNodeTag}"):
> ...

Well, right now the message is:

```python
            interactionMessage = " {} interacting with {} ".format(
                interactionName, interface.name
            )
```

And I'm not sure I like "interacting with". I was thinking something more like:

```python
f"{interface.name}.{interactionName}()"
```

would be nicer, e.g.: `GlobalFluxInterface.interactBOL()`, `FuelHandlerInterface.interactEveryNode()`, etc.

But I think if we put "cycleNodeTag" in there it would break the timer reports. The "average" and "cummulative" columns for these calls would be broken.

albeanth:
> But I think if we put "cycleNodeTag" in there it would break the timer reports. The "average" and "cummulative" columns for these calls would be broken.

Not sure I follow how the average and cumulative would be broken. Wouldn't the timers just be separated out by cycle and time node? So if we have 1 cycle and 2 time nodes for a simulation, it would go from something like  

 TIMER REPORTS | CUMULATIVE (s) | AVERAGE (s) | NUM ITERS 
 ----- | ------- | ------ | ---------------- 
GlobalFluxInterface.interactEveryNode | 100 | 50 | 2 

to something like 

 TIMER REPORTS | CUMULATIVE (s) | AVERAGE (s) | NUM ITERS 
 ----- | ------- | ------ | ---------------- 
GlobalFluxInterface, interactEveryNode, c0n0 | 50 | 50 | 1
GlobalFluxInterface, interactEveryNode, c0n1 | 50 | 50 | 1

albeanth:
> with
> 
> ```python
> with self.timer.getTimer(f"{interface.name}, {interactionName}, {cycleNodeTag}"):
> ```
> 
> ?

I'd change this to be 
```python
with self.timer.getTimer(f"{interface.name}, {interactionName}, c{self.r.p.cycle}n{self.r.p.timeNode}"):
```

that won't separate out iters for interactCoupled, but I think it's ok to show those with numIters > 1

john-science:
> TIMER REPORTS	CUMULATIVE (s)	AVERAGE (s)	NUM ITERS
> GlobalFluxInterface, interactEveryNode, c0n0	50	50	1
> GlobalFluxInterface, interactEveryNode, c0n1	50	50	1

Tight.... so why have the average/cumm/iters columns if everything is by cycle and node?   For a 6-cycle run with dozens of time nodes, it'll be a lot harder to read the final table and see what percentage of the run was due to... some particular interface. 

And, I think, that is what Mark wants. He wants to be able to look at an interface line and the new "total time" on the bottom, and get a rough estimate at a glance of what percentage of the run was due to a particular interface. (I am like 95% on this.)

albeanth:
> Tight.... so why have the average/cumm/iters columns if everything is by cycle and node?

In the example I gave where there is an even split, there's is arguably no utility. _But_, if one cycle/timenode takes way longer than another, there's arguably utility. But there's also a timer report at EOC so there's at least some breakup between cycles.

I would settle for   

 TIMER REPORTS | CUMULATIVE (s) | AVERAGE (s) | NUM ITERS 
 ----- | ------- | ------ | ---------------- 
GlobalFluxInterface.interactEveryNode | 100 | 50 | 2 

though. I won't put my foot down on splitting by cycle time node.

albeanth (approved):

Issue #1923, opened 2024-10-03
HexBlock.autoCreateSpatialGrids raises lots of spurious exceptions


drewj-tp:
It's not uncommon for me to see messages like
```
[warn] Could not create a spatialGrid for block BLOCK, multiplicities are not 1 or N they are {1}
```
which, what? I read this as "multiplicities are not 1 they are 1" which enforces a sense of distrust in warning messages.

Tracking this down, we try/except `Block.autoCreateSpatialGrids` when creating a block in blueprints

https://github.com/terrapower/armi/blob/f1f3dbea05c2564404a03d8e053f72c2af15daa9/armi/reactor/blueprints/blockBlueprint.py#L243-L247

and the `HexBlock` implementation throws this error

https://github.com/terrapower/armi/blob/f1f3dbea05c2564404a03d8e053f72c2af15daa9/armi/reactor/blocks.py#L2353-L2358

The error appears to be triggered because we don't have two multiplicities. Which I guess would be what you'd expect an assembly with pins to have: single multiplicity for a duct or other bounding component, and then a multiplicity > 1 for pin-like components.

## Remedy

If there is only a single multiplicity, and it's _close to one_, do not attempt to make a spatial grid on the block. This feels acceptable. There's nothing here to make a spatial grid so we won't. It's not that we cannot make a spatial grid (e.g., invalid number of pins), just that it doesn't make sense to create on here. Maybe the autocreate method raises an exception that is silently discarded in the block blueprint?

keckler:
+1 on this issue. I see this warning ~10 bazillion times whenever I run, and the warning message is not very useful. 

I would also say that the warning is not entirely justified, considering that nobody guaranteed the user that a spatial grid would be created on their block.

drewj-tp:
> I would also say that the warning is not entirely justified, considering that nobody guaranteed the user that a spatial grid would be created on their block.

Totally get the confusion, especially for things that don't make sense to have grids.

Surprise surprise, there's a setting, defaulting to `True`, that controls this

https://github.com/terrapower/armi/blob/f1f3dbea05c2564404a03d8e053f72c2af15daa9/armi/settings/fwSettings/globalSettings.py#L832-L839

john-science:
Well, I like an easy one.

I think, perhaps, we should:

> Make the default for `autoGenerateBlockGrids` / `CONF_BLOCK_AUTO_GRID` `False`.

[...]

PR #2117, opened 2025-04-10
Reducing the warnings from Block.autoCreateSpatialGrids

john-science:

Reducing the number of warnings from `Block.autoCreateSpatialGrids()`. Essentially, I used a multi-pronged approach here:

1. I changed many of these from "warning" level to "extra" level
2. I fixed an issue where we were looking for "1", but not finding "1.0" in the list of multiplicities.
3. I isolated the case where all the multiplicities are 1 or 1.0.

I found this to be a particularly annoying thing to unit test, as I had to cover huge numbers of cases with realistic Blocks with various different collections of multiplicities. I was like 20 tests in and several hundred lines of code and... What do you say we just look at your log statements and see if I fixed the problem?  What do you think, is that crazy?

## SCR Information

Change Type: trivial

One-Sentence Description: Reducing the warnings from Block.autoCreateSpatialGrids

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -2186,7 +2186,7 @@ class HexBlock(Block):
 
         In this case, a simple block would be one that has either multiplicity of components equal
         to 1 or N but no other multiplicities. Also, this should only happen when N fits exactly
-        into a given number of hex rings.  Otherwise, do not create a grid for this block.
+        into a given number of hex rings. Otherwise, do not create a grid for this block.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


armi/reactor/blocks.py
```diff
@@ -2211,12 +2211,21 @@ def autoCreateSpatialGrids(self, systemSpatialGrid=None):
...
+        if len(multz) == 1 and list(multz)[0] == 1.0:
+            runLog.extra(
+                f"Will not create a spatialGrid for block {self.p.type}, multiplicities are are all 1.",
+                single=True,
             )
+            return
+        elif len(multz) != 2 or 1.0 not in multz:
+            runLog.extra(
+                f"Could not create a spatialGrid for block {self.p.type}, multiplicities are not 1 "
+                f"or N they are {mults}",
```
drewj-tp, line 2225: We will create the spatial grid if we have a `{1, N}` multiplicity. I think the set notation makes the expectations explicit, and why this block did not support that behavior.
```suggestion
                f"Could not create a spatialGrid for block {self.p.type}, multiplicities are not {{1, N}} "
                f"they are {mults}",
```

armi/reactor/blocks.py
```diff
@@ -2211,12 +2211,21 @@ def autoCreateSpatialGrids(self, systemSpatialGrid=None):
...
         mults = {c.getDimension("mult") for c in self.iterComponents()}
 
-        if len(mults) != 2 or 1 not in mults:
-            raise ValueError(
-                "Could not create a spatialGrid for block {}, multiplicities are not 1 or N they are {}".format(
-                    self.p.type, mults
-                )
+        # Do some validation: Should we try to create a spatial grid?
+        multz = {float(m) for m in mults}
+        if len(multz) == 1 and list(multz)[0] == 1.0:
```
drewj-tp, line 2216:
```suggestion
        if len(multz) == 1 and 1.0 in multz:
```

drewj-tp (changes requested):
Thanks @john-science! The logs are in a better place now

keckler:
Thank you for doing this.

Can we try the suggestion that I mentioned [here](https://github.com/terrapower/armi/issues/1923#issuecomment-2571693352)?

I ask because, for models with mixed assemblies, it will be necessary to define the spatial grid manually. So for those blocks it (1) is not necessary to even try to create a spatial grid, and (2) will be misleading because a warning will be printed even though the block already has a grid.

keckler (approved):

drewj-tp (approved):

PR #2118, opened 2025-04-11
Adding a FilletedHexagon shape

### LINKED ISSUE #1301: Add support for hexagonal ducts with filleted corners
ARMI currently supports Hexagonal ducts with sharp corners, but in many practical cases the hex ducts have rounded (filleted) corners with a specified corner radius, which reduces both the flow area and wetted perimeter of these ducts and increases the hydraulic diameter by as much as ~2.4% (neglecting pins). Capability should be added to support either specifying a corner radius when defining a component of shape `Hexagon`, or perhaps a new `FilletedHexagon` shape should be added with that capability. 

If $D$ denotes the flat-to-flat inner duct diameter (i.e., the value returned by `duct.getDimension('ip')`) and $r$ denotes the corner radius ($0 \le r \le D/2$), then the perimeter, $P$, of the inner duct surface and the cross-sectional area $A$ inside the duct are given, respectively, by:

$$P(D,r) = \underbrace{2\sqrt{3}D}_{P(D,r=0)} \left[1 - \left(1 - \frac{\pi}{2\sqrt{3}}\right)\left(\frac{2r}{D}\right)\right]$$

and

$$A(D,r) = \underbrace{\frac{\sqrt{3}}{2}D^2}_{A(D,r=0)} \left[1 - \left(1 - \frac{\pi}{2\sqrt{3}}\right)\left(\frac{2r}{D}\right)^2 \right]$$

where $P(D, r=0)$ and $A(D, r=0)$ denote the perimeter and cross-sectional area assuming unrounded corners. Note that in the limiting case where $r = D/2$, the hexagon reduces to a circle and we have $A(D, D/2) = \pi D^2 /4$ and $P(D, D/2) = \pi D$ as expected. 

The hydraulic diameter of a filleted duct is given by

$$D_{hyd} = D \left[ \frac{1 - \left(1 - \frac{\pi}{2\sqrt{3}} \right) \left( \frac{2r}{D} \right)^2}{ 1 - \left(1 - \frac{\pi}{2\sqrt{3}} \right) \left( \frac{2r}{D} \right) } \right]$$

The quantity in brackets is maximum when $r \approx 0.256 D$, in which case $D_{hyd} \approx 1.0224 D$

john-science:

Adding a FilletedHexagon shape, which is just a hexagon with rounded corners. It behaves mostly like a Hexagon, but it requires one more input, the radius of curvature of the inner and outer rounded corners.

close #1301 

## SCR Information

Change Type: features

One-Sentence Description: Adding a FilletedHexagon shape, which is just a hexagon with rounded corners.

---

## Checklist


armi/reactor/components/complexShapes.py
```diff
@@ -140,6 +137,69 @@ def getCircleInnerDiameter(self, Tc=None, cold=False):
         return self.getDimension("holeOP", Tc, cold)
 
 
+class FilletedHexagon(basicShapes.Hexagon):
+    """Hexagon with rounded corners.
+
+    This shape has a hexagonal hole with rounded corners cut out of the center of it. By default,
```
alexhjames, line 143: This description is really confusing to me. Is this trying to model a hexagon where the corners are filleted on both the inner and outer surface?

john-science: Yup!

armi/reactor/blocks.py
```diff
@@ -2414,14 +2407,21 @@ def getHydraulicDiameter(self):
...
 
 
+class FilletedHexBlock(HexBlock):
+    """Defines a Block shaped like a hexagon with rounded corners."""
+
+    PITCH_COMPONENT_TYPE: ClassVar[_PitchDefiningComponent] = (
+        components.FilletedHexagon,
+    )
+
+
```
keckler, line 2424: Is a block with rounded corners something that is wanted? This is confusing to me, because `Block`s are used to define the core lattice, so what is then happening to the empty space between blocks near the rounded corners?

armi/reactor/components/basicShapes.py
```diff
@@ -149,27 +151,14 @@ def getCircleInnerDiameter(self, Tc=None, cold=False):
...
         area = math.sqrt(3.0) / 2.0 * (op**2 - ip**2)
         area *= mult
         return area
 
-    def getPerimeter(self, Tc=None):
-        """Computes the perimeter of the hexagon component in cm."""
-        ip = self.getDimension("ip", Tc)
-        mult = self.getDimension("mult", Tc)
-        perimeter = 6 * (ip / math.sqrt(3)) * mult
-        return perimeter
```
keckler, line 171: Why is this being deleted?

armi/reactor/components/complexShapes.py
```diff
@@ -140,6 +137,69 @@ def getCircleInnerDiameter(self, Tc=None, cold=False):
...
+    that inner hole has a diameter of zero, making this a solid object with no hole.
+    """
```
keckler, line 145: I agree with Alex, this description is not very clear to me. I have a suggestion, feel free to take it or leave it.
```suggestion
    """A hexagon with a hexagonal hole cut out of the center of it, where the corners of both the outer and inner hexagons are rounded.

    By default, the inner hole has a diameter of zero, making this a solid object with no hole.
    """
```

armi/reactor/components/complexShapes.py
```diff
@@ -140,6 +137,69 @@ def getCircleInnerDiameter(self, Tc=None, cold=False):
...
+
+        area = 1.0 - (1.0 - (math.pi / (2.0 * math.sqrt(3)))) * (2 * r / D) ** 2
+        area *= (math.sqrt(3.0) / 2.0) * D**2
+        return area
+
+    def getComponentArea(self, cold=False, Tc=None):
+        """Computes the area for the rounded hexagon component in cm^2."""
+        op = self.getDimension("op", cold=cold, Tc=None)
+        ip = self.getDimension("ip", cold=cold, Tc=None)
+        r = self.getDimension("cornerR", cold=cold, Tc=None)
```
keckler, line 195: Why are these three calls not passing through the value of `Tc`?

armi/reactor/components/componentParameters.py
```diff
@@ -239,6 +239,17 @@ def getHexHoledCircleParameterDefinitions():
...
+def getFilletedHexagonParameterDefinitions():
+    """Return parameters for FilletedHexagon."""
+    pDefs = parameters.ParameterDefinitionCollection()
+    with pDefs.createBuilder(location=ParamLocation.AVERAGE, saveToDB=True) as pb:
+        pb.defParam(
+            "cornerR", units=units.CM, description="Radius of curvature of the corners"
+        )
+
+    return pDefs
+
```
keckler, line 251: To be honest I don't fully understand this part of ARMI, so I might be misunderstanding. But why do we not want to also store other defining aspects of this shape, like `ip` and `op`? We store those things for the "normal" hexagon, but apparently not for the filleted hexagon or holed hexagon. Is this intentional?

armi/reactor/components/basicShapes.py
john-science: It turns out, no one is using perimeter on a hex shape, I checked.

keckler: I don't see how that is related to the purpose of this PR, though. 

I'm not trying to be provocative, this just seems a bit random and unwarranted unless there is something actually wrong with the perimeter calculation. Maybe there is a reason that I am not aware of for removing stuff like this.

john-science: I don't know if this calculation is correct. Should the perimeter of a hexagon with a hexagonal hole cut out of it include the internal perimeter AND the external one?

Well, for SOME reason, this only includes the internal and not the external perimeter. Why is that? What does that mean?

At the very least, this method needs a long docstring telling me what it is for. But it doesn't have that, so I can only guess.

But, @keckler , to answer your question, I am subclassing this class. And thus would need to subclass this method and re-implement it. But how can I reimplement something whose purpose I don't understand. That's how I got to this point.

Thoughts?

keckler: Makes sense, thank you both for the context.

armi/reactor/blocks.py
john-science: So, your notion is that we should have `FilletedHexagon` as a shape for Components, but not have a Block of with that shape?

@dlangewisch Will having just the "rounded hexagon" _shape_ be enough for you to build components of this shape? Or do you need entire blocks with this shape?

armi/reactor/components/complexShapes.py
john-science: typo! Fixing!

armi/reactor/components/componentParameters.py
john-science: > But why do we not want to also store other defining aspects of this shape, like ip and op?

Don't worry! Those come from subclasses the hexagon.

armi/reactor/blocks.py
keckler: Correct, that is what I would expect. I do not know the full motivation for having this new shape, but it doesn't seem to make sense to have it define a block.

john-science:
@dlangewisch I can change this shape to have different radii of curvature on the inner and outer hexagons, if you like. BUT that will leave room for people to do silly things, that are non-physical, like the inner hexagon being so rounded its a circle and overlapping the outer hexagon:

image

I identified 6 special cases I can solve for to trap such conditions. But the general case (where both hexagons have slightly rounded corners) is hard to solve for generally.

So, if I add that extra radii, I won't be able to catch ALL non-physical conditions, just most.

dlangewisch (approved):

PR #2120, opened 2025-04-15
Cylindrical Cross Section model updates

mgjarrett:

I have recently run into a few corner-case type issues when exercising the 1D cylindrical cross section model. This PR is intended to address those issues to make the 1D cylindrical model more robust and able to be used in more scenarios.

The problems and how they are being addressed are listed below:
 * The representative block for a `CylindricalComponentsAveageBlockCollection` has component temperatures that come from the selected candidate block. This PR changes the `_makeRepresentativeBlock` method to set component temperatures based on the average over the block collection.
 * `BlockConverter.dissolveComponentIntoComponent` loses the nuclides from the solute component when the solute is defined by linked dimensions that are linked to the solvent. This was an order of operations issue, which was resolved by moving up the `solute.mergeNuclidesInto(solvent)` within `dissolveComponentIntoComponent`.
 * The `HexComponentsToCylConverter` determines whether a component is associated with a pin based on the multiplicity of that component. This logic breaks down when a pin has multiple components that have the same flags. The implementation was changed to look through a hard-coded set of flags that are associated with pin-level components.

## SCR Information

Change Type: fixes

One-Sentence Description: Updates to the CylindricalComponentsAverageBlockCollection representative block temperature treatment and the HexComponentsToCylConverter.

One-line Impact on Requirements: The implementation of R_ARMI_BLOCKCONV_HEX_TO_CYL is being updated. The current functionality should be preserved, but it should be more robust across more situations and corner cases with these updates. The implementation of R_ARMI_XSGM_CREATE_REPR_BLOCKS for cylindrical XS models (I_ARMI_XSGM_CREATE_REPR_BLOCKS1) is also being updated to set the representative block's component temperatures based on the block collection average temperatures.
---

## Checklist

### DIFF
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -1448,12 +1448,10 @@ class Component(composites.Composite, metaclass=ComponentType):
         density = composites.Composite.density(self)
 
         if not density and not isinstance(self.material, void.Void):
-            # possible that there are no nuclides in this component yet. In that case,
-            # defer to Material. Material.density is wrapped to warn if it's attached
-            # to a parent. Avoid that by calling the inner function directly
-            density = self.material.density.__wrapped__(
-                self.material, Tc=self.temperatureInC
-            )
+            # possible that there are no nuclides in this component yet. In that case, defer to
+            # Material. Material.density is wrapped to warn if it's attached to a parent. Avoid that
+            # by calling the inner function directly
+            density = self.material.density(Tc=self.temperatureInC)
 
         return density
 
--- a/armi/reactor/converters/blockConverters.py
+++ b/armi/reactor/converters/blockConverters.py
@@ -90,6 +90,7 @@ class BlockConverter:
         # shape volume computations will fail
         newBlock.remove(solute, recomputeAreaFractions=False)
         self._sourceBlock = newBlock
+        solute.mergeNuclidesInto(solvent)
 
         # adjust new shape area.
         if solvent.__class__ is components.DerivedShape:
@@ -122,7 +123,6 @@ class BlockConverter:
             self._verifyExpansion(solute, solvent)
 
         solvent.changeNDensByFactor(oldArea / solvent.getArea())
-        solute.mergeNuclidesInto(solvent)
 
     def _checkInputs(self, soluteName, solventName, solute, solvent):
         if solute is None or solvent is None:
@@ -773,7 +773,9 @@ class HexComponentsToCylConverter(BlockAvgToCylConverter):
             od=outDiameter,
             mult=1,
         )
-        circle.setNumberDensities(baseComponent.getNumberDensities())
+        # set number densities directly to avoid material expansion
+        circle.p.nuclides = baseComponent.p.nuclides
+        circle.p.numberDensities = baseComponent.p.numberDensities
         circle.p.flags = baseComponent.p.flags
         return circle
 
--- a/armi/reactor/converters/blockConverters.py
+++ b/armi/reactor/converters/blockConverters.py
@@ -642,9 +642,13 @@ class HexComponentsToCylConverter(BlockAvgToCylConverter):
 
         pinComponentFlags = [
             Flags.FUEL,
+            Flags.ANNULAR | Flags.VOID,
             Flags.GAP,
+            Flags.BOND,
             Flags.LINER,
             Flags.CLAD,
+            Flags.WIRE,
+            Flags.CONTROL,
             Flags.REFLECTOR,
             Flags.SHIELD,
             Flags.SLUG,
--- a/armi/reactor/converters/blockConverters.py
+++ b/armi/reactor/converters/blockConverters.py
@@ -640,6 +640,17 @@ class HexComponentsToCylConverter(BlockAvgToCylConverter):
         """
         pinComponents, nonPins = [], []
 
+        pinComponentFlags = [
+            Flags.FUEL,
+            Flags.GAP,
+            Flags.LINER,
+            Flags.CLAD,
+            Flags.REFLECTOR,
+            Flags.SHIELD,
+            Flags.SLUG,
+            Flags.PIN,
+        ]
+
         for c in self._sourceBlock:
 
             # If the area of the component is negative than this component should be skipped
@@ -649,10 +660,7 @@ class HexComponentsToCylConverter(BlockAvgToCylConverter):
             if c.getArea() < 0.0:
                 continue
 
-            if (
-                self._sourceBlock.getNumComponents(c.p.flags)
-                == self._sourceBlock.getNumPins()
-            ):
+            if any(c.hasFlags(f) for f in pinComponentFlags):
                 pinComponents.append(c)
             elif (
                 c.name != "coolant"
--- a/armi/nuclearDataIO/nuclearFileMetadata.py
+++ b/armi/nuclearDataIO/nuclearFileMetadata.py
@@ -110,9 +110,12 @@ class _Metadata:
             otherVal = other[key]
             mergedVal = None
             if not properties.numpyHackForEqual(selfVal, otherVal):
-                raise exceptionClass(
+                exceptionMsg = (
                     "{libType} {key} metadata differs between {lib1} and {lib2}; Cannot Merge\n"
-                    "{key} has values of {val1} and {val2}".format(
+                    "{key} has values of {val1} and {val2}"
+                )
+                raise exceptionClass(
+                    exceptionMsg.format(
                         libType=fileType,
                         lib1=selfContainer,
                         lib2=otherContainer,
--- a/armi/physics/neutronics/crossSectionGroupManager.py
+++ b/armi/physics/neutronics/crossSectionGroupManager.py
@@ -570,7 +570,9 @@ class CylindricalComponentsAverageBlockCollection(AverageBlockCollection):
         repBlock.p.percentBu = self._calcWeightedBurnup()
         componentsInOrder = self._orderComponentsInGroup(repBlock)
 
-        for i, (c, allSimilarComponents) in enumerate(zip(sorted(repBlock), componentsInOrder)):
+        for i, (c, allSimilarComponents) in enumerate(
+            zip(sorted(repBlock), componentsInOrder)
+        ):
             allNucsNames, densities = self._getAverageComponentNucs(
                 allSimilarComponents, bWeights
             )
--- a/armi/physics/neutronics/crossSectionGroupManager.py
+++ b/armi/physics/neutronics/crossSectionGroupManager.py
@@ -521,7 +521,7 @@ def getBlockNuclideTemperatureAvgTerms(block, allNucNames):
     return nvt, nv
 
 
-class CylindricalComponentsAverageBlockCollection(BlockCollection):
+class CylindricalComponentsAverageBlockCollection(AverageBlockCollection):
     """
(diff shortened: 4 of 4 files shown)


drewj-tp:
@mgjarrett bump

armi/reactor/components/component.py
```diff
@@ -1151,6 +1151,9 @@ def mergeNuclidesInto(self, compToMergeWith):
         """
         # record pre-merged number densities and areas
         aMe = self.getArea()
+        # if negative-area gap, treat is as 0.0 and return
+        if aMe <= 0.0:
```
drewj-tp, line 1155: This feels bad. What would cause the gap to have a negative area? I could understand a zero area being something we don't want to mess with. But negative? Does that imply the block has too much stuff in it, so something needs to have a negative area?

armi/reactor/converters/blockConverters.py
```diff
@@ -100,22 +102,39 @@ def dissolveComponentIntoComponent(self, soluteName, solventName, minID=0.0):
...
+                    solvent.setDimension("id", soluteID, cold=False)
+                if solvent.getDimension("od", cold=False) < soluteOD:
+                    runLog.debug("Increasing OD of {} to accommodate {}.".format(solvent, solute))
+                    solvent.setDimension("od", soluteOD, cold=False)
+                if solvent.getDimension("id", cold=False) < minID:
+                    runLog.debug("Updating the ID of {} the the specified min ID: {}.".format(solvent, minID))
+                    solvent.setDimension("id", minID, cold=False)
+            else:
+                # can only merge a negative-area component if one of the dimensions is linked
+                matchedDimesion = False
```
drewj-tp: Okay I'm more confused. There's a legitimate operation where we get negative areas but only if we have linked dimensions? 

Would this be when fuel expands into and over the gap, functionally removing the gap? Which doesn't seem like it's problem with this PR, more an issue of how ARMI handles expansion.

armi/reactor/converters/blockConverters.py
```diff
@@ -570,6 +593,22 @@ def _classifyComponents(self):
...
+            Flags.LINER,
+            Flags.CLAD,
+            Flags.WIRE,
+            Flags.CONTROL,
+            Flags.REFLECTOR,
+            Flags.SHIELD,
+            Flags.SLUG,
+            Flags.PIN,
+            Flags.POISON,
+        ]
```
drewj-tp: Very broad but still bespoke set of flags. And there's no real way to configure this outside of a code change.

I would prefer if we could promote this to a class variable. Would something like `HexComponentsToCylConverter.PIN_COMPONENT_FLAGS = (Flags.FUEL, ..., )` satisfy the needs here?

armi/reactor/components/component.py
drewj-tp: This may be handled by https://github.com/terrapower/armi/pull/2120#discussion_r2192865233 and linked dims and stuff

mgjarrett: Yes, a negative gap can happen if the blueprints have a zero-area gap where two components meet, (e.g, ``liner.od == gap.id == gap.od == clad.id`)

In this case, if the `liner` and `clad` components are different materials or are at different temperatures, the thermal expansion of the `liner` can encroach on the `clad` (i.e. `liner.od > clad.id` thus `gap.id > gap.od` and `gap.getArea() < 0.0`.

We can still get rid of the gap and update the `solvent` component's dimensions accordingly, but we can't transfer any nuclides from the gap to the `solvent` component.

armi/reactor/converters/blockConverters.py
mgjarrett: Yes, that's basically the scenario being treated here

armi/reactor/converters/blockConverters.py
mgjarrett: Yeah I don't see any reason for it not to be a class variable.

The alternative to a hard-coded list of flags is to have the user enter what flags they want to be considered pins as a setting, but that seems like an unnecessary burden..

armi/tests/armiRun.yaml
```diff
@@ -35,6 +35,22 @@ settings:
...
+      mergeIntoClad:
+        - gap2
+        - inner liner
+        - gap3
+        - outer liner
+        - gap4
+      mergeIntoFuel:
+        - gap1
+      numInternalRings: 1
+      numExternalRings: 1
```
drewj-tp, line 53: I have no idea the answer to this question but will this have knock-on effects to other tests that use `armiRun.yaml`? I guess we'll see sooner or later 🙃

armi/reactor/converters/blockConverters.py
drewj-tp: > The alternative to a hard-coded list of flags is to have the user enter what flags they want to be considered pins as a setting, but that seems like an unnecessary burden..

💯 Just wanted to afford someone the option, somehow, to change these variables. A user setting is not what I'm after

drewj-tp (approved):

armi/tests/armiRun.yaml
mgjarrett: I did this to facilitate a unit test for the new features. If it has unmanageable knock-on effects, I can always just make a copy of this test input within the downstream repo where it's needed and make these modifications there. But I don't necessarily expect there to be any side effects, so if all of the unit tests continue to pass with this change then it's easier to do this and avoids unnecessary duplication of test inputs.

armi/reactor/converters/blockConverters.py
```diff
@@ -100,22 +102,39 @@ def dissolveComponentIntoComponent(self, soluteName, solventName, minID=0.0):
...
+                    solvent.setDimension("id", soluteID, cold=False)
+                    matchedDimension = True
+                if solvent.getDimension("od", cold=False) == soluteID:
+                    runLog.debug("Decreasing OD of {} to accommodate {}.".format(solvent, solute))
+                    solvent.setDimension("od", soluteOD, cold=False)
+                    matchedDimension = True
+                if not matchedDimension:
+                    errorMsg = (
+                        "Cannot merge negative-area component {solute} into {solvent} without the two being linked."
+                    )
```
john-science:
```suggestion
            if soluteArea >= 0.0:
                if solvent.getDimension("id", cold=False) > soluteID:
                    runLog.debug(f"Decreasing ID of {solvent} to accommodate {solute}.")
                    solvent.setDimension("id", soluteID, cold=False)
                if solvent.getDimension("od", cold=False) < soluteOD:
                    runLog.debug(f"Increasing OD of {solvent} to accommodate {solute}.")
                    solvent.setDimension("od", soluteOD, cold=False)
                if solvent.getDimension("id", cold=False) < minID:
                    runLog.debug(f"Updating the ID of {solvent} the the specified min ID: {minID}.")
                    solvent.setDimension("id", minID, cold=False)
            else:
                # can only merge a negative-area component if one of the dimensions is linked
                matchedDimension = False
                if solvent.getDimension("id", cold=False) == soluteOD:
                    runLog.debug(f"Increasing ID of {solvent} to accommodate {solute}.")
                    solvent.setDimension("id", soluteID, cold=False)
                    matchedDimension = True
                if solvent.getDimension("od", cold=False) == soluteID:
                    runLog.debug(f"Decreasing OD of {solvent} to accommodate {solute}.")
                    solvent.setDimension("od", soluteOD, cold=False)
                    matchedDimension = True
                if not matchedDimension:
                    errorMsg = (
                        "Cannot merge negative-area component {solute} into {solvent} without the two being linked."
                    )
```

john-science: I just switched from format to f-string here.

I want to do more of this in ARMI.

armi/reactor/converters/blockConverters.py
```diff
@@ -132,9 +151,13 @@ def _checkInputs(self, soluteName, solventName, solute, solvent):
                 "Components are not of compatible shape to be merged solute: {}, solvent: {}".format(solute, solvent)
             )
         if solute.getArea() < 0:
-            raise ValueError(
-                "Cannot merge solute with negative area into a solvent. {} area: {}".format(solute, solute.getArea())
-            )
+            # allow negative-area gap / void
+            if not solute.hasFlags(Flags.GAP):
```
john-science: Sorry, did you mean to check for `Flags.GAP` OR `Flags.VOID`?

armi/reactor/converters/blockConverters.py
```diff
@@ -462,6 +485,22 @@ class HexComponentsToCylConverter(BlockAvgToCylConverter):
     and ``driverFuelBlock`` must be instances of HexBlocks.
     """
 
+    PIN_COMPONENT_FLAGS = (
```
john-science, line 488: @drewj-tp Do you see my concern with not having a `FUEL_PIN` flag or something else to simplify this logic?

Everyone is forced to generate their own logical for these things. It seems like we're risking inconsistency here.

Should ARMI have a function `isFuelPin()`, so at least we're all consistent?

armi/reactor/converters/blockConverters.py
```diff
@@ -462,6 +485,22 @@ class HexComponentsToCylConverter(BlockAvgToCylConverter):
...
+        Flags.FUEL,
+        Flags.ANNULAR | Flags.VOID,
+        Flags.GAP,
+        Flags.BOND,
+        Flags.LINER,
+        Flags.CLAD,
+        Flags.WIRE,
+        Flags.CONTROL,
+        Flags.REFLECTOR,
+        Flags.SHIELD,
```
john-science, line 498: Sorry, @mgjarrett , surely we aren't saying anything that's marked as a `SHIELD` is a pin.

What am I missing?

armi/tests/armiRun.yaml
john-science: `armiRun.yaml` is used in LOTS of tests in ARMI and downstream. I'm pretty surprised this didn't require fixing dozens of tests in ARMI.

But if this didn't break any unit tests in ARMI or downstream, great, I guess. Surprising, but great.

john-science (changes requested):

armi/reactor/converters/blockConverters.py
mgjarrett: I think the check here was originally more general but then I narrowed it down to just `Flags.GAP`. I'll update the comment.

armi/reactor/converters/blockConverters.py
mgjarrett: Any component within a block that has `Flags.SHIELD` is a typically a shield pin.

armi/reactor/converters/blockConverters.py
mgjarrett: `Flags.FUEL_PIN` or `Flags.FUEL | Flags.PIN` doesn't solve the problem here.

This function needs to identify any component within a block that is associated with a pin. These are all of the component flag types that are typically found in pins.

In theory we could add 'Flags.PIN' to every single component (fuel, clad, bond, wire, liner, etc.) that is associated with a pin. It could work, but that would be somewhat cumbersome in the input.

armi/reactor/converters/blockConverters.py
mgjarrett: You could in theory have a `SHIELD` component that is some kind of shape other than a pin. But this `HexComponentsToCylConverter` class would only be used on a block that has pins.

john-science (approved):
Nice work!

PR #2121, opened 2025-04-16
Supporting growing DBto full core on db load

### LINKED ISSUE #2089: Support injecting growToFullCore after DB load
There should be a way to grow to full core after database load for snapshots (and maybe restart runs)

Analysis often need to grow to full core for more detailed analysis. ARMI already has `growToFullCoreAfterLoad` setting but it doesn't do anything in the framework. this should do something, probably within database interface.

Putting this hook at the end of `dbi.loadState` would get both snapshots and restarts since `snashots.py` calls it directly, and restart calls `prepRestartRun` which calls in indirectly.

drewj-tp:

The setting `growToFullCoreAfterLoad` is used to expand the reactor to full core when loading from a database. This supports a workflow where someone needs to perform snapshot analysis on a database that studied a symmetric core, but their specific analyses require operations on the full core. For example, determining the worth a single control rod by moving it independently of all other control rods in the reactor.

Tests are added showing the behavior on `Database.load` but also the `OperatorSnapshots` gets a full core reactor if this setting is active at run time.

Closes #2089 

## SCR Information

Change Type: features

One-Sentence Description: The setting `growToFullCoreAfterLoad` is used to expand the reactor to full core when loading from a database.

One-line Impact on Requirements: Functionality added to I_ARMI_DB_TIME1 the implementation of R_ARMI_DB_TIME

---

## Checklist


john-science:
@drewj-tp If it helps:

> Okay so we don't sort imports that way. Cool

I just use `ruff check . --fix` to sort the imports.

armi/operators/operatorMPI.py
```diff
@@ -81,21 +81,19 @@ def operate(self):
                 )
                 raise
             finally:
-                if context.MPI_SIZE > 0:
+                # If there are other processes, tell them to stop
+                if context.MPI_SIZE > 1:
```
john-science, line 85: So ARMI will no longer work if you run in MPI mode with only one thread?

That'll affect people's testing downstream. That doesn't really feel like a change that is related to this PR. At the very least, we need to test this downstream to make sure no one is doing that kind of unit testing.

armi/operators/tests/test_operatorSnapshots.py
```diff
@@ -100,3 +102,48 @@ def test_getOperatorClassFromSettings(self):
...
+    """Test that a snapshot operator can do full core analysis with a 1/3 core DB."""
+
+    DB_PATH = Path("test_operator_snapshot_full_core_expansion.h5")
+
+    @classmethod
+    def setUpClass(cls):
+        o, cls.symmetricReactor = loadTestReactor(
+            inputFileName="smallestTestReactor/armiRunSmallest.yaml",
+        )
+        reduceTestReactorRings(cls.symmetricReactor, o.cs, maxNumRings=2)
```
john-science, line 117:
```suggestion
```

armi/operators/operatorMPI.py
drewj-tp: I came across this because something didn't work in my local testing on windows w/o mpi without this change. You're correct: it's not related to this PR so I will revert and try to reproduce the original problem

armi/operators/tests/test_operatorSnapshots.py
drewj-tp: Yes that is correct. I'll need to update my testing to use a non-single assembly reactor that is defined with 1/3 symmetry. The `smallestTestReactor/armiRunSmallest.yaml` has full symmetry which makes sense: it's just one assembly.

armi/operators/operatorMPI.py
drewj-tp: Found the problem. In the tests I'm adding, I call `operator.operate` with an `OperatorSnapshots` which inherits from `OperatorMPI`. When I run this w/o mpi4py locally (which is an optional dependency), I get the following error
```
  File "c:\Users\anjohnson\codes\nala\framework\armi\operators\tests\test_operatorSnapshots.py", line 144, in test_fullCoreFromThirdCore
    o.operate()
    ~~~~~~~~~^^
  File "c:\Users\anjohnson\codes\nala\framework\armi\operators\operatorMPI.py", line 88, in operate
    context.MPI_COMM.bcast(
    ^^^^^^^^^^^^^^^^^^^^^^
```

I'll see if I can get around this with instead calling `DatabaseInterface.loadState` which we do in `OperatorSnapshots.operate`. But it feels kind of unsatisfying to not be able to show that, from the highest level, `OperatorSnapshots.operate` can grow to full core based on this setting. We still have the breadcrumbs to show the behavior works, just feels (to me) unsatisfying.

This also implies that no one can do `OperatorSnapshots.operate` without `mpi4py` installed, regardless of if you want to run that in parallel or serial.

I think this change is still fine through. When you run w/ one MPI process, `context.MPI_SIZE` will be one. The logic under that conditional guard is required to shut down the other MPI workers. But, since `context.MPI_SIZE == 1`, we don't have any MPI workers.

armi/operators/tests/test_operatorSnapshots.py
drewj-tp: 57704a6deee254cf3decfa935fd9fa5979be38e6

armi/operators/tests/test_operatorSnapshots.py
```diff
@@ -100,3 +104,50 @@ def test_getOperatorClassFromSettings(self):
...
+
+class TestOperatorSnapshotFullCoreExpansion(unittest.TestCase):
+    """Test that a snapshot operator can do full core analysis with a 1/3 core DB."""
+
+    DB_PATH = Path("test_operator_snapshot_full_core_expansion.h5")
+
+    @classmethod
+    def setUpClass(cls):
+        o, cls.symmetricReactor = loadTestReactor(TEST_ROOT)
+        reduceTestReactorRings(cls.symmetricReactor, o.cs, maxNumRings=2)
```
john-science, line 117:
```suggestion
```

john-science (approved):

PR #2131, opened 2025-04-25
Update docstring for `Settings` class to reflect mutability

keckler:

The docstring for the `Settings` class had a note that was not technically true. This PR fixes the docstring but maintains the intent of the original in terms of discouraging users from changing settings mid-run.

## SCR Information

Change Type: docs

One-Sentence Description: Update the docstring for the `Settings` class to reflect their mutability

One-line Impact on Requirements: None

---

## Checklist

### DIFF
--- a/armi/settings/caseSettings.py
+++ b/armi/settings/caseSettings.py
@@ -61,7 +61,10 @@ class Settings:
 
     Notes
     -----
-    The actual settings in any instance of this class are immutable.
+    While it is possible to modify case settings during the course of a run, this
+    is highly discouraged because there will be no record of this happening in your
+    results or in the database produced from your run. There is no guarantee that
+    doing so will not cause unexpected problems with your calculation.
     """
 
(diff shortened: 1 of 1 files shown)


keckler:
Note, this is already acknowledged back in the original ticket on this topic: https://github.com/terrapower/armi/issues/371#issuecomment-1230710501

john-science (approved):
I was thinking of adding, "ARMI doesn't not guarantee correctness not will ARMI pretend to support parameters changed during a run ". 

Thoughts?

keckler:
> I was thinking of adding, "ARMI doesn't not guarantee correctness not will ARMI pretend to support parameters changed during a run ".
> 
> Thoughts?

I thought my wording kinda covered that already with "There is no guarantee that doing so will not cause unexpected problems with your calculation."

But go ahead and make a suggestion to the code if you want to add something, I am not necessarily against doing so.

PR #2137, opened 2025-05-15
Improving description of rateProdNet parameter

john-science:

The description for the `rateProdNet` parameter was insufficient, and people asked to improve it.

## SCR Information

Change Type: docs

One-Sentence Description: Improving descriptiion of rateProdNet parameter,

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/parameters.py
+++ b/armi/physics/neutronics/parameters.py
@@ -389,7 +389,7 @@ def _getNeutronicsBlockParams():
         pb.defParam(
             "rateProdNet",
             units=f"1/{units.CM}^3/{units.SECONDS}",
-            description="Net production rate of neutrons",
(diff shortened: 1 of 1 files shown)


drewj-tp (approved):

keckler:
Thanks for the change. As a side note, this parameter, especially since we are now defining it as something very specific to DIF3D, does not belong in the framework. It would be good to move this internal, IMO.

john-science:
> Thanks for the change. As a side note, this parameter, especially since we are now defining it as something very specific to DIF3D, does not belong in the framework. It would be good to move this internal, IMO.

This parameter is currently used in the `GlobalFluxInterface` in ARMI.

So, THAT would have to be moved out of ARMI first.

PR #2138, opened 2025-05-19
Allowing the BOL orientations to be set in the blueprints

john-science:

This was a new feature request. In the past, all orientations for high-level BP / Core children start with the same default, because all parameters do.

But there was a design request to be able to start the orientations of each Composite at the start of the run. So here I copy the `grid contents` parameter over to the `orientationBOL` parameter, so you can define the orientation parameter of any high-level Composite on any grid.

## SCR Information

Change Type: features

One-Sentence Description: Allowing the BOL orientations to be set in the blueprints

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/gridBlueprint.py
+++ b/armi/reactor/blueprints/gridBlueprint.py
@@ -297,8 +297,8 @@ class GridBlueprint(yamlize.Object):
         runLog.extra("Creating the spatial grid")
         if geom in (geometry.RZT, geometry.RZ):
             if self.gridBounds is None:
-                # This check is regrettably late. It would be nice if we could validate
-                # that bounds are provided if R-Theta mesh is being used.
+                # This check is regrettably late. It would be nice if we could validate that bounds
+                # are provided if R-Theta mesh is being used.
                 raise InputError(
                     f"Grid bounds must be provided for `{self.name}` to specify a grid with "
                     "r-theta components."
@@ -370,7 +370,6 @@ class GridBlueprint(yamlize.Object):
         else:
             return 6
 
-    # TODO: JOHN: does this need to be tweaked?
     def expandToFull(self):
         """
         Unfold the blueprints to represent full symmetry.
@@ -382,13 +381,14 @@ class GridBlueprint(yamlize.Object):
         for some scenarios, such as when expanding fuel shuffling paths or the like. Future work may
         make this more sophisticated.
         """
+        # if we are already full core, don't throw an error, just return passively
         if (
             geometry.SymmetryType.fromAny(self.symmetry).domain
             == geometry.DomainType.FULL_CORE
         ):
-            # No need!
             return
 
+        # fill the new grid contents
         grid = self.construct()
 
         newContents = copy.copy(self.gridContents)
@@ -398,6 +398,8 @@ class GridBlueprint(yamlize.Object):
                 newContents[idx2] = contents
 
         self.gridContents = newContents
+
+        # set the grid symmetry
         split = geometry.THROUGH_CENTER_ASSEMBLY in self.symmetry
         self.symmetry = str(
             geometry.SymmetryType(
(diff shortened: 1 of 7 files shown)


john-science:
@alexhjames My only question here is:

> In this PR I only set the Assembly orientation, I don't go through and set orientation parameter for each Block inside the Assembly. Is that... okay?

alexhjames:
> @alexhjames My only question here is:
> 
> > In this PR I only set the Assembly orientation, I don't go through and set orientation parameter for each Block inside the Assembly. Is that... okay?

That's fine. Only the assembly orientation is needed.

alexhjames (approved):
Shockingly, I have no comments. This looks like it does what we need and I can't think of any problems.

armi/reactor/blueprints/reactorBlueprint.py
```diff
@@ -200,12 +202,21 @@ def _constructComposites(self, cs, bp, loadComps, system, gridDesign):
...
 
-    def _loadComposites(self, cs, container, gridContents, bp):
+    def _loadComposites(self, cs, bp, container, gridContents, orientationBOL):
         runLog.header(f"=========== Adding Composites to {container} ===========")
         badLocations = set()
         for locationInfo, aTypeID in gridContents.items():
+            # create a new Composite to add to the grid
             newAssembly = bp.constructAssem(cs, specifier=aTypeID)
 
+            # TODO: JOHN: Or this could be in `constructAssem`???
```
keckler, line 212: Note the remaining TODO

armi/reactor/blueprints/gridBlueprint.py
```diff
@@ -179,6 +179,9 @@ class GridBlueprint(yamlize.Object):
     gridContents : dict
         A {(i,j): str} dictionary mapping spatialGrid indices in 2-D to string specifiers of what's
         supposed to be in the grid.
+    orientationBOL : dict
+        A {(i,j): float} dictionary mapping spatialGrid indices in 2-D to the orientation of
+        what's supposed to be in the grid.
```
keckler, line 184: I want to confirm, this new parameter does not actually cause any rotation to take place, right? From looking at the code, my understanding is that the assembly is put in at whatever rotation that it is defined in the blueprints, and this new value is simply a bookkeeping parameter.

For instance, if I have an asymmetric assembly that is defined with a block grid pointed "up" (whatever that means), my understanding is that the reactor will be instantiated with that assembly pointed "up" no matter if `orientationBOL` is 0, 1, 2, 3, etc. The use of `orientationBOL` is simply to say that pointed "up" corresponds to an orientation value of `x`.

Is my understanding correct?

armi/reactor/blueprints/reactorBlueprint.py
john-science: LOL

Yikes!

Thanks!

armi/reactor/blueprints/gridBlueprint.py
john-science: That's a good point. I have adding in rotating the assemblies instead of just setting the `orientation` parameter.

I guess I was hyper-focused on a potential issue with 1/3 core conversions and got blind. Thanks for the assist.

alexhjames (changes requested):
Per @mgjarrett  testing: `gridContents` need to be converted from `(i, j)` to `(ring, pos)` so that the user doesn't have to come up with the indexing in the blueprints.

armi/reactor/blueprints/gridBlueprint.py
john-science: fixed

armi/reactor/blueprints/gridBlueprint.py
```diff
@@ -431,12 +416,9 @@ def _readGridContentsLattice(self):
...
         iOffset = 0
         jOffset = 0
-        if (
-            geom == geometry.GeomType.CARTESIAN
-            and symmetry.domain == geometry.DomainType.FULL_CORE
-        ):
+        if geom == geometry.GeomType.CARTESIAN and symmetry.domain == geometry.DomainType.FULL_CORE:
             # asciimaps is not smart about where the center should be, so we need to offset
-            # appropriately to get (0,0) in the middle
+            # ppropriately to get (0,0) in the middle
```
alexhjames: typo

john-science: fixed

alexhjames (approved):

john-science:
> FYI I do not plan to do another detailed review of this PR, unless you want me to. I gave my feedback without being asked in the first place, so please don't let me hold this up :-)

No worries!  Alex and Michael are reviewing this feature. You're free.

PR #2140, opened 2025-05-19
Cleaning up Block constructor

### LINKED ISSUE #2124: We manually set some parameters in the Block constructor
This makes the code harder to maintain and we don't know why this is necessary:

https://github.com/terrapower/armi/blob/ac9a81c70ecc8b48dd790643b3e9a98b67c098a4/armi/reactor/blocks.py#L104-L121

We should try reverting this change and seeing if any downstream code breaks.

john-science:

The constructor for `Block` manually sets a small set of parameters to zero. But we handle this in the parameter definitions already, by giving them default values. 

So this PR removes this old default-generating logic. I am sure it is just a hold-over from ancient history.

close #2124

## SCR Information

Change Type: trivial

One-Sentence Description: Removing old, defunct code from Block constructor.

---

## Checklist

### DIFF
--- a/armi/reactor/blockParameters.py
+++ b/armi/reactor/blockParameters.py
@@ -363,7 +363,7 @@ def getBlockParameterDefinitions():
             "eqRegion",
             units=units.UNITLESS,
             description="Equilibrium shuffling region. Corresponds to how many full cycles fuel here has gone through.",
-            default=-1,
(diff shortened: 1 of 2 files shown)


john-science:
Bump. This should be a fast one.

opotowsky (approved):
Easy!

PR #2142, opened 2025-05-22
Quieting warnings from Block.getComponent

alexhjames:

Currently, there is a mountain of warnings that are printed to the log from `getComponent` for blocks that don't have wire/clad due to a variety of methods. Those methods all attempt to grab the wire or clad component with `getComponent`, then check if the return value was `None`. The initial call to `getComponent` doesn't use the `quiet` argument though so it prints a warning to log that says it is returning None which the method is already handling. This change adds the quiet flag in select spots to reduce the unnecessary warning output.  

## SCR Information

Change Type: trivial

One-Sentence Description: Reducing warning output from `getComponent`

One-line Impact on Requirements: None

---

## Checklist

- N/A [ ] [Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify any new/changed code.

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -2030,9 +2030,13 @@ class HexBlock(Block):
     def verifyBlockDims(self):
         """Perform some checks on this type of block before it is assembled."""
         try:
-            wireComp = self.getComponent(Flags.WIRE)
+            wireComp = self.getComponent(
+                Flags.WIRE, quiet=True
+            )  # Quiet because None case is checked for below
             ductComps = self.getComponents(Flags.DUCT)
-            cladComp = self.getComponent(Flags.CLAD)
+            cladComp = self.getComponent(
+                Flags.CLAD, quiet=True
+            )  # Quiet because None case is checked for below
         except ValueError:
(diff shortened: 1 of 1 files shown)


john-science:
I don't really want to be in the business of putting `quiet` in the method declaration of every method in ARMI. That would get messy and cumbersome.

Happily, we already have another tool in ARMI that does this. All `runLog` statements can use `single=True`, so the log statement only gets printed once. And we use this widely throughout ARMI:

https://github.com/terrapower/armi/blob/e4d34d37ae7cf32585f2125a5163844616bb446d/armi/reactor/cores.py#L425-L429

https://github.com/terrapower/armi/blob/e4d34d37ae7cf32585f2125a5163844616bb446d/armi/reactor/blocks.py#L792-L795

https://github.com/terrapower/armi/blob/e4d34d37ae7cf32585f2125a5163844616bb446d/armi/reactor/blocks.py#L2006-L2009

I think this is a better approach, because it only happens AT the log message, and doesn't clutter up the ARMI API. 

Worst case scenario, you need to tweak the log statement a little bit, so it is fully duplicate.

Thoughts?!?

PR #2144, opened 2025-05-23
Improving the default value for Assembly.getArea()

### LINKED ISSUE #1682: Bad Default: Assembly.getArea() is 1.0
Over on [this PR](https://github.com/terrapower/armi/pull/1670), I noted a strange point in the ARMI API.

When an `Assembly` has no children/`Blocks`, this method returns a "default area" of `1.0`.  Surely that should be an error? Or it should return `None`. Or it should return `0.0`. Returning a valid numerical value just seems crazy.

john-science:

ARMI had an insane default that I am worried could cause non-physical results downstream if people are not careful.

(As a side note, I ran ALL the downstream testing, and it all passed fine. So perhaps this PR is just paranoia. I still think it's worth doing; I don't want people HAVE to be extra careful to use ARMI safely.)

close #1682 

## SCR Information

Change Type: trivial

One-Sentence Description: Improving the default value for Assembly.getArea()

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_assemblies.py
+++ b/armi/reactor/tests/test_assemblies.py
@@ -399,7 +399,7 @@ class Assembly_TestCase(unittest.TestCase):
         """Tests area calculation for hex assembly."""
         # Default case: for assemblies with no blocks
         a = HexAssembly("TestAssem", assemNum=10)
-        self.assertEqual(a.getArea(), 1)
+        self.assertEqual(a.getArea(), None)
 
         # more realistic case: a hex block/assembly
         cur = self.assembly.getArea()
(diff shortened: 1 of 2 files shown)


armi/reactor/tests/test_assemblies.py
```diff
@@ -399,7 +399,7 @@ def test_getArea(self):
         """Tests area calculation for hex assembly."""
         # Default case: for assemblies with no blocks
         a = HexAssembly("TestAssem", assemNum=10)
-        self.assertEqual(a.getArea(), 1)
+        self.assertEqual(a.getArea(), None)
```
albeanth, line 402:
```suggestion
        self.assertIsNone(a.getArea())
```

albeanth (approved):

opotowsky (approved):

PR #2145, opened 2025-05-23
Make axial linking aware of block grids for axial expansion

albeanth:

This is the replacement for #1376. 

What: This change enables the axial linking class supporting axial expansion to be aware of the grid blocks. This enables users to have assembly definitions that contain multiple pin-components (e.g., mixing fuel and reflector pins). 

Why: Increases the capabilities of ARMI by allowing more assembly types to be modeled.

## SCR Information

Change Type: features

One-Sentence Description: Enable multi-pin component support for axial expansion.

One-line Impact on Requirements: N/A

---

## Checklist


armi/tests/detailedAxialExpansion/refSmallCoreGrid.yaml
albeanth, line 1: Updated the lattice map to be consistent with the new assembly defs and added the block-grid.

armi/reactor/converters/tests/test_uniformMesh.py
```diff
@@ -84,7 +82,8 @@ def test_makeAssemWithUniformMesh(self):
         )
 
         prevB = None
-        for newB, sourceB in zip(newAssem, sourceAssem):
+        for newB in newAssem:
+            sourceB = sourceAssem.getBlockAtElevation(newB.p.z)
```
albeanth, line 86: This change was needed because with the new blueprints def, the `newAssem` didn't have the same number of blocks as the `sourceAssem`.

armi/reactor/converters/tests/test_axialExpansionChanger.py
albeanth, line 1: I would LOVE to clean this up and split it into multiple files.... But I didn't want to make the changes _so_ drastic is was harder to review than necessary

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -117,8 +152,7 @@ class AxialLink(typing.Generic[Comp]):
     * :attr:`AxialAssemblyLinkage.linkedComponents`
     """
 
-    lower: typing.Optional[Comp] = dataclasses.field(default=None)
-    upper: typing.Optional[Comp] = dataclasses.field(default=None)
```
albeanth, line 121: So turns out, we never use `upper` and having it complicated the new linking. We can bring it back if you feel strongly.

armi/tests/detailedAxialExpansion/refSmallCoreGrid.yaml
```diff
@@ -2,12 +2,34 @@ core:
...
+              AF  MC
+            IC  LA  MC
+              IC  LA
+            PC  IC  MC
+
+fourPin:
+  geom: hex_corners_up
+  symmetry: full
+  lattice map: |
+    - - - - - - - - 1 1 1 1 1 1 1 1 1
```
john-science, line 19: This seems like a really good time to reduce the size of the axial expansion test reactor.

Surely we could test this feature with fewer rings?

albeanth: reduce the size of the core map or this block grid?

john-science: I mean... both in an ideal world.

These are unit tests, they should be fast and small.

But doing either would make the tests faster, so "perfect is the enemy of better", I'll take either.

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -48,40 +41,83 @@ def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
 
     Notes
     -----
-    - Requires that shapes have the getCircleInnerDiameter and getBoundingCircleOuterDiameter
-      defined
-    - When component dimensions are retrieved, cold=True to ensure that dimensions are evaluated
-      at cold/input temperatures. At temperature, solid-solid interfaces in ARMI may produce
-      slight overlaps due to thermal expansion. Handling these potential overlaps are out of scope.
+    If componentA and componentB are both solids and the same type, geometric overlap can be checked via
+    getCircleInnerDiameter and getBoundingCircleOuterDiameter. Five different cases are accounted for.
```
keckler, line 45:
```suggestion
    getCircleInnerDiameter and getBoundingCircleOuterDiameter. Four different cases are accounted for.
```

armi/reactor/assemblies.py
```diff
@@ -262,15 +262,14 @@ def getLocation(self):
...
+        """Return the location of the assembly in the plane using cartesian global coordinates.
 
         .. impl:: Assembly coordinates are retrievable.
             :id: I_ARMI_ASSEM_POSI1
             :implements: R_ARMI_ASSEM_POSI
 
-            In this method, the spatialLocator of an Assembly is leveraged to return
-            its physical (x,y) coordinates in cm.
+            In this method, the spatialLocator of an Assembly is leveraged to return its physical
+            (x,y) coordinates in cm.
```
drewj-tp: A good number of lines changed due to a line length discrepancy in some formatter. Please check with @john-science that these are acceptable

armi/reactor/assemblies.py
```diff
@@ -285,34 +284,36 @@ def getArea(self):
...
-            cladId = b.getComponent(Flags.CLAD).getDimension("id")
             length = b.getHeight()
-            plenumVolume += (
-                math.pi * (cladId / 2.0) ** 2.0 * length * 1e-6
-            )  # convert cm^3 to m^3
+            for c in b.getComponents(Flags.CLAD):
+                cladId = c.getDimension("id")
+                # convert vol from cm^3 to m^3
+                plenumVolume += math.pi * (cladId / 2.0) ** 2.0 * length * 1e-6
         return plenumVolume
```
drewj-tp: Micro-opt: increment the squared radii for each clad component and then scale by `pi * length * 1e-6` at the end

armi/reactor/assemblies.py
```diff
@@ -285,34 +284,36 @@ def getArea(self):
...
+        float: Total plenum volume for an assembly.
         """
         plenumVolume = 0.0
         for b in self.iterChildrenWithFlags(Flags.PLENUM):
-            cladId = b.getComponent(Flags.CLAD).getDimension("id")
             length = b.getHeight()
-            plenumVolume += (
-                math.pi * (cladId / 2.0) ** 2.0 * length * 1e-6
-            )  # convert cm^3 to m^3
+            for c in b.getComponents(Flags.CLAD):
```
drewj-tp: We don't need the list of things returned by `b.getComponents`
```suggestion
            for c in b.iterChildrenWithFlags(Flags.CLAD):
```

armi/reactor/blocks.py
drewj-tp, line 1: All the changes here seem to be related to https://github.com/terrapower/armi/pull/2142

Does the base branch need to be updated to track main?

armi/reactor/components/component.py
drewj-tp, line 1: All the changes here seem to be `f-string` related changes. Possibly from https://github.com/terrapower/armi/pull/2146 // https://github.com/terrapower/armi/pull/2141

Can you update the base branch?

armi/reactor/composites.py
drewj-tp, line 1: Most the changes here seem to be from https://github.com/terrapower/armi/pull/2146 // https://github.com/terrapower/armi/pull/2141

Can you update the base branch?

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -48,40 +41,83 @@ def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
...
             runLog.warning(
                 f"Components {componentA} and {componentB} are UnshapedComponents "
-                "and do not have 'getCircleInnerDiameter' or getBoundingCircleOuterDiameter "
-                "methods; nor is it physical to do so. Instead of crashing and raising an error, "
+                "and do not have 'getCircleInnerDiameter' or getBoundingCircleOuterDiameter methods; "
+                "nor is it physical to do so. Instead of crashing and raising an error, "
                 "they are going to be assumed to not be linked.",
                 single=True,
             )
-            return False
```
drewj-tp, line 75: Why not return false here? Why do we need to save the `linked` variable? We don't seem to use it in this method except to return it later

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -48,40 +41,84 @@ def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
...
-        odB = componentB.getBoundingCircleOuterDiameter(cold=True)
-        biggerID = max(idA, idB)
-        smallerOD = min(odA, odB)
-        return biggerID < smallerOD
-    return False
+            linked = False
+        elif isinstance(componentA.spatialLocator, MultiIndexLocation) and isinstance(
+            componentB.spatialLocator, MultiIndexLocation
+        ):
+            ## Case 2
```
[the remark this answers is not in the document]
drewj-tp: Could we use a `Set[tuple[int, int]]` where each tuple is the `ij` location for that index? Then we can compare the sets directly. Maybe
```python
fromA = set((index.i, index.j) for index in componentA.spatialLocator.indices)
fromB = set((index.i, index.j) for index in componentB.spatialLocator.indices)
if fromA == fromB:
    linked = _checkOverlap(componentA, componentB)
else:
    linked = False
```

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -48,40 +41,83 @@ def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
...
+            if len(componentAIndices) < len(componentBIndices):
+                if all(index in componentBIndices for index in componentAIndices):
+                    linked = _checkOverlap(componentA, componentB)
+                else:
+                    linked = False
+            else:
+                if all(index in componentAIndices for index in componentBIndices):
+                    linked = _checkOverlap(componentA, componentB)
+                else:
+                    linked = False
```
drewj-tp: This, as its written, would allow for the components to have different multiplicities since we're only checking a subset of indices. If

- `componentAIndices = [[a, b, c]]`
- `componentBIndices = [[a, b, c], [x, y, z]]`

Then we would go into the `_checkOverlap` check because every index in A is found in the indices of B.

By the docstring

>     Case #2: Blocks with specified grids. If componentA and componentB share common grid indices (cannot be a partial
>     case, ALL of the indices must be contained by one or the other), then overlap can be checked.

It's not _super_ clear if we support one component having a subset of the indices. They would still "share common grid indices" and "ALL of the indices" from one would be contained in the other. And since Case 3 requires their multiplicity to be identical, I get the sense we do want a strict equality of the indices.

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -48,40 +41,83 @@ def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
...
-      defined
-    - When component dimensions are retrieved, cold=True to ensure that dimensions are evaluated
-      at cold/input temperatures. At temperature, solid-solid interfaces in ARMI may produce
-      slight overlaps due to thermal expansion. Handling these potential overlaps are out of scope.
+    If componentA and componentB are both solids and the same type, geometric overlap can be checked via
+    getCircleInnerDiameter and getBoundingCircleOuterDiameter. Four different cases are accounted for.
+    If they do not meet these initial criteria, linkage is assumed to be False.
+    Case #1: Unshaped Components. There is no way to determine overlap so they're assumed to be not linked.
+    Case #2: Blocks with specified grids. If componentA and componentB share common grid indices (cannot be a partial
+    case, ALL of the indices must be contained by one or the other), then overlap can be checked.
```
drewj-tp: Stricter and more explicit
```suggestion
    Case #2: Blocks with specified grids. If componentA and componentB have identical grid indices (cannot be a partial
    case, ALL of the indices must be contained by one or the other), then overlap can be checked.
```

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -48,40 +41,83 @@ def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
...
+        ):
+            ## Case 2
+            componentAIndices = [
+                list(index) for index in componentA.spatialLocator.indices
+            ]
+            componentBIndices = [
+                list(index) for index in componentB.spatialLocator.indices
+            ]
+            # check for common indices between components. If either component has indices within its counterpart,
+            # then they are candidates to be linked and overlap should be checked.
```
drewj-tp: Comment here implies it's okay to have agreement on some but not necessarily all. So long as all the indices in one are found in the other, it's okay. But that leads to the problem of different multiplicities I bring up in other comments

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -48,40 +41,83 @@ def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
...
+def _checkOverlap(componentA: Component, componentB: Component) -> bool:
+    """Check two components for geometric overlap.
+
+    Notes
+    -----
+    When component dimensions are retrieved, cold=True to ensure that dimensions are evaluated
+    at cold/input temperatures. At temperature, solid-solid interfaces in ARMI may produce
+    slight overlaps due to thermal expansion. Handling these potential overlaps are out of scope.
+    """
+    # Check if one component could fit within the other
```
drewj-tp: Comment is implied by the function docstring and/or could be placed as the main docstring signature
```suggestion
```

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
drewj-tp: If we weren't using it, probably seems we don't need it. I guess you would want to use it to understand what this component is pushing into and/or sending information into. But we still understand the overall linking based on every component knowing what's below it. Kind of list a singly linked list vs. doubly linked list

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
```diff
@@ -263,7 +265,28 @@ def setAssembly(self, a: Assembly, setFuel=True, expandFromTinputToThot=False):
...
+        Makes sure that there aren't any blocks (other than the top-most dummy block)
+        that are entirely fluid filled, unless all blocks in the assembly are only
+        fluids. The expansion changer doesn't know what to do with such mixed assemblies.
+        """
+        for b in self.linked.a[:-1]:
+            # the topmost block has already been confirmed as the dummy block
+            solidCompsInBlock = list(
+                filter(lambda c: not isinstance(c.material, Fluid), b.iterComponents())
+            )
+            if len(solidCompsInBlock) == 0:
```
drewj-tp: Maybe
```python
areAllFluid = all(isinstance(c.material, Fluid) for c in b.iterComponents())
if areAllFluid:
    raise ...
```

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
```diff
@@ -263,7 +265,28 @@ def setAssembly(self, a: Assembly, setFuel=True, expandFromTinputToThot=False):
...
+
+    def _checkAssemblyConstructionIsValid(self):
         self._isTopDummyBlockPresent()
+        self._checkForBlocksWithoutSolids()
+
+    def _checkForBlocksWithoutSolids(self):
+        """
+        Makes sure that there aren't any blocks (other than the top-most dummy block)
+        that are entirely fluid filled, unless all blocks in the assembly are only
+        fluids. The expansion changer doesn't know what to do with such mixed assemblies.
```
drewj-tp: The implementation here does not check if all blocks in the assembly are only fluid, just that all blocks have at least one non-fluid component

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
```diff
@@ -263,7 +265,28 @@ def setAssembly(self, a: Assembly, setFuel=True, expandFromTinputToThot=False):
...
+        """
+        for b in self.linked.a[:-1]:
+            # the topmost block has already been confirmed as the dummy block
+            solidCompsInBlock = list(
+                filter(lambda c: not isinstance(c.material, Fluid), b.iterComponents())
+            )
+            if len(solidCompsInBlock) == 0:
+                raise InputError(
+                    f"Assembly {self.linked.a} is constructed improperly for use with the axial expansion changer.\n"
+                    "Consider using the assemFlagsToSkipAxialExpansion case setting."
```
drewj-tp: Can we include something that tells the user why this assembly failed? Something to the effect of "this block was found to not have solids"

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -110,20 +109,17 @@ def setExpansionFactors(self, components: list["Component"], expFrac: list[float
             If components and expFrac are different lengths
         """
         if len(components) != len(expFrac):
-            runLog.error(
+            raise RuntimeError(
                 "Number of components and expansion fractions must be the same!\n"
                 f"     len(components) = {len(components)}\n"
                 f"        len(expFrac) = {len(expFrac)}"
             )
-            raise RuntimeError
```
drewj-tp, line 118: Yeah removing an empty `RuntimeError` is a good call. 

Non-blocking but I think the `runLog.error` is still useful. Maybe a shared message string goes to both?
```python
message = "Number of components ..."
runLog.error(message)
raise RuntimeError(message)
```
We have a mixed bag of guidance on this (runLog with message then error, or just error) so I'm okay having just the exception with the error. 

Related to https://github.com/terrapower/armi/discussions/2056

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -239,18 +234,21 @@ def _setTargetComponents(self, setFuel: bool):
...
-            elif b.hasFlags(Flags.DUMMY):
-                # Dummy blocks are intended to contain only fluid and do not need a target component
-                pass
-            elif setFuel and b.hasFlags(Flags.FUEL):
-                self._isFuelLocked(b)
-            else:
-                self.determineTargetComponent(b)
+            self.setTargetComponent(b, setFuel)
+
+    def setTargetComponent(self, b: "Block", setFuel: bool):
```
drewj-tp, line 237: Methods, especially public methods, should have docstrings

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+class TestAreAxiallyLinked(AxialExpansionTestBase):
+    """Provide test coverage for the different cases in assemblyAxialLinkage.areAxiallyLinked."""
+
+    def test_mismatchComponentType(self):
+        """Case 4; component type mismatch."""
+        compDims = ("test", "FakeMat", 25.0, 25.0)  # name, material, Tinput, Thot
+        comp1 = Circle(*compDims, od=1.0, id=0.0)
+        comp2 = UnshapedComponent(*compDims, area=1.0)
```
drewj-tp: Since we do special logic for unshaped components, I would recommend making the second component a hexagon or something. Just so we don't have any overlap (pun intended) with case 1: unshaped components

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
```diff
@@ -0,0 +1,332 @@
...
+            lowerFuels = lowerB.getComponents(Flags.FUEL)
+            upperFuels = upperB.getComponents(Flags.FUEL)
+            # same grid locs, are linked
+            self.assertTrue(
+                AssemblyAxialLinkage.areAxiallyLinked(lowerFuels[0], upperFuels[0])
+            )
+            # different grid locs, are not linked
+            self.assertFalse(
+                AssemblyAxialLinkage.areAxiallyLinked(lowerFuels[1], upperFuels[0])
+            )
```
drewj-tp: Readability improvement to understand how these relate to the components in the blueprints
```suggestion
            lowerFuel1, lowerFuel2 = lowerB.getComponents(Flags.FUEL)
            upperfuel1, upperFuel2 = upperB.getComponents(Flags.FUEL)
            # same grid locs, are linked
            self.assertTrue(
                AssemblyAxialLinkage.areAxiallyLinked(lowerFuel1, upperFuel1)
            )
            # different grid locs, are not linked
            self.assertFalse(
                AssemblyAxialLinkage.areAxiallyLinked(lowerFuel1, upperFuel2)
            )
```

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
```diff
@@ -0,0 +1,332 @@
...
+        compDims = ("test", "FakeMat", 25.0, 25.0)
+        comp1 = Circle(*compDims, od=1.0, id=0.0)
+        comp2 = Circle(*compDims, od=1.0, id=0.0)
+        # mult are same, comp1 and comp2 are linked
+        self.assertTrue(AssemblyAxialLinkage.areAxiallyLinked(comp1, comp2))
+        # mult is different, now they are not linked
+        comp2.p.mult = 2
+        self.assertFalse(AssemblyAxialLinkage.areAxiallyLinked(comp1, comp2))
+
+    def test_multiIndexLocation(self):
```
drewj-tp, line 199: Request for a test that shows either

1. Locations *can* be a partial match, or
2. Locations *cannot* be a partial match

I'm inclined to think we want the latter (exact match in locations) but other comments/parts of the code have me unsure

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
```diff
@@ -0,0 +1,332 @@
...
+        for method, dims in componentsToTest.items():
+            typeA = method(*self.common, **dims[0])
+            typeB = method(*self.common, **dims[1])
+            msg = f"{self._testMethodName} failed for component type {str(method)}!"
+            if assertionBool:
+                self.assertTrue(_checkOverlap(typeA, typeB), msg=msg)
+                self.assertTrue(_checkOverlap(typeB, typeA), msg=msg)
+            else:
+                self.assertFalse(_checkOverlap(typeA, typeB), msg=msg)
+                self.assertFalse(_checkOverlap(typeB, typeA), msg=msg)
```
drewj-tp: What if we made the assertion function an argument to the test? e.g., `self.runTest(..., self.assertTrue)`? Non-blocking discussion not _really_ a request but it would collapse six lines to two without sacrificing readability in the later tests

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
```diff
@@ -0,0 +1,332 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        """Contains common dimensions for all component class types."""
+        super().setUp(cls)
+        cls.common = ("test", "FakeMat", 25.0, 25.0)  # name, material, Tinput, Thot
+
+    def runTest(
+        self,
+        componentsToTest: dict,
```
drewj-tp: (sarcastic) come on, we can do better w/ the types!

I think `dict[Type[Component], dict[str, float]]` would be what we want

armi/reactor/assemblies.py
albeanth: met ya halfway. I thought keeping the volume calc together (pi r^2 h) was nice for readability but did pull the unit conversion out to the end
 https://github.com/terrapower/armi/pull/2145/commits/386b8df0f3e79bb71083242af183d86fade146fa

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
albeanth: https://github.com/terrapower/armi/pull/2145/commits/e40089a64db3956be3dc9116538b7621a6092f2e don't need it, good catch

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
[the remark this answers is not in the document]
albeanth: https://github.com/terrapower/armi/pull/2145/commits/e40089a64db3956be3dc9116538b7621a6092f2e
much simpler actually matches what is intended -- all indices must mach. I should have looked closer here! I must have been trying to get fancy when I originally wrote this

armi/reactor/converters/axialExpansionChanger/expansionData.py
albeanth: I switched everything to be just the exception with the error.... The error message _should_ get printed in the stdout though. The stderr will include it again, along with the traceback. 

If we decide to go the direction prescribed in #2056, I'll happily reverse course here.

armi/reactor/converters/axialExpansionChanger/expansionData.py
albeanth: https://github.com/terrapower/armi/pull/2145/commits/bc098455a1f89b2296c4ccd168bed2e4f40adfc6

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
albeanth: 0080e099e81d816ade913e7a45b5412db8cbd917 good call

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
albeanth: you are correct, we want the latter. They have to match exactly.

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
albeanth: 💯 577a88aeeab8d5ec45713d48e432a3dce04d5881

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
albeanth: 12418952c845ac285dd7f4a089ab300fce29040a agreed totally

armi/reactor/blocks.py
albeanth: should be good now

armi/reactor/components/component.py
albeanth: should be good now

armi/reactor/composites.py
albeanth: should be good now

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
albeanth: ~8536793321d6a08cce3a0fa11a1dbe7ee55eac5f~ 11ee728636c1f47b0840ec485890cf74e5cb827a

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
albeanth: 11ee728636c1f47b0840ec485890cf74e5cb827a

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
albeanth: 11ee728636c1f47b0840ec485890cf74e5cb827a

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
albeanth: this was updated, should be ok now

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
albeanth: >  I get the sense we do want a strict equality of the indices.

Correct. This has been corrected and should be ok now

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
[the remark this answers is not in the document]
albeanth: sooooo I was actually mistaken. we do not want to always require a perfect match. I will make updates and ping you.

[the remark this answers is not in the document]
drewj-tp: I'm very curious to find out why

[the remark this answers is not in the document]
albeanth: Below is a snippet from the docs addition. If we want to keep this ability, then we need to be able to query for a partial match. 
image

Put another way
image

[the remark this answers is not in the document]
albeanth: Conversely, we could not allow this and require exact matches. This might better represent reality as the pin stacks for fuel 1 and fuel 2 are likely different designs, so it makes less sense to say they share the same shield pin component.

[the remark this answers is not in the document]
drewj-tp: Hmmm okay, I would want to bring in people who'd be making the inputs. But I feel like a partial match is going to be more trouble than it's worth.

Especially as we look at updating things between linked components. That feels more prone to stumbling than if we enforce 1:1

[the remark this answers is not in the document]
albeanth: I chatted with @mgjarrett and we will enforce 1:1 and split the shield component into two different components.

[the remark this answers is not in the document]
albeanth: https://github.com/terrapower/armi/pull/2145/commits/e84c0182d1851788b306fc51782f595ab2eff1f7

armi/tests/detailedAxialExpansion/refSmallCoreGrid.yaml
albeanth: @john-science https://github.com/terrapower/armi/pull/2145/commits/a27f0259318b6f7de5539f226b865bd87faacfab has a _much_ smaller pin lattice that works

albeanth: @john-science https://github.com/terrapower/armi/pull/2145/commits/3e6c1d5feac0835f313e91c74e7e73c4de581415 has as small of an assembly lattice as I think we're going to get!

albeanth:
@drewj-tp GH won't let me re-request your review, so I am doing it here. It's ready!

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -72,29 +71,14 @@ def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
...
-                    linked = False
-            else:
-                if all(index in componentAIndices for index in componentBIndices):
-                    linked = _checkOverlap(componentA, componentB)
-                else:
-                    linked = False
+            fromA = list(list(index) for index in componentA.spatialLocator.indices)
+            fromB = list(list(index) for index in componentB.spatialLocator.indices)
+            if all(index in fromB for index in fromA):
+                linked = _checkOverlap(componentA, componentB)
```
drewj-tp: This is still only checking a subset: only checking everything in `fromA` is found in `fromB`. There could still be items in `fromB` that don't exist in `fromA`

What about
```suggestion
            fromA = set(tuple(index) for index in componentA.spatialLocator.indices)
            fromB = set(tuple(index) for index in componentB.spatialLocator.indices)
            if fromA == fromB:
                linked = _checkOverlap(componentA, componentB)
```
This would throw away duplicates. But if you have a component that has duplicated indices, I think there are other problems...

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -236,10 +219,10 @@ def _findComponentLinkedTo(self, c: Component, otherBlock: typing.Optional[Block
...
             else:
                 errMsg = (
-                    "Multiple component axial linkages have been found for "
-                    f"Component {c} in Block {c.parent} in Assembly {c.parent.parent}. "
-                    "This is indicative of an error in the blueprints! Linked "
-                    f"components found are {candidate} and {otherComp} in {otherBlock}"
+                    "Multiple component axial linkages have been found for the following component!\n"
+                    f"\tComponent {c}\n\t  -> Block {c.parent}\n\t  -> Assembly {c.parent.parent}.\n"
+                    f"This is indicative of an error in the blueprints! Candidate components in {otherBlock}:\n"
+                    f"\t{candidate}\n\t{otherComp}"
```
drewj-tp: At this point, since there's a lot of `\n` and `\t`, can we just write a triple string?
```python
errMsg = f"""Mulitple component axial linkages have been found for:
    Component {c}
    -> Block {c.parent}
...
"""
```
Then we have a much better understanding of how this error message will look when it's logged

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
```diff
@@ -271,18 +269,17 @@ def _checkAssemblyConstructionIsValid(self):
     def _checkForBlocksWithoutSolids(self):
         """
         Makes sure that there aren't any blocks (other than the top-most dummy block)
-        that are entirely fluid filled, unless all blocks in the assembly are only
-        fluids. The expansion changer doesn't know what to do with such mixed assemblies.
+        that consist of entirely fluid components. The expansion changer doesn't know
```
drewj-tp:
```suggestion
        that consist entirely of fluid components. The expansion changer doesn't know
```

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -223,28 +221,41 @@ def getExpansionFactor(self, c: "Component"):
         value = self._expansionFactors.get(c, 1.0)
         return value
 
-    def _setTargetComponents(self, setFuel: bool):
-        """Sets target component for each block.
+    def _setAllTargetComponents(self, setFuel: bool):
+        """Sets axial expansion target component on each block in self._a.
```
drewj-tp: Feels weird to reference a protected variable in a docstring. Albeit for a protected method.
```suggestion
        """Sets axial expansion target component on each block in the expanded assembly.
```
idk not really better and don't want to bike shed this 🤷

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -297,8 +308,8 @@ def determineTargetComponent(self, b: "Block", flagOfInterest: Optional[Flags] =
             raise RuntimeError(f"No target component found!\n   Block {b}")
         if len(candidates) > 1:
             raise RuntimeError(
-                "Cannot have more than one component within a block that has the target flag!"
-                f"Block {b}\nflagOfInterest {flagOfInterest}\nComponents {candidates}"
+                "Cannot have more than one component within a block that has the target flag!\n"
+                f"Block {b}\n\tflagOfInterest {flagOfInterest}\n\tComponents {candidates}"
```
drewj-tp: Same soft request to use a triple string to capture the desired `\n` and `\t` behavior

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
drewj-tp: Okay. Please add a test demonstrating this behavior. Not saying it doesn't work now, but we should have a test to confirm the expectations we need from the implementation

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
```diff
@@ -0,0 +1,322 @@
...
+        """Test for multiple component axial linkage."""
+        linked = AssemblyAxialLinkage(buildTestAssemblyWithFakeMaterial("FakeMat"))
+        b = linked.a.getFirstBlockByType("fuel")
+        c = b.getComponent(typeSpec=Flags.FUEL)
+        c.setDimension("od", 0.785, cold=True)
+        with self.assertRaisesRegex(
+            RuntimeError,
+            expected_regex="Multiple component axial linkages have been found for ",
+        ):
+            linked._getLinkedComponents(b, c)
```
drewj-tp: I don't understand why this should fail by being linked to multiple components. There's just the two fuel pins in the block, one that doesn't have the same locations. So, it should just link to itself.

drewj-tp: Ooooo the assembly is from a different file nvm.

Yeah, it would link against itself and the clad

https://github.com/terrapower/armi/blob/14392c84b872477ced67ad833234dbe866d7ad55/armi/reactor/converters/tests/test_axialExpansionChanger.py#L1225-L1226

Soft request to grab the dimension with something like
```python
clad = b.getComponent(Flags.CLAD)
c.setDimension("od", 0.5 * (clad.getDimension("id") + clad.getDimension("od")))
```
Then it's clearer why the fuel is linked to two things. Non-blocking request

armi/reactor/converters/tests/test_axialExpansionChanger.py
drewj-tp: > I did split out axial linking testing into its own file. Was easier to manage

I missed this comment and that would have changed some of my earlier comments like type hints on the assertion helper 😬

armi/reactor/converters/tests/test_axialExpansionChanger.py
```diff
@@ -660,42 +643,50 @@ def test_isTopDummyBlockPresent(self):
...
+            self.obj.expansionData.updateComponentTempsBy1DTempField(tempGrid, tempField)
 
     def test_AssemblyAxialExpansionException(self):
         """Test that negative height exception is caught."""
         # manually set axial exp target component for code coverage
         self.a[0].p.axialExpTargetComponent = self.a[0][0].name
         temp = Temperature(self.a.getTotalHeight(), numTempGridPts=11, tempSteps=10)
-        with self.assertRaisesRegex(ArithmeticError, "has a negative height!") as cm:
+        with self.assertRaisesRegex(ArithmeticError, "has a negative height!"):
             for idt in range(temp.tempSteps):
```
drewj-tp: While we're here, we should flip the order here. The way this is currently written, so long as any `idt` raises the error, this check will pass. Some of them may not raise the error and that will be silently missed

doc/user/radial_and_axial_expansion.rst
```diff
@@ -314,24 +314,27 @@ properly linking, ``block_fuel_axial_shield`` needs to be redefined with the Blo
...
 
-The following example builds upon the above and provides a case in which the Component axial linking implementation will
-interpret multiple Component linking and raise a ``RuntimeError``. In this example, the axial shield and fuel block grids
-are well defined and link properly. However, though ``block_plenum_multiPin`` uses a Block-grid,
-``clad 1`` will try to link to both ``fuel 1`` and ``fuel 2`` below it, raising a ``RuntimeError`` (this would also occur
-if a multiplicity-based definition were used for ``clad 1``).
+The following incorporates the fix for ``block_fuel_axial_shield`` and illustrates another situation in which the
+Component axial linking implementation will interpret multiple axial links and raise a ``RuntimeError``. In this
+example, a plenum block is added above the fuel and while it does utilize a Block-grid, ``clad`` will try to link to
+both the ``fuel 1`` and ``fuel 2`` components below it, raising a ``RuntimeError`` (this would also occur if a
```
drewj-tp: I would expect we would get a different problem in this case. The clad would not be linked to anything since it's locations don't match.

The reason we would get a multi-match error is because we are (presently) allowing for a subset of locations in order to check overlap. All the locations of `fuel1` exist in the locations of `clad`, and all the locations of `fuel2` exist in the locations of `clad`. That's why it would match both of them.

Can you revise this section of the documentation accordingly? Yes, axial expansion will have an error in the case of clad having the union of lattice IDs of two fuel components in the same block. But not for the reason listed here.

drewj-tp: Also, please put the `(this would also ...)` sentence as a non-parenthetical aside. It's still valuable information

drewj-tp (changes requested):
It still looks like the implementation allows for a subset of locations in linking. I would recommend to add a test showing this behavior is not supported. We can use the clad with two lattice IDs from the docs.

If that test passes, e.g., shows the clad is not matched to either fuel and neither fuel is matched to the clad, then we don't need any changes. But by my understanding of the current approach, 
```
areAxiallyLinked(fuel1, clad)
```
would pass because all the indices of `clad` are in the locations of `fuel1`. This is not what I think we want

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
albeanth: I added a test to show that the `all(index in from B for index in fromA)` approach works as we intend it to. 

https://github.com/terrapower/armi/pull/2145/commits/7b40e1081f2614288252c7c1474ebb2a7371bbb0

I also switched it over to the `set/tuple` approach since the `fromA == fromB` is more clear from a quick glance. 
https://github.com/terrapower/armi/pull/2145/commits/c1ed201b4515b888e050b348ed638444f9b269b0

armi/reactor/converters/axialExpansionChanger/expansionData.py
albeanth: https://github.com/terrapower/armi/pull/2145/commits/d99d58eee827a3a18a7a5eb743538df17747e61b

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
albeanth: https://github.com/terrapower/armi/pull/2145/commits/d99d58eee827a3a18a7a5eb743538df17747e61b

armi/reactor/converters/tests/test_assemblyAxialLinkage.py
albeanth: https://github.com/terrapower/armi/pull/2145/commits/1425791bc668baa1ab85c10a04a9ad4562168551

armi/reactor/converters/tests/test_axialExpansionChanger.py
albeanth: Not every single tempStep is supposed to create the error. It doesn't actually fail until towards the end of tempSteps. This is an older test that could be updated but I'd rather not do that right now....

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
albeanth: https://github.com/terrapower/armi/pull/2145/commits/af634b37d44ebe5adda2bdb92aaaba10f3068b4d to settle it! nice nice

doc/user/radial_and_axial_expansion.rst
albeanth: https://github.com/terrapower/armi/pull/2145/commits/efa7aa046dcc21c2b26bd2789382b013e2d3924d

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -218,12 +218,15 @@ def _findComponentLinkedTo(self, c: Component, otherBlock: typing.Optional[Block
...
-                )
+                errMsg = f"""
+                    Multiple component axial linkages have been found for the following component!
+                        Component {c}
+                          -> Block {c.parent}
+                          -> Assembly {c.parent.parent}
+                    This is indicative of an error in the blueprints! Candidate components in {otherBlock}:
+                        {candidate}
+                        {otherComp}
+                """
```
drewj-tp, line 230: The de-indenting is relative to the current indentation level. Meaning I think we'll get a lot of leading empty space when we print this message
```python
>>> def f():
...     s = """
...               Hello!
...           Look at me!
...     I am right under the s def
... I'm all the way back at column zero
...     """
...     print(s)
...

>>> f()

              Hello!
          Look at me!
    I am right under the s def
I'm all the way back at column zero
```
Non-blocking request to de-indent this message

```suggestion
                errMsg = f"""
Multiple component axial linkages have been found for the following component!
    Component {c}
        -> Block {c.parent}
        -> Assembly {c.parent.parent}
This is indicative of an error in the blueprints! Candidate components in {otherBlock}:
    {candidate}
    {otherComp}
    """
```

armi/reactor/converters/axialExpansionChanger/expansionData.py
```diff
@@ -307,9 +307,12 @@ def determineTargetComponent(self, b: "Block", flagOfInterest: Optional[Flags] =
...
             raise RuntimeError(f"No target component found!\n   Block {b}")
         if len(candidates) > 1:
-            raise RuntimeError(
-                "Cannot have more than one component within a block that has the target flag!\n"
-                f"Block {b}\n\tflagOfInterest {flagOfInterest}\n\tComponents {candidates}"
+            raise RuntimeError(f"""
+                Cannot have more than one component within a block that has the target flag!
+                "Block {b}
+                    flagOfInterest {flagOfInterest}
+                    Components {candidates}
```
drewj-tp: Similar leading empty space request
```suggestion
Cannot have more than one component within a block that has the target flag!
Block {b}
    flagOfInterest {flagOfInterest}
    Components {candidates}
```

drewj-tp (changes requested):
Super soft edits and then I'm content 💯

armi/reactor/converters/axialExpansionChanger/expansionData.py
albeanth: how about using `dedent`? https://github.com/terrapower/armi/pull/2145/commits/db723a97289e2ebe486d8915359f32e6abd205a1

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
albeanth: how about using `dedent`? https://github.com/terrapower/armi/pull/2145/commits/db723a97289e2ebe486d8915359f32e6abd205a1

drewj-tp (approved):

Issue #1765, opened 2024-07-04
Allow pin pitch in `HexBlock`s to fall back to pitch defined by a grid


keckler:
The pin pitch is gotten by `HexBlock.getPinPitch()` by examining the clad and wire dimensions:
https://github.com/terrapower/armi/blob/8768dec747f288a2abd537c209c08e6c0055d367/armi/reactor/blocks.py#L2420-L2464

This implicitly assumes that every `HexGrid` has a wire wrap. This is not universally true. It also assumes that there is only one type of clad and one type of wire in the block. This is certainly not universally true.

If we look at the parent class method, `Block.getPinPitch()` will actually check if the block has a grid defined, and it will use that to get the pin pitch within the block:
https://github.com/terrapower/armi/blob/8768dec747f288a2abd537c209c08e6c0055d367/armi/reactor/blocks.py#L1270-L1276

This implementation on the parent class allows for the pin pitch to be gathered, independent of the assumptions that I outlined above. It just requires the user to define a grid in the blueprints.

I propose that the `HexBlock` subclass of `getPinPitch()` falls back to the parent method implementation if it fails to find the pin pitch using the current method. So basically just adding another `try/except` clause in there, and then pointing to `super().getPinPitch()`.

jakehader:
Maybe related #1402?

keckler:
> Maybe related #1402?

Yeah, you're definitely correct, I forgot about that older ticket 😄

john-science:
Thanks, Chris!  I see what you mean about the assumed wire wrap, and I like your solution.

I think it will take more time to write sufficient unit tests to prove the solution works, then to actually implement the solution. But that's fine, it still won't be that onerous.

drewj-tp:
See also #252

john-science:
Juuuust waiting for someone to take the ticket. :)

drewj-tp:
Plan is going to be

1. `Block.getPinPitch` remains the same: pull `self.spatialGrid.pitch`
2. `HexBlock.getPinPitch` will not exist, instead falling back to `Block.getPinPitch`
3. But, we need a way for `HexBlock.autoCreateSpatialGrids` to know the pitch. In that case, I will move the current `HexBlock.getPinPitch` logic into a separate method, maybe `HexBlock.inferPinPitch` (?)

john-science:
@drewj-tp I am reviewing my release plan. Do you think you will be able to do this ticket in the next, say, two months?

[...]

Issue #1402, opened 2023-09-05
Pin pitch in `HexBlock`s can be inconsistent with grid definitions


keckler:
`HexBlock` has a `getPinPitch()` method which infers the pin pitch based on the (possibly) temperature-dependent diameters of the clad+wire:
https://github.com/terrapower/armi/blob/360791847227df3f3a337a996ef561e00f846a09/armi/reactor/blocks.py#L2160-L2195

No such specialized method exists for a `CartesianBlock`, and in that case `getPinPitch()` will look to to an underlying grid for definition.

If a user also defines a pin grid for the block, they must also define a pin pitch there using the `lattice pitch` attribute (defaults to 1.0, I think?). Since the defined grid forces the user to input a lattice pitch that is NOT temperature dependent and is not based on any built-in assumptions on what defines the pin pitch, I guess you might think that this is the "real" pin pitch. That would certainly be my expectation, since I would have input that value explicitly.

But regardless, it is likely that the two pin pitches do not agree, in which case the question of which is right must be asked. 

My expectation would be that `getPinPitch()` first checks if a grid has been defined, and if not then tries to find the pitch based on the clad+wire diameters.

For most "real" assembly definitions the difference is probably very small. But the reason that I bring this up is because I'm modeling assemblies that don't always have wires, or for one reason or another the assumptions in `getPinPitch()` are tripped up by my assembly definitions. Basically anytime that an assembly has multiple pin definitions, `HexBlock.getPinPitch()` will fail, and this is a problem that can be easily rectified by pulling the pin pitch from an available grid definition.

keckler:
Related: https://github.com/terrapower/armi/issues/252

drewj-tp:
Plan of attack: https://github.com/terrapower/armi/issues/1765#issuecomment-2462741515

drewj-tp:
This is a duplicate of #1765 that has had more work done / more thoughts written / more problems discovered. Crucially, this [problem of hot/cold and complex blocks](https://github.com/terrapower/armi/issues/1765#issuecomment-2932979425)

PR #2147, opened 2025-06-02
Update the way that block grids are built in HexBlocks

keckler:
## SCR Information

Change Type: fixes

One-Sentence Description: 

This change updates the reactor building process such that grids are assigned to blocks earlier in the process.

One-line Impact on Requirements: 

Requirements are not impacted.

---

## Checklist

### DIFF
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -1239,21 +1239,6 @@ class Assembly(composites.Composite):
         grid = self.parent.spatialGrid
         return grid.overlapsWhichSymmetryLine(self.spatialLocator.getCompleteIndices())
 
-    def orientBlocks(self, parentSpatialGrid):
-        """Add special grids to the blocks inside this Assembly, respecting their orientation.
-
-        Parameters
-        ----------
-        parentSpatialGrid : Grid
-            Spatial Grid of the parent of this Assembly (probably a system-level grid).
-        """
-        for b in self:
-            if b.spatialGrid is None:
-                try:
-                    b.autoCreateSpatialGrids(parentSpatialGrid)
-                except (ValueError, NotImplementedError) as e:
-                    runLog.extra(str(e), single=True)
-
 
 class HexAssembly(Assembly):
     """An assembly that is hexagonal in cross-section."""
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -2177,7 +2177,7 @@ class HexBlock(Block):
             return
 
         grid = grids.HexGrid.fromPitch(
-            self.inferPinPitch(cold=True),
+            self.getPinPitch(cold=True),
             numRings=0,
             armiObject=self,
             cornersUp=cornersUp,
@@ -2220,7 +2220,13 @@ class HexBlock(Block):
         pinCenterFlatToFlat = math.sqrt(3.0) / 2.0 * pinCenterCornerToCorner
         return pinCenterFlatToFlat
 
-    def inferPinPitch(self, cold=False):
+    def hasPinPitch(self):
+        """Return True if the block has enough information to calculate pin pitch."""
+        return (self.getComponent(Flags.CLAD, quiet=True) is not None) and (
+            self.getComponent(Flags.WIRE, quiet=True) is not None
+        )
+
+    def getPinPitch(self, cold=False):
         """
         For a "simple" HexBlock, infer the pin pitch in cm.
 
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -2243,6 +2243,11 @@ class HexBlock(Block):
         -------
         pinPitch : float
             pin pitch in cm
+
+        Notes
+        -----
+        This method assumes a close-packed bundle. If the blueprints specify a grid
+        for this block, the value returned will not necessarily align with that pitch.
         """
         try:
             clad = self.getComponent(
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -2187,7 +2187,7 @@ class HexBlock(Block):
             cornersUp = False
 
         grid = grids.HexGrid.fromPitch(
-            self.getPinPitch(cold=True),
+            self.inferPinPitch(cold=True),
             numRings=0,
             armiObject=self,
             cornersUp=cornersUp,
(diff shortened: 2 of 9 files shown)


armi/reactor/assemblies.py
```diff
@@ -1231,21 +1231,6 @@ def isOnWhichSymmetryLine(self):
...
-        parentSpatialGrid : Grid
-            Spatial Grid of the parent of this Assembly (probably a system-level grid).
-        """
-        for b in self:
-            if b.spatialGrid is None:
-                try:
-                    b.autoCreateSpatialGrids(parentSpatialGrid)
-                except (ValueError, NotImplementedError) as e:
-                    runLog.extra(str(e), single=True)
-
```
keckler, line 1248: I removed this method entirely, because it really doesn't do anything at this point. The creation of spatial grids, accounting for the corners-up vs flats-up of the core, is now performed when the blocks are instantiated from blueprints.

armi/reactor/blocks.py
```diff
@@ -269,28 +269,16 @@ def getSmearDensity(self, cold=True):
...
-            Spatial Grid of the system-level parent of this Assembly that contains this Block.
-
-        Raises
-        ------
-        ValueError
-            If the multiplicities of the block are not only 1 or N or if generated ringNumber leads
-            to more positions than necessary.
         """
-        if self.spatialGrid is None:
-            self.spatialGrid = systemSpatialGrid
```
keckler, line 293: I removed this action in favor of a `pass`. I'm not sure why it ever made sense to give a block the spatial grid of the core.

The `HexBlock` has a more specific implementation of this method, so for those blocks, something real will happen.

armi/reactor/blocks.py
```diff
@@ -2133,7 +2121,7 @@ def getSymmetryFactor(self):
                     return 2.0
         return 1.0
 
-    def autoCreateSpatialGrids(self, systemSpatialGrid=None):
+    def autoCreateSpatialGrids(self, cornersUp=False, *args, **kwargs):
```
keckler, line 2124: In the past, we were passing the entire spatial grid down to this level just so that we can determine if the grid is corners-up. I simplified this method so that it could be more generally callable from places that might not have the system spatial grid.

armi/reactor/blocks.py
```diff
@@ -269,28 +269,16 @@ def getSmearDensity(self, cold=True):
...
-        Parameters
-        ----------
-        systemSpatialGrid : Grid, optional
-            Spatial Grid of the system-level parent of this Assembly that contains this Block.
-
-        Raises
-        ------
-        ValueError
-            If the multiplicities of the block are not only 1 or N or if generated ringNumber leads
-            to more positions than necessary.
```
keckler, line 290: This was never true.

armi/reactor/blocks.py
```diff
@@ -269,28 +269,16 @@ def getSmearDensity(self, cold=True):
...
 
-    def autoCreateSpatialGrids(self, systemSpatialGrid=None):
+    def autoCreateSpatialGrids(self, *args, **kwargs):
         """
         Creates a spatialGrid for a Block.
 
         Blocks do not always have a spatialGrid from Blueprints, but some Blocks can have their
         spatialGrids inferred based on the multiplicity of their components. This would add the
         ability to create a spatialGrid for a Block and give its children the corresponding
         spatialLocators if certain conditions are met.
```
keckler, line 279: This probably needs to be updated.

armi/reactor/blocks.py
```diff
@@ -2235,18 +2220,13 @@ def getPinCenterFlatToFlat(self, cold=False):
...
-    def getPinPitch(self, cold=False):
+    def inferPinPitch(self, cold=False):
         """
-        Get the pin pitch in cm.
+        For a "simple" HexBlock, infer the pin pitch in cm.
 
         Assumes that the pin pitch is defined entirely by contacting cladding tubes and wire wraps.
-        Grid spacers not yet supported.
+        Grid spacers not yet supported. Will only work if the block has a single
+        clad and a single wire component.
```
keckler, line 2236: This change is implementing what was suggested in https://github.com/terrapower/armi/issues/1765#issuecomment-2462741515

I changed the name of this method to `inferPinPitch()`. `HexBlock` no longer has a direct implementation of `getPinPitch()`, so it will fall back to `Block.getPinPitch()`, which looks directly for a spatial grid.

`HexBlock.inferPinPitch()` is now called from within `HexBlock.autoCreateSpatialGrid()`.

The upshot of this is that the behavior is unchanged from before, EXCEPT:

>  When a block has a grid defined in the blueprint, that block should use that grid to define the pitch, and only try to infer the pitch based on the component definitions if a grid does not exist. This was not being done in the past, as highlighted by https://github.com/terrapower/armi/issues/1402 . This PR fixes this inconsistency problem.

In practice, this change is not expected to impact hardly anybody, because not that many people define grids for their blocks, so there was not an opportunity for conflict.

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -265,8 +265,20 @@ def construct(
...
+        else:
+            try:
+                cornersUpCore = (
+                    blueprint.gridDesigns[blueprint.systemDesigns["core"].gridName].geom
+                    == geometry.HEX_CORNERS_UP
+                )
+                b.autoCreateSpatialGrids(cornersUp=not cornersUpCore)
+            except (TypeError, ValueError, NotImplementedError) as e:
+                runLog.extra(str(e), single=True)
+
```
keckler, line 280: Instead of using `orientBlocks()`, this is where we now will add the spatial grid to individual blocks. This ensures that all blocks (to the extent possible) have spatial grids which are aligned with the corners-up vs flats-up state of the core.

armi/reactor/converters/tests/test_blockConverter.py
```diff
@@ -211,7 +211,6 @@ def test_convert(self):
             .core.getAssemblies(Flags.FUEL)[2]
             .getFirstBlock(Flags.FUEL)
         )
-        block.spatialGrid = grids.HexGrid.fromPitch(1.0)
```
keckler, line 214: This is no longer needed.

armi/reactor/converters/tests/test_blockConverter.py
```diff
@@ -255,9 +254,6 @@ def test_convertHexWithFuelDriver(self):
         # add depletable flag to see if it is carried
         control.p.flags |= Flags.DEPLETABLE
 
-        driverBlock.spatialGrid = None
-        block.spatialGrid = grids.HexGrid.fromPitch(1.0)
-
```
keckler, line 260: This is no longer needed.

armi/reactor/tests/test_composites.py
```diff
@@ -41,6 +41,9 @@ class MockBP:
     inactiveNuclides = set()
     elementsToExpand = set()
     customIsotopics = {}
+    gridDesigns = unittest.mock.MagicMock()
+
+    systemDesigns = unittest.mock.MagicMock()
```
keckler, line 46:
```suggestion
    gridDesigns = unittest.mock.MagicMock()
    systemDesigns = unittest.mock.MagicMock()
```

armi/reactor/blocks.py
keckler: Nevermind, I undid the funtionality changes here.

PR #2153, opened 2025-06-11
Hiding duplicate warning messages

### LINKED ISSUE #2152: Warnings are no longer being filtered by `single=True`
A recent PR (https://github.com/terrapower/armi/pull/2135) updated this block of code and removed a `return False` statement that would cause duplicate warning messages to be filtered out of the logs.

https://github.com/terrapower/armi/blob/a91a5da1b6ae77f9cf9fe3c4e4c37a1e01ff1f57/armi/runLog.py#L471-L473

Warnings like the one below that used to be filtered to just a single instance now appear thousands of times in the stdout:

Image

The fix should be simple (add the `return False` back), but the unit testing should also be updated to verify the correct behavior.

john-science:

Duplicate warning messages were not being correctly hidden when they should be, so I fixed that.

close #2152

1. The first thing I did was write the new unit test. It failed, which was correct.
2. Then I fixed the bug in `runLog.py`.
3. THEN I realized our unit tests were using some test version of the `runLog.DeduplicationFilter`, which means the tests were still failing. I believe it will be more correct if `MockRunLogs` uses the `DeduplicationFilter` directly, so I did that.
4. Then I noticed that `DeduplicationFilter.singleMessageCounts` was a `dict`, but it only needed to be a `set`, so I made that change.

## SCR Information

Change Type: fixes

One-Sentence Description: Hiding duplicate warning messages.

---

## Checklist

### DIFF
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -442,7 +442,7 @@ class DeduplicationFilter(logging.Filter):
 
     def __init__(self, *args, **kwargs):
         logging.Filter.__init__(self, *args, **kwargs)
-        self.singleMessageCounts = {}
+        self.singleMessageCounts = set()
         self.warningCounts = {}
 
     def filter(self, record):
@@ -468,9 +468,8 @@ class DeduplicationFilter(logging.Filter):
             # in sub-warning cases, hash the label, for faster lookup
             label = hash(label)
             if label not in self.singleMessageCounts:
-                self.singleMessageCounts[label] = 1  # TODO: Could be a set
+                self.singleMessageCounts.add(label)
             else:
-                self.singleMessageCounts[label] += 1
                 return False
 
         # Handle some special string-mangling we want to do, for multi-line messages
--- a/armi/runLog.py
+++ b/armi/runLog.py
@@ -393,19 +393,19 @@ def raw(msg):
 
 
 def extra(msg, single=False, label=None):
-    LOG.log("extra", msg, single=single)
+    LOG.log("extra", msg, single=single, label=label)
 
 
 def debug(msg, single=False, label=None):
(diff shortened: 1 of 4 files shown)


armi/runLog.py
```diff
@@ -442,35 +442,35 @@ class DeduplicationFilter(logging.Filter):
...
         # Track all warnings, for warning report
         if record.levelno in (logging.WARNING, logging.CRITICAL):
-            # if this is from the custom logger, it will have a "label"
-            label = getattr(record, "label", msg)
-            # if the "label" default is None, it needs to be replaced
-            label = msg if label is None else label
-
-            # add the warning to the tracker
             if label not in self.warningCounts:
                 self.warningCounts[label] = 0
```
mgjarrett:
```suggestion
                self.warningCounts[label] = 1
```
Now that the increment is in an `else`, we should start the count at 1 instead of 0

armi/runLog.py
```diff
@@ -442,35 +442,35 @@ class DeduplicationFilter(logging.Filter):
 
     def __init__(self, *args, **kwargs):
         logging.Filter.__init__(self, *args, **kwargs)
-        self.singleMessageCounts = {}
+        self.singleMessageCounts = set()
```
mgjarrett:
```suggestion
        self.singleMessageLabels = set()
```

Recommend renaming the variable since the occurrences are not being counted anymore.

john-science: Ah!  That's a good point. Sure, I'll fix that.

john-science: Done!

armi/runLog.py
john-science: done!

mgjarrett (approved):

PR #2155, opened 2025-06-13
Cleaning up strange counter line

john-science:

Cleaning out a strange counter increment in the plugin manager. @opotowsky pointed this out. I removed it and ran all the internal testing I could find, and everything passed.

I can only conclude that this is not useful now. Perhaps it was once, in the distant past.

## SCR Information

Change Type: trivial

One-Sentence Description: Cleaning out a strange counter increment in the plugin manager.

---

## Checklist


opotowsky (approved):
This appears only used in `getParameterRenames`, which is only used in the Database class during `_readParams`...I think in addition to the testing, this is safe. Although, no one unregisters plugins downstream, so maybe it's just irrelevant to 99.9% of use cases.

PR #2163, opened 2025-06-24
Alter loop in `gamiso.addDummyNuclidesToLibrary` to use gamma groups

opotowsky:

The code in `gamiso.addDummyNuclidesToLibrary` was looping over the isotxs metadata, which is giving things like `jband` the neutron energy group data structure. When someone wanted to increase the number of gamma energy groups in their simulation beyond 33, the simulation failed. 

So the loop was using the `jband` data structure from isotxs, but it is a different structure for gamiso. This fixes that problem by getting the necessary numbers from the gamiso lib, and making a new dictionary based on that. 

The isotxs metadata for dummy nuclides is still useful. It comes prepopulated with all the useful nuclide metadata that would otherwise be blank if I was starting from scratch with gamiso.

## SCR Information

Change Type: fixes

Alter loop in `gamiso.addDummyNuclidesToLibrary` to use gamma groups instead of neutron groups.

---

## Checklist

### DIFF
--- a/armi/nuclearDataIO/cccc/tests/test_gamiso.py
+++ b/armi/nuclearDataIO/cccc/tests/test_gamiso.py
@@ -19,7 +19,7 @@ import unittest
 from copy import deepcopy
 
 from armi.nuclearDataIO import xsLibraries
-from armi.nuclearDataIO.cccc import gamiso
+from armi.nuclearDataIO.cccc import gamiso, isotxs
 from armi.nuclearDataIO.xsNuclides import XSNuclide
 from armi.utils.directoryChangers import TemporaryDirectoryChanger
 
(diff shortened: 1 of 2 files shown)


armi/nuclearDataIO/cccc/tests/test_gamiso.py
```diff
@@ -69,3 +69,14 @@ def test_addDummyNuclidesToLibrary(self):
...
         self.assertEqual(list(diff)[0].xsId, "38")
+
+    def test_addDummyNuclidesToLibraryNumGroups(self):
+        isoLib = isotxs.readBinary(os.path.join(FIXTURE_DIR, "ISOAA"))
+        gamLib = gamiso.readBinary(GAMISO_AA)
+        gamLib.gamisoMetadata["numGroups"] = 50
+        dummyNuc = XSNuclide(isoLib, "DMP1AA")
+        dummyNuc.isotxsMetadata = isoLib.getNuclides("AA")[0].isotxsMetadata
+        gamiso.addDummyNuclidesToLibrary(gamLib, [dummyNuc])
+        # Why doesn't  lib.getNuclide("DMP1", "AA") work?
```
opotowsky: TODO: HALP

john-science: Let me look.

john-science: Oh god, dummy nuclides are such a pain.  I can confirm in your branch this works:

```python
gamLib["DUMP1AA"]
```

But these both fail:

```python
gamLib.getNuclide("DMP1", "AA")
# or...
gamLib.getNuclide("DUMP1", "AA")
```

with the same error:

```bash
KeyError: 'DMP1AA'
```

Because `nuclideBases.byName` stores dummies as `DMP`, but if I do `print(gamLib.nuclideLabels)` in your test here I get the nuclide saved as `DUMP1AA`. So, this is a `DMP` vs `DUMP` problem with dummy variables.

I guess no one has noticed this before, because people don't use dummy nuclides for much. I can say this will be A LOT easier to remedy after global nuclides are gone. Which I am working on today!

opotowsky: nice! I thought I had tried all combos of `DMP`, `DUMMY`, and `DUMP`, but I must have missed the last one. 

I'll push up the update. It's much cleaner. Thanks for the sleuthing

opotowsky: AH so it's `gamLib["DMP1AA"]`. That's what ended up working. OMG. So silly.

PR #2173, opened 2025-07-03
Improving Core.libs to look for the current cycle and node

### LINKED ISSUE #2094: Core.lib should try to load the appropriate cycle if it exists
Core.lib (shown below), calls `getExpectedISOTXSFileName`. but it doesn't send any args into it. But we DO have cycle dependent cross section files (eg `ISOTXS-c0n0`) so if the `core.parent.p.cycle` and `core.parent.p.node`, it should see if a cycle specific file exists.

And then if cycle specific does not, send no arguments. Thanks!

https://github.com/terrapower/armi/blob/e263c26014f56fc5b8fccd22429b0a83c3cd0f85/armi/reactor/cores.py#L210-L219

john-science:

When `Core.lib` called `nuclearDataIO.getExpectedISOTXSFileName()` it did it without passing in the `cycle` and `timeNode` parameters. But that could be a very helpful thing to do. So we are passing that option in now.

close #2094 

## SCR Information

Change Type: features

One-Sentence Description: Improving Core.libs to look for the current cycle and node.

---

## Checklist

### DIFF
--- a/armi/nuclearDataIO/__init__.py
+++ b/armi/nuclearDataIO/__init__.py
@@ -15,8 +15,7 @@
 """Read and/or write data files associated with nuclear data and reactor physics data."""
 # ruff: noqa: F401
 
-# export the cccc modules here to keep external clients happy,
-# though prefer full imports in new code
+# Export the cccc modules here for backward compatibility, though prefer full imports in new code.
 from armi.nuclearDataIO.cccc import (
     compxs,
     dif3d,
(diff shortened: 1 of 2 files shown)


opotowsky (approved):
Let me know if there were any downstream impacts we should discuss. Otherwise, easy to approve!

john-science:
Yeah, @onufer , this breaks some tests downstream. TBD

PR #2175, opened 2025-07-07
Blocking duplicate flags from being added

### LINKED ISSUE #2172: Flags sometimes can't be written to a database
When certain amounts of flags are defined, they can't be written to the database. It implies that the math calculating the number of bytes is slightly off and sometimes calculates the wrong value. 

For example, writing a database when this set of flags is defined: 

Image

results in this error: 

Image

john-science:

Here I block a duplicate flag from being added to a collection of flags. I can see no reason to support such a thing. And, indeed, doing so causes some trouble in the current code. So here I just silently pass any time someone tries to add a duplicate flag. I can't think of any good reason for that to raise an error.

close #2172

## SCR Information

Change Type: features

One-Sentence Description: Blocking duplicate flags from being added.

One-line Impact on Requirements: Changes the implementation `I_ARMI_FLAG_EXTEND0` of `R_ARMI_FLAG_EXTEND`, so that if someone tries to add a flag that already exists, it is silently ignored.

---

## Checklist

### DIFF
--- a/armi/utils/flags.py
+++ b/armi/utils/flags.py
@@ -246,9 +246,8 @@ class Flag(metaclass=_FlagMeta):
         <MyFlags.SUPER: 8>
         """
         # add explicit values first, so that autos know about them
-        toAdd = {}
         for field, value in ((f, v) for f, v in fields.items() if isinstance(v, int)):
-            toAdd[field] = cls._registerField(field, value)
+            cls._registerField(field, value)
 
         # find auto values (ignore if they already exist)
         toResolve = [field for field, val in fields.items() if isinstance(val, auto)]
--- a/armi/utils/flags.py
+++ b/armi/utils/flags.py
@@ -15,42 +15,39 @@
 """
 A Flag class, similar to ``enum.Flag``.
 
-This is an alternate implementation of the standard-library ``enum.Flag`` class. We use
-this to implement :py:class:`armi.reactor.flags.Flags`. We used to use the
-standard-library implementation, but that became limiting when we wanted to make it
-possible for plugins to define their own flags; the standard implementation does not
-support extension. We also considered the ``aenum`` package, which permits extension of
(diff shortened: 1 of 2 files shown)


armi/utils/flags.py
```diff
@@ -254,9 +246,13 @@ def extend(cls, fields: Dict[str, Union[int, auto]]):
         <MyFlags.SUPER: 8>
         """
         # add explicit values first, so that autos know about them
+        toAdd = {}
```
alexhjames: why are you storing this dictionary? It doesn't seem to be used for anything?

alexhjames:
I can also confirm this change fixes the test ordering problems I was seeing before.

armi/utils/flags.py
john-science: Ah! A dev item that was left-over. Thanks!

john-science: Removed!

armi/utils/tests/test_flags.py
```diff
@@ -152,3 +153,45 @@ def test_hashable(self):
...
+
+        F.extend({"FLAG0": auto()})
+        for i in range(1, 12):
+            F.extend({f"FLAG{i}": auto()})
+            num = F.len()
+            F.extend({f"FLAG{i - 1}": auto()})
+            self.assertEqual(F.len(), num)
+
+            ff = getattr(F, f"FLAG{i}")
+            FlagSerializer._packImpl(
```
alexhjames, line 175: Can you add a comment here about why these lines are included for future people? They aren't actually used (asserted) in the test, but they are showing that this can be done without raising an error like it did prior to this change.

alexhjames (approved):
one small comment asking for a comment in the code.

armi/utils/tests/test_flags.py
john-science: Fair enough. Done.

alexhjames (approved):

PR #2186, opened 2025-07-15
Ensuring full core BPs aren't converted like 1/3 core

### LINKED ISSUE #2184: Why are we scaling "edge assemblies" in full core models?
I am running into what seems like a significant bug.

I've attached a MWE. The example is full core, with just two assembly types in the core. Below is some code that shows that the `molesHmBOL` parameter gets borked during the process of building the core.

```python
>> import armi; armi.configure()
>> from armi.reactor.reactors import loadFromCs
>> from armi.settings.caseSettings import Settings

>> cs = Settings("anl-afci-177.yaml")
>> r = loadFromCs(cs)

>> a = r.core.getAssemblyWithStringLocation("003-012")
>> a
<inner fuel Assembly A0006 at 003-012>

>> a[2]
<fuel B0006-002 at 003-012-002 XS: A ENV GP: A>

>> a[2].p.molesHmBOL
0.2693242563304581

>> a[2].getHMMoles()
0.23358565163092637
```

Those two last values should be **exactly** the same.

So which one is right?

```python
>> r.blueprints.constructAssem(cs, "inner fuel")[2].getHMMoles()
0.23358565163092637

>> r.blueprints.constructAssem(cs, "inner fuel")[2].p.molesHmBOL
0.23358565163092637
```

So, why is it that the `molesHmBOL` parameter gets screwed up sometime during the process of building the core? I believe that it is related to us calling this:
https://github.com/terrapower/armi/blob/a69b905e8a0210c4f396a052141bd96c1d4571b5/armi/reactor/blueprints/reactorBlueprint.py#L239

which if you trace downwards, will call to this:
https://github.com/terrapower/armi/blob/a69b905e8a0210c4f396a052141bd96c1d4571b5/armi/reactor/converters/geometryConverters.py#L1372

It appears that we ALWAYS scale a selection of parameters for assemblies that are along the symmetry line, even in full core cases. This is a bug that leads to many very surprising results downstream. I was surprised that we have not caught this by now, but looking through the test reactors, it seems like all of the hex cores are defined with 1/3 symmetry, in which case nothing is wrong.

The fix for this is to only do symmetry scaling when we have a non-full core.

john-science:

There was a bug where, when reading a reactor core from blueprints, they were always going through geometry conversion. But for full core, that is unnecessary.

close #2184

## SCR Information

Change Type: fixes

One-Sentence Description: TBD

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/reactorBlueprint.py
+++ b/armi/reactor/blueprints/reactorBlueprint.py
@@ -191,8 +191,9 @@ class SystemBlueprint(yamlize.Object):
             self._loadComposites(cs, bp, system, gridDesign.gridContents, gridDesign.orientationBOL)
 
             if isinstance(system, Core):
-                summarizeMaterialData(system)
+                # if not system.isFullCore:
                 self._modifyGeometry(system, gridDesign)
+                summarizeMaterialData(system)
                 system.processLoading(cs)
 
         return system
--- a/armi/reactor/blueprints/reactorBlueprint.py
+++ b/armi/reactor/blueprints/reactorBlueprint.py
@@ -191,8 +191,8 @@ class SystemBlueprint(yamlize.Object):
             self._loadComposites(cs, bp, system, gridDesign.gridContents, gridDesign.orientationBOL)
 
             if isinstance(system, Core):
-                # if not system.isFullCore:
-                self._modifyGeometry(system, gridDesign)
+                if not system.isFullCore:
+                    self._modifyGeometry(system, gridDesign)
                 summarizeMaterialData(system)
                 system.processLoading(cs)
 
--- a/armi/tests/smallHexReactor/smallHexReactor-bp.yaml
+++ b/armi/tests/smallHexReactor/smallHexReactor-bp.yaml
@@ -1,0 +1,613 @@
+# A small, hex-based, full-core reactor
+nuclide flags:
+  U234:
+    burn: true
+    xs: true
+    expandTo:
+  U235:
+    burn: true
+    xs: true
+    expandTo:
+  U236:
+    burn: true
+    xs: true
+    expandTo:
+  U238:
+    burn: true
+    xs: true
+    expandTo:
+  NP237:
+    burn: true
+    xs: true
+    expandTo:
+  NP238:
+    burn: true
+    xs: true
+    expandTo:
+  PU236:
+    burn: true
+    xs: true
+    expandTo:
+  PU238:
+    burn: true
+    xs: true
+    expandTo:
+  PU239:
+    burn: true
+    xs: true
+    expandTo:
+  PU240:
+    burn: true
+    xs: true
+    expandTo:
+  PU241:
+    burn: true
+    xs: true
+    expandTo:
+  PU242:
+    burn: true
+    xs: true
+    expandTo:
+  AM241:
+    burn: true
+    xs: true
+    expandTo:
+  AM242:
+    burn: true
+    xs: true
+    expandTo:
+  AM243:
+    burn: true
+    xs: true
+    expandTo:
+  CM242:
+    burn: true
+    xs: true
+    expandTo:
+  CM243:
+    burn: true
+    xs: true
+    expandTo:
+  CM244:
+    burn: true
+    xs: true
+    expandTo:
+  CM245:
+    burn: true
+    xs: true
+    expandTo:
+  CM246:
+    burn: true
+    xs: true
+    expandTo:
+  CM247:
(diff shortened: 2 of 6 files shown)


armi/tests/smallHexReactor/smallHexReactor.yaml
```diff
@@ -0,0 +1,26 @@
+# A simple test reactor, not physicially functional
+#
+# * pin-type reactor with hex assemblies
+# * sodium-cooled, fast
+# * full core symmetry
+# * The core grid is corners up
+# * The symmetric core positions have different assembly types
+# * The different assembly types have different amounts of molesHmBOL
+# * There are 3 rings (the assemblies in the 2nd ring don't actually fall along the symmetry line that is checked)
```
keckler:
```suggestion
# * There are 3 rings (necessary for checking problems with symmetry because the assemblies in the 2nd ring don't actually fall along the symmetry line that is checked)
```

armi/tests/smallHexReactor/smallHexReactor.yaml
```diff
@@ -0,0 +1,26 @@
+# A simple test reactor, not physicially functional
+#
+# * pin-type reactor with hex assemblies
+# * sodium-cooled, fast
+# * full core symmetry
+# * The core grid is corners up
+# * The symmetric core positions have different assembly types
+# * The different assembly types have different amounts of molesHmBOL
```
keckler:
```suggestion
# * The different assembly types have different amounts of molesHmBOL because they had different TD_frac material modifications
```

armi/tests/smallHexReactor/smallHexReactor-bp.yaml
```diff
@@ -0,0 +1,613 @@
...
+        height: [307.74]
+        axial mesh points: [1]
+        xs types: [A]
+# end-assemblies-rr
+    radial shield:
+        specifier: SH
+        blocks: [*block_shield]
+        height: [307.74]
+        axial mesh points: [1]
+        xs types: [A]
```
keckler: These are unused
```suggestion
```

armi/tests/smallHexReactor/smallHexReactor-bp.yaml
```diff
@@ -0,0 +1,613 @@
...
+    - B10
+    - B11
+
+
+custom isotopics:
+    pu_vector:
+        input format: number densities
+        PU239: 1
+        U235: 1
+        ZR: 1
```
keckler: Unused
```suggestion
```

armi/tests/smallHexReactor/smallHexReactor-bp.yaml
```diff
@@ -0,0 +1,613 @@
+# A small, hex-based, full-core reactor
+nuclide flags:
```
keckler, line 2: I think the nuclide flags can be very much simplified, but I didn't try and its not that important.

keckler (approved):
I left some suggestions, but nothing that NEEDS to be fixed.

I think this fixes the problem that I ran into, but honestly I would not be surprised if there are still problematic cases out there. It seems very suspect to me that we are just doing this "edge assembly" conversion process on any reactor that isn't full core. What about non-full cartesian cores? They don't have the same symmetry lines, but the edge assembly converter has the symmetry lines for a 1/3 hex core baked in. Something to think about.

armi/reactor/blueprints/reactorBlueprint.py
```diff
@@ -191,8 +191,9 @@ def _constructComposites(self, cs, bp, loadComps, system, gridDesign):
             self._loadComposites(cs, bp, system, gridDesign.gridContents, gridDesign.orientationBOL)
 
             if isinstance(system, Core):
+                if not system.isFullCore:
+                    self._modifyGeometry(system, gridDesign)
```
mgjarrett, line 195: Are we sure that we want to skip all of `_modifyGeometry` when the system is full core, and not just the `scaleParamsRelatedToSymmetry` part?

john-science: For the moment, "modify geometry" doesn't do anything else relevant. 

It could, perhaps, have a much less general-sounding name.

keckler: Mike was right, this chunk of code is now being skipped, but it shouldn't be:
https://github.com/terrapower/armi/blob/38305de56bfa0d7d4ccdf9792560188ae6713eac/armi/reactor/blueprints/reactorBlueprint.py#L243-L251

PR #2191, opened 2025-07-22
Fixing issue with full core BP geometry conversion

john-science:

This is technically still part of the bug fix for this ticket: https://github.com/terrapower/armi/issues/2184

But a problem was found with the solution I provided: https://github.com/terrapower/armi/pull/2186/files#r2223048951

## SCR Information

Change Type: fixes

One-Sentence Description: Fixing issue with full core BP geometry conversion

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/reactorBlueprint.py
+++ b/armi/reactor/blueprints/reactorBlueprint.py
@@ -18,14 +18,13 @@ Definitions of top-level reactor arrangements like the Core (default), SFP, etc.
 See documentation of blueprints in :ref:`bp-input-file` for more context. See example in
 :py:mod:`armi.reactor.blueprints.tests.test_reactorBlueprints`.
 
-This was built to replace the old system that loaded the core geometry from the ``cs['geometry']``
-setting. Until the geom file-based input is completely removed, this system will attempt to migrate
-the core layout from geom files. When geom files are used, explicitly specifying a ``core`` system
-will result in an error.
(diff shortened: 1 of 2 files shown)


armi/reactor/blueprints/reactorBlueprint.py
```diff
@@ -235,10 +234,11 @@ def _modifyGeometry(self, container, gridDesign):
         # all cases should have no edge assemblies. They are added ephemerally when needed
         from armi.reactor.converters import geometryConverters
 
-        runLog.header("=========== Applying Geometry Modifications ===========")
-        converter = geometryConverters.EdgeAssemblyChanger()
-        converter.scaleParamsRelatedToSymmetry(container)
-        converter.removeEdgeAssemblies(container)
+        if not container.isFullCore:
+            runLog.header("=========== Applying Geometry Modifications ===========")
```
keckler: My recommendation would be to leave this header statement outside of the conditional. That way if more geometry conversions/scaling/etc. are done in the future from within other conditionals, then the header doesn't need to be repeated.

keckler (commented):
Thank you!

It might be worth adding some discussion to the docstring for `_constructComposites()` related to the geometry conversions that happen within it.

But no major complaints from me, I think this should work.

armi/reactor/blueprints/reactorBlueprint.py
john-science: done

PR #2202, opened 2025-07-29
Provide Component.pinIndices for helping understand where pins are

### LINKED ISSUE #2200: Method for looking up pin powers by index in data array
Use case: `Block.p.linPowbyPin` is a `(N, )` vector with a scalar per "pin". If there is one pin type or three, it is always a vector. `linPowByPin[j]` is the pin power for the `j`th pin, as determined by `Block.getPinLocations`. There are other parameters defined inside and outside the framework that use a similar structure.

I now have a fuel pin component that I know exists on the block. For assemblies with more than one pin type (common across lots of reactors), this component exists on a subset of the lattice grid and each of the "pins" it represents have a pin power somewhere in `Block.p.linPowByPin`. How do we determine the pin powers for each of the pins represented by a given component?

---

I would propose `Component.p.pinIndices` which is an integer vector `(M, )` where `M <= N`. (M would be the multiplicity of this component).
`ix = Component.p.pinIndices[k`] tells me that the `k`th pin of this component type has a pin power of `Block.p.linPowByPin[ix]`

In the trivial case where there is one fuel pin component per block, this `Component.p.pinIndices` would be `np.arange(N)` or `np.arange(1, N + 1)` if we want the center pin to be position 1 as is the history of ARMI. But, for the case where we have > 1 fuel pins per block, each `Component.p.pinIndicies` would be a non-overlapping set of integers.

I would prefer to have it be zero indexed so I / we don't need to write something like
```python
ix = c.p.pinIndices[k]
pow = c.parent.p.linPowByPin[ix - 1]
```

drewj-tp:
🚨 Draft for some initial stakeholder review 🚨 

Provides `Block.assignPinIndices` which is responsible for setting the new `Component.p.pinIndices` parameter. `Component.p.pinIndices` helps us know two things:

1. Where, in `Block.getPinLocations` and `Block.getPinCoordinates`, do each of the pins for a given `Component` reside?
2. Where, in data arrays derived from `Block.getPinLocations`, do values for pins on a given `Component` reside?

The idea is, for a block with `locations = b.getPinLocations()` as a
list of `IndexLocations` for all the pins in the block, circular
components have a `Component.p.pinIndices` parameter. This data is used
to determine where each of the pins reside in space, and where data for
the pins reside in various data arrays like `linPowByPin`. For example,
```python
locations = b.getPinLocations()
fuel = b.getComponents(Flags.FUEL)[0]
fuelLocations = list(fuel.spatialLocator)
assert len(fuelLocations) == fuel.p.pinIndices.size
for ix, loc in zip(fuel.p.pinIndices, fuelLocations):
    assert loc == locations[ix]
```
One could then find all the pin-powers with
```python
pinPows = b.p.linPowByPin
powersForThisPin = pinPows[fuel.p.pinIndices]
```

Closes #2200 

## SCR Information

Change Type: features

One-Sentence Description: Provides `Block.assignPinIndices` which is responsible for setting the new `Component.p.pinIndices` parameter. 

Do we need new requirements for this?

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -23,6 +23,7 @@ import collections
 import copy
 import functools
 import math
+import operator
 from typing import ClassVar, Optional, Tuple, Type
 
 import numpy as np
@@ -2146,6 +2147,31 @@ class HexBlock(Block):
                     c.spatialLocator = spatialLocators
                 elif c.getDimension("mult") == 1:
                     c.spatialLocator = grids.CoordinateLocation(0.0, 0.0, 0.0, grid)
+
+    def assignPinIndices(self):
+        """Assign pin indices for pin components on the block."""
+        if self.spatialGrid is None:
+            return
+        locations = self.getPinLocations()
+        if not locations:
+            return
+        ijGetter = operator.attrgetter("i", "j")
+        allIJ: tuple[tuple[int, int]] = tuple(map(ijGetter, locations))
+        # Flags for components that we want to set this parameter
+        # Usually things are linked to one of these "primary" flags, like
+        # a cladding component having linked dimensions to a fuel component
+        targetFlags = (Flags.FUEL, Flags.CONTROL, Flags.SHIELD)
+        for c in self.iterChildren(predicate=lambda c: c.hasFlags(targetFlags) and isinstance(c, Circle)):
+            localLocations = c.spatialLocator
+            if isinstance(localLocations, grids.MultiIndexLocation):
+                localIJ = list(map(ijGetter, localLocations))
+            elif isinstance(localLocations, grids.IndexLocation):
+                localIJ = [ijGetter(localLocations)]
+            else:
+                continue
+            localIndices = list(map(allIJ.index, localIJ))
+            # TODO Handle trivial case where one pin lives at all locations
+            c.p.pinIndices = localIndices
 
     def getPinCenterFlatToFlat(self, cold=False):
         """Return the flat-to-flat distance between the centers of opposing pins in the outermost ring."""
--- a/armi/reactor/components/componentParameters.py
+++ b/armi/reactor/components/componentParameters.py
@@ -14,6 +14,8 @@
 
 """Component parameter definitions."""
 
+import numpy as np
+
 from armi.reactor import parameters
 from armi.reactor.parameters import ParamLocation
 from armi.reactor.parameters.parameterDefinitions import isNumpyArray, isNumpyF32Array
@@ -196,6 +198,25 @@ def getCircleParameterDefinitions():
 
         pb.defParam("op", units=units.CM, description="Outer pitch")
 
+        def _validatePinIndices(self, val):
+            if val is not None:
+                # unsigned "short" integer, typically 8 bytes
+                # holds [0, 65_535] so at most, 65_535 pins per block
+                self._p_pinIndices = np.array(val, dtype=np.uint8)
+            else:
+                self._p_pinIndices = None
+
+        pb.defParam(
+            "pinIndices",
+            default=None,
+            description=(
+                "Indices within data arrays where values for this component are stored. "
+                "The array is zero indexed and structured such that the j-th pin on this "
+                "component can be found at ``Block.getPinLocations()[pinIndices[j]]``"
+            ),
+            units=units.UNITLESS,
+            setter=_validatePinIndices,
+        )
     return pDefs
 
 
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -1222,6 +1222,9 @@ class Assembly(composites.Composite):
                     b.autoCreateSpatialGrids(parentSpatialGrid)
                 except (ValueError, NotImplementedError) as e:
                     runLog.extra(str(e), single=True)
+            # Do more grid initializations from a manual or auto created grid
+            if b.spatialGrid is not None:
+                b.assignPinIndices()
 
 
 class HexAssembly(Assembly):
--- a/armi/reactor/tests/test_blocks.py
+++ b/armi/reactor/tests/test_blocks.py
@@ -35,7 +35,6 @@ from armi.physics.neutronics.settings import (
     CONF_XS_KERNEL,
 )
 from armi.reactor import blocks, blueprints, components, geometry, grids
-from armi.reactor.blueprints import blockBlueprint
 from armi.reactor.components import basicShapes, complexShapes
 from armi.reactor.flags import Flags
 from armi.reactor.tests.test_assemblies import makeTestAssembly
--- a/armi/reactor/tests/test_blocks.py
+++ b/armi/reactor/tests/test_blocks.py
@@ -35,6 +35,7 @@ from armi.physics.neutronics.settings import (
     CONF_XS_KERNEL,
 )
 from armi.reactor import blocks, blueprints, components, geometry, grids
+from armi.reactor.blueprints import blockBlueprint
 from armi.reactor.components import basicShapes, complexShapes
 from armi.reactor.flags import Flags
 from armi.reactor.tests.test_assemblies import makeTestAssembly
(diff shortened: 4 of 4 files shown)


armi/reactor/components/basicShapes.py
```diff
@@ -90,6 +94,15 @@ def isEncapsulatedBy(self, other):
         myID, myOD = self.getDimension("id"), self.getDimension("od")
         return otherID <= myID < otherOD and otherID < myOD <= otherOD
 
+    def getPinIndices(self) -> typing.Optional[np.ndarray[tuple[int], int]]:
+        pdata = self.p.pinIndices
+        if pdata is not None:
+            return pdata
+        if self.parent.spatialGrid is not None and self.hasFlags([Flags.FUEL, Flags.CONTROL]):
+            # Save space (RAM and HDD) by computing on the fly
+            return np.arange(self.getDimension("mult"), dtype=np.ushort)
```
john-science: Okay, this is curious.

If this works without any edge cases, it feels like we don't need this parameter, we just do this.

But if that is not true... should we back-fill the param value here before returning it?

drewj-tp: We can't do this all the time, grid or not. Otherwise if we have two different pins in the same block, they would both claim they should have the same data. We can't have an overlap between indices for the same block.

I'm okay setting this parameter here. I was just reading some tea leaves about not wanting this parameter to be defined in the trivial case where a pin occupies the entire lattice grid

https://github.com/terrapower/armi/issues/2200#issuecomment-3119706815

drewj-tp:
Having some good luck with this feature. Going to push some doc updates to help explain how to use this and how to set things like `Block.p.linPowByPin` to be consistent across the ecosystem

drewj-tp:
So now that the parameter could be a `DimensionLink`, the problem is databases don't like writing things that could be None, could be an array, or could be a tuple (component, str). Hitting "ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part" in this part

https://github.com/terrapower/armi/blob/1c16cb8837c2ea4312fd15c657535dd745fa68c5/armi/bookkeeping/db/database.py#L880-L906

Soooo

1. Make this a dimension so it's resolvable,
2. Make a serializer for this parameter

(1) feels not great because we don't want `Component.getDimension` to try and thermally expand this. Which it won't unless someone adds `pinIndices` to a `THERMAL_EXPANSION_DIMS` set. Which, well that would be fun.

(2) feels elegant but cumbersome. Defining a serializer for this, idk seems like overkill but better than even needing to cross the bridge of "this is a dimension so treat it as such." But since this is a `DimensionLink`, that might be too late?

armi/reactor/blockParameters.py
```diff
@@ -37,15 +37,6 @@ def getBlockParameterDefinitions():
             default=None,
         )
 
-        pb.defParam(
-            "pinLocation",
```
john-science, line 41: I just realized this will require some downstream changes. I assume you have those well in hand.

armi/reactor/blocks.py
```diff
@@ -1777,37 +1769,41 @@ def getDuctOP(self):
...
-        """Initialize pin locations."""
-        nPins = self.getNumPins()
-        self.p.pinLocation = list(range(1, nPins + 1))
+        """Initialize pin locations.
+
+        Deprecated. Use :meth:`assignPinIndices` to additionally update component parameters.
+        """
+        # stacklevel=2 means the warning traceback, file, and line numbers will reflect the caller
+        # of this method, not this method itself.
+        warnings.warn(
```
john-science, line 1778: Why are you not just doing `runLog.warn()`?

armi/reactor/components/componentParameters.py
```diff
@@ -175,6 +177,26 @@ def _assignTDFrac(self, val):
...
+                self._p_pinIndices = np.array(val, dtype=np.ushort)
+            else:
+                self._p_pinIndices = None
+
+        pb.defParam(
+            "pinIndices",
+            default=None,
+            description=(
+                "Indices within data arrays where values for this component are stored. "
+                "The array is zero indexed and structured such that the j-th pin on this "
```
john-science, line 192: Isn't everything in Python zero-indexed? Why are we calling that out?  Just curious.

armi/reactor/components/componentParameters.py
```diff
@@ -175,6 +177,26 @@ def _assignTDFrac(self, val):
             description="Total number of moles of heavy metal at BOL.",
         )
 
+        def _validatePinIndices(self, val):
+            if val is not None:
+                # unsigned short integer
+                # holds [0, 65_535] so at most, 65_535 pins per block
```
john-science: Actually, this is only 16-bit on an x86_64 machine: https://numpy.org/doc/1.22/user/basics.types.html

On a 32-bit machine (with 32-bit Python and NumPy), I am like 95% sure this is an 8-bit short.

armi/reactor/blockParameters.py
drewj-tp: Yep.

armi/reactor/blocks.py
drewj-tp: Lots of thoughts.
1. `DeprecationWarning` will be silent during a normal ARMI run, but pytest may print it during someone's testing. So someone building an ARMI application will see this if they call `HexBlock.initializePinLocations`. But not during runtime.
2. The `stacklevel=2` call means the developer who sees this message will know what called `HexBlock.initializePinLocations` from the warning message. They don't need to grep through their codebase to find it. But they still can. Conversely, `runLog.warn` has no machinery for this.
3. A `runLog.warn` of this type, deprecation notice, is maybe not helpful for users to see in their logs. It should be raised and caught by the developers building and calling the to-be-deprecated functions.

armi/reactor/components/componentParameters.py
drewj-tp: Because ARMI has one-indexed things running around too. The first ring of a hexagon is ring one. The first position in a ring is position one.

Maybe being painfully explicit.

armi/reactor/components/componentParameters.py
drewj-tp: I originally had a `np.unit8` but it was not as portable as I like - f87415da4cdd33aa49c5127148a3bb82f107ff8d

And, on recent reflection, it doesn't work on my windows machine like it did prior to that commit...
```python
>>> np.array([300], dtype=np.uint8)
  File "<python-input-2>", line 1, in <module>
    np.array([300], dtype=np.uint8)
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
OverflowError: Python integer 300 out of bounds for uint8
```
`np.ushort` resolves to `np.uint16` on my windows and linux machines

john-science: > np.ushort resolves to np.uint16 on my windows and linux machines

Not to belabor the point, but I don't think so. I think the size of `ushort` depends on the system architecture, as stated in the numpy docs: https://numpy.org/doc/1.22/user/basics.types.html

I guess my question is: Why not just use `np.uint16`?

armi/reactor/blocks.py
john-science: So, you want the `DeprecationWarning` to NOT appear in the ARMI logs? That's interesting. 

Okay, I am not totally convinced, but you seem very convinced. So whatever.

armi/reactor/components/componentParameters.py
drewj-tp: ec4f870e4e783f2ccdb2f5b7b05f6907c6718844

armi/reactor/components/component.py
```diff
@@ -1385,6 +1372,31 @@ def getPinMgFluxes(self, adjoint: Optional[bool] = False, gamma: Optional[bool]
             runLog.error(ee)
             raise ValueError(msg) from ee
 
+    def getPinIndices(self) -> Optional[np.ndarray[tuple[int], np.ushort]]:
```
john-science: And this `np.ushort` doesn't have to match the one you changed?

drewj-tp: a8c67c56

PR #2162, opened 2025-06-24
Make axial linking aware of block grids for axial expansion

albeanth:
## What is the change? Why is it being made?

See https://github.com/terrapower/armi/pull/2145. Myself and @drewj-tp were holding this off on a side branch for some other potential downstream work, but have decided to just pull this into ARMI `main`. This PR only brings in #2145, which has already been reviewed, so there is nothing new to review with this PR. 

## SCR Information

Change Type: features

One-Sentence Description: Enable multi-pin component support for axial expansion.

One-line Impact on Requirements: NA

---

## Checklist

- [x] This PR has only [one purpose or idea](https://terrapower.github.io/armi/developer/tooling.html#one-idea-one-pr).
- [x] [Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify any new/changed code.
- [x] The [documentation](https://terrapower.github.io/armi/developer/tooling.html#document-it) is still up-to-date in the `doc` folder.
- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The dependencies are still up-to-date in `pyproject.toml`.

PR #2203, opened 2025-07-29
Slight refactor on b.getSmearDensity to accommodate downstream work

albeanth:

To accommodate downstream work supporting assemblies with multiple pin types, this PR refactors block.getSmearDensity to allow users to provide an explicit list of components rather than relying on the block itself. 

## SCR Information

Change Type: trivial

One-Sentence Description: Refactor block.getSmearDensity to allow users to provide an explicit list of components rather than relying on the block itself. 

---

## Checklist


opotowsky (approved):

albeanth:
> @opotowsky @albeanth Ideally, we would add a unr test when we add a feature like this.
> 
> Open a ticket for that, maybe?

@john-science sure. There's test coverage technically with the existing test for `getSmearDensity`, but I can add one. I will add one once #2162 is merged.

PR #2208, opened 2025-08-01
Making ParamMapper symmetry-aware

bsculac:

The parameter mapper did not account for symmetry when mapping parameters between blocks. This was an issue when using the parameter mapper to assign parameters to an assembly that was not yet placed in the reactor grid from an assembly that had a non-unity symmetry factor. Now, the ParamMapper will internally convert volume integrated parameters to full-block values when getting parameters, and reduce parameters to partial values when assigning.

This specifically fixes an issue with some current development that @mgjarrett is doing, and prevents having a hyper-specific local fix for symmetry handling.

## SCR Information

Change Type: features

One-Sentence Description: ParamMapper now converts volume integrated parameters to full size when getting parameters (when appropriate), and converts parameters to partial size when setting parameters (when appropriate).

---

## Checklist

### DIFF
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -57,8 +57,8 @@ import collections
 import copy
 import glob
 import re
-from timeit import default_timer as timer
 import typing
+from timeit import default_timer as timer
 
 import numpy as np
 
@@ -1441,10 +1441,16 @@ class ParamMapper:
             multiplier = self.getFactorSymmetry(paramName, symmetryFactor)
             val = block.p[paramName]
             # list-like should be treated as a numpy array
-            if isinstance(val, (tuple, list, np.ndarray)):
-                paramVals.append(np.array(val) * multiplier if len(val) > 0 else None)
+            if val is None:
+                if isinstance(val, (tuple, list, np.ndarray)):
+                    paramVals.append(np.array(val) if len(val) > 0 else None)
+                else:
+                    paramVals.append(val)
             else:
-                paramVals.append(val * multiplier)
+                if isinstance(val, (tuple, list, np.ndarray)):
+                    paramVals.append(np.array(val) * multiplier if len(val) > 0 else None)
+                else:
+                    paramVals.append(val * multiplier)
 
         return np.array(paramVals, dtype=object)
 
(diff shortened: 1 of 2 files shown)


armi/reactor/converters/uniformMesh.py
```diff
@@ -1416,45 +1420,63 @@ def __init__(self, reactorParamNames, blockParamNames, b):
...
+            multiplier = self.getFactorSymmetry(paramName, symmetryFactor)
             val = block.p[paramName]
             # list-like should be treated as a numpy array
-            if isinstance(val, (tuple, list, np.ndarray)):
-                paramVals.append(np.array(val) if len(val) > 0 else None)
+            if val is None:
+                if isinstance(val, (tuple, list, np.ndarray)):
+                    paramVals.append(np.array(val) if len(val) > 0 else None)
+                else:
+                    paramVals.append(val)
```
mgjarrett, line 1448: I'm confused about the need for this whole block. If `val is None`, then it's not going to be a `tuple`, `list`, or `np.ndarray`.

I think one layer of the `if`/`else` could be removed here.

```
            if isinstance(val, (tuple, list, np.ndarray)):
                paramVals.append(np.array(val) * multiplier if len(val) > 0 else None)
            elif val is not None:
                paramVals.append(val * multiplier)
            else:
                paramVals.append(val)
```

armi/reactor/converters/uniformMesh.py
```diff
@@ -1416,45 +1420,63 @@ def __init__(self, reactorParamNames, blockParamNames, b):
...
         paramVals = []
+        symmetryFactor = block.getSymmetryFactor()
         for paramName in paramNames:
+            multiplier = self.getFactorSymmetry(paramName, symmetryFactor)
             val = block.p[paramName]
             # list-like should be treated as a numpy array
-            if isinstance(val, (tuple, list, np.ndarray)):
-                paramVals.append(np.array(val) if len(val) > 0 else None)
+            if val is None:
+                if isinstance(val, (tuple, list, np.ndarray)):
```
mgjarrett: If `val is None` then this `if` here wouldn't ever be true.

You could simplify lines 1444-1453 down to something like this:

```python
            if isinstance(val, (tuple, list, np.ndarray)):
                paramVals.append(np.array(val) * multiplier if len(val) > 0 else None)
            elif param is None:
                paramVals.append(val)
            else:
                paramVals.append(val * multiplier)
```

bsculac: updated, thanks!

armi/reactor/converters/uniformMesh.py
```diff
@@ -1416,45 +1420,59 @@ def __init__(self, reactorParamNames, blockParamNames, b):
...
             # Skip setting None values.
             if val is None:
                 continue
 
             if isinstance(val, (tuple, list, np.ndarray)):
-                ParamMapper._arrayParamSetter(block, [val], [paramName])
+                self._arrayParamSetter(block, [val], [paramName])
             else:
-                ParamMapper._scalarParamSetter(block, [val], [paramName])
+                print(paramName)
```
mgjarrett:
```suggestion
```

mgjarrett (changes requested):

mgjarrett (approved):

PR #2209, opened 2025-08-01
Block collection nuclides

mgjarrett:

When creating a representative block for a block collection with `averageByComponent: true`, each component is assigned only the nuclides present in that component.

Previously, the component number densities were set for all nuclides in the problem. For many components, most of the nuclides are not present and were thus stored at a 0.0 number density.

## SCR Information

Change Type: features

One-Sentence Description: reduce the number of nuclides stored on each component in a representative block by ignoring those with 0.0 number density.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/crossSectionGroupManager.py
+++ b/armi/physics/neutronics/crossSectionGroupManager.py
@@ -390,11 +390,11 @@ class AverageBlockCollection(BlockCollection):
         numberDensities : dict
             nucName, ndens data (atoms/bn-cm)
         """
-        nuclides = self.allNuclidesInProblem
         blocks = self.getCandidateBlocks()
         weights = np.array([self.getWeight(b) for b in blocks])
         weights /= weights.sum()  # normalize by total weight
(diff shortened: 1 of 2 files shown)


mgjarrett:
> It's a small change and I like it.
> 
> Out of curiosity, will it matter downstream that the representative blocks change the number of nuclides they have over time?
> 
> I'm thinking that at BOC a fuel block will have fewer nuclides than it will have at EOC, due to depletion/decay.

I don't expect that this would matter, but I haven't fully completed downstream testing yet. Once that's complete I can say it doesn't matter with more confidence.

mgjarrett:
> @mgjarrett - you may want to check that nuclides with zero density get omitted from cross sections since one goal of keeping infinitely dilute cross sections would be to for follow on one-group XS or reaction rate estimations. Having many nuclides increases the calculation burden, but could be used or expected behavior.

Depletable components will still have all of the possible nuclides present at infinite dilute number densities. The goal of this change is for non-depletable components to not inadvertently acquire those nuclides when creating a representative block.

PR #2217, opened 2025-08-08
Improving documentation of axial expansion

john-science:

There were some end-user requests for a little more information about axial expansion. I am attempting to improve that documentation here.

## SCR Information

Change Type: docs

One-Sentence Description: Improving documentation of axial expansion

---

## Checklist

### DIFF
--- a/doc/user/radial_and_axial_expansion.rst
+++ b/doc/user/radial_and_axial_expansion.rst
@@ -2,10 +2,13 @@
 Radial and Axial Expansion and Contraction
 ******************************************
 
-ARMI natively supports linear expansion in both the radial and axial dimensions. These expansion
-types function independently of one another and each have their own set of underlying assumptions
-and use-cases. Radial expansion happens by default but axial expansion only occurs if the setting
-``inputHeightsConsideredHot`` is set to ``False``.
+ARMI natively supports linear expansion in both the radial and axial dimensions for pin-type reactors. These expansion
+types function independently of one another and each have their own set of underlying assumptions and use-cases. Radial
+expansion happens by default but there are several settings that control axial expansion:
+
+* ``inputHeightsConsideredHot`` - Set this to ``False`` or axial expansion will not occur.
+* ``assemFlagsToSkipAxialExpansion`` - Assemblies with a flag in this list will not be axially expanded.
+* ``detailedAxialExpansion`` - Allow each assembly to expand independently. This will result in a non-uniform mesh.
 
 Thermal Expansion
 =================
--- a/doc/user/radial_and_axial_expansion.rst
+++ b/doc/user/radial_and_axial_expansion.rst
@@ -10,6 +10,11 @@ expansion happens by default but there are several settings that control axial e
 * ``assemFlagsToSkipAxialExpansion`` - Assemblies with a flag in this list will not be axially expanded.
 * ``detailedAxialExpansion`` - Allow each assembly to expand independently. This will result in a non-uniform mesh.
 
+By default, ARMI runs radial and axial expansion when objects are created from blueprints. That is, when the reactor is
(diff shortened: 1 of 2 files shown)


doc/user/radial_and_axial_expansion.rst
```diff
@@ -2,10 +2,18 @@
...
+types function independently of one another and each have their own set of underlying assumptions and use-cases. Radial
+expansion happens by default but there are several settings that control axial expansion:
+
+* ``inputHeightsConsideredHot`` - Set this to ``False`` or axial expansion will not occur.
+* ``assemFlagsToSkipAxialExpansion`` - Assemblies with a flag in this list will not be axially expanded.
+* ``detailedAxialExpansion`` - Allow each assembly to expand independently. This will result in a non-uniform mesh.
+
+By default, ARMI runs radial and axial expansion when objects are created from blueprints. That is, when the reactor is
+created from blueprints at BOL, these calculations are performed. But also at BOC if new assemblies are added to the
+core, then expansion will happen again when the assembly object is created from blueprints.
```
albeanth, line 15: This is not exactly true. We perform axial expansion at BOL on the blueprints assemblies if `inputHeightsConsideredHot: False`. When assemblies are added to the core, e.g., shuffling or building out the core at BOL, copies of these (potentially already expanded) blueprints assemblies are added.

doc/user/radial_and_axial_expansion.rst
```diff
@@ -2,10 +2,18 @@
...
 
-ARMI natively supports linear expansion in both the radial and axial dimensions. These expansion
-types function independently of one another and each have their own set of underlying assumptions
-and use-cases. Radial expansion happens by default but axial expansion only occurs if the setting
-``inputHeightsConsideredHot`` is set to ``False``.
+ARMI natively supports linear expansion in both the radial and axial dimensions for pin-type reactors. These expansion
+types function independently of one another and each have their own set of underlying assumptions and use-cases. Radial
+expansion happens by default but there are several settings that control axial expansion:
+
+* ``inputHeightsConsideredHot`` - Set this to ``False`` or axial expansion will not occur.
```
albeanth: We should add a note here that this setting only controls the behavior at BOL and if the blueprints assemblies are expanded. It has no control over downstream ARMI plugins/apps.

clstocking: I would reword this description to be something along the lines of "Indicates whether blueprints heights have already been thermally expanded. If ``False``, ARMI will expand components at BOL consistent with provided temperatures"

john-science: Changed

doc/user/radial_and_axial_expansion.rst
john-science: Updated.

But, just to be clear, the reason anything happens at "BOL" is because your logic is injected whenever something is created from the blueprints. That is at BOL or even at BOC if a new Assembly is created.

See what you think of the new version.

albeanth: I suppose I'm confused on what you are trying to communicate with these sentences? The first two sentences are more or less repetitions from what's already been said and I do not think that expansion happens again at BOC when an assembly is shuffled into a core. The blueprint assembly is already expanded and then copies of it are brought into the core directly without a secondary round of expansion.

john-science: Well, the code you used to inject axial expansion connects up to ARMI in the blueprints Python code, right?

So, "when" this axial expansion happens is when that code is called. And that code is executed whenever an assembly is created from the blueprints. So that is at BOL when a new set of assemblies is created for the core. But that code is also called when a new assembly is created at BOC for new assemblies rotated into the core.

Does that make sense?

Thoughts?

PR #2218, opened 2025-08-08
Adding a method to get cycle/node combinations for a time interval

### LINKED ISSUE #2216: Add Method for getting Database Cycle-Nodes Corresponding to a Time Range
It's not always clear before the fact what cycle, node pairs of a database correspond to a reactor time in years from `r.p.time`. It would make analyses easier if there was a method that returned the cycle,node pairs that fall within a given time range in years without having to load a database or manually examine the database file. The method would also return an error message if no cycle,node pair falls within the time range selected.

For example:

Given a database method `getCycleNodeAtTime(armiDatabase, startTime, endTime, errorIfNotExactlyOne=True)`

This should not actually load the DB in ARMI, it should just investigate the h5 directly (to stay fast)

I could call something like `armi.bookkeeping.db.getCycleNodeAtTime("./database.h5", 0, 1)` on an existing database and assuming this database only has 1 cycle of 3 nodes with a total cycle length of 365 EFPD (and corresponding reactor time r.p.time of ~1.0) this method would return a list of cycle,node tuples [(0,0), (0,1), (0,2)]. Likewise, if I called `armi.bookkeeping.db.getCycleNodeAtTime("./database.h5", 1, 2)` this would result in an error stating that no cycle,node pair exist on this database for this time range.

The implementation of this method could read all values from the h5 directly and not need to use an expensive method like `db.loadOperator`

john-science:

There was a user request for a small utility method to get a list of cycle/node combinations for a given time interval. 

close #2216 

## SCR Information

Change Type: features

One-Sentence Description: Adding a small utility method to get a list of cycle/node combinations for a given time interval. 

One-line Impact on Requirements: This is a small change to `I_ARMI_DB_H5` for `R_ARMI_DB_H5`, in that we added code. But this is just an added convenience utility to safe users doing math, it does not meaningfully change the Database interface.

---

## Checklist


john-science:
@wcscherer I see that you are not a member of the TerraPower org on github yet. Would you like to be? I can't assign PRs for review to accounts outside our org.

armi/bookkeeping/db/database.py
```diff
@@ -1463,6 +1463,73 @@ def _applyComponentNumberDensitiesMigration(comps, unpackedData):
             c.p["nuclides"] = nuclides
             c.p["numberDensities"] = numberDensities
 
+    @staticmethod
+    def getCycleNodeAtTime(dbPath, startTime, endTime, errorIfNotExactlyOne=False):
```
onufer: from a user perspective we typically are just looking for 1 time node, so i would prefer if this defaulted to "true"

armi/bookkeeping/db/database.py
```diff
@@ -1463,6 +1463,73 @@ def _applyComponentNumberDensitiesMigration(comps, unpackedData):
...
+        errorIfNotExactlyOne : boolean
+            Raise an error if more than one cycle/node combination is returned. Default is False.
+
+        Returns
+        -------
+        list of two-tuples of integers
+            A list of (cycle, node) combinations corresponding to the desired time interval.
+        """
+        # basic sanity checks
+        assert startTime >= 0.0, f"The start time cannot be negative: {startTime}."
```
onufer, line 1488: i love assert statements, but was told we dont really want them in the code base. maybe this has changed.

armi/bookkeeping/db/database.py
```diff
@@ -1463,6 +1463,73 @@ def _applyComponentNumberDensitiesMigration(comps, unpackedData):
...
+        assert endTime >= startTime, f"The end time {endTime} is not greater than the start time {startTime}."
+
+        # open the H5 file directly
+        h5 = h5py.File(dbPath, "r")
+
+        # load Settings object
+        cs = settings.Settings()
+        cs.caseTitle = os.path.splitext(os.path.basename(dbPath))[0]
```
onufer: could we access the h5 file directly? DMed you some more info

armi/bookkeeping/db/database.py
```diff
@@ -1463,6 +1463,73 @@ def _applyComponentNumberDensitiesMigration(comps, unpackedData):
...
+        timePassed = 0
+        endFound = False
+        cycleNodes = []
+        for cycle, nodes in enumerate(stepLengths):
+            for node, timeStep in enumerate(nodes):
+                timePassed += timeStep
+                if not cycleNodes and timePassed >= startTime:
+                    cycleNodes.append((cycle, node))
+                elif cycleNodes:
+                    cycleNodes.append((cycle, node))
```
onufer: i think we sometimes have "named" timesteps that have more than just cycle/node. how would that work?

armi/bookkeeping/db/database.py
john-science: There's nothing wrong with having error handling in the code base. ARMI certainly has never put such a guideline in the docs.

armi/bookkeeping/db/database.py
john-science: Updated

armi/bookkeeping/db/database.py
john-science: updated

armi/bookkeeping/db/database.py
```diff
@@ -1463,6 +1463,60 @@ def _applyComponentNumberDensitiesMigration(comps, unpackedData):
...
+        -------
+        list of strings
+            A list of strings to the desired time interval, e.g.: ["c01n08", "c14n18EOL"]
+        """
+        # basic sanity checks
+        assert startTime >= 0.0, f"The start time cannot be negative: {startTime}."
+        assert endTime >= startTime, f"The end time ({endTime}) is not greater than the start time ({startTime})."
+
+        # open the H5 file directly
+        h5 = h5py.File(dbPath, "r")
```
onufer: does `h5py.File` have a `with open` option?

armi/bookkeeping/db/database.py
```diff
@@ -1463,6 +1463,60 @@ def _applyComponentNumberDensitiesMigration(comps, unpackedData):
...
+        # read time steps in H5 file
+        thisTime = 0.0
+        cycleNodes = []
+        for h5Key in h5.keys():
+            if h5Key == "inputs":
+                continue
+
+            thisTime = h5[h5Key]["Reactor"]["time"][0]
+            if thisTime >= endTime:
+                cycleNodes.append(h5Key)
```
onufer: i dont think we would want to append in this case since its after the end time

armi/bookkeeping/db/database.py
```diff
@@ -1463,6 +1463,60 @@ def _applyComponentNumberDensitiesMigration(comps, unpackedData):
...
+                break
```
onufer: this break assumes that the database nodes are all in order of ascending time. that's probably a pretty good assumption. But do you think it might be robust to just check all the times and add the ones within the range?

armi/bookkeeping/db/database.py
```diff
@@ -1463,6 +1463,60 @@ def _applyComponentNumberDensitiesMigration(comps, unpackedData):
...
+        h5 = h5py.File(dbPath, "r")
+
+        # read time steps in H5 file
+        thisTime = 0.0
+        cycleNodes = []
+        for h5Key in h5.keys():
+            if h5Key == "inputs":
+                continue
+
+            thisTime = h5[h5Key]["Reactor"]["time"][0]
```
onufer: i think this will fail for many of our databases. for c00n00 time has typically not yet been written to the database. can you test it on our intern ones. this is why i had to write the second if statement here. but it should probably say if "c00n00" in k since it could have EOL or some other name on the end 
image

armi/bookkeeping/db/database.py
onufer: what do you think about this?

armi/bookkeeping/db/database.py
john-science: We recently merged a change in ARMI that sets the `r.p.time = 0` at `c00n000`:

https://github.com/terrapower/armi/blob/fcc30f4c78604dd9fbd21fe8cfdc4332dcb519c9/armi/operators/operator.py#L407-L411

So, this shouldn't be a problem any more.

armi/bookkeeping/db/database.py
john-science: done

armi/bookkeeping/db/database.py
john-science: This was here because if the target time is between the start and end time in a given time node, we would want that time node counted. It's just how the loop is structured.

armi/bookkeeping/db/database.py
john-science: If the time nodes aren't in order, I'm not sure the H5 file approach works. Because the `r.p.time` saved to the database during each time node is the time at the END of that timenode. We don't save the `r.p.time` at the START of the time node. The only way we could get that would be to go back to the old `r.p.stepLengths` solution, before we were readying from the DB. And that solution fails for named time nodes.

If it helps, ARMI only ever runs the reactor simulation forward through time. So we have no reason to write data to the DB out of chronological order.

Thoughts?

armi/bookkeeping/db/database.py
john-science: done

armi/bookkeeping/db/database.py
```diff
@@ -1463,6 +1463,55 @@ def _applyComponentNumberDensitiesMigration(comps, unpackedData):
...
+            # read time steps in H5 file
+            thisTime = 0.0
+            cycleNodes = []
+            for h5Key in h5.keys():
+                if h5Key == "inputs":
+                    continue
+
+                thisTime = h5[h5Key]["Reactor"]["time"][0]
+                if thisTime >= endTime:
+                    cycleNodes.append(h5Key)
```
onufer, line 1502: do we really want to append here? i think it should maybe just be break because its after the requested end time.

Although, this code still assumes that all the points in the database are in time order.. i think some simplification is possible....

armi/bookkeeping/db/database.py
```diff
@@ -1463,6 +1463,55 @@ def _applyComponentNumberDensitiesMigration(comps, unpackedData):
...
+                    break
+                elif thisTime >= startTime:
+                    cycleNodes.append(h5Key)
```
onufer, line 1505:
```suggestion
                if startTim <= thisTime <= endTime:
                    cycleNodes.append(h5Key)
```

PR #2219, opened 2025-08-11
Refactoring Shuffle Logic Inputs to YAML

HunterPSmith:

Changes made to make assembly shuffling and rotation more extensible and not require the shuffle logic python script.

- Introduced `CONF_SHUFFLE_SEQUENCE_FILE` setting for custom shuffle sequences.
- Updated `FuelHandler` to process shuffle sequences from a YAML file.
- Enhanced settings validation for shuffle sequence file existence.
- Updated documentation to reflect changes in shuffle management.
- Update testing for new methods
- Implement `AssemblyMove` data class to improve clarity and maintainability for handling assembly movements

## SCR Information

Change Type: features

One-Sentence Description: Implement YAML shuffle file input with ability to manually rotate individual assemblies.

---

## Checklist


armi/reactor/cores.py
```diff
@@ -1599,11 +1599,13 @@ def _getReflectiveDuplicateAssembly(self, neighborLoc):
 
     def setMoveList(self, cycle, oldLoc, newLoc, enrichList, assemblyType, assemName):
         """Tracks the movements in terms of locations and enrichments."""
-        data = (oldLoc, newLoc, enrichList, assemblyType, assemName)
+        from armi.physics.fuelCycle.fuelHandlers import AssemblyMove
```
john-science, line 1605: Can this import be at the top of the file?

Or did you try that and it caused an import loop failure?

armi/settings/settingsValidation.py
```diff
@@ -430,6 +430,17 @@ def _willBeCopiedFrom(fName):
             self.NO_ACTION,
         )
 
+        self.addQuery(
```
drewj-tp, line 442: I think we should have a validator if you define multiple of
1. shuffle logic file,
2. explicit repeat shuffles file,
3. new shuffle sequence file

doc/user/inputs.rst
```diff
@@ -366,7 +366,41 @@ A few examples of restart cases:
...
+   ``cascade`` chains, ``misloadSwap`` entries, and optional manual
+   ``rotations``. Cascades describe a chain of assembly displacements 
+   beginning with a fresh fuel assembly and ending with the final location's
+   assembly being discharged. Optional ``fuelEnrichment`` lists specify the
+   U235 weight percent enrichment for each axial block in the fresh assembly,
+   from bottom to top, including zeroes for non-fuel blocks. ``misloadSwap``
+   swaps the assemblies at two locations after all cascades are processed but
+   before any rotations are applied. Manual ``rotations`` map final location
+   labels to relative angles in degrees and are performed after any
+   algorithmic rotation routines. Valid angles depend on the assembly's
```
drewj-tp: We should reference the setting/settings that configure the algorithmic rotation routines, e.g.,
```suggestion
   algorithmic rotation routines defined with the ``assemblyRotationAlgorithm`` setting. Valid angles depend on the assembly's
```

doc/user/inputs.rst
```diff
@@ -366,7 +366,41 @@ A few examples of restart cases:
...
+   geometry. Rotations apply to the assembly that ends up at the given location
+   once all moves and swaps in that cycle are
+   complete. A cascade with no final destination defaults to discharging the
```
drewj-tp: Trying to wrap my head around how a user is intended to interpret the rotations field. There's a lot going on.

1. Whatever assembly ends up at the listed location gets the rotation
2. Rotation is relative to however the assembly is oriented
3. Rotation happens after the algorithmic shuffling

(1) is okay. But putting the rotating directly in the cascade section implies, to me, 
1. I create fresh fuel
2. I move it to 009-045
3. I then rotate the fresh assembly at 009-045 60 degrees CCW

(2) is also fine but gets a bit muddy with (3)

Is the intent for the rotations to more support some accident analysis scenario like a misload? What happens if we accidentally over-rotate some assembly? If that is the case, I would recommend we break the rotations field out into it's own sub-field like misload swap
```yaml
sequence:
  1:
    - cascade: [...]
    - cascase: [...]
    - misloadSwap: ...
    - extraRotations: {"009-045": 60}
```

doc/user/inputs.rst
```diff
@@ -366,7 +366,41 @@ A few examples of restart cases:
...
+   ``SFP`` it is only retained if the ``trackAssems`` setting is True;
+   ``ExCore`` always deletes the assembly.
+   For example::
+
+       sequence:
+        1:
+          - cascade: ["outer fuel", "009-045", "008-004", "SFP"]
+            fuelEnrichment: [0, 12, 14, 15, 0]  # wt% U235 by block
+            rotations: {"009-045": 60}
+          - misloadSwap: ["009-045", "008-004"]
```
drewj-tp: The paragraph explanation you have above is great at explaining these fields. I think users will be aided by us writing out what this sequence would look like. Something like

1. Assembly A, originally at `008-004`, is discharged to SFP
2. Assembly B, originally at `009-045`, is moved to `008-004`
3. A fresh `outer fuel` assembly is created with an axial enrichment profile
4. The fresh assembly is placed at `009-045`
5. Fresh assembly and assembly B are swapped such that
   - Fresh assembly now resides at `008-004`
   - Assembly B is now at it's original location of `009-045`
6. Assembly B, located at `009-045` is rotated an additional 60 degrees CCW

Please confirm this behavior is what we expect to happen because this is what I gathered from reading this doc.

doc/user/inputs.rst
```diff
@@ -366,7 +366,41 @@ A few examples of restart cases:
...
+   swaps the assemblies at two locations after all cascades are processed but
+   before any rotations are applied. Manual ``rotations`` map final location
+   labels to relative angles in degrees and are performed after any
+   algorithmic rotation routines. Valid angles depend on the assembly's
+   geometry. Rotations apply to the assembly that ends up at the given location
+   once all moves and swaps in that cycle are
+   complete. A cascade with no final destination defaults to discharging the
+   assembly to the spent fuel pool ``SFP``. Assemblies can also be removed from
+   the model entirely by ending with ``ExCore``. When an assembly is sent to the
+   ``SFP`` it is only retained if the ``trackAssems`` setting is True;
```
drewj-tp:
```suggestion
   ``SFP`` it is only retained in memory if the ``trackAssems`` setting is True;
```

doc/user/inputs.rst
```diff
@@ -366,7 +366,41 @@ A few examples of restart cases:
...
+   ``ExCore`` always deletes the assembly.
+   For example::
```
drewj-tp: I _think_ we need to do
```suggestion
   For example
   
.. code:: yaml
```
to get this to render as yaml and not python

doc/user/inputs.rst
```diff
@@ -366,7 +366,41 @@ A few examples of restart cases:
...
+       sequence:
+        1:
+          - cascade: ["outer fuel", "009-045", "008-004", "SFP"]
+            fuelEnrichment: [0, 12, 14, 15, 0]  # wt% U235 by block
+            rotations: {"009-045": 60}
+          - misloadSwap: ["009-045", "008-004"]
+        2:
+          - cascade: ["outer fuel", "010-046", "009-045", "ExCore"]
+            fuelEnrichment: [0, 12, 14, 15, 0]
+
```
drewj-tp: A gentle `.. note::` block to point users to the user of yaml anchors here will be very helpful. We don't need to provide a full example, but something like
```rst
.. note:: 
    Consider using yaml anchors ``&`` and aliases ``*`` to reduce repetition.
```

armi/tests/armiRun-SHUFFLES.yaml
```diff
@@ -0,0 +1,14 @@
...
+      fuelEnrichment: [0, 12, 14, 15, 0]
+      rotations: {"009-045": 60}
+    - cascade: ["outer fuel", "010-046", "011-046", "012-046", "ExCore"]
+      fuelEnrichment: [0, 12, 14, 15, 0]
+      rotations: {}
+  2: *cycle_1
+  3:
+    - *shuffle_1_9_45
+    - misloadSwap: ["009-045", "008-004"]
+    - misloadSwap: ["008-004", "009-045"]
```
drewj-tp: Is this intentional? Swapping two assemblies and then swapping them back? It's probably fine but that feels odd

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -532,22 +639,66 @@ def test_readMoves(self):
...
+                AssemblyMove("011-046", "012-046", [], None, None),
+                AssemblyMove("012-046", "ExCore", [], None, None),
+            ],
+            3: [
+                AssemblyMove("LoadQueue", "009-045", [0.0, 12.0, 14.0, 15.0, 0.0], "outer fuel", None),
+                AssemblyMove("009-045", "008-004", [], None, None),
+                AssemblyMove("008-004", "007-001", [], None, None),
+                AssemblyMove("007-001", "006-005", [], None, None),
+                AssemblyMove("006-005", "SFP", [], None, None),
+                AssemblyMove("009-045", "009-045", [], None, None, 60.0),
```
drewj-tp: Here we have a rotation happening before all the swaps are processed, which goes against what we have in the docs

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -532,22 +639,66 @@ def test_readMoves(self):
...
+        self.assertEqual(sfpMove.movingAssemName, "A0073")  # name of assem in SFP
 
         # make sure we fail hard if the file doesn't exist
         with self.assertRaises(RuntimeError):
             fh.readMoves("totall_fictional_file.txt")
 
+    def test_readMovesYaml(self):
+        fh = fuelHandlers.FuelHandler(self.o)
+        moves = fh.readMovesYaml("armiRun-SHUFFLES.yaml")
+        self.maxDiff = None
```
drewj-tp, line 673: Un-used?
```suggestion
```

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -532,22 +639,66 @@ def test_readMoves(self):
...
+        expected = {
+            1: [
+                AssemblyMove("LoadQueue", "009-045", [0.0, 12.0, 14.0, 15.0, 0.0], "outer fuel", None),
+                AssemblyMove("009-045", "008-004", [], None, None),
+                AssemblyMove("008-004", "007-001", [], None, None),
+                AssemblyMove("007-001", "006-005", [], None, None),
+                AssemblyMove("006-005", "SFP", [], None, None),
+                AssemblyMove("009-045", "009-045", [], None, None, 60.0),
```
drewj-tp: Another rotation that happens after all cascases are performed, contrary to the docs

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -532,22 +639,66 @@ def test_readMoves(self):
...
-        self.assertEqual(firstMove[0], "002-001")
-        self.assertEqual(firstMove[1], "SFP")
-        self.assertEqual(len(firstMove[2]), numblocks)
-        self.assertEqual(firstMove[3], "igniter fuel")
-        self.assertEqual(firstMove[4], None)
+        self.assertEqual(firstMove.fromLoc, "002-001")
+        self.assertEqual(firstMove.toLoc, "SFP")
+        self.assertEqual(len(firstMove.enrichList), numblocks)
+        self.assertEqual(firstMove.assemType, "igniter fuel")
+        self.assertIsNone(firstMove.movingAssemName)
```
drewj-tp: so so much more readable 🤩

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -45,6 +47,111 @@
...
+            self._run(yaml_text)
+
+    def test_missingCycle(self):
+        yaml_text = "sequence:\n  1: []\n  3: []\n"
+        with self.assertRaisesRegex(InputError, "Missing cycle 2"):
+            self._run(yaml_text)
+
+    def test_duplicateCycle(self):
+        yaml_text = "sequence:\n  1: []\n  1: []\n"
+        with self.assertRaisesRegex(InputError, "Duplicate cycle"):
```
drewj-tp: I'm surprised we need to write code and a custom exception to handle this. yaml will throw a pretty detailed exception if you have duplicated keys
```python
import ruamel.yaml as yaml
y = yaml.YAML()
y.load("""parent:
  child: 1
  child: 2
""")
DuplicateKeyError: while constructing a mapping
  in "<unicode string>", line 2, column 3:
      child: 1
      ^ (line: 2)
found duplicate key "child" with value "2" (original value: "1")
  in "<unicode string>", line 3, column 3:
      child: 2
      ^ (line: 3)
```

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -45,6 +47,111 @@
...
+            fname = tf.name
+        try:
+            return fuelHandlers.FuelHandler.readMovesYaml(fname)
+        finally:
+            os.remove(fname)
+
+    def test_cycles_out_of_order(self):
+        yaml_text = "sequence:\n  1: []\n  2: []\n  4: []\n  3: []\n"
+        moves = self._read(yaml_text)
+        self.assertEqual(list(moves.keys()), [1, 2, 4, 3])
```
drewj-tp: Dictionaries are already iterable so we can write
```suggestion
        self.assertEqual(list(moves), [1, 2, 4, 3])
```

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -532,22 +639,66 @@ def test_readMoves(self):
...
+                AssemblyMove("LoadQueue", "009-045", [0.0, 12.0, 14.0, 15.0, 0.0], "outer fuel", None),
+                AssemblyMove("009-045", "008-004", [], None, None),
+                AssemblyMove("008-004", "007-001", [], None, None),
+                AssemblyMove("007-001", "006-005", [], None, None),
+                AssemblyMove("006-005", "SFP", [], None, None),
+                AssemblyMove("009-045", "009-045", [], None, None, 60.0),
+                AssemblyMove("LoadQueue", "010-046", [0.0, 12.0, 14.0, 15.0, 0.0], "outer fuel", None),
+                AssemblyMove("010-046", "011-046", [], None, None),
+                AssemblyMove("011-046", "012-046", [], None, None),
+                AssemblyMove("012-046", "ExCore", [], None, None),
```
drewj-tp: Some of these fields are optional so we don't need to write them here. Save some characters

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -27,21 +27,56 @@
...
+        Type of assembly that is moving.
+    movingAssemName : str, optional
+        Name of the assembly moving (for SFP/ExCore interactions).
+    rotation : float, optional
+        Degrees of manual rotation to apply after shuffling.
+    """
+
+    fromLoc: str
+    toLoc: str
+    enrichList: List[float]
```
drewj-tp: We can make `enrichList` optional as well with `dataclasses.field(default_factory=list)`. From the [py docs](https://docs.python.org/3/library/dataclasses.html#dataclasses.field)
> If provided, it must be a zero-argument callable that will be called when a default value is needed for this field. Among other purposes, this can be used to specify fields with mutable default values, as discussed below. It is an error to specify both default and default_factory
```python
@dataclass
class C:
    mylist: list[int] = field(default_factory=list)

c = C()
c.mylist += [1, 2, 3]
```
This would improve readability but not really impact internal usage since we're almost always processing the entire cascade and providing these fields explicitly, even if they're empty

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -27,21 +27,56 @@
...
+    fromLoc : str
+        Original location label.
+    toLoc : str
+        Destination location label.
+    enrichList : list[float]
+        Axial U235 weight percent enrichment values for each block.
+    assemType : str, optional
+        Type of assembly that is moving.
+    movingAssemName : str, optional
+        Name of the assembly moving (for SFP/ExCore interactions).
```
drewj-tp: `movingAssemName` feels at odds with `toLoc`. Maybe "nameAtDischarge` ? because we only want to use this if we're discharging, and we're re-naming the assembly upon move?

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -91,7 +127,16 @@ def outage(self, factor=1.0):
             raise ValueError("Cannot perform two outages with same FuelHandler instance.")
 
         # determine if a repeat shuffle is occurring or a new shuffle pattern
-        if self.cs["explicitRepeatShuffles"]:
+        if self.cs[CONF_SHUFFLE_SEQUENCE_FILE]:
+            if not os.path.exists(self.cs[CONF_SHUFFLE_SEQUENCE_FILE]):
+                raise RuntimeError(
```
drewj-tp: The pattern is already to use a `RuntimeError` but I would also accept (and prefer) `FileNotFoundError`. Non-blocking

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
         return moves
 
+    @staticmethod
+    def readMovesYaml(fname):
+        """Read a shuffle file in YAML format."""
+        try:
+            from ruamel.yaml import YAML
+            from ruamel.yaml.constructor import DuplicateKeyError
+        except ImportError:
+            raise RuntimeError("ruamel.yaml is required to read YAML shuffle files")
```
drewj-tp: `ruamel.yaml` is part of the ARMI install process so

1. We definitely have this, don't need to guard against not having it
2. We should move the import to the top of file

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+
+        try:
+            with open(fname, "r") as stream:
+                yaml = YAML(typ="safe")
+                yaml.allow_duplicate_keys = False
+                data = yaml.load(stream)
+        except DuplicateKeyError:
+            raise InputError("Duplicate cycle in shuffle YAML sequence")
```
drewj-tp: As I mentioned in another comment, the yaml duplicate key error traceback is great. We should use that still. My preference would be to remove this logic
```suggestion
```
and then assert the `DuplicateKeyError` is raised in the testing.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+        except OSError:
+            raise RuntimeError(
+                "Could not find/open repeat shuffle file {} in working directory {}".format(fname, os.getcwd())
+            )
```
drewj-tp: Python has a pattern of (what I call) nested exceptions, e.g.,
```python
try:
    with open(...):
        yaml.load(...)
except OSError as ee:
    raise RuntimeError(...) from ee
```
This does two things
1. It captures the original traceback down to where in the code the original exception was raised, 
2. Allows us to raise a more user friendly exception, without losing the original information.

Non-blocking request to use this pattern here.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+
+            def _validateLoc(loc):
+                if loc in {"SFP", "ExCore", "LoadQueue"}:
+                    return
+                try:
+                    grids.locatorLabelToIndices(loc)
+                except Exception:
+                    raise InputError(f"Invalid location label {loc} in shuffle YAML")
+
+            for action in actions or []:
```
drewj-tp: Hmm `actions` will be None if there's nothing in this mapping. Could we skip this entire block above instead?
```python
for cycleKey, actions in data["sequence"].items():
    if actions is None:
        continue
```
Maybe give the user a warning / info message if there's an empty cycle too. Not sure if that's something we want to be a hard error or not

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+
+                    enrich = []
+                    if "fuelEnrichment" in action:
+                        enrichList = action["fuelEnrichment"]
+                        try:
+                            enrich = [float(e) for e in enrichList]
+                        except (TypeError, ValueError):
+                            raise InputError("fuelEnrichment values must be numeric")
+                        if any(e < 0 or e > 100 for e in enrich):
+                            raise InputError("fuelEnrichment values must be between 0 and 100")
```
drewj-tp: `enrichList = action.get("fuelEnrichment", [])` would also be sufficient here and it would reduce the indentation down a level

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+                    if not isinstance(swap, list) or len(swap) != 2:
+                        raise InputError("misloadSwap must be a list of two location labels")
+                    if any(not isinstance(item, str) for item in swap):
+                        raise InputError("misloadSwap entries must be strings")
+                    for loc in swap:
+                        _validateLoc(loc)
+                        seenLocs.add(loc)
+                    loc1, loc2 = swap
+                    moves[cycle].append(AssemblyMove(loc1, loc2, [], None, None))
+                    moves[cycle].append(AssemblyMove(loc2, loc1, [], None, None))
```
drewj-tp: In general, we can choose to not provide the optional arguments when making the `AssemblyMove` to save some characters and help us be flexible with future changes to the API, should they ever come

Please update to the following

1. For swaps, only provide the locations of the swapped assemblies, not empty enrichment, assem type, rotation
2. For rotations, only provide the rotation assembly and the rotation, not empty enrichment list nor assem type
3. Only the initial fresh move should have the enrichment list and assem type provided to the move constructor

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+                    moves[cycle].append(AssemblyMove("LoadQueue", locs[0], enrich, assemType, None))
+                    for i in range(len(locs) - 1):
+                        moves[cycle].append(AssemblyMove(locs[i], locs[i + 1], [], None, None))
+                    if locs[-1] not in ["SFP", "ExCore"]:
+                        moves[cycle].append(AssemblyMove(locs[-1], "SFP", [], None, None))
+
+                if "misloadSwap" in action:
+                    swap = action["misloadSwap"]
+                    if not isinstance(swap, list) or len(swap) != 2:
+                        raise InputError("misloadSwap must be a list of two location labels")
```
drewj-tp: We should tell people what the offending `swap` was to help with fixing input files
```suggestion
                        raise InputError(f"misloadSwap must be a list of two location labels, got {swap}")
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+                    if any(not isinstance(item, str) for item in swap):
+                        raise InputError("misloadSwap entries must be strings")
```
drewj-tp:
```suggestion
                        raise InputError(f"misloadSwap entries must be strings, got {swap}")
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+            moves[cycle] = []
+            seenLocs = set()
+
+            def _validateLoc(loc):
+                if loc in {"SFP", "ExCore", "LoadQueue"}:
+                    return
+                try:
+                    grids.locatorLabelToIndices(loc)
+                except Exception:
+                    raise InputError(f"Invalid location label {loc} in shuffle YAML")
```
drewj-tp: Please move this outside this function. Otherwise we're re-defining the function every iteration over the cycles in the yaml file

drewj-tp: I think future users would appreciate more information about the invalid location. But also, `locatorLabelToIndices` will only fail if it
1. finds something that can't be an integer,
2. is given an empty string

So those are probably things a user could search for in their text file.

Non-blocking request to add at least the current cycle being processed to this error message. IDK, what information would an analyst want to see if there's some invalid data in their input file, other than "hey we found this invalid data?"

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+                        raise InputError("misloadSwap entries must be strings")
+                    for loc in swap:
+                        _validateLoc(loc)
+                        seenLocs.add(loc)
+                    loc1, loc2 = swap
+                    moves[cycle].append(AssemblyMove(loc1, loc2, [], None, None))
+                    moves[cycle].append(AssemblyMove(loc2, loc1, [], None, None))
+
+                if "rotations" in action:
+                    for loc, angle in (action.get("rotations") or {}).items():
```
drewj-tp: Could also use `for loc, angle in action.get("rotations", {}).items()`

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+                    if any(not isinstance(item, str) for item in chain):
+                        raise InputError("cascade entries must be strings")
+
+                    assemType = chain[0]
+                    locs = chain[1:]
+                    for loc in locs:
+                        _validateLoc(loc)
+                        if loc not in {"SFP", "ExCore"}:
+                            if loc in seenLocs:
+                                raise InputError(f"Location {loc} appears in multiple cascades in cycle {cycle}")
```
drewj-tp: Non-blocking recommendation: combine these two conditions so we can reduce the indentation
```python
if loc not in {"SFP", "ExCore"} and loc in seenLocs:
    raise InputError(...)
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+                    raise InputError(f"Missing cycle {missing[0]} in shuffle sequence")
+                raise InputError(f"Missing cycles {missing} in shuffle sequence")
+
+        for cycleKey, actions in data["sequence"].items():
+            cycle = int(cycleKey)
+            moves[cycle] = []
+            seenLocs = set()
+
+            def _validateLoc(loc):
+                if loc in {"SFP", "ExCore", "LoadQueue"}:
```
drewj-tp: Does this imply `LoadQueue` could exist in the yaml shuffle logic file? I don't think we want to support that
```suggestion
                if loc in {"SFP", "ExCore"}:
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+                        if any(e < 0 or e > 100 for e in enrich):
+                            raise InputError("fuelEnrichment values must be between 0 and 100")
+
+                    moves[cycle].append(AssemblyMove("LoadQueue", locs[0], enrich, assemType, None))
+                    for i in range(len(locs) - 1):
+                        moves[cycle].append(AssemblyMove(locs[i], locs[i + 1], [], None, None))
+                    if locs[-1] not in ["SFP", "ExCore"]:
+                        moves[cycle].append(AssemblyMove(locs[-1], "SFP", [], None, None))
+
+                if "misloadSwap" in action:
```
drewj-tp: I think this logic supports putting misloadSwaps and cascades in the same map. We really have two pathways we want to support (presently, maybe a third specific for rotations)

```python
if "cascade" in action:
    # build the cascade
elif "misloadSwap" in action:
    # do the misload
elif "rotations" in action:
    # make the final rotations map
else:
    raise InputError(f"Unable to process {action} in {cycle}")
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+                    swap = action["misloadSwap"]
+                    if not isinstance(swap, list) or len(swap) != 2:
+                        raise InputError("misloadSwap must be a list of two location labels")
+                    if any(not isinstance(item, str) for item in swap):
+                        raise InputError("misloadSwap entries must be strings")
+                    for loc in swap:
+                        _validateLoc(loc)
+                        seenLocs.add(loc)
```
drewj-tp: Hmm I'm not sure we want to add these to `seenLocs`. If you always do `misloadSwaps` at the end of the file, we have no more cascades to process, so the updated `seenLocs` doesn't matter.

But if you re-arranged the example input to put the misloads before the cascades, e.g.,
```yaml
sequence:
  1:
    - misloadSwap: ["009-045", "008-004"]
    - cascade: ["outer fuel", "009-045", "008-004", "007-001", "006-005"]
      fuelEnrichment: [0, 12, 14, 15, 0]
      rotations: {"009-045": 60}
```
then you would have an error processing the cascade because we've "seen" `009-045` and `008-004` already.

If I'm understanding this correctly, I recommend removing this update to `seenLocs`
```suggestion
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -873,14 +927,16 @@ def swapCascade(self, assemList):
                 continue
             self.swapAssemblies(assemList[0], assemList[level + 1])
 
-    def repeatShufflePattern(self, explicitRepeatShuffles):
+    def repeatShufflePattern(self, shuffleFile, yaml=False):
```
drewj-tp: Calling this method `repeatShufflePattern` doesn't feel correct anymore. I believe the intent of the YAML file is to have users write that by hand not be produced by ARMI, so it's not a record of something to be repeated in the same way the explicit repeats file is.

So, what do? Simplest approach would be to rename this `performShuffle` and change various documentation (e.g., `"The file name that contains the shuffle history"`) to be more general, removing the assumption this is a shuffle someone has done before. This would be acceptable.

The internal logic isn't too hard that it would merit, at this point, two separate methods
1. `self.repeatShufflePattern(explicitRepeatShuffleFile)` and
2. `self.doYamlShuffle(shuffleFile)`

but that feels better. The only difference, really, is _how_ the moves list if being generated and we've got the two internal parsers for that already.

I'm gonna put this thought away but still would like to see this method renamed and documentation that aligns w/ shuffle "history" and "repeated shuffles" to be scrubbed except for when literally interacting with the explicit shuffle file

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
                 newLoc = m.group(2)
                 enrichList = [float(i) for i in m.group(3).split()]
                 # old loading style, just assume that there is a booster as our surrogate
-                moves[cycle].append((oldLoc, newLoc, enrichList, None))
+                moves[cycle].append(AssemblyMove(oldLoc, newLoc, enrichList, None))
```
drewj-tp: `assemType` is default `None` and we don't need it here
```suggestion
                moves[cycle].append(AssemblyMove(oldLoc, newLoc, enrichList))
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+        try:
+            from ruamel.yaml import YAML
+            from ruamel.yaml.constructor import DuplicateKeyError
+        except ImportError:
+            raise RuntimeError("ruamel.yaml is required to read YAML shuffle files")
+
+        try:
+            with open(fname, "r") as stream:
+                yaml = YAML(typ="safe")
+                yaml.allow_duplicate_keys = False
```
drewj-tp: `allow_duplicate_keys` is already false by default
```suggestion
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1181,7 +1349,14 @@ def processMoveList(self, moveList):
 
         # go through again, looking for stuff that isn't in chains.
         # put them in loop type 3 moves (arbitrary order)
-        for fromLoc, toLoc, _enrichList, assemType, _movingAssemName in moveList:
+        for move in moveList:
+            fromLoc = move.fromLoc
+            toLoc = move.toLoc
+            assemType = move.assemType
+            rot = move.rotation
```
drewj-tp: `assemType` and `rot` are not used in this loop
```suggestion
```

doc/user/inputs.rst
```diff
@@ -366,7 +366,41 @@ A few examples of restart cases:
 
 .. note:: The ``skipCycles`` setting is related to skipping the lattice physics calculation specifically, it is not required to do a restart run.
 
-.. note:: The X-SHUFFLES.txt file is required to do explicit repeated fuel management.
+.. note:: The ``explicitRepeatShuffles`` setting points to a ``*-SHUFFLES.txt``
+          file that records moves from a previous run for exact repetition.
+
+   Users may also define a custom shuffle plan in a YAML file referenced by the
+   ``shuffleSequenceFile`` setting. The YAML format organizes data by cycle in a
+   ``sequence`` mapping. Each cycle contains a list of high-level actions such as
```
drewj-tp: We should be explicit that the cycle keys are for the *start* of that cycle. I could see someone wanting to put a cycle `0` in the input (maybe we should explicitly throw an error there) because they imagine those changes happening during the outage that occurs in cycle 0.

It's kind of wishy/washy (is the end of cycle 0 or _between_ cycle 0 and 1) but since we treat this as a beginning-of-cycle action, we should be explicit about that. Something to the effect of
```rst
Keys are the cycle where the shuffling should occur during the beginning-of-cycle step. The first
available cycle where shuffling will occur is cycle 1
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1058,120 @@ def readMoves(fname):
...
+                    if len(chain) < 2:
+                        raise InputError("cascade must contain an assembly type and at least one location")
+                    if any(not isinstance(item, str) for item in chain):
+                        raise InputError("cascade entries must be strings")
+
+                    assemType = chain[0]
+                    locs = chain[1:]
+                    for loc in locs:
+                        _validateLoc(loc)
+                        if loc not in {"SFP", "ExCore"}:
```
drewj-tp: This pattern exists a lot in this file. To keep things [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), I would propose some [`frozenset`](https://docs.python.org/3/library/stdtypes.html#frozenset) for these "discharge locations". Perhaps attached as a class attribute on the fuel handler? Some benefits:

1. We don't need python to make the set for the conditional and then throw it away everytime
2. Every thing that needs to check if a location is a discharge location has a standard way of doing it
3. If/when we add new discharge locations or change what those magic strings are, we have one spot to do it

Perhaps
```python
class FuelHander:
    DISHARGE_LOCS = frozenset({"SFP", "ExCore"})
    """Special strings to indicate an assembly is no longer in the core."""
```
then write things like
```python
if loc in self.DISCHARGE_LOCS:
```
This is a non-blocking request but I think it would be great to see it. We can move it to a separate ticket if you want.

armi/settings/settingsValidation.py
HunterPSmith: addressed in [6068755](https://github.com/terrapower/armi/pull/2219/commits/60687556ef57f046b92e7992311634dad9a16223)

doc/user/inputs.rst
HunterPSmith: updated in [6068755](https://github.com/terrapower/armi/pull/2219/commits/60687556ef57f046b92e7992311634dad9a16223)

doc/user/inputs.rst
HunterPSmith: addressed in [1154831](https://github.com/terrapower/armi/pull/2219/commits/11548319801a7a662ee12e0dea41b47c0fef2e15)

armi/physics/fuelCycle/tests/test_fuelHandlers.py
HunterPSmith: this is just testing the yaml processor, we dont apply the rotations until after all moves within [`outage`](https://github.com/HunterPSmith/armi/blob/11548319801a7a662ee12e0dea41b47c0fef2e15/armi/physics/fuelCycle/fuelHandlers.py#L120)

doc/user/inputs.rst
HunterPSmith: yes, this is what we expect. updated with example in [5658ea7](https://github.com/terrapower/armi/pull/2219/commits/5658ea7490f6815db9c9025b37c796eb906808a1)

doc/user/inputs.rst
HunterPSmith: updated in [5658ea7](https://github.com/terrapower/armi/pull/2219/commits/5658ea7490f6815db9c9025b37c796eb906808a1)

doc/user/inputs.rst
HunterPSmith: fixed in [5658ea7](https://github.com/terrapower/armi/pull/2219/commits/5658ea7490f6815db9c9025b37c796eb906808a1)

doc/user/inputs.rst
HunterPSmith: added note in [5658ea7](https://github.com/terrapower/armi/pull/2219/commits/5658ea7490f6815db9c9025b37c796eb906808a1)

armi/tests/armiRun-SHUFFLES.yaml
HunterPSmith: switched it up [5658ea7](https://github.com/terrapower/armi/pull/2219/commits/5658ea7490f6815db9c9025b37c796eb906808a1)

armi/physics/fuelCycle/tests/test_fuelHandlers.py
HunterPSmith: removed in [a17ca4c](https://github.com/terrapower/armi/pull/2219/commits/a17ca4c8d7f5f17889f53fa6f26004986e4ca182)

armi/physics/fuelCycle/tests/test_fuelHandlers.py
HunterPSmith: same response as https://github.com/terrapower/armi/pull/2219#discussion_r2286021367

armi/physics/fuelCycle/tests/test_fuelHandlers.py
HunterPSmith: relying on yamls exception now [5e31e14](https://github.com/terrapower/armi/pull/2219/commits/5e31e14d111dface76b9a9b5f53a5de703f00269)

armi/physics/fuelCycle/tests/test_fuelHandlers.py
HunterPSmith: fixed ! [5e31e14](https://github.com/terrapower/armi/pull/2219/commits/5e31e14d111dface76b9a9b5f53a5de703f00269)

armi/physics/fuelCycle/tests/test_fuelHandlers.py
HunterPSmith: remvoed unused fields [5e31e14](https://github.com/terrapower/armi/pull/2219/commits/5e31e14d111dface76b9a9b5f53a5de703f00269)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: made enrichList optional! [5e31e14](https://github.com/terrapower/armi/pull/2219/commits/5e31e14d111dface76b9a9b5f53a5de703f00269)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: renamed to nameAtDischarge [5e31e14](https://github.com/terrapower/armi/pull/2219/commits/5e31e14d111dface76b9a9b5f53a5de703f00269)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: switched to FileNotFoundError [5e31e14](https://github.com/terrapower/armi/pull/2219/commits/5e31e14d111dface76b9a9b5f53a5de703f00269)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: cleaned this up [5e31e14](https://github.com/terrapower/armi/pull/2219/commits/5e31e14d111dface76b9a9b5f53a5de703f00269)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: utilizing yamls traceback now, [5e31e14](https://github.com/terrapower/armi/pull/2219/commits/5e31e14d111dface76b9a9b5f53a5de703f00269)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: added nested exceptions [5e31e14](https://github.com/terrapower/armi/pull/2219/commits/5e31e14d111dface76b9a9b5f53a5de703f00269)

armi/physics/fuelCycle/tests/test_fuelHandlers.py
HunterPSmith: actually I readded. it is necessary in order to view the large diffs (otherwise it hides the diffs).

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: added warning [03f286c](https://github.com/terrapower/armi/pull/2219/commits/03f286c74397c70f425370ac415263f388bed542)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: implemented in [03f286c](https://github.com/terrapower/armi/pull/2219/commits/03f286c74397c70f425370ac415263f388bed542)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: cleaned this up in [03f286c](https://github.com/terrapower/armi/pull/2219/commits/03f286c74397c70f425370ac415263f388bed542)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: added more to error message [03f286c](https://github.com/terrapower/armi/pull/2219/commits/03f286c74397c70f425370ac415263f388bed542)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: [03f286c](https://github.com/terrapower/armi/pull/2219/commits/03f286c74397c70f425370ac415263f388bed542)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: moved out [03f286c](https://github.com/terrapower/armi/pull/2219/commits/03f286c74397c70f425370ac415263f388bed542)

HunterPSmith: added cycle to error message [03f286c](https://github.com/terrapower/armi/pull/2219/commits/03f286c74397c70f425370ac415263f388bed542)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: [6a93d61](https://github.com/terrapower/armi/pull/2219/commits/6a93d61f9dd1f3c715513def9c0c2cbab85e6991)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: reduced indentation [6a93d61](https://github.com/terrapower/armi/pull/2219/commits/6a93d61f9dd1f3c715513def9c0c2cbab85e6991)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: removed "loadqueue" [6a93d61](https://github.com/terrapower/armi/pull/2219/commits/6a93d61f9dd1f3c715513def9c0c2cbab85e6991)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: updated structure here [6a93d61](https://github.com/terrapower/armi/pull/2219/commits/6a93d61f9dd1f3c715513def9c0c2cbab85e6991)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: totally agree. removed [6a93d61](https://github.com/terrapower/armi/pull/2219/commits/6a93d61f9dd1f3c715513def9c0c2cbab85e6991)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: updated to performShuffle. [74a6e89](https://github.com/terrapower/armi/pull/2219/commits/74a6e891a16a746c30c597ac7f241666500746db)

also should we change doRepeatShuffle to doShuffle?

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: updated [e79d75e](https://github.com/terrapower/armi/pull/2219/commits/e79d75ee99726a9c3a4b0c833bef11b4e102e94f)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: [e79d75e](https://github.com/terrapower/armi/pull/2219/commits/e79d75ee99726a9c3a4b0c833bef11b4e102e94f)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: [e79d75e](https://github.com/terrapower/armi/pull/2219/commits/e79d75ee99726a9c3a4b0c833bef11b4e102e94f)

doc/user/inputs.rst
HunterPSmith: updated in[e79d75e](https://github.com/terrapower/armi/pull/2219/commits/e79d75ee99726a9c3a4b0c833bef11b4e102e94f) 

now throws an input error if user inputs a cycle 0 in the yaml

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: did this refactoring [095dc49](https://github.com/terrapower/armi/pull/2219/commits/095dc4955921418c55fec0ae285a088781b121ea)

armi/reactor/cores.py
HunterPSmith: moved to the top [be47731](https://github.com/terrapower/armi/pull/2219/commits/be47731e8a8f52dfd962beaf34049e516aa72dae)

armi/physics/fuelCycle/tests/test_fuelHandlers.py
drewj-tp: huh! today I learned something - https://docs.python.org/3/library/unittest.html#unittest.TestCase.maxDiff

doc/user/inputs.rst
```diff
@@ -371,36 +371,58 @@ A few examples of restart cases:
...
+             fuelEnrichment: [0, 12, 14, 15, 0]
+
+.. note:: Consider using yaml anchors ``&`` and aliases ``*`` to reduce repetition.
+
+For cycle 1 above, the actions execute in the following order:
+
+   1. The assembly originally at ``008-004`` is discharged to the spent fuel pool ``SFP``.
+   2. The assembly originally at ``009-045`` moves to ``008-004``.
+   3. A fresh ``outer fuel`` assembly is created with the specified axial enrichment profile and inserted at ``009-045``.
+   4. The fresh assembly and the moved assembly are swapped, leaving the fresh assembly at ``008-004`` and the moved assembly back at ``009-045``.
```
drewj-tp:
```suggestion
   4. The fresh assembly and the assembly at ``008-004`` are swapped, leaving the fresh assembly at ``008-004`` and the other assembly back at ``009-045``.
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1130,45 +1159,44 @@ def _validateLoc(loc):
...
-                            raise InputError("fuelEnrichment values must be numeric")
-                        if any(e < 0 or e > 100 for e in enrich):
-                            raise InputError("fuelEnrichment values must be between 0 and 100")
-
-                    moves[cycle].append(AssemblyMove("LoadQueue", locs[0], enrich, assemType, None))
+                    enrichList = action.get("fuelEnrichment", [])
+                    try:
+                        enrich = [float(e) for e in enrichList]
+                    except (TypeError, ValueError):
+                        raise InputError("fuelEnrichment values must be numeric")
```
drewj-tp:
```suggestion
                        raise InputError(f"fuelEnrichment values must be numeric. Got {enrichList}")
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1130,45 +1159,44 @@ def _validateLoc(loc):
...
+                    if any(e < 0 or e > 100 for e in enrich):
+                        raise InputError("fuelEnrichment values must be between 0 and 100")
```
drewj-tp:
```suggestion
                        raise InputError(f"fuelEnrichment values must be between 0 and 100. Got {enrich}")
```

drewj-tp (changes requested):
Some very minor comments requests. Thank you so much for the update 🏆 

Let me know how you want to proceed re "integration" testing

armi/reactor/cores.py
HunterPSmith: actually that caused import loop failure I believe, reverting commit [6ecab27](https://github.com/terrapower/armi/pull/2219/commits/6ecab279ff8eaae5bb211abb8d56c313a670d258)

doc/user/inputs.rst
HunterPSmith: [365c5a8](https://github.com/terrapower/armi/pull/2219/commits/365c5a8dfcb2ac6c6d44a956abcf9404dc9d3bf0)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: [365c5a8](https://github.com/terrapower/armi/pull/2219/commits/365c5a8dfcb2ac6c6d44a956abcf9404dc9d3bf0)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: [365c5a8](https://github.com/terrapower/armi/pull/2219/commits/365c5a8dfcb2ac6c6d44a956abcf9404dc9d3bf0)

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -661,41 +662,94 @@ def test_readMovesYaml(self):
...
+    def test_performShuffleYamlIntegration(self):
+        fh = fuelHandlers.FuelHandler(self.o)
+        yaml_text = (
+            "sequence:\n"
+            "  1:\n"
+            "    - misloadSwap: [\"005-023\", \"006-029\"]\n"
+            "    - cascade: [\"igniter fuel\", \"009-045\", \"008-004\", \"007-001\", \"006-005\"]\n"
+            "      fuelEnrichment: [0, 0.12, 0.14, 0.15, 0]\n"
+            "    - extraRotations: {\"009-045\": 60}\n"
+        )
```
drewj-tp: This test is excellent. Thank you very much for adding it!

I cleaned up the readability by writing something like
```suggestion
        yaml_text = """
sequence:
  1:
    - misloadSwap: ["005-023", "006-029"]
    - cascade: ["igniter fuel", "009-045", "008-004", "007-001", "006-005"]
      fuelEnrichment: [0, 0.12, 0.14, 0.15, 0]
    - extraRotations: {"009-045": 60}
"""
```

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -661,41 +662,94 @@ def test_readMovesYaml(self):
...
             ],
         }
         self.assertEqual(moves, expected)
 
+    def test_performShuffleYamlIntegration(self):
+        fh = fuelHandlers.FuelHandler(self.o)
+        yaml_text = (
+            "sequence:\n"
+            "  1:\n"
+            "    - misloadSwap: [\"005-023\", \"006-029\"]\n"
```
drewj-tp: So I got curious and wanted to know why this differs from the example file and does not swap two assemblies in a shuffle cascade.

And that failed, and I don't think it should. But I can't immediately find out why.

If I change this to swapping `009-045` and `008-004`, the `assemblyList` in `FuelHandler.doRepeatShuffle` is
```
[<igniter fuel Assembly A0045 at 008-004>, <igniter fuel Assembly A0031 at 007-001>, <feed fuel Assembly A0052 at 009-045>, <radial shield Assembly A0004 at 006-005>, <radial shield Assembly A0004 at 006-005>, <feed fuel Assembly A0052 at 009-045>]
```

If I keep the `misloadSwap` as is, the `assemblyList` is
```
[<igniter fuel Assembly A0045 at 006-005>, <igniter fuel Assembly A0031 at 007-001>, <feed fuel Assembly A0052 at 008-004>, <radial shield Assembly A0004 at 009-045>]
```

I say this because I think we want the ability to swap assemblies in a shuffle, our example docs in this PR call that out, but I don't think it's working as advertised

drewj-tp: Could be something about trying to swap a fuel and shield assembly. Investigating

HunterPSmith: the issue has to do with misloadSwaps and cascades and how it works with ``trackChain``. The YAML parser currently builds a single move list that contains both cascade moves and the misload swap. When ``trackChain`` reconstructs a load chain, it appends the origin of every move whose destination matches the current location, without stopping after the first match. Because both the cascade and the misload swap contain moves to "008-004", the chain ends with a duplicated "009-045" entry.  I fixed this in [11a26e7](https://github.com/terrapower/armi/pull/2219/commits/11a26e75455cd04589873c3912d13aa3c02c8320)

I basically separated out the misload swaps from the movesList to fix it.

armi/physics/fuelCycle/tests/test_fuelHandlers.py
HunterPSmith: fixed in [11a26e7](https://github.com/terrapower/armi/pull/2219/commits/11a26e75455cd04589873c3912d13aa3c02c8320)

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1199,7 +1212,7 @@ def readMovesYaml(fname):
                 else:
                     raise InputError(f"Unable to process {action} in {cycle}")
 
-        return moves
+        return moves, dict(misloadSwaps)
```
drewj-tp: Please consider adding a description on the return type since now we have a slightly more complicated return pattern

drewj-tp (approved):
I'm very happy with the state of this PR. Thank you so much for working with me through this.

I have one minor comment that is non-blocking

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: [cb0941b](https://github.com/terrapower/armi/pull/2219/commits/cb0941ba48c690206df009420ac47d51eeb5c4ab)

doc/user/inputs.rst
```diff
@@ -366,7 +366,63 @@ A few examples of restart cases:
 
 .. note:: The ``skipCycles`` setting is related to skipping the lattice physics calculation specifically, it is not required to do a restart run.
 
-.. note:: The X-SHUFFLES.txt file is required to do explicit repeated fuel management.
+.. note:: The ``explicitRepeatShuffles`` setting points to a ``*-SHUFFLES.txt``
```
john-science, line 372:
```suggestion
Fuel Shuffling
^^^^^^^^^^^^^^

.. note:: The ``explicitRepeatShuffles`` setting points to a ``*-SHUFFLES.txt``
```

john-science: This may require some other edits to the order of all these `notes` around here, but this should be its own section. This is too much info to just be a bullet under restarts. IMO

armi/settings/settingsValidation.py
```diff
@@ -431,6 +431,33 @@ def _willBeCopiedFrom(fName):
...
+            lambda: self.cs["shuffleSequenceFile"]
```
john-science: If you _can_, it would be nice to use the global `CONF_SHUFFLE_SEQUENCE_FILE` variables for the settings in these validators. If you aren't using them because of import loops, that's fine.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1104,133 @@ def readMoves(fname):
...
+        -------
+        moves : dict
+            Mapping of cycle numbers to lists of
+            :class:`~armi.physics.fuelCycle.fuelHandlers.AssemblyMove` objects that
+            describe the shuffle sequence.
+        misloadSwaps : dict
+            Mapping of cycle numbers to lists of location-pair tuples describing
+            assemblies to be swapped in order to simulate a misload.
+        """
+        try:
```
john-science, line 1136:
```suggestion
        # 1. load YAML file
        try:
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1104,133 @@ def readMoves(fname):
...
+                yaml = YAML(typ="safe")
+                data = yaml.load(stream)
+        except DuplicateKeyError as e:
+            raise InputError(str(e)) from e
+        except OSError as ee:
+            raise RuntimeError(
+                f"Could not find/open repeat shuffle file {fname!r} in working directory {os.getcwd()}: {ee}"
+            ) from ee
+
+        if "sequence" not in data:
```
john-science, line 1148:
```suggestion
        # 2. perform various validation tests on the YAML data
        if "sequence" not in data:
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1104,133 @@ def readMoves(fname):
...
+                        moves[cycle].append(AssemblyMove(locs[-1], "SFP"))
+
+                elif "misloadSwap" in action:
+                    swap = action["misloadSwap"]
+                    if not isinstance(swap, list) or len(swap) != 2:
+                        raise InputError("misloadSwap must be a list of two location labels, got {swap}")
+                    if any(not isinstance(item, str) for item in swap):
+                        raise InputError("misloadSwap entries must be strings, got {swap}")
+                    for loc in swap:
+                        FuelHandler._validateLoc(loc, cycle)
```
john-science: We are accessing `_validateLoc()` outside the FuelHandler? Than this should be a public state method, not a private method.

Can we change that? You added that in this PR, right?

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -997,14 +1104,133 @@ def readMoves(fname):
...
+        cycleNums = {int(c) for c in data["sequence"].keys()}
+        if cycleNums:
+            expected = set(range(min(cycleNums), max(cycleNums) + 1))
+            missing = sorted(expected - cycleNums)
+            if missing:
+                if len(missing) == 1:
+                    raise InputError(f"Missing cycle {missing[0]} in shuffle sequence")
+                raise InputError(f"Missing cycles {missing} in shuffle sequence")
+
+        for cycleKey, actions in data["sequence"].items():
```
john-science, line 1164:
```suggestion
        # 3. parse YAML file into shuffle data
        for cycleKey, actions in data["sequence"].items():
```

john-science: The idea here with these three little suggestions is this method is just too long. We could break it up into smaller, helper methods. But I know you are on a deadline. I think the quicker solution is to add some high-level bullet-point comments.

I'm willing to hear other ideas, of course.

HunterPSmith: added the comments! 

[ceec293](https://github.com/terrapower/armi/pull/2219/commits/ceec29390618d08d972964c41880182f276673cb)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: moved to public [ceec293](https://github.com/terrapower/armi/pull/2219/commits/ceec29390618d08d972964c41880182f276673cb)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: [ceec293](https://github.com/terrapower/armi/pull/2219/commits/ceec29390618d08d972964c41880182f276673cb)

armi/physics/fuelCycle/fuelHandlers.py
HunterPSmith: [ceec293](https://github.com/terrapower/armi/pull/2219/commits/ceec29390618d08d972964c41880182f276673cb)

armi/settings/settingsValidation.py
HunterPSmith: [ceec293](https://github.com/terrapower/armi/pull/2219/commits/ceec29390618d08d972964c41880182f276673cb)

doc/user/inputs.rst
HunterPSmith: [ceec293](https://github.com/terrapower/armi/pull/2219/commits/ceec29390618d08d972964c41880182f276673cb)

PR #2223, opened 2025-08-12
Move the `zonesFile` setting to the framework and add building of zones to the interface stack

keckler:
## SCR Information

One-Sentence Description: TBD

---

## Checklist

### DIFF
--- a/armi/reactor/cores.py
+++ b/armi/reactor/cores.py
@@ -2167,8 +2167,8 @@ class Core(composites.Composite):
 
     def buildManualZones(self, cs):
         """
-        Build the Zones that are defined manually in the given Settings file, in the
-        `zoneDefinitions` setting.
+        Build the Zones that are defined in the given Settings, in the
+        `zoneDefinitions` or `zonesFile` case setting.
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


keckler:
For my consideration -- do I need to do anything with copyInterfaceFiles?

armi/settings/settingsValidation.py
```diff
@@ -346,6 +346,7 @@ def _inspectSettings(self):
             CONF_XS_KERNEL,
             CONF_XS_SCATTERING_ORDER,
         )
+        from armi.settings.fwSettings.globalSettings import CONF_ZONE_DEFINITIONS, CONF_ZONES_FILE
```
john-science, line 349: This can't be at the top of the file because it causes an import loop?

keckler: That's correct.

This particular module is a bit wonky. It serves both to define the machinery which allows the settings validation, and then also proceeds to use that machinery and do a bunch of settings validation. But doing the settings validation from within that module is an anti-pattern, because it requires a bunch of circular imports due to the fact that the actual settings are defined in other modules.

Ideally, the settings validation would be moved to the associated modules which define the settings themselves.

armi/reactor/cores.py
```diff
@@ -2189,20 +2192,40 @@ def buildManualZones(self, cs):
...
+                self.zones.addZone(zone)
+
+        elif cs[CONF_ZONES_FILE]:
+            runLog.info(f"Custom zoning strategy applied from {CONF_ZONES_FILE}.")
+
+            self.zones = Zones()
+            with open(cs[CONF_ZONES_FILE]) as stream:
+                zonesDict = YAML(typ="safe").load(stream)
+
+            for assemblyLocation, zoneName in zonesDict["customZonesMap"].items():
```
john-science: Does this "zone" logic only work with pin-type reactors?

If now, can we just change "assemblyLocation" to "location"? We have other reactors around that aren't pin-type.

john-science (approved):
I have one outstanding request, but it's small and not a show stopper.

armi/reactor/cores.py
keckler: Done

PR #2225, opened 2025-08-13
Advancing r.p.time in the Operator

john-science:

For some reason, in the past, the Operator advanced `r.p.cycle`, `r.p.node`, and `r.p.timeNode`, but not `r.p.time`. This keeps surprising my power users, which is a sign it is not a good design. And I agree, it was awkward that ARMI did not contain the code to advance time.

## SCR Information

Change Type: features

One-Sentence Description: Advancing r.p.time in the Operator.

One-line Impact on Requirements: A minor change to `I_ARMI_OPERATOR_COMM` for `R_ARMI_OPERATOR_COMM`, in that we are adding the feature that the `Operator` now advance time.

---

## Checklist


armi/operators/operator.py
```diff
@@ -379,6 +357,7 @@ def _cycleLoop(self, cycle, startingCycle):
         self.r.p.availabilityFactor = self.availabilityFactors[cycle]
         self.r.p.cycle = cycle
         self.r.core.p.coupledIteration = 0
+        # self.r.p.time += self.r.p.cycleLength * (1.0 - self.r.p.availabilityFactor) / units.DAYS_PER_YEAR  # TODO
```
keckler: Should this be deleted?

armi/operators/operator.py
```diff
@@ -415,6 +394,7 @@ def _cycleLoop(self, cycle, startingCycle):
     def _timeNodeLoop(self, cycle, timeNode):
         """Run the portion of the main loop that happens each subcycle."""
         self.r.p.timeNode = timeNode
+        self.r.p.time += self.r.o.stepLengths[cycle][timeNode - 1] / units.DAYS_PER_YEAR  # TODO: JOHN
```
keckler: Why TODO?

armi/operators/operator.py
```diff
@@ -622,6 +601,10 @@ def interactAllEveryNode(self, cycle, tn, excludedInterfaceNames=()):
 
     def interactAllEOC(self, cycle, excludedInterfaceNames=()):
         """Interact end of cycle for all enabled interfaces."""
+        # TODO: JOHN
```
keckler: TODO

armi/operators/operator.py
john-science: whoops

john-science: Removed my dev crumb. Sorry.

armi/operators/operator.py
john-science: Removed my dev crumb. Sorry.

armi/operators/operator.py
john-science: Removed my dev crumb. Sorry.

armi/operators/operator.py
```diff
@@ -622,6 +602,9 @@ def interactAllEveryNode(self, cycle, tn, excludedInterfaceNames=()):
 
     def interactAllEOC(self, cycle, excludedInterfaceNames=()):
         """Interact end of cycle for all enabled interfaces."""
+        if not math.isclose(self.r.p.availabilityFactor, 1):
+            self.r.p.time += self.r.p.cycleLength * (1 - self.r.p.availabilityFactor) / units.DAYS_PER_YEAR
```
keckler: I do not think there is any benefit to wrapping this in an `if` statement. Just update `time` every time, no matter what `availabilityFactor` is.

keckler (approved):
Great! Do we require any new requirement on the framework for this?

armi/operators/operator.py
john-science: I just copied this over. But that's an interesting point.

john-science: I would guess this exists so that if there is a floating point rounding error and the `r.p.availabilityFactor` is `0.9999999999995`, it is actually `1.0` and we don't want to change the `r.p.time` by one microsecond.

keckler: I think the `if` was there in the place that you copied the code from because in that circumstance, a call to the depletion solver was also wrapped in the same `if`, which would be expensive to call for such a tiny time step.

I don't feel super strongly, but I am 99% sure that this `if` statement can be removed.

john-science: Okay, sure. Done.

PR #2226, opened 2025-08-14
Fix equality of MultiIndexLocator and CoordinateLocation

### LINKED ISSUE #2224: MultiIndexLocations with different locations can be equal
Stumbled across this today. If you have two multi index locations, they can be considered equal even if their locations are different. I found this with a test that has two fuel pins with different multiplicities, but somehow the spatial locators were equal. Despite these existing at different lattice sites within the block.

Here's a reproducible example
```python
>>> from armi.reactor.grids import IndexLocation, MultiIndexLocation
>>> a = MultiIndexLocation(None)
>>> a.append(IndexLocation(0, 0, 0, None))
>>> b = MultiIndexLocation(None)
>>> b.append(IndexLocation(1, 1, 1, None))
>>> a == b
True
```

---

If you also compare the spatial locator of a fuel pin with multiplicity > 1 and something that does not live on the spatial grid, like the duct, you get the spatial locators are also equal.

```python
>>> import armi
>>> armi.configure()
>>> _, r = loadTestReactor()
>>> from armi.reactor.flags import Flags
>>> b = r.core.getFirstBlock(Flags.FUEL)
>>> list(b)
[<Circle: fuel>,
 <Circle: bond>,
 <Circle: clad>,
 <Helix: wire>,
 <DerivedShape: coolant>,
 <Hexagon: duct>,
 <Hexagon: intercoolant>]
>>> b[0].spatialLocator
<MultiIndexLocation with 169 locations>
>>> b[-1].spatialLocator
<CoordinateLocation @ (0.0,0.0,0.0)>
>>> b[0].spatialLocator == b[-1].spatialLocator
True
```
---

The root seems to be the shared inheritance of `LocationBase` which defines equality

https://github.com/terrapower/armi/blob/35b3dcc6834aa82c18a1f7c261e1a0f988119d4f/armi/reactor/grids/locations.py#L98-L103

But a `MultIndexLocation` always sends `(i, j, k) = (0, 0, 0)` at construction

https://github.com/terrapower/armi/blob/35b3dcc6834aa82c18a1f7c261e1a0f988119d4f/armi/reactor/grids/locations.py#L339-L341

Any two `MultiIndexLocation`s on the same grid will be considered equal.

drewj-tp:

Closes #2224 

- `MultiIndexLocator` objects are equal if their grids are equal and all their locations are equal, including ordering.
- `CoordinateLocation` objects are equal if their grids are equal and their misnamed `(i, j, k)` coordinate locations (real positions in space, not an index like ijk would imply) are equal

## SCR Information

Change Type: fixes

One-Sentence Description: Fix equality of `MultiIndexLocator` and `CoordinateLocation`

One-line Impact on Requirements: Strengthens behavior related to `R_ARMI_GRID_MULT` but does not directly modify that requirement.

---

## Checklist

### DIFF
--- a/armi/reactor/grids/locations.py
+++ b/armi/reactor/grids/locations.py
@@ -435,6 +435,10 @@ class CoordinateLocation(IndexLocation):
             return self.grid == other.grid and self.i == other.i and self.j == other.j and self.k == other.k
         return NotImplemented
 
+    def __hash__(self):
+        """Hash based on the coordinates but not the grid."""
+        return hash((self.i, self.j, self.k))
+
     def getLocalCoordinates(self, nativeCoords=False):
         """Return x,y,z coordinates in cm within the grid's coordinate system."""
         return self.indices
--- a/armi/reactor/grids/locations.py
+++ b/armi/reactor/grids/locations.py
@@ -340,6 +340,17 @@ class MultiIndexLocation(IndexLocation):
         IndexLocation.__init__(self, 0, 0, 0, grid)
         self._locations = []
 
+    def __eq__(self, other):
+        """Considered equal if the grids are identical and contained locations are identical.
+
+        Two ``MultiIndexLocation`` objects with the same total collection of locations, but in
+        different orders, will not be considered equal.
+        """
+        if isinstance(other, type(self)):
+            return self.grid == other.grid and self._locations == other._locations
+        # Different objects -> let other.__eq__(self) handle it
+        return NotImplemented
+
     def __getstate__(self) -> List[IndexLocation]:
         """Used in pickling and deepcopy, this detaches the grid."""
         return self._locations
--- a/armi/reactor/grids/locations.py
+++ b/armi/reactor/grids/locations.py
@@ -427,6 +427,14 @@ class CoordinateLocation(IndexLocation):
 
     __slots__ = ()
 
+    def __eq__(self, other):
+        if isinstance(other, type(self)):
+            # Mainly to avoid comparing against MultiIndexLocations
+            # Fuel pins may have a multi index location and the duct may
+            # have a coordinate location and we don't want them to be equal
+            return self.grid == other.grid and self.i == other.i and self.j == other.j and self.k == other.k
+        return NotImplemented
+
     def getLocalCoordinates(self, nativeCoords=False):
         """Return x,y,z coordinates in cm within the grid's coordinate system."""
         return self.indices
--- a/armi/reactor/grids/tests/test_grids.py
+++ b/armi/reactor/grids/tests/test_grids.py
@@ -84,6 +84,10 @@ class TestSpatialLocator(unittest.TestCase):
         b.append(grids.IndexLocation(1, 1, 1, None))
         self.assertNotEqual(a, b)
 
+        c = grids.MultiIndexLocation(None)
+        c.append(grids.IndexLocation(0, 0, 0, None))
+        self.assertEqual(a, c)
+
     def test_multiIndexEqWithLocations(self):
         """Two multi index locators on the same grid are equal."""
         grid = MockStructuredGrid()
--- a/armi/reactor/grids/tests/test_grids.py
+++ b/armi/reactor/grids/tests/test_grids.py
@@ -76,6 +76,14 @@ class TestSpatialLocator(unittest.TestCase):
         loc2 = grids.IndexLocation(2, 2, 0, None)
         self.assertEqual(loc1 + loc2, grids.IndexLocation(3, 4, 0, None))
 
+    def test_multiIndexEq(self):
+        """Check multi index locations are only true if they live on the same grid and have the same locations."""
+        a = grids.MultiIndexLocation(None)
+        a.append(grids.IndexLocation(0, 0, 0, None))
+        b = grids.MultiIndexLocation(None)
+        b.append(grids.IndexLocation(1, 1, 1, None))
+        self.assertNotEqual(a, b)
+
     def test_recursion(self):
         """
         Make sure things work as expected with a chain of locators/grids/locators.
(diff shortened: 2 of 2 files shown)


armi/reactor/grids/locations.py
```diff
@@ -416,6 +427,18 @@ class CoordinateLocation(IndexLocation):
 
     __slots__ = ()
 
+    def __eq__(self, other):
+        if isinstance(other, type(self)):
+            # Mainly to avoid comparing against MultiIndexLocations
+            # Fuel pins may have a multi index location and the duct may
+            # have a coordinate location and we don't want them to be equal
+            return self.grid == other.grid and self.i == other.i and self.j == other.j and self.k == other.k
+        return NotImplemented
```
john-science, line 436:
```suggestion
        return False
```

john-science: Am I missing something here?

If they are different types, then they are not equal, right?

armi/reactor/grids/locations.py
```diff
@@ -340,6 +340,17 @@ def __init__(self, grid: "Grid"):
...
+    def __eq__(self, other):
+        """Considered equal if the grids are identical and contained locations are identical.
+
+        Two ``MultiIndexLocation`` objects with the same total collection of locations, but in
+        different orders, will not be considered equal.
+        """
+        if isinstance(other, type(self)):
+            return self.grid == other.grid and self._locations == other._locations
+        # Different objects -> let other.__eq__(self) handle it
+        return NotImplemented
```
john-science, line 352:
```suggestion
        return False
```

john-science: Am I missing something?

If they are different types, then they are not equal, right?

armi/reactor/grids/locations.py
```diff
@@ -416,6 +427,18 @@ class CoordinateLocation(IndexLocation):
...
+        if isinstance(other, type(self)):
+            # Mainly to avoid comparing against MultiIndexLocations
+            # Fuel pins may have a multi index location and the duct may
+            # have a coordinate location and we don't want them to be equal
+            return self.grid == other.grid and self.i == other.i and self.j == other.j and self.k == other.k
+        return NotImplemented
+
+    def __hash__(self):
+        """Hash based on the coordinates but not the grid."""
+        return hash((self.i, self.j, self.k))
```
john-science, line 440: Out of curiosity, why is this needed?

armi/reactor/grids/tests/test_grids.py
```diff
@@ -76,6 +76,50 @@ def test_add(self):
         loc2 = grids.IndexLocation(2, 2, 0, None)
         self.assertEqual(loc1 + loc2, grids.IndexLocation(3, 4, 0, None))
 
+    def test_multiIndexEq(self):
+        """Check multi index locations are only true if they live on the same grid and have the same locations."""
+        a = grids.MultiIndexLocation(None)
+        a.append(grids.IndexLocation(0, 0, 0, None))
+        b = grids.MultiIndexLocation(None)
+        b.append(grids.IndexLocation(1, 1, 1, None))
+        self.assertNotEqual(a, b)
```
john-science, line 85:
```suggestion
        self.assertNotEqual(a, b)

        c = grids.MultiIndexLocation(None)
        c.append(grids.IndexLocation(0, 0, 0, None))
        self.assertEqual(a, c)
```

john-science: My thought here is that if we are testing the "not equal" situation, we should also test the "equal" situation.

armi/reactor/grids/locations.py
drewj-tp: This will end up there eventually.

Consider comparing a list to a set. `[1] == {1, }` is False. Makes sense: how would you compare a list to a set? But, if we check `[1].__eq__({1, })`, we get `NotImplemented`. 

`NotImplemented` is a special signal for python that "`a` doesn't know how to compare to `b`, ask if `b` knows how to compare to `a`. So if `a == b` "fails" by returning `NotImplemented`, then we check `b == a`. If that also returns `NotImplemented`, we get `False`. Same for other comparative operators, if `a < b` fails, ask `b > a`. Otherwise fail.

If I have a list-like class, it will never be able to compare equal to a list, even if they're holding the same objects. Here's an example.
```python
>>> class MyList:
...     def __init__(self, *items):
...         self._items = list(items)
...     def __repr__(self):
...         return f"MyList({self._items})"
...     def __eq__(self, other):
...         return self._items == other
...
>>> ml = MyList(1)
>>> ml
MyList([1])
>>> [1].__eq__(ml)
NotImplemented
>>> [1] == ml
True
```

john-science: Hmm... I think `a == b` returning `False` makes more sense than raising a `NotImplemented` error. The `==` and `__eq__` operators are meant to return a boolean. This is how the equality operator works in all types defined in the Python standard library, and I think that makes it the more standard and predictable approach:

```python
>>> 1 == "1"
False
>>> 1 == [1]
False
>>> {1:1} == "1"
False
>>> (1, 1) == {1:1}
False
```

If you like, you can put `isinstance(a, type(b))` in the type check, to cover the case where one is a subclass of the other and they are more comparable.

armi/reactor/grids/locations.py
john-science: Ok, we talked about this.

armi/reactor/grids/locations.py
drewj-tp: We talked a bit about the nuance of `list == set` and `list.__eq__(set)`. I think we landed on keeping `return NotImplemented`

armi/reactor/grids/locations.py
drewj-tp: See thread - https://github.com/terrapower/armi/pull/2226#discussion_r2279646049

PR #2227, opened 2025-08-15
Symmetry testing

bsculac:

This PR adds an easy way for plugin developers to audit their parameters to determine if said parameters are symmetry-aware or not. A test fixture is added that takes some attribute configuration and will run pre-written tests comparing parameter values before and after a partial to full core transformation.

## SCR Information

Change Type: features

One-Sentence Description: Add test fixture to allow plugins to check if parameters are symmetry-aware.

---

## Checklist


armi/testing/symmetryTesting.py
```diff
@@ -0,0 +1,324 @@
...
+        return {param: obj.p[param] for param in paramList}
+
+    @staticmethod
+    def _getParamNamesFromDefs(pdefs: parameters.ParameterDefinitionCollection):
+        return set([p.name for p in pdefs])
+
+    def _loadDefaultParameters(self):
+        self.defaultCoreParameterDefs = set(reactorParameters.defineCoreParameters())
+        self.defaultAssemblyParameterDefs = set(assemblyParameters.getAssemblyParameterDefinitions())
+        self.defaultBlockParameterDefs = set(blockParameters.getBlockParameterDefinitions())
```
mgjarrett: What would this be used for?

armi/testing/symmetryTesting.py
```diff
@@ -0,0 +1,324 @@
...
+        self.defaultParameterValue = 2
+        # load default armi parameters for each object type
+        self._loadDefaultParameters()
+        # some parameters have validation on their inputs and need specific settings
+        self.parameterOverrides = parameterOverrides
+
+        self.testObject = testObject
+        self.expectedSymmetricCoreParams = expectedSymmetricCoreParams
+        self.expectedSymmetricAssemblyParams = expectedSymmetricAssemblyParams
+        self.expectedSymmetricBlockParams = expectedSymmetricBlockParams
```
mgjarrett: It's confusing to have the same set of parameter names `expectedSymmetric[object]Params` at both the `BasicArmiSymmetryTestHelper` level and also one level below it, in `SymmetryFactorTester`. 

The one at the `SymmetryFactorTester` level seems like it would be better described by a name like `blockParamsToTest` rather than `expectedSymmetricBlockParams`. I'm inferring this partly based on the initialization of the `SymmetryFactorTester` where `BasicArmiSymmetryTestHelper.blockParamsToTest` is passed in to the `expectedSymmetricBlockParams` kwarg.

armi/testing/symmetryTesting.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def setUp(self):
+        self._preprocessPluginParams()
+        self.symTester = SymmetryFactorTester(
+            self,
+            expectedSymmetricCoreParams=self.coreParamsToTest,
+            expectedSymmetricAssemblyParams=self.assemblyParamsToTest,
+            expectedSymmetricBlockParams=self.blockParamsToTest,
+            parameterOverrides=self.parameterOverrides,
+            paramsToIgnore=self.paramsToIgnore,
```
mgjarrett:
```suggestion
```

If you are already passing in `self`, there is no need to separately pass in all of these instance attributes of `self`.

armi/testing/symmetryTesting.py
```diff
@@ -0,0 +1,324 @@
...
+    """
+
+    def __init__(
+        self,
+        testObject: unittest.TestCase,
+        expectedSymmetricCoreParams: list[str] = [],
+        expectedSymmetricAssemblyParams: list[str] = [],
+        expectedSymmetricBlockParams: list[str] = [],
+        parameterOverrides: dict[str:Any] = {},
+        paramsToIgnore: list[str] = [],
```
mgjarrett:
```suggestion
```

see comment above

armi/testing/symmetryTesting.py
```diff
@@ -0,0 +1,324 @@
...
+        # there is exactly one assembly with 3-symmetry in the test core
+        self.partialAssembly = [a for a in self.r.core.getAssemblies() if a.getSymmetryFactor() == 3][0]
+        self.partialBlock = self.partialAssembly.getBlocks()[0]
+        # expectedSymmetry describes the ratio of (post-expansion / pre-expansion) values
+        self.expectedSymmetry = 3
+        self.defaultParameterValue = 2
+        # load default armi parameters for each object type
+        self._loadDefaultParameters()
+        # some parameters have validation on their inputs and need specific settings
+        self.parameterOverrides = parameterOverrides
```
mgjarrett:
```suggestion
        self.parameterOverrides = testObject.parameterOverrides
```

armi/testing/symmetryTesting.py
```diff
@@ -0,0 +1,324 @@
...
+        self.expectedSymmetricAssemblyParams = expectedSymmetricAssemblyParams
+        self.expectedSymmetricBlockParams = expectedSymmetricBlockParams
+        self._initializeCore()
+        self._initializeAssembly()
+        self._initializeBlock()
+
+        # Some parameters change because of symmetry but are not "volume integrated"
+        # so this marks them for skipping in the compare.
+        # Also allows plugins the flexibility to skip some parameters if needed.
+        self.paramsToIgnore = paramsToIgnore
```
mgjarrett:
```suggestion
        self.paramsToIgnore = testObject.paramsToIgnore
```

armi/testing/symmetryTesting.py
mgjarrett:
```suggestion
        self.expectedSymmetricCoreParams = testObject.coreParamsToTest
        self.expectedSymmetricAssemblyParams = testObject.assemblyParamsToTest
        self.expectedSymmetricBlockParams = testObject.blockParamsToTest
```

armi/testing/symmetryTesting.py
```diff
@@ -0,0 +1,324 @@
...
+                self.testObject.assertEqual(
+                    ratio,
+                    self.expectedSymmetry,
+                    f"The after-to-before expansion ratio of parameter '{paramName}' was expected to be "
+                    f"{self.expectedSymmetry} but was instead {ratio} for the {scopeName}.",
+                )
+
+    @contextmanager
+    def _checkCore(self, expectedParams: Iterable[str]):
+        coreReferenceParameters = self._getParameters(self.core, self.expectedSymmetricCoreParams)
```
mgjarrett:
```suggestion
        coreReferenceParameters = self._getParameters(self.core, self.coreParamsToTest)
```

I won't add all of these suggestions, but this would go with the suggested changes above. The idea here is that we want to pass all `coreParamsToTest`, not just the parameters that are expected to be symmetric.

armi/testing/symmetryTesting.py
```diff
@@ -0,0 +1,324 @@
...
+        ]
+        self.expectedSymmetricAssemblyParams = [
+            p if isinstance(p, str) else p.name for p in self.expectedSymmetricAssemblyParams
+        ]
+        self.expectedSymmetricBlockParams = [
+            p if isinstance(p, str) else p.name for p in self.expectedSymmetricBlockParams
+        ]
+
+    def test_defaultSymmetry(self):
+        self.symTester.runSymmetryFactorTests(expectedBlockParams=self.expectedSymmetricBlockParams)
```
mgjarrett: Why not pass core and assembly parameters by default as well? If the user defines them, they'll be tested. If not, the lists will be empty, so it won't matter, right?

armi/testing/symmetryTesting.py
```diff
@@ -0,0 +1,324 @@
...
+            Dictionary of assembly parameters that the user expects to be symmetry aware.
+        blockParams : Iterable[str], optional
+            Dictionary of block parameters that the user expects to be symmetry aware.
+        """
+        with (
+            self._checkCore(expectedCoreParams),
+            self._checkAssembly(expectedAssemblyParams),
+            self._checkBlock(expectedBlockParams),
+        ):
+            self.r.core.growToFullCore(self.o.cs)
```
mgjarrett: It would be a good addition to capture the `converter` returned here and then run `converter.restorePreviousGeometry()` to verify the symmetry is applied correctly in reverse.

armi/testing/symmetryTesting.py
mgjarrett: I ask because I think I see some core parameters that _should_ be `ParamLocation.VOLUME_INTEGRATED` but are not, so they would fail this test in their current form.

The specific parameters I see are `fissileMass` and `heavyMetalMass`. There could be more (maybe volume-integrated reaction rates).

armi/testing/symmetryTesting.py
bsculac: These are vestigial and I shall remove them!

armi/testing/symmetryTesting.py
bsculac: I shall change the names on the `SymmetryFactorTester`, and change how they are assigned.

armi/testing/symmetryTesting.py
bsculac: I was on the fence but you've convinced me!

armi/testing/symmetryTesting.py
bsculac: Good point, this was leftover when it was expected that plugins would write their own test methods and was not revisited when I decided to make this simpler for plugin implementation. Changing to pass all expected params.

mgjarrett (approved):

.github/workflows/docs.yaml
```diff
@@ -22,6 +22,8 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: 3.13
+      - name: Update package index
+        run: sudo apt-get update
```
opotowsky, line 26: @john-science this update was needed for this to pass CI. I am not sure if we will run into issues with its removal from other workflows, but I just added what was needed to get this PR to pass CI. The build still ran in 17 min.

armi/testing/symmetryTesting.py
```diff
@@ -0,0 +1,315 @@
...
+
+    coreParamsToTest = []
+    assemblyParamsToTest = []
+    blockParamsToTest = []
+    expectedSymmetricCoreParams = []
+    expectedSymmetricAssemblyParams = []
+    expectedSymmetricBlockParams = []
+    parameterOverrides = {}
+    paramsToIgnore = []
+    customSettings = {}
```
drewj-tp, line 117: I know this has been merged already, but I came across this (very useful!) tool is being used internally.

I'm very cautious about mutable class attributes on something we expect subclasses to modify. We will never have a problem if someone only ever sets these attributes. But if someone appends or removes or modifies in place, those changes will persist.

<details>

<summary> Example of foot guns in wait </summary>

```python
>>> class Base:
...     items = []
...     def __repr__(self):
...         return f"{type(self)} {self.items=}"
>>> class SafeChild(Base):
...     def __init__(self):
...         self.items = ["hello"]
...

>>> class UnsafeChild1(Base):
...     def __init__(self):
...         self.items.append("from 1")
...

>>> class UnsafeChild2(Base):
...     def __init__(self):
...         self.items.append("from 2")

>>> UnsafeChild1()
<class '__main__.UnsafeChild1'> self.items=['from 1']

>>> SafeChild()
<class '__main__.SafeChild'> self.items=['hello']

>>> UnsafeChild1()
<class '__main__.UnsafeChild1'> self.items=['from 1', 'from 1']

>>> UnsafeChild2()
<class '__main__.UnsafeChild2'> self.items=['from 1', 'from 1', 'from 2']
```
</details>

A safer pattern would be to put these inside an `__init__` method.

bsculac: The thought was to make these attributes findable with IDE tab complete tools, but if we think this is a problem then it shouldn't be a problem to swap it.

opotowsky: Oh yes let's get it in an init method! Ping me when the PR is up

bsculac: PR open https://github.com/terrapower/armi/pull/2241.

.github/workflows/docs.yaml
john-science: @opotowsky 

Humph. This adds 3-5 minutes to CI.

I am confused... how did you come across this? Because I saw the docs build on the PR that removed this, and then another couple PRs after that.

opotowsky: docs builds began to fail, reliably. I re-added it (failure message said to run `update`) and they began passing.

opotowsky: super bummer it adds 5 min though. I don't know of a way around it :-(

john-science: awwww

Drat. And also shucks.

Okay, well, I'll have to look for other ways to speed our the docs build.

PR #2198, opened 2025-07-24
Updating Axial Expansion Changer for improved mass redistribution

albeanth:
## What is the change? Why is it being made?

These changes support the ongoing the multi-pin assembly effort.

The current axial expansion changer was built assuming there would be a single pin type. When introducing a second pin type (presumably fuel or other neutronically important material) it is no longer generally possible to conserve mass of the second pin component who is the non-axial expansion target component. The lone exception is if the second component is isothermal and contains the same number densities throughout. 

To address this limitation, this PR improves on the mass redistribution that was previously implicitly done by explicitly passing mass between axially linked components and using this new mass and volume to compute new number densities. This new redistribution requires that the axially linked components have the same nuclides; this is used to ensure that they are the same material. If the nuclides were different, they would presumably be different materials and material properties would become largely unknown. 

### High-level description of changes:

1. `assemblyAxialLinkage.py`: bring back upper
2. `axialExpansionChanger`: new redistrubtion algorithm
3. `expansionData: add `getTargetComponent` method
4. `test_axialExpansionChanger.py`: add tests
5. `densityTools.py`: type hints
6. `detailedAxialExpansion` blueprints: simplify simplify simplify

## SCR Information

Change Type: features

One-Sentence Description:  New axial expansion changer that manually redistributes mass. 

One-line Impact on Requirements: Minor change to the implementation `I_ARMI_INP_COLD_HEIGHT` for `R_ARMI_INP_COLD_HEIGHT` and major change to the implementation `I_ARMI_ASSEM_HEIGHT_PRES` for `R_ARMI_ASSEM_HEIGHT_PRES`; adding logic for block-level mass redistribution. 

---

## Checklist

- [x] This PR has only [one purpose or idea](https://terrapower.github.io/armi/developer/tooling.html#one-idea-one-pr).
- [x] [Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify any new/changed code.
- [x] The [documentation](https://terrapower.github.io/armi/developer/tooling.html#document-it) is still up-to-date in the `doc` folder.
- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The dependencies are still up-to-date in `pyproject.toml`.

PR #2228, opened 2025-08-15
Fixing bug in Air.pseudoDensity when given Celsius T

albeanth:

While working on #2198 I found a bug in air.pseudoDensity. If you supply `Tc` to `air.pseudoDensity`, then on line 86 you get `Tk`. Then two lines later when `1/Tk` is calculated, you supply both `Tc` and `Tk` to `getTk` and that errors out. A fix here is to simply just use `Tk` and not try to convert it again. 

## SCR Information

Change Type: fixes

One-Sentence Description: Fixing bug in Air.pseudoDensity when given Celsius temperature.

---

## Checklist

### DIFF
--- a/armi/materials/air.py
+++ b/armi/materials/air.py
@@ -85,7 +85,7 @@ class Air(material.Fluid):
         """
         Tk = getTk(Tc, Tk)
         self.checkPropertyTempRange("pseudoDensity", Tk)
-        inv_Tk = 1.0 / getTk(Tc, Tk)
+        inv_Tk = 1.0 / Tk
         rho_kgPerM3 = 1.15675e03 * inv_Tk**2 + 3.43413e02 * inv_Tk + 2.99731e-03
         return rho_kgPerM3 / G_PER_CM3_TO_KG_PER_M3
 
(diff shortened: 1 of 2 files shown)


armi/materials/air.py
```diff
@@ -85,7 +85,7 @@ def pseudoDensity(
         """
         Tk = getTk(Tc, Tk)
         self.checkPropertyTempRange("pseudoDensity", Tk)
-        inv_Tk = 1.0 / getTk(Tc, Tk)
+        inv_Tk = 1.0 / Tk
```
john-science, line 88: This is obviously _better_, but it is does not change the result, right?

`getTk()` will just return `Tk` here. So there is no difference in the mathematical result of this method? (I ask because this says "fix bug" in the title".

john-science (approved):

armi/materials/air.py
albeanth: > So there is no difference in the mathematical result of this method?

That's my understanding, correct. `getTk` is just a helper method to convert C to K. If you provide `Tc` you do the conversion on line 86 and get Tk. Line 87 does not modify the temperature, so line 88 just uses the thing you already converted. 

Come hunt me down if you ever find otherwise 😀

PR #2229, opened 2025-08-15
Change initialization of modArea for database load

mgjarrett:

Initialize all Component dimensions with `None` instead of 0 when loading from a database. This prevents `modArea`, which is usually `None`, from becoming 0 and being written to the database as a result.

This is being done because the `modArea` Component parameter is typically `None` when loading a model from inputs, but it will be assigned as 0.0 when loading from a database. These 0.0 values would then be written to the database, whereas `None` would be skipped. This change is being made so that restart runs can more closely replicate the original database from which they are restarted.

There is also a strange interaction with h5py where the `modArea` values of 0 are sometimes stored as -(2^63) in the database. This seems to be some kind of confusion between signed and unsigned integers. This PR will avoid this issue.

## SCR Information

Change Type: fixes

One-Sentence Description: change Component dimension initialization value from 0 to None in layout.py.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/layout.py
+++ b/armi/bookkeeping/db/layout.py
@@ -342,7 +342,8 @@ class Layout:
                 comp = Klass(name)
             elif issubclass(Klass, Component):
                 # init all dimensions to None, they will be loaded and assigned after load
-                kwargs = dict.fromkeys(Klass.DIMENSION_NAMES, None)
+                kwargs = dict.fromkeys(Klass.DIMENSION_NAMES, 0)
+                kwargs["modArea"] = None
                 kwargs["material"] = material
                 kwargs["name"] = name
                 kwargs["Tinput"] = temperatures[0]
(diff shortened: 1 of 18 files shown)


mgjarrett:
Here is the weird issue with `modArea` being stored as -(2^63) when it should be identically 0 everywhere.

image

armi/bookkeeping/db/layout.py
```diff
@@ -341,8 +341,9 @@ def _initComps(self, caseTitle, bp):
             elif issubclass(Klass, ExcoreStructure):
                 comp = Klass(name)
             elif issubclass(Klass, Component):
-                # init all dimensions to 0, they will be loaded and assigned after load
+                # init all dimensions to None, they will be loaded and assigned after load
```
mgjarrett:
```suggestion
                # init all dimensions to 0, they will be loaded and assigned after load
```

PR #2230, opened 2025-08-18
Adding Composite.getFirstComponent method

john-science:

Adding Composite.getFirstComponent method, because it would be useful for people that have this a lot in their code:

```python
c = X.getComponents()[0]
```

I just want to add the new method to help performance.

## SCR Information

Change Type: trivial

One-Sentence Description: Adding Composite.getFirstComponent method

---

## Checklist

### DIFF
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -2656,7 +2656,42 @@ class Composite(ArmiObject):
         return list(items)
 
     def getComponents(self, typeSpec: TypeSpec = None, exact=False):
+        """
+        Return a list of Component objects within this Composite.
+
+        Parameters
+        ----------
+        typeSpec : TypeSpec
+            Component flags. Will restrict Components to specific ones matching the flags specified.
+        exact : bool, optional
+            Only match exact component labels (names). If True, 'coolant' will not match 'interCoolant'. This has no
+            impact if typeSpec is None.
+
+        Returns
+        -------
+        list of Component
+            items matching typeSpec and exact criteria
+        """
         return list(self.iterComponents(typeSpec, exact))
+
+    def getFirstComponent(self, typeSpec: TypeSpec = None, exact=False):
+        """
+        Returns a single Component object within this Composite.
+
+        Parameters
+        ----------
+        typeSpec : TypeSpec
+            Component flags. Will restrict Components to specific ones matching the flags specified.
+        exact : bool, optional
+            Only match exact component labels (names). If True, 'coolant' will not match 'interCoolant'. This has no
+            impact if typeSpec is None.
+
+        Returns
+        -------
+        Component
+            The first item matching typeSpec and exact criteria
+        """
+        return next(self.iterComponents(typeSpec, exact))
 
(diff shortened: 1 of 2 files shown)


armi/reactor/composites.py
```diff
@@ -2656,8 +2656,43 @@ def getChildren(
...
+        exact : bool, optional
+            Only match exact component labels (names). If True, 'coolant' will not match 'interCoolant'. This has no
+            impact if typeSpec is None.
+
+        Returns
+        -------
+        Component
+            The first item matching typeSpec and exact criteria
+        """
+        return next(self.iterComponents(typeSpec, exact))
```
drewj-tp: If there is no component that matches the spec, this will raise a `StopIteration` error, which is not super helpful for users.

Please consider catching that with, e.g.,
```python
try:
    return next(self.iterComponents(typeSpec, exact))
except StopIteration:
    raise ValueError(f"No component matches {typeSpec=} {exact=}")
```
or something

john-science: Done

armi/reactor/tests/test_composites.py
```diff
@@ -473,6 +473,28 @@ def test_getReactionRates(self):
             self.assertAlmostEqual(rRatesCore[key], val)
             self.assertAlmostEqual(rRatesReactor[key], val)
 
+    def test_getFirstComponent(self):
+        c = self.container.getComponents()[0]
+        c0 = self.container.getFirstComponent()
+        self.assertEqual(c, c0)
```
drewj-tp: Stronger check that we get the same objects, not identical objects
```suggestion
        self.assertIs(c, c0)
```

armi/reactor/tests/test_composites.py
```diff
@@ -473,6 +473,28 @@ def test_getReactionRates(self):
...
+        self.assertIsInstance(c0, composites.Composite)
+
+        c = self.cladChild.getComponents()[0]
+        c0 = self.cladChild.getFirstComponent()
+        self.assertEqual(c, c0)
```
drewj-tp:
```suggestion
        self.assertIs(c, c0)
```

armi/reactor/tests/test_composites.py
```diff
@@ -473,6 +473,28 @@ def test_getReactionRates(self):
...
+        self.assertIsInstance(c0, composites.Composite)
+
+        c = self.secondGen.getComponents()[0]
+        c0 = self.secondGen.getFirstComponent()
+        self.assertEqual(c, c0)
```
drewj-tp:
```suggestion
        self.assertIs(c, c0)
```

armi/reactor/tests/test_composites.py
```diff
@@ -473,6 +473,28 @@ def test_getReactionRates(self):
...
+        self.assertIsInstance(c0, composites.Composite)
+
+        b = loadTestBlock()
+        c = b.getComponents()[0]
+        c0 = b.getFirstComponent()
+        self.assertEqual(c, c0)
```
drewj-tp:
```suggestion
        self.assertIs(c, c0)
```

PR #2236, opened 2025-08-20
Handle pinIndices for blocks that don't have fuel

drewj-tp:

In a discussion w @mgjarrett and @john-science, we realized you could have a plenum block that has pins but no fuel. Nor any of the "magic" / "important" flags used to set `Component.p.pinIndices`. But it's still a pin so it should have `pinIndices`. If we don't find any components with the "important" flags, we'll assign pin indices to the cladding components.

We could just as easily only assign to cladding components always. The rationale is fuel and control are things where you are more likely to care about the pin power or pin mg flux directly on that component. So, having direct access to `Component.p.pinIndices` on fuel is valuable. Otherwise, everytime you query `fuel.getPinIndices`, it will loop over all siblings and compare the spatial locators. Not the worst, but I speculate asking for `fuel.getPinIndices` will be more common than asking for `clad.getPinIndices`. 

Additionally, having `Component.getPinIndices` is not super helpful. What does that mean? Is it not a pin? Does it occupy all the lattice sites? Now it will raise an exception.

## SCR Information

Change Type: fixes

One-Sentence Description: Assign pinIndices to clad components if there are no fuel nor control components

---

## Checklist

### DIFF
--- a/armi/reactor/components/component.py
+++ b/armi/reactor/components/component.py
@@ -1372,16 +1372,22 @@ class Component(composites.Composite, metaclass=ComponentType):
             runLog.error(ee)
             raise ValueError(msg) from ee
 
-    def getPinIndices(self) -> Optional[np.ndarray[tuple[int], np.uint16]]:
+    def getPinIndices(self) -> np.ndarray[tuple[int], np.uint16]:
         """Find the indices for the locations where this component can be found in the block.
 
         Returns
         -------
-        np.array[int] or None
-            None if this object is not a pin, or if this object is not the central component
-            in the pin. Otherwise, return the indices in various Block-level pin methods,
+        np.array[int]
+            The indices in various Block-level pin methods,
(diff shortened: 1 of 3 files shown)


armi/reactor/components/component.py
```diff
@@ -1395,7 +1401,8 @@ def getPinIndices(self) -> Optional[np.ndarray[tuple[int], np.uint16]]:
         for sibling in withPinIndices:
             if sibling.spatialLocator == self.spatialLocator:
                 return sibling.p.pinIndices
-        return None
+        msg = f"{self} on {self.parent} has no pin indices."
+        raise ValueError(msg)
```
john-science, line 1405: I'm pretty soft on this change. I think returning a numpy array of length zero or None will be just as informative and easy to handle downstream.

But if it's important to you, you can make folks downstream handle the error.

john-science (approved):

PR #2243, opened 2025-08-26
Updated wetted perim calc

clstocking:

This change updates the `getWettedPerimeter` function to work properly for blocks with multiple pin types. This function now uses the `mult` dimension for each component and grabs all components with `Flags.CLAD` and `Flags.WIRE` without the `exact=True` option.

## SCR Information

Change Type: fixes

One-Sentence Description: This PR updates the getWettedPerimeter function for blocks with multiple pin types.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -2254,8 +2254,6 @@ class HexBlock(Block):
         wettedPinComponentFlags = (
             Flags.CLAD,
             Flags.WIRE,
-            Flags.CLAD | Flags.DEPLETABLE,
-            Flags.WIRE | Flags.DEPLETABLE,
         )
 
         # flags pertaining to components where both the interior and exterior are wetted
@@ -2272,8 +2270,8 @@ class HexBlock(Block):
 
         wettedPinComponents = []
         for flag in wettedPinComponentFlags:
-            c = self.getComponent(flag, exact=True)
-            wettedPinComponents.append(c) if c else None
+            comps = self.getComponents(flag)
+            wettedPinComponents += comps if comps else []
 
         wettedHollowCircleComponents = []
         wettedHollowHexComponents = []
@@ -2285,7 +2283,6 @@ class HexBlock(Block):
                 wettedHollowCircleComponents.append(c) if c else None
 
         # calculate wetted perimeters according to their geometries
-
         # hollow hexagon = 6 * ip / sqrt(3)
         wettedHollowHexagonPerimeter = 0.0
         for c in wettedHollowHexagonComponents:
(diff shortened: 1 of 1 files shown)


armi/testing/singleMixedAssembly.py
```diff
@@ -0,0 +1,258 @@
...
+             - 1 1 2 2 1 1
+              1 1 2 2 2 1 1
+               1 1 2 2 1 1
+                1 1 2 1 1
+                 1 1 1 1
+"""
+
+
+def buildMixedPinAssembly():
+    FULL_BP = BLOCK_DEFINITIONS + REGULAR_ASSEMBLY_DEF + GRID_DEFINITION
```
john-science, line 255: Can we put a docstring here that says this is a mixed-pin assembly for a sodium-fast-reactor hex-shaped assembly? Or whatever other information you think would describe the test assembly to someone outside TerraPower.

Thanks!

john-science: @albeanth Apparently, this is your test assembly.  Just thinking aloud.

albeanth: @clstocking if you want me to do this, let me know

clstocking: 569fc32896e362e3a33ea28d08b74daa61a92259

john-science:
It looks like we need to run `ruff` on this PR:

```bash
ruff format .
ruff check . --fix
```

armi/reactor/blocks.py
```diff
@@ -2272,8 +2270,8 @@ def getWettedPerimeter(self):
 
         wettedPinComponents = []
         for flag in wettedPinComponentFlags:
-            c = self.getComponent(flag, exact=True)
-            wettedPinComponents.append(c) if c else None
+            comps = self.getComponents(flag)
```
john-science, line 2273: Why isn't this `exact=True` any more?

clstocking: Because the different pin clads will likely have extra flags to differentiate themselves. If using the `exact=True` option they will not be picked up in this calc.

armi/reactor/blocks.py
```diff
@@ -2272,8 +2270,8 @@ def getWettedPerimeter(self):
...
+            wettedPinComponents += comps if comps else []
```
albeanth:
```suggestion
            wettedPinComponents.extend(comps)
```

armi/reactor/blocks.py
albeanth: We need to include some flexibility here to account for multiple clad/wire comps within a block. If there are two clads on a block where the first has flags (Flags.CLAD, Flags.A) and the second has (Flags.CLAD, Flags.B), we need this method to pick up both. In fact, if we keep exact=True, this method would miss both of these and would be wrong. So this arguably is fixing a deficiency that we always had.

PR #2247, opened 2025-08-26
Logging a table of important reactor parameters

clemekay:

It's been pointed out that a simple table of parameters at every node would be useful to make quick comparisons after code changes. This PR adds a feature to the reporting interface that prints a table of parameters to the stdout file at the end of every node. 

Note -- this is missing testing. I've manually tested the functionality using the script get-parameter-table.py; once I have real testing added, I'll remove that script. 

## SCR Information

Change Type: features

One-Sentence Description: Adds function to print parameter table to stdout file.

---

## Checklist


get-parameter-table.py
```diff
@@ -0,0 +1,49 @@
+# Meant to be run from directory containing database
```
john-science: Every Python file in the project needs the same "license" header comment. Check it out.

get-parameter-table.py
```diff
@@ -0,0 +1,49 @@
...
+
+import sys
+import nala
+nala.configure()
```
john-science:
```suggestion
```

john-science: Not open-source. Please remove.

get-parameter-table.py
```diff
@@ -0,0 +1,49 @@
...
+    repInt.interactEveryNode(cycle, node)
+    tableParameters = repInt.tableParameters.copy()
+    return tableParameters
+
+"""
+Expects something like
+    python ../get-parameter-table.py refTestBase
+to be run from refTestBase/
+"""
+case = sys.argv[1]
```
john-science: All this code can't be in the global scope. This gets run when people do `import armi`, which is not good.

Just encapsulate this logic in a function please.

armi/bookkeeping/report/reportInterface.py
```diff
@@ -94,6 +95,54 @@ def interactEveryNode(self, cycle, node):
...
+            self.r.p.cycle,
+            self.r.p.timeNode,
+            self.r.core.p.coupledIteration,
+            self.r.core.p.keffUnc,
+            self.r.core.p.THdeltaPCore / 1e6,
+            self.r.core.getMaxBlockParam("percentBuPeak"),
+            self.r.core.getMaxBlockParam("puFrac"),
+            self.r.core.getMaxParam("THmassFlowRate"),
+        ]
+        paramNames = [
```
john-science, line 111: You can get the "description" field from a parameter. So you don't have to maintain this list manually.

get-parameter-table.py
john-science: Or you know, This is just one or two functions. Just put it in the `reportInterface.py` file.

armi/bookkeeping/report/reportInterface.py
jakehader: It might be good to implement this type of summary on the application level rather than in ARMI since different applications may care about different or arbitrary parameters.

john-science: Actually, Jake makes a good point. I hadn't thought of that, but yeah totally.

get-parameter-table.py
clemekay: Did not mean to include this file in the PR! Sorry you spent time going through it

armi/bookkeeping/report/reportInterface.py
john-science: Yeah, Jake makes a good point. ARMI is used in several reactor projects that don't have blocks at all. This table would not be generally useful to those projects. 

I say we move this into your reactor Application itself, Kayla.

john-science:
I believe we are closing this PR. And following Jake's idea.

If I am wrong, just reopen this PR.

PR #2249, opened 2025-08-27
Improving docs on entry points creation

### LINKED ISSUE #2192: Confusing detail with setting up EntryPoints in a plugin
When you set up a plugin, there's a hook, `defineEntryPoints`, which defines the entry points associated with a plugin that is being registered with an app. 

For some undocumented reason, you have to import these entrypoint modules **_inside_** the hook, not at the plugin module level. Look at an example in ARMI:
https://github.com/terrapower/armi/blob/19c946fe7f5e512aa2920979ed2338ab7b4b438a/armi/cli/__init__.py#L51-L67

When I placed the import at the top of the file, the entrypoint registered but my `APP_NAME` in the code was `armi`, not the app that was loaded. But hen I placed the import back inside the hook impl, everything magically worked. 

I'm not sure if this is a bug or something else, but I'm not sure if this can be recreated inside armi....maybe it could be recreated with a test app.

john-science:

Improving docs on entry points creation.

close #2192

## SCR Information

 Change Type: docs

One-Sentence Description: Improving docs on entry points creation.

---

## Checklist

### DIFF
--- a/doc/developer/entrypoints.rst
+++ b/doc/developer/entrypoints.rst
@@ -84,5 +84,5 @@ To add entry points, ``ArmiPlugin``s can subclass the ``defineEntryPoints`` meth
 has an extensive :py:class:`~armi.cli.EntryPointsPlugin` that comes with several CLI entry points. 
 It is important to note that if you are building your own ARMI ``Application``, the
 ``EntryPointsPlugin`` must be registered to access these entry points either by registration in the application or 
-subclassing :py:class:`~armi.apps.App`. If you do not want them or if only want
+subclassing :py:class:`~armi.apps.App`. If you do not want them or if you only want
 some of them you can build your own list in a custom ``defineEntryPoints`` method.


doc/developer/entrypoints.rst
```diff
@@ -81,7 +76,12 @@ or (if ``myapp`` is not in your ``PYTHONPATH``)::
...
-    don't change. And such changes tend to hide data from other developers.
+.. tip:: The settings file will be read into a ``Settings`` object. This ``Settings`` object will be passed widely
+    around the code. Please do not edit these settings during a run. The idea of "run settings" is a lot simpler to
+    understand when they don't change. And such changes tend to hide data from other developers.
+
+Every ``ArmiPlugin`` can subclass the ``defineEntryPoints`` method. This is the place to add all your entry points. ARMI
+comes with an extensive :py:class:`~armi.cli.EntryPointsPlugin` that comes with several CLI entry points that you will
+want. It is important to note that if you are building your own ARMI ``Application``, if you do not include the
+``EntryPointsPlugin``, you will not have these entry points. But that is fine if you do not want them or if only want
+some of them you can build your own list in a custom ``defineEntryPoints`` method.
```
opotowsky: The last sentence needs a little grammar love here. But also I have questions!

I have experience with an app that only defines `defineEntryPoints` and those entrypoints subclass ARMI's `EntryPoint`. So it doesn't have the `EntryPoint` plugin registered. That app does have access to ARMI's `run` entrypoint via `<app> run`. So i'm not sure what you're saying here is true (unsure if that's a bug, I'm misunderstanding what you're saying, or something else)

john-science: This is where the default ARMI App gets its entry points:

https://github.com/terrapower/armi/blob/3b9c42babaa6511babc902f6a76b78fa8385df7a/armi/cli/__init__.py#L51-L66

If you subclass the default ARMI `App`, then you get all of those. If you don't subclass this, then you have to create a `defineEntryPoints` yourself.

opotowsky: Ah Ok since we do `from armi.apps import App; MyApp(App)` we get this registered. Cool cool. Ok I'll resolve this then go suggest some rewording in this paragraph

opotowsky:
```suggestion
To add entry points, ``ArmiPlugin``s can subclass the ``defineEntryPoints`` method. ARMI
has an extensive :py:class:`~armi.cli.EntryPointsPlugin` that comes with several CLI entry points. 
It is important to note that if you are building your own ARMI ``Application``, the
``EntryPointsPlugin`` must be registered to access these entry points either by registration in the application or 
subclassing :py:class:`~armi.apps.App`. If you do not want them or if only want
some of them you can build your own list in a custom ``defineEntryPoints`` method.
```

You may want to do this yourself (1) because you may only want a part of these edits and (2) the formatting could be weird , but here's some thoughts on rewording this.

doc/developer/entrypoints.rst
```diff
@@ -81,7 +76,13 @@ or (if ``myapp`` is not in your ``PYTHONPATH``)::
...
+.. tip:: The settings file will be read into a ``Settings`` object. This ``Settings`` object will be passed widely
+    around the code. Please do not edit these settings during a run. The idea of "run settings" is a lot simpler to
+    understand when they don't change. And such changes tend to hide data from other developers.
+
+To add entry points, ``ArmiPlugin``s can subclass the ``defineEntryPoints`` method. ARMI
+has an extensive :py:class:`~armi.cli.EntryPointsPlugin` that comes with several CLI entry points. 
+It is important to note that if you are building your own ARMI ``Application``, the
+``EntryPointsPlugin`` must be registered to access these entry points either by registration in the application or 
+subclassing :py:class:`~armi.apps.App`. If you do not want them or if only want
+some of them you can build your own list in a custom ``defineEntryPoints`` method.
```
opotowsky:
```suggestion
subclassing :py:class:`~armi.apps.App`. If you do not want them or if you only want
some of them you can build your own list in a custom ``defineEntryPoints`` method.
```

Ah dangit I missed the grammar thing that started the whole edit in the first place

PR #2250, opened 2025-08-27
Logging redundant material modifications

### LINKED ISSUE #1820: `material modifications` can be out of sync with the use of `custom isotopics`
A lot of our material classes utilize `material modifications` specified in the blueprints to do things like specify the enrichments of materials in the blocks. For instance:
image

We also now have the ability to specify custom isotopic vectors that are paired with material classes, in which case the custom isotopic will take precedence over the isotopics from the material class (see #1745 ). 

To my knowledge, these two methods of specifying isotopics in the blueprints are able to conflict with each other without any warning being printed to the user. I assume that the material modifications are simply ignored with respect to the component's isotopics.

It would seem useful if the user were somehow informed of this situation, though it seems pretty difficult to make that happen without just blasting a warning anytime that the custom isotopic is specified on a component irrespective of if a material modification is used or not.

john-science:

In the past we made to sure to log the error of someone specifying a `TD_frac` and `custom isotopics` for the same component's material in the blueprints. But now we are extending this same error logging to include a wider range of material modifications: `B10_wt_frac`, `LI_wt_frac`, `U233_wt_frac`, `U235_wt_frac`, and so on. Anything that ends in `_frac`. 

While this is not a guarantee against people naming material modifications crazy things in the future. It's certainly more visibility than we've ever had in the past.

close #1820

## SCR Information

Change Type: docs

One-Sentence Description: Logging redundant material modifications.

One-line Impact on Requirements: Trivial logging change to `I_ARMI_BP_COMP` for `R_ARMI_BP_COMP`.

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/componentBlueprint.py
+++ b/armi/reactor/blueprints/componentBlueprint.py
@@ -248,38 +248,37 @@ class ComponentBlueprint(yamlize.Object):
 
         if densityFromCustomIsotopic <= 0:
             runLog.error(
-                "A zero or negative density was specified in a custom isotopics input. "
-                "This is not permitted, if a 0 density material is needed, use 'Void'. "
-                f"The component is {comp} and the isotopics entry is {self.isotopics}."
+                "A zero or negative density was specified in a custom isotopics input. This is not permitted, if a 0 "
+                f"density material is needed, use 'Void'. The component is {comp} and the isotopics entry is "
+                f"{self.isotopics}."
             )
             raise ValueError("A zero or negative density was specified in the custom isotopics for a component")
 
         mat = materials.resolveMaterialClassByName(self.material)()
         if not isinstance(mat, materials.Custom):
             # check for some problem cases
-            if "TD_frac" in matMods.keys():
+            overSpecs = [k for k in matMods if k.endswith("_frac")]
+            if len(overSpecs):
                 runLog.error(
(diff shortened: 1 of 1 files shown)


armi/reactor/blueprints/componentBlueprint.py
```diff
@@ -248,38 +248,44 @@ def _setComponentCustomDensity(self, comp, blueprint, matMods, inputHeightsConsi
...
-                "This is not permitted, if a 0 density material is needed, use 'Void'. "
-                f"The component is {comp} and the isotopics entry is {self.isotopics}."
+                "A zero or negative density was specified in a custom isotopics input. This is not permitted, if a 0 "
+                f"density material is needed, use 'Void'. The component is {comp} and the isotopics entry is "
+                f"{self.isotopics}."
             )
             raise ValueError("A zero or negative density was specified in the custom isotopics for a component")
+        elif len(matMods):
+            runLog.warning(
+                "Custom isotopics and material modifications have both been defined for {self.material}. Please "
```
keckler:
```suggestion
                f"Custom isotopics and material modifications have both been defined for {self.material}. Please "
```

keckler: Also it might be better to tell the user which component this is, not just the material.

keckler (approved):

armi/reactor/blueprints/componentBlueprint.py
john-science: done

PR #2256, opened 2025-08-29
Providing ways to check if an assembly or block is "detailed"

### LINKED ISSUE #2156: Add a generator/iterator method to retrieve detailed assemblies
The history tracker has two `get*` methods for retrieving detailed assemblies and blocks

- `getDetailAssemblies`
- `getDetailBlocks`

https://github.com/terrapower/armi/blob/41e169a945290f632b9538f4bee019604bd71f14/armi/bookkeeping/historyTracker.py#L246-L266

An iterator approach would be useful for cases where a client (me in this case) may not want a list of all the assemblies / blocks because that could be manifold.

drewj-tp:

Detailed assemblies are those designated to get special treatment in physics codes. Sometimes, you have a specific `Assembly` or `Block` and you want to check if it is eligible for this detailed treatment. Before, you would have to do something like:

```python
allDetailed = hti.getDetailedAssemblies()
for a in r.core:
    if a in allDetailed:
        # do detailed thing
    else:
        # don't
```

This PR introduces `HistoryTrackerInterface.isDetailedAssembly` to check if a given assembly is considered detailed or not. Similarly, the `HistoryTrackerInterface.isDetailedBlock` will check if a block is in a detailed assembly. Mostly by just checking if the parent `isDetailedAssembly` so maybe not a lot of value?

Doing this, you can pretty easily write iterators/generators to find all the detailed assemblies in the `Core` right now. I added one example in ba054c699fafd11afd38e1a4901cbbdad0b54b3d:

```python
for a in filter(hti.isDetailedAssembly, r.core):
    # do detailed thing
```

You could use [`itertools.chain.from_iterable`](https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable) to get the detailed blocks with e.g.,:

```python
asms = filter(hti.isDetailedAssembly, r.core)
blocks = itertools.chain.from_iterable(asms)
```

We could move those into the full API if we want. 

Closes #2156 

At least for the scope I was looking for when I wrote that. I've scoped out internal usage and these methods will be useful and remove the need (well, my need) for an iterator of detailed assemblies.

## SCR Information

Change Type: features

One-Sentence Description: Provide `HistoryTrackerInterface.isDetailedAssembly` and `isDetailedBlock` for checking if an assembly or block is considered "detailed"
One-Sentence Rationale: People would like more tools for finding "detailed assemblies".

One-line Impact on Requirements: N/A

---

## Checklist


john-science (approved):

john-science:
@drewj-tp What's up with this PR? It is like 8 months old, but the CI is passing. Where are you at on this?

drewj-tp:
I do not intend to complete this. We don't have a pressing need for it.

PR #2259, opened 2025-09-03
Re-assigning pin indices when sorting a Block

drewj-tp:

The component pin indices are assigned when being added to an assembly. This is done well in advance of the reactor being completed, e.g., in `reactors.factory()`. One of the last things the factory method does is call `Reactor.sort` which will eventually call `Block.sort` and change the order of the child components in the block. This, in turn, changes the pin locations since now we could find clad components in different order. Which would render `Component.getPinIndices` meaningless because the locations are relative to the previous ordering of components.

Now, `Block.sort` will re-assign pin indices if there is a spatial grid.

Not sure on the change type since it's related to a new feature, but it's a bug fix in said feature.

## SCR Information

Change Type: fixes

One-Sentence Description: Re-assigning pin indices when sorting a Block

---

## Checklist


armi/reactor/tests/test_blocks.py
```diff
@@ -2789,6 +2793,19 @@ def test_nonFueledBlock(self):
...
 
+    def test_reassignOnSort(self):
+        """Show the pin indices are reassigned when the block is sorted."""
+        # Make sure we get new block-level pin locations or else this test is meaningless
+        with patch.object(self.block, "assignPinIndices") as patchAssign:
+            self.block.sort()
+        newPinLocations = self.block.getPinLocations()
+        if newPinLocations == self.allLocations:
+            # raise an error rather than assert to signal as a test error not failure
+            raise RuntimeError("Test requires new pin locations post-sort.")
```
john-science: I think this situation WOULD be a test error.

drewj-tp: which it is?

The counter point would be to do `self.assertNotEqual(newLocations, self.allLocations)` but then we have pytest treat this as a failed test. And I think we do want to capture this as an error in the test, kind of like a failed set up

john-science: Yes, change it to an assertion please. Otherwise, this test doesn't really test that the pinLocations are being set correctly during the sort method.

Thanks!

john-science (approved):

PR #2260, opened 2025-09-03
Fixing `compareLines` so that it doesn't trip on zeros

keckler:

The `compareLines` utility has a little quirk that if the two lines you're trying to compare has a bunch of zeros on it, the comparison will fail because you'll encounter a `DivideByZeroError` here: 
https://github.com/terrapower/armi/blob/aba9095f4d78a98813a2027b0531bfe24954ca44/armi/tests/__init__.py#L296

This PR makes a small change to the method to avoid this scenario.

## SCR Information

Change Type: fixes

One-Sentence Description: Fixing `compareLines()` so that it doesn't trip on zeros

---

## Checklist

### DIFF
--- a/armi/tests/__init__.py
+++ b/armi/tests/__init__.py
@@ -293,7 +293,10 @@ class ArmiTestHelper(unittest.TestCase):
 
             if actualVal is not None:
                 # we have two floats and can compare them
-                if abs(actualVal - expectedVal) / expectedVal > eps:
+                if actualVal == expectedVal:
+                    # will catch the case where they are zero
+                    return True
+                elif abs(actualVal - expectedVal) / expectedVal > eps:
                     return False
             else:
                 # strings, compare directly
--- a/armi/tests/test_tests.py
+++ b/armi/tests/test_tests.py
@@ -21,17 +21,17 @@ from armi import tests
 
 class TestCompareFiles(unittest.TestCase):
     def test_compareFileLine(self):
-        expected = "oh look, a number! 3.14 and some text and another number 1.5"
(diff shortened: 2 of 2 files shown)


armi/tests/__init__.py
```diff
@@ -293,7 +293,10 @@ def compareLines(actual: str, expected: str, eps: Optional[float] = None):
 
             if actualVal is not None:
                 # we have two floats and can compare them
-                if abs(actualVal - expectedVal) / expectedVal > eps:
+                if actualVal == expectedVal:
+                    # will catch the case where they are zero
+                    return True
+                elif abs(actualVal - expectedVal) / expectedVal > eps:
```
drewj-tp, line 299: Would [`math.isclose`](https://docs.python.org/3/library/math.html#math.isclose) be useful here? It's kind of doing this, but doesn't need the additional division and logic branch
```python
math.isclose(actualVal, expectedVal, rel_tol=eps)
```
looks like it should do the trick

keckler: Probably yes, but this method is used pretty widely and I am worried that changing the diff mechanism will cause many tests to fail downsteam.

john-science: Looks like it works to me:

```python
>>> import math
>>> math.isclose(0, 0, rel_tol=1e-9)
True
>>> math.isclose(1.00000012, 1.00000012, rel_tol=1e-9)
True
>>> math.isclose(1.00000012, 1.00050012, rel_tol=1e-9)
False
```

drewj-tp: > Probably yes, but this method is used pretty widely and I am worried that changing the diff mechanism will cause many tests to fail downsteam.

fair

PR #2263, opened 2025-09-03
Adding tests to improve code coverage

john-science:

Adding tests to improve code coverage.

## SCR Information

Change Type: trivial

One-Sentence Description: Adding tests to improve code coverage.

---

## Checklist


armi/bookkeeping/db/database.py
```diff
@@ -429,9 +429,8 @@ def writeInputsToDB(self, cs, csString=None, bpString=None):
         """
         Write inputs into the database based the Settings.
 
-        This is not DRY on purpose. The goal is that any particular Database implementation should
-        be very stable, so we dont want it to be easy to change one Database implementation's
-        behavior when trying to change another's.
+        This is not DRY on purpose. The goal is that any particular Database implementation should be very stable, so we
+        dont want it to be easy to change one Database implementation's behavior when trying to change another's.
```
drewj-tp: Why these changes? I thought we had already formatted the repo.

In a PR titled "Adding tests to improve code coverage" I'm surprised to see changes outside of test files

armi/bookkeeping/db/tests/test_database3.py
```diff
@@ -125,22 +125,28 @@ def test_load(self):
...
+        # show we can delete a specify H5 key.
         del self.db.h5db["c00n00/Reactor/missingParam"]
         _r = self.db.load(0, 0, allowMissing=False)
 
-        # we shouldn't be able to set the fileName if a file is open
+        # show we can delete an entire time now from the DB.
+        del self.db[0, 0, ""]
+        with self.assertRaises(KeyError):
+            self.db.load(0, 0, allowMissing=False)
+
```
drewj-tp, line 148: TIL about `__delitem__`

Cool, I guess? IDK why we have it but now it's tested 😅 

Non-blocking observation: weird that we have `__delitem__` but not `__getitem__`.

armi/bookkeeping/db/tests/test_database3.py
```diff
@@ -185,6 +191,13 @@ def test_history(self):
         self.assertIn((2, 0), hist["chargeTime"].keys())
         self.assertEqual(hist["chargeTime"][(2, 0)], 2)
 
+        # test edge case: ancient DB file
+        v = self.db._versionMinor
+        self.db._versionMinor = 3
+        with self.assertRaises(ValueError):
+            self.db.getHistoriesByLocation([testBlock], params=["serialNum"], timeSteps=[(0, 0), (1, 0)])
+        self.db._versionMinor = v
```
drewj-tp: Could I interest you in [`patch.object`](https://docs.python.org/3/library/unittest.mock.html#patch-object) for the version minor? It'll handle the getting of old value, setting of new value, and restoring of old value

Something like
```python
with patch.object(self.db, "_versionMinor", 3), self.assertRaises(ValueError):
    self.db.getHistoriesByLocation(...)
```

Non-blocking

armi/bookkeeping/db/tests/test_database3.py
```diff
@@ -930,3 +948,16 @@ def test_badData(self):
...
             with Database(self.db.fileName, "r") as db:
                 _r = db.load(0, 0, allowMissing=True)
+
+
+class TestSimplestDatabaseItems(unittest.TestCase):
+    """The tests here are simple, direct tests of Database, that don't need a DatabaseInterface or Reactor."""
+
+    def test_open(self):
+        dbPath = "test_open.h5"
+        db = Database(dbPath, "w")
```
drewj-tp, line 962: Is this going to leave a very small `test_open.h5` laying around?

armi/reactor/tests/test_composites.py
```diff
@@ -336,6 +336,24 @@ def test_hasFlagsNoTypeSpecified(self):
             self.assertTrue(self.container.hasFlags(t))
             self.assertFalse(self.container.hasFlags(t, exact=True))
 
+    def test_calcTotalParam(self):
+        refMinTotal = 21.0
```
drewj-tp: What is this value? `refMinTotal` implies reference minimum total ... what?

Is this a value to make sure we get some meaningful non-zero number?

armi/reactor/tests/test_composites.py
```diff
@@ -630,6 +652,48 @@ def test_ordering(self):
...
+        comp = self.block.getFirstComponent()
+        c = comp.getAncestorWithFlags(Flags.FUEL)
+        self.assertIsNone(c)
+
+        compos = self.block.getChildrenWithFlags(Flags.FUEL)[0]
+        compon = compos.getFirstComponent()
+        c = compon.getAncestorWithFlags(Flags.FUEL)
+        self.assertEqual(c, compon)
+        c = compos.getAncestorWithFlags(Flags.FUEL)
+        self.assertEqual(c, compos)
```
drewj-tp: This block is hard to follow, in part due to `c`, `compos`, and `compon`. I think what we're trying to do is
1. Get a child with the fuel flag -> `compos`
2. Get a grandchild -> `compon`
3. Show `compon` has an ancestor with fuel, but it's `compon`?
4. Show that `compos` has an ancestor with fuel, but it's `compos`?

I can see the check being
1. `compos` has fuel flag
2. `compon` is a child of `compos`
3. `compon` should have an ancestor with the fuel flag
4. that ancestor should be `compos`

Do I have that right? 

If I understand this check right, `Composite.getAncestorWithFlags` could return itself? That feels like it goes against the notion of an ancestor

Suggestions

1. Var names like `child` and `grandchild` to help show the composite tree (non-blocking)
2. Use `self.assertIs` to show we have the same object, not just an equivalent object

armi/reactor/tests/test_composites.py
```diff
@@ -630,6 +652,48 @@ def test_ordering(self):
...
+    def test_changeNDensByFactor(self):
+        c = deepcopy(self.block.getComponents(Flags.FUEL)[0])
+
+        # test inital state
+        dens = c.getNumberDensities()
+        self.assertAlmostEqual(dens["ZR"], 0.03302903991506813, delta=1e-6)
+        self.assertAlmostEqual(dens["U235"], 0.012819005788784095, delta=1e-6)
+        self.assertAlmostEqual(dens["U238"], 0.10125669470078642, delta=1e-6)
+
+        # change N dens
```
drewj-tp: Comment is a restatement of the line below without additional context. Non-blocking request stemming from my personal preference
```suggestion
```

armi/reactor/tests/test_composites.py
```diff
@@ -630,6 +652,48 @@ def test_ordering(self):
...
+        self.assertEqual(c, compos)
+
+    def test_changeNDensByFactor(self):
+        c = deepcopy(self.block.getComponents(Flags.FUEL)[0])
+
+        # test inital state
+        dens = c.getNumberDensities()
+        self.assertAlmostEqual(dens["ZR"], 0.03302903991506813, delta=1e-6)
+        self.assertAlmostEqual(dens["U235"], 0.012819005788784095, delta=1e-6)
+        self.assertAlmostEqual(dens["U238"], 0.10125669470078642, delta=1e-6)
```
drewj-tp: I would welcome these not being hardcoded values. We (maybe me) have some recent scars of subtle tweaks to materials changing very precise hard coded values.

If we stored e.g., `originalZr = dens["ZR"]`, we can use that later in the check by half call

armi/reactor/tests/test_composites.py
```diff
@@ -735,6 +799,36 @@ def test_getFPMass(self):
...
+
+        # test new state
+        self.assertEqual(c.getFPMass(), 0.0)
+        self.assertAlmostEqual(c.getHMMass(), 7.178895593948443, delta=1e-6)
+        self.assertAlmostEqual(c.getMass(), 7.186784402639666, delta=1e-6)
+
+        # test edge case were zero density
+        for nucName in c.getMassFracs().keys():
+            c.setNumberDensity(nucName, 0.0)
+
```
drewj-tp: If we want to clear the composition
```suggestion
        c.setNumberDensities({})
```
should work.

armi/tests/test_mpiActions.py
```diff
@@ -154,9 +163,12 @@ def test_runActionsDistributionAction(self):
         self.assertEqual(len(results), 1)
         self.assertIsNone(results[0])
 
+        o.cs["verbosity"] = "debug"
```
drewj-tp, line 166: Do we need this debug change? We're not querying the logs that something has happened

armi/tests/test_mpiActions.py
```diff
@@ -154,9 +163,12 @@ def test_runActionsDistributionAction(self):
...
+        res = act.invokeHook()
+        self.assertIsNone(res)
```
drewj-tp, line 168: Is this the behavior we want to test? That a function returns `None`? Is there something stronger we can check? Like that the reactor state is distributed? Or something in the logs?

armi/tests/test_mpiActions.py
```diff
@@ -165,6 +177,17 @@ def test_runActionsDistributeStateAction(self):
...
+    @patch("armi.context.MPI_SIZE", 4)
+    @patch("armi.context.MPI_DISTRIBUTABLE", True)
+    def test_runActionsDistStateActionParallel(self):
+        o, r = test_reactors.loadTestReactor(inputFileName="smallestTestReactor/armiRunSmallest.yaml")
+
+        act = DistributeStateAction([self.action])
+        results = runActions(o, r, o.cs, [act])
+        self.assertEqual(len(results), 1)
+        self.assertIsNone(results[0])
+
```
drewj-tp, line 190: What value does this provide over the other nearly identical distribute state action test?

armi/tests/test_mpiActions.py
```diff
@@ -187,6 +210,9 @@ def test_diagnosePickleErrorTestReactor(self):
             self.assertIn("Scanning blocks by name", mock.getStdout())
             self.assertIn("Scanning the ISOTXS library", mock.getStdout())
 
+    def test_invokeAsMaster(self):
+        self.assertEqual(7, MockMpiAction.invokeAsMaster(1, 2, 3))
+
```
drewj-tp: Can you provide a comment / docstring to this test. I know what you're going for (invoke as master calls the invoke method)

But, at least in a lot of production runs, I don't think we care about what the invoke method returns, just that it runs. So maybe we assert that `MockMpiAction.invokeAsMaster` calls `MockMpiAction.invoke` via a mock or patch?

armi/bookkeeping/db/database.py
john-science: Reverting my docstring and f-string changes.

armi/bookkeeping/db/tests/test_database3.py
john-science: We could consider `__getitem__`, but as I recall it's been tried and people immediately started using the DB like it was a dictionary with lots of nasty performance problems and wild feature requests.

That's all vague recollection though.

armi/bookkeeping/db/tests/test_database3.py
john-science: Sure, good point.

armi/bookkeeping/db/tests/test_database3.py
john-science: Nice catch!  Fixed.

armi/reactor/tests/test_composites.py
john-science: Okay, I changed the variable name to "minSerialNumberCount".

The problem is the only numeric parameter in this test example is serial number. So I'm summing that.

BUT.... it's a "minimum" because of some real complexity when these tests are run in parallel.

armi/reactor/tests/test_composites.py
john-science: You have the right of it. I have modified the variable names and add comments to clarify.

armi/reactor/tests/test_composites.py
john-science: Sure, great point. I rewrote the test a bit.

Thanks for that!

armi/reactor/tests/test_composites.py
john-science: Done

armi/tests/test_mpiActions.py
john-science: This is just trying to get code coverage on a block of code that is only run in debug mode.

armi/tests/test_mpiActions.py
john-science: This is just the untested debug code block.  #shrug

armi/tests/test_mpiActions.py
john-science: It's nearly identical, but the `MPI_DISTRIBUTABLE` flag leads to a bunch of untested code.

armi/tests/test_mpiActions.py
john-science: Added a docstring, sure.

armi/reactor/tests/test_composites.py
```diff
@@ -660,6 +682,52 @@ def test_ordering(self):
...
+        otherBlock.spatialLocator = None
+        with self.assertRaises(ValueError):
+            otherBlock < self.block
+
+    def test_getAncestorWithFlags(self):
+        # this test block is not part of an assembly, so it should not have a parent/ancestor
+        parent = self.block.getAncestorWithFlags(Flags.FUEL)
+        self.assertIsNone(parent)
+
+        # pick a component that is not part of a fuel composite, so it should not have a fuel ancestor
```
opotowsky, line 699: I love all these comments, makes it super easy to know what you're testing when I come in without context!

armi/tests/test_mpiActions.py
opotowsky: noted!

drewj-tp: Following up on this.

I don't think this is a good strategy. Sure, we've increased the code coverage. But these aren't meaningful unit tests. If the underlying behavior changes, we won't catch that here.

armi/tests/test_mpiActions.py
drewj-tp: I disagree with this strategy.

armi/tests/test_mpiActions.py
opotowsky: I don't think the intent here was a unit test revamp or audit or anything. John just saw uncovered lines of code (mostly error catching) and added tests to cover those lines.

PR #2268, opened 2025-09-04
Fixing Uranium enrichment calculations

mgjarrett:

Change uranium enrichment calculations from considering only U-235 and U-238 to considering all uranium isotopes.
Uranium materials can have minor fractions of other non-fissile isotopes (U-232, U-234, U-236) and other fissile isotopes (U-233).

The methods `Block.getUraniumMassEnrich()` and `ArmiObject.getUraniumNumEnrich()` now account for all uranium nuclides present.

## SCR Information

Change Type: fixes

One-Sentence Description: Consider all uranium isotopes in the enrichment calculation.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -31,7 +31,7 @@ import numpy as np
 
 from armi import nuclideBases, runLog
 from armi.bookkeeping import report
-from armi.nucDirectory import elements, nucDir
+from armi.nucDirectory import nucDir
 from armi.nuclearDataIO import xsCollections
 from armi.physics.neutronics import GAMMA, NEUTRON
 from armi.reactor import (
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -1583,15 +1583,18 @@ class Block(composites.Composite):
         return b10 / total
 
     def getUraniumMassEnrich(self):
-        """Returns U-235 mass fraction."""
-        u5 = self.getMass("U235")
+        """Returns fissile mass fraction of Uranium."""
         U = elements.bySymbol["U"]
-        blockNucs = self.getNuclides()
-        uraniumNucsPresent = [nuc.name for nuc in U.nuclides if nuc.name in blockNucs]
-        if u5 < 1e-10:
+        compNucs = set(self.getNuclides())
+        uraniumNucs = set(nuc.name for nuc in U.nuclides)
+        uraniumNucsPresent = compNucs & uraniumNucs
+        fissileNucsPresent = [nucName for nucName in uraniumNucsPresent if nuclideBases.byName[nucName].isFissile()]
+        fissileU = self.getMass(fissileNucsPresent)
+        totalU = self.getMass(list(uraniumNucsPresent))
+        if totalU < 1e-10:
             return 0.0
-        totalU = self.getMass(uraniumNucsPresent)
-        return u5 / totalU
+
+        return fissileU / totalU
 
     def getInputHeight(self) -> float:
         """Determine the input height from blueprints.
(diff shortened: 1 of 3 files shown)


armi/reactor/blocks.py
```diff
@@ -1582,12 +1582,18 @@ def getBoronMassEnrich(self):
...
 
     def getUraniumMassEnrich(self):
-        """Returns U-235 mass fraction assuming U-235 and U-238 only."""
-        u5 = self.getMass("U235")
-        if u5 < 1e-10:
+        """Returns fissile mass fraction of uranium."""
+        U = elements.bySymbol["U"]
+        compNucs = set(self.getNuclides())
+        uraniumNucs = set(nuc.name for nuc in U.nuclides)
+        uraniumNucsPresent = compNucs & uraniumNucs
```
john-science: I see what you're doing here, and it seems correct.

But couldn't we also just do:

```python
self.getMass("U235") / self.getMass("U")
```

Aside from some error handling for zeros, obviously. But does `block.getMass("U")` do something subtlety different that I'm not thinking of?

mgjarrett: `self.getMass("U")` works for the denominator.

For the numerator, we want to add both fissile nuclides, `U233` and `U235`. There isn't a convenient method that does `getFissileNuclides()`, but since this method is specific to uranium we could just hard code it.

armi/reactor/tests/test_blocks.py
```diff
@@ -1277,6 +1282,22 @@ def test_getUraniumNumEnrich(self):
...
         self.assertEqual(cur, 0.0)
 
+    def test_getUraniumNumEnrichWith233(self):
+        fuel = self.block.getComponent(Flags.FUEL)
+        u5 = fuel.getNumberDensity("U235")
+        fuel.setNumberDensity("U233", 0.005)
+        self.block.adjustUEnrich(0.25)
+
+        cur = self.block.getUraniumNumEnrich()
+        print(cur)
```
john-science:
```suggestion
```

john-science: debug print out left over

armi/reactor/composites.py
```diff
@@ -1564,12 +1564,18 @@ def getFissileMassEnrich(self):
...
 
     def getUraniumNumEnrich(self):
-        """Returns U-235 number fraction."""
-        u8 = self.getNumberDensity("U238")
-        if u8 < 1e-10:
+        """Returns fissile uranium number fraction."""
+        U = elements.bySymbol["U"]
+        compNucs = set(self.getNuclides())
+        uraniumNucs = set(nuc.name for nuc in U.nuclides)
+        uraniumNucsPresent = compNucs & uraniumNucs
```
john-science: Same simplification as the number enrichment?

mgjarrett: Looking at it closer, `c.getMass("U")` or `c.getNuclideNumberDensities("U")` attempts to get a number density for all possible uranium nuclides, of which there are 33:

image

If we pare down the list to only the nuclides present before calling `getMass()` or `getNuclideNumberDensities()`, it will usually only be 4-5 nuclides (232, 234, 235, 236, and 238).

I'm not sure whether it's faster to call `getMass("U")` and let it attempt to get number densities for all of those isotopes that aren't present (`U215`, `U216`, `U217`, etc.) or to spend the time figuring out which nuclides are actually present.

john-science: Well, if they were approximately the same speed, I would definitely vote the shorter more readable version.

mgjarrett: The timing is probably inconsequential here, so I'll go with the simpler version.

john-science (approved):

PR #2269, opened 2025-09-05
Track assemblies if discharged to SFP

HunterPSmith:

Currently, assemblies are only tracked after discharge if `trackAssems == True`. However, this forces you to track every discharged assembly which can be costly with respect to DB size. 

This PR changes the default discharge location (if the user does not specify a discharge location in their cascades) to be deleted after discharge. If the user specifies `SFP` at the end of a cascade, the assembly will now be stored in the database in the SFP, even if `trackAssems == False`. This allows us to selectively track assemblies. 

The situation will now be:

* If `trackAssems == True`: store all discharged assemblies
* If `trackAssems == False`:
  * If the user specifies `SFP` at the end of a cascade, some Assemblies will be placed in the SFP, and thus in the DB
  * If the user does not specify `SFP` at the end of a cascade, no Assemblies will be tracked.

## SCR Information

Change Type: features

One-Sentence Description: Enhance fuel handling logic to support flexible assembly discharge destinations and update documentation accordingly. 

---

## Checklist

### DIFF
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -67,7 +67,7 @@ class AssemblyMove:
     assemType : str, optional
         Type of assembly that is moving.
     nameAtDischarge : str, optional
-        Name of the assembly moving (for SFP/ExCore interactions).
+        Name of the assembly moving (for SFP/delete interactions).
     rotation : float, optional
         Degrees of manual rotation to apply after shuffling.
     """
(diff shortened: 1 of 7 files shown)


armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1427,15 +1449,26 @@ def processMoveList(self, moveList):
...
+        return (
+            loadChains,
+            loopChains,
+            enriches,
+            loadChargeTypes,
+            loadNames,
+            dischargeDests,
+            rotations,
+            alreadyDone,
+        )
```
drewj-tp: This should probably be a class with attributes. Then, we can change the attributes or the ordering with a little more freedom to people who are receiving the output of this

HunterPSmith: [7884c27](https://github.com/terrapower/armi/pull/2269/commits/7884c272ae836a1ef16fb5ad6a1ba6a5d27fb83e)

PR #2270, opened 2025-09-05
Fixing sphinx warnings in the doc build

john-science:

Fixing nearly all the Sphinx warnings in the docs build. There are still 3 left that I can't fix, that I think come from one of out plugins. So MAYBE we could fix that by updating the version of one or several of our sphinx plugins. But that is a much bigger tasks, so I'll fix what I can now.

> You can check the doc build on this PR for verification that I didn't break the docs build. Or if you feel like hand verifying the changes I made.

## SCR Information

Change Type: docs

One-Sentence Description: Fixing sphinx warnings in the doc build

---

## Checklist

### DIFF
--- a/doc/tutorials/data_model.nblink
+++ b/doc/tutorials/data_model.nblink
@@ -1,3 +1,3 @@
 {
-	"path": "../../armi/tests/tutorials/data_model.ipynb",
+	"path": "../../armi/tests/tutorials/data_model.ipynb"
 }


armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
```diff
@@ -494,7 +494,7 @@ def manageCoreMesh(self, r):
     def _checkComponentHeight(self, c):
         if c.zbottom > c.ztop:
             msg = f"""
-            {c} has a negative height! This is unphysical.
```
opotowsky, line 497: sphinx won't let us have exclamation points?!?!?! I didn't realize we were under a sphinx dictatorship

armi/reactor/cores.py
```diff
@@ -2172,8 +2172,7 @@ def processLoading(self, cs, dbLoad: bool = False):
 
     def buildManualZones(self, cs):
         """
-        Build the Zones that are defined in the given Settings, in the
-        `zoneDefinitions` or `zonesFile` case setting.
+        Build the Zones that are defined in the given Settings, in the `zoneDefinitions` or `zonesFile` case setting.
```
opotowsky, line 2175: do you know if ` or `` is needed here or if it doesn't matter?

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
john-science: Yeah, this throws an error I couldn't get rid of. Boooo

There problem is SOME way to solve this. But I gave it ~10 minutes and gave up. Stupid Sphinx.

armi/reactor/cores.py
john-science: Tony has strong feelings on this. But they both render in the HTML, so I don't know what the difference is.

Issue #2258, opened 2025-09-03
Provide Component.p.enrichmentBOL


drewj-tp:
It's decently common for reactors to have assemblies with differently enriched pins residing in the same assembly. Our C5G7 docs allude to this with three different MOX pins in the same block [^mox]

https://github.com/terrapower/armi/blob/aba9095f4d78a98813a2027b0531bfe24954ca44/armi/tests/c5g7/c5g7-blueprints.yaml#L5-L40

However, ARMI only provides a `Block.p.enrichmentBOL` parameter. 

https://github.com/terrapower/armi/blob/aba9095f4d78a98813a2027b0531bfe24954ca44/armi/reactor/blocks.py#L740-L751

It would be extremely beneficial to have this on individual components as well

[^mox]: aside: "enrichment" for [MOX fuel](https://en.wikipedia.org/wiki/MOX_fuel) is kind of a misnomer because the interesting fraction is weight fraction PuO2 in UO2 but that's beside the point

drewj-tp:
I agree that this is not an ideal way to track BOL data. But this is the paradigm we have right now.

Alternatives could be using the history tracker to track these BOL data. Interfaces needing e.g., `Component.p.enrichmentBOL` would need a way to fetch that from the history tracker.

Another alternative could be an alternative `h5py.Group` in the root of the database that stores these data. Would need some thinking to make it I/O efficient and not mean we store the entire BOL reactor and need to re-load the BOL reactor at a later state.

Additionally, BOL is kind of a red herring. For assemblies that are shuffled into the core in a later cycle, their BOL is not `(cycle, node)=  (0, 0)` so we would not be able to do e.g., `db.load(0, 0)` in this case. We'd need to find their BOL based on when they were shuffled into the core

PR #2272, opened 2025-09-08
Adding new Component param enrichmentBOL

HunterPSmith:

Addresses: https://github.com/terrapower/armi/issues/2258

ARMI only provides the `enrichmentBOL` parameter at the block level. Since it is desirable to support assemblies with fuel pins of varying enrichment, in `completeInitialLoading`, where the block-level enrichment is set via `self.p.enrichmentBOL`, the loop now assigns an `enrichmentBOL` value to all components of that block, regardless of pin count. Expanded block bookkeeping tests to assert component-level enrichment values and ensure block values are propagated correctly.

## SCR Information

Change Type: features

One-Sentence Description: Add component level `enrichmentBOL` parameter.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -748,11 +748,11 @@ class Block(composites.Composite):
             if isinstance(child, components.Component):
                 child.p.massHmBOL = hmMass
                 child.p.molesHmBOL = child.getHMMoles()
-                child.p.enrichmentBOL = child.getFissileMassEnrich()
                 if child.p.molesHmBOL:
                     child.p.hmNuclidesBOL = [nuc for nuc in child.p.nuclides if nucDir.isHeavyMetal(nuc.decode())]
                     child.p.hmNumberDensitiesBOL = child.getNuclideNumberDensities(child.p.hmNuclidesBOL)
                     child.p.temperatureInCBOL = child.temperatureInC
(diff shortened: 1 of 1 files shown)


armi/reactor/blocks.py
```diff
@@ -744,10 +744,11 @@ def completeInitialLoading(self, bolBlock=None):
             hmMass = child.getHMMass()
             massHmBOL += hmMass
             # Components have the following parameters but not every composite will massHmBOL,
-            # molesHmBOL, puFrac
+            # molesHmBOL, puFrac, enrichmentBOL
             if isinstance(child, components.Component):
                 child.p.massHmBOL = hmMass
                 child.p.molesHmBOL = child.getHMMoles()
+                child.p.enrichmentBOL = child.getFissileMassEnrich()
```
drewj-tp: Super trivial change, but if we move this into the logic branch below, we won't set the param nor calculate the enrichment for things that don't have heavy metal.

Defaults will still be zero for non-fuel. But we could save some DB size / MPI communication size if we're not passing around zeros that have been explicitly set as zeros

HunterPSmith: [e9a8efd](https://github.com/terrapower/armi/pull/2272/commits/e9a8efd03bbb987ec9620ce0a70edd2e1c9a09cf)

PR #2273, opened 2025-09-10
Change `setVerbosity` location in `mockRunLog`

opotowsky:

This moves the `setVerbosity` call from the `init` method to the `log` method.

But WHY?

I was making a unit test downstream and was using `loadTestReactor` and grabbing that log. And only the headers were printing and it was driving me insane! I printed out the current log verbosity and it was 40 ("error"). I tried a few things, eventually moving where the verbosity was set. 0 --> 10 when you do it right. Then I got my logs. The end. 

Running downstream testing now, since this is a cornerstone of many a unit test.

## SCR Information

Change Type: fixes

One-Sentence Description: Change `setVerbosity` location in `mockRunLog` to `log` method rather than `init`.

---

## Checklist

### DIFF
--- a/armi/tests/mockRunLogs.py
+++ b/armi/tests/mockRunLogs.py
@@ -37,7 +37,6 @@ class BufferLog(runLog._RunLog):
         self._errStream = io.StringIO()
         self._deduplication = runLog.DeduplicationFilter()
         sys.stderr = self._errStream
-        self.setVerbosity(0)
 
     def __enter__(self):
         self.originalLog = runLog.LOG
@@ -54,6 +53,7 @@ class BufferLog(runLog._RunLog):
         This is a wrapper around logger.log() that does most of the work and is
         used by all message passers (e.g. info, warning, etc.).
         """
+        self.setVerbosity(0)
         # the message label is only used to determine unique for single-print warnings
         if label is None:
             label = msg


opotowsky:
There's a couple of broken unit tests here
image

If this is the scene downstream I will close this PR and just set the verbosity on my particular test of interest. But I'd imagine this is a larger issue if the log level is only being set to error much of the time.

opotowsky:
> This is a very minor change in ARMI. My major concern is that this doesn't break a lot of things downstream. People's tests or workflows.

I ran the big downstream testing job and it was good!
 
> Though... I think having verbosity turned WAY down during unit testing is good too. It should make the tests run faster, right? Can we test that?

I just left it at 0, which somewhere becomes debug. I don't think the tests NEED it to be debug but I was trying to change as little as possible.

armi/tests/mockRunLogs.py
```diff
@@ -37,7 +37,6 @@ def __init__(self, *args, **kwargs):
         self._errStream = io.StringIO()
         self._deduplication = runLog.DeduplicationFilter()
         sys.stderr = self._errStream
-        self.setVerbosity(0)
```
john-science, line 40: Yeah, I would have expected this to be `self.setVerbsoity(logging.INFO)`. That's a little odd.

PR #2277, opened 2025-09-11
Enhance fuel handler logic to support module imports

HunterPSmith:

Changes are made to ``fuelHandlerFactory`` in order to allow for users to support module imports for custom fuel handlers. ``CONF_SHUFFLE_LOGIC`` can now be set to a dotted module path to be imported instead of only being able to process a file path. ``CONF_FUEL_HANDLER_NAME`` still behaves the same. Tests have been added to ensure that both options work. User documentation is updated.

## SCR Information

Change Type: features

One-Sentence Description: Enhance fuel handler logic to support module imports

---

## Checklist

### DIFF
--- a/armi/physics/fuelCycle/tests/test_fuelHandlerFactory.py
+++ b/armi/physics/fuelCycle/tests/test_fuelHandlerFactory.py
@@ -19,6 +19,7 @@ from pathlib import Path
 
 from armi.physics.fuelCycle import fuelHandlerFactory
 from armi.physics.fuelCycle.settings import CONF_FUEL_HANDLER_NAME, CONF_SHUFFLE_LOGIC
+from armi.physics.fuelCycle.tests import _customFuelHandlerModule
 
 
 class _DummySettings(dict):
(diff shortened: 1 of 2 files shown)


armi/physics/fuelCycle/tests/test_fuelHandlerFactory.py
```diff
@@ -0,0 +1,76 @@
...
+        self.settings.update(
+            {
+                CONF_FUEL_HANDLER_NAME: "FileFuelHandler",
+                CONF_SHUFFLE_LOGIC: str(modulePath),
+            }
+        )
+
+        handler = fuelHandlerFactory.fuelHandlerFactory(self.operator)
+
+        self.assertEqual(handler.__class__.__name__, "FileFuelHandler")
```
mgjarrett: This effectively works just the same, but is there any particular reason not to use `assertIsInstance` here like is done below for the `ModuleFuelHandler` case?

mgjarrett (approved):
Looks good, no significant comments.

armi/physics/fuelCycle/tests/test_fuelHandlerFactory.py
HunterPSmith: since the `fileFuelHandler` loads through `pathTools.importCustomPyModule()` instead of `importlib.import_module()` it produces a distinct module (and `FileFuelHandler` class) that makes `isinstance` fail, hencew hy the test checks only the class name there, while the “module path” test uses `importlib.import_module()` and can safely assert `isinstance`.

PR #2287, opened 2025-09-19
Updating Block BOL HM params during axial expansion

drewj-tp:

The axial expansion changer did not update block level "BOL" parameters to reflect the changed component level BOL parameters. As we do axial expansion, we perform bookkeeping to keep e.g., `Component.p.molesHmBOL` consistent by moving around BOL composition between adjacent components. However, these updates were not applied to the block level. This means that you could have a block w/ fuel pins, and `Block.p.molesHmBOL` would not equal the sum of `c.p.molesHmBOL`.

This PR adds a test to show this behavior and a corresponding fix.

## SCR Information

Change Type: fixes

One-Sentence Description: Update block level BOL heavy metal parameters in addition to component BOL heavy metal parameters through axial expansion

One-line Impact on Requirements: Added logic to handle heavy metal BOL parameters to `I_ARMI_ASSEM_HEIGHT_PRES` for `R_ARMI_ASSEM_HEIGHT_PRES`.

---

## Checklist

### DIFF
--- a/armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
+++ b/armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
@@ -452,16 +452,20 @@ class AxialExpansionChanger:
             mesh.append(b.p.ztop)
             b.spatialLocator = self.linked.a.spatialGrid[0, 0, ib]
 
-            # Update block-level BOL params to account for changes to the child params
-            # Some params are zero, some are None. filter will discard them from the summation
-            for param in ("molesHmBOL", "massHmBOL"):
-                values = (getattr(c.p, param, 0) for c in b)
-                updated = sum(filter(None, values))
-                setattr(b.p, param, updated)
+            self._updateBlockParamsPostExpansion(b)
 
         bounds = list(self.linked.a.spatialGrid._bounds)
         bounds[2] = array(mesh)
         self.linked.a.spatialGrid._bounds = tuple(bounds)
+
+    def _updateBlockParamsPostExpansion(self, b: "Block"):
+        """Update any parameters on the block after the child components have been expanded."""
+        for param in ("molesHmBOL", "massHmBOL"):
+            values = (getattr(c.p, param, 0) for c in b)
+            # Update block-level BOL params to account for changes to the child params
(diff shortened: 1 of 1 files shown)


armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
```diff
@@ -18,7 +18,7 @@
 from math import isclose
 from textwrap import dedent
 
-from numpy import array, sum
```
drewj-tp, line 21: This was causing super weird behavior. Python `builtins.sum` can handle a filter but `numpy.sum` does not
```
>>> import numpy as np
>>> sum(filter(None, [0, 1, 2]))
3
>>> np.sum(filter(None, [0, 1, 2]))
<filter at 0x1d77f169780>
>>> list(_)
[1, 2]
```

john-science: `filter()` returns a "filter object", not an "iterable".

drewj-tp: sure but that's like saying `list()` returns a list object not an iterable
```python
>>> from collections.abc import Iterable, Iterator
>>> isinstance(filter(None, [0, 1]), Iterable)
True
>>> isinstance(list(), Iterable)
True
```
numpy, I think for good reason, doesn't like taking in iterators. Which are not the same as iterables.
```python
>>> from collections.abc import Iterable, Iterator
>>> isinstance(filter(None, [0, 1]), Iterator)
True
>>> isinstance(list(), Iterator)
False
```

albeanth:
it's looking like #2291 is going to capture this too

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
john-science: Hmm.. filter is a subclass of iterable?  Hmm... Okay, so I guess it must be that numpy doesn't support all iterables?  I mean, that seems wrong, but must be true. Odd.

john-science (approved):

john-science:
Holler, if you need a weekend merge. You know how to find me.

PR #2292, opened 2025-09-20
Support mixed Blocks for smear density calculation

mgjarrett:

The smear density calculation in `Block.getSmearDensity()` made several assumptions that are not true for mixed assemblies.

1. The inner cladding area was calculated by averaging the `id` of all `clad`components without taking into account the `mult` of each component. The calculation was updated to correctly calculate area inside cladding.
2. The smear density was calculated from all components inside a randomly `pop`'ed component from the list of `clad`; this might not be the largest clad component, in which case some components could be missed. The `clad` components are now sorted before running `pop` to get the largest one.
3. When there are multiple `clad` components, some can be inside of the others and thus show up in the calculation of component areas inside the cladding. Any `clad` component needs to be excluded from this calculation, otherwise the smear density will be overestimated.

## SCR Information

Change Type: features

One-Sentence Description: Fix the `Block.getSmearDensity()` calculation for assemblies with multiple different types of `clad` and/or `fuel` components.

---

## Checklist

### DIFF
--- a/armi/reactor/tests/test_blocks.py
+++ b/armi/reactor/tests/test_blocks.py
@@ -427,7 +427,8 @@ class Block_TestCase(unittest.TestCase):
         ref = (self.block.getDim(Flags.FUEL, "od") ** 2 - self.block.getDim(Flags.FUEL, "id") ** 2) / self.block.getDim(
             Flags.LINER, "id"
         ) ** 2
-        self.assertAlmostEqual(cur, ref, places=10)
+        places = 10
+        self.assertAlmostEqual(cur, ref, places=places)
 
         # test with liner instead of clad
         ref = (self.block.getDim(Flags.FUEL, "od") ** 2 - self.block.getDim(Flags.FUEL, "id") ** 2) / self.block.getDim(
--- a/armi/reactor/tests/test_blocks.py
+++ b/armi/reactor/tests/test_blocks.py
@@ -463,6 +463,45 @@ class Block_TestCase(unittest.TestCase):
             msg="Incorrect getSmearDensity with annular fuel. Got {0}. Should be {1}".format(cur, ref),
         )
 
+    def test_getSmearDensityMultipleClads(self):
+        # add clad of different size
+        clad = self.block.getComponent(Flags.CLAD)
+        self.block.remove(clad)
+        cladDims = {
+            "Tinput": 273.0,
+            "Thot": 273.0,
+            "od": clad.getDimension("od") + 0.02,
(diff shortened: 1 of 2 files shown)


john-science:
I see the need. Sure.

I'd like a test, or one test for each of your three bullets in the PR description.

If you need help with tests, I can probably slap something together.

john-science:
I'm starting to wish I had been involved in the early discussions for mixed assemblies.

At this point, it seems like, making a subclass for `class MixedAssembly(Assembly)` of maybe `class MixedHexBlock(HexBlock)` might have been a cleaner solution that all these changes trickling in.

armi/reactor/tests/test_blocks.py
```diff
@@ -463,6 +463,115 @@ def test_getSmearDensity(self):
...
+        )
+        innerArea = 0.0
+        for clad in clads:
+            innerArea += math.pi / 4.0 * clad.getDimension("id", cold=True) ** 2 * clad.getDimension("mult")
+        for liner in self.block.getComponents(Flags.LINER):
+            innerArea -= liner.getArea(cold=True)
+
+        ref = fuelArea / innerArea
+        places = 10
+        self.assertAlmostEqual(cur, ref, places=places)
```
john-science:
```suggestion
        # calculate reference smear density
        fuel = self.block.getComponent(Flags.FUEL, exact=True)
        annularFuel = self.block.getComponent(Flags.FUEL | Flags.ANNULAR)
        liner = self.block.getComponent(Flags.LINER | Flags.INNER)
        clads = self.block.getComponents(Flags.CLAD)
        fuelArea = math.pi / 4.0 * fuel.getDimension("od", cold=True) ** 2 * fuel.getDimension("mult")
        fuelArea += (
            math.pi
            / 4.0
            * (annularFuel.getDimension("od", cold=True) ** 2 - annularFuel.getDimension("id", cold=True) ** 2)
            * annularFuel.getDimension("mult")
        )
        innerArea = 0.0
        for clad in clads:
            innerArea += math.pi / 4.0 * clad.getDimension("id", cold=True) ** 2 * clad.getDimension("mult")
        for liner in self.block.getComponents(Flags.LINER):
            innerArea -= liner.getArea(cold=True)

        ref = fuelArea / innerArea
        cur = self.block.getSmearDensity()
        self.assertAlmostEqual(cur, ref, places=10)
```

john-science: Minor, opinionated cleanup. 

I won't die on this hill.

john-science (approved):

PR #2294, opened 2025-09-21
Fix DB load then write

alexhjames:

Currently, a reactor loaded from a database cannot be written to a new database file that is operable. The new database file will be missing the blueprints causing errors when loading that database. This change updates the blueprints location when loading from a database to point at the database file instead of the original blueprints file (which may no longer be accessible). On writing to a database, the blueprints are either read from the blueprints file (what it used to do) or read from the database (if the blueprints file is pointing at an h5).  

## SCR Information

Change Type: fixes

One-Sentence Description: Allow reactors loaded from a database to be written to a working database file.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database.py
+++ b/armi/bookkeeping/db/database.py
@@ -172,11 +172,15 @@ class Database:
         return self._versionMinor
 
     def __repr__(self):
-        return "<{} {}>".format(self.__class__.__name__, repr(self.h5db).replace("<", "").replace(">", ""))
+        return "<{} {}>".format(
+            self.__class__.__name__, repr(self.h5db).replace("<", "").replace(">", "")
+        )
 
     def open(self):
         if self.h5db is not None:
-            raise ValueError("This database is already open; make sure to close it before trying to open it again.")
+            raise ValueError(
+                "This database is already open; make sure to close it before trying to open it again."
+            )
         filePath = self._fileName
         self._openCount += 1
 
@@ -193,7 +197,9 @@ class Database:
 
         else:
             runLog.error("Unrecognized file permissions `{}`".format(self._permission))
-            raise ValueError("Cannot open database with permission `{}`".format(self._permission))
+            raise ValueError(
+                "Cannot open database with permission `{}`".format(self._permission)
+            )
 
         # open the database, and write a bunch of metadata to it
         runLog.info("Opening database file at {}".format(os.path.abspath(filePath)))
@@ -207,7 +213,10 @@ class Database:
         app = getApp()
         self.h5db.attrs["appName"] = app.name
         plugins = app.pluginManager.list_name_plugin()
-        ps = [(os.path.abspath(sys.modules[p[1].__module__].__file__), p[1].__name__) for p in plugins]
+        ps = [
+            (os.path.abspath(sys.modules[p[1].__module__].__file__), p[1].__name__)
+            for p in plugins
+        ]
         ps = np.array([str(p[0]) + ":" + str(p[1]) for p in ps]).astype("S")
         self.h5db.attrs["pluginPaths"] = ps
         self.h5db.attrs["localCommitHash"] = Database.grabLocalCommitHash()
@@ -302,7 +311,9 @@ class Database:
             newPath = safeMove(self._fullPath, self._fileName)
             self._fullPath = os.path.abspath(newPath)
 
-    def splitDatabase(self, keepTimeSteps: Sequence[Tuple[int, int]], label: str) -> str:
+    def splitDatabase(
+        self, keepTimeSteps: Sequence[Tuple[int, int]], label: str
+    ) -> str:
         """
         Discard all data except for specific time steps, retaining old data in a separate file.
 
@@ -353,7 +364,9 @@ class Database:
 
             if not set(keepTimeSteps).issubset(timeSteps):
                 raise ValueError(
-                    "Not all desired time steps ({}) are even present in the database".format(keepTimeSteps)
+                    "Not all desired time steps ({}) are even present in the database".format(
+                        keepTimeSteps
+                    )
                 )
 
             minCycle = next(iter(sorted(keepTimeSteps)))[0]
@@ -390,8 +403,14 @@ class Database:
         custom Application you do not have access to, the DB may not be usable.
         """
         cs = settings.Settings()
-        cs.caseTitle = os.path.splitext(os.path.basename(self.fileName))[0]
-        cs.loadFromString(self.h5db["inputs/settings"].asstr()[()], handleInvalids=handleInvalids)
+        cs.path = self.fileName
+        cs.loadFromString(
+            self.h5db["inputs/settings"].asstr()[()], handleInvalids=handleInvalids
+        )
+
+        # Need to update the blueprint file to be the database so that its not pointing at a source
+        # that doesn't exist anymore (the original blueprints yaml).
+        cs[CONF_LOADING_FILE] = os.path.basename(self.fileName)
 
         return cs
 
@@ -457,7 +476,9 @@ class Database:
         file contains a hash of each files' contents. In the future, we should be able to reproduce
         a calculation with confidence that the inputs are identical.
         """
-        caseTitle = cs.caseTitle if cs is not None else os.path.splitext(self.fileName)[0]
+        caseTitle = (
+            cs.caseTitle if cs is not None else os.path.splitext(self.fileName)[0]
+        )
         self.h5db.attrs["caseTitle"] = caseTitle
         if csString is None:
             # don't read file; use what's in the cs now.
@@ -469,12 +490,24 @@ class Database:
 
         if bpString is None:
             bpPath = pathlib.Path(cs.inputDirectory) / cs[CONF_LOADING_FILE]
-            # only store blueprints if we actually loaded from them
-            if bpPath.exists() and bpPath.is_file():
-                # Ensure that the input as stored in the DB is complete
-                bpString = resolveMarkupInclusions(pathlib.Path(cs.inputDirectory) / cs[CONF_LOADING_FILE]).read()
+            if bpPath.suffix == ".h5":
+                # The blueprints are in a database file, they need to be read
+                try:
+                    db = h5py.File(bpPath, "r")
+                    bpString = db["inputs/blueprints"].asstr()[()]
+                except KeyError:
+                    # not all reactors need to be created from blueprints, so they may not exist
+                    bpString = ""
             else:
-                bpString = ""
+                # The blueprints are a standard blueprints yaml that can be read.
+                if bpPath.exists() and bpPath.is_file():
+                    # only store blueprints if we actually loaded from them
+                    # Ensure that the input as stored in the DB is complete
+                    bpString = resolveMarkupInclusions(
+                        pathlib.Path(cs.inputDirectory) / cs[CONF_LOADING_FILE]
+                    ).read()
+                else:
+                    bpString = ""
 
         self.h5db["inputs/settings"] = csString
         self.h5db["inputs/blueprints"] = bpString
@@ -549,7 +582,9 @@ class Database:
         self, timeSteps: Sequence[Tuple[int, int]] = None
     ) -> Generator[h5py._hl.group.Group, None, None]:
         """Returns a generator of HDF5 Groups for all time nodes, or for the passed selection."""
-        assert self.h5db is not None, "Must open the database before calling genTimeStepGroups"
+        assert (
+            self.h5db is not None
+        ), "Must open the database before calling genTimeStepGroups"
         if timeSteps is None:
             for groupName, h5TimeNodeGroup in sorted(self.h5db.items()):
                 match = self.timeNodeGroupPattern.match(groupName)
@@ -568,7 +603,9 @@ class Database:
 
     def genTimeSteps(self) -> Generator[Tuple[int, int], None, None]:
         """Returns a generator of (cycle, node) tuples that are present in the DB."""
-        assert self.h5db is not None, "Must open the database before calling genTimeSteps"
+        assert (
+            self.h5db is not None
+        ), "Must open the database before calling genTimeSteps"
         for groupName in sorted(self.h5db.keys()):
             match = self.timeNodeGroupPattern.match(groupName)
             if match:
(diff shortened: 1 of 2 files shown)


armi/utils/textProcessors.py
```diff
@@ -232,6 +232,7 @@ def _resolveMarkupInclusions(
 
     if isinstance(src, pathlib.Path):
         # this is inefficient, but avoids having to play with io buffers
+        print(src)
```
john-science:
```suggestion
```

john-science: Debug print statement.

armi/bookkeeping/db/database.py
```diff
@@ -390,9 +390,13 @@ def loadCS(self, handleInvalids=True):
         custom Application you do not have access to, the DB may not be usable.
         """
         cs = settings.Settings()
-        cs.caseTitle = os.path.splitext(os.path.basename(self.fileName))[0]
+        cs.path = self.fileName
```
john-science, line 393: This is an API-breaking change, but I do not see the need for it.

People use the caseTitle, and this change would break it for most people's workflows. I think if you want to remove the Settings.caseTitle, that should be a different PR.

armi/bookkeeping/db/database.py
```diff
@@ -469,12 +473,22 @@ def writeInputsToDB(self, cs, csString=None, bpString=None):
 
         if bpString is None:
             bpPath = pathlib.Path(cs.inputDirectory) / cs[CONF_LOADING_FILE]
-            # only store blueprints if we actually loaded from them
-            if bpPath.exists() and bpPath.is_file():
-                # Ensure that the input as stored in the DB is complete
-                bpString = resolveMarkupInclusions(pathlib.Path(cs.inputDirectory) / cs[CONF_LOADING_FILE]).read()
+            if bpPath.suffix == ".h5":
```
john-science: This is not sufficient. A lot of times people capitalize it such as `.H5`. And on occasion people use `.hdf5` and `.HDF5`.

armi/bookkeeping/report/reportingUtils.py
```diff
@@ -102,13 +102,20 @@ def _listInputFiles(cs):
         pathToLoading = pathlib.Path(cs.inputDirectory) / cs[CONF_LOADING_FILE]
 
         if pathToLoading.is_file():
-            includedBlueprints = [inclusion[0] for inclusion in textProcessors.findYamlInclusions(pathToLoading)]
+            if pathToLoading.suffix == ".h5":
```
john-science:
```suggestion
            if pathToLoading.suffix.lower() in (".h5", ".hdf5"):
```

john-science: I believe this is more correct.

armi/bookkeeping/db/database.py
alexhjames: I'm not getting rid of `caseTitle`. The path is just the correct thing to set here so that the caseTitle is populated. The old way of setting the `caseTitle` directly makes the path incorrect. At this point in the code, the input folder is blank so by setting `caseTitle` you get a path that is just a file in the local folder. See https://github.com/terrapower/armi/blob/2c9ea3c371db4589b98c78e48e4de999b7434156/armi/settings/caseSettings.py#L113-L138

armi/utils/textProcessors.py
alexhjames: removed

armi/bookkeeping/report/reportingUtils.py
alexhjames: Thanks, updated

armi/bookkeeping/db/database.py
alexhjames: updated

armi/bookkeeping/db/database.py
john-science: Okay, that setting/getter logic is at least 6 years old. So changing that is a probably a bad solution. ok

armi/bookkeeping/db/tests/test_databaseInterface.py
```diff
@@ -233,6 +233,70 @@ def test_noSyncDbAfterWrite(self):
...
+        # Write this reactor to a database file.
+        dbi = DatabaseInterface(self.r, self.o.cs)
+        dbi.initDB(fName="testDB1.h5")
+        db = dbi.database
+        db.writeToDB(self.r)
+        db.close()
+
+        # Now load the db again
+        with Database("testDB1.h5", "r") as db:
+            cs2 = db.loadCS()
```
john-science, line 252:
```suggestion
            cs2 = db.loadCS()
            self.assertTrue(isinstance(cs2, settings.Settings))
```

john-science: Tests should have asserts.

armi/bookkeeping/db/tests/test_databaseInterface.py
```diff
@@ -233,6 +233,70 @@ def test_noSyncDbAfterWrite(self):
...
+            r2 = db.load(0, 0, cs=cs2)
```
john-science, line 253: I want to suggest this here:

```python
            self.assertTrue(isinstance(r2, Reactor))
```

But you will need to `from armi.reactor.reactors import Reactor` at the top of the file, for it to work.

armi/bookkeeping/db/tests/test_databaseInterface.py
```diff
@@ -233,6 +233,70 @@ def test_noSyncDbAfterWrite(self):
...
+        dbi = DatabaseInterface(r2, cs2)
+        dbi.initDB(fName="testDB2.h5")
+        db = dbi.database
+        db.writeToDB(r2)
+        db.close()
+
+        # Now load this db. It should load
+        with Database("testDB2.h5", "r") as db:
+            cs3 = db.loadCS()
+            _ = db.load(0, 0, cs=cs3)
```
john-science: I want to suggest this here:

```python
            self.assertTrue(isinstance(r3, Reactor))
```

But you will need to `from armi.reactor.reactors import Reactor` at the top of the file, for it to work.

armi/bookkeeping/db/tests/test_databaseInterface.py
```diff
@@ -233,6 +233,70 @@ def test_noSyncDbAfterWrite(self):
...
+        # Now write this db to this folder
+        dbi = DatabaseInterface(r2, cs2)
+        dbi.initDB(fName="testDB2.h5")
+        db = dbi.database
+        db.writeToDB(r2)
+        db.close()
+
+        # Now load this db. It should load
+        with Database("testDB2.h5", "r") as db:
+            cs3 = db.loadCS()
```
john-science, line 267:
```suggestion
            cs3 = db.loadCS()
            self.assertTrue(isinstance(cs3, settings.Settings))
```

john-science: The ARMI docs says all tests should have asserts.

armi/bookkeeping/db/tests/test_databaseInterface.py
```diff
@@ -233,6 +233,70 @@ def test_noSyncDbAfterWrite(self):
...
+            _ = db.load(0, 0, cs=cs3)
```
john-science: Ditto above, with the asserts.

armi/bookkeeping/report/reportingUtils.py
```diff
@@ -102,13 +102,20 @@ def _listInputFiles(cs):
...
+                includedBlueprints = [inclusion[0] for inclusion in textProcessors.findYamlInclusions(pathToLoading)]
         else:
             includedBlueprints = []
 
         inputInfo = []
         inputFiles = [
-            ("Case Settings", cs.caseTitle + ".yaml"),
+            (
+                "Case Settings",
+                os.path.basename(cs.path),
```
john-science:
```suggestion
                os.path.basename(cs.path) if cs.path else cs.caseTitle + ".yaml",
```

john-science: I believe people run these reports even on cases that are entirely in memory, and there are no actual input files. Which, yes, the name of the method makes that sound crazy. But here we are.

I'm not married to this implementation. But I think `cs.path` will be `None` sometimes, and so the `os.path.basename()` will throw an error. That's all.

armi/bookkeeping/db/tests/test_databaseInterface.py
alexhjames: updated to have asserts

armi/bookkeeping/db/tests/test_databaseInterface.py
alexhjames: updated to have asserts

armi/bookkeeping/db/tests/test_databaseInterface.py
alexhjames: updated to have asserts

armi/bookkeeping/db/tests/test_databaseInterface.py
alexhjames: updated to have asserts

armi/bookkeeping/db/tests/test_databaseInterface.py
alexhjames: updated to have asserts

armi/reactor/blueprints/__init__.py
```diff
@@ -116,19 +119,35 @@ def loadFromCs(cs, roundTrip=False):
...
-                        "Please run the `modify` entry point on this case to automatically convert."
-                        "".format(cs[CONF_LOADING_FILE])
-                    )
-                raise
+        bpPath = pathlib.Path(cs[CONF_LOADING_FILE])
+        print(f"loading file {bpPath}")
+        if bpPath.suffix.lower() in (".h5", ".hdf5"):
+            # This is a case settings from a database so the blueprints are also in the database.
+            # try:
+            print(os.listdir())
```
john-science:
```suggestion
```

armi/reactor/blueprints/__init__.py
```diff
@@ -116,19 +119,35 @@ def loadFromCs(cs, roundTrip=False):
...
+            db = h5py.File(bpPath, "r")
+            bpString = db["inputs/blueprints"].asstr()[()]
+            # print(f"bpString: {bpString}")
```
john-science:
```suggestion
```

armi/reactor/blueprints/__init__.py
```diff
@@ -116,19 +119,35 @@ def loadFromCs(cs, roundTrip=False):
...
+            stream = io.StringIO(bpString)
+            stream = Blueprints.migrate(stream)
+            bp = Blueprints.load(stream)
+            # except KeyError:
```
john-science:
```suggestion

```

armi/reactor/blueprints/__init__.py
```diff
@@ -116,19 +119,35 @@ def loadFromCs(cs, roundTrip=False):
...
+        else:
+            with open(cs[CONF_LOADING_FILE], "r") as bpYaml:
+                root = bpPath.parent.absolute()
+                bpYaml = textProcessors.resolveMarkupInclusions(bpYaml, root)
+                try:
+                    bp = Blueprints.load(bpYaml, roundTrip=roundTrip)
+                except yamlize.yamlizing_error.YamlizingError as err:
+                    if "cross sections" in err.args[0]:
+                        runLog.error(
+                            "The loading file {} contains invalid `cross sections` input. "
```
john-science, line 142: Can we make this an f-string?

Thanks!

armi/reactor/blueprints/__init__.py
```diff
@@ -116,19 +119,35 @@ def loadFromCs(cs, roundTrip=False):
...
-                        "The loading file {} contains invalid `cross sections` input. "
-                        "Please run the `modify` entry point on this case to automatically convert."
-                        "".format(cs[CONF_LOADING_FILE])
-                    )
-                raise
+        bpPath = pathlib.Path(cs[CONF_LOADING_FILE])
+        print(f"loading file {bpPath}")
+        if bpPath.suffix.lower() in (".h5", ".hdf5"):
+            # This is a case settings from a database so the blueprints are also in the database.
+            # try:
```
john-science:
```suggestion
```

armi/bookkeeping/db/database.py
```diff
@@ -469,12 +475,22 @@ def writeInputsToDB(self, cs, csString=None, bpString=None):
...
-                # Ensure that the input as stored in the DB is complete
-                bpString = resolveMarkupInclusions(pathlib.Path(cs.inputDirectory) / cs[CONF_LOADING_FILE]).read()
+            if bpPath.suffix.lower() in (".h5", ".hdf5"):
+                # The blueprints are in a database file, they need to be read
+                try:
+                    db = h5py.File(bpPath, "r")
+                    bpString = db["inputs/blueprints"].asstr()[()]
+                except KeyError:
+                    # not all reactors need to be created from blueprints, so they may not exist
+                    bpString = ""
```
keckler, line 485: This change apparently allows for one to specify an h5 file for the `loadingFile` case setting.

Why?

The description for `loadingFile` says that this file should be a blueprints file:
image

john-science: I believe @alexhjames has a workflow where you open one H5/DB file and use it to create a new one.

Is that right Alex?

alexhjames: Yes, if you load a Reactor using a database file, then ARMI extracts the case settings and blueprints from the database file to make the Reactor object. In that case, the database file is tracked as the "source" of the case settings and database information so that things are linked properly if you were to write that Reactor to a new database. Previously the code was tracking the original blueprints/case setting file which meant that when writing the new database it wouldn't write the blueprint information if the original blueprints file wasn't accessible anymore. The lack of blueprint information would make the written database unable to be opened. 

The workflow for this is: 
1) Someone using a case settings yaml/blueprints runs an ARMI app to create a database file (database 1).
2) Database 1 is moved somewhere else or given to someone else. 
3) New person loads a Reactor from Database 1
4) New person modifies the data on the Reactor (adds information)
5) New person writes a database file (database 2) from the Reactor (this database in the old code would be missing blueprints information but the operation will succeed)
6) New person 2 tries to open database 2 to create a Reactor to look at the information (in the old code, this load operation would fail unless the original case settings and blueprint yaml files were accessible at the same path they were when database 1 was written).

PR #2298, opened 2025-09-22
Preserve loading of CoordinateLocation in db load

### LINKED ISSUE #2295: CoordinateLocations are converted to IndexLocations during database load
If you do not define an ASCII grid and rely on `mult` to control the auto creation of the spatial grid, things like duct and coolant that don't live on the spatial grid will be assigned a "coordinate location" via `HexBlock.autoCreateSpatialGrid`

https://github.com/terrapower/armi/blob/2c9ea3c371db4589b98c78e48e4de999b7434156/armi/reactor/blocks.py#L2161-L2168

These are attributed to the center of the block.

If you dump this to database and reload the reactor, you don't get a `CoordinateLocation` object out: you get an `IndexLocation`. It's still assigned to the center of the block, but the object type is different.

I got bit by this because I have some logic that handles behavior differently if a component is assigned a `CoordinateLocation`. During a restart calc, where the reactor is read from DB, the duct and coolant were assigned an `IndexLocation` not a `CoordinateLocation`.

My minimal confirmation is loading a reactor from DB and looking at the spatial locators for the components on a block that does not have an ASCII grid. The block `bf` is such a block loaded from the DB
```python
>>> [(c, c.spatialLocator) for c in bf]
[(<Circle: slug>, <MultiIndexLocation with 61 locations>),
 (<Circle: clad>, <MultiIndexLocation with 61 locations>),
 (<Helix: wire>, <MultiIndexLocation with 61 locations>),
 (<DerivedShape: coolant>, <IndexLocation @ (0,0,0)>),
 (<Hexagon: duct>, <IndexLocation @ (0,0,0)>),
 (<Hexagon: intercoolant>, <IndexLocation @ (0,0,0)>)]
```

drewj-tp:
When loading the DB, pass around the locator type for the spatial locator.

This helps us know what type of locator we've got. For index locations
and multi-index locations, we can directly query the grid[i,j,k] and get
the correct IndexLocation / MultiIndexLocation object out. But, if we
pass the indices of the coordinate location in, we get an index location
out. And that's problematic because we don't get a 1:1 identical reactor
on the reload.

We dump the location type information to DB already. This changes the
layout `_initComps` to produce that information in sync w/ the location
information. Then we can do a check if we're looking at a coordinate
location or not. This means if a coordinate location goes in, a
coordinate location comes out of the database.

Closes #2295 

## SCR Information

Change Type: fixes

One-Sentence Description: Recover any CoordinateLocation spatial locators during DB load

One-line Impact on Requirements: Correct behavior during DB load covered by R_ARMI_DB_TIME

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database.py
+++ b/armi/bookkeeping/db/database.py
@@ -61,6 +61,7 @@ from armi import context, getApp, getPluginManagerOrFail, meta, runLog, settings
 from armi.bookkeeping.db.jaggedArray import JaggedArray
 from armi.bookkeeping.db.layout import (
     DB_VERSION,
+    LOC_COORD,
     Layout,
     replaceNonesWithNonsense,
     replaceNonsenseWithNones,
@@ -800,7 +801,7 @@ class Database:
 
     def _compose(self, comps, cs, parent=None):
         """Given a flat collection of all of the ArmiObjects in the model, reconstitute the hierarchy."""
-        comp, _, numChildren, location = next(comps)
+        comp, _, numChildren, location, locationType = next(comps)
 
         # attach the parent early, if provided; some cases need the parent attached for the rest of
         # _compose to work properly.
@@ -821,7 +822,14 @@ class Database:
         # set the spatialLocators on each component
         if location is not None:
             if parent is not None and parent.spatialGrid is not None:
-                comp.spatialLocator = parent.spatialGrid[location]
+                if locationType != LOC_COORD:
+                    # We can directly index into the spatial grid for IndexLocation
+                    # and MultiIndexLocators to get equivalent spatial locators
+                    comp.spatialLocator = parent.spatialGrid[location]
+                else:
+                    comp.spatialLocator = grids.CoordinateLocation(
+                        location[0], location[1], location[2], parent.spatialGrid
+                    )
             else:
                 comp.spatialLocator = grids.CoordinateLocation(location[0], location[1], location[2], None)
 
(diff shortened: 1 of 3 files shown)


armi/bookkeeping/db/database.py
```diff
@@ -821,7 +822,14 @@ def _compose(self, comps, cs, parent=None):
...
             if parent is not None and parent.spatialGrid is not None:
-                comp.spatialLocator = parent.spatialGrid[location]
+                if locationType != LOC_COORD:
+                    # We can directly index into the spatial grid for IndexLocation
+                    # and MultiIndexLocators to get equivalent spatial locators
+                    comp.spatialLocator = parent.spatialGrid[location]
+                else:
+                    comp.spatialLocator = grids.CoordinateLocation(
+                        location[0], location[1], location[2], parent.spatialGrid
+                    )
```
john-science, line 832: Okay, I see how this fixes your problem.

Are we sure there is never a situation where `"C"` is read from the database and it's correct?

drewj-tp: >Are we sure there is never a situation where "C" is read from the database and it's correct?

I don't follow. If we write in a `"C"` that means we want a `CoordinateLocation`. And if we directly index into the grid with `grid[location]`, we get either an `IndexLocation` or `MultiIndexLocation`

https://github.com/terrapower/armi/blob/0487a293d091c63d6722c2b9dff8cdd0ca662d95/armi/reactor/grids/structuredGrid.py#L244-L278

PR #2301, opened 2025-09-24
Set pin indices on more types of pin components

mgjarrett:

`pinIndices` was recently added (#2202 and #2236) as a way to identify the indices for a given component within a block that has multiple types of pins. However, the `pinIndices` parameter would only be set for some types of component within the block. This update sets `pinIndices` for all components within a block that match one of the target flags.

## SCR Information

Change Type: features

One-Sentence Description: Add `pinIndices` to more types of components.

---

## Checklist


john-science:
Closed in favor of: https://github.com/terrapower/armi/pull/2302

Issue #473, opened 2021-11-12
Make NuclideBases not global


youngmit:
The directory in `nuclideBases` is global, which leads to considerable pain:
 - Import order is very sensitive, and lots of things have to happen at import time, and cant happen again or else 💥.
 - The actual contents change when a burn chain is imposed, which is really case dependent and can only happen once, or else 💥
 - Some code changes the LFP properties, and these changes probably don't survive round-tripping through the database. Not exactly related to global state, but part of the reason we haven't been thinking of it. See `collapseFissionProducts` and `updateYieldVector` methods within the `fissionProductModel`.

So, we should be thinking of another place for this to live, and spin up an object containing the valid nuclides for that scope. Reactor? Case? Somewhere else? Even a global database that is actually an object instead of a magical singleton-esque horror would be better than it is now. Also, if it is mutable beyond the initial setup with a burn chain, then we need to think about how to recover it's state from the database.

jakehader:
We made a step in a direction where global nuclide bases are less bad than they used to be by adding more nuclides in #998. Unfortunately, the nuclide data is still global and the burn chain modifications still may occur when it's being imposed. I specifically added logging statements if a user overwrites any default parameter from the `burn-chain.yaml` or custom file, but the transmutation and decay data that is a subset of the nuclides still gets modified based on the user-defined burn chain data. 

We can probably take this further still.

john-science:
@opotowsky 

This has reared it's head again, and will have to be put on the calendar for ARMI `0.4.0`.  It turns out ARMI enforces everyone to use RIPL nuclear data, but we have some users that use ENDF in other parts of their modeling chain.  So I would LOVE to give people a setting to choose which nuclear data they want.  But we can't do that (easily) if the nuclear data lives in the global scope at import time.

So, finally, I think we have enough interest to make this ticket happen.

john-science:
I notice that it is not just `nuclideBases` that is global, but also `elements` and `thermalScattering`. We even have [an unrelated ticket open](https://github.com/terrapower/armi/issues/1804) to rip `thermalScattering` out of the global scope.

So, let's try to itemize the nuclide information that is currently in the global scope, and where (outside the `nucDirectory` area) it is used:

## nuclideBases

| Name | Usage |
|---|---|
[...]

Issue #1195, opened 2023-02-24
New Setting to swap in different Nuclide data


john-science:
We recently [updated our RIPL nuclides](https://github.com/terrapower/armi/pull/998).  But some people are asking for the ability to use different nuclide data files in ARMI:

* [armi/resources/nuclides.dat](https://github.com/terrapower/armi/blob/3e47ec6e6359870866502c9356747cf1da4f9515/armi/resources/nuclides.dat)
* [armi/resources/elements.dat](https://github.com/terrapower/armi/blob/3e47ec6e6359870866502c9356747cf1da4f9515/armi/resources/elements.dat)

Reasons for this include backporting to older versions of the nuclide data to check and make sure old results haven't changed. Though some people might be working on improving nuclide data, and want to test that out.

This seems like a good use for a `Setting`, but I might also accept some reasonably easy code solution for changing out the nuclide library.

john-science:
@jakehader @drewj-usnctech  FYI.  Here is my first draft of this ticket.  Feel free to holler if we need to change the verbage.

john-science:
This ticket will be easily doable, but only after we remove global nuclides: https://github.com/terrapower/armi/issues/473

So, that is my current plan.

PR #2303, opened 2025-09-24
Step0 in removing global nuclides: create classes

john-science:

> This PR does not change the ARMI API.

This is just step zero of the plan to remove global nuclides as outlined in issue #473.

This PR does not actually remove global nuclides, it sets the groundwork in a non-API-breaking way. Essentially, the files `nuclideBases.py` and `elements.py` have a bunch of data saved to the global scope. So this PR creates the classes `NuclideBases` and `Elements` to contain that data.

## SCR Information

Change Type: features

One-Sentence Description: Making classes to contain the global data in `nuclideBases.py` and `elements.py`.

One-line Impact on Requirements: The code related to nuclideBases and elements is being refactored to be more contained: `I_ARMI_ND_ELEMENTS0`, `I_ARMI_ND_ELEMENTS1`, `I_ARMI_ND_DATA0`, `I_ARMI_ND_DATA1`, `I_ARMI_ND_ISOTOPES0`, `I_ARMI_ND_ISOTOPES1`, `I_ARMI_ND_ISOTOPES2`, `I_ARMI_ND_ISOTOPES3`, `I_ARMI_ND_ISOTOPES4`, `I_ARMI_ND_ISOTOPES5`, `I_ARMI_ND_ISOTOPES6`, and `I_ARMI_ND_ISOTOPES7` for `R_ARMI_ND_ELEMENTS`, `R_ARMI_ND_DATA`, and `R_ARMI_ND_ISOTOPES`.

---

## Checklist

### DIFF
--- a/armi/nucDirectory/elements.py
+++ b/armi/nucDirectory/elements.py
@@ -13,26 +13,22 @@
 # limitations under the License.
 
 """
-This module provides fundamental element information to be used throughout the framework
-and applications.
+This module provides fundamental element information to be used throughout the framework and applications.
 
 .. impl:: A tool for querying basic data for elements of the periodic table.
     :id: I_ARMI_ND_ELEMENTS0
     :implements: R_ARMI_ND_ELEMENTS
 
     The :py:mod:`elements <armi.nucDirectory.elements>` module defines the
-    :py:class:`Element <armi.nucDirectory.elements.Element>` class which acts as
-    a data structure for organizing information about an individual element,
-    including number of protons, name, chemical symbol, phase (at STP), periodic
-    table group, standard weight, and a list of isotope :py:class:`nuclideBase
-    <armi.nucDirectory.nuclideBases.NuclideBase>` instances. The module includes
-    a factory that generates the :py:class:`Element
-    <armi.nucDirectory.elements.Element>` instances by reading from the
-    ``elements.dat`` file stored in the ARMI resources folder.  When an
-    :py:class:`Element <armi.nucDirectory.elements.Element>` instance is
-    initialized, it is added to a set of global dictionaries that are keyed by
-    number of protons, element name, and element symbol. The module includes
-    several helper functions for querying these global dictionaries.
+    :py:class:`Element <armi.nucDirectory.elements.Element>` class which acts as a data structure for organizing
+    information about an individual element, including number of protons, name, chemical symbol, phase (at STP),
+    periodic table group, standard weight, and a list of isotope
+    :py:class:`nuclideBase <armi.nucDirectory.nuclideBases.NuclideBase>` instances. The module includes a factory that
+    generates the :py:class:`Element <armi.nucDirectory.elements.Element>` instances by reading from the
+    ``elements.dat`` file stored in the ARMI resources folder. When an
+    :py:class:`Element <armi.nucDirectory.elements.Element>` instance is initialized, it is added to a set of global
+    dictionaries that are keyed by number of protons, element name, and element symbol. The module includes several
+    helper functions for querying these global dictionaries.
 
 The element class structure is outlined :ref:`here <elements-class-diagram>`.
 
@@ -136,9 +132,10 @@ from typing import List
 from armi import context
 from armi.utils.units import HEAVY_METAL_CUTOFF_Z
 
-byZ = {}
-byName = {}
-bySymbol = {}
+elements = None
+byZ = None
+byName = None
+bySymbol = None
 
 
 class ChemicalPhase(Enum):
@@ -242,8 +239,7 @@ class Element:
 
         Notes
         -----
-        This method will filter out any NaturalNuclideBases from the `nuclides`
-        attribute.
+        This method will filter out any NaturalNuclideBases from the `nuclides` attribute.
         """
         return [nuc for nuc in self.nuclides if nuc.abundance > 0.0 and nuc.a > 0]
 
@@ -253,199 +249,74 @@ class Element:
 
         Notes
         -----
-        Heavy metal in this instance is not related to an exact weight or density
-        cut-off, but rather is designated for nuclear fuel burn-up evaluations, where
-        the initial heavy metal mass within a component should be tracked. It is typical
-        to include any element/nuclide above Actinium.
+        Heavy metal in this instance is not related to an exact weight or density cut-off, but rather is designated for
+        nuclear fuel burn-up evaluations, where the initial heavy metal mass within a component should be tracked. It is
+        typical to include any element/nuclide above Actinium.
         """
         return self.z > HEAVY_METAL_CUTOFF_Z
 
 
 def getElementsByChemicalPhase(phase: ChemicalPhase) -> List[Element]:
-    """
-    Returns all elements that are of the given chemical phase.
-
-    Parameters
-    ----------
-    phase: ChemicalPhase
-        This should be one of the valid options from the `ChemicalPhase` class.
-
-    Returns
-    -------
-    elems : List[Element]
-        A list of elements that are associated with the given chemical phase.
-    """
-    elems = []
-    if not isinstance(phase, ChemicalPhase):
-        raise ValueError(f"{phase} is not an instance of {ChemicalPhase}")
-    for element in byName.values():
-        if element.phase == phase:
-            elems.append(element)
-    return elems
+    """Pass through to Elements.getElementsByChemicalPhase() for the global Elements object."""
+    global elements
+    return elements.getElementsByChemicalPhase(phase)
 
 
 def getElementsByChemicalGroup(group: ChemicalGroup) -> List[Element]:
-    """
-    Returns all elements that are of the given chemical group.
-
-    Parameters
-    ----------
-    group: ChemicalGroup
-        This should be one of the valid options from the `ChemicalGroup` class.
-
-    Returns
-    -------
-    elems : List[Element]
-        A list of elements that are associated with the given chemical group.
-    """
-    elems = []
-    if not isinstance(group, ChemicalGroup):
-        raise ValueError(f"{group} is not an instance of {ChemicalGroup}")
-    for element in byName.values():
-        if element.group == group:
-            elems.append(element)
-    return elems
+    """Pass through to Elements.getElementsByChemicalGroup() for the global Elements object."""
+    global elements
+    return elements.getElementsByChemicalGroup(group)
 
 
 def getName(z: int = None, symbol: str = None) -> str:
-    r"""
-    Returns element name.
-
-    Parameters
-    ----------
-    z : int
-        Atomic number
-    symbol : str
-        Element abbreviation e.g. 'Zr'
-
-    Examples
-    --------
-    >>> elements.getName(10)
-    'Neon'
-    >>> elements.getName(symbol='Ne')
-    'Neon'
-    """
-    element = None
-    if z:
-        element = byZ[z]
-    else:
-        element = byName[symbol.upper()]
-    return element.name
+    """Pass through to Elements.getName() for the global Elements object."""
+    global elements
+    return elements.getName(z, symbol)
 
 
 def getSymbol(z: int = None, name: str = None) -> str:
-    r"""
-    Returns element abbreviation given atomic number Z.
-
-    Parameters
-    ----------
-    z : int
-        Atomic number
-    name : str
-        Element name E.g. Zirconium
-
-    Examples
-    --------
-    >>> elements.getSymbol(10)
-    'Ne'
-    >>> elements.getSymbol(name='Neon')
-    'Ne'
-
-    """
-    element = None
-    if z:
-        element = byZ[z]
-    else:
-        element = byName[name.lower()]
-    return element.symbol
+    """Pass through to Elements.getSymbol() for the global Elements object."""
+    global elements
+    return elements.getSymbol(z, name)
 
 
 def getElementZ(symbol: str = None, name: str = None) -> int:
-    """
-    Get element atomic number given a symbol or name.
-
-    Parameters
-    ----------
-    symbol : str
-        Element symbol e.g. 'Zr'
-    name : str
-        Element name e.g. 'Zirconium'
-
-    Examples
-    --------
-    >>> elements.getZ('Zr')
-    40
-    >>> elements.getZ(name='Zirconium')
-    40
-
-    Notes
-    -----
-    Element Z is stored in elementZBySymbol, indexed by upper-case element symbol.
-    """
-    if not symbol and not name:
-        return None
-    element = None
-    if symbol:
-        element = bySymbol[symbol.upper()]
-    else:
-        element = byName[name.lower()]
-    return element.z
+    """Pass through to Elements.getElementZ() for the global Elements object."""
+    global elements
+    return elements.getElementZ(symbol, name)
 
 
 def factory():
-    """
-    Generate the :class:`Elements <Element>` instances.
-
-    .. warning::
-        This method gets called by default when loading the module, so don't call it
-        unless you know what you're doing.
-        Any existing :class:`Nuclides <armi.nucDirectory.nuclide.Nuclide>`
-        may lose their reference to the underlying :class:`Element`.
-    """
-    destroyGlobalElements()
-    with open(os.path.join(context.RES, "elements.dat"), "r") as f:
-        for line in f:
-            # Skip header lines
-            if line.startswith("#") or line.startswith("Z"):
-                continue
-            # read z, symbol, name, phase, and chemical group
-            lineData = line.split()
-            z = int(lineData[0])
-            sym = lineData[1].upper()
-            name = lineData[2]
-            phase = lineData[3]
-            group = lineData[4]
-            standardWeight = lineData[5]
-            e = Element(z, sym, name, phase, group)
-            if standardWeight != "Derived":
-                e.standardWeight = float(standardWeight)
-
-
-def addGlobalElement(element: Element):
-    """Add an element to the global dictionaries."""
-    if element.z in byZ or element.name in byName or element.symbol in bySymbol:
-        raise ValueError(f"{element} has already been added and cannot be duplicated.")
-
-    byZ[element.z] = element
-    byName[element.name] = element
-    bySymbol[element.symbol] = element
-
-
-def destroyGlobalElements():
-    """Delete all global elements."""
+    """Pass through to Elements.factory() for the global Elements object."""
+    print("xxxxxxxxxxxxxxxxxxxxxxxxxx Elements factory xxxxxxxxxxxxxxxxxxxxxxxxxx")
+    global elements
     global byZ
     global byName
     global bySymbol
 
-    byZ.clear()
-    byName.clear()
-    bySymbol.clear()
+    elements = Elements()
+    elements.factory()
+
+    byZ = elements.byZ
+    byName = elements.byName
+    bySymbol = elements.bySymbol
 
 
-factory()
+def addGlobalElement(element: Element):
+    """Pass through to Elements.addElement() for the global Elements object."""
+    global elements
+    elements.addElement(element)
+
+
+def destroyGlobalElements():
+    """Pass through to Elements.clear() for the global Elements object."""
+    global elements
+    elements.clear()
 
 
 """
+TODO: This split is only temporary. Global nuclides are going away. Document that.
+
 TODO: Above this point is the old "global nuclides" code. Soon to be deleted.
       Below this point is the new code.
 """
@@ -629,3 +500,6 @@ class Elements:
             element = self.byName[name.lower()]
 
         return element.z
+
+
+factory()
--- a/armi/nucDirectory/elements.py
+++ b/armi/nucDirectory/elements.py
@@ -165,7 +165,7 @@ class ChemicalGroup(Enum):
 class Element:
     """Represents an element defined on the Periodic Table."""
 
-    def __init__(self, z, symbol, name, phase="UNKNOWN", group="UNKNOWN"):
+    def __init__(self, z, symbol, name, phase="UNKNOWN", group="UNKNOWN", skipGlobal=False):
         """
         Creates an instance of an Element.
 
@@ -206,7 +206,8 @@ class Element:
         self.group = ChemicalGroup[group]
         self.standardWeight = None
         self.nuclides = []
-        addGlobalElement(self)
+        if not skipGlobal:
+            addGlobalElement(self)
 
     def __repr__(self):
         return f"<Element {self.symbol:>3s} (Z={self.z}), {self.name}, {self.group}, {self.phase}>"
@@ -442,3 +443,189 @@ def destroyGlobalElements():
 
 
 factory()
+
+
+"""
+TODO: Above this point is the old "global nuclides" code. Soon to be deleted.
+      Below this point is the new code.
+"""
+
+
+class Elements:
+    """TODO."""
+
+    def __init__(self):
+        self.byZ = {}
+        self.byName = {}
+        self.bySymbol = {}
+        self.elementsFile = os.path.join(context.RES, "elements.dat")
+
+    def clear(self):
+        """Empty all the data in this collection."""
+        self.byZ.clear()
+        self.byName.clear()
+        self.bySymbol.clear()
+
+    def addElement(self, element: Element):
+        """Add an element to this collection."""
+        if element.z in self.byZ or element.name in self.byName or element.symbol in self.bySymbol:
+            raise ValueError(f"{element} has already been added and cannot be duplicated.")
+
+        self.byZ[element.z] = element
+        self.byName[element.name] = element
+        self.bySymbol[element.symbol] = element
+
+    def factory(self):
+        """Generate the :class:`Elements <Element>` instances."""
+        self.clear()
+        with open(self.elementsFile, "r") as f:
+            for line in f:
+                # Skip header lines
+                if line.startswith("#") or line.startswith("Z"):
+                    continue
+                # read z, symbol, name, phase, and chemical group
+                lineData = line.split()
+                z = int(lineData[0])
+                sym = lineData[1].upper()
+                name = lineData[2]
+                phase = lineData[3]
+                group = lineData[4]
+                standardWeight = lineData[5]
+                e = Element(z, sym, name, phase, group, skipGlobal=True)
+                if standardWeight != "Derived":
+                    e.standardWeight = float(standardWeight)
+                self.addElement(e)
+
+    def getElementsByChemicalPhase(self, phase: ChemicalPhase) -> List[Element]:
+        """
+        Returns all elements that are of the given chemical phase.
+
+        Parameters
+        ----------
+        phase: ChemicalPhase
+            This should be one of the valid options from the `ChemicalPhase` class.
+
+        Returns
+        -------
+        elems : List[Element]
+            A list of elements that are associated with the given chemical phase.
+        """
+        elems = []
+        if not isinstance(phase, ChemicalPhase):
+            raise ValueError(f"{phase} is not an instance of {ChemicalPhase}")
+
+        for element in self.byName.values():
+            if element.phase == phase:
+                elems.append(element)
+
+        return elems
+
+    def getElementsByChemicalGroup(self, group: ChemicalGroup) -> List[Element]:
+        """
+        Returns all elements that are of the given chemical group.
+
+        Parameters
+        ----------
+        group: ChemicalGroup
+            This should be one of the valid options from the `ChemicalGroup` class.
+
+        Returns
+        -------
+        elems : List[Element]
+            A list of elements that are associated with the given chemical group.
+        """
+        elems = []
+        if not isinstance(group, ChemicalGroup):
+            raise ValueError(f"{group} is not an instance of {ChemicalGroup}")
+
+        for element in self.byName.values():
+            if element.group == group:
+                elems.append(element)
+
+        return elems
+
+    def getName(self, z: int = None, symbol: str = None) -> str:
+        r"""
+        Returns element name.
+
+        Parameters
+        ----------
+        z : int
+            Atomic number
+        symbol : str
+            Element abbreviation e.g. 'Zr'
+
+        Examples
+        --------
+        >>> elements.getName(10)
+        'Neon'
+        >>> elements.getName(symbol='Ne')
+        'Neon'
+        """
+        element = None
+        if z:
+            element = self.byZ[z]
+        else:
+            element = self.byName[symbol.upper()]
+
+        return element.name
+
+    def getSymbol(self, z: int = None, name: str = None) -> str:
+        r"""
+        Returns element abbreviation given atomic number Z.
+
+        Parameters
+        ----------
+        z : int
+            Atomic number
+        name : str
+            Element name E.g. Zirconium
+
+        Examples
+        --------
+        >>> elements.getSymbol(10)
+        'Ne'
+        >>> elements.getSymbol(name='Neon')
+        'Ne'
+
+        """
+        element = None
+        if z:
+            element = self.byZ[z]
+        else:
+            element = self.byName[name.lower()]
+
+        return element.symbol
+
+    def getElementZ(self, symbol: str = None, name: str = None) -> int:
+        """
+        Get element atomic number given a symbol or name.
+
+        Parameters
+        ----------
+        symbol : str
+            Element symbol e.g. 'Zr'
+        name : str
+            Element name e.g. 'Zirconium'
+
+        Examples
+        --------
+        >>> elements.getZ('Zr')
+        40
+        >>> elements.getZ(name='Zirconium')
+        40
+
+        Notes
+        -----
+        Element Z is stored in elementZBySymbol, indexed by upper-case element symbol.
+        """
+        if not symbol and not name:
+            return None
+
+        element = None
+        if symbol:
+            element = self.bySymbol[symbol.upper()]
+        else:
+            element = self.byName[name.lower()]
+
+        return element.z
--- a/armi/nucDirectory/elements.py
+++ b/armi/nucDirectory/elements.py
@@ -313,16 +313,12 @@ def destroyGlobalElements():
     elements.clear()
 
 
-"""
-TODO: This split is only temporary. Global nuclides are going away. Document that.
-
-TODO: Above this point is the old "global nuclides" code. Soon to be deleted.
-      Below this point is the new code.
-"""
-
-
 class Elements:
-    """TODO."""
+    """
+    A container for all the elements information in the simulation.
+
+    By design, you would only expect to have one instance of this object in memory during a simulation.
+    """
 
     def __init__(self):
         self.byZ = {}
--- a/armi/nucDirectory/elements.py
+++ b/armi/nucDirectory/elements.py
@@ -288,7 +288,6 @@ def getElementZ(symbol: str = None, name: str = None) -> int:
 
 def factory():
     """Pass through to Elements.factory() for the global Elements object."""
-    print("xxxxxxxxxxxxxxxxxxxxxxxxxx Elements factory xxxxxxxxxxxxxxxxxxxxxxxxxx")
     global elements
     global byZ
     global byName
--- a/armi/nucDirectory/elements.py
+++ b/armi/nucDirectory/elements.py
@@ -170,18 +170,14 @@ class Element:
             :id: I_ARMI_ND_ELEMENTS1
             :implements: R_ARMI_ND_ELEMENTS
 
-            The :py:class:`Element <armi.nucDirectory.elements.Element>` class
-            acts as a data structure for organizing information about an
-            individual element, including number of protons, name, chemical
-            symbol, phase (at STP), periodic table group, standard weight, and a
-            list of isotope
-            :py:class:`nuclideBase <armi.nucDirectory.nuclideBases.NuclideBase>`
-            instances.
+            The :py:class:`Element <armi.nucDirectory.elements.Element>` class acts as a data structure for organizing
+            information about an individual element, including number of protons, name, chemical symbol, phase (at STP),
+            periodic table group, standard weight, and a list of isotope
+            :py:class:`nuclideBase <armi.nucDirectory.nuclideBases.NuclideBase>` instances.
 
-            The :py:class:`Element <armi.nucDirectory.elements.Element>` class
-            has a few methods for appending additional isotopes, checking
-            whether an isotope is naturally occurring, retrieving the natural
-            isotopic abundance, or whether the element is a heavy metal.
+            The :py:class:`Element <armi.nucDirectory.elements.Element>` class has a few methods for appending
+            additional isotopes, checking whether an isotope is naturally occurring, retrieving the natural isotopic
+            abundance, or whether the element is a heavy metal.
 
         Parameters
         ----------
--- a/armi/nucDirectory/nuclideBases.py
+++ b/armi/nucDirectory/nuclideBases.py
@@ -1312,7 +1312,9 @@ class NuclideBases:
         matches = [nuc for nuc in self.instances if predicate(nuc)]
         if len(matches) != 1:
             raise IndexError(
-                f"Expected single match, but got {len(matches)} matches:\n  {'\n  '.join(str(mo) for mo in matches)}"
+                "Expected single match, but got {} matches:\n  {}".format(
+                    len(matches), "\n  ".join(str(mo) for mo in matches)
+                )
             )
 
         return matches[0]
--- a/armi/nucDirectory/nuclideBases.py
+++ b/armi/nucDirectory/nuclideBases.py
@@ -1024,7 +1024,6 @@ def imposeBurnChain(burnChainStream):
 
 def factory():
     """Pass through to NuclideBases.factory() for the global NuclideBases object."""
-    print("xxxxxxxxxxxxxxxxxxxxxxxxxx NuclideBases factory xxxxxxxxxxxxxxxxxxxxxxxxxx")
     global nuclideBases
     global burnChainImposed
     global instances
@@ -1201,7 +1200,7 @@ class NuclideBases:
         missingActiveNuclides = set()
         memo = set()
         nucNames = [nucName.decode() for nucName in nuclides]
-        difference = set(numberDensities).difference(memo)
+        difference = set(nucNames).difference(memo)
         while any(difference):
             newNucs = set()
             nuclide = difference.pop()
@@ -1217,8 +1216,8 @@ class NuclideBases:
                     # Interaction nuclides can only be added to the number density dictionary if they are a part of the
                     # user-defined active nuclides
                     productNuclide = interaction.getPreferredProduct(activeNuclides)
-                    if productNuclide not in numberDensities:
-                        numberDensities[productNuclide] = 0.0
+                    if productNuclide not in nucNames:
+                        newNucs.add(productNuclide.encode())
                 except KeyError:
                     # Keep track of the first production nuclide
                     missingActiveNuclides.add(interaction.productNuclides)
--- a/armi/nucDirectory/nuclideBases.py
+++ b/armi/nucDirectory/nuclideBases.py
@@ -100,20 +100,13 @@ import numpy as np
 from ruamel.yaml import YAML
 
 from armi import context, runLog
-from armi.nucDirectory import transmutations
+from armi.nucDirectory import elements, transmutations
 from armi.utils.units import HEAVY_METAL_CUTOFF_Z
-
-# Used to prevent multiple applications of burn chains, which would snowball unphysically. This is a bit of a crutch for
-# the global state that is the nuclide directory.
-burnChainImposed = False
-
-instances = []
-# The elements must be imported after the instances list is established to allow for simultaneous initialization of the
-# nuclides and elements together to maintain self-consistency.
-from armi.nucDirectory import elements  # noqa: E402
 
 # Global nuclide and nuclideBases data
 nuclideBases = None
+instances = []
+burnChainImposed = False
 byName = None
 byDBName = None
 byLabel = None
@@ -123,7 +116,6 @@ byMcc3IdEndfbVII0 = None
 byMcc3IdEndfbVII1 = None
 byMcnpId = None
 byAAAZZZSId = None
-
 
 # lookup table from https://t2.lanl.gov/nis/data/endf/endfvii-n.html
 BASE_ENDFB7_MAT_NUM = {
--- a/armi/nucDirectory/nuclideBases.py
+++ b/armi/nucDirectory/nuclideBases.py
@@ -19,25 +19,29 @@ framework and applications.
     :id: I_ARMI_ND_ISOTOPES0
     :implements: R_ARMI_ND_ISOTOPES
 
-    The :py:mod:`nuclideBases <armi.nucDirectory.nuclideBases>` module defines the
-    :py:class:`NuclideBase <armi.nucDirectory.nuclideBases.NuclideBase>` class which is used to
-    organize and store metadata about each nuclide. The metadata is read from ``nuclides.dat`` file
-    in the ARMI resources folder, which contains metadata for 4,614 isotopes. The module also
-    contains classes for special types of nuclides, including :py:class:`DummyNuclideBase
+    The :py:mod:`nuclideBases <armi.nucDirectory.nuclideBases>` module defines
+    the :py:class:`NuclideBase <armi.nucDirectory.nuclideBases.NuclideBase>`
+    class which is used to organize and store metadata about each nuclide. The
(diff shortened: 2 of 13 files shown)


armi/nucDirectory/elements.py
```diff
@@ -127,6 +123,12 @@ def getAttributes(element):
                                     tableFmt='rst'),
                            caption='List of elements',
                            label='nuclide-bases-table')
+
+Notes
+-----
+Currently, this module contains a lot of data in the global scope. But ARMI is in the process of encapsulating this
+data, moving it out of the global scope, making it part of the reactor data model, and making it configurable via
+Settings. Pardon the mess during this transition.
```
drewj-tp, line 131: I know you don't like putting links to things in source code. But I believe, in this case, a link to #473 would be appropriate. That would help people know when this "transition" is done

Non-blocking request.

armi/nucDirectory/elements.py
```diff
@@ -165,26 +168,22 @@ class ChemicalGroup(Enum):
 class Element:
     """Represents an element defined on the Periodic Table."""
 
-    def __init__(self, z, symbol, name, phase="UNKNOWN", group="UNKNOWN"):
+    def __init__(self, z, symbol, name, phase="UNKNOWN", group="UNKNOWN", skipGlobal=False):
```
drewj-tp: I know it's a transitory thing. But the double negative "skipGlobal=False" is not my favorite. I'm thinking "okay so it's false which means we don't skip adding to globals so we do add to globals"

Non-blocking request to change to `addToGlobal=True` to reduce cognitive load
```suggestion
    def __init__(self, z, symbol, name, phase="UNKNOWN", group="UNKNOWN", addToGlobal=True):
```

drewj-tp: This is also not documented in the docstring. Maybe it's not intended to be called by users. Just wanted to call this out

armi/nucDirectory/elements.py
```diff
@@ -206,7 +205,8 @@ def __init__(self, z, symbol, name, phase="UNKNOWN", group="UNKNOWN"):
         self.group = ChemicalGroup[group]
         self.standardWeight = None
         self.nuclides = []
-        addGlobalElement(self)
+        if not skipGlobal:
```
drewj-tp:
```suggestion
        if addToGlobal:
```

armi/nucDirectory/elements.py
```diff
@@ -252,193 +251,252 @@ def isHeavyMetal(self):
...
-            group = lineData[4]
-            standardWeight = lineData[5]
-            e = Element(z, sym, name, phase, group)
-            if standardWeight != "Derived":
-                e.standardWeight = float(standardWeight)
 
+    def __init__(self):
+        self.byZ = {}
+        self.byName = {}
+        self.bySymbol = {}
```
drewj-tp: These attributes should be documented. Please consider adding them to an `Attributes` section of the class docstring

armi/nucDirectory/elements.py
```diff
@@ -252,193 +251,252 @@ def isHeavyMetal(self):
...
+        self.elementsFile = os.path.join(context.RES, "elements.dat")
+
+    def clear(self):
+        """Empty all the data in this collection."""
+        self.byZ.clear()
+        self.byName.clear()
+        self.bySymbol.clear()
+
+    def addElement(self, element: Element):
+        """Add an element to this collection."""
```
drewj-tp: Non-blocking request to document the `ValueError` in a `Raises` section of the method docstring

armi/nucDirectory/elements.py
drewj-tp: Bonus points for in-line type hints 🤩 e.g.,
```
self.byZ: dict[int, Element]
self.byName: dict[str, Element]
```

drewj-tp: Some additional thoughts.

It's possible for someone to add something directly into the `byZ` dictionary since it's public. That's not ideal.

We could get around this by making these
- protected attributes, e.g., `self._byZ`,
- exposing them as public properties wrapped in a [`types.MappingProxyType`](https://docs.python.org/3/library/types.html#types.MappingProxyType)

Then, the only way to modify these dictionaries is to go through `addElement`

drewj-tp: This can be a follow-on ticket as well. If so, this can be resolved by creating the ticket and linking it here.

armi/nucDirectory/elements.py
```diff
@@ -252,193 +251,252 @@ def isHeavyMetal(self):
...
+                lineData = line.split()
+                z = int(lineData[0])
+                sym = lineData[1].upper()
+                name = lineData[2]
+                phase = lineData[3]
+                group = lineData[4]
+                standardWeight = lineData[5]
+                e = Element(z, sym, name, phase, group, skipGlobal=True)
+                if standardWeight != "Derived":
+                    e.standardWeight = float(standardWeight)
```
drewj-tp, line 381: Potential follow-on ticket: allow pass this work off to an `Element` classmethod e.g., `e = Element.fromLine(line)`

Non-blocking request.

armi/nucDirectory/elements.py
```diff
@@ -252,193 +251,252 @@ def isHeavyMetal(self):
...
+                    continue
+                # read z, symbol, name, phase, and chemical group
+                lineData = line.split()
+                z = int(lineData[0])
+                sym = lineData[1].upper()
+                name = lineData[2]
+                phase = lineData[3]
+                group = lineData[4]
+                standardWeight = lineData[5]
+                e = Element(z, sym, name, phase, group, skipGlobal=True)
```
drewj-tp: If my suggestion re double negatives is accepted
```suggestion
                e = Element(z, sym, name, phase, group, addToGlobal=False)
```

armi/nucDirectory/elements.py
```diff
@@ -252,193 +251,252 @@ def isHeavyMetal(self):
...
+            A list of elements that are associated with the given chemical phase.
+        """
+        elems = []
+        if not isinstance(phase, ChemicalPhase):
+            raise ValueError(f"{phase} is not an instance of {ChemicalPhase}")
 
-    byZ[element.z] = element
-    byName[element.name] = element
-    bySymbol[element.symbol] = element
+        for element in self.byName.values():
```
drewj-tp, line 402: Should this class have an `__iter__` method? We have this pattern of `for e in self.byName.values()` a few times to do just that.

Could be a follow-on ticket, if you like. Non-blocking request.

armi/nucDirectory/elements.py
```diff
@@ -252,193 +251,252 @@ def isHeavyMetal(self):
...
-    if element.z in byZ or element.name in byName or element.symbol in bySymbol:
-        raise ValueError(f"{element} has already been added and cannot be duplicated.")
+        Returns
+        -------
+        elems : List[Element]
+            A list of elements that are associated with the given chemical phase.
+        """
+        elems = []
+        if not isinstance(phase, ChemicalPhase):
+            raise ValueError(f"{phase} is not an instance of {ChemicalPhase}")
```
drewj-tp: :shrug: It was a value error before but the error is raised b/c the thing is not the correct type so 🤷 
```suggestion
            raise TypeError(f"{phase} is not an instance of {ChemicalPhase}")
```
Non-blocking as this would also make it an API breaking change _somewhere_ _maybe_

armi/nucDirectory/nuclideBases.py
```diff
@@ -88,6 +92,11 @@
 >>> nuclideBases.byAAAZZZSId['2350920']
 <NuclideBase U235:  Z:92, A:235, S:0, W:2.350439e+02, Label:U235>, HL:2.22160758861e+16, Abund:7.204000e-03>
 
+Notes
+-----
+Currently, this module contains a lot of data in the global scope. But ARMI is in the process of encapsulating this
+data, moving it out of the global scope, making it part of the reactor data model, and making it configurable via
+Settings. Pardon the mess during this transition.
```
drewj-tp, line 99: Same non-blocking comment re linking #473 here.

armi/nucDirectory/nuclideBases.py
```diff
@@ -286,21 +289,25 @@ class INuclide(NuclideInterface):
...
     label : str
-        ARMI's unique 4 character label for the nuclide. These are not human readable, but do not
-        lose any information. The label is effectively the
-        :attr:`Element.symbol `armi.nucDirectory.elements.Element.symbol` padded to two characters,
-        plus the mass number (A) in base-26 (0-9, A-Z). Additional support for meta-states is
-        provided by adding 100 * the state to the mass number (A).
+        ARMI's unique 4 character label for the nuclide.
+        These are not human readable, but do not lose any information.
+        The label is effectively the
+        :attr:`Element.symbol `armi.nucDirectory.elements.Element.symbol`
```
drewj-tp: This is not a valid link. It doesn't render as a clickable reference the API docs - https://terrapower.github.io/armi/.apidocs/armi.nucDirectory.nuclideBases.html#armi.nucDirectory.nuclideBases.INuclide

image

Non-blocking since it was here before. Could be a separate ticket

armi/nucDirectory/nuclideBases.py
```diff
@@ -321,6 +328,7 @@ def __init__(
         mcc2id=None,
         mcc3idEndfbVII0=None,
         mcc3idEndfbVII1=None,
+        skipGlobal=False,
```
drewj-tp:
```suggestion
        addToGlobal=True,
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -356,7 +364,8 @@ def __init__(
         self.mcc2id = mcc2id or ""
         self.mcc3idEndfbVII0 = mcc3idEndfbVII0 or ""
         self.mcc3idEndfbVII1 = mcc3idEndfbVII1 or ""
-        addGlobalNuclide(self)
+        if not skipGlobal:
```
drewj-tp:
```suggestion
        if addToGlobal:
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -499,15 +500,15 @@ class NuclideBase(INuclide, IMcnpNuclide):
...
+        The :py:class:`NuclideBase <armi.nucDirectory.nuclideBases.NuclideBase>` class provides a data structure for
+        information about a single nuclide, including the atom number, atomic weight, element, isomeric state,
+        half-life, and name. The class contains static methods for creating an internal ARMI name or label for a
+        nuclide. There are instance methods for generating the nuclide ID for external codes, e.g. MCNP or Serpent, and
+        retrieving the nuclide ID for MC\ :sup:`2`-2 or MC\ :sup:`2`-3. There are also instance methods for generating
+        an AAAZZZS ID and an ENDF MAT number.
     """
 
-    def __init__(self, element, a, weight, abundance, state, halflife):
+    def __init__(self, element, a, weight, abundance, state, halflife, skipGlobal=False):
```
drewj-tp:
```suggestion
    def __init__(self, element, a, weight, abundance, state, halflife, addToGlobal=True):
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -519,6 +520,7 @@ def __init__(self, element, a, weight, abundance, state, halflife):
             halflife=halflife,
             name=NuclideBase._createName(element, a, state),
             label=NuclideBase._createLabel(element, a, state),
+            skipGlobal=skipGlobal,
```
drewj-tp:
```suggestion
            addToGlobal=addToGlobal,
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -674,21 +677,20 @@ def getSerpentId(self):
         Returns
         -------
         id: str
-            The ID of this nuclide based on it's elemental name, weight, and state, eg ``U-235``,
-            ``Te-129m``.
+            The ID of this nuclide based on it's elemental name, weight,
+            and state, eg ``U-235``, ``Te-129m``,
```
drewj-tp: Period to end the sentence
```suggestion
            and state, eg ``U-235``, ``Te-129m``.
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -708,7 +710,7 @@ def getEndfMatNum(self):
 
         isotopeNum = (a - smallestStableA) * 3 + self.state + 25
         mat = z * 100 + isotopeNum
-        return "{0}".format(mat)
+        return f"{mat}"
```
drewj-tp: Might as well just
```suggestion
        return str(mat)
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -717,12 +719,11 @@ class NaturalNuclideBase(INuclide, IMcnpNuclide):
...
     -----
-    This is meant to represent the combination of all naturally occurring nuclides within an
-    element. The abundance is forced to zero here so that it does not have any interactions with the
-    NuclideBase objects.
+    This is meant to represent the combination of all naturally occurring nuclides within an element. The abundance is
+    forced to zero here so that it does not have any interactions with the NuclideBase objects.
     """
 
-    def __init__(self, name, element):
+    def __init__(self, name, element, skipGlobal=False):
```
drewj-tp:
```suggestion
    def __init__(self, name, element, addToGlobal=True):
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -733,6 +734,7 @@ def __init__(self, name, element):
             halflife=np.inf,
             name=name,
             label=name,
+            skipGlobal=skipGlobal,
```
drewj-tp:
```suggestion
            addToGlobal=addToGlobal,
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -788,17 +790,17 @@ def getSerpentId(self):
...
             runLog.warning(
-                "The only elemental in ENDF/B VII.1 is carbon. "
-                "ENDF mat num was requested for the elemental {} and will not be helpful "
-                "for working with ENDF/B VII.1. Try to expandElementalsToIsotopics".format(self)
+                f"The only elemental in ENDF/B VII.1 is carbon. ENDF mat num was requested for the elemental {self} and"
+                "will not be helpful for working with ENDF/B VII.1. Try to expandElementalsToIsotopics"
             )
-        return "{0}".format(self.z * 100)
+
+        return f"{self.z * 100}"
```
drewj-tp:
```suggestion
        return str(self.z * 100)
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -807,11 +809,11 @@ class DummyNuclideBase(INuclide):
...
     Notes
     -----
-    This may be used to store mass from a depletion calculation, specifically in the instances where
-    the burn chain is truncated.
+    This may be used to store mass from a depletion calculation, specifically in the instances where the burn chain is
+    truncated.
     """
 
-    def __init__(self, name, weight):
+    def __init__(self, name, weight, skipGlobal=False):
```
drewj-tp:
```suggestion
    def __init__(self, name, weight, addToGlobal=True):
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -822,6 +824,7 @@ def __init__(self, name, weight):
             halflife=np.inf,
             name=name,
             label="DMP" + name[4],
+            skipGlobal=skipGlobal,
```
drewj-tp:
```suggestion
            addToGlobal=addToGlobal,
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -885,7 +888,7 @@ class LumpNuclideBase(INuclide):
         Describes what nuclides LumpNuclideBase is expend to.
     """
 
-    def __init__(self, name, weight):
+    def __init__(self, name, weight, skipGlobal=False):
```
drewj-tp:
```suggestion
    def __init__(self, name, weight, addToGlobal=True):
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -896,6 +899,7 @@ def __init__(self, name, weight):
             halflife=np.inf,
             name=name,
             label=name[1:],
+            skipGlobal=skipGlobal,
```
drewj-tp:
```suggestion
            addToGlobal=addToGlobal,
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
+        self.instances = []
+        self.byName = {}
+        self.byDBName = {}
+        self.byLabel = {}
+        self.byMcc2Id = {}
+        self.byMcc3Id = {}
+        self.byMcc3IdEndfbVII0 = {}
+        self.byMcc3IdEndfbVII1 = {}
+        self.byMcnpId = {}
+        self.byAAAZZZSId = {}
```
drewj-tp, line 1139: I don't think this will clear the global state. The global values will still point to the previous dictionaries, and the instance here will have new dictionaries. Mini-example
```python
>>> class NB:
...     def __init__(self):
...         self.byName = {}
...     def add(self, k, v):
...         self.byName[k] = v
...     def clear(self):
...         self.byName = {}
...
>>> bases = NB()
>>> bases.add("hello", "world")
>>> byName = bases.byName
>>> bases.clear()
>>> bases.byName
{}
>>> byName
{'hello': 'world'}
```
In the elements, we use `dict.clear` which won't have this problem because the same _object_ exists, it's just been wiped out.

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
+        self.instances = []
+        self.byName = {}
+        self.byDBName = {}
+        self.byLabel = {}
+        self.byMcc2Id = {}
+        self.byMcc3Id = {}  # for backwards compat. Identical to byMcc3IdEndfbVII1
+        self.byMcc3IdEndfbVII0 = {}
+        self.byMcc3IdEndfbVII1 = {}
+        self.byMcnpId = {}
+        self.byAAAZZZSId = {}
```
drewj-tp: Similar, non-blocking potential future ticket to make these protected (`self._instances`, `self._byName`, etc.) and expose read-only variants via properties

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
+        self.nuclidesFile = os.path.join(context.RES, "nuclides.dat")
+        self.mccNuclidesFile = os.path.join(context.RES, "mcc-nuclides.yaml")
+        self.factory()
```
drewj-tp, line 1111: This is well outside of the scope of this PR so I'm putting down some thoughts here that we can move to a follow up issue if you want.

---

Do we really want it to be impossible to make one of these empty? I'm generally not a fan of lots of work being done in the constructor of a class. It can make testing difficult. Right now, if we want to test that we can add a nuclide to one of these bases, we have to fully construct the entire set (or mock around `.factory()`) just to do a small operation.

What if we have the `factory` be a class method, e.g., `nuclideBases = NuclideBases.factory()`? Which would
1. Constuct an empty `NuclideBases` instance,
2. Populate that class the same way, just outside the constructor

But the current `factory` method has a lot of helper methods, some protected some private (?), and it intrinsically tied to one data source / data type. Maybe this should be further broken out

1. Class that just holds the nuclide objects, simple add/get/clear methods
2. Builder class responsible for doing the `factory` work and configuring the thing.

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
-    """
-    Apply transmutation and decay information to each nuclide.
+        self.addNuclideBases()
+        self.__addNaturalNuclideBases()
+        self.__addDummyNuclideBases()
+        self.__addLumpedFissionProductNuclideBases()
+        self.updateNuclideBasesForSpecialCases()
+        self.readMCCNuclideData()
+        self.__renormalizeNuclideToElementRelationship()
+        self.__deriveElementalWeightsByNaturalNuclideAbundances()
```
drewj-tp, line 1200: No action needed. But this is a weird mix of public / private methods that still mutate the base object. Why some are public, some are private is not clear.

I don't think any of these should be public: we don't want someone calling e.g., `nb.addNuclideBases` multiple times. So maybe all private / protected is the move?

:shrug:

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
+        >>> len(isomers90)
+        3
+        >>> for iso in isomers: print(iso)
+        <NuclideBase MO95: Z:42, A:95, S:0, label:MO2N>
+        <NuclideBase NB95: Z:41, A:95, S:0, label:NB2N>
+        <NuclideBase ZR95: Z:40, A:95, S:0, label:ZR2N>
+        """
+        for nuc in self.instances:
+            if predicate(nuc):
+                yield (nuc)
```
drewj-tp: Pretty sure this can be reduced to
```suggestion
        return filter(predicate, self.instances)
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
+        """
+        nuclideBase.label = newLabel
+        self.byLabel[newLabel] = nuclideBase
 
-    .. impl:: The special case name Am242g is supported.
-        :id: I_ARMI_ND_ISOTOPES6
-        :implements: R_ARMI_ND_ISOTOPES
+    def getDepletableNuclides(self, activeNuclides, obj):
+        """Get nuclides in this object that are in the burn chain."""
+        return sorted(set(activeNuclides) & set(obj.getNuclides()))
```
drewj-tp: This doesn't use `self` nor the "global" state. Can this function be left alone in the refactor?

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
 
-    # Update the pointer of `AM242` to refer to `AM242M`.
-    am242m = byName["AM242M"]
-    byName["AM242"] = am242m
-    byDBName["nAm242"] = am242m
-    byDBName[byName["AM242"].getDatabaseName()] = am242m
+        Notes
+        -----
+        We believe the transmutation information would probably be better stored on a less fundamental place (e.g. not
+        on the NuclideBase).
```
drewj-tp: What do we want to do with this note? It seems we're pretty keen on keeping the transmutation information on this fundamental state. Looks like this aside has existed since this method was added.

1. Should we remove this note?
2. Should we make a ticket to put this information in a better home?

Non-blocking. Just a thought that we've been okay (maybe begrudgingly so) with this information being here for three years

armi/nucDirectory/nuclideBases.py
drewj-tp: Similar request to document these in the class docstring. There is presently little to no documentation on what `byName` means, both in the global sense nor on this attribute.

drewj-tp: Please also consider type hints for these. I have the lack of type hint information on `byName` and `byMcnpId` to be a source of confusion / slow down on more than one occasion.

Yes, I know they're nuclides. But my IDE doesn't so when I try to do e.g., `byName[name].getMccId` I need to pull up the documentation to find out the exact name of the method (is it getMccID, getMccId?) when my IDE could tell me exactly

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
+                abun = float(lineData[6])
+                halflife = lineData[7]
+                if halflife == "inf":
+                    halflife = np.inf
+                else:
+                    halflife = float(halflife)
+                nuSF = float(lineData[8])
+
+                element = elements.bySymbol[sym]
+                nb = NuclideBase(element, a, mass, abun, state, halflife, skipGlobal=True)
```
drewj-tp:
```suggestion
                nb = NuclideBase(element, a, mass, abun, state, halflife, addToGlobal=False)
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
+                nb.nuSF = nuSF
+                self.addNuclide(nb)
+
+    def __addNaturalNuclideBases(self):
+        """Generates a complete set of nuclide bases for each naturally occurring element."""
+        for element in elements.byZ.values():
+            if element.symbol not in self.byName:
+                if element.isNaturallyOccurring():
+                    self.addNuclide(NaturalNuclideBase(element.symbol, element, skipGlobal=True))
```
drewj-tp:
```suggestion
                    self.addNuclide(NaturalNuclideBase(element.symbol, element, addToGlobal=False))
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
-        byMcnpId[nuclide.getMcnpId()] = nuclide
-    if not isinstance(nuclide, (NaturalNuclideBase, LumpNuclideBase, DummyNuclideBase)):
-        # There are no AZS ID for elements / natural nuclides, or fictitious lump or dummy nuclides
-        byAAAZZZSId[nuclide.getAAAZZZSId()] = nuclide
+        Notes
+        -----
+        These nuclides can be used to truncate a depletion / burn-up chain within the MC2 program.
+        """
+        self.addNuclide(DummyNuclideBase(name="DUMP1", weight=10.0, skipGlobal=True))
+        self.addNuclide(DummyNuclideBase(name="DUMP2", weight=240.0, skipGlobal=True))
```
drewj-tp:
```suggestion
        self.addNuclide(DummyNuclideBase(name="DUMP1", weight=10.0, addToGlobal=False))
        self.addNuclide(DummyNuclideBase(name="DUMP2", weight=240.0, addToGlobal=False))
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
+
+    def __addLumpedFissionProductNuclideBases(self):
+        """Generates a set of nuclides for use as lumped fission products."""
+        self.addNuclide(LumpNuclideBase(name="LFP35", weight=233.273, skipGlobal=True))
+        self.addNuclide(LumpNuclideBase(name="LFP38", weight=235.78, skipGlobal=True))
+        self.addNuclide(LumpNuclideBase(name="LFP39", weight=236.898, skipGlobal=True))
+        self.addNuclide(LumpNuclideBase(name="LFP40", weight=237.7, skipGlobal=True))
+        self.addNuclide(LumpNuclideBase(name="LFP41", weight=238.812, skipGlobal=True))
+        self.addNuclide(LumpNuclideBase(name="LREGN", weight=1.0, skipGlobal=True))
```
drewj-tp:
```suggestion
        self.addNuclide(LumpNuclideBase(name="LFP35", weight=233.273, addToGlobal=False))
        self.addNuclide(LumpNuclideBase(name="LFP38", weight=235.78, addToGlobal=False))
        self.addNuclide(LumpNuclideBase(name="LFP39", weight=236.898, addToGlobal=False))
        self.addNuclide(LumpNuclideBase(name="LFP40", weight=237.7, addToGlobal=False))
        self.addNuclide(LumpNuclideBase(name="LFP41", weight=238.812, addToGlobal=False))
        self.addNuclide(LumpNuclideBase(name="LREGN", weight=1.0, addToGlobal=False))
```

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
+                self.byMcc2Id[nb.getMcc2Id()] = nb
+            if mcc3idEndfbVII0 is not None:
+                nb.mcc3idEndfbVII0 = mcc3idEndfbVII0
+                self.byMcc3IdEndfbVII0[nb.getMcc3IdEndfbVII0()] = nb
+            if mcc3idEndfbVII1 is not None:
+                nb.mcc3idEndfbVII1 = mcc3idEndfbVII1
+                self.byMcc3IdEndfbVII1[nb.getMcc3IdEndfbVII1()] = nb
+
+        # Have the byMcc3Id dictionary be VII.1 IDs.
+        self.byMcc3Id = self.byMcc3IdEndfbVII1
```
drewj-tp, line 1501: Note: if `byMcc3Id` is a property, we don't need to set it here. We can just have e.g.,
```python
@property
def byMcc3Id(self):
    return self.byMcc3IdEndfbVIII1
```
and we don't need another dictionary running around. Other comments on read-only via `MappingProxyType` and protected attributes apply

armi/nucDirectory/nuclideBases.py
```diff
@@ -950,458 +954,566 @@ def getMcc3IdEndfbVII1(self):
...
+
+        # Update the pointer of `AM242` to refer to `AM242M`.
+        am242m = self.byName["AM242M"]
+        self.byName["AM242"] = am242m
+        self.byDBName["nAm242"] = am242m
+        self.byDBName[self.byName["AM242"].getDatabaseName()] = am242m
+
+    def __renormalizeNuclideToElementRelationship(self):
+        """Fill in the missing element data for each nuclide."""
+        for nuc in self.instances:
```
drewj-tp, line 1536: Non-blocking thing for us to consider later: `__iter__` looping over `self.instances`

armi/nucDirectory/tests/test_nuclideBases.py
```diff
@@ -22,31 +22,37 @@
...
-            nuclideBases.byName["Cat"]
+            self.nuclideBases.byName["Cat"]
 
     def test_nucBase_AllAbundancesAddToOne(self):
         for zz in range(1, 102):
-            nuclides = nuclideBases.isotopes(zz)
+            nuclides = elements.byZ[zz].nuclides
             # We only process nuclides with measured masses. Some are purely theoretical, mostly over z=100
-            self.assertTrue(len(nuclides) > 0, msg="z={} unexpectedly has no nuclides".format(zz))
+            self.assertTrue(len(nuclides) > 0, msg=f"z={zz} unexpectedly has no nuclides")
```
drewj-tp: Non-blocking since this existed before: [`TestCase.assertGreater`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertGreater)
```suggestion
            self.assertGreater(len(nuclides), 0, msg=f"z={zz} unexpectedly has no nuclides")
```
The check is functionally the same, but the default error message will be more helpful

armi/nucDirectory/tests/test_nuclideBases.py
```diff
@@ -59,18 +65,18 @@ def test_nucBase_AllAbundancesAddToOne(self):
 
     def test_nucBases_AllLabelsAreUnique(self):
         labels = []
-        for nn in nuclideBases.instances:
-            self.assertTrue(nn.label not in labels, "Label already exists: {}".format(nn.label))
+        for nn in self.nuclideBases.instances:
+            self.assertTrue(nn.label not in labels, f"Label already exists: {nn.label}")
```
drewj-tp: Non-blocking since this existed before: [`TestCase.assertNotIn`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertNotIn)
```suggestion
            self.assertNotIn(nn.label, labels, f"Label already exists: {nn.label}")
```
The check is functionally the same, but the default error message will be more helpful

armi/nucDirectory/tests/test_nuclideBases.py
```diff
@@ -127,16 +133,16 @@ def test_nucBasesLabelAndNameCollsAreForSameNuc(self):
...
         count = 0
-        for nuc in nuclideBases.where(lambda nn: nn.name == nn.label):
+        for nuc in self.nuclideBases.where(lambda nn: nn.name == nn.label):
             count += 1
-            self.assertEqual(nuc, nuclideBases.byName[nuc.name])
-            self.assertEqual(nuc, nuclideBases.byDBName[nuc.getDatabaseName()])
-            self.assertEqual(nuc, nuclideBases.byLabel[nuc.label])
+            self.assertEqual(nuc, self.nuclideBases.byName[nuc.name])
+            self.assertEqual(nuc, self.nuclideBases.byDBName[nuc.getDatabaseName()])
+            self.assertEqual(nuc, self.nuclideBases.byLabel[nuc.label])
```
drewj-tp: Non-blocking since this was here before
```suggestion
            self.assertIs(nuc, self.nuclideBases.byName[nuc.name])
            self.assertIs(nuc, self.nuclideBases.byDBName[nuc.getDatabaseName()])
            self.assertIs(nuc, self.nuclideBases.byLabel[nuc.label])
```

armi/nucDirectory/tests/test_nuclideBases.py
```diff
@@ -233,28 +239,28 @@ def test_nucBases_Am242m(self):
             :id: T_ARMI_ND_ISOTOPES1
             :tests: R_ARMI_ND_ISOTOPES
         """
-        am242m = nuclideBases.byName["AM242"]
-        self.assertEqual(am242m, nuclideBases.byName["AM242M"])
+        am242m = self.nuclideBases.byName["AM242"]
+        self.assertEqual(am242m, self.nuclideBases.byName["AM242M"])
```
drewj-tp:
```suggestion
        self.assertIs(am242m, self.nuclideBases.byName["AM242M"])
```

armi/nucDirectory/tests/test_nuclideBases.py
```diff
@@ -233,28 +239,28 @@ def test_nucBases_Am242m(self):
...
         self.assertEqual("nAm242m", am242m.getDatabaseName())
-        self.assertEqual(am242m, nuclideBases.byDBName["nAm242"])
+        self.assertEqual(am242m, self.nuclideBases.byDBName["nAm242"])
```
drewj-tp:
```suggestion
        self.assertIs(am242m, self.nuclideBases.byDBName["nAm242"])
```

armi/nucDirectory/tests/test_nuclideBases.py
```diff
@@ -233,28 +239,28 @@ def test_nucBases_Am242m(self):
...
         self.assertAlmostEqual(am242m.weight, 242.059601666)
 
-        am242g = nuclideBases.byName["AM242G"]
-        self.assertEqual(am242g, nuclideBases.byName["AM242G"])
+        am242g = self.nuclideBases.byName["AM242G"]
+        self.assertEqual(am242g, self.nuclideBases.byName["AM242G"])
```
drewj-tp:
```suggestion
        self.assertIs(am242g, self.nuclideBases.byName["AM242G"])
```

armi/nucDirectory/tests/test_nuclideBases.py
```diff
@@ -233,28 +239,28 @@ def test_nucBases_Am242m(self):
...
         self.assertEqual("nAm242g", am242g.getDatabaseName())
-        self.assertEqual(am242g, nuclideBases.byDBName["nAm242g"])
+        self.assertEqual(am242g, self.nuclideBases.byDBName["nAm242g"])
```
drewj-tp:
```suggestion
        self.assertIs(am242g, self.nuclideBases.byDBName["nAm242g"])
```

armi/nucDirectory/thermalScattering.py
```diff
@@ -195,64 +189,78 @@ def _genACELabel(self):
             label = f"{element.symbol.lower()}-{first.a:d}"
         else:
             raise ValueError(f"{self} label cannot be generated")
+
         return label
 
 
-def factory():
+def factory(byName):
```
drewj-tp: New parameter should be documented

armi/nucDirectory/thermalScattering.py
```diff
@@ -195,64 +189,78 @@ def _genACELabel(self):
...
+
+    if "FE56" in byName:
+        fe56 = byName["FE56"]
+        byNbAndCompound[fe56, None] = ThermalScattering(fe56)
+
+    if "H2" in byName:
+        h = byName["H"]
+
+    if "H2" in byName:
+        d = byName["H2"]
```
drewj-tp, line 243:
```suggestion
        h = byName["H"]
        d = byName["H2"]
```

armi/nucDirectory/thermalScattering.py
```diff
@@ -195,64 +189,78 @@ def _genACELabel(self):
...
+            c, CRYSTALLINE_GRAPHITE, f"tsl-{CRYSTALLINE_GRAPHITE}.endf", "grph"
+        )
+
+    if "FE56" in byName:
+        fe56 = byName["FE56"]
+        byNbAndCompound[fe56, None] = ThermalScattering(fe56)
+
+    if "H2" in byName:
+        h = byName["H"]
+
```
drewj-tp: Can be combined w/ below
```suggestion
```

armi/nucDirectory/thermalScattering.py
```diff
@@ -195,64 +189,78 @@ def _genACELabel(self):
...
+        )
+        byNbAndCompound[be, BEO] = ThermalScattering(be, BEO, endf8Label=BEO, aceLabel="be-beo")
+
+    if "C" in byName:
+        c = byName["C"]
+        byNbAndCompound[c, SIC] = ThermalScattering(c, SIC)
+        byNbAndCompound[c, GRAPHITE_10P] = ThermalScattering(c, GRAPHITE_10P, f"tsl-{GRAPHITE_10P}.endf", "grph10")
+        byNbAndCompound[c, GRAPHITE_30P] = ThermalScattering(c, GRAPHITE_30P, f"tsl-{GRAPHITE_30P}.endf", "grph30")
+        byNbAndCompound[c, CRYSTALLINE_GRAPHITE] = ThermalScattering(
+            c, CRYSTALLINE_GRAPHITE, f"tsl-{CRYSTALLINE_GRAPHITE}.endf", "grph"
```
drewj-tp, line 234: These are new? Well, new to this function. I'm sure they existed elsewhere in the global state of things

armi/nucDirectory/thermalScattering.py
```diff
@@ -195,64 +189,78 @@ def _genACELabel(self):
...
-    byNbAndCompound[u, UN] = ThermalScattering(u, UN)
-    byNbAndCompound[zr, ZRH] = ThermalScattering(zr, ZRH)
-    byNbAndCompound[si, SIC] = ThermalScattering(si, SIC)
-    byNbAndCompound[c, CRYSTALLINE_GRAPHITE] = ThermalScattering(
-        c, CRYSTALLINE_GRAPHITE, f"tsl-{CRYSTALLINE_GRAPHITE}.endf", "grph"
-    )
-    byNbAndCompound[c, GRAPHITE_10P] = ThermalScattering(c, GRAPHITE_10P, f"tsl-{GRAPHITE_10P}.endf", "grph10")
-    byNbAndCompound[c, GRAPHITE_30P] = ThermalScattering(c, GRAPHITE_30P, f"tsl-{GRAPHITE_30P}.endf", "grph30")
+    if "AL27" in byName:
+        al27 = byName["AL27"]
```
drewj-tp, line 218: Controversial take, no action needed.

---

This is where the walrus operator would be excellent!
```suggestion
    if (al27 := byName.get("AL27")) is not None:
```
It's been around since py38 so it'll be supported in released ARMI.

drewj-tp (changes requested):
Overall this is very clean. Thank you @john-science!

I have only a few real requests and a lot of things that are more future improvements now that we have these encapsulations and/or thoughts that I had while reading a part of the code base I don't frequent.

## Real requests

### Double negatives

Consider `addToGlobal` over `skipGlobal`. I find the double negatives are sources of confusion / cognitive gymnastics. I _think_ I covered all the usages in my suggestions so _hopefully_, if you agree, you can accept them and it won't cost you any time.
### Clearing global nuclide
I don't think `NuclideBases.clear` is going to _really_ clear the global state since the previous globals will retain references to the old dictionaries, while the attributes on the `NuclideBase` object has been cleared. You could probably verify this by clearing the global state and showing that the global `nuclideBases.byName` attribute is empty while the `byName` dictionary is not

## Things for us to really consider later
### `NuclideBasesBuilder`
The `NuclideBases` class is also responsible for building itself off a single very particular stream of data. Granted, this data ships w/ ARMI so that's not the worst assumption. But, in the future, if we do really want to decouple the storage of data at run time from the source of the data (e.g., bring your own nuclear data #1195), then we should consider pulling out the data parsing from the data storage.

No actions now. Just food for thought // a separate ticket for us to work on later

armi/nucDirectory/thermalScattering.py
john-science: These are not new, it's just a refactor of what was there before. This is a pretty ugly class, and this method is nasty. I had done more refactoring, but it's API breaking. So this is the non-API-breaking part.

armi/nucDirectory/elements.py
john-science: Yeah, I am pretty dead set against HTML links in code. If a docstring doesn't stand on its own, you are demanding people who read the code do research even if just casually glancing at a method.

That is a big request for people who are definitely busy.

armi/nucDirectory/elements.py
john-science: This is very much a temporary argument, that will be deleted in a couple of weeks. I guess we can document it in the mean time. Sure.

armi/nucDirectory/elements.py
john-science: There is a metric ton of work coming. I will add that to the next PR.

armi/nucDirectory/nuclideBases.py
john-science: Well, let me think on it. I did not create this public/private mix. I am just moving existing functions into a class.

Let me look at it. If you don't mind reviewing a bigger PR, I can certainly make more changes.

armi/nucDirectory/nuclideBases.py
john-science: Maybe. It is used a lot downstream. I can look again at my previous refactors and see why I needed to bring this into the class.

I certainly agree it could be a `staticmethod`.

armi/nucDirectory/elements.py
john-science: hmmm...

No, I don't think `byName` is more central to the class than `byZ` or `bySymbol`.

I took a look around the various codebases I have access to, and got usage counts that are APPROXIMATELY like this:

* `byName` = 693
* `byZ` = 334
* `bySymbol` = 228

So, sure, `byName` is the most popular, but by the number is used about the same as the other two combined. I don't want to design the class to make it look like people "should want" or "should use" one of these three over the other. They are simply different ways to find the same data.

armi/nucDirectory/nuclideBases.py
john-science: Sorry, no. If it makes you feel better, this note will be gone in the next 4-8 weeks. So it won't be here long anyway.

armi/nucDirectory/nuclideBases.py
john-science: We don't need a different ticket. I already have a branch open for the next PR in this series. I can throw that into the mix. It's a good idea. Sure.

john-science: Yeah, this was all in the module-level docstring. I was going to do that in the next PR but I can move them down in this one. NBD.

armi/nucDirectory/nuclideBases.py
john-science: The `factory()` method is in the constructor because I am trying to exactly reproduce the old functionality without making any API changes.

Not the next PR, but the following one (the one were I break the API), the factory method _probably_ won't be in the constructor. It will be separated out and called directly by the `Reactor.__init__()` method.

But I can't do any of that right now. The goal is to make this PR small, both for the reviewer's time and to make as few changes as humanly possible. This PR is supposed to have one goal: "encapsulate the current global nuclides logic and data into classes". The next PR I will start making meaningful changes to that logic. Like your setters and getters.

armi/nucDirectory/nuclideBases.py
john-science: A good catch.  It's funny that this didn't break any unit tests in ARMI. Also, I ran all the unit and integration tests I could find downstream, and none of them broke.

Well, I can make the change anyway.

armi/nucDirectory/nuclideBases.py
john-science: All three of the non-private methods here have requirements on them. So that is clearly why they are public. Because of SQA and the fact that we use Sphinx, but hide all private methods from the API.

So, unless we abandon sphinx-needs, or heavily reconfigure how we use Sphinx, these three methods have to remain public for SQA purposes.

armi/nucDirectory/nuclideBases.py
john-science: Okay, I moved it back out to module-level.

armi/nucDirectory/elements.py
drewj-tp: That's not really what I was getting at.

We have a need to iterate over all the elements stored in the container. That is currently handled with `for e in self.byName.values()`. You're correct in saying that would be identical to `for e in self.byZ.values()` since the values in the dicts all point to the same things.

But I'm proposing
```python
def __iter__(self):
    return iter(self.byName.values())
```
to produce an iterator over all the `Element` objects

armi/nucDirectory/nuclideBases.py
john-science: It appears Nick added this comment just about exactly 5 years ago: https://github.com/terrapower/armi/commit/7052135742944fb22b56c3970af32a38446a986b#diff-375fa44c7d0ae0771451fae28972107c7cb321179a932d51569e8b85c5656f99R386

I am not sure I agree. Especially in the near future, I want it to be easy for a Reactor to find a reference to the nuclide bases, and then for a given `INuclide`, easily find its transmutation and decay information.

I think this comment comes from the days when all this information was going to be stored as separate global data.  I don't think this really applies in our upcoming Brave New World.

Yeah, I'm going with (1), I'll just delete it.

armi/nucDirectory/nuclideBases.py
john-science: I'll make them all properties in my other branch. Yeah. (That's the next PR, not this one.)

That's a good simplification.

armi/nucDirectory/nuclideBases.py
john-science: hmmm

By the numbers, I think `instances` is the least used of the class attributes, from outside the class.

Within the class, we loop over `instances` twice and `byName` once.

Hmm... I'm not sure if we can immediately add an `def __iter__` or `def __contains__`, `__getitem__`, or the like. The class is a collection of different data collections, all used to address the same data. It's hard to envision how to make one of them more central than the others.

I'm going to hold off for now. Until we get a better view of what the final class interface might be. If that's okay.

armi/nucDirectory/elements.py
john-science: done

armi/nucDirectory/elements.py
john-science: done

armi/nucDirectory/elements.py
john-science: Somehow I missed this one in my first pass. 

It's a fine idea. Sure.

armi/nucDirectory/elements.py
john-science: By that iterates over the `byName` and not the `byZ`, is what I'm saying.

armi/nucDirectory/nuclideBases.py
john-science: fixed

armi/nucDirectory/nuclideBases.py
john-science: done

armi/nucDirectory/thermalScattering.py
john-science: I don't know. It's not used anywhere else in ARMI. And I feel like code should be consistent. 

Using a cool new tool in one place and not others ends up making code messy.

Maybe. Let me sleep on it.

drewj-tp: no actions needed. just food for though

armi/nucDirectory/thermalScattering.py
john-science: done

armi/nucDirectory/elements.py
drewj-tp: Let's take this offline. Not worth blocking this PR

drewj-tp (approved):
Thanks for the awesome cleanup @john-science

armi/nucDirectory/nuclideBases.py
drewj-tp: Same thread to follow in https://github.com/terrapower/armi/pull/2303#discussion_r2399365832

Resolving b/c it's outside the scope of this PR

PR #2310, opened 2025-09-30
Removing unnecessary code from Assembly.getBlocksBetweenElevations

### LINKED ISSUE #2308: `Assembly.getBlocksBetweenElevations` can raise unwarranted errors
The code in `Assembly.getBlocksBetweenElevations` was updated in #2221 . The update that I am particularly interested in is below:
https://github.com/terrapower/armi/pull/2221/commits/a9d289cf036ffd34777a85f46b3afb3005718efb

The way that the comparisons against `eps` are done in this code allows for errors to be raised that are unwarranted. Admittedly, the window in which this problem might occur is very narrow, but I've had the unfortunately luck of running into it.

My assembly has blocks with the following `ztop` values:
```
54.61
64.77000000000001
79.29170643171996
81.31986788878203
99.64000022387279
117.96013255896355
136.2802648940543
154.60039722914507
172.92052956423584
174.95609982369038
189.6057219568908
209.66609829577732
299.15543714355005
398.45000000000005
```

Alternatively here are the block heights:
```
54.61
10.16000000000001
14.521706431719949
2.028161457062069
18.32013233509076
18.32013233509076
18.32013233509076
18.32013233509076
18.32013233509076
2.035570259454545
14.64962213320041
20.060376338886527
89.48933884777273
99.29456285645
```

If you make an assembly with those block heights and then run `a.getBlocksBetweenElevations(205.47842037674482, 209.66609830000013)` you will get the following error:
```python
ValueError: The cumulative height of [(<DriverPlenumBlock B0216-011 at 006-012-011 XS: N ENV GP: A>, 4.187677919032495)] is 4.187677919032495 cm and does not equal the expected height of 4.187677923255308 cm.
All mesh points: [189.6057219568908, 209.66609829577732, 299.15543714355005]
Upper mesh point: 209.66609830000013 cm
Lower mesh point: 205.47842037674482 cm
```

This error is not warranted, because the difference between the sliver that is cut off from the upper block is, as a fraction of that block's height, `abs(209.66609829577732 - 209.66609830000013) / 89.48933884777273 = 4.7e-11` , which is smaller than `eps=1e-10`.

There are two issues that I identify with the algorithm in that method:
1. When blocks are checked to be potentially thrown out or kept [here](https://github.com/terrapower/armi/blob/7486ca0ac200a0473da0e3f5bc22c7d25c050a32/armi/reactor/assemblies.py#L991-L993), they are checked as relative to the height of the block itself. However when the method [checks again later](https://github.com/terrapower/armi/blob/7486ca0ac200a0473da0e3f5bc22c7d25c050a32/armi/reactor/assemblies.py#L1004-L1011), the comparison is relative to the `expectedHeight`, which could be considerably smaller than any given block's height. This means that a sliver that was judged to be small in comparison to its block height may be thrown out even though it is not that small in comparison to the `zLower` and `zUpper` specified by the user.
2. In the [second check](https://github.com/terrapower/armi/blob/7486ca0ac200a0473da0e3f5bc22c7d25c050a32/armi/reactor/assemblies.py#L1004-L1011), the comparison is actually done against `10*eps`. This is presumably necessary because multiple slivers could have been thrown out, each sliver possibly being as large as `eps`. So the 10x provides a bit of buffer. But this buffer allows for false positives, although the window for this is typically pretty dang small.

My overall suggestion would be to just throw out this chunk of code:
https://github.com/terrapower/armi/blob/7486ca0ac200a0473da0e3f5bc22c7d25c050a32/armi/reactor/assemblies.py#L995-L1011

As far as I can tell, it serves no purpose.

john-science:

The warning we are removing from this method is not strictly useful, and has led to some false warnings being printed in the wild.

close #2308

## SCR Information

Change Type: trivial

One-Sentence Description: Removing unnecessary code from Assembly.getBlocksBetweenElevations.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks.py
+++ b/armi/reactor/blocks.py
@@ -319,12 +319,12 @@ class Block(composites.Composite):
     def assignPinIndices(self):
         pass
 
-    def getMgFlux(self, adjoint=False, average=False, volume=None, gamma=False):
+    def getMgFlux(self, adjoint=False, average=False, gamma=False):
         """
         Returns the multigroup neutron flux in [n/cm^2/s].
 
-        The first entry is the first energy group (fastest neutrons). Each additional
-        group is the next energy group, as set in the ISOTXS library.
+        The first entry is the first energy group (fastest neutrons). Each additional group is the next energy group, as
+        set in the ISOTXS library.
 
         It is stored integrated over volume on self.p.mgFlux
 
@@ -332,15 +332,8 @@ class Block(composites.Composite):
         ----------
         adjoint : bool, optional
             Return adjoint flux instead of real
-
         average : bool, optional
-            If true, will return average flux between latest and previous. Doesn't work
-            for pin detailed yet
-
-        volume: float, optional
-            If average=True, the volume-integrated flux is divided by volume before being returned.
-            The user may specify a volume, or the function will obtain the block volume directly.
-
+            If true, will return average flux between latest and previous. Doesn't work for pin detailed yet.
         gamma : bool, optional
             Whether to return the neutron flux or the gamma flux.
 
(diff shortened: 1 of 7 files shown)


john-science:
I have proven to myself that this breaks nothing downstream.

PR #2314, opened 2025-10-01
Documenting ARMI's testing tools for devs

john-science:
## What is the change? Why is it being made?

ARMI comes with a lot of good tools to help people build unit tests of ARMI code. It's about time we started documenting these.

close #1746

## SCR Information

Change Type: docs

One-Sentence Description: Documenting ARMI's testing tools for devs

One-line Impact on Requirements: NA

---

## Checklist

- [x] This PR has only [one purpose or idea](https://terrapower.github.io/armi/developer/tooling.html#one-idea-one-pr).
- [x] [Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify any new/changed code.
- [x] The [documentation](https://terrapower.github.io/armi/developer/tooling.html#document-it) is still up-to-date in the `doc` folder.
- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The dependencies are still up-to-date in `pyproject.toml`.

PR #2315, opened 2025-10-02
Refactor some testing utils to use testing module

### LINKED ISSUE #2313: Move getEmptyHexReactor and getEmptyCartesianReactor to testing
I think `getEmptyHexReactor` and `getEmptyCartesianReactor` are useful enough they should be moved from `armi.tests` to `armi.testing`. Currently though, this would cause an import loop. So some care will have to be taken.

Also, maybe these should move into `armi.testing`:

* `from armi.reactor.tests.test_blocks import loadTestBlock`
* `from armi.reactor.tests.test_blocks import buildSimpleFuelBlock`
* `from armi.reactor.tests.test_assemblies import buildTestAssemblies`

john-science:

> NOTE: This PR will be in DRAFT until another PR is merged: https://github.com/terrapower/armi/pull/2314

I think some of our testing tools should be raised to the level of the `testing` module.

close #2313

## SCR Information

Change Type: trivial

One-Sentence Description: Refactor some testing utils to use testing module

---

## Checklist


armi/testing/__init__.py
```diff
@@ -26,45 +25,43 @@
...
 
-from armi import operators, runLog, settings
-from armi.reactor import reactors
-from armi.tests import ARMI_RUN_PATH, TEST_ROOT
+from armi import runLog
+from armi.reactor import geometry, grids, reactors
 
-_THIS_DIR = os.path.dirname(__file__)
+TEST_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "tests")
+TESTING_ROOT = os.path.dirname(os.path.abspath(__file__))
```
opotowsky, line 32: I hate to be the person that harps on naming esp in a testing file, but can you explain why we are making both a `TEST_ROOT` and `TESTING_ROOT` in `testing/__init__.py` and `test/__init__.py`? I think the ecosystem commonly imports `TEST_ROOT` from the latter file, but I think this might contribute to some confusion.

john-science: It's duplication, for sure. I did it to avoid a circular import.

There are more `armi/tests/` tools I would like to move to `armi/testing/`, but it causes a bunch of circular imports. 

It's going to be a process.

opotowsky: AH ok! I'm cool with that as long as the end goal is more clarity here.

PR #2317, opened 2025-10-06
Simplify thermal scattering

### LINKED ISSUE #1804: Better way to subclass thermal scattering labels for plugins
To use thermal scattering data in the ARMI [OpenMC plugin](https://github.com/terrapower/armicontrib-openmc), the labels for the thermal scattering data in OpenMC's syntax need to be generated somewhere. I recently submitted pull request #1792 in which I generate these labels with ARMI's existing ENDF and ACE label generators. However, @jakehader made a good point that this involves introducing some semi-hardcoded names into the framework and it's safer to keep these labels with the plugin that uses them.

Implementing these labels directly in the plugin misses out on many of the features of the ARMI `ThermalScattering` object. The best long-term solution from the plugin side would be a way to store the labels themselves in the plugin but have them behave like the existing ENDF and ACE labels (present in the material library in the `ThermalScattering` objects in `armi.material.thermalScatteringLaws` at import or configure time, accessible via `armi.nucDirectory.ThermalScattering.byNbAndCompound`, etc.). I don't know off the top of my head how that could work. Future physics plugins that make use of thermal scattering data will need something similar since they all use slightly difference labeling schemes.

For now, I have a temporary solution in the OpenMC plugin, so I think it makes sense to close the pull request and move the discussion here.

john-science:

Simplifying the API and architecture for the old thermal scattering laws. The code was FAR too complicated. And it looked like at some point it would describe the "laws" governing thermal scattering. But all the code has ever done the past ~decade is store two hard-code strings for labels for 19 compounds.

Also, as a side note, this closes #1804

## SCR Information

Change Type: features

One-Sentence Description: Simplifying the API and architecture for the old thermal scattering laws.

---

## Checklist

### DIFF
--- a/armi/nucDirectory/thermalScattering.py
+++ b/armi/nucDirectory/thermalScattering.py
@@ -71,7 +71,7 @@ BY_NAME_AND_COMPOUND = {
 }
 
 
-@dataclass
+@dataclass(frozen=True)
 class ThermalScatteringLabels:
     """Container for the labels for a particular nuclide/compound combination.
 


armi/materials/be9.py
```diff
@@ -19,15 +19,14 @@
...
-from armi.nucDirectory import nuclideBases as nb
 from armi.nucDirectory import thermalScattering as tsl
 from armi.utils.units import getTk
 
 
 class Be9(Material):
     """Beryllium."""
 
-    thermalScatteringLaws = (tsl.byNbAndCompound[nb.byName["BE"], tsl.BE_METAL],)
+    thermalScatteringLaws = (tsl.fromNameAndCompound("BE", tsl.BE_METAL),)
```
drewj-tp, line 29: I immediately like where this is going

armi/nucDirectory/thermalScattering.py
```diff
@@ -42,228 +30,91 @@ class attributes on :py:class:`~armi.materials.material.Material` subclasses to
...
+    ("O", D2O): (f"tsl-Oin{D2O}.endf", "o-d2o"),
+    ("O", UO2): ("tsl-OinUO2.endf", "o-uo2"),
+    ("SI", SIC): ("tsl-SIinSiC.endf", "si-sic"),
+    ("U", UN): ("tsl-UinUN.endf", "u-un"),
+    ("U", UO2): ("tsl-UinUO2.endf", "u-uo2"),
+    ("ZR", ZRH): ("tsl-ZRinZrH.endf", "zr-zrh"),
+}
+
+
+@dataclass
```
drewj-tp: I saw you had to remove `slots=True` for py3.8 support which is sad. I would recommend `frozen` as well to make these immutable
```suggestion
@dataclass(frozen=True)
```

john-science: Actually, my commit message is a typo. We only have to support 3.9.

Still, that means no "slots".

PR #2318, opened 2025-10-06
Move tutorial tests out of history tracker set up class

drewj-tp:
Move the execution of the `data_model.ipynb` from outside the `setUpClass` for the `TestHistoryTracker` tests and into an actual `test_*` method.

Now we don't have a test hiding inside the setup class for other tests. Looking at some [test timing reports](https://github.com/terrapower/armi/actions/runs/18232455948/job/51918891054#step:5:370), having the notebooks here leads to about 30 seconds of "set up" time for other, unrelated history tracker tests
```
============================= slowest 30 durations =============================
25.56s setup    armi/bookkeeping/tests/test_historyTracker.py::TestHistoryTracker::test_calcMGFluence
```

## SCR Information

Change Type: trivial

One-Sentence Description: Make a dedicated test for testing tutorial notebook, not hiding inside setup class for history tracker.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/tests/test_historyTracker.py
+++ b/armi/bookkeeping/tests/test_historyTracker.py
@@ -32,11 +32,13 @@ from armi.cases import case
 from armi.context import ROOT
 from armi.reactor import blocks, grids
 from armi.reactor.flags import Flags
-from armi.tests import ArmiTestHelper
+from armi.tests import TEST_ROOT, ArmiTestHelper
 from armi.utils.directoryChangers import TemporaryDirectoryChanger
 
 CASE_TITLE = "anl-afci-177"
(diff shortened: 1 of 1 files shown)


drewj-tp:
Okay looks like the notebook is responsible for setting some "physics data" on the reactor and then storing dumping it to db. These are necessary pre-conditions for the history tracker tests.

I think that's silly

john-science:
> Okay looks like the notebook is responsible for setting some "physics data" on the reactor and then storing dumping it to db. These are necessary pre-conditions for the history tracker tests.
> 
> I think that's silly

ARMI uses jupyter notebooks as examples in our docs. But, also, they serve as the closest thing to an "integration test" that ARMI can have (not being an executable program).

PR #2320, opened 2025-10-06
Providing interface method interactRestart for managing restarts

drewj-tp:

This spawned from https://github.com/terrapower/armi/discussions/2316

Interfaces can now provide the `interactRestart` method. 

The method can be used to solve the problem introduced in #2316: an interface could implement that hook to set `Core.lib` with the correct cross section library. Note that since no interface handles this behavior, no such method is implemented here.

There was some minor refactoring to accomplish this. The restart behavior was mostly hidden inside the `MainInterface.interactBOL`. Now it's handled by `Operator.interactAllRestart` which is engaged between the `interactAllInit` and `interactAllBOL` hooks. `MainInterface.interactBOL` existed to
-  jump the interface stack, grab the database interface, and do restarting behavior first,
- if necessary, call `interactAllEOC` for restarting at the start of one cycle

This is split out into `Operator.interactAllRestart` now.

1. Call the database interface to load from the previous time point using `DatabaseInterface.prepRestartRun`
2. Iterate through all remaining interfaces and call their `interactRestart` method

## SCR Information

Change Type: features

One-Sentence Description: Provide interactRestart interface method for managing restarts

One-line Impact on Requirements: R_ARMI_SNAPSHOT_RESTART implementation is modified to use this new behavior

---

## Checklist

### DIFF
--- a/armi/interfaces.py
+++ b/armi/interfaces.py
@@ -25,7 +25,7 @@ armi.plugins : Register various interfaces
 """
 
 import copy
-from typing import Dict, List, NamedTuple, Union
+from typing import Dict, List, NamedTuple, Tuple, Union
 
 import numpy as np
 from numpy.linalg import norm
@@ -498,6 +498,25 @@ class Interface:
     def interactDistributeState(self):
         """Called after this interface is copied to a different (non-primary) MPI node."""
         pass
+
+    def interactRestart(self, startNode: Tuple[int, int], previousNode: Tuple[int, int]):
+        """Perform any actions prior to simulating a restart.
+
+        Interfaces may want to restore some state that would have existed at the start of
+        ``startNode`` prior to calling :meth:`interactBOL` for the desired start point.
+        The database interface will be used prior to any interfaces calling this method,
+        so you can assume the reactor state has been correctly loaded from the database
+        from the ``previousNode``. This helps ensure that interfaces restart at e.g.,
+        ``(cycle, node)=(4, 3)`` would see the same data compared to the nominal simulation
+        without a restart.
+
+        Parameters
+        ----------
+        startNode
+            Pair of ``(cycle, node)`` for the requested restart point.
+        previousNode
+            Pair of ``(cycle, node)`` for the time node immediately preceeding ``startNode``.
+        """
 
     def isRequestedDetailPoint(self, cycle=None, node=None):
         """


armi/operators/operator.py
```diff
@@ -360,6 +366,23 @@ def _mainOperate(self):
                 break
         self.interactAllEOL()
 
+    def _onRestart(self):
```
john-science: I understand this is a private method. But you have made it pretty important. Can we get a one-line docstring telling people what this is for?

armi/plugins.py
```diff
@@ -699,6 +700,22 @@ def getAxialExpansionChanger() -> type["AxialExpansionChanger"]:
 
         """
 
+    @staticmethod
+    @HOOKSPEC
+    def prepRestart(o: "Operator", startTime: tuple[int, int], previousTime: tuple[int, int]):
```
john-science: Can you update the documentation to include a quick one-sentence description?

We list the plugin hooks for developers here:

https://github.com/terrapower/armi/blob/2bdd2f37e930ea412a455338987a9d4bce33fb59/doc/developer/guide.rst?plain=1#L207-L229

drewj-tp: Hmm those are interface methods which are not the same as the plugin hooks. Most of the docs I can find on how to build a plugin point devs to the module docs for `armi.plugins`, which includes docstrings for these `HOOKSPEC` methods - https://terrapower.github.io/armi/.apidocs/armi.plugins.html#armi.plugins.ArmiPlugin.defineBlockTypes

armi/operators/operator.py
drewj-tp: 43988ce1

armi/plugins.py
john-science: Oh, I didn't even notice you made your change to `Plugin`.  I guess I just assumed it fit better in `Interface` and _assumed_ it was there.

That's interesting. This didn't turn out how I expected. Sorry, let me look at this for a hot minute.

drewj-tp: Revised to be an interface method now

armi/bookkeeping/tests/test_historyTracker.py
```diff
@@ -93,11 +93,12 @@ def setUp(self):
...
 
-        self.o.getInterface("main").interactBOL()
-
         dbi = self.o.getInterface("database")
-        # Get to the database state at the end of stack of time node 1.
-        # The end of the interface stack is when history tracker tends to run.
+        # Make sure we have a database to use
+        dbi.initDB()
+        # Load the previous time point and merge histories to align with the restart point
+        dbi.prepRestartRun()
```
john-science, line 100: Sorry, I'm not sure I understand this change. Is it necessary for your PR? Previously, the `interactBOL` loaded a bunch of stuff into the Reactor state so we could test the history tracker. I can see the tests below still pass, but I'm not sure what i happening here.

drewj-tp: Yep. Before, `MainInterface.interactBOL` (now removed) called these things directly

https://github.com/terrapower/armi/blob/df71b0e807b0103f264a997a5b12be7fbc983512/armi/bookkeeping/mainInterface.py#L61-L105

Now that `MainInterface.interactBOL` does not do these things anymore, we can call the things we explicitly want to do

john-science: Okay. Thanks!

PR #2321, opened 2025-10-06
Provide Core.hasLib to check if there is a cross section library

drewj-tp:

In working on #2320 // #2316 I came across a weird pattern. If I want to check if the core currently has a cross section library, I could write `if self.r.core.lib is None`. But the `Core.lib` property will attempt to load a library if there is not one. This could have some unintentional side effects (and some have been handled e.g., #2173)

It's also not _great_ to encourage people to access what should be a protected attribute and check `if self.r.core._lib is None`. So, now `Core.hasLib() -> bool` will do this for you.

## SCR Information

Change Type: features 

One-Sentence Description: Provide Core.hasLib to check if there is a cross section library

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/latticePhysics/latticePhysicsInterface.py
+++ b/armi/physics/neutronics/latticePhysics/latticePhysicsInterface.py
@@ -383,7 +383,7 @@ class LatticePhysicsInterface(interfaces.Interface):
         if executeXSGen and not idsChangedBurnup:
             executeXSGen = False
 
-        if self.r.core._lib is not None:
+        if self.r.core.hasLib():
             # justification=r.core.lib property can raise exception or load pre-generated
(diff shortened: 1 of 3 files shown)


john-science (approved):

drewj-tp:
> @drewj-tp If I'm being picky, I have never liked "lib" as a name here. Surely it should be "xsLib" or something else that mentions cross sections, right?
> 
> I suppose that would require downstream PRs though. And I think a certain downstream team is code frozen right now.
> 
> Oh well. One day.

If you put up a ticket to rename it, we can hash it out there! I think a lot of things consider themselves to be data libraries attached to the thing you're simulating. So, some disambiguation would help there

I don't have as strong of opinions though

PR #2326, opened 2025-10-20
Revert "Moving some testing utils to testing module"

john-science:
Reverts terrapower/armi#2315


opotowsky:
We will attempt to not need this PR, but we will keep this in our back pocket for now

opotowsky:
#2331 usurps this one

PR #2334, opened 2025-10-21
Refactoring loadTestReactor to pickle multiple test reactors

### LINKED ISSUE #1636: Does the test reactor need to be cached/pickled?
`loadTestReactor` pickles the reactor. This caused a big bug in some downstream tests. Just wondering if this functionality is needed.

### LINKED ISSUE #2333: `isPickledReactor` check requests
## Request # 1: Make the check more robust

A little history:

#2315 broke the `isPickledReactor` check because even though the file path was valid, the STRINGS did not match

#2331 Fixed the fact that the strings didn't match. In it, I attempted to make the check a little more robust via wrapping the file paths being compared in `os.path.realpath()`, which actually broke the check (so I reverted it). 

I think we need to consider this more thoughtfully, and make the check more robust. 

## Request # 2: Make reactor pickling more flexible

Right now the only reactors that get pickled are when `armiRun.yaml` is loaded. Many downstream tests use the same test reactor based off of but not exactly `armiRun`. So can we pickle those and potentially speed up those tests?

john-science:

Essentially, when `loadTestReactor()` was first written, there was only ONE test reactor in ARMI. But now there are all kinds of them in ARMI and in downstream ARMI projects.

This PR helps cache/pickle all such test reactors. Though it provides a method to not do that, say for instance if you know that a certain test reactor will only ever be used in one test. You can save the RAM by not pickling it.

close #2333
close #1636

## SCR Information

Change Type: trivial

One-Sentence Description: Refactoring loadTestReactor to pickle multiple test reactors
One-Sentence Rationale: This substantially decreases unit testing time.

---

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -11,8 +11,8 @@
 <!-- Change Type: trivial -->
 <!-- Change Type: docs -->
 
-<!-- MANDATORY: Describe the change in one sentence -->
-One-Sentence Description: TBD
+<!-- MANDATORY: Describe why this change is needed, in one sentence -->
+One-Sentence Rationale: TBD
 
 <!-- MANDATORY: Describe any impact on the requirements, all on one line -->
 One-line Impact on Requirements: NA
--- a/doc/.static/automateScr.py
+++ b/doc/.static/automateScr.py
@@ -133,7 +133,7 @@ def _buildScrLine(prNum: str):
         scrType = "trivial"
 
     # grab one-line description
-    desc = _findOneLineData(lines, prNum, "One-Sentence Description:")
+    desc = _findOneLineData(lines, prNum, "One-Sentence Rationale:")
 
     # grab impact on requirements
     impact = _findOneLineData(lines, prNum, "One-line Impact on Requirements:")
--- a/doc/qa_docs/scr/0.6.rst
+++ b/doc/qa_docs/scr/0.6.rst
@@ -3,6 +3,63 @@ Release Notes for ARMI 0.6
 
 Here you will find the release notes for previous ARMI releases.
 
+ARMI v0.6.1
+-----------
+Release Date: 2025-11-05
+
+This was a minor release. While a lot of technical debt was addressed, no major feature work was done. There were some minor bugs fixed, but again nothing worth a release. This is being tagged as a release because the API is stable and this commit was tested heavily downstream and is trustworthy.
+
+Code Changes, Features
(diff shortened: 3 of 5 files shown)


opotowsky (commented):
Oh this is great! I especially love that pickle is now being spelled properly

We know this gives a speedup, but we need to monitor memory too since we are pickling so many reactors (depending on the repo). I will run testing tomorrow for memory use.

armi/testing/__init__.py
```diff
@@ -62,37 +65,35 @@ def loadTestReactor(inputFilePath=TEST_ROOT, customSettings=None, inputFileName=
...
-    fName = os.path.join(inputFilePath, inputFileName)
+    global _TEST_REACTORS
+    fName = os.path.abspath(os.path.join(inputFilePath, inputFileName))
     customSettings = customSettings or {}
-    isPickeledReactor = fName == ARMI_RUN_PATH and customSettings == {}
+    reactorHash = hash(fName + str(customSettings))
 
-    if isPickeledReactor and _TEST_REACTOR:
+    if useCache and reactorHash in _TEST_REACTORS:
         # return test reactor only if no custom settings are needed.
```
opotowsky: I think this comment here no longer applies

armi/testing/__init__.py
```diff
@@ -49,6 +49,9 @@ def loadTestReactor(inputFilePath=TEST_ROOT, customSettings=None, inputFileName=
         given in customSettings for that key.
     inputFileName : str, default="armiRun.yaml"
         Name of the input file to run.
+    useCache : bool, default=True
+        Do not bother pickling this reactor or looking for a pickled reactor. (This is useful for reactors where you are
```
opotowsky: I think this is now defaulting to true so it's the opposite: DO bother to pickle the reactor, and set to false for reactors that are only used once.

armi/testing/__init__.py
john-science:
```suggestion
        # return test reactor from cache
```

armi/testing/__init__.py
john-science:
```suggestion
        Look for a copy of this Reactor in the cache, if not in the cache, put it there. (Set to False when you are
```

john-science: Agreed. Done.

armi/testing/__init__.py
john-science: Agreed. Done.

PR #2365, opened 2025-11-14
Improving and correcting the ARMI version semantics

john-science:

Improving and correcting the ARMI version semantics.

## SCR Information

Change Type: docs

One-Sentence Rationale: It has been noted that the version semantics in the ARMI docs did not match the reality of the ARMI releases.

---

## Checklist

### DIFF
--- a/armi/nucDirectory/nuclideBases.py
+++ b/armi/nucDirectory/nuclideBases.py
@@ -157,19 +157,19 @@ class NuclideInterface:
 
     def getMcc2Id(self):
         """Return the MC2-2 nuclide identification label based on the ENDF/B-V.2 cross section library."""
-        raise NotImplementedError
+        return NotImplementedError
 
     def getMcc3Id(self):
         """Return the MC2-3 nuclide identification label based on the ENDF/B-VII.1 cross section library."""
-        raise NotImplementedError
+        return NotImplementedError
 
     def getMcc3IdEndfbVII0(self):
         """Return the MC2-3 nuclide identification label based on the ENDF/B-VII.0 cross section library."""
-        raise NotImplementedError
+        return NotImplementedError
 
     def getMcc3IdEndfbVII1(self):
         """Return the MC2-3 nuclide identification label based on the ENDF/B-VII.1 cross section library."""
-        raise NotImplementedError
+        return NotImplementedError
 
     def getSerpentId(self):
         """Get the Serpent nuclide identification label."""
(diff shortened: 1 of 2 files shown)


doc/developer/tooling.rst
```diff
@@ -136,9 +136,9 @@ Releasing a New Version of ARMI
 We use the common ``major.minor.bump`` version scheme where a version string might look like
 ``0.1.7``, ``1.0.0``, or ``12.3.123``. Each number has a specific meaning:
 
-* ``major`` - Revved for DOE/NRC-sanctioned release or at the end of a long development cycle.
-* ``minor`` - Revved when we decide the code or our API has reached a stable point.
-* ``bump`` - Revved every time we modify the API, and any other time we want.
+* ``major`` - Revved for major milestones of the ARMI project.
+* ``minor`` - Revved for the usual release, real feature work completed.
+* ``bump`` - Revved for very small releases, but still well-tested with a stable API.
```
drewj-tp, line 141: I don't know how to interpret these. Both from a developer (if I was to make a release, what should the next number be) nor as a user (if I saw a new release, what information does the numbering tell me)

By this logic, 0.6.1 contains "real feature work" (thermal scattering, nuclide bases class, mpi distribution of lattice physics) compared to 0.6.0.

idk, version numbers mean whatever people say they mean. People like [semantic versioning](https://semver.org/) b/c there is meaningful information for users conveyed just by looking at the change in the version number.

- "I was on version 1.15.4 and no 1.15.5 is out. That should be fine to upgrade, no problem"
- "Now 1.16.0 is out. Maybe should peep the changelog but should be stable for my purposes"
- "Oh 2.0.0 is out. Definitely need to read the changelog and do testing before upgrading"

This feels like we want to follow semantic versioning, but that doesn't align with our history. Is that the intent going forward?

drewj-tp (approved):
I'm approving because these changes convey more information than before and are "better"

I think, for now, until we hit armi 1.0.0, it would be sufficient to say

- major: bumped for major milestones
- minor: bumped whenever

and just progress 0.6.0 -> 0.7.0 -> 0.8.0

These are my thoughts. But I'm still approving the PR. Happy to talk more if you'd like.

doc/developer/tooling.rst
opotowsky: Since the API isn't stable, I don't think we should use the typical approach (otherwise we'd be major revving every release).

PR #2366, opened 2025-11-14
Raising instead of returning errors

john-science:

There were some places in the repo where I found we were returning, rather than raising errors.

## SCR Information

Change Type: trivial

One-Sentence Rationale: The exceptions in Python are meant to be raised, not returned, in most cases.

---

## Checklist

### DIFF
--- a/armi/reactor/grids/locations.py
+++ b/armi/reactor/grids/locations.py
@@ -32,15 +32,6 @@ class LocationBase(ABC):
     A namedtuple-like object for storing location information.
 
     It's immutable (you can't set things after construction) and has names.
-
-    Notes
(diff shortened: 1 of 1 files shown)


armi/reactor/grids/locations.py
```diff
@@ -32,15 +32,6 @@ class LocationBase(ABC):
     A namedtuple-like object for storing location information.
 
     It's immutable (you can't set things after construction) and has names.
-
-    Notes
-    -----
-    This was originally a namedtuple but there was a somewhat unbelievable
```
opotowsky, line 38: HA more historical tales! nice catch

PR #2373, opened 2025-12-04
Fixing docs deploying across GitHub repos

john-science:

A [recent change](https://github.com/terrapower/armi/pull/2369/files#diff-dd8b4830ef963dd868caa4707cb38e9ddd98382738c573275f9453c4e5fdc373R4) to the location of certain test reactors has caused problems with the automated docs build on GitHub. This PR attempts to address that problem.

## SCR Information

Change Type: docs

One-Sentence Rationale: The docs need to build correctly on GitHub, not just locally.

---

## Checklist

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,0 +1,52 @@
+name: Coverage
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - 'doc/**'
+  pull_request:
+    paths-ignore:
+      - 'doc/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    # Deploying coverage to coveralls.io should not happen on forks
+    if: github.repository == 'terrapower/armi'
+    runs-on: ubuntu-24.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.11'
+      - name: Install ARMI and MPI
+        run: |
+          sudo apt-get -y install libopenmpi-dev
+          pip install -e .[memprof,mpi,test]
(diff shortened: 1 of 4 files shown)


drewj-tp:
Is this why https://terrapower.github.io/armi/ is giving a 404? I can still find the latest release version - https://terrapower.github.io/releases/armi/v0.6.1/index.html

armi/tests/tutorials/param_sweep.ipynb
```diff
@@ -48,11 +48,17 @@
...
+    "import os\n",
+    "\n",
     "from armi import cases, settings\n",
     "from armi.cases import suiteBuilder\n",
     "from armi.cases.inputModifiers import inputModifiers\n",
     "\n",
-    "cs = settings.Settings(\"../../testing/reactors/anl-afci-177/anl-afci-177.yaml\")\n",
+    "fPath = \"../../armi/testing/reactors/anl-afci-177/anl-afci-177.yaml\"\n",
+    "if not os.path.exists(fPath):\n",
+    "    fPath = \"../../testing/reactors/anl-afci-177/anl-afci-177.yaml\"\n",
```
keckler, line 59: Is this so that the old versions of the docs don't break?

If so, maybe you can add a comment to indicate that. It looks arbitrary, without that context.

john-science: It turns out these files are used for tests and for the docs build. The jupyter notebooks are run from two different places for these.

BUT, the only reason it worked before was accidentally both of those place were the same number of directories away (`../`) from where they were being run. So there were two paths before, but because of an accident with the relative paths, they looked the same.

Does that make sense?

doc/tutorials/param_sweep.nblink
```diff
@@ -1,9 +1,9 @@
 {
 	"path": "../../armi/tests/tutorials/param_sweep.ipynb",
```
opotowsky, line 2: Is this path correct too? Or does it need one `..` removed?

john-science: Yup, this is working. You can download the docs in the GitHub CI workflow (they are archived) and you will see the tutorials / param_sweep is working as intended:

https://github.com/terrapower/armi/actions/runs/19970538705?pr=2373

armi/tests/tutorials/param_sweep.ipynb
keckler: I see.

Well, again, a comment could be useful. Not blocking by any means.

PR #2385, opened 2025-12-12
Set materials.FAIL_ON_RANGE to True by default

### LINKED ISSUE #1077: Set `material.materials.FAIL_ON_RANGE` to `True`
The subject `material.materials` constant is used in the `checkTempRange` function to determine whether the program should fail, or throw a warning, if material properties are requested for temperatures outside their appropriate correlation bounds:

https://github.com/terrapower/armi/blob/42726110a72606b9a45e64238582a2841e889f2e/armi/materials/material.py#L568-L596

It is set to `False` here:

https://github.com/terrapower/armi/blob/42726110a72606b9a45e64238582a2841e889f2e/armi/materials/material.py#L41

This means that, if running the code as-is, the program will run to completion even if improper correlation values are queried. A warning is logged, but this can be easy to miss. The default behavior should be changed so that the program fails instead.

The easiest way would be to set `material.materials.FAIL_ON_RANGE` to `True`. Or, we could remove this constant and replace it with a case setting instead.

john-science:

We want to set `armi.material.materials.FAIL_ON_RANGE` to `True` by default. For the past ~14 years, there has been an uncomfortable situation where if someone called a material property (say Uranium density) outside a temperature range that where we know the values, the simulation would just continue. That meant ARMI simulations could continue even if they have invalid inputs.

We are finally changing the default value, and it feels great.

close #1077 

## SCR Information

Change Type: features

One-Sentence Rationale: A simulation that has found that it has invalid or insufficient material data should fail by default.

---

## Checklist

### DIFF
--- a/armi/materials/b4c.py
+++ b/armi/materials/b4c.py
@@ -34,7 +34,7 @@ class B4C(material.Material):
     DEFAULT_THEORETICAL_DENSITY_FRAC = 0.90
     enrichedNuclide = "B10"
     NATURAL_B10_NUM_FRAC = 0.199
-    propertyValidTemperature = {"linear expansion percent": ((25, 500), "C")}
+    propertyValidTemperature = {"linear expansion percent": ((25, 600), "C")}
 
     def __init__(self):
         self.b10NumFrac = self.NATURAL_B10_NUM_FRAC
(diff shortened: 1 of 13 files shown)


john-science:
* @keckler / @aaronjamesreynolds / @sombrereau / @alexhjames 

This could break your codebases. I argue this is a good thing. But if that's a problem for you, holler and I can add one line to your repo (or even just `conftest.py`) to recover the old ARMI functionality:

```python
armi.material.materials.FAIL_ON_RANGE = False
```

I will happily make such PRs if you are negatively affected, or just hate me for this in some way.

opotowsky (commented):
Code review is good to go. I know we are waiting on the ecosystem to be able to merge, but I figured I should comment that i reviewed the PR!

Issue #2384, opened 2025-12-11
Users too frequently have to make an ARMI PR to touch Settings and Parameters


john-science:
@keckler noted that ARMI developers too frequently have to move PRs into ARMI to deal with `Setting`s and `Parameter`s defined in ARMI.

That is a system design problem. It seems to me the only solution here is to make sure ARMI includes fewer mandatory `Setting`s and `Parameter`s that we force on every downstream ARMI application. As such, one important ticket to help with this will be to start making `App` subclasses for different usages, as in [this ticket](https://github.com/terrapower/armi/issues/2382).

Other than that, I think the order-of-operations here goes something like this:

* [ ] Review all `Parameter`s in ARMI, and remove anything that appears unused in the world.
    * https://github.com/terrapower/armi/pull/981
    * https://github.com/terrapower/armi/pull/986
    * https://github.com/terrapower/armi/pull/988
    * https://github.com/terrapower/armi/pull/989
    * https://github.com/terrapower/armi/pull/990
    * https://github.com/terrapower/armi/pull/997
    * https://github.com/terrapower/armi/pull/2130
    * https://github.com/terrapower/armi/pull/2132
    * https://github.com/terrapower/armi/pull/2134
    * https://github.com/terrapower/armi/pull/2136
* [ ] Review all `Setting`s in ARMI, and remove anything that appears unused in the world.
    * https://github.com/terrapower/armi/pull/1393
    * https://github.com/terrapower/armi/pull/1758
    * https://github.com/terrapower/armi/pull/2127
* [ ] Review all `Parameter`s in ARMI, if they are only used by one (internal TerraPower) project, move it out of ARMI and into that project.
* [ ] Review all `Settings`s in ARMI, if they are only used by one (internal TerraPower) project, move it out of ARMI and into that project.

**NOTE**: Even if I can't find anyone using a `keff` `Parameter`, some `Setting`s and `Parameter`s are useful to a wide nuclear reactor audience, and we will just keep them on general principles.

john-science:
# Settings Review

## Unused

* ✅ `automaticVariableMesh`
* ✅ `assemPowSummary` - defaults to `False`, and I can't find any history of anyone ever setting it to `True`
* ✅ `operatorLocation`
* ✅ `thKernel`
* ✅ `timelineInclusionCutoff` - one plot in ReportInterface


## Only used in one place downstream

* ✅ `doTh`
* ✅ `circularRingOrder` - equal-ibrium
[...]

PR #2387, opened 2025-12-13
Removing Five Unused Settings

john-science:

I have asked my power users and we believe the following `Setting`s are unused and can safely be deleted:

* `automaticVariableMesh`
* `operatorLocation`
* `thKernel`
* `summarizePower`
* `timelineInclusionCutoff`

**NOTE**: I ran all the downstream unit and integration tests I could find, and they passed without problem.

Progress on #2384 

## SCR Information

Change Type: trivial

One-Sentence Rationale: This is the start of work to reduce the amount of fussing with Settings and Parameters in ARMI that our users need to do.

---

## Checklist

### DIFF
--- a/armi/settings/fwSettings/reportSettings.py
+++ b/armi/settings/fwSettings/reportSettings.py
@@ -14,7 +14,6 @@
 
 """Settings related to the report generation."""
 
-
 from armi.settings import setting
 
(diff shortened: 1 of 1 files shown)


opotowsky (approved):
Approving, since I just checked the downstream tests pass!

PR #2390, opened 2025-12-17
Ensuring that SCRs will build correctly during main branch pushes

### LINKED ISSUE #2397: Are the SCR docs broken on main?
During this last release push @opotowsky found that a couple of PRs were missing from the SCR in the PDF docs that were produced from the `main` branch.

BUT when I looked at the PDF produced from the PR that _caused_ that `main` branch commit, it was perfect.

My conclusion here is that there is a problem in the file:

> doc/.static/automateScr.py

And the reasons we haven't found this problem before are probably something like:

a. It is only a small discrepancy, like 2 of 24 PRs were missing
b. Since it is just a small fraction of PRs / SCRs missing, this might be harder to detect at the begin of a release cycle, rather than the end.
c. This will be hard to reproduce or test, as it requires a `main` branch commit.

john-science:

This PR improves the ARMI docs in a few different ways:

* I believe the SCR was not building correctly on ARMI `main`, because the `PR_NUMBER` env variable was being set incorrectly in the `doc.yaml` file. This is why I had to pull PDFs from the PRs for this past ARMI release.
* There was a gallery example that broke when we made the `FAIL_ON_RANGE=True` change.
* Also, I refactored the `automateSCR.py` script to make it easier to run locally (I just made one of the arguments optional).

closes #2397

## SCR Information

Change Type: docs

One-Sentence Rationale: The SCR should, ideally, build from PRs and from the main branch.

---

## Checklist

### DIFF
--- a/doc/.static/automateScr.py
+++ b/doc/.static/automateScr.py
@@ -111,7 +111,6 @@ def _buildScrLine(prNum: str):
     str
         RST-formatted list item.
     """
-    print(prNum)
     txt = subprocess.check_output(["gh", "pr", "view", prNum]).decode("utf-8")
     lines = [ln.strip() for ln in txt.split("\n") if ln.strip()]
 
@@ -239,7 +238,6 @@ def buildScrListing(pastCommit: str, thisPrNum: int = -1):
 
     # 3. Build a list for each SCR
     data = {"docs": [], "features": [], "fixes": [], "trivial": []}
-    print(sorted(prNums))
     for prNum in sorted(prNums):
         if not isMainPR(prNum):
             continue


pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
     "pluggy>=1.2.0", # Central tool behind the ARMI Plugin system
     "pyDOE>=0.3.8", # We import a Latin-hypercube algorithm to explore a phase space
     "pyevtk>=1.2.0", # Handles binary VTK visualization files
-    "ruamel.yaml ; python_version >= '3.11.0'", # Our foundational YAML library
+    "ruamel.yaml<0.19.0 ; python_version >= '3.11.0'", # Our foundational YAML library
```
john-science, line 43: A new release of `ruamel.yaml` happened just before the New Year. Now all open ARMI PRs and branches need this line or the CI will fail.

First PR to merge in wins!

opotowsky (approved):
SWEET so excited you have this fixed. You have 3 print statements in the PR still, unsure if you want to keep those or if they were just there for troubleshooting. I'll leave that up to you to delete, but it's non-blocking to my approval.

doc/.static/automateScr.py
```diff
@@ -110,6 +111,7 @@ def _buildScrLine(prNum: str):
     str
         RST-formatted list item.
     """
+    print(prNum)
```
john-science:
```suggestion
```

doc/.static/automateScr.py
```diff
@@ -237,6 +239,7 @@ def buildScrListing(thisPrNum: int, pastCommit: str):
 
     # 3. Build a list for each SCR
     data = {"docs": [], "features": [], "fixes": [], "trivial": []}
+    print(sorted(prNums))
```
john-science:
```suggestion
```

PR #2399, opened 2025-12-19
New exception for mpi4py import

opotowsky:

ARMI has a try/except for the from `mpi4py import MPI` that has historically worked for us downstream. The exception, `ImportError`, has captured all our needs up until now. 
 
Now (mpi4py 4.1.1 upgrade to go with the pkl5 change), that import has a different exception: `RuntimeError` (`RuntimeError: cannot load MPI library`). We are failing upon all invocations of armi as a result, unless we have an OMPI module loaded.

## SCR Information

Change Type: features

One-Sentence Rationale: Upgrading mpi4py caused a new error; this update handles that situation.

---

## Checklist

### DIFF
--- a/armi/context.py
+++ b/armi/context.py
@@ -111,7 +111,8 @@ except ImportError:
     # stick with defaults
     pass
 except RuntimeError:
-    # likely from MPI not being on system, this is OK for many armi invocations
+    # likely from MPI not being on system, this is OK for many ARMI invocations
+    # Note this exception was introduced upon upgrading to mpi4py 4.1.1
     pass
 
 if sys.platform.startswith("win"):


armi/context.py
```diff
@@ -110,6 +110,9 @@ def setMode(cls, mode):
 except ImportError:
     # stick with defaults
     pass
+except RuntimeError:
+    # likely from MPI not being on system, this is OK for many armi invocations
```
john-science:
```suggestion
    # likely from MPI not being on system, this is OK for many ARMI invocations
    # Note this exception was introduced upon upgrading to mpi4py 4.1.1
```

john-science: Also... Should we mention that this was new for mpi4py version whatever?

Your call.

Issue #2401, opened 2025-12-30
`cleanPath` behavior has troubling issues


opotowsky:
# Problem Description

Here's the code:
https://github.com/terrapower/armi/blob/8586f0021339b3c1dac23b58e700af3526c19111/armi/utils/pathTools.py#L189-L220

Note the variable `valid`. I assume this means "valid for deletion". The `if` statement on line 219 backs up this assumption of mine. If `valid = False`, you get the warning that the path cannot be deleted, and if `valid = True`, the path is deleted. 

But the checks that set `valid` to `True` are the opposite of what you'd expect! This caused me to lose a lot of time until I took a 3rd look at the code. 

## Issue 1

Look at the check on line 212. It is checking for paths in `DO_NOT_CLEAN_PATHS`:
https://github.com/terrapower/armi/blob/8586f0021339b3c1dac23b58e700af3526c19111/armi/utils/pathTools.py#L29-L37

These are actually being processed as `YES_CLEAN_THESE_PATHS`. And anything under `tests` _should_ be cleaned (these are always from `TempDirChanger` uses). But maybe some of these others we'd want to keep around, although I assume they are mostly irrelevant / out of date.

## Issue 2

The check on line 216 is problematic. It's checking if `APP_DATA` is in ANY parent of `path`. Who cares about `APP_DATA` at all? Except....it's a default for `activateLocalFastPath`:
https://github.com/terrapower/armi/blob/8586f0021339b3c1dac23b58e700af3526c19111/armi/context.py#L177-L184

But, like, it's only a part of the default fast path. OR it's not even fast path at all because much of the time it's set to the `cwd`:
https://github.com/terrapower/armi/blob/8586f0021339b3c1dac23b58e700af3526c19111/armi/context.py#L145. Either way, this has no bearing on the delete-ability of `path`.

Also, `APP_DATA` will always be in `parents` if they are on the same system. E.g., `parents` includes the file path `/`, and if `APP_DATA` is `/tmp/.armi` then it's in parents.

Anyway, I think I've made the case that we shouldn't care if `APP_DATA` is in any parent of `path`. Can we just remove the check? I don't think it's doing anything and maybe if nothing else could be dangerous?

# Solution

1. Obviously, clear up the code that is giving people the wrong idea. Update `DO_NOT_CLEAN_PATHS` and make sure it actually makes sense for modern ARMI uses. Make sure the code is processing it properly (right now the code is allowing deletions of anything in this list).
[...]

PR #2402, opened 2025-12-30
Improving `cleanPath` and related tooling

opotowsky:

_Update: I'm marking this as a feature request instead of a bug for the SCR because although I'm calling it a bug, it wasn't hurting anyone. It's just super unclear behavior we are now correcting._ 

See #2401 for a full description of the non-ideal behavior.

1. The `DO_NOT_CLEAN_PATHS` check is removed entirely as it is not needed. It was doing the opposite of what was indicated--it was greenlighting those deletions. We don't need any greenlit deletions other than (2) and (3) below.
2. The `APP_DATA` check is replaced with a check that the path for deletion is under `_FAST_PATH`. `_FAST_PATH` is amended often (e.g. if the code changes directory) and is not guaranteed to be under `APP_DATA`, but `_FAST_PATH` and therefore anything in it is always safe to delete. 
3. The `TemporaryDirectoryChanger` root is not guaranteed to be under `_FAST_PATH`, so an argument is added to indicate to `cleanPath` the dir slated for deletion comes from a `tempDir` instance (and this was also applied to output caches as well).
4. This work exposed some confusion with `cleanTempDirs`, so we made a ticket for that: #2408. I'm resolving that here because it was found during the course of this work. 

## SCR Information

Change Type: features

One-Sentence Rationale: `cleanPath` is not behaving like the variable names would indicate, so issues with naming were corrected in `cleanPath` as well as in `cleanTempDirs`.

---

## Checklist

### DIFF
--- a/armi/utils/pathTools.py
+++ b/armi/utils/pathTools.py
@@ -192,13 +192,12 @@ def cleanPath(path, mpiRank=0, forceClean=False):
     if not os.path.exists(path):
         return True
 
-    # Any forceClean can be deleted
     if forceClean:
+        # Any forceClean can be deleted
         valid = True
-
-    # If the path slated for deletion is a subdirectory of _FAST_PATH, then cool, delete.
-    # _FAST_PATH itself gets deleted on program exit.
-    if pathlib.Path(path).is_relative_to(pathlib.Path(context.getFastPath())):
+    elif pathlib.Path(path).is_relative_to(pathlib.Path(context.getFastPath())):
+        # If the path slated for deletion is a subdirectory of _FAST_PATH, then cool, delete.
+        # _FAST_PATH itself gets deleted on program exit.
         valid = True
 
     if not valid:


armi/utils/pathTools.py
```diff
@@ -209,25 +204,33 @@ def cleanPath(path, mpiRank=0):
...
-            valid = True
+    # Any tempDir can be deleted
+    if tempDir:
+        valid = True
 
-    if pathlib.Path(context.APP_DATA) in pathlib.Path(path).parents:
+    # If the path slated for deletion is a subdirectory of _FAST_PATH, then cool, delete.
+    # _FAST_PATH itself gets deleted on program exit.
+    if path.is_relative_to(context.getFastPath()):
         valid = True
```
opotowsky: We have only 2 scenarios that allow us to delete a path (temp dir, and underneath FAST_PATH), so I think this function is pretty safe. 

I will need to do a lot of testing to make sure this doesn't mess up anything in the downstream code.

armi/utils/pathTools.py
```diff
@@ -26,14 +26,14 @@
 from armi import context, runLog
 from armi.utils import safeCopy
 
+# TODO: This needs to be updated, since I don't think any of these are in common use anymore
```
john-science: Just delete this list then.

opotowsky: Oh sure I can do that 😈

opotowsky:
@john-science  this is a feature addition that exposed a bug, so I'm treating it as a bug ticket. I could separate the feature addition  into a different PR if you prefer for SCR clarity. The feature is just this:
image

Also I see a bunch of CI failed. I'll look into that later today....bodes well for my downstream testing 🙃

armi/utils/pathTools.py
```diff
@@ -209,25 +204,33 @@ def cleanPath(path, mpiRank=0):
...
-        if validPath in path.lower():
-            valid = True
+    # Any tempDir can be deleted
+    if tempDir:
+        valid = True
 
-    if pathlib.Path(context.APP_DATA) in pathlib.Path(path).parents:
+    # If the path slated for deletion is a subdirectory of _FAST_PATH, then cool, delete.
+    # _FAST_PATH itself gets deleted on program exit.
+    if path.is_relative_to(context.getFastPath()):
```
opotowsky:
```suggestion
    if pathlib.Path(path).is_relative_to(pathlib.Path(context.getFastPath())):
```

pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
     "pluggy>=1.2.0", # Central tool behind the ARMI Plugin system
     "pyDOE>=0.3.8", # We import a Latin-hypercube algorithm to explore a phase space
     "pyevtk>=1.2.0", # Handles binary VTK visualization files
-    "ruamel.yaml ; python_version >= '3.11.0'", # Our foundational YAML library
+    "ruamel.yaml<0.19.0; python_version >= '3.11.0'", # Our foundational YAML library
```
opotowsky: we got one of those "someone released a new package and now everything broke" situations. 

I didn't look into fixes yet, just testing that this is working first

armi/utils/outputCache.py
```diff
@@ -176,7 +176,8 @@ def deleteCache(cachedFolder):
     if "cache" not in str(cachedFolder).lower():
         raise RuntimeError("Cache location must contain keyword: `cache`.")
 
-    cleanPath(cachedFolder)
+    # We can consider caches temporary directories, since they are not important containers for simulation results.
+    cleanPath(cachedFolder, tempDir=True)
```
opotowsky: this makes me think we could call it something else, like safeToDelete

but of course anyone using cleanPath thinks their path is safe to delete!

armi/utils/pathTools.py
opotowsky: 😈 d11bc73a65e4357c362f1659f319b122968f5876

pyproject.toml
john-science: Uh oh.

`ruamel.yaml` is used widely though ARMI, and another downstream materials repo we know. 

This would be a really unfortunate restriction> BUT, I guess, if you're only changing the version of ruamel.yaml for old versions of Python, then this isn't a game stopped.

armi/utils/pathTools.py
```diff
@@ -186,19 +176,12 @@ def moduleAndAttributeExist(pathAttr):
...
-
-    !!! Be careful with this !!! It can delete the entire cluster.
-
-    We add copious os.path.exists checks in case an MPI set of things is trying to delete everything
-    at the same time. Always check filenames for some special flag when calling this, especially
-    with full permissions on the cluster. You could accidentally delete everyone's work with one
-    misplaced line! This doesn't ask questions.
+def cleanPath(path, mpiRank=0, tempDir=False):
+    """Recursively delete a path. This function checks for a few cases we know to be OK to delete: (1) Any
+    `TemporaryDirectoryChanger` instance and (2) anything under the ARMI `_FAST_PATH`.
```
john-science: This is not what cleanPath originally was meant for. It's worth a discussion.

pyproject.toml
opotowsky: actually this is for >= 3.11

So it's pretty ugly. :-(

opotowsky: this was merged into main already, and we will have to address it in the future. I'll make a ticket and resolve this convo about it

armi/utils/pathTools.py
```diff
@@ -209,25 +192,27 @@ def cleanPath(path, mpiRank=0):
...
-        if validPath in path.lower():
-            valid = True
+    # Any tempDir can be deleted
+    if tempDir:
+        valid = True
 
-    if pathlib.Path(context.APP_DATA) in pathlib.Path(path).parents:
+    # If the path slated for deletion is a subdirectory of _FAST_PATH, then cool, delete.
+    # _FAST_PATH itself gets deleted on program exit.
+    if pathlib.Path(path).is_relative_to(pathlib.Path(context.getFastPath())):
```
opotowsky: TODO double check this does what I think it does

armi/utils/pathTools.py
opotowsky: @john-science will check on if this can be deleted. Could be cruft from the old days.

Does shutil only return once it's complete? If yes, can delete

armi/utils/pathTools.py
opotowsky: yes, it does: 

Image

armi/utils/pathTools.py
john-science: I can confirm:

* This is cruft from the ancient past.
* Their docs do not say that `shutil.rmtree()` is blocking. But digging around the code for a while convinces me it is.

armi/utils/outputCache.py
john-science: Honest Question. I don't see why `tempDir=True` is necessary here. It wasn't before.

Is this needed in vanilla ARMI, or is this to support your container work?

armi/utils/pathTools.py
```diff
@@ -186,19 +176,12 @@ def moduleAndAttributeExist(pathAttr):
...
-def cleanPath(path, mpiRank=0):
-    """Recursively delete a path.
-
-    !!! Be careful with this !!! It can delete the entire cluster.
-
-    We add copious os.path.exists checks in case an MPI set of things is trying to delete everything
-    at the same time. Always check filenames for some special flag when calling this, especially
-    with full permissions on the cluster. You could accidentally delete everyone's work with one
-    misplaced line! This doesn't ask questions.
+def cleanPath(path, mpiRank=0, tempDir=False):
```
john-science: I'm 50/50 on this. Should this new parameter be called `tempDir`, or something like `force`?

armi/utils/outputCache.py
opotowsky: output caches are often in a non-`_FAST_PATH` location, so they will fail to be deleted with the new `cleanPath` updates. 

But they are certainly always safe to clean up. So I'm thinking we can justify passing in the `tempDir` arg here (and or call that arg something else to be more inclusive of the 2 use cases we now have)

armi/utils/pathTools.py
opotowsky: Oh yea, I'm very willing to go over ideas here (see comment reply here https://github.com/terrapower/armi/pull/2402#discussion_r2670396419)

Really, as reviewer you can tell me what to do here. 

`force`
`forceDelete`
`forceClean`
`allowDelete`
`yesIHaveAPermit`
`AUTHORIZED`
`clearedForDeletion`
`blessedBeTheDeletion`
`outputCacheOrTempDirDeletionONLY`
`hallPassForDestruction`

ok wow I need to stop

john-science: I would go with `force` or `forceClean`, if I was picking from your list. #shrug

opotowsky: i like `forceClean` between the two, just cause I have a fear reaction with `force` alone 😅

armi/utils/pathTools.py
opotowsky: So from google, it looks like this is here to support behavior we see on Windows. Since ARMI supports Windows, maybe we should just keep it. i can just add a comment to this effect? 

An alternative (doesn't need to be decided with this PR) is removing the deletion check from `cleanPath`. Failed fast path or temp dir deletions aren't catastrophic, and one can always run `armi clean-temps` occasionally to clear out temp dirs on a local machine running armi.

What do you think?

armi/utils/pathTools.py
opotowsky: https://github.com/terrapower/armi/pull/2402/commits/5febad9f9aed0d3e6465f524820ce211b19909e2

armi/utils/pathTools.py
john-science: Eww. If this is still a problem on Windows then I guess we'll keep it.

armi/tests/test_mpiFeatures.py
```diff
@@ -297,50 +297,50 @@ class MpiPathToolsTests(unittest.TestCase):
     def test_cleanPathMpi(self):
         """Simple tests of cleanPath(), in the MPI scenario."""
         with TemporaryDirectoryChanger():
-            # TEST 0: File is not safe to delete, due to name pathing
+            # TEST 0: File is not safe to delete, due to due not being a temp dir or under FAST_PATH
```
john-science:
```suggestion
            # TEST 0: File is not safe to delete, due to not being a temp dir or under FAST_PATH
```

armi/tests/test_mpiFeatures.py
```diff
@@ -297,50 +297,50 @@ class MpiPathToolsTests(unittest.TestCase):
...
-            # delete the directory and test
             self.assertTrue(os.path.exists(dir3))
-            self.assertTrue(os.path.exists(os.path.join(dir3, "file1.txt")))
-            self.assertTrue(os.path.exists(os.path.join(dir3, "file2.txt")))
-            pathTools.cleanPath(dir3, mpiRank=context.MPI_RANK)
+            pathTools.cleanPath(dir3, mpiRank=context.MPI_RANK, forceClean=True)
             MPI_COMM.barrier()
             self.assertFalse(os.path.exists(dir3))
 
+            # TEST 3: Delete a directory with two files inside with forceClean=True
```
john-science, line 332:
```suggestion
            # TEST 3: Delete a directory with two files inside the FAST_PATH with forceClean=True
```

john-science: Is this what you meant?

armi/utils/tests/test_pathTools.py
```diff
@@ -104,46 +104,46 @@ def test_moduleAndAttributeExist(self):
...
-
-            # delete the directory and test
             self.assertTrue(os.path.exists(dir3))
-            self.assertTrue(os.path.exists(os.path.join(dir3, "file1.txt")))
-            self.assertTrue(os.path.exists(os.path.join(dir3, "file2.txt")))
-            pathTools.cleanPath(dir3, mpiRank=0)
+            pathTools.cleanPath(dir3, mpiRank=0, forceClean=True)
             self.assertFalse(os.path.exists(dir3))
 
+            # TEST 4: Delete a directory with two files inside with forceClean=True
```
john-science, line 135:
```suggestion
            # TEST 4: Delete a directory with two files inside the FAST_PATH with forceClean=True
```

john-science: I think this is what you mean?

armi/utils/pathTools.py
```diff
@@ -209,25 +192,27 @@ def cleanPath(path, mpiRank=0):
...
-        if validPath in path.lower():
-            valid = True
+    # Any forceClean can be deleted
+    if forceClean:
+        valid = True
 
-    if pathlib.Path(context.APP_DATA) in pathlib.Path(path).parents:
+    # If the path slated for deletion is a subdirectory of _FAST_PATH, then cool, delete.
+    # _FAST_PATH itself gets deleted on program exit.
+    if pathlib.Path(path).is_relative_to(pathlib.Path(context.getFastPath())):
```
john-science: It's a very minor performance change, hardly matters, but this could be an `elif` statement.

Your call.

armi/utils/pathTools.py
```diff
@@ -209,25 +192,27 @@ def cleanPath(path, mpiRank=0):
...
+    # Delete the file/directory from only one process
     if mpiRank == context.MPI_RANK:
         if os.path.exists(path) and os.path.isdir(path):
             shutil.rmtree(path)
         elif not os.path.isdir(path):
             # it's just a file. Delete it.
             os.remove(path)
 
-    # Potentially, wait for the deletion to finish.
+    # Deletions are not immediate, so wait for it to finish.
```
john-science:
```suggestion
    # Deletions may not be immediate on Windows, so wait for it to finish.
```

armi/tests/test_mpiFeatures.py
opotowsky: nope I think this dir isn't in the _FAST_PATH, so it shows that `forceClean` still works to delete it

armi/utils/tests/test_pathTools.py
opotowsky: Same as above

armi/utils/outputCache.py
```diff
@@ -176,7 +176,8 @@ def deleteCache(cachedFolder):
     if "cache" not in str(cachedFolder).lower():
         raise RuntimeError("Cache location must contain keyword: `cache`.")
 
-    cleanPath(cachedFolder)
+    # We can consider caches temporary directories, since they are not important containers for simulation results.
```
opotowsky:
```suggestion
    # Output caches need to pass in `forceClean` in order to greenlight the deletion.
```

armi/utils/pathTools.py
opotowsky: ah true! agreed it's prob not impactful but might as well do the right thing :-)

opotowsky: done! https://github.com/terrapower/armi/pull/2402/commits/116bd6b60cb37bc896cac9dc4322db8ddadb90e6

PR #2405, opened 2026-01-05
Remove `ruamel.yaml` dependency pin maximum

opotowsky:

We depend on `yamlize` for blueprints loading, which is a dead project. So when some underlying pieces of ruamel.yaml changed, our code broke. Right now we are choosing to drop CLoader (since ruamel.yaml has dropped clib and clibz as dependencies) and we are sidestepping the API issue by setting the `max_depth` attribute to `None`. This allows us to unpin the dependency and also drop the clib dependency, but we are still left with the dependence on a dead project. 

Note that this results in the unit tests running about 20 seconds slower. This is a hit we probably have to take for now since we need to move in the direction of using updated packages. We can add back in the `CLoader` code if this is unacceptable right now, but it will have to be addressed eventually.

## SCR Information

Change Type: features

One-Sentence Rationale: The ruamel.yaml project released a new version with updated API and changed dependencies that we need to conform to.

---

## Checklist

- [x] ~[Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify any new/changed code.~ Tests pass whereas before they failed.

### DIFF
--- a/armi/reactor/blueprints/__init__.py
+++ b/armi/reactor/blueprints/__init__.py
@@ -521,8 +521,7 @@ class Blueprints(yamlize.Object, metaclass=_BlueprintsPluginCollector):
         # AttributeError: 'RoundTripLoader' object has no attribute 'max_depth'
         # Setting that attribute to `None` solved the issue. However, it would be prudent to rework blueprints loading
         # to side step the issue entirely. This occurs because of the way `yamlize` works when it calls
-        # `get_single_node`. Reworking blueprints loading to drop yamlize (a dead project) is the only real solution
-        # here.
+        # `get_single_node`.
         RoundTripLoader.max_depth = None
         return super().load(stream, Loader=RoundTripLoader)
 


armi/reactor/blueprints/__init__.py
```diff
@@ -522,6 +522,7 @@ def load(cls, stream, roundTrip=False):
         the `CLoader` class is 10x faster, but doesn't allow for "round trip" (read-
         write) access to YAMLs; for that we have the `RoundTripLoader`.
         """
+        RoundTripLoader.max_depth = None
         loader = RoundTripLoader if roundTrip else CLoader
         return super().load(stream, Loader=loader)
```
opotowsky: we do not have a need for CLoader. we only ever use `load` with `typ="safe"` or `"typ=rt"`. speed isn't a concern for these 1 off yaml loading events...we aren't doing thousands of these

I think if we drop the CLoader thing and clean up this method, we sidestep the entire issue with the deprecation of `max_depth` attribute for RTLoader class

opotowsky: @john-science whatcha think about a mini cleanup job here, and then we can 1. unpin ruamel.yaml and 2. likely drop the `clib` dep, which appears to be a dep ruamel.yaml dropped

john-science: > for these 1 off yaml loading events...we aren't doing thousands of these

Aren't we doing exactly that, in the unit tests? In ARMI and in the big packages with lots of tests downstream?

Honest question. How big a hit are we talking?

>  likely drop the clib dep

I thought that was the goal here, but the pyproject.toml doesn't have that change yet. I would love to lose the depenendency though. For sure.

pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
     "pluggy>=1.2.0", # Central tool behind the ARMI Plugin system
     "pyDOE>=0.3.8", # We import a Latin-hypercube algorithm to explore a phase space
     "pyevtk>=1.2.0", # Handles binary VTK visualization files
-    "ruamel.yaml<0.19.0 ; python_version >= '3.11.0'", # Our foundational YAML library
+    "ruamel.yaml>=0.19.1 ; python_version >= '3.11.0'", # Our foundational YAML library
     "ruamel.yaml.clib ; python_version >= '3.11.0'", # C-based core of ruamel below
     "ruamel.yaml.clib<=0.2.7 ; python_version < '3.11.0'", # C-based core of ruamel below
```
john-science:
```suggestion
```

john-science: Is this the goal?

armi/reactor/blueprints/__init__.py
opotowsky: that's a great question. let me get this PR running with that change and we can compare runtime for ARMI CI as well as downstream.

pyproject.toml
opotowsky: Yea -- Let me spend a few minutes on this! I dropped th ball yesterday. Maybe It'll be all for nothing but won't know until we try

john-science: No rush. I'm just making sure I know the idea. I think the ticket maybe confuses me? But if this is the goal of the PR then great, sure, love it. I understand this.

armi/reactor/blueprints/__init__.py
```diff
@@ -522,6 +522,12 @@ def load(cls, stream, roundTrip=False):
         the `CLoader` class is 10x faster, but doesn't allow for "round trip" (read-
         write) access to YAMLs; for that we have the `RoundTripLoader`.
         """
+        # With the release of ruamel.yaml 0.19.1, we began getting the following error:
+        # AttributeError: 'RoundTripLoader' object has no attribute 'max_depth'
+        # Setting that attribute to `None` solved the issue. However, it would be prudent to rework blueprints loading
+        # to side step the issue entirely.
+        if roundTrip:
+            RoundTripLoader.max_depth = None
```
opotowsky:
```suggestion
        RoundTripLoader.max_depth = None
```

pyproject.toml
opotowsky: Ok this is likely no longer a goal of this PR and I made a new ticket with this goal in mind: #2449

armi/reactor/blueprints/__init__.py
opotowsky: This work is being moved to another ticket so I'll resolve this convo

armi/reactor/blueprints/__init__.py
```diff
@@ -522,6 +522,11 @@ def load(cls, stream, roundTrip=False):
         the `CLoader` class is 10x faster, but doesn't allow for "round trip" (read-
         write) access to YAMLs; for that we have the `RoundTripLoader`.
         """
+        # With the release of ruamel.yaml 0.19.1, we began getting the following error:
+        # AttributeError: 'RoundTripLoader' object has no attribute 'max_depth'
+        # Setting that attribute to `None` solved the issue. However, it would be prudent to rework blueprints loading
+        # to side step the issue entirely.
+        RoundTripLoader.max_depth = None
```
opotowsky, line 525: Ok this is super weird @john-science 

If I hide this under an `if roundTrip` (Cause that's the only time it's needed, RIGHT?!) tests still fail with `Roundtrip loader has no attribute max_depth` (See prior commits)

But if I leave it out in the open like this then the code works. So why, on non-round trip loads, would this be needed?

I am tempted to walk back on my separation of tickets and test the performance hit of getting rid of CLoader (dropping clib dep) rather than chasing this particular problem

opotowsky: This PR now removes all `CLoader` so this comment is moot. But it was still super weird and I think the underlying code still ended up using `RoundTripLoader` somewhere somehow even when `CLoader` was passed in. Since I'm not gonna spend my time updating yamlize, I'm not going to spend more time on this particular issue. We are using the default now since ruamel.yaml dropped clib/cliz support anyway.

armi/reactor/blueprints/__init__.py
```diff
@@ -516,14 +516,15 @@ def migrate(cls, inp: typing.TextIO):
...
-        """
-        loader = RoundTripLoader if roundTrip else CLoader
-        return super().load(stream, Loader=loader)
+        """This method is a wrapper around the `yamlize.Object.load()` method."""
+        # With the release of ruamel.yaml 0.19.1, we began getting the following error:
+        # AttributeError: 'RoundTripLoader' object has no attribute 'max_depth'
+        # Setting that attribute to `None` solved the issue. However, it would be prudent to rework blueprints loading
+        # to side step the issue entirely. This occurs because of the way `yamlize` works when it calls
+        # `get_single_node`. Reworking blueprints loading to drop yamlize (a dead project) is the only real solution
+        # here.
```
john-science:
```suggestion
        # `get_single_node`.
```

john-science: That last sentence sounds more like a ticket than a code comment. IMO.

opotowsky: Yea there is a ticket for it too :-) Happy to remove the comment if you want!

https://github.com/terrapower/armi/issues/2449

PR #2407, opened 2026-01-06
Removing unused Inconel material properties

john-science:

There are three redundant and unused material properties in the `Inconel` materials:

* `polyfitThermalConductivity`
* `polyfitLinearExpansionPercent`
* `polyfitHeatCapacity`

These are redundant with other properties / methods in those materials, just just don't have "polyFit" in the name. And after thorough testing, I don't see that anyone has used these in several years.

## SCR Information

Change Type: trivial

One-Sentence Rationale: Removing unused code is a best practice, but in this case we are simplifying upcoming materials work.

---

## Checklist

### DIFF
--- a/armi/materials/inconel600.py
+++ b/armi/materials/inconel600.py
@@ -19,7 +19,6 @@ this file for a fully worked example of an ARMI material. And this material has 
 contained in this file should not be used in production simulations.
 """
 
-
 from armi.materials.material import Material
 from armi.utils.units import getTc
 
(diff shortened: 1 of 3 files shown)


dpham-materials (approved):
I've reviewed all of the polyfits and ensured that the respective properties represent the polyfit results appropriately. If the polyfits aren't actually used anywhere else, I think they can be removed. They do form the data basis for the material property equations, so someone may want to look back for that.

PR #2410, opened 2026-01-07
Adding Environment Variable Checks to support unit tests in read-only environments

opotowsky:

_NOTE: This PR is built on top of the `clean-path` branch via a squash commit for my own testing. Once that is merged, this PR will show a cleaner diff._

This PR adds a check for an env variable in two places: During the `logs` dir creation and as a `root` arg override in `TemporaryDirectoryChanger`. This is to support read-only containerization efforts. ARMI's unit tests wouldn't all be able to run with these edits, but they don't have to. The downstream tests can. 

## SCR Information

Change Type: features

One-Sentence Rationale: These changes are needed to support ARMI being run in read-only containers downstream.

---

## Checklist

### DIFF
--- a/armi/tests/test_runLog.py
+++ b/armi/tests/test_runLog.py
@@ -20,8 +20,6 @@ from io import StringIO
 from pathlib import Path
 from shutil import rmtree
 
-import pytest
-
 from armi import runLog
 from armi.tests import mockRunLogs
 from armi.utils.directoryChangers import TemporaryDirectoryChanger
@@ -464,6 +462,12 @@ class TestRunLogEnvEdits(unittest.TestCase):
     """Tests that will use monkeypatch to alter an environment variable."""
 
     def setUp(self):
+        # We cannot import pytest at the top of the file right now. The ARMI unit tests are currently imported at
+        # runtime, and until that is changed, we don't want pytest to be a runtime dependency. For now, hide the import
+        # down here. Once the testing module is complete and ARMI's unit tests aren't all imported, the pytest import
+        # can move up to where it belongs.
+        import pytest
+
         self.monkeypatch = pytest.MonkeyPatch()
 
     def tearDown(self):
--- a/armi/utils/tests/test_directoryChangers.py
+++ b/armi/utils/tests/test_directoryChangers.py
@@ -19,8 +19,6 @@ import shutil
 import unittest
 from pathlib import Path
 
-import pytest
-
 from armi.utils import directoryChangers, directoryChangersMpi
 
 
@@ -158,6 +156,12 @@ class TestDirectoryChangersEnvEdits(unittest.TestCase):
     """Tests that will use monkeypatch to alter an environment variable."""
 
     def setUp(self):
+        # We cannot import pytest at the top of the file right now. The ARMI unit tests are currently imported at
+        # runtime, and until that is changed, we don't want pytest to be a runtime dependency. For now, hide the import
+        # down here. Once the testing module is complete and ARMI's unit tests aren't all imported, the pytest import
+        # can move up to where it belongs.
+        import pytest
+
         self.monkeypatch = pytest.MonkeyPatch()
 
     def tearDown(self):


armi/runLog.py
```diff
@@ -286,6 +285,13 @@ def startLog(self, name):
             sys.stderr = self.stderrLogger
 
 
+def getLogDir():
```
john-science, line 288: Can we get a one-line docstring?

doc/gallery-src/framework/run_blockVolumeFractions.py
```diff
@@ -83,7 +83,7 @@ def plotVolFracsWithComponentTemps(b, uniformTemps):
     plt.show()
 
 
-uniformTempsInC = [400.0, 500.0, 600.0, 700.0, 800.0, 900.0, 1200.0]
+uniformTempsInC = [300.0, 400.0, 500.0, 600.0, 700.0]
```
john-science, line 86: I think you need to merge ARMI `main` into this branch?

opotowsky: So I pulled in `clean-path` then squashed the commits so the commit history would be easier to see on this branch. 

I need this + clean-path together for my testing. 

You can review the last 2 commits alone, or you can wait until `clean-path` is merged and this PR is updated with `main`

armi/runLog.py
opotowsky: fo sho! Wasn't sure you'd like this approach so didn't document it yet :-)

opotowsky: https://github.com/terrapower/armi/pull/2410/commits/b0225d8e37626b41ec65bcc663f63cf23f8e94a4

armi/tests/test_runLog.py
```diff
@@ -17,8 +17,11 @@
 import os
 import unittest
 from io import StringIO
+from pathlib import Path
 from shutil import rmtree
 
+import pytest
```
john-science: Oh, well, actually. We need to talk about this.

So, your team really wants to be able to import the ARMI unit tests. And they do that all over the place. And right now `pytest` is not part of the ARMI runtime dependencies.

But if you add this line here, then we also need to add `pytest` to the ARMI runtime dependencies. I'm not THRILLED about that. If you think this is super important, we can do it. I'll need a little convincing, maybe.

opotowsky: this is a unit test file. I don't think there's any runtime implications

opotowsky:
> I believe your new `monkeypatch` test leaves some crumbs around in the ARMI working area, outside any temp dirs. That's why CI is failing.

Wow -- only happening on windows! I'll look into it.

armi/tests/test_runLog.py
opotowsky: Talked offline. I forgot ALL unit tests were imported at runtime until we can get the testing module migration complete. I will push up a fix

doc/gallery-src/framework/run_blockVolumeFractions.py
opotowsky: cleaned up!

armi/tests/test_runLog.py
opotowsky: https://github.com/terrapower/armi/pull/2410/changes/15d36dac576fcb7adb070ee4477252bb7512c580

armi/runLog.py
```diff
@@ -286,6 +285,16 @@ def startLog(self, name):
...
+    """This returns a file path for the `logs` directory, first checking if the user set the ARMI_TEMP_ROOT_PATH
+    environment variable.
+    """
+    if os.environ.get("ARMI_TEMP_ROOT_PATH"):
```
john-science, line 292: Should we document `ARMI_TEMP_ROOT_PATH` in the docs somewhere?

Thoughts on the best way to document this?

opotowsky: It wasn't immediately obvious to me. It's used to support running tests in a read-only environment. But ARMI, even if using this env variable, still wouldn't be able to run its unit tests in a container, since many tests create files not in a temp dir. So it feels weird to document something that doesn't even work. And there's no existing container docs. So while I had the plan to note it in the docs, I felt weird documenting a half-functional "feature". Do you have thoughts on what you'd like to see?

john-science: Well, ARMI runs in containers every day. The GitHub Actions are spawned onto Docker containers. 

But they are just writable containers, that's all.

armi/utils/directoryChangers.py
```diff
@@ -252,6 +254,11 @@ def __init__(
             outputPath,
         )
 
+        # If an application sets this environment variable, all root args in all `TempDirChanger` uses are overriden
+        # with a different root path. This is useful for running unit tests in a read-only environment.
+        if os.environ.get("ARMI_TEMP_ROOT_PATH"):
```
john-science, line 259: Honest question.

Should this be something like:

```python
if root is None and os.environ.get("ARMI_TEMP_ROOT_PATH"):
```

The version you have completely overrides the input `root` parameter. Is that a necessary part of the design?

armi/runLog.py
opotowsky: For posterity: the use case of read-only containers is currently limited to one user base. We will keep the env var documentation in the docstrings only for now. Should there be more of a need to support read-only containers, we can add a section to the dev manual on it. (This still won't allow ARMI's unit tests to run in one without a lot of work, but we can write some general documentation).

armi/utils/directoryChangers.py
opotowsky: Yes that is a necessary part of the design. Downstream, all unit tests pass in the root arg, which needs to be overridden in read-only environments.

PR #2411, opened 2026-01-08
Refactoring Shuffling Logic

jasonbmeng:

These changes enable different ARMI applications to share a common shuffling input definition utilizing `ring` & `position` at `cycle` instead of `assemblyName`. This is particularly useful for movements where an assembly which has been discharged from the core (e.g. to the spent fuel pool) needs to be reintroduced into the core in a later cycle of operation.

Currently, the input format to specify an assembly moving from the spent fuel pool back into the core uses the `assemblyName`. However, `assemblyName` (i.e. assembly number) is not guaranteed to be consistent across ARMI Apps. For example, one App may run a full core model while another leverages core symmetry and models a 1/3 symmetric core model. Both applications should be able to interrogate a common shuffling definition and parse consistent assembly movements.

This PR moves spent fuel pool assembly identification away from using the `assemblyName`. Instead, the assembly location history is used. The input is now `ring`, `position`, and `cycle`. This is used to identify the target assembly which resided at `(ring, position)` during `cycle`.

Constraints Discussed: New assembly params used to track location history need to be arrays instead of dictionaries.

## SCR Information

Change Type: features

One-Sentence Rationale: Different ARMI Applications should be able to share common shuffling input definitions which are independent of application specifics such as assembly numbering.

One-line Impact on Requirements: The requirement R_ARMI_SHUFFLE is touched, including the implementation I_ARMI_SHUFFLE and testing T_ARMI_SHUFFLE.

---

## Checklist


john-science:
@jasonbmeng Some questions:

> However, assemblyName (i.e. assembly number) is not guaranteed to be consistent across ARMI Apps. 

* Why can't you just use the object ID?
* So, have you gotten an agreement from [Other Team] that they will switch their shuffling logic tools to use the new naming?

> This PR moves spent fuel pool assembly identification away from using the assemblyName.

Okay, I will need you to make sure this doesn't break the functionality for [Other Team] before you merge. (Standard stuff.)

> You still have broken tests. How's that going?

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -230,6 +257,63 @@ def outage(self, factor=1.0):
...
+            a.p.ringPosHist = []
+        # param length is shorter than expected
+        # (data from previous cycles is missing or shuffling was not performed
+        # on a previous cycle)
+        while len(a.p.ringPosHist) < cycle:
+            a.p.ringPosHist.append(None)
+        # param length is longer than expected. perhaps a restart analysis of some sort.
+        # trim trailing data to correct length
+        while len(a.p.ringPosHist) > cycle:
+            a.p.ringPosHist.pop(-1)
```
john-science: I don't think we need this loop. It is easy to crop items off the end of a list:

```python
>>> x = list(range(10))
>>> x
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> x[:7]
[0, 1, 2, 3, 4, 5, 6]
```

Probably, something like this would work:

```python
if len(a.p.ringPosHist) > cycle:
    a.p.ringPosHist = a.p.ringPosHist[:cycle]
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -230,6 +257,63 @@ def outage(self, factor=1.0):
...
+            cycle number at BOC to update assembly location history
+        """
+        # init list if needed
+        if a.p.ringPosHist is None:
+            a.p.ringPosHist = []
+        # param length is shorter than expected
+        # (data from previous cycles is missing or shuffling was not performed
+        # on a previous cycle)
+        while len(a.p.ringPosHist) < cycle:
+            a.p.ringPosHist.append(None)
```
john-science: Lists are pretty easy to bulk append to, I think we can speed up this loop:

```python
>>> x = list(range(10))
>>> x
>>> x + [None]*(12-10)
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, None, None]
```

We could do SOMETHING like:

```python
if len(a.p.ringPosHist) < cycle:
    a.p.ringPosHist = a.p.ringPosHist + [None]*(cycle - len(a.p.ringPosHist))
```

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -230,6 +257,63 @@ def outage(self, factor=1.0):
...
+        else:
+            ring, pos, _ = grids.locatorLabelToIndices(a.getLocation())
+            a.p.ringPosHist.append((ring, pos))
+    
+    def updateAllLocationHistParams(self, cycle):
+        """
+        update location history param for all assemblies with current assembly locations
+        for specified cycle number
+        index of a.p.ringPosHist corresponds to the cycle number BOC assembly location
+        e.g. i=0 is the initial position, i=1 is the position at BOC1, etc.
```
john-science:
```suggestion
        Update location history param for all assemblies with current assembly locations for specified cycle number.

        The index of a.p.ringPosHist corresponds to the cycle number BOC assembly location e.g. i=0 is the initial
        position, i=1 is the position at BOC1, etc.
```

john-science: Just some linting and formatting.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -230,6 +257,63 @@ def outage(self, factor=1.0):
...
+        """
+        for a in self.r.core:
+            self._updateAssemLocationHistParam(a, cycle)
+        sfpAssems = list(self.r.excore["sfp"])
+        for a in sfpAssems:
```
john-science:
```suggestion

        for a in list(self.r.excore["sfp"]):
```

john-science: Just style points.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -1087,19 +1174,19 @@ def readMoves(fname):
                 # due to legacy reasons, the assembly type will be put into group 4
                 pat = (
                     r"([A-Za-z0-9!\-]+) moved to ([A-Za-z0-9!\-]+) with assembly type "
-                    + r"([A-Za-z0-9!\s]+)\s*(ANAME=\S+)?\s*with enrich list: (.+)"
+                    + r"([A-Za-z0-9!\s]+)\s*(ringPosCycle=\[.*\])?\s*with enrich list: (.+)"
```
john-science, line 1172: Can you explain this to me? Along with the change to line 1165.

Thanks!

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -565,6 +568,52 @@ def runShuffling(self, fh):
         self.o.cs.caseTitle = "armiRun2"
         fh.interactBOL()
 
+        # expected assembly position history based on shuffling specification of this test.
+        # do not blindly rebase these reference values. test failures using this dict
+        # imply that the assembly shuffling definition has changed.
+        expPosHist = {}
```
john-science, line 574: Honest Question.

Why is this so long / verbose?

john-science: We discussed this.

armi/physics/fuelCycle/fuelHandlers.py
jasonbmeng: Agree, slicing the list is better. Incorporated.

armi/physics/fuelCycle/fuelHandlers.py
jasonbmeng: Agree, incorporated.

armi/physics/fuelCycle/fuelHandlers.py
jasonbmeng: Incorporated.

armi/physics/fuelCycle/fuelHandlers.py
jasonbmeng: Incorporated. Was following style of one of the existing loops through assemblies in the SFP. Changed all to be consistent with this.

armi/physics/fuelCycle/fuelHandlers.py
jasonbmeng: Sure thing.
https://github.com/terrapower/armi/blob/4e221aaea54a1d339ed18d6b7bdddbc53044de16/armi/physics/fuelCycle/fuelHandlerInterface.py#L162-L169
This modifies the movement output log, so this line is a consistency change to look for the corresponding pattern.

The additional whitespace in line 1165 avoids accidentally hitting `cycle` in the newly added `ringPosCycle=`

jasonbmeng:
> Why can't you just use the object ID?

Upstream analysis cases are sometimes of varying core symmetry. Object ID is not guaranteed to be consistent.

> So, have you gotten an agreement from [Other Team] that they will switch their shuffling logic tools to use the new naming?

[Related Team] is aware of the change

> Okay, I will need you to make sure this doesn't break the functionality for [Other Team] before you merge. (Standard stuff.)

[Related Team] integration tests appear to be passing

> You still have broken tests. How's that going?

Updates pushed to address jagged arrays

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -230,6 +258,61 @@ def outage(self, factor=1.0):
...
+        # (data from previous cycles is missing or shuffling was not performed
+        # on a previous cycle)
+        if len(a.p.ringPosHist) < cycle:
+            a.p.ringPosHist = a.p.ringPosHist + [(a.NOT_CREATED_YET, a.NOT_CREATED_YET)] * (
+                cycle - len(a.p.ringPosHist)
+            )
+        # param length is longer than expected. perhaps a restart analysis of some sort.
+        # trim trailing data to correct length
+        if len(a.p.ringPosHist) > cycle:
+            a.p.ringPosHist = a.p.ringPosHist[:cycle]
```
john-science:
```suggestion
        # Param length is shorter than expected (data from previous cycles is missing or shuffling was not performed
        # on a previous cycle)
        if len(a.p.ringPosHist) < cycle:
            a.p.ringPosHist += [(a.NOT_CREATED_YET, a.NOT_CREATED_YET)] * (cycle - len(a.p.ringPosHist))

        # Param length is longer than expected. perhaps a restart analysis of some sort. trim trailing data
        if len(a.p.ringPosHist) > cycle:
            a.p.ringPosHist = a.p.ringPosHist[:cycle]
```

john-science: Two minor things:

1. ARMI uses lines of length 120, so the comments can be on fewer lines
2. I used `list += list` to shorten one of your statements.

Nothing exciting.

armi/reactor/assemblies.py
```diff
@@ -76,6 +78,10 @@ def __init__(self, typ, assemNum=None):
         self.lastLocationLabel = self.LOAD_QUEUE
         self.p.orientation = np.array((0.0, 0.0, 0.0))
 
+        # Manually set some parameters at BOL
+        for problemParam in ["ringPosHist"]:
+            self.p[problemParam] = []
```
john-science: 1. There are a few other parameters above this also being set to default values.
2. This isn't exactly "BOL", it's at assembly creation. Which might be 40 years after BOL.
3. This doesn't need to be a loop, there is only one value.

I think you can just add `self.p["ringPosHist"] = []` to line 80, and skip the comment and the loop.

Thoughts?

john-science: Are there any other parameters which are lists, and not numpy arrays?

Honest question.

armi/reactor/assemblyParameters.py
```diff
@@ -277,6 +277,16 @@ def _enforceNotesRestrictions(self, value):
             saveToDB=True,
         )
 
+        pb.defParam(
+            "ringPosHist",
+            units=units.UNITLESS,
+            description=(
+                "Ring and position history for this assembly written at BOC. Index 1 corresponds to position at BOC1."
```
john-science, line 284:
```suggestion
                "Ring and position history for this assembly, by cycle."
```

john-science: I'm just trying to tighten up the language.

Thoughts?

armi/reactor/blueprints/reactorBlueprint.py
```diff
@@ -224,6 +224,18 @@ def _loadComposites(self, cs, bp, container, gridContents, orientationBOL):
...
+        for a in container:
+            loc = a.getLocation()
+            if loc in a.NOT_IN_CORE:
+                a.p.ringPosHist = [(loc, loc)]
+            else:
+                try:
+                    ring, pos, _ = grids.locatorLabelToIndices(a.getLocation())
+                    a.p.ringPosHist = [(ring, pos)]
+                except ValueError:  # some ex-core structures do not appear to have valid locator label indices.
+                    a.p.ringPosHist = [(a.NOT_CREATED_YET, a.NOT_CREATED_YET)]
```
john-science:
```suggestion
                except ValueError:
                    # some ex-core structures may not have valid locator label indices
                    a.p.ringPosHist = [(a.NOT_CREATED_YET, a.NOT_CREATED_YET)]
```

armi/reactor/converters/geometryConverters.py
```diff
@@ -1080,6 +1081,7 @@ def convert(self, r: reactors.Reactor):
                 if thisZone:
                     thisZone.addLoc(newAssem.getLocation())
                 self._newAssembliesAdded.append(newAssem)
+                self.updateThirdToFullCoreLocHist(newAssem, count - 2)
```
john-science: I think this should be a private method. Because you don't want people calling this method before `self.grid` is populated and correct.

The alternate solution is to make this a `staticmethod` and pass `grid` in (it would no longer be a `self.grid`).

armi/reactor/cores.py
```diff
@@ -1442,6 +1442,33 @@ def getAssemblyWithStringLocation(self, locationString):
...
+        # search core
+        for a in self:
+            if self._checkIfAssemAtRingPosCycle(a, ring, pos, cycleNum):
+                return a
+        # search sfp
+        if self.parent.excore.get("sfp") is not None:
+            for a in list(self.r.excore["sfp"]):
+                if self._checkIfAssemAtRingPosCycle(a, ring, pos, cycleNum):
+                    return a
+        return
```
john-science: Wait. Two things:

1. This empty `return` statement returns... `None`. So, if that's your idea, and that's on purpose, this should be `return None`, make it explicit and easy to read.
2. If this isn't _supposed_ to return `None`, than this is probably an error and we need to beef up our unit testing to catch this case.

What do you think?

armi/physics/fuelCycle/fuelHandlers.py
jasonbmeng: Incorporated.

armi/reactor/assemblies.py
jasonbmeng: Followed similar format to [Related Team] plugins for initing problematic block params. But I agree with these comments.

Removed loop and switched to initing the param directly with `self.p.ringPosHist = []`

jasonbmeng: I'm not aware of others as far as assembly parameters. There are some block parameters registered by [Related Team] plugins that are initialized in this way.

armi/reactor/assemblyParameters.py
jasonbmeng: I do think some clarification in the block description is helpful here. It appears that some external physics kernels may interpret cycle 0 to actually be the first cycle of operation. This description makes it clear that the list index of this param corresponds to the cycle number.

armi/reactor/blueprints/reactorBlueprint.py
jasonbmeng: Incorporated

armi/reactor/converters/geometryConverters.py
jasonbmeng: Agree, this would be better as a private method. Incorporated.

armi/reactor/cores.py
jasonbmeng: `None` was the intent here and is handled when this method is called in `FuelHandler`.

Incorporated `return None` for readability.

armi/physics/fuelCycle/fuelHandlers.py
john-science: It's funny that GitHub doesn't seem to see your changes. Are you manually copy/pasting these changes? I just want to be sure that you know that you can click "Commit suggestion" or do multiple with "add suggestion to batch". It's an easy solution, if you don't know about it, that's all.

armi/physics/fuelCycle/fuelHandlers.py
```diff
@@ -230,6 +258,57 @@ def outage(self, factor=1.0):
...
+        else:
+            ring, pos, _ = grids.locatorLabelToIndices(a.getLocation())
+            a.p.ringPosHist.append((ring, pos))
+
+    def updateAllLocationHistParams(self, cycle):
+        """
+        Update location history param for all assemblies with current assembly locations
+        for specified cycle number
+        Index of a.p.ringPosHist corresponds to the cycle number BOC assembly location
+        e.g. i=0 is the initial position, i=1 is the position at BOC1, etc.
```
john-science, line 305:
```suggestion
        Update assembly location history parameter with current assembly location for specified cycle number.

        Index of a.p.ringPosHist corresponds to the cycle number BOC assembly location, e.g. i=0 is the initial
        position, i=1 is the position at BOC1, etc.
        """
        a = self._preconditionLocationHistParam(a, cycle)
        # assem param should now be the correct len. append data at correct index.
        if a.getLocation() in a.NOT_IN_CORE:
            a.p.ringPosHist.append((a.getLocation(), a.getLocation()))
        else:
            ring, pos, _ = grids.locatorLabelToIndices(a.getLocation())
            a.p.ringPosHist.append((ring, pos))

    def updateAllLocationHistParams(self, cycle):
        """
        Update location history param for all assemblies with current assembly locations for specified cycle number.

        Index of a.p.ringPosHist corresponds to the cycle number BOC assembly location, e.g. i=0 is the initial
        position, i=1 is the position at BOC1, etc.
```

armi/reactor/assemblies.py
john-science: Well, long-term, maybe this param should be a numpy array and not a list. But I guess that can be a follow on PR some time.

doc/user/inputs.rst
```diff
@@ -1426,22 +1429,26 @@ The code will crash if materials used in :ref:`blocks-and-components` contain nu
 Fuel Management Input
 =====================
 
-Fuel management in ARMI is specified through custom Python scripts that often reside
+Fuel management in ARMI is specified through custom Python scripts or yaml file definitions that often reside
```
john-science:
```suggestion
Fuel management in ARMI is specified through custom Python scripts or YAML files that often reside
```

doc/user/inputs.rst
```diff
@@ -1426,22 +1429,26 @@ The code will crash if materials used in :ref:`blocks-and-components` contain nu
...
 ``fuelHandlerName``
    The name of a FuelHandler class that ARMI will look for in the Fuel Management Input module or file
    specified by ``shuffleLogic``. Since it's input, it's the user's responsibility
    to design and place that object in that module or file.
 
 .. note:: We consider the limited syntax needed to express fuel management in Python
    code itself to be sufficiently expressive and simple for non-programmers to
-   actually use. Indeed, this has been our experience.
+   actually use. Indeed, this has been our experience. Yaml input format options for fuel management
+   are additionally available for Users.
```
john-science:
```suggestion
   actually use. Also, fuel management options are available through YAML input files.
```

armi/reactor/cores.py
```diff
@@ -1442,6 +1442,33 @@ def getAssemblyWithStringLocation(self, locationString):
...
+        # search core
+        for a in self:
+            if self._checkIfAssemAtRingPosCycle(a, ring, pos, cycleNum):
+                return a
+        # search sfp
+        if self.parent.excore.get("sfp") is not None:
+            for a in list(self.r.excore["sfp"]):
+                if self._checkIfAssemAtRingPosCycle(a, ring, pos, cycleNum):
+                    return a
+        return None
```
john-science, line 1474:
```suggestion
        for a in self:
            if self._checkIfAssemAtRingPosCycle(a, ring, pos, cycleNum):
                return a

        # search sfp
        if self.parent.excore.get("sfp") is not None:
            for a in list(self.r.excore["sfp"]):
                if self._checkIfAssemAtRingPosCycle(a, ring, pos, cycleNum):
                    return a

        return None
```

armi/reactor/cores.py
```diff
@@ -1442,6 +1442,33 @@ def getAssemblyWithStringLocation(self, locationString):
...
+        nCycles = len(a.p.ringPosHist)
+        if nCycles >= cycleNum:  # requested cycleNum has data populated
+            if a.p.ringPosHist[cycleNum] == (ring, pos):
+                return True
+        return False
+
+    def getAssemblyWithRingPosHist(self, ring, pos, cycleNum):
+        """
+        Search the Core and SFP for assembly which resided at specified ring and position at specified cycle.
+        This is an alternative to getting an assembly by number or string location.
```
john-science, line 1461:
```suggestion
        Search the Core and SFP for assembly which resided at specified ring and position at specified cycle.

        This is an alternative to getting an assembly by number or string location.
```

armi/reactor/cores.py
```diff
@@ -1442,6 +1442,33 @@ def getAssemblyWithStringLocation(self, locationString):
         assem = self.childrenByLocator.get(loc)
         return assem
 
+    def _checkIfAssemAtRingPosCycle(self, a, ring, pos, cycleNum):
+        """
+        Interrogate location history param of specified assembly object
+        Return True if assembly was at specified (ring, pos) at specified cycleNum BOC.
```
john-science:
```suggestion
        Interrogate location history param of specified assembly object

        Return True if assembly was at specified (ring, pos) at specified cycleNum BOC.
```

armi/reactor/cores.py
```diff
@@ -1442,6 +1442,37 @@ def getAssemblyWithStringLocation(self, locationString):
         assem = self.childrenByLocator.get(loc)
         return assem
 
+    def _checkIfAssemAtRingPosCycle(self, a, ring, pos, cycleNum):
+        """
+        Interrogate location history param of specified assembly object
```
john-science:
```suggestion
        Interrogate location history param of specified assembly object.
```

PR #2424, opened 2026-01-14
Removing deprecation warning from test

opotowsky:

I noticed a deprecation warning for a test, and then noticed it was the last use of the deprecation here and downstream. Easy to clean up and remove the deprecation warning too.

## SCR Information

Change Type: trivial

One-Sentence Rationale: Code maintenance is important to well functioning code.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database.py
+++ b/armi/bookkeeping/db/database.py
@@ -1437,8 +1437,8 @@ class Database:
         for c, ndensDict in zip(comps, unpackedData):
             nuclides = np.array(list(ndensDict.keys()), dtype="S6")
             numberDensities = np.array(list(ndensDict.values()), dtype=np.float64)
-            c.p["nuclides"] = nuclides
-            c.p["numberDensities"] = numberDensities
+            c.p.nuclides = nuclides
+            c.p.numberDensities = numberDensities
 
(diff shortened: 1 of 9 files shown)


armi/reactor/blocks.py
```diff
@@ -1777,20 +1776,6 @@ def getDuctOP(self):
...
-        Deprecated. Use :meth:`assignPinIndices` to additionally update component parameters.
-        """
-        # stacklevel=2 means the warning traceback, file, and line numbers will reflect the caller
-        # of this method, not this method itself.
-        warnings.warn(
-            "HexBlock.initializePinLocations is deprecated. Please use assignPinIndices",
-            category=DeprecationWarning,
-            stacklevel=2,
-        )
-        self.assignPinIndices()
```
john-science, line 1792: You also removed this line?

opotowsky: Yea the `initializePinLocations` --> `assignPinIndices` is no longer needed since there's no other use of `initializePinLocations` I can find downstream. So the deprecation is now complete!

john-science: Okay, I agree. I can find no other usage of this method downstream.

PR #2428, opened 2026-01-16
Copying assembly flags when creating a new assembly from type

mgjarrett:

The uniform mesh converter has a private method `_createNewAssembly` that creates a new instance of an assembly based on the `class` and `type` of an input `sourceAssembly`.

Previously, the new assembly was not explicitly assigned flags, so it would just assume the flags produced by `Flags.fromStringIgnoreErrors(sourceAssembly.getType())`. If `sourceAssembly` had explicitly defined flags in the blueprints that were different from what would be generated from the type string, then the new assembly flags would differ from the `sourceAssembly`, which is undesirable. This change copies the flags from the `sourceAssembly` to the new assembly.

## SCR Information

Change Type: features

One-Sentence Rationale: When creating a copy of a given assembly, it is desirable to preserve whatever flags have been assigned to the original source assembly.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database.py
+++ b/armi/bookkeeping/db/database.py
@@ -1437,8 +1437,8 @@ class Database:
         for c, ndensDict in zip(comps, unpackedData):
             nuclides = np.array(list(ndensDict.keys()), dtype="S6")
             numberDensities = np.array(list(ndensDict.values()), dtype=np.float64)
-            c.p["nuclides"] = nuclides
-            c.p["numberDensities"] = numberDensities
+            c.p.nuclides = nuclides
+            c.p.numberDensities = numberDensities
 
     @staticmethod
     def getCycleNodeAtTime(dbPath, startTime, endTime, errorIfNotExactlyOne=True):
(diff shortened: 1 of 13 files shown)


armi/reactor/converters/uniformMesh.py
```diff
@@ -969,6 +969,7 @@ def _createNewAssembly(sourceAssembly):
         a = sourceAssembly.__class__(sourceAssembly.getType())
         a.spatialGrid = grids.AxialGrid.fromNCells(len(sourceAssembly))
         a.setName(sourceAssembly.getName())
+        a.p.flags = sourceAssembly.p.flags
```
john-science, line 972: Okay, I'm wondering if this needs to be a `copy.copy())`.  I'm not sure, but what we are doing is copying a pointer to a Flags collection over. So, if after this mesh conversion is done we add a flag to one assembly, it will magically appear in the other assembly.

Is that what we want? My guess is no. But what do you think?

mgjarrett: I think it is already a unique copy of the Flags collection. I added a couple lines to the unit test to verify that changes to `sourceAssembly.p.flags` after the assignment don't affect `a.p.flags`.

john-science: Yeah, I see that. 

Okay, wild. Great.

john-science:
@mgjarrett Holler when you are ready for this to be merged.

PR #2430, opened 2026-01-16
Moving 27 TH parameters out of ARMI

john-science:

ARMI currently has one `Setting` listed as for thermal hydraulics, but it is not used in ARMI: `doTH`. This setting is being moved out of ARMI.

ARMI currently contains these 31 thermal hydraulics Parameters, which are not used in ARMI and are only needed by one project downstream:

### Block Params

* TH0SigmaCladIDT
* TH0SigmaCladODT
* TH2SigmaCladIDT
* TH2SigmaCladODT
* TH3SigmaCladIDT
* TH3SigmaCladODT
* THaverageCladODT
* THaverageCladTemp
* THaverageDuctTemp
* THcoolantStaticT
* THhotChannel
* THhotChannelCladIDT
* THhotChannelCladODT
* THhotChannelFuelCenterlineT
* THhotChannelFuelODT
* THhotChannelHeatTransferCoeff
* THhotChannelOutletT
* THlocalDTout
* THlocalDToutFuel
* THorificeZone
* THTfuelCL
* THTfuelOD

### Assembly Params

* THdeltaPNoGrav
* THdeltaPPump
* THlocalDTout
* THlocalDToutFuel
* THorificeZone

These Parameters are not used in ARMI, and were deemed not general enough to be interesting to a wide nuclear engineering audience. These Parameters are used by a single, downstream project. And for the sake of quality, that downstream project should be responsible for maintaining these parameters.

## SCR Information

Change Type: trivial

One-Sentence Rationale: Ideally, ARMI should not be responsible for maintain features that are not generally useful to a wide audience, and only used by one downstream project.

---

## Checklist

### DIFF
--- a/armi/reactor/converters/tests/test_uniformMesh.py
+++ b/armi/reactor/converters/tests/test_uniformMesh.py
@@ -71,6 +71,8 @@ class TestAssemblyUniformMesh(unittest.TestCase):
 
     def test_makeAssemWithUniformMesh(self):
         sourceAssem = self.r.core.getFirstAssembly(Flags.IGNITER)
+        # assign different flags to test flag preservation
+        sourceAssem.p.flags = Flags.FUEL | Flags.IGNITER | Flags.TEST
 
         self.converter._generateUniformMesh(minimumMeshSize=0.01)
         b = sourceAssem.getFirstBlock(Flags.FUEL)
@@ -80,6 +82,10 @@ class TestAssemblyUniformMesh(unittest.TestCase):
             paramMapper=uniformMesh.ParamMapper([], ["power"], b),
             mapNumberDensities=True,
         )
+        self.assertEqual(newAssem.p.flags, sourceAssem.p.flags)
+        # chnage sourceAssem flags to verify that a unique copy was made
+        sourceAssem.p.flags = Flags.FUEL | Flags.IGNITER
+        self.assertNotEqual(newAssem.p.flags, sourceAssem.p.flags)
 
         prevB = None
         for newB in newAssem:
--- a/armi/reactor/converters/uniformMesh.py
+++ b/armi/reactor/converters/uniformMesh.py
@@ -969,6 +969,7 @@ class UniformMeshGeometryConverter(GeometryConverter):
         a = sourceAssembly.__class__(sourceAssembly.getType())
         a.spatialGrid = grids.AxialGrid.fromNCells(len(sourceAssembly))
         a.setName(sourceAssembly.getName())
+        a.p.flags = sourceAssembly.p.flags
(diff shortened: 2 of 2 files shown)


drewj-tp:
:shrug:

I get there are some very niche / bespoke parameters that we don't think have broad applicability. But saying that the assembly inlet temperature and block pressure drop "are not generally useful to a wide audience" and "deemed not general enough to be interesting to a wide nuclear engineering audience" and therefore should not be included in a general purpose framework, well why do we have any parameters? (hyperbolically) 

Every liquid fueled reactor that has assemblies that are cooled will care about inlet temperature and pressure drop. LWRs, SFRs, HTGRs, PBRs (maybe stretching the limits here). I feel it's sufficient to call them decently fundamental. 0/2/3 sigma clad IDT / ODT? Sure, more niche.

(strawman argument)
ARMI doesn't (really) do anything with `Core.p.keff`. It floats around a bit, and it's used in the `GlobalFluxInterface` which isn't added to the main app in any meaningful way. So why do we keep that parameter? It's (IMO) "generally useful to a wide audience" and "general enough to be interesting to a wide nuclear engineering audience"

(/strawman argument)

There is certainly a balance between keeping ARMI general purpose so someone has a decent footing on which to stand. And sure, maybe people don't want to call the assembly inlet temperature `THcoolantInletT` and would prefer `coolantInletTemp`. But at that point (okay one more strawman argument) why do we define any settings and parameters?

This comment is not entirely helpful, I get it. I know ARMI has a lot of the niche params that don't need to be in ARMI. I think, at a bare minimum, we should keep these parameters

1. assembly inlet coolant temperature,
2. total assembly pressure drop,
3. block inlet coolant temperature,
4. total block pressure drop

Those, to me, feel "generally useful to a wide audience" and "general enough to be interesting to a wide nuclear engineering audience"

jakehader:
I agree with keeping several of these parameters in the framework, for instance pressure drop, inlet temperatures but component type dependent parameters for the framework ought to at least be removed or replaced with a more general version. If anything perhaps it's a good idea to have a basic plugin that is neutronics, thermal hydraulics, etc available in the open source that the framework doesn't include by default, but as a starter pack.

john-science:
Okay, I can keep some of these in ARMI. That's a fine and easy change to make. 

I created this PR, in part, because all of these params have the same lack of documentation and information. Most of these could be anything, and several of them are totally unused in any plugins I can see.

@drewj-tp and @jakehader just list the parameters that you want me to keep in ARMI. NBD.

opotowsky (approved):
This looks good, but let's stay on touch regarding when this should be merged!

PR #2432, opened 2026-01-20
Adding `forceClean` to snapshot deletions

opotowsky:

Given the recent `cleanPath` changes, the snapshot directories were no longer greenlit for deletion like they were before. `snapshot` was in the `DO_NOT_CLEAN_PATHS` list, which opposite to its name, was actually a list of greenlit deletions. At the time of removing that list, we didn't think any of those paths were accurate to more recent ARMI uses. Well, `snapshot` proved that theory wrong! This fixes that.

## SCR Information

Change Type: features

One-Sentence Rationale: This is a change that was determined to be needed based on recent feature updates to `cleanPath`.

---

## Checklist


john-science (approved):
Holler when the downstream testing is finished.

PR #2436, opened 2026-01-27
Optimizing the HistoryTracker EOL interaction

mgjarrett:

The HistoryTracker interface writes out the history for detailed assemblies (defined by the `detailAssemNums` settings). The interface was written in a way that it had to reload the HDF5 file for each individual assembly that it writes a history for.  This PR refactors the access to the HDF5 database so that data for all assemblies is pulled simultaneously. This will speed up the interface by roughly a factor of the number of detailed assemblies, which can be on the order of 100 or more for some cases.

## SCR Information

Change Type: features

One-Sentence Rationale: The HistoryTracker interface could take several days to run for a large full core model with 100-200 detailed assemblies being tracked.

---

## Checklist


john-science (approved):
I like the change. Am I still waiting for tests?

PR #2461, opened 2026-02-25
Renaming _Material_Test to AbstractMaterialTest

### LINKED ISSUE #2394: Add comment on the name `_Material_Test`
In our materials unit tests we have a class named `_Material_Test`:

https://github.com/terrapower/armi/blob/95d94a4d58b175cd88e25a64670c1cdf1ea02cd4/armi/materials/tests/test_materials.py#L30

This class name starts with an underscore, so apparently it is meant to be private to that file. But it is also clearly a base class, imported and used ~50 times throughout ARMI. So, we should at least get rid of that leading underscore.

Also, our test name convention in ARMI is `CamelCase` not `Snake_Case`, and we usually start test class names with `TestX`. So you would expect this class to be named:

```python
TestMaterial
```

## NOTE

The name was originally chosen to prevent pytest auto-finding this base class. So, that's fine. We just need to explain that in the comments.

john-science:

We have a test base class `_Material_Test`, which violates a bunch of our naming conventions, so we are changing the name to `AbstractMaterialTest`.

close #2394

## SCR Information

Change Type: trivial

One-Sentence Rationale: We follow a naming convention because it makes the code easier to read and understand.

One-line Impact on Requirements: Trivial implementation changes to the tests: T_ARMI_MAT_FLUID2, T_ARMI_MAT_FRACS0, T_ARMI_MAT_FRACS1, T_ARMI_MAT_FRACS2, T_ARMI_MAT_FRACS3, and T_ARMI_MAT_FRACS4.

---

## Checklist

### DIFF
--- a/armi/materials/tests/test_materials.py
+++ b/armi/materials/tests/test_materials.py
@@ -27,13 +27,7 @@ from armi.utils import units
 
 
 class AbstractMaterialTest:
-    """
-    Base for all specific material test cases.
-
-    Notes
-    -----
-    This name does not match our naming conventions because we are hiding this base class from pytest auto-discovery.
-    """
+    """Base for material tests."""
 
     MAT_CLASS = None
     VALID_TEMP_K = 500


armi/materials/tests/test_materials.py
```diff
@@ -27,7 +27,13 @@
 
 
 class _Material_Test:
-    """Base for all specific material test cases."""
+    """
+    Base for all specific material test cases.
+
+    Notes
```
opotowsky: I am willing to support downstream if you want to actually fix/change this.

john-science: Oh yeah?

I'm not sure I'm feeling it right now.  

I guess the problem is: we still can't have it be called "TestMaterial" or anything, because pytest will find it.

BUT I still do kind of dislike that it has a leading underscore. That's strange, considering how many places downstream it is imported.

I suppose we could rename it to something like `MaterialTestingBaseClass` or something without the word "`Test`" in it. Some trial and error and we could find something that works.

john-science: Okay, you convinced me. I'll just fix the name.

opotowsky (approved):

PR #2468, opened 2026-03-08
Fixing shuffle file reading in reload runs

mgjarrett:

Ensure that the `ring` and `pos` from the `a.p.ringPosHist` parameter are converted to `int` before being compared against `int` values. This is necessary because these values get stored as byte strings in the database.

As part of this PR, the shuffling related unit tests (`test_fuelHandlers.py`) are being updated. The `shuffleLogic` file for the unit tests had shuffling happening in cycle 0, which isn't a real use case. Shuffling should only occur for cycle >= 1. The logic was updated, and some corresponding updates were made to the reference results.

## SCR Information

Change Type: fixes

One-Sentence Rationale: The existing implementation will cause ARMI apps to crash during a restart run that loads from the database, because `a.p.ringPostHist` will be stored as byte strings instead of integers.

---

## Checklist


armi/cases/tests/test_cases.py
```diff
@@ -192,7 +192,7 @@ def test_run(self):
             newSettings = {
                 "branchVerbosity": "important",
                 "coverage": False,
-                "nCycles": 1,
+                "nCycles": 2,
```
john-science, line 195: I don't notice any over-all slow down in the ARMI unit tests. So I guess this is fine.

But, ideally, we try to keep our unit tests lightweight and fast. This looks like we are doubling the length of the run. Why's this necessary?

armi/physics/fuelCycle/tests/test_fuelHandlers.py
```diff
@@ -178,7 +178,7 @@ def setUp(self):
         """
         self.o, self.r = test_reactors.loadTestReactor(
             self.directoryChanger.destination,
-            customSettings={"nCycles": 3, "trackAssems": True},
+            customSettings={"nCycles": 4, "trackAssems": True},
```
john-science, line 181: What's the purpose of this change?

mgjarrett: The shuffling logic in refSmallReactorShuffleLogic.py has 3 shuffling cascades here: 

https://github.com/terrapower/armi/blob/b40ec077ac83a2fb441ab25b8538d0517c33e9b5/armi/tests/refSmallReactorShuffleLogic.py#L22-L26

Previously, these were executed at cycles 0, 1, and 2 in the unit test. But in real ARMI runs, we don't shuffle during cycle 0, because shuffling happens at the beginning of a cycle and nothing has happened yet at the beginning of cycle 0. So you'd have to run 4 cycles to get through 3 rounds of shuffling.

This change pairs with the changes that I made to refSmallReactorShuffleLogic.py, and to fuelHandlers.py::FuelHandler::performShuffle

https://github.com/terrapower/armi/blob/b40ec077ac83a2fb441ab25b8538d0517c33e9b5/armi/physics/fuelCycle/fuelHandlers.py#L1093

I think setting `cycle = self.r.p.cycle + 1` was very confusing here. This is done to read shuffling from a text file, where we print `cycle + 1` to the file:

https://github.com/terrapower/armi/blob/b40ec077ac83a2fb441ab25b8538d0517c33e9b5/armi/physics/fuelCycle/fuelHandlerInterface.py#L155

 If we just eliminate `+ 1` in both instances, then we don't need this weird treatment of the cycle index.

mgjarrett: There is a comment that says we put cycle 0 in so we could do BOL branch searches, but that is a capability that is probably long deprecated. It's not clear to me whether the rest of the features needed for a BOL branch search are still functional.

https://github.com/terrapower/armi/blob/b40ec077ac83a2fb441ab25b8538d0517c33e9b5/armi/physics/fuelCycle/fuelHandlerInterface.py#L152-L154

armi/cases/tests/test_cases.py
mgjarrett: See comment below. The unit test shuffle logic was updated to not shuffle during cycle 0, because that doesn't match up with how ARMI is used in practice.

armi/physics/fuelCycle/tests/test_fuelHandlers.py
john-science: Thanks!

PR #2477, opened 2026-03-16
Adding a block converter for mixed pin assemblies

mgjarrett:

The `MultipleComponentMerger` is capable of merging a list of components into another component within a block. This class has a private method `_verifyExpansion` that checks to ensure the result of this merging operation still has a valid and consistent geometry. The logic within this method assumes that there is only a single type of pin (i.e. a set of concentric circular components) within the block. The check can throw an error when there are multiple types of pins within the block.

This change adds a `MixedAssemblyMerger`, which has a unique geometry check (`_verifyPinExpansion`) that checks the geometry only for the pin containing the components being merged. This component merger can operate on blocks with mixed pin types.

## SCR Information

Change Type: features

One-Sentence Rationale: The existing `MultipleCompnentMerger` is not compatible with mixed-pin blocks.

One-line Impact on Requirements: A new implementation tag, `I_ARMI_BLOCKCONV2`, was added to the `MixedAssemblyMerger` class to denote the implementation of a new aspect of the requirement `R_ARMI_BLOCKCONV` for mixed-pin blocks.

---

## Checklist

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -1,0 +1,39 @@
+name: Build Wheel
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build:
+    if: github.repository == 'terrapower/armi'
+
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+      - name: Install PIP Packages
+        run: |
+          pip install -U pip
+          pip install -e .
+          pip install -U wheel
+      - name: Build Wheels
+        run: |
+          mkdir dist
+          pip wheel . -w dist/
+          chmod 664 dist/armi*.whl
+      - name: Archive PIP wheel artifacts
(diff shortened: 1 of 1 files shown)


armi/reactor/converters/blockConverters.py
```diff
@@ -312,6 +312,78 @@ def convert(self):
...
+                raise ValueError(
+                    "There is a non void component {} in the location where component {} was expanded "
+                    "to absorb component solute {}. solvent dims {}, {} comp dims {} {}.".format(
+                        c, solvent, solute, solvent.p.id, solvent.p.od, c.p.id, c.p.od
+                    )
+                )
+            if c.getArea() < 0.0:
+                runLog.warning(
+                    "Component {} still has negative area after {} was dissolved into {}".format(c, solute, solvent),
+                    single=True,
```
john-science, line 383: single?

I feel like this message can't be hit twice, in practice, right?

mgjarrett: 🤔 

The components don't have unique names -- that is, a "fuel" component in one block will have the same name as the same component in another block. So in theory, the identical string could be produced here multiple times if multiple instances of the same type of block are run through this converter.

PR #2478, opened 2026-03-19
Checking for unrepresented XS IDs

mgjarrett:

When the method `CrossSectionGroupManager.createRepresentativeBlocks()` is invoked, it checks for unrepresented XS IDs and modifies them as necessary. However, when `CrossSectionGroupManager.makeCrossSectionGroups()` is called outside of `createRepresentativeBlocks()`, it can generate unrepresented XS IDs without addressing them. This change ensures that unrepresented XS IDs are always identified and modified to avoid errors.

## SCR Information

Change Type: features

One-Sentence Rationale: This change prevents a condition where a block has a XS type that is not represented by any representative block in the `CrossSectionGroupManager`, which would cause a XS generation application to fail.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/crossSectionGroupManager.py
+++ b/armi/physics/neutronics/crossSectionGroupManager.py
@@ -1376,9 +1376,9 @@ class CrossSectionGroupManager(interfaces.Interface):
 
         Notes
         -----
-        This should be run after `CrossSectionGroupManager._updateEnvironmentGroups()`, which resets
-        `b.p.envGroup` and can result in unrepresented cross section IDs. This is usually invoked
-        as a result of a call to `CrossSectionGroupManager.makeCrossSectionGroups()`
+        This should be run after :meth:`CrossSectionGroupManager._updateEnvironmentGroups`, which resets
+        ``b.p.envGroup`` and can result in unrepresented cross section IDs. This is usually invoked
+        as a result of a call to :meth:`CrossSectionGroupManager.makeCrossSectionGroups`
         """
         self._unrepresentedXSIDs = []
         for xsID, collection in blockCollectionsByXsGroup.items():
--- a/armi/physics/neutronics/tests/test_crossSectionManager.py
+++ b/armi/physics/neutronics/tests/test_crossSectionManager.py
@@ -906,10 +906,8 @@ class TestXSGM(unittest.TestCase):
 
         # set valid flags to something the fuel block would not have to trigger unrepresented block
         fuelXStype = "AD"
-        blocksWithType = [b for b in self.csm.r.core.getBlocks(Flags.FUEL) if b.getMicroSuffix() == fuelXStype]
-        print(blockCollectionsByXsGroup)
+        blocksWithType = [b for b in self.csm.r.core.iterBlocks(Flags.FUEL) if b.getMicroSuffix() == fuelXStype]
         fuelCollection = blockCollectionsByXsGroup[fuelXStype]
-        print(fuelCollection)
         fuelCollection._validRepresentativeBlockTypes = Flags.CLAD
 
         # check for unrepresented XS ID, assert that it is found


armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -900,6 +900,29 @@ def test_getRepresentativeBlocks(self):
...
         self.assertIsNone(blocks[1].p.detailedNDens)
 
+    def test_checkForUnrepresentedXSIDs(self):
+        blockCollectionsByXsGroup = self.csm.makeCrossSectionGroups()
+        self.csm.createRepresentativeBlocks()
+
+        # set valid flags to something the fuel block would not have to trigger unrepresented block
+        fuelXStype = "AD"
+        blocksWithType = [b for b in self.csm.r.core.getBlocks(Flags.FUEL) if b.getMicroSuffix() == fuelXStype]
+        print(blockCollectionsByXsGroup)
```
drewj-tp:
```suggestion
```

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -900,6 +900,29 @@ def test_getRepresentativeBlocks(self):
...
+        fuelCollection = blockCollectionsByXsGroup[fuelXStype]
+        print(fuelCollection)
```
drewj-tp:
```suggestion
```

armi/physics/neutronics/tests/test_crossSectionManager.py
```diff
@@ -900,6 +900,29 @@ def test_getRepresentativeBlocks(self):
         self.assertIsNone(blocks[0].p.detailedNDens)
         self.assertIsNone(blocks[1].p.detailedNDens)
 
+    def test_checkForUnrepresentedXSIDs(self):
+        blockCollectionsByXsGroup = self.csm.makeCrossSectionGroups()
+        self.csm.createRepresentativeBlocks()
+
+        # set valid flags to something the fuel block would not have to trigger unrepresented block
+        fuelXStype = "AD"
+        blocksWithType = [b for b in self.csm.r.core.getBlocks(Flags.FUEL) if b.getMicroSuffix() == fuelXStype]
```
drewj-tp:
```suggestion
        blocksWithType = [b for b in self.csm.r.core.iterBlocks(Flags.FUEL) if b.getMicroSuffix() == fuelXStype]
```

armi/physics/neutronics/crossSectionGroupManager.py
drewj-tp: Setting `self._unrepresentedXSIDs` here is maybe not necessary anymore, since we handle it entirely w/in `self._checkForUnrepresentedXSIDs`

drewj-tp: You can see how it's not cleared in `createRepresentativeBlocksUsingExistingBlocks` as well

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -1367,6 +1366,32 @@ def _getModifiedReprBlocks(self, blockList, originalRepresentativeBlocks):
...
+        Parameters
+        ----------
+        blockCollectionsByXsGroup: dict[str, BlockCollection]
+            Dict of BlockCollection keyed by the XS group they belong to.
+
+        Notes
+        -----
+        This should be run after `CrossSectionGroupManager._updateEnvironmentGroups()`, which resets
+        `b.p.envGroup` and can result in unrepresented cross section IDs. This is usually invoked
+        as a result of a call to `CrossSectionGroupManager.makeCrossSectionGroups()`
```
drewj-tp: IDK, these don't end up in our API docs (I don't think) so the value of formatting is 🤷 


```suggestion
        This should be run after :meth:`CrossSectionGroupManager._updateEnvironmentGroups`, which resets
        ``b.p.envGroup`` and can result in unrepresented cross section IDs. This is usually invoked
        as a result of a call to :meth:`CrossSectionGroupManager.makeCrossSectionGroups`
```

armi/physics/neutronics/crossSectionGroupManager.py
mgjarrett: Good catch, I'll remove that line

mgjarrett:
> Only real blockers are removing the print statements.
> 
> Formatting comments are pedantic RST formatting nits that might be of questionable value
> 
> > One-line Impact on Requirements: NA
> 
> Is this true? We have `R_ARMI_XSGM_CREATE_XS_GROUPS` that is related to `createRepresentativeBlocks`

I've always read this as N/A unless the requirement itself changes, or the implementation of the requirement is substantially modified. In this case, I did modify some things in `createRepresentativeBlocks` which has the implementation tag for this requirement, but the method is still doing the same thing.

If I had to draft a one-line impact, it would be:

One-line Impact on Requirements: Update the implementation of `R_ARMI_XSGM_CREATE_XS_GROUPS` to always handle unrepresented XS IDs.

PR #2499, opened 2026-04-10
Removing matProps dataSchema

john-science:

The `matProps/dataSchema` package is not used by `matProps`, nor is it imported by anyone downstream.

But I have already found that it is mildly out-of-date, and kind of hard to keep track when and where to update it. It's NOT a bad package, but will cause a lot of hard-to-automate work over time. And, in the end, I think what people are really going to want is matProps documentation. It will be more helpful to people to read the docs to see what a matProps YAML file can look like; our users don't want to wade into the dataSchema format JSON file to try to figure it out themselves.

## SCR Information

Change Type: trivial

One-Sentence Rationale: We should only include code that our users want.

---

## Checklist


opotowsky (approved):
I've clarified offline that this will be documented in a near-future PR. I've also made a ticket to bring this feature back and have it actually be used https://github.com/terrapower/armi/issues/2504. I support deleting unused/broken code in the interim.

PR #2503, opened 2026-04-14
Reducing warnings from settings version and Composite.getComponent

john-science:

This PR reduces the frequency of certain log messages in an ARMI run:

1. The Settings Validation section is optional, so if the user doesn't touch it, they don't want a `warning`.
2. The `Composite.getComponent()` method was raising too many warnings.

close #2496

## SCR Information

Change Type: trivial

One-Sentence Rationale: Too many warnings can cause Alarm Fatigue, and the user base singled out these two warnings as being particularly fatiguing.

---

## Checklist

### DIFF
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -2208,22 +2208,20 @@ class ArmiObject(metaclass=CompositeModelType):
         else:
             return components[0]
 
-    def getComponent(self, typeSpec: TypeSpec, exact: bool = False, quiet: bool = False) -> Optional["Component"]:
+    def getComponent(self, typeSpec: TypeSpec, exact: bool = False, quiet: bool = True) -> Optional["Component"]:
         """
         Get a particular component from this object.
+
+        Be careful with multiple similar names in one object.
 
         Parameters
         ----------
         typeSpec : flags.Flags or list of Flags
             The type specification of the component to return
-
         exact : boolean, optional
             Demand that the component flags be exactly equal to the typespec. Default: False
-
         quiet : boolean, optional
-            Warn if the component is not found. Default: False
-
-        Careful with multiple similar names in one object
+            Log if the component is not found. Default: True
 
         Returns
         -------
@@ -2238,7 +2236,7 @@ class ArmiObject(metaclass=CompositeModelType):
             return results[0]
         elif not results:
             if not quiet:
-                runLog.warning(
+                runLog.extra(
                     f"No component matched {typeSpec} in {self}. Returning None",
                     single=True,
                     label=f"None component returned instead of {typeSpec}",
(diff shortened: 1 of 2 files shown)


armi/reactor/composites.py
```diff
@@ -2238,7 +2236,7 @@ def getComponent(self, typeSpec: TypeSpec, exact: bool = False, quiet: bool = Fa
             return results[0]
         elif not results:
             if not quiet:
-                runLog.warning(
+                runLog.extra(
```
opotowsky: I'd vote debug here.

armi/settings/settingsIO.py
```diff
@@ -120,7 +120,7 @@ def renameSetting(self, name) -> Tuple[str, bool]:
             msg = "\n".join(
                 ["   {}: {}".format(expiredRename, date) for expiredRename, date in expiredCandidates.items()]
             )
-            runLog.warning(
+            runLog.info(
```
opotowsky: I'm ok leaving this as info because I think a user should fix the setting name immediately and then it would go away

armi/settings/settingsIO.py
```diff
@@ -111,7 +111,7 @@ def renameSetting(self, name) -> Tuple[str, bool]:
 
         activeRename = self._activeRenames.get(name, None)
         if activeRename is not None:
-            runLog.warning(f"Invalid setting {name} found. Renaming to {activeRename}.")
+            runLog.extra(f"Invalid setting {name} found. Renaming to {activeRename}.")
```
opotowsky: Why do we have this runLog + the one below? Aren't they accomplishing the same thing?

armi/reactor/composites.py
john-science:
```suggestion
                runLog.debug(
```

john-science: Done!

armi/settings/settingsIO.py
```diff
@@ -228,7 +228,7 @@ def _checkInvalidSettings(self):
         if not proceed:
             raise InvalidSettingsStopProcess(self)
         else:
-            runLog.warning(f"Ignoring invalid settings: {invalidNames}")
+            runLog.info(f"Ignoring invalid settings: {invalidNames}")
```
opotowsky, line 221: This can be kept as info due to the same logic as above: the intention is that the user will fix it immediately (so it's not a useless `info` level statement like many in the ecosystem are)

armi/settings/settingsIO.py
john-science: Okay, I cleaned up this method some, and removed the duplication log logic.

opotowsky (approved):

Issue #2467, opened 2026-03-06
Material.refDens is an instance variable, but should not be


john-science:
There is a variable value in the ARMI `Material` class: 

https://github.com/terrapower/armi/blob/ec8e2b06b4baef63b8eaafee786a6dff540337cc/armi/materials/material.py#L136

> But this being a variable (which is updated in `Material.updateMassFrac()`) can break restarting a run from a DB.

Also, this can cause scientifically invalidate results in a couple ways:

1. In the `refDens` logic, the density of a material is calculated from `refDens` and linear expansion. But if you change the `refDens`, you can invalidate the linear expansion. For instance, you could easily set the `refDens` for a value that corresponds to a temperature outside the valid range of the linear expansion method. (This would be silent, hidden error.)
2. If you change the mass fractions of a material such that you change the reference density, you SHOULD have to change the curve for linear expansion. (Our material references only measure linear expansion in the lab for a know and very specific set of mass fractions that define a that material.)

The easiest solution is to make `refDens` a CLASS attribute, that is not updated in `updateMassFracs()`.

> Also: This `refDens` logic could be moved to a `matProps` `Property`.

john-science:
I think the best solution here is to add a reference density to the YAML data format.

opotowsky:
> I think the best solution here is to add a reference density to the YAML data format.

Agree!

PR #2507, opened 2026-04-15
Refactoring the ARMI materials to be based on matProps

### LINKED ISSUE #2353: Redesign ARMI Materials to be based on data files
After much [discussion](https://github.com/terrapower/armi/discussions/767), we have determined the ARMI materials are hard to use well and they do not sufficiently separate data from code. 

> The best solution seems to be to store ARMI material data in data files, next to the code.

ARMI will still need to support fully in-code, special case materials. But we do not expect that will be the usual case.

## Change Process

Because this change is so large, we are going to break it into three ARMI PRs / steps:

* Step 1: Add `armi.matProps`, a material library that will allow us to define materials in YAML format. [#2453](https://github.com/terrapower/armi/pull/2453)
* Step 2: Refactor the current `armi.materials` to use `matProps`.
* Step 3: Redesign the ARMI materials requirements.

## Side Notes

1. This is a good time for ARMI to transition to measurable density data. "density" and "linear expansion" are things people measure in the real world, and good data can be found. But "pseudo-density" should be moved to be the second-class citizen it is.
2. ARMI's current materials should be moved into YAML files in `armi/testing/materials/`, to further cement that ARMI's materials are not production-worthy.
3. The design should allow for people to put references in every line of the data file.
4. The design should allow for a set collection of curves / properties: density, linear expansion, heat capacity, and so on. But ARMI `Plugins` should allow people to add to that list.
5. Each curve / property for a material should be able to take an arbitrary number of independent variables.
6. Curves / Properties should be able to be defined based on look up tables, if desired.

john-science:

The major change here is that the old ARMI `Material` class is now a subclass of the matProps material class. This means that materials can now be defined in YAML, instead of in pure Python code. This gives a better separation between code and data. Though the old "pure python" approach is still supported, for people who want to add custom logic.

Using the YAML approach to defining a material as data is the preferred approach in ARMI now. And as such, the following materials were 100% converted from Python to YAML:

> Air, Alloy200,  Californium, CaH2, Concrete, Cs, Inconel, Inconel600, InconelPE16, Lead, LeadBismuth, Magnesium, MOX, NaCl, Potassium, Sodium, TZM, Void, and ZnO

And the following materials were converted from pure Python to nearly pure YAML:

> Be9, Graphite, HastelloyN, HT9, Inconel625, Inconel800, InconelX750, Lithium, MgO, Sc2O3, Thorium, ThoriumOxide / ThO2, ThU, Uranium, UraniumOxide / UO2, Y2O3, Zr

(NOTE: Most of these "mostly YAML" materials are only not pure Python because they use a `self.refDens` property that is not supported by matProps. I could easily use matProps to support that property and fix this issue. But I did not want to make this PR even bigger.)

This PR also makes some small improvements to matProps:

* added back in support for simple "constant" functions, that do not use sympy.
* added support for a "void" material type, needed by ARMI
* added new material types "SimpleSolid" and "Composite", needed by ARMI

This PR also greatly improves the quality of the materials included in ARMI. All the ARMI materials used to come with a disclaimer that they were only of "test" and not "production" quality. Well, @dpham-materials and I reviewed the materials. David even found better sources to attribute the data in many cases. And I just straight deleted some materials and curves that I could not personally validate. So the ARMI materials a lot more trustworthy now.

close #2353

## SCR Information

Change Type: features

One-Sentence Rationale: The ARMI material class used to be flexible, but also problem in a number of ways; here we provide much better support to materials.

One-line Impact on Requirements: This is a redesign for the implementation of the materials requirements: R_ARMI_MAT_PROPERTIES / I_ARMI_MAT_PROPERTIES , R_ARMI_MAT_FRACS / I_ARMI_MAT_FRACS, and R_ARMI_MAT_FLUID / I_ARMI_MAT_FLUID. Removed now defunct tests T_ARMI_MAT_PROPERTIES2 and T_ARMI_MAT_PROPERTIES3.

---

## Checklist


armi/materials/tests/test_air.py
```diff
@@ -195,9 +195,10 @@ def test_pseudoDensity(self):
 
         for Tk, densKgPerM3 in zip(REFERENCE_Tk, REFERENCE_DENSITY_KG_PER_M3):
             if Tk < 2400:
-                error = math.fabs((air.pseudoDensityKgM3(Tk=Tk) - densKgPerM3) / densKgPerM3)
+                print(air.pseudoDensity(Tk=Tk), air.pseudoDensity(Tk=Tk) * 1000.0, densKgPerM3)
```
john-science:
```suggestion
```

armi/materials/__init__.py
```diff
@@ -39,9 +40,10 @@
 from typing import List
 
 from armi.materials.material import Material
+from armi.materials.pureYaml import Void  # noqa: F401
 
-# This will frequently be updated by the CONF_MATERIAL_NAMESPACE_ORDER setting
-# during reactor construction (see armi.reactor.reactors.factory).
+# This will frequently be updated by the CONF_MATERIAL_NAMESPACE_ORDER setting during reactor construction (see
+# armi.reactor.reactors.factory).
```
john-science:
```suggestion
# This will frequently be updated by the CONF_MATERIAL_NAMESPACE_ORDER setting
# during reactor construction (see armi.reactor.reactors.factory).
```

armi/materials/__init__.py
```diff
@@ -54,60 +56,56 @@ def setMaterialNamespaceOrder(order):
...
-        is chosen, i.e. earlier namespaces in the list have precedence.
+        An ARMI application will need materials. Materials can be imported from any code the application has access to,
+        like plugin packages. This leads to the situation where one ARMI application will want to import multiple
+        collections of materials. To handle this, ARMI keeps a list of material namespaces. This is an ordered list of
+        importable packages that ARMI can search for a particular material by name.
+
+        This automatic exploration of an importable package saves the user the tedium have having to import or include
+        hundreds of materials manually somehow. But it comes with a caveat; the list is ordered. If two different
+        namespaces in the list include a material with the same name, the first one found in the list is chosen, i.e.
+        earlier namespaces in the list have precedence.
```
john-science:
```suggestion
        An ARMI application will need materials. Materials can be imported from
        any code the application has access to, like plugin packages. This leads to
        the situation where one ARMI application will want to import multiple
        collections of materials. To handle this, ARMI keeps a list of material
        namespaces. This is an ordered list of importable packages that ARMI
        can search for a particular material by name.

        This automatic exploration of an importable package saves the user the
        tedium have having to import or include hundreds of materials manually somehow.
```

pyproject.toml
```diff
@@ -325,6 +325,44 @@ armi = [
     "physics/neutronics/tests/rzmflxYA",
     "resources/burn-chain.yaml",
     "resources/elements.dat",
+    "resources/materials/Air.yaml",
```
opotowsky: Can you replace this with `resources/materials/*.yaml` or `resources/materials/*`? I see wild cards being used above so assumed it was possible.

armi/reactor/blueprints/blockBlueprint.py
```diff
@@ -266,13 +266,15 @@ def _getMaterialModsFromBlockChildren(self, c: Composite) -> Set[str]:
...
             for materialParentClass in material.__class__.__mro__:
-                # we must loop over parents as well, since applyInputParams
-                # could call to Parent.applyInputParams()
+                # we must loop over parents as well, since applyInputParams could call to Parent.applyInputParams()
                 if issubclass(materialParentClass, Material):
                     perChildModifiers.update(signature(materialParentClass.applyInputParams).parameters.keys())
-        # self is a parameter to methods, so it gets picked up here
-        # but that's obviously not a real material modifier
+                    if materialParentClass.enrichedNuclide:
+                        perChildModifiers.add(f"{materialParentClass.enrichedNuclide}_wt_frac")
```
opotowsky, line 273: I'm not 100% sure why this is needed. Commenting here to come back to this or get an explanation from John.

armi/reactor/components/component.py
```diff
@@ -1400,11 +1392,9 @@ def density(self) -> float:
...
         density = composites.Composite.density(self)
 
-        if not density and not isinstance(self.material, void.Void):
-            # possible that there are no nuclides in this component yet. In that case,
-            # defer to Material. Material.density is wrapped to warn if it's attached
-            # to a parent. Avoid that by calling the inner function directly
-            density = self.material.density.__wrapped__(self.material, Tc=self.temperatureInC)
+        if not density:
+            # It is possible that there are no nuclides in this component yet. In that case, we defer to the Material.
+            density = self.material.density(Tc=self.temperatureInC)
```
opotowsky, line 1397: Comment to come back and understand this better.

armi/materials/tests/test_thoriumOxide.py
```diff
@@ -29,14 +29,11 @@ def setUp(self):
         self.ThoriumOxide_TD_frac = ThoriumOxide()
         self.ThoriumOxide_TD_frac.applyInputParams(TD_frac=0.4)
 
-    def test_theoretical_pseudoDensity(self):
+    def test_pseudoDensity(self):
         ref = self.mat.pseudoDensity(500)
 
         reduced = self.ThoriumOxide_TD_frac.pseudoDensity(500)
-        self.assertAlmostEqual(ref * 0.4, reduced)
+        self.assertAlmostEqual(ref * 0.16, reduced)
```
opotowsky, line 36: Did the default TD_frac change?

armi/materials/tests/test_uZr.py
```diff
@@ -78,28 +78,6 @@ def test_cache(self):
...
-        dens = self.mat.density(500)
-        densKgM3 = self.mat.densityKgM3(500)
-        self.assertEqual(dens * 1000.0, densKgM3)
-
-    def test_pseudoDensityKgM3(self):
-        """Test the pseudo density for kg/m^3.
-
-        .. test:: Test the material base class has temp-dependent 2D density.
-            :id: T_ARMI_MAT_PROPERTIES3
-            :tests: R_ARMI_MAT_PROPERTIES
```
opotowsky, line 97: Would you want to call out these 2 deleted test tags in the requirements affected section in the PR descrip? I agree they are extraneous.

armi/materials/tests/test_uZr.py
opotowsky, line 96: why is this deleted sometimes and not others? (I did not survey for consistency here, just noticed this one)

alexhjames:
@john-science maybe I'm misunderstanding how `refDens` works, but wouldn't the equivalent `matProps` behavior be for you to set the reference temperature on the density property (you can assign a reference temperature to any property not just thermal expansion) and then the `refDens` would be the density property evaluated at the reference temperature?

armi/materials/b4c.py
```diff
@@ -170,16 +168,6 @@ def getMassEnrichmentFromNumEnrich(
 
         return b10NumFrac * b10AtomicMass / (b10NumFrac * b10AtomicMass + (1.0 - b10NumFrac) * b11AtomicMass)
 
-    def pseudoDensity(self, Tk: float = None, Tc: float = None) -> float:
```
opotowsky, line 173: haven't looked into this too much but it why are we deleting pseudo density here but not, eg, in water?

armi/matProps/constantFunction.py
```diff
@@ -0,0 +1,55 @@
...
+        Parameters
+        ----------
+        mat: Material
+            Material object with which this ConstantFunction is associated
+        prop: Property
+            Property that is represented by this ConstantFunction
+        """
+        super().__init__(mat, prop)
+        self.value = None
+        """Constant value that is returned by ConstantFunction."""
```
opotowsky: does this do anything here when its inside the init?

armi/materials/tests/test_lithium.py
```diff
@@ -11,61 +11,43 @@
...
         self.mat = Lithium()
 
-        self.Lithium_LI_wt_frac = Lithium()
-        self.Lithium_LI_wt_frac.applyInputParams(LI6_wt_frac=0.5)
-
-        self.Lithium_LI6_wt_frac = Lithium()
-        self.Lithium_LI6_wt_frac.applyInputParams(LI6_wt_frac=0.6)
-
-        self.Lithium_both = Lithium()
-        self.Lithium_both.applyInputParams(LI6_wt_frac=0.8)
```
opotowsky, line 36: are we OK doing away with the special input params that were originally implemented for this material?

armi/resources/materials/Air.yaml
```diff
@@ -0,0 +1,83 @@
+file format: 3.0
+# Copyright 2019 TerraPower, LLC
```
opotowsky, line 2: All the copyrights are 2019, FYI

armi/materials/material.py
```diff
@@ -18,72 +18,44 @@
...
 from armi.utils import densityTools
 from armi.utils.units import getTc, getTk
 
-# globals
+# Setting this to False will globally allow you to access material properties outside their defined bounds. This only
+# works for material properties defined purely in Python.
 FAIL_ON_RANGE = True
 
-
-def parentAwareDensityRedirect(f):
```
opotowsky, line 38: so we are guaranteed to have material.density match component.density now? I actually wasn't expecting that with this PR! We can chat about it offline so I can understand better.

armi/materials/material.py
```diff
@@ -438,94 +446,80 @@ def pseudoDensity(self, Tk: float = None, Tc: float = None) -> float:
...
+
+        Since refDens is specified at the material-dep reference case, we don't need to specify the reference
+        temperature. It is already consistent with linearExpansion Percent.
         - p*(dp/p(T) + 1) =p*( p + dp(T) )/p = p + dp(T) = p(T)
         - dp/p = (1-(1 + dL/L)**3)/(1 + dL/L)**3
         """
+        # try the YAML file first
+        if hasattr(self, "rho") and self.rho is not None:
+            Tc = getTc(Tc, Tk)
+            # matProps does density is in kg/m3, and this method is in g/cm3
```
opotowsky, line 481: are the yamls in kg/m3 and then we convert to g/cm3?

Either way: grammar edit. But I think we should clear up what exactly is giving density in kg/m3

armi/materials/material.py
```diff
@@ -827,28 +684,30 @@ class FuelMaterial(Material):
...
-    class1_wt_frac = None
-    class1_custom_isotopics = None
-    class2_custom_isotopics = None
+    def __init__(self):
+        Material.__init__(self)
+        self.class1_wt_frac = None
+        self.class1_custom_isotopics = None
+        self.class2_custom_isotopics = None
 
     def applyInputParams(
```
opotowsky, line 693: when did we decide that editable input params would be snake case? Not requesting a change, but it's odd given the camel case choice the repo made.

opotowsky (commented):
First pass review

pyproject.toml
john-science: Actually, yes, I can. 

I guess I have been waffling on which is best practice: the convenience of wild cards or the specificity of listing everything manually. 

I should pick one though, you're right. One sec.

john-science: Okay, let me switch this to using wild cards. That would reduce a lot of busy work as people add and remove material YAML files. Good idea! Thanks!

Done!

armi/reactor/blueprints/blockBlueprint.py
john-science: Oh, right, I'm sorry. That is not clear.

The old `Material.applyInputParams()` was empty by default:

https://github.com/terrapower/armi/blob/main/armi/materials/material.py#L317-L319

And a BUNCH of different materials subclassed this method and did thing that were all essentially the same, like setting "U238_wt_frac" to a value. And that ended up being a lot of the reason why our half/most of our materials couldn't be converted to YAML. 

So I solved this by realizing that "{X}_wt_frac" was an easy solution:

https://github.com/terrapower/armi/blob/mat_refactor/armi/materials/material.py#L337

And that alllmost solved the whole situation, and allowed a dozen materials become YAMLS. But I had to add a couple of small "enriched nuclide" hooks in the code. Like this one.

armi/reactor/components/component.py
john-science: So, the logic here is that the Component might not have enough information to calculate it's own density. (It's not important, but the Component might not have any nuclides defined, it's quite common when building a reactor and things are still in flux.)

In that situation, we fall back on just asking the `Material` what the density is.

But that used to be a LOT more complicated than it is now. Look at the old version, there was this Python-level `__wrapped__` logic that was hard to follow. And we had to make an exception for `Void`. 

All that is simpler now, you just ask for `Material.density(Tc)`.

armi/materials/material.py
john-science: Sadly, I still can't guarantee that while the reactor is being built.

This is not as big a change as all that.  I just reduced a ton of overly-complicated code.  Instead of this whole __wrapped__ nonsense, I simplified the logic that calls density when-and-only-when the problem this is trying to solve occurs. (I have already pointed out one such case, when you asked a comment on blockBlueprint.py.)

BUT the benefit of my code simplification is this right here was REALLY slow. The new code is much faster, because it can skip all this namespace hacking.

armi/materials/tests/test_uZr.py
john-science: Oh, sure, I can do that. I was probably just not thinking enough about requirements because that's really the next PR.

Done.

armi/materials/tests/test_uZr.py
john-science: Okay, so "propertyValidTemperature" is part of the old logic for defining a material in Python. The YAML files now have a different (better) solution to this.

So any material that is defined in Python now doesn't have this field anymore, so we can / should remove this test.

The only material where I might not have removed these are the ones that are still in Python. (UZr is still Python.)

armi/materials/b4c.py
john-science: Oh! Good find!

In ARMI, fluids do not expand (linear expansion is zero). So `density()` and `pseudoDensity()` are always zero.

But the `Water` subclasses used to have all this tricky logic they don't need any more. I can simplify that a lot now.  I will go delete those from the various `Water`s.

Thanks!

Done!

armi/matProps/constantFunction.py
john-science: Ah, no, you're right.

Fixed!

armi/materials/tests/test_lithium.py
john-science: hmmm...

Okay, I vaguely remember removing these because they aren't used in but one unit test.

And I THOUGHT the base class (`AbstractMaterialTest`) took care of testing this logic for us. But maybe it doesn't?  Lemme look.

john-science: Okay, I added it back in. (With some improvements.)

Done.

armi/resources/materials/Air.yaml
john-science: Yeah... I was copying over from the Python files. I guess I was thinking this was just a translation and not a truly "new" thing.

But... I didn't put much thought into that. And I'm not going to die on this hill.

What do you think?

armi/materials/material.py
john-science: Yeah, the units of matProps YAML do not 100% agree with the units ARMI wants.

And we could change this, but that would mean talking to people on your materials team and telling them to change all their data files. Which they COULD do, but it would take time and I don't want to start that fight in the middle of this PR.

Certainly, that will be documented in the matProps User Manual, when we bring that into ARMI in the next PR.

Do you not think this is clear enough? Thoughts?

```python
# matProps does density is in kg/m3, and this method is in g/cm3
```

armi/materials/material.py
john-science: LOL

Well those are class attributes and not instance variables:

https://github.com/terrapower/armi/blob/ee1993987db7ed9201322c0e5657c141ed091494/armi/materials/material.py#L830-L832

So, we are okay with snake case for global variables and class attributes, I think.

Listen, I'm not going to die on this hill. A quick look at the history says these names are 7 years old, and predate my time on the project.  They are referenced in downstream projects, so I think we would want a separate ticket / PR for that work. But if you want that ticket, let's open it. NBD.

john-science:
> @john-science maybe I'm misunderstanding how `refDens` works, but wouldn't the equivalent `matProps` behavior be for you to set the reference temperature on the density property (you can assign a reference temperature to any property not just thermal expansion) and then the `refDens` would be the density property evaluated at the reference temperature?

Well, yes, there is an equivalent approach in matProps.

But there are downstream MANY downstream projects that would get diffs in their results with even the tiniest, smallest change in this logic.

For instance, for better (or definitely) worse, this `refDens` is changed by some downstream codes. Which is just a nightmare, because that information is not recorded in the ARMI database. But I can't _make_ people change, so here is a ticket: https://github.com/terrapower/armi/issues/2467

armi/materials/tests/test_uZr.py
opotowsky: ah ok. So this is still a valid thing for py materials but not yaml materials. Got it.

armi/resources/materials/Air.yaml
opotowsky: Meh, I don't have too many opinions on this, I just thought you'd care about the year. but I see your logic!

armi/materials/material.py
opotowsky: Yea I don't think we should change units. That's a lot of mess for very little gain (and possibly no gain / all mess). 

But the kg/m3 units are coming from the yaml files (or someones mat DB), and mat props is actually whats spitting it out in g/cm3, so we just need to correct the comment to say that

armi/materials/material.py
opotowsky: Yea it's a lot of mess for little gain to do something about it now :-)

armi/materials/material.py
```diff
@@ -120,54 +85,52 @@ def __init_subclass__(cls) -> None:
...
+
         self.parent = None
         self.massFrac = {}
         self.refDens = 0.0
         self.theoreticalDensityFrac = 1.0
         self.cached = {}
         self._backupCache = None
-        self._name = self.__class__.__name__
+
+        if self.name is None:
```
alexhjames, line 114: I'm seeing some weirdness surrounding a couple material names. For example `resolveMaterialClassByName("Alloy718")` works and gives you the material back but the name of the material you got back is `Alloy_718` because that is what the matProps YAML name was. This causes problems in a few places for us, although they are fixable if needed. It would be easiest though if the name of the material continued to align with the class name like it used it, but I'm not sure if that causes other problems.

alexhjames: an example of this problem is if you do something like this it will error:

```
name = component.getProperties().getName()
resolveMaterialClassByName(name) # This will error because the name property doesn't match the class
```

armi/materials/material.py
opotowsky: Before, when component.density didn't work and we fell back to material.density, this emitted a warning. Is that important? (I don't want unneeded warnings but also I'm not sure if it's OK to delete this one)

I'll trust you on the speedup! Did you do any testing or did you just recall that adding this slowed things down?

armi/materials/water.py
```diff
@@ -11,18 +11,16 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Basic water material.
+"""
+Basic water materials: SaturatedSteam and SaturatedWater.
 
-The data in this file exists for testing and demonstration purposes only. Developers of ARMI applications can refer to
-this file for a fully worked example of an ARMI material. And this material has proven useful for testing. The data
-contained in this file should not be used in production simulations.
+This is an example of a material implement purely in Python, without any matProps YAML inputs.
```
opotowsky:
```suggestion
This is an example of a material implemented purely in Python, without any matProps YAML inputs.
```

armi/materials/material.py
```diff
@@ -120,54 +85,52 @@ def __init_subclass__(cls) -> None:
     """Name of enriched nuclide to be interpreted by enrichment modification methods"""
 
     modelConst = {}
-    """Constants that may be used in interpolation functions for property lookups"""
+    """Constants that may be used in interpolation functions for property lookups. Recommendation: don't use this."""
 
     propertyValidTemperature = {}
     """Dictionary of valid temperatures over which the property models are valid in the format
     'Property Name': ((Temperature_Lower_Limit, Temperature_Upper_Limit), Temperature_Units)"""
```
opotowsky, line 92: So as I understand it we use this for python materials and not for yaml materials. Is this what you expect to see?

Image

For example, one of those tests in test_materials.py is californium and that is a yaml material.

armi/resources/materials/Void.yaml
```diff
@@ -12,19 +13,32 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Tantalum.
+# Void / Vacuum
```
opotowsky, line 16: I do not see the void.py file as a deletion in the diffs. Can you check that it is deleted? Super odd git is showing this as a rename of tantalum.py but I do not claim to know the magics

opotowsky: OH nope nevermind it's showing void.py -> magnesium.yaml

I will not stare into the sun trying to understand this.

armi/materials/material.py
```diff
@@ -18,72 +18,44 @@
...
-            )
-        return f(self, *args, **kwargs)
-
-    return inner
+# Need for an memoization optimization to cache YAML-mased materials
+_YAML_MATERIALS = {}
 
 
-class Material:
+class Material(MatPropsMaterial):
```
alexhjames, line 38: Non-YAML based materials appear to be missing the `rho` property even though they have `density`. This property should be redirected to `density` for those materials. Similar thought for other properties that are defined with both legacy ARMI and legacy mat_props function names: `linear expansion percent` vs `alpha_mean` and `heat_capacity` vs `k`.

armi/resources/materials/Void.yaml
john-science: So, `armi/materials/Void.py` got moved to `armi/resources/materials.Void.yaml`.

And that's exactly as simple as it sounds.

This Magnesium / Tantulum / Void confusion is 100% the GitHub UI, and nothing to do with git or reality.  

Sorry.

armi/materials/material.py
john-science: If you do not create a YAML file, you will not be able to call the YAML properties.

There is nothing I can do about that.

The correlation between things like `mat.density()` and `mat.rho()` needs-must be one directional. I can't guarantee correctness if I try to reverse-engineer `mat.rho()` from `mat.density()`. 

Sorry, this is just not possible, there are too many limitations in the system. I can only go one direction.

Though, `resolveMaterialClassByName()` will always return the ARMI `Material` subclass, which should help your underlying concern, I think.

alexhjames: Yes, `resolveMaterialClassByName()` changes solve this concern.

armi/materials/__init__.py
```diff
@@ -63,51 +78,111 @@ def setMaterialNamespaceOrder(order):
...
+        # tests this is a valid material file AND preps for adding to matProp registry
+        mat = Material()
+        try:
+            mat.loadFile(yamlPath)
+        except Exception:
+            continue
+
+        # TODO: print(os.path.basename(yamlPath))
+        addMaterial(yamlPath, mat)
+        addYamlMaterialToThisNamespace(yamlPath, overwriteExisting=overwriteExisting)
```
alexhjames, line 147: Right after this loop, there needs to be something to the effect of: 

`matProps.loadedRootDirs.append(dirPath)`

The `addMaterial` call stores it in the matProps material dictionary so the loaded directories needs an update.

armi/materials/__init__.py
```diff
@@ -188,19 +256,21 @@ def resolveMaterialClassByName(name: str, namespaceOrder: List[str] = None):
...
         modPath, clsName = name.split(":")
         mod = importlib.import_module(modPath)
         return getattr(mod, clsName)
 
+    # 2. Try to import the material from a namespace defined above
     namespaceOrder = namespaceOrder or _MATERIAL_NAMESPACE_ORDER
     for namespace in namespaceOrder:
         mod = importlib.import_module(namespace)
         if hasattr(mod, name):
+            # TODO: Does this work? I appear to be returning modules, not just material classes.
```
alexhjames: I am observing the same thing about importing modules. This occurs if one namespace has a material that is the same as a module name in a different (higher preference) namespace. This affect is independent of the latest changes, it just was never triggered before.

I think you need these lines of code: 
```
mod = importlib.import_module(namespace)
materialsList = inspect.getmembers(mod, lambda c: inspect.isclass(c))
materialsList = [material[0] for material in materialsList]
if name in materialsList:
    return getattr(mod, name)
```

armi/materials/__init__.py
```diff
@@ -63,51 +78,111 @@ def setMaterialNamespaceOrder(order):
...
+    name = Path(yamlPath).stem
 
-    This allows devs to use ``from armi.materials import HT9``
+    # If a class with this name already exists in the package, continue
+    if not overwriteExisting and name in globals():
+        return
 
-    This can be used in plugins for similar purposes.
+    # Build a custom YAML material class and add it to this package
+    code = """
```
alexhjames: Optional, it is the same net result, but these couple lines of code may be cleaner: 

```
MaterialClass = type(name, (Material,), {"YAML_PATH": yamlPath})
globals()[name] = MaterialClass
```

armi/materials/__init__.py
john-science: Is that used? Is it necessary?

armi/materials/__init__.py
john-science: Sure, I love it. Done.

armi/materials/__init__.py
john-science: Done

armi/materials/__init__.py
john-science: I don't want to break the existing functionality, and I have no idea how to unit test this change. Seems tricky.

So I am reluctant here. But I'm looking at it.

Also, I don't understand your zero index here: `materialsList = [material[0]...`

alexhjames: This code grabs the module like previous, then filters its members to just the ones that are classes. The list of filtered members that are returned look like this where it is a list of tuples that are class name and full namespace. 
image
So the zero index is just grabbing the name of the class out to compare against the thing requested. 

In the old code that just looked at the attributes of the modules, I was getting modules back because something high in the namespace order would have a file that is say `air.py` with a class `Air` inside it. So the module would have attributes that are a child module `air` as well as a class `Air`. Then some less preferential namespace would have a material called `air`. `resolveMaterialClassByName` would see that `air` exists on the higher preference namespace (as a module) and return that instead of the class with that name lower in the namespace preference. I think we didn't run into this before because we never had a class named the same as a module but this is happening with one or two matProps materials.

john-science: Done.

armi/materials/material.py
john-science: Boy, those tests seem pretty goofy to me if they are all zero.

The only material in ARMI that still uses that little feature is B4C. I think I'm going to refactor this testing to demonstrate that fact.

john-science: Done.

armi/materials/water.py
john-science: Fixed

armi/materials/material.py
john-science: Okay, so matProps says `rho` is in kg/m3:

https://github.com/terrapower/armi/blob/b643fbcae5e14febf481e8edf3cc10827a06e615/armi/matProps/prop.py#L50

And ARMI Materials say `density` is in g/cm3:

https://github.com/terrapower/armi/blob/b643fbcae5e14febf481e8edf3cc10827a06e615/armi/materials/material.py#L466

But you are saying this: "and mat props is actually whats spitting it out in g/cm3".  I don't think that's true. Am I missing something?

armi/materials/material.py
john-science: Oh, I accidentally ended up doing performance testing on this branch.

My first working version of this branch TRIPLED the run time of the ARMI unit tests. I got real worried I'd going down the wrong path. BUT, I was following my order of operations:

1. Make it work.
2. Make it right.
3. Make it fast.

So I spent a while trying to find ways to speed up the code. This was one of the performance improvements I found. Eventually, I realized `sympy` was the problem. When building a reactor from blueprints, every time we needed a material, we were going through with `sympy` and re-parsing the YAML file. Which involves file IO, and is very slow.

Now, every time we parse a YAML file with `sympy`, I save off the result in a cache, and that solves  my performance problem:

https://github.com/terrapower/armi/blob/b643fbcae5e14febf481e8edf3cc10827a06e615/armi/materials/material.py#L35

https://github.com/terrapower/armi/blob/b643fbcae5e14febf481e8edf3cc10827a06e615/armi/materials/material.py#L94-L102

Now, this branch actually runs ARMI unit tests faster than `main`.

armi/materials/material.py
john-science: Done!

armi/materials/tests/test_thoriumOxide.py
opotowsky: ping @john-science

john-science: Oh!
Okay, let me look.

john-science: Okay, fixed. I removed `SimpleSolid` as a base class for `ThoriumOxide`: https://github.com/terrapower/armi/commit/65df22d3b5a900a637d48158a737c5a97e3440cc

opotowsky (approved):

PR #2510, opened 2026-04-17
Improving error message in settings validation

### LINKED ISSUE #2509: Improve failure mode when user selects "no" during settings validation
I've had some user-reported confusion over opting "no" during settings validation. When a case is submitted and the following scenario occurs:

```
[prmt] INSPECTOR: blah blah grumble blah
[prmt] Continue? (YES, NO CANCEL, Y, N)
```

If the user chooses no, then you get a sys exit with a keyboard interrupt. Can we have this exit more gracefully so the user knows they have done the right thing?

john-science:

We have had a request to improve the error message when a user cancels a run during settings validation.

close #2509

## SCR Information

Change Type: trivial

One-Sentence Rationale: Error messages are only useful if they are clear and understandable.

---

## Checklist

### DIFF
--- a/armi/settings/settingsValidation.py
+++ b/armi/settings/settingsValidation.py
@@ -110,7 +110,7 @@ class Query:
                             self.corrected = True
                         self._passed = True
                     except RunLogPromptCancel:
-                        raise KeyboardInterrupt("You have cancelled the submission.")
+                        raise SystemExit("You have cancelled the submission.")
                 else:
                     try:
                         continueSubmission = prompt(
@@ -121,9 +121,9 @@ class Query:
                             "CANCEL",
                         )
                         if not continueSubmission:
-                            raise KeyboardInterrupt
+                            raise SystemExit("You have cancelled the submission.")
                     except RunLogPromptCancel:
-                        raise KeyboardInterrupt("You have cancelled the submission.")
+                        raise SystemExit("You have cancelled the submission.")
             except RunLogPromptUnresolvable:
                 self.autoResolved = False
                 self._passed = True
--- a/armi/settings/settingsValidation.py
+++ b/armi/settings/settingsValidation.py
@@ -109,8 +109,8 @@ class Query:
                             self.correction()
                             self.corrected = True
                         self._passed = True
-                    except RunLogPromptCancel as ki:
-                        raise KeyboardInterrupt from ki
+                    except RunLogPromptCancel:
+                        raise KeyboardInterrupt("You have cancelled the submission.")
                 else:
                     try:
                         continueSubmission = prompt(
--- a/armi/settings/settingsValidation.py
+++ b/armi/settings/settingsValidation.py
@@ -13,13 +13,11 @@
 # limitations under the License.
 
 """
-A system to check user settings for validity and provide users with meaningful
-suggestions to fix.
+A system to check user settings for validity and provide users with meaningful suggestions to fix.
 
-This allows developers to define a rich set of rules and suggestions for user settings.
-These then pop up during initialization of a run, either on the command line or as
-dialogues in the GUI. They say things like: "Your ___ setting has the value ___, which
-is impossible. Would you like to switch to ___?"
+This allows developers to define a rich set of rules and suggestions for user settings. These then pop up during
+initialization of a run, either on the command line or as dialogues in the GUI. They say things like: "Your ___ setting
+has the value ___, which is impossible. Would you like to switch to ___?"
 """
 
 import itertools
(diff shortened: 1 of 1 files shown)


armi/settings/settingsValidation.py
```diff
@@ -105,45 +98,44 @@ def resolve(self):
...
                             "NO_DEFAULT",
                             "CANCEL",
                         )
-                        if make_correction:
+                        if makeCorrection:
                             self.correction()
                             self.corrected = True
                         self._passed = True
                     except RunLogPromptCancel as ki:
                         raise KeyboardInterrupt from ki
```
opotowsky: I think we want the message here too?

armi/settings/settingsValidation.py
```diff
@@ -105,45 +98,44 @@ def resolve(self):
...
                             "YES_NO",
                             "NO_DEFAULT",
                             "CANCEL",
                         )
-                        if not continue_submission:
+                        if not continueSubmission:
                             raise KeyboardInterrupt
-                    except RunLogPromptCancel as ki:
-                        raise KeyboardInterrupt from ki
+                    except RunLogPromptCancel:
```
opotowsky, line 125: so this is effective for a CANCEL response but is it for the NO response? Would this also be needed on line 124?

john-science: The exception on line 24 is never seen. It is hidden inside a try/except.

armi/settings/settingsValidation.py
john-science: whoops

armi/settings/settingsValidation.py
opotowsky: So this clearly happens if a user types "cancel", but what if they type "no"?

armi/settings/settingsValidation.py
john-science: Fixed.

armi/settings/settingsValidation.py
john-science: What would you like it to say there?

I guess I was thinking that the user "cancelled" the run either way, if we hit this error and it halts. That's all. I'm not married to it.

opotowsky: Oh I have no issue with the message, it's perfectly fine. I just want to know if this is ALSO the code pathway for when a user types "no" (it's not clear to me if no==cancel in terms of the code)

opotowsky: let me pull this in and test it out

opotowsky: Ok I was right:

When I typed `CANCEL` I got:
image

And when I typed `NO` I got:
image

armi/settings/settingsValidation.py
```diff
@@ -105,45 +98,44 @@ def resolve(self):
...
                         )
-                        if make_correction:
+                        if makeCorrection:
                             self.correction()
                             self.corrected = True
                         self._passed = True
-                    except RunLogPromptCancel as ki:
-                        raise KeyboardInterrupt from ki
+                    except RunLogPromptCancel:
+                        raise KeyboardInterrupt("You have cancelled the submission.")
```
opotowsky:
```suggestion
                        raise SystemExit("You have cancelled the submission.")
```

armi/settings/settingsValidation.py
```diff
@@ -105,45 +98,44 @@ def resolve(self):
...
                             "NO_DEFAULT",
                             "CANCEL",
                         )
-                        if not continue_submission:
+                        if not continueSubmission:
                             raise KeyboardInterrupt
-                    except RunLogPromptCancel as ki:
-                        raise KeyboardInterrupt from ki
+                    except RunLogPromptCancel:
+                        raise KeyboardInterrupt("You have cancelled the submission.")
```
opotowsky:
```suggestion
                        raise SystemExit("You have cancelled the submission.")
```

armi/settings/settingsValidation.py
```diff
@@ -105,45 +98,44 @@ def resolve(self):
...
+                        continueSubmission = prompt(
                             "INSPECTOR: " + self.statement,
                             "Continue?",
                             "YES_NO",
                             "NO_DEFAULT",
                             "CANCEL",
                         )
-                        if not continue_submission:
+                        if not continueSubmission:
                             raise KeyboardInterrupt
```
opotowsky:
```suggestion
                            raise SystemExit("You have cancelled the submission.")
```

opotowsky (commented):
Ok the suggested changes in this review give me what I want. 

No:
Image

Cancel:
Image

Using `SystemExit` instead of `KeyboardInterrupt` exits nicely and avoids the traceback that was confusing users. The traceback doesn't provide any useful information (the settings validator does that). So I think this is a defensible change!

opotowsky (approved):

PR #2515, opened 2026-04-24
Adding CircleHoledCircle shape

### LINKED ISSUE #2512: Create CircleHoledCircle Shape
We have a `HoledHexagon` (single hex with >= 1 circle holes)

And we have a `HexHoledCircle` (single circle with  1 hex hole)

It would be good to add a `HoledCircle` (single circle with >= 1 circle holes)

john-science:

By user request, a new complex shape was added to ARMI.

This shape is a circle with N circles cut out of it.

close #2512

## SCR Information

Change Type: features

One-Sentence Rationale: This new shape could be used for a lot of things in pin-type reactor design.

---

## Checklist

### DIFF
--- a/armi/reactor/components/complexShapes.py
+++ b/armi/reactor/components/complexShapes.py
@@ -138,6 +138,9 @@ class CircleHoledCircle(basicShapes.Circle):
             modArea=modArea,
         )
 
+        assert (holeRadFromCenter + holeOD) < od, "Holes extend past outer diameter"
+        assert 2 * holeRadFromCenter * math.sin(math.pi / nHoles) > holeOD, "Internal holes overlap"
+
     def getComponentArea(self, cold=False, Tc=None):
         """Computes the area for the circle with n circular holes."""
         od = self.getDimension("od", cold=cold, Tc=Tc)
--- a/armi/reactor/components/complexShapes.py
+++ b/armi/reactor/components/complexShapes.py
@@ -154,7 +154,7 @@ class CircleHoledCircle(basicShapes.Circle):
         """
         For the special case of only one single hole, returns the diameter of that hole.
 
-        For any other case, returns 0.0 because an "circle inner diameter" becomes undefined.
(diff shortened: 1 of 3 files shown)


armi/reactor/components/complexShapes.py
```diff
@@ -95,6 +95,73 @@ def getCircleInnerDiameter(self, Tc=None, cold=False):
...
+        innerArea = nHoles * math.pi * ((holeOD / 2.0) ** 2)
+
+        mult = self.getDimension("mult")
+        return mult * (outerArea - innerArea)
+
+    def getCircleInnerDiameter(self, Tc=None, cold=False):
+        """
+        For the special case of only one single hole, returns the diameter of that hole.
+
+        For any other case, returns 0.0 because an "circle inner diameter" becomes undefined.
```
mgjarrett:
```suggestion
        For any other case, returns 0.0 because a "circle inner diameter" becomes undefined.
```

mgjarrett (approved):
I only noticed one minor typo in a docstring.

I have two additional ideas, both of which are optional:

1. It might be worthwhile to discourage (warning) or disallow (error) the user from specifying a `CircleHoledCircle` with `nHoles: 1`. Such a shape can be modeled using the existing `Circle` class with an `id`. Allowing two different types of shape to model the same thing feels like it could be error prone.
2. Should there be some kind of check that the geometry is physically meaningful? For example, we could enforce the following
  - holes don't extend past the outer diameter: `holeRadFromCenter + holeOD / 2.0 < od / 2.0`.
  - holes don't overlap with the origin: `holeRadFromCenter > holeOD / 2.0`
  - Enforce non-overlapping holes. Maybe something like "chord length between the center of two holes" > `holeOD`, e.g., `2 * holeRadFromCenter * sin(pi / nHoles) > holeOD`

For (2), we're not currently doing this for the `HoledHexagon` shape, so it's probably not necessary to add it here. But it could make this class more robust against non-physical component definitions.

PR #2521, opened 2026-05-05
Move material source summary table to Bookkeeping Plugin

opotowsky:

Having the materials summary table in the blueprints loading is likely one of many places a table or report is outside of the bookkeeping plugin. In this particular case, we wanted the summary attached to a plugin hook so that it could be edited in a downstream bookkeeping plugin. 

## SCR Information

Change Type: features

One-Sentence Rationale: This moves us in the direction of all reporting being done via the Bookkeeping plugin.

---

## Checklist


armi/bookkeeping/__init__.py
```diff
@@ -92,3 +93,33 @@ def mpiActionRequiresReset(cmd) -> bool:
...
+
+
+    @staticmethod
+    @plugins.HOOKIMPL
+    def onProcessCoreLoading(core, cs, dbLoad):
+        """Called whenever a Core object is newly built."""
+        summarizeMaterialData(core)
+
+
+def summarizeMaterialData(container):
```
john-science, line 104: Are you sure you don't want to make this a `staticmethod` inside the class? That would make it easier for you to subclass later.

#not-dying-on-this-hill

opotowsky: I wasn't sure which would be better. I can't really subclass it since I need to edit inside the loop, and the runlog prints the table created. 

And this seemed to be the norm around the ecosystem with helper functions attached to hooks, so I just went with this. Happy to switch it up, but it doesn't matter to me either way.

armi/bookkeeping/tests/test_bookkeepingPlugin.py
```diff
@@ -0,0 +1,83 @@
...
+
+class TestBookkeepingPlugin(TestPlugin):
+    plugin = BookkeepingPlugin
+
+
+class TestBookkeepingPluginHooks(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        """Create a reactor core with case settings, blueprints, reactor."""
+        # TODO: Just copied over old test setup b/c I assume this is a much smaller reactor than the test reactors.
```
john-science: We don't merge `TODO`s in ARMI. The ARMI repo currently has zero.

opotowsky: Oh yea, that's cause I want to ask you about this. The setup is fairly complex b/c I stole it from the original unit test. But I could make it like one line of code by loading a test reactor. We lose the `Custom` material in that process though. What do you think? Should I just simplify it for maintainability?

john-science: Well, if all you want is a test reactor, this one is only one block and it's super fast to build:

https://github.com/terrapower/armi/blob/17419973764444ddd48886ed19e45f2c01e816b0/armi/bookkeeping/report/tests/test_report.py#L179

That wouldn't have your `Custom` material, but it seems easier to maintain to me.

opotowsky: Yea I agree, I just wanted to make sure I wasn't removing some special code coverage or something first :-). I'll push up the cleanup and re-request a review!

opotowsky: (done)

john-science (approved):
Looks  great!

I will just trust that you are going to test this downstream to make sure you still like the look of the log files.

armi/tests/refSmallReactorBase.yaml
```diff
@@ -340,7 +340,7 @@ blocks:
     outer liner:
       shape: Circle
       material: Zr
-      Tinput: 20.0
+      Tinput: 25.0
```
john-science, line 343: This is all we needed this whole time?!?

AWESOME

Thanks for that.

opotowsky: Yea there's just one material property that is in the yaml that is only allowed down to something like 24.85 deg, meaning the 20 deg asks were failing downstream with the test reactor.

PR #2523, opened 2026-05-06
Clean up runtime warnings

john-science:

I spoke with some of our analysts and found that they were having to disposition warnings that appeared in the run time of their code that they believed were not real warnings and not interesting.

So, out of that discussion came this PR, to remove or reduce in priority several warnings that were bothering our user base. 

Thanks to @wcscherer !

## SCR Information

Change Type: trivial

One-Sentence Rationale: ARMI should not be in the business of causing warning fatigue.

---

## Checklist

### DIFF
--- a/armi/materials/b4c.py
+++ b/armi/materials/b4c.py
@@ -40,8 +40,7 @@ class B4C(material.Material):
 
     def applyInputParams(self, B10_wt_frac=None, theoretical_density=None, TD_frac=None, *args, **kwargs):
         if B10_wt_frac is not None:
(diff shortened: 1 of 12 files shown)


armi/materials/b4c.py
```diff
@@ -50,7 +49,7 @@ def applyInputParams(self, B10_wt_frac=None, theoretical_density=None, TD_frac=N
                 single=True,
```
opotowsky, line 49: You could officially deprecate `theoretical_density` and clean up this warning and the runLog item below as well.

opotowsky (approved):

PR #2527, opened 2026-05-07
Cleanup reports

### LINKED ISSUE #793: Clean up the Reporting Tools
## Original Problem Statement

At some point, @youngmit and @chris10mckenz built some new reporting tools. Awesome! However, the project was left unfinished. We we have the original/old reporting tools:

* `armi/bookkeeping/report/data.py`
* `armi/bookkeeping/report/html.py`
* `armi/bookkeeping/report/reportingUtils.py`
* `armi/bookkeeping/report/reportInterface.py`

But we also have the "new" reporting tools:

* `armi/bookkeeping/report/newReports.py`
* `armi/bookkeeping/report/newReportUtils.py`
* `armi/bookkeeping/report/report.js`
* `armi/bookkeeping/report/style.css`

Obviously, leaving the word "new" on those files indefinitely seems silly. This is a confusing state of affairs for our user base. We should combine these tools into one set, and simplify our API.

## Progress Update

The layout of the reporting tools is now much simpler:

* `armi/bookkeeping/report/data.py`
* `armi/bookkeeping/report/reportingUtils.py`
* `armi/bookkeeping/report/reportInterface.py`
* `armi/utils/reportPlotting.py`

But there are still some common concerns I hear about ARMI's reporting tools:

1. Some of the log info or plots are not used, and never have been.
2. Some plots are either broken, or too fragile.
3. Some tools are hyper-specific to only work with Fast or Hex-based, pin-type reactors.
    * Separate these out and make them easily identified.
4. ARMI should just provide TOOLS for reporting things, in a `ReportInterface`. But leave all the detailed choices of what reactor-specific reports and plots should be created to downstream subclasses of the `ReportInterface`.

john-science:

In this PR I removed some plots that our analysts deemed not useful, and cleaned up another plot. You will find each of these changes clearly separated by their own commit:

* Axe the code timing plot
* Axe the maxPD plot
* Axe the SFP report when it is not needed
* Reduce y-axis noise on plotAssemblyTypes

**NOTE**: There was also talk of moving the comprehensive reporting log out of ARMI, and just leaving a simpler version in ARMI. But that change is much more complicated than the four trivial changes that ended up in this PR, so it will have to be handled separately, working with our downstream stakeholder.

close #793 

## SCR Information

Change Type: trivial

One-Sentence Rationale: If our analysts and users do not find a default report helpful, then we should not force the plots upon them.

---

## Checklist


armi/utils/plotting.py
```diff
@@ -768,12 +768,17 @@ def plotAssemblyTypes(
...
+
     yBlockHeights.insert(0, 0.0)
     yBlockHeights.sort()
     yBlockHeightDiffs = np.diff(yBlockHeights)  # Compute differential heights between each block
-    ax.set_yticks([0.0] + list(set(np.cumsum(yBlockHeightDiffs))))
-    ax.xaxis.set_visible(False)
+    yTicks = np.array([0.0] + list(set(np.cumsum(yBlockHeightDiffs))))
+    if len(yTicks) > 20:
+        # keep the number of y-axis ticks to a legible maximum
+        yTicks = yTicks[np.linspace(0, yTicks.size - 1, 20, dtype=int)]
```
drewj-tp: I would love to see how this looks for some of the pathological cases. I wonder if it would be more aesthetically pleasing to take 20 equally spaced values between the min/max z value? Like

```suggestion
        yTicks = np.linspace(yTicks[0], yTicks[-1], 20, dtype=float)]
```

opotowsky (approved):
This sat for 2 weeks and is easy to review, so I reviewed it. :-)

PR #2531, opened 2026-05-12
Remove `callReactorConstructionHook` argument from `db.load` and `loadOperator`

opotowsky:

In #2115 this argument was added, and to preserve backwards compatibility it was defaulted to False. 

However, we have identified downstream that this really should default to True. The `beforeReactorConstruction` hook was added to support flexible materials database usage, and if that hook is missed during reactor loading, the default materials database will be loaded, not necessarily the one the user has chosen via their case settings. This potential mismatch could cause some materials differences one may not catch easily. 

The solution was to remove the argument entirely and always call the hook. The onus will be on developers to use the hook responsibly; documentation is added to the docstring to strongly encourage that.

## SCR Information

Change Type: features

One-Sentence Rationale: To support reactor loading during a `db.load` operation, it makes more sense to call the `beforeReactorConstruction` than for that to default to False. 

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/database.py
+++ b/armi/bookkeeping/db/database.py
@@ -647,17 +647,7 @@ class Database:
         # Reload the file in append mode and continue on our merry way
         self.h5db = h5py.File(self._fullPath, "r+")
 
-    def load(
-        self,
-        cycle,
-        node,
-        cs=None,
-        bp=None,
-        statePointName=None,
-        allowMissing=False,
-        handleInvalids=True,
-        callReactorConstructionHook=True,
-    ):
+    def load(self, cycle, node, cs=None, bp=None, statePointName=None, allowMissing=False, handleInvalids=True):
         """Load a new reactor from a DB at (cycle, node).
 
         Case settings and blueprints can be provided, or read from the database. Providing  can be useful for snapshot
@@ -690,8 +680,6 @@ class Database:
             with undefined parameters. Default False.
         handleInvalids : bool
             Whether to check for invalid settings. Default True.
-        callReactorConstructionHook : bool
-            Flag for whether the beforeReactorConstruction plugin hook should be executed. Default is True.
 
         Returns
         -------
@@ -705,8 +693,7 @@ class Database:
         if bp is None:
             bp = self.loadBlueprints(cs)
 
-        if callReactorConstructionHook:
-            getPluginManagerOrFail().hook.beforeReactorConstruction(cs=cs)
+        getPluginManagerOrFail().hook.beforeReactorConstruction(cs=cs)
 
         if node < 0:
             numNodes = getNodesPerCycle(cs)[cycle]
--- a/armi/bookkeeping/db/database.py
+++ b/armi/bookkeeping/db/database.py
@@ -656,7 +656,7 @@ class Database:
         statePointName=None,
         allowMissing=False,
         handleInvalids=True,
-        callReactorConstructionHook=False,
+        callReactorConstructionHook=True,
     ):
         """Load a new reactor from a DB at (cycle, node).
 
@@ -691,7 +691,7 @@ class Database:
         handleInvalids : bool
             Whether to check for invalid settings. Default True.
         callReactorConstructionHook : bool
-            Flag for whether the beforeReactorConstruction plugin hook should be executed. Default is False.
+            Flag for whether the beforeReactorConstruction plugin hook should be executed. Default is True.
 
         Returns
         -------
--- a/armi/plugins.py
+++ b/armi/plugins.py
@@ -263,8 +263,10 @@ class ArmiPlugin:
         decorator with the following:
 
         .. code::
+            from <somewhere> import <PLUGIN>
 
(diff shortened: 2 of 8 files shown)


john-science:
Well, we have a unit test for this flag here:

https://github.com/terrapower/armi/blob/76b5eee6a2d7dc7033336e99f5af4a94aed586a8/armi/tests/test_plugins.py#L155

Is it enough? Up to you.

john-science (approved):
Can't wait to see the unit test. You rock.

alexhjames (approved):

drewj-tp:
Pulling a point from conversation into the public.

It is entirely likely that users will make successive calls to `db.load` from the same database in the same python environment. If you want to look at time evolution of your reactor, you need to load the reactor from the database at multiple cycle/node pairs. That means there will be multiple calls to this hook. Which can be noisy if a plugin prints a lot of information and/or take unnecessary time (e.g., loading the same data files over and over again)

How can we communicate this information to implementors of plugins? That they should implement some caching / logic to know if they _really_ need to perform some update. Or if they've gotten the same settings as the last call?

alexhjames:
For caching purposes, the user would be looking at `loadedRootDirs` list from matProps.  The current logic is a little wrong because if `clearFirst=True` then the directory would be added to the list, then cleared, then loaded so the user wouldn't know they loaded that directory. Perhaps we can clean that up during this PR as well. https://github.com/terrapower/armi/blob/76b5eee6a2d7dc7033336e99f5af4a94aed586a8/armi/materials/__init__.py#L130-L133

drewj-tp:
Users calling `db.load` could also pass `callReactorConstructionHook=False` if they are aware of how this will impact their loaded reactor. In some cases, it's probably fine (it was the previous behavior before this PR and before #2115)

But there are very real cases where you'd want to call this hook loading a database. And then it should be up to the implementors of plugins to know when to not do potentially noisy / slow things (imo)

armi/plugins.py
```diff
@@ -237,7 +237,28 @@ def onProcessCoreLoading(core, cs, dbLoad) -> None:
     @staticmethod
     @HOOKSPEC
     def beforeReactorConstruction(cs) -> None:
-        """Function to call before the reactor is constructed."""
+        """
+        Function to call before the reactor is constructed.
+
+        Warning
+        -------
+
```
john-science:
```suggestion
        Notes
        -----
```

john-science: Just an opinion

opotowsky: This is more like a "you shall do this or else" thing. would `Important` work?

armi/utils/__init__.py
```diff
@@ -745,3 +746,15 @@ def safeMove(src: str, dst: str) -> None:
 
     runLog.extra(f"Moved {src} -> {dst}")
     return dst
+
+
+def onlyRunOnce(method):
+    @functools.wraps(method)
```
john-science, line 754:
```suggestion
def onlyRunOnce(method):
    """A decorator to ensure a method is only run once."""
    @functools.wraps(method)
```

john-science: The wording is up to you, I just thought it could use a docstring of some kind.

opotowsky: Oh thanks!

armi/plugins.py
opotowsky: I put `Attention` :-D

drewj-tp: From the [numpydoc docs](https://numpydoc.readthedocs.io/en/latest/format.html#warns)
image

opotowsky: warnings with an s rather than without 🤯

oops, and interesting. 

I dont have strong opinions here, I just need plugin devs to know this is critical info and if they don't follow it they COULD RUIN EVERYTHING (everything meaning db loading speed)

armi/plugins.py
```diff
@@ -237,7 +237,37 @@ def onProcessCoreLoading(core, cs, dbLoad) -> None:
     @staticmethod
     @HOOKSPEC
     def beforeReactorConstruction(cs) -> None:
-        """Function to call before the reactor is constructed."""
+        """
+        Function to call before the reactor is constructed, typically implemented as related to global materials
+        loading.
+
+        Attention
+        ---------
```
opotowsky, line 245:
```suggestion
        Warnings
        --------
```

armi/bookkeeping/db/database.py
```diff
@@ -647,17 +647,7 @@ def syncToSharedFolder(self):
...
-        cycle,
-        node,
-        cs=None,
-        bp=None,
-        statePointName=None,
-        allowMissing=False,
-        handleInvalids=True,
-        callReactorConstructionHook=False,
-    ):
+    def load(self, cycle, node, cs=None, bp=None, statePointName=None, allowMissing=False, handleInvalids=True):
```
john-science, line 650: I love that this got shortened. :)

john-science (approved):
Great!

PR #2536, opened 2026-05-15
Reducing reliance on global nuclides

john-science:

This PR does not solve, but merely makes progress on, removing Global Nuclides.

Global nuclides were removed from:

* `Case._initBurnChain()`
* `test_xsNuclides.py`

Progress on #473

## SCR Information

Change Type: features

One-Sentence Rationale: We are working toward a major redesign goal where we remove "nuclideBases" from the global scope; many users have requested this.

One-line Impact on Requirements: Minor docstring change to I_ARMI_R.

---

## Checklist


armi/reactor/reactors.py
```diff
@@ -28,27 +28,23 @@ class Reactor(composites.Composite):
     """
     Top level of the composite structure, potentially representing all components in a reactor.
 
-    This class contains the core and any ex-core structures that are to be represented in the ARMI
-    model. Historically, the ``Reactor`` contained only the core. To support better representation
-    of ex-core structures, the old ``Reactor`` functionality was moved to the newer `Core` class,
-    which has a ``Reactor`` parent.
+    This class contains the core and any ex-core structures that are to be represented in the ARMI model. Historically,
+    the ``Reactor`` contained only the core. To support better representation of ex-core structures, the old ``Reactor``
+    functionality was moved to the newer `Core` class, which has a ``Reactor`` parent.
```
opotowsky: I'm not a fan of historical info in docstrings. We can just say Reactor is the parent, it has a Core child, and other ex-core structures are also children

armi/reactor/reactors.py
```diff
@@ -28,27 +28,23 @@ class Reactor(composites.Composite):
...
-        :py:class:`Assembly <armi.reactor.assemblies.Assembly>` objects arranged in a hexagonal or
-        Cartesian grid. Each Assembly consists of a stack of
-        :py:class:`Block <armi.reactor.blocks.Block>` objects, which are each composed of one or
-        more :py:class:`Component <armi.reactor.components.component.Component>` objects. Each
-        :py:class:`Interface <armi.interfaces.Interface>` is able to interact with the reactor and
-        its child :py:class:`Composites <armi.reactor.composites.Composite>` by retrieving data from
-        it or writing new data to it. This is the main medium through which input information and
-        the output of physics calculations is exchanged between interfaces and written to an ARMI
-        database.
+        The Reactor is the top level of the composite structure, which can represent all components within a reactor
```
opotowsky, line 36: I thought since this was edited there would be something about nuclides being attached to the reactor object, or is that a change you'll make later?

opotowsky (commented):
One request, one question, and then there's TODOs in the PR description. But otherwise this looks good!

armi/reactor/reactors.py
john-science: Done. Love it.

armi/reactor/reactors.py
john-science: Right now, I am still faking that by making a class property that calls global nuclides. I can't get rid of that until global nuclides are finally, truly dead. Which this PR doesn't do; this is just a progress PR (what I had time for).

You're right though, when global nuclides are finally dead, this should get updated.

opotowsky (approved):

Issue #2471, opened 2026-03-10
codecov.io fails when coverage drops 0.02%


john-science:
On a recent ARMI PR into `main`, the code coverage of the repo went down by one line, but `codecov.io` says this is a [failure](https://app.codecov.io/gh/terrapower/armi/commit/b40ec077ac83a2fb441ab25b8538d0517c33e9b5):

<img width="486" height="140" alt="code cov failure" src="https://github.com/user-attachments/assets/06c0391f-fc8d-49bd-8cd1-3d659c08fb11" />

That is obviously ridiculous, so we will need to fix that.

[This](https://github.com/codecov/codecov-action) is the action we are using. I do not explicitly see anything configurable here to stop this nonsense. The only option I see that _might_ be it is `fail_ci_if_error: false`. But I am not sure that will work.

Maybe we would need a [codecov.io config file](https://docs.codecov.com/docs/common-recipe-list) in our repo to solve this problem? Eww.

john-science:
This is still happening, sigh: https://app.codecov.io/gh/terrapower/armi/commit/f6b78c5a000ae54ca2b5111d8d7721b1c70b93b9

john-science:
Codecov is still failing the build, which I have tried to totally block for any reason, and it is not working:

<img width="558" height="89" alt="Image" src="https://github.com/user-attachments/assets/898afd84-a6ad-4a57-99a9-749c3d044903" />

This is doubly annoying, because if it weren't for this, I would otherwise love codecov.io.

john-science:
Okay, I haven't had this problem in a while. This might be fixed.

We will wait a while longer and see.

john-science:
Drat:

<img width="816" height="161" alt="Image" src="https://github.com/user-attachments/assets/babe7730-9408-4e79-bf14-dcaedb4fa8a6" />

A [recent PR](https://github.com/terrapower/armi/pull/2574) added some code that was not covered by unit tests, so the ARMI coverage job failed, even though the coverage is still above 90%.

This is so frustrating.

john-science:
I think perhaps the original goal of this ticket has been solved.

But now we are getting codecov.io failures if less than 90% of "new lines of code added in this PR" are covered by tests. And that's... not an interesting metric to me.

So now we need to fix THAT problem.

john-science:
This hasn't failed in a while.  Fingers crossed.

john-science:
This hasn't happened in over a month.

I believe my PR #2586 solved the problem. 

[...]

PR #2537, opened 2026-05-18
Trying to stop codecov.io from failing the build

john-science:

We are still having the problem with codecov.io failing our build. I like their website, but they fail our build WAY too aggressively.

This time I pushed a couple of bad coverage reports to codecov.io, by added and removing test files from ARMI and testing the next .codecov.io settings file. 
 
I am reasonably sure this will work this time.

progress on #2471

Still, I won't close the Issue until codecov.io passes without failing for a couple weeks. I'm just too paranoid to trust  my solution at this point.

## SCR Information

Change Type: trivial

One-Sentence Rationale: Code coverage should not fail ARMI CI all the time, that will create alarm fatigue.

---

## Checklist


john-science:
Okay, I got the build to fail by changing the `.codecov.yml` file:

image

So I think my new branch here is working.

I made some changes to `.codecov.yml` should do the trick. #fingers-crossed

.github/.codecov.yml
```diff
@@ -0,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 1%    # the required coverage value: stop codecov.io from failing the build
+        threshold: 50%  # allows a 50% drop from the previous base commit coverage
```
opotowsky: The previous PR deleted this file:

Image

So are you just making these values wildly permissive so it'll never fail the build and we still get our reporting?

I'm curious as to why the build was failing when this file was missing entirely

john-science: Yeah, these are WILDLY permissive.

I guess I could put that in a comment at the top of the file.

I think the "target" is stupid. Say you rewrite 4 lines of code that already exist. CodeCov.io considers all of these "new" lines. Now if only 3 of the 4 of those lines are covered by tests, codecov.io says you only have 75% coverage of "new lines" and compares that to our old 80% target and fails the build.

Which is all kinds of stupid. And means unless you have 100% code coverage, your build will fail just all the time.

john-science: Okay, I added a comment to the file. See what you think.

opotowsky: OMG THAT IS SO DUMB! (the codecov behavior, not your comment)

opotowsky (approved):

PR #2543, opened 2026-05-21
Adding equality to Settings and Settings

### LINKED ISSUE #2535: Support equality of settings objects
There are some cases where you may want to know if two `Settings` objects are identical with `==`. Right now, this is not possible, since `Settings.__eq__` is not implemented, comparisons are made off object identity
```python
>>> from copy import deepcopy
>>> cs = Settings(PATH)
>>> cs == cs
True
>>> copied == Settings(PATH)
False
>>> cs == deepcopy(cs)
False
```

john-science:

Adding the Python-standard dunder method `__eq__` to `Setting` and `Settings`, so that we can check for the equality of both.

close #2535

## SCR Information

Change Type: features

One-Sentence Rationale: Adding the ability to check for the equality of settings objects.

One-line Impact on Requirements: Adding a new feature that adds code to I_ARMI_SETTING0.

---

## Checklist

### DIFF
--- a/armi/settings/caseSettings.py
+++ b/armi/settings/caseSettings.py
@@ -145,11 +145,17 @@ class Settings:
 
         return f"<{self.__class__.__name__} name:{self.caseTitle} total:{total} altered:{altered}>"
 
-    def __eq__(self, other): # TODO
+    def __eq__(self, other):
         if type(self) is not type(other):
             return False
-        
-        TODOreturn self.__members() == other.__members()
+        elif self.keys() != other.keys():
+            return False
+
+        for key in self.keys():  # noqa: SIM110
+            if self[key] != other[key]:
+                return False
+
+        return True
 
     def _directAccessOfSettingAllowed(self, key):
         """
@@ -159,11 +165,10 @@ class Settings:
 
         Notes
         -----
-        Checking the validity of grabbing specific settings at this point, as is done for the
-        SIMPLE_CYCLES_INPUT's, feels a bit intrusive and out of place. In particular, the fact that
-        the check is done every time that a setting is reached for, no matter if it is the setting
-        in question, is quite clunky. In the future, it would be desirable if the settings system
-        were more flexible to control this type of thing at a deeper level.
+        Checking the validity of grabbing specific settings at this point, as is done for the SIMPLE_CYCLES_INPUT's,
+        feels a bit intrusive and out of place. In particular, the fact that the check is done every time that a setting
+        is reached for, no matter if it is the setting in question, is quite clunky. In the future, it would be
+        desirable if the settings system were more flexible to control this type of thing at a deeper level.
         """
         if key not in self.__settings:
             return False, NonexistentSetting(key)
--- a/armi/settings/caseSettings.py
+++ b/armi/settings/caseSettings.py
@@ -13,13 +13,12 @@
 # limitations under the License.
 
 """
-This defines a Settings object that acts mostly like a dictionary. It
-is meant so that each ARMI run has one-and-only-one Settings object. It records
-user settings like the core power level, the input file names, the number of cycles to
-run, the run type, the environment setup, and hundreds of other things.
+This defines a Settings object that acts mostly like a dictionary. It is meant so that each ARMI run has
+one-and-only-one Settings object. It records user settings like the core power level, the input file names, the number
+of cycles to run, the run type, the environment setup, and hundreds of other things.
 
-A Settings object can be saved as or loaded from an YAML file. The ARMI GUI is designed to
-create this settings file, which is then loaded by an ARMI process on the cluster.
+A Settings object can be saved as or loaded from an YAML file. The ARMI GUI is designed to create this settings file,
+which is then loaded by an ARMI process on the cluster.
 """
 
 import io
@@ -53,19 +52,17 @@ class Settings:
         :id: I_ARMI_SETTING0
         :implements: R_ARMI_SETTING
 
-        The Settings object is accessible to most ARMI objects through self.cs
-        (for 'case settings'). It acts largely as a dictionary, and setting values
-        are accessed by keys.
+        The Settings object is accessible to most ARMI objects through self.cs (for 'case settings'). It acts largely as
+        a dictionary, and setting values are accessed by keys.
 
-        The Settings object has a 1-to-1 correspondence with the ARMI settings
-        input file. This file may be created by hand or by a GUI.
+        The Settings object has a 1-to-1 correspondence with the ARMI settings input file. This file may be created by
+        hand or by a GUI.
 
     Notes
     -----
-    While it is possible to modify case settings during the course of a run, this
-    is highly discouraged because there will be no record of this happening in your
-    results or in the database produced from your run. There is no guarantee that
-    doing so will not cause unexpected problems with your calculation.
+    While it is possible to modify case settings during the course of a run, this is highly discouraged because there
+    will be no record of this happening in your results or in the database produced from your run. There is no guarantee
+    that doing so will not cause unexpected problems with your calculation.
     """
 
     defaultCaseTitle = "armi"
@@ -85,11 +82,10 @@ class Settings:
         self._failOnLoad = False
         """This is state information.
 
-        The command line can take settings, which override a value in the current
-        settings file; however, if the settings file is listed after a setting value,
-        the setting from the settings file will be used rather than the one explicitly
-        provided by the user on the command line.  Therefore, _failOnLoad is used to
-        prevent this from happening.
+        The command line can take settings, which override a value in the current settings file; however, if the
+        settings file is listed after a setting value, the setting from the settings file will be used rather than the
+        one explicitly provided by the user on the command line.  Therefore, _failOnLoad is used to prevent this from
+        happening.
         """
         from armi import getApp
 
@@ -118,14 +114,11 @@ class Settings:
             :id: I_ARMI_SETTINGS_META0
             :implements: R_ARMI_SETTINGS_META
 
-            Every Settings object has a "case title"; a string for users to
-            help identify their run. This case title is used in log file
-            names, it is printed during a run, it is frequently used to
-            name the settings file. It is designed to be an easy-to-use
-            and easy-to-understand way to keep track of simulations. The
-            general idea here is that the average analyst that is using
-            ARMI will run many ARMI-based simulations, and there needs
-            to be an easy to identify them all.
+            Every Settings object has a "case title"; a string for users to help identify their run. This case title is
+            used in log file names, it is printed during a run, it is frequently used to name the settings file. It is
+            designed to be an easy-to-use and easy-to-understand way to keep track of simulations. The general idea here
+            is that the average analyst that is using ARMI will run many ARMI-based simulations, and there needs to be
+            an easy to identify them all.
         """
         if not self.path:
             return self.defaultCaseTitle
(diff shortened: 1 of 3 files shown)


armi/settings/caseSettings.py
```diff
@@ -150,7 +143,19 @@ def __repr__(self):
         isAltered = lambda s: 1 if s.value != s.default else 0
         altered = sum([isAltered(setting) for setting in self.__settings.values()])
 
-        return "<{} name:{} total:{} altered:{}>".format(self.__class__.__name__, self.caseTitle, total, altered)
+        return f"<{self.__class__.__name__} name:{self.caseTitle} total:{total} altered:{altered}>"
+
+    def __eq__(self, other):
+        if type(self) is not type(other):
```
drewj-tp, line 149: A side-effect here, maybe intentional, is we will only be able to compare a literal `Settings` object to another `Settings` object. Nominally that's okay and I can get behind not comparing settings to non-settings (though dict _might_ be nice!) But subclasses of settings will not work b/c they have different types.

I cannot do equality check on a list and tuple b/c (as expected) they have different types. But I can do an equality on a list and a subclass of list
```python
>>> class MyList(list):
...     pass
...
>>> [1] == MyList([1])
True
```
This is well and good; I just wanted to communicate that. Please consider adding this to the docstring, even as simple as `Compare on ``Settings`` object to another`

armi/settings/caseSettings.py
```diff
@@ -150,7 +143,19 @@ def __repr__(self):
...
+            return False
```
drewj-tp, line 150: The `return False` here means we don't have the option for a symmetric equality check. If `a == b` is not supported by `a`, python allows the call to `b == a` (same for a lot of other options like addition and subtraction). This is done by

```suggestion
            return NotImplemented
```
Again, if we're only interested in comparing literal `Settings` objects against each other and they shall not be identical otherwise, then `False` is okay. Returning `NotImplemented` allows other libraries to write their own settings equality, if they desire.
```python
>>> class A:
...     def __eq__(self, other):
...         print("Hello from A")
...         if type(self) is not type(other):
...             return NotImplemented
...         return True
...
>>> class B:
...     def __eq__(self, other):
...         print("Hello from B")
...         if type(self) is not type(other):
...             return NotImplemented
...         return True
...
>>> A() == A()
Hello from A
True
>>> A() == B()
Hello from A
Hello from B
False
```
So `A.__eq__(B)` returns `NotImplemented` which then checks `B.__eq__(A)`.

armi/settings/setting.py
```diff
@@ -315,6 +315,30 @@ def __copy__(self):
...
+            self.subLabels,
+            self.isEnvironment,
+            self.oldNames,
+            self._default,
+            self._value,
+        )
+
+    def __eq__(self, other):
+        if type(self) is type(other):
+            return self.__members() == other.__members()
```
drewj-tp, line 335: I am very surprised we're able to access `other.__members()`. I would have expected that be private and mangled and something another object cannot touch. But that seems to be different if it's one class trying to access the private attribute of an identical class.

Looking at the disassembled bytecode
```python
>>> class A:
...     def __init__(self, d):
...         self.__data = d
...     def __members(self):
...         return self.__data
...     def __eq__(self, other):
...         if type(self) is type(other):
...             return self.__members() == other.__members()
...         return False
...

>>> dis.dis(A.__eq__)
  6           RESUME                   0

  7           LOAD_GLOBAL              1 (type + NULL)
              LOAD_FAST                0 (self)
              CALL                     1
              LOAD_GLOBAL              1 (type + NULL)
              LOAD_FAST                1 (other)
              CALL                     1
              IS_OP                    0
              POP_JUMP_IF_FALSE       33 (to L1)

  8           LOAD_FAST                0 (self)
              LOAD_ATTR                3 (_A__members + NULL|self)
              CALL                     0
              LOAD_FAST                1 (other)
              LOAD_ATTR                3 (_A__members + NULL|self)
              CALL                     0
              COMPARE_OP              72 (==)
              RETURN_VALUE

  9   L1:     RETURN_CONST             1 (False)
```
The demangling is handled correctly. And you have tests that show it's working. 

Could we use [`vars`](https://docs.python.org/3/library/functions.html#vars) instead?
```python
>>> s = armi.settings.Setting(name="my setting", default=None, description="this is a silly setting")
>>> vars(s)
{'name': 'my setting',
 'description': 'this is a silly setting',
 'label': 'my setting',
 'options': None,
 'enforcedOptions': False,
 'subLabels': None,
 'isEnvironment': False,
 'oldNames': [],
 '_default': None,
 '_value': None,
 '_customSchema': None,
 'schema': <Schema(Coerce(NoneType, msg=None), extra=PREVENT_EXTRA, required=False) object at 0x18c10b81b70>}

>>> vars(s) == vars(s)
True
```

armi/settings/setting.py
```diff
@@ -315,6 +315,30 @@ def __copy__(self):
...
+            self.name,
+            self.description,
+            self.label,
+            self.options,
+            self.enforcedOptions,
+            self.subLabels,
+            self.isEnvironment,
+            self.oldNames,
+            self._default,
+            self._value,
```
drewj-tp, line 330: Not equal based on `schema`? Maybe it's a huge stretch that someone finds themselves with two otherwise identical `Setting` objects but they have two different validation schemas.

armi/settings/setting.py
```diff
@@ -315,6 +315,30 @@ def __copy__(self):
...
+
+    def __eq__(self, other):
+        if type(self) is type(other):
+            return self.__members() == other.__members()
+        else:
+            return False
+
+    def __hash__(self):
+        return hash(str(self.__members()))
+
```
drewj-tp, line 341: I'm not sure how to feel about the hashability of a `Setting`. Do we envision people using a `Setting` object as e.g., a key in a dictionary? a. la
```python
>>> s = armi.settings.Setting(name="my setting", default=None, description="this is a silly setting")
>>> {s: value}
```
This is _kind of_ what the `Settings` object aspires to do, but we store the `Setting` objects as the _values_ in the `Settings.__settings` dictionary, with their setting names as keys

drewj-tp (changes requested):
I think the `Settings.__eq__` implementation is excellent. I have two non-blocking comments. I think we are justified in taking a strict stance of "Only two literal settings objects can be compared" and disallow subclasses. That should be documented.

I have some thoughts about `Setting.__eq__`. I don't think we need it for `Settings.__eq__` because we're comparing against the _value_ of the `Setting` `cs[k] == other[k]` not against the `Setting` object (with description and old names and default and ...)

armi/settings/caseSettings.py
john-science: Well, IMO, if someone is comparing a vanilla `Setting` object to a `NeutronicsSetting(Setting)` object, than we can just say they are not equal.

And we have VERY few `Setting` subclasses, at least in code that I have seen.

armi/settings/caseSettings.py
john-science: > Returning NotImplemented allows other libraries to write their own settings equality, if they desire.

What's the use-case for that?

armi/settings/setting.py
john-science: If I implement `__eq__` without implementing `__hash__`, a bunch of things break.  It's both or neither.

armi/settings/caseSettings.py
drewj-tp: This comment thread is about the `Settings` object and potential subclasses, e.g., `MySettings(Settings)`, not the individual `Setting` object

armi/settings/caseSettings.py
drewj-tp: Who knows! But this type of pattern opens the door for people to extend functionality. `float == np.ndarray` works with this behavior. float equality method doesn't support comparison to an array, but it doesn't return False, it returns `NotImplemented` so python can switch and see if an array knows how to compare to a float
```python
>>> import numpy as np
>>> a = 2.3
>>> x = np.arange(5, dtype=float)
>>> a.__eq__(x)
NotImplemented
>>> a == x
array([False, False, False, False, False])
```
If both objects return `NotImplemented`, python handles it automagically
```python
>>> class UnsureEq:
...     def __eq__(self, other):
...         return NotImplemented
...

>>> UnsureEq() == UnsureEq()
False

>>> UnsureEq() == 1.2
False
```

Do _we_ have a use case for this? Probably not! Would someone else want to? Maybe! And this lets them do it without needing to petition us for a code change or always make sure the ARMI Settings object is on the right side of the `==` to avoid triggering our `__eq__`.

armi/settings/setting.py
drewj-tp: That's surprising. I know things that hash identically should pass an equality check, but I don't think _python_ requires `__hash__` to be implemented if `__eq__` is implemented. Maybe it's an us problem?

That said, I don't think we need `Setting.__eq__` in order for `Settings.__eq__` to make sense. I am okay if we scrub `Setting.__eq__`

armi/settings/caseSettings.py
john-science: I don't think I have ever seen anyone do this: `MySettings(Settings)`.

drewj-tp: Okay

armi/settings/setting.py
john-science: So, I tried this to be sure, and yeah it's as I thought.

If you use `vars()` instead of my `__members()` solution, then this fails:

```python
s1 = setting.Setting("testSetting", 897, description="whatever")
s1Copy = copy.deepcopy(s1)
self.assertEqual(s1, s1Copy)  # fails
```

And I think it would very much confuse users if that fails. So I wanted it to pass, thus my solution. (Which we can call `_members()` if that double underscore wrinkles your brain.

armi/settings/setting.py
john-science: No, the Schema instance is too unique and will fail the equality if you try and copy a settings object. (See discussion below on `vars()`.

armi/settings/setting.py
john-science: The deal is that in `settingsIO.py` we store off settings in a dictionary at several steps.

So, that's doable if you don't implement `__hash__` or `__eq__` (like the current tip of ARMI `main`). But if you implement `__eq__()` then Python will demand that your custom class has `__hash__` on it to take the hash of the object.

armi/settings/caseSettings.py
john-science: I can't imagine a use-case for it, and readability counts. 

I think having a method that is easy to read and understand is super important.

armi/settings/setting.py
john-science: I could put it in, if you really want.

But `.schema` isn't in the constructor, so I don't want to trust that attribute will always exist.

armi/settings/caseSettings.py
drewj-tp: Okay

drewj-tp (commented):
My remaining comments hinge on the implementation of `Setting.__eq__`. I'm not convinced we need it in order to support `Settings.__eq__` which is what #2535 asks for.

Please consider removing `Setting.__eq__` or helping me understand why it's necessary to support `Settings.__eq__`

john-science:
> My remaining comments hinge on the implementation of `Setting.__eq__`. I'm not convinced we need it in order to support `Settings.__eq__` which is what #2535 asks for.
> 
> Please consider removing `Setting.__eq__` or helping me understand why it's necessary to support `Settings.__eq__`

This is just a practical matter.  If you want to test for the equality of a `Settings` object, that is little more than a collection of `Setting` objects.  I don't know how you could test for the equality of `Settings` without testing for the equality for `Setting`.

How do you suggest that happen?

drewj-tp:
> > My remaining comments hinge on the implementation of `Setting.__eq__`. I'm not convinced we need it in order to support `Settings.__eq__` which is what #2535 asks for.
> > Please consider removing `Setting.__eq__` or helping me understand why it's necessary to support `Settings.__eq__`
> 
> This is just a practical matter. If you want to test for the equality of a `Settings` object, that is little more than a collection of `Setting` objects. I don't know how you could test for the equality of `Settings` without testing for the equality for `Setting`.
> 
> How do you suggest that happen?

The implementation you've proposed for `Settings.__eq__` is sufficient and does not use `Setting.__eq__`.  It

1. Checks the `Settings.keys()` are equivalent -> keys of the underlying dictionary of settings -> names of `Setting`
2. Check the **values** of the settings are equivalent `cs[x} == other[x]`

I think this is exactly what we would expect to compare the equality of two `Settings` mappings: they would produce equivalent physics results if given to an operator / application. And it does not use `Setting.__eq__`

PR #2548, opened 2026-06-01
Reducing warnings from B10 Block calculation

john-science:

This change comes out of a conversation with @alexhjames and @mgjarrett.

In practice, we think it is possible that multiple components have B10 in them. So we opt for reducing the likely of unneeded warnings.

This is part of our effort to fight Alarm Fatigue.

## SCR Information

Change Type: trivial

One-Sentence Rationale: Alarm fatigue is contrary to safety culture.

---

## Checklist

### DIFF
--- a/armi/reactor/blocks/block.py
+++ b/armi/reactor/blocks/block.py
@@ -757,7 +757,7 @@ class Block(composites.Composite):
         Parameters
         ----------
         heightHot : bool
-            True if self.height() is cold height.
+            True if self.height() is hot height.
         """
         # Exclude fuel components since they could have slight B10 impurity and this metric is not relevant for fuel.
         b10Comps = [c for c in self if c.getNumberDensity("B10") and not c.isFuel()]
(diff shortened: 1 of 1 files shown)


armi/reactor/blocks/block.py
```diff
@@ -747,30 +747,32 @@ def completeInitialLoading(self, bolBlock=None):
...
 
+        For situations where multiple components in a Block contain B10, this method only coniders the component that
+        has the highest B10 density. That is, hopefully, it selects the one component that has non-trace amounts of B10.
+
         Parameters
         ----------
-        heightHot : Boolean
-            True if self.height() is cold height
+        heightHot : bool
+            True if self.height() is cold height.
```
mgjarrett: I think the existing docstring was wrong here.
```suggestion
            True if self.height() is hot height.
```

alexhjames (approved):

armi/reactor/blocks/block.py
john-science: It's been wrong this whole time???

PR #2551, opened 2026-06-05
Updating ParamLocation for `heightBOL`

mgjarrett:

Change the `ParamLocation` of the block parameter `heightBOL` from `AVERAGE` to `VOLUME_INTEGRATED`.

## SCR Information

Change Type: fixes

One-Sentence Rationale: `heightBOL` is a volume-integrated property of a block.

---

## Checklist

### DIFF
--- a/armi/reactor/blockParameters.py
+++ b/armi/reactor/blockParameters.py
@@ -368,7 +368,7 @@ def getBlockParameterDefinitions():
             "heightBOL",
             units=units.CM,
             description="As-fabricated height of this block (as input). Used in fuel performance. Should be constant.",
-            location=ParamLocation.AVERAGE,
+            location=ParamLocation.VOLUME_INTEGRATED,
             categories=[parameters.Category.retainOnReplacement],
         )
 


mgjarrett:
On second thought, this won't work, because `heightBOL` is volume-integrated only in the height dimension and not in the radial dimension. 🤔  Not sure what to do about this. I don't want to create another `ParamLocation` for `HEIGHT_INTEGRATED`.

john-science:
The set of ParamLocations is meant to be small. Because the more of these we add, the more places in ARMI and downstream will have to support the new location.

If only one parameter needs this new "height-integrated" value, then I'm not in favor of it. Because what we have them is a parameter that just needs it's own bespoke / unique logic to calculate. Most parameters do.

Let's not create unnecessary abstractions.

john-science:
Bump.

mgjarrett:
Closing this pull request. The `ParamLocation.VOLUME_INTEGRATED` construct doesn't work for `heightBOL`.

PR #2554, opened 2026-06-18
Improving documentation

yardasol:

Makes typo fixes to the docs and changes a block from a tip to a note.

## SCR Information

Change Type: docs

One-Sentence Rationale: The documentation is important, so we must make sure they are as high quality as possible.

---

## Checklist

### DIFF
--- a/doc/user/inputs.rst
+++ b/doc/user/inputs.rst
@@ -9,10 +9,10 @@ ARMI input files define the initial state of the reactor model and tell ARMI wha
 There are several input files:
 
 Settings file
-    Contains simulation parameters (like full power, cycle length, and which physics modules to activate) and all kind of modeling approximation settings (e.g. convergence criteria)
+    Contains simulation parameters (like full power, cycle length, and which physics modules to activate) and all kinds of modeling approximation settings (e.g. convergence criteria)
 
 Blueprints file
-    Contains dimensions and composition of the components/blocks/assemblies in your reactor systems, from fuel pins to heat exchangers
+    Contains dimensions and compositions of the components/blocks/assemblies in your reactor systems, from fuel pins to heat exchangers
 
 Fuel management file
     Describes how fuel moves around during a simulation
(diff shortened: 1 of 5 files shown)


john-science:
@yardasol So, fun fact, we use the title of this PR as the commit message that goes into `main`. And... I'm not sure if this is important, but nearly every single commit in ARMI main is in active voice (the verb ends in "ing").  

This is a stupid little detail, so just FYI. We COULD, of course, change it. Because I'm not sure if I love this. But since this has been the way for 7 years now, we are going to keep going this way. Unless we purposefully break the pattern, of course.

#shrug #details

doc/user/inputs.rst
```diff
@@ -515,7 +515,7 @@ ARMI adds an ``!include`` YAML tag, which can be used to include the contents of
    grids:
        core: !include path/to/core_grid.yaml
 
-Would have the effect of copy-pasting the contents of ``path/to/core_grid.yaml`` into the main blueprints file. The rules that ARMI uses to handle things like indentation of the included text are usually rather intuitive, but sometimes it can be useful to witness the behavior first-hand. The ``expand-bp`` command can be used to do a dry run for testing inputs with !includes.
+would have the effect of copy-pasting the contents of ``path/to/core_grid.yaml`` into the main blueprints file. The rules that ARMI uses to handle things like indentation of the included text are usually rather intuitive, but sometimes it can be useful to witness the behavior first-hand. The ``expand-bp`` command can be used to do a dry run for testing inputs with !includes.
```
john-science, line 518: Maybe this is an incomplete sentence as is? This could be something like:

"This would have the effect of whatever stuff things..."

Thoughts? Opinions?

yardasol: I see this as a continuation of the previous sentence with a code-block example in the middle, hence the lower-casing of the "W". Maybe that's not proper grammar. I can rework if we want complete, concrete sentences.

john-science:
@yardasol Can you add your name and username to the "AUTHORS" file at the top of the repo?  :)

john-science (approved):

PR #2555, opened 2026-06-18
Improving setMaterialNamespaceOrder

### LINKED ISSUE #2520: Move matProps venv dir logic into `setMaterialNamespaceOrder`
The idea I had was to move to something like this:
 
```python
materials.setMaterialNamespaceOrder([
     "my.awesome.project",
     "venv:materials_data",
     "armi.materials"])
 ```
 
Here we could have special keywords like `"venv:x"` or `"dir:y"` that explicitly load directories of matProps YAML files, even from a directory in the venv (which we currently do in matProps).
 
This would be a lot easier to configure and understand than the venv-dir logic that has been hiding in matProps. It would also let ARMI be the driver for when materials get loaded into ARMI, rather than depending on downstream codes.

My opinion is this is a cleaner, easier design to understand and reason about.

### PR DESCRIPTION
**Pull request by john-science**

Here I improve the functionality of `setMaterialNamespaceOrder()` so the user can add paths to directories and even paths inside the Python virtual environment right into the namespace order. These are things people are already doing, but finally they will be supported.

Also, this feature makes the true namespace ordering much clearer.

NOTE: The new directory / venv logic only supports directories of YAML files, not materials defined in Python.

close #2520

### Description of Design Differences

1. We already had a cache for pure matProps YAML materials in the matProps area. Now we have one for more ARMI-centric YAML-based materials in the ARMI area: `_loadedYamlDirs`.
2. I learned the hard way that we can't have this new logic work with the old `resolveMaterialClassByName()`. It breaks pickling. So now we have `createMaterialByName()`. (We can change the name of that function if you want.)
3. Less critically, I created the static method `MatPropsMaterial.getMatNameFromYamlPath()` to help us do something we often want to do.

## SCR Information

Change Type: features

One-Sentence Rationale: This change should make the material namespace much more transparent.

One-line Impact on Requirements: Minor name-change implementation difference to T_ARMI_MAT_NAMESPACE0. Logic added to implementation of I_ARMI_MAT_ORDER. Major rewrite of implementation of resolveMaterialClassByName to become createMaterialByName. Minor implementation change to I_ARMI_MAT_PROPERTIES1. Very minor documentation change to I_ARMI_MAT_USER_INPUT1.

---

## Checklist

### CODE REVIEW

File: armi/materials/__init__.py [hunk 1]
```diff
@@ -39,112 +39,107 @@
...
+        collections of materials. To handle this, ARMI keeps a list of material namespaces. This is an ordered list of
+        importable packages that ARMI can search for a particular material by name.
+
+        This automatic exploration of an importable package saves the user the tedium have having to import or include
+        hundreds of materials manually somehow. But it comes with a caveat; the list is ordered. If two different
+        namespaces in the list include a material with the same name, the first one found in the list is chosen, i.e.
+        An ARMI application will need materials. Materials can be imported from any code the application has access to,
+        like plugin packages. This leads to the situation where one ARMI application will want to import multiple
+        collections of materials. To handle this, ARMI keeps a list of material namespaces. This is an ordered list of
+        importable packages that ARMI can search for a particular material by name.
```
opotowsky: You have some repeated sentences in here (in the first 2 paragraphs) that need some cleaning up

armi/materials/__init__.py
```diff
@@ -244,30 +239,48 @@ def resolveMaterialClassByName(name: str, namespaceOrder: List[str] = None):
...
     namespaceOrder = namespaceOrder or _MATERIAL_NAMESPACE_ORDER
+
+    # 3. Try to import the material from a namespace defined above
     for namespace in namespaceOrder:
-        mod = importlib.import_module(namespace)
-        materialsList = inspect.getmembers(mod, lambda c: inspect.isclass(c) and issubclass(c, MatPropsMaterial))
-        materialsList = [material[0] for material in materialsList]
-        if name in materialsList:
-            return getattr(mod, name)
+        if namespace.startswith("venv:") or namespace.startswith("dir:"):
```
opotowsky, line 280: ah I see what you're getting up to....I am liking this so far!

armi/materials/__init__.py
```diff
@@ -198,9 +193,9 @@ def iterAllMaterialClassesInNamespace(namespace):
                 yield obj
 
 
-def resolveMaterialClassByName(name: str, namespaceOrder: List[str] = None):
+def createMaterialByName(name: str, namespaceOrder: List[str] = None):
```
opotowsky, line 217: I hate long names, but we are basically grabbing the material given a namespace although that arg is optional. `createMaterialByName` doesn't capture the namespace aspect. `resolve` in the prior function did somewhat. 

Tossing out ideas: 

- Could we keep the original name resolveMaterialClassByName?
- createMaterialByNameAndNamespace
- supercalifragilisticexpialidociousMaterial
- I dunno, it's my friday

john-science: > Could we keep the original name resolveMaterialClassByName?

I'm happy to change the name here. But... the difference here is that method used to return a "class" but now it returns an "instance" of that class. 

I swear, this is a necessary difference. It was soooo annoying to figure out.

armi/materials/__init__.py
john-science: yikes. Sorry. 

Done.

armi/materials/__init__.py
opotowsky: how about `getMaterialFromMaterialNamespace`?

john-science: `getMaterialFromMaterialNamespaces` plural, because that's the goal?

Or, you know... just `getMaterial()`? Is that better?

opotowsky: > I swear, this is a necessary difference. It was soooo annoying to figure out.

I bet that was!!!

opotowsky: > `getMaterialFromMaterialNamespaces` plural, because that's the goal?
> 
> Or, you know... just `getMaterial()`? Is that better?

yes!

opotowsky: I think this isn't important enough to hold up the PR over. matProps also has a `getMaterial` so I'm not sure it's a good idea for armi.materials to also have one

opotowsky (approved):

armi/matProps/__init__.py
```diff
@@ -82,16 +82,15 @@
 """
 
 import os
-import sysconfig
 from glob import glob
 
 from armi.matProps.material import MatPropsMaterial
 
+# list of all the directories of purely matProps materials that have been loaded
```
opotowsky, line 91: I'm not seeing downstream pure matProps materials get added to this. That isn't the point of this PR but it is a point of confusion for me

If it's not needed we can just delete it

armi/materials/__init__.py
```diff
@@ -39,112 +39,96 @@
...
-from armi.matProps import addMaterial, clear, loadedRootDirs
+from armi.matProps import MatPropsMaterial
 from armi.matProps import getPaths as getYamlPaths
 
 # This can be updated by the CONF_MATERIAL_NAMESPACE_ORDER setting during reactor construction (see
 # armi.reactor.reactors.factory).
 _MATERIAL_NAMESPACE_ORDER = ["armi.materials"]
 
+# dictionary of loaded materials: d[mat directory][mat name] = instance of Material object
+_loadedYamlDirs = {}
```
opotowsky: it's a point of confusion for me the difference between `loadedRootDirs` in matProps and `_loadedYamlDirs` in materials.

also this lists loaded root dirs but figuring out this with the namespace order might be important?

opotowsky: e.g. if youre inspecting this, you could have: `{"dir1": {"mat1": mat1, "mat2": mat2}, "dir2": {"mat1": mat1, "mat3": mat3}}`, and you wouldn't know which `mat1` would be used

opotowsky: piling on more questions. 

I'm working in a downstream project with a materials plugin. I have the following:

`namespace = [plugin.materials, venv:yaml_dir, armi.materials]`

After the materials are loaded, I'd expect to see:
`_loadedYamlDirs = {"path/to/yaml_dir": {mats1}, "path/to/armi/yamls": {mats2}}`

But I only see `_loadedYamlDirs = {"path/to/yaml_dir": {mats1})`

armi/materials/__init__.py
```diff
@@ -39,112 +39,96 @@
...
     # recursively get all the *.yaml and *.yml files from the provided directory
+    _loadedYamlDirs[dirPath] = {}
     paths = getYamlPaths(dirPath)
     for yamlPath in paths:
-        # tests this is a valid material file AND preps for adding to matProp registry
         mat = Material()
         try:
             mat.loadFile(yamlPath)
         except Exception:
             continue
```
albeanth: (out of my lane, but am going to leave a note cause I don't understand this) 

this feels odd. within this try/except there is a "partial data check of the material data file" in material.py::dataCheckMaterialFile that has specific catches for Value and Key errors. Why would you want to silently continue on other failures?

armi/materials/__init__.py
```diff
@@ -39,112 +39,96 @@
...
 
-        # add the material to two namespaces, to support different user workflows
-        addMaterial(yamlPath, mat)
-        addYamlMaterialToThisNamespace(yamlPath, overwriteExisting=overwriteExisting)
+        # If a class with this name already exists in the package, continue
+        _loadedYamlDirs[dirPath][mat.name] = mat
+
+
+def setMaterialNamespaceOrder(order):
```
alexhjames, line 108: Can we add a method for seeing what the material namespace order is? Now that we are unifying around setMaterialNamespaceOrder as the way to control material selection, it is useful to be able to see the current order in testing without having to access the private variable `_MATERIAL_NAMESPACE_ORDER`. In particular, very useful in writing a pytest fixture to ensure it stays set at the right value across tests. 

fyi - @opotowsky

opotowsky: I feel the same way about `_loadedYamlDirs`. 

Hopefully a couple of helper functions is all we need.

armi/materials/__init__.py
opotowsky: One more pile on:

Since this is the format `_loadedYamlDirs[<MAT DB PATH>][<MAT NAME>] = <MAT OBJ>`

I'd argue that the name is a little misleading and it should be more focused on the fact that the material objects themselves are stored in this dict. 

Also, I didn't see this in this chain, so apologies if I'm repeating myself: `loadedRootDirs` is empty which is another confusing piece of this for me.

armi/materials/__init__.py
opotowsky: good to call this out @albeanth 

I found an issue with this code last night and Alex found an issue today. So this will get some editing once we figure out what needs to happen

armi/materials/material.py
alexhjames, line 77: This DATA_SOURCE variable doesn't appear to be updated for YAML materials loaded from a DIR or VENV

armi/materials/__init__.py
alexhjames: To document what I found, the `setDefaultMassFracs` method is called during the material constructor which runs first. For YAML materials, this means the composition is blank so there is no mass fractions set. The composition is then populated from the YAML file in `loadFile`.

armi/materials/__init__.py
opotowsky: (I added both to the PR I have into this feature branch BTW)

armi/materials/material.py
opotowsky: Yea perhaps for dir-only or venv-only materials they should get a blanket data source (armi and plugin materials will have their defined of course)

I'll put in something into my PR and we can argue about the details of it there.

alexhjames: Shouldn't the source just be the directory path or venv path that it was sourced from?

armi/materials/__init__.py
john-science: TODO: Document this vs the global in armi.matProps better

armi/matProps/__init__.py
```diff
@@ -82,16 +82,15 @@
...
 loadedRootDirs = []
```
john-science, line 92: TODO: Better document the different with the similar global in armi.materials. After Arrielle's PR merges.

armi/matProps/__init__.py
john-science: All of this stuff is only needed for people who are using matProps directly, without ARMI.

As we talked about earlier, I will document that once your PR merges into mine.

I only know of one downstream customer that is using matProps directly without the rest of ARMI right now. But, still, that one customer is sufficient to have me keep supporting this. At least for now.

armi/materials/__init__.py
john-science: We talked about this, and I will add documentation. 

Though I am open to a better name than `_loadedYamlDirs`. Obviously. Naming things is hard.

armi/materials/__init__.py
john-science: > this feels odd.

@albeanth This check does two things:

1. Proves that the material YAML file is valid, and CAN be loaded. If someone provides a mal-formed YAML file we don't want to add it to the run. 
2. Calling `mat.loadFile()` loads the material into the ARMI material cache. Which we very much want to happen in order, when loading collections of materials in order using `setMaterialNamespaceOrder()`.

So, TLDR; I'm definitely leaving this check in. It is crucial. But I can document these two reasons in a comment (as soon as Arrielle's branch is merged into this one).

albeanth: > Proves that the material YAML file is valid, and CAN be loaded. If someone provides a mal-formed YAML file we don't want to add it to the run.

Agreed. But maybe a runLog statement saying the YAML file is bunk and not getting added? Otherwise, from my limited view, it feels like it silently doesn't add it and a user wouldn't know. (again, I'm not up to speed on all the changes, so take this with somewhat of a grain of salt...)

john-science: Yeah, the `loadFile()` method will log if it fails to load a YAML file. For sure.

albeanth: Cool. My spidey senses were tingling seeing "partial data check" and then the silent catch all except. So it felt like there are corner cases that wouldn't be caught in the "partial data check" within `loadFile` and then go silently ignored...

armi/materials/__init__.py
opotowsky: loadedYamlMaterials (cause the dict contains a file path, a name, and a mat object)?

armi/materials/__init__.py
opotowsky: So this particular try/except is removed in my PR into this feature branch. BUT the code still calls `loadFile`, just via initializing the `Material` class instead. 

So we still have the loadFile checks happening

armi/materials/material.py
opotowsky: this is addressed in my PR into this feature branch, so I will resolve this comment.

armi/materials/__init__.py
john-science: Done

armi/matProps/__init__.py
john-science: done

armi/materials/__init__.py
john-science: done

PR #2558, opened 2026-06-23
Increasing test coverage - matProps loadAll exception

yardasol:

Adds a test to cover an untested exception in `matProps.loadAll()`

## SCR Information

Change Type: features

One-Sentence Rationale: Covers an exception that with testing.

---

## Checklist

### DIFF
--- a/armi/matProps/tests/test_parsing.py
+++ b/armi/matProps/tests/test_parsing.py
@@ -15,6 +15,7 @@
 """Test YAML parsers for all files in the matProps data directory to ensure that there are no parsing errors."""
 
 import os
+import sysconfig
 import tempfile
 import unittest
 from os import path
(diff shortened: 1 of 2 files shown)


armi/matProps/tests/test_parsing.py
```diff
@@ -76,6 +77,18 @@ def test_multiDataLoadingLoadingAll(self):
         armi.matProps.clear()
         self.assertEqual(0, len(armi.matProps.loadedMaterials()))
 
+    def test_loadAllBadRootDir(self):
```
john-science, line 82: We frequently run the ARMI unit tests in parallel. So every computer runs the ARMI unit tests in a different, parallel order.

Can this change break other unit tests, if this test is run in a different order on a different machine? Thoughts?

john-science: Could we refactor the given code to have a class attribute or global variable that defined the "materials_data" string?

THEN we could use `mock` in this test to mock that change to get the result you want, without worrying about this test breaking any other tests due to parallel test woes?

Or is that crazy? I'm just thinking aloud here. Thoughts?

yardasol: Sorry for the late response on this. I was not thinking with parallel execution in mind. If there exists a test that relies on the `materials_data` directory, then it is highly likely that the implementation here will cause that test to fail.

yardasol: > Could we refactor the given code to have a class attribute or global variable that defined the "materials_data" string?
>
>THEN we could use mock in this test to mock that change to get the result you want, without worrying about this test breaking any other tests due to parallel test woes?

Love this solution, I'll give it a whirl.

yardasol: Okay, this is my first time using mock, let me know what you think @john-science. One way I think this could be made better would be to have the mock `_DEFAULT_ROOT_DIR` be a simple path that 100% will never exist, but I don't know if that's possible.

john-science: > If there exists a test that relies on the materials_data directory

Several hundred tests in ARMI do, actually. Everything that uses `loadTestReactor()`.

> this is my first time using mock

Awesome, than this exercise was well worth the time.

john-science (approved):
Love it!

Thanks for the revision.

PR #2568, opened 2026-07-05
Suggested edits to namespace order work

opotowsky:

Adding some helper functions in to access some global variables, and adding some logic in to `loadTestReactor` to allow for a change in material namespace order should the unit test demand it.

## SCR Information

One-Sentence Rationale: TBD

---

## Checklist

### DIFF
--- a/armi/matProps/material.py
+++ b/armi/matProps/material.py
@@ -200,6 +200,9 @@ class MatPropsMaterial:
             sha1.update(materialFile.read())
         self._sha1 = sha1.hexdigest()
 
+        if not self.YAML_PATH:
+            self.YAML_PATH = self.yamlPath
+
         # build the material, from the file
         self.dataCheckMaterialFile(yamlPath, node)
         self.loadNode(node)
--- a/armi/matProps/material.py
+++ b/armi/matProps/material.py
@@ -67,15 +67,13 @@ class MatPropsMaterial:
         self._sha1 = None
         """SHA1 value of parsed material file."""
 
-        self.yamlPath = yamlPath if yamlPath else self.YAML_PATH
-        """Path to the YAML file that was loaded to build this instance."""
+        if yamlPath:
+            self.YAML_PATH = yamlPath
 
         # Load the material, if the YAML was provided.
-        if self.yamlPath:
-            self.loadFile(self.yamlPath)
-            if not self.YAML_PATH:
-                # handle some edge-cases with how materials can be loaded
-                self.YAML_PATH = self.yamlPath
+        if self.YAML_PATH:
+            self.loadFile(self.YAML_PATH)
+
 
     def __repr__(self):
         """Provides string representation for MatPropsMaterial class."""
@@ -187,7 +185,7 @@ class MatPropsMaterial:
             Path containing name of YAML file to parse.
         """
         # load the file path
-        self.yamlPath = yamlPath
+        self.YAML_PATH = yamlPath
         y = YAML(pure=True)
         node = y.load(Path(yamlPath))
 
@@ -199,9 +197,6 @@ class MatPropsMaterial:
         with open(yamlPath, "rb") as materialFile:
             sha1.update(materialFile.read())
         self._sha1 = sha1.hexdigest()
-
-        if not self.YAML_PATH:
-            self.YAML_PATH = self.yamlPath
 
         # build the material, from the file
         self.dataCheckMaterialFile(yamlPath, node)
--- a/armi/matProps/material.py
+++ b/armi/matProps/material.py
@@ -183,8 +183,9 @@ class MatPropsMaterial:
         yamlPath: str
             Path containing name of YAML file to parse.
         """
-        # load the file path
+        # Set class attr
         self.YAML_PATH = yamlPath
+        # load the file path
         y = YAML(pure=True)
         node = y.load(Path(yamlPath))
 
--- a/armi/matProps/material.py
+++ b/armi/matProps/material.py
@@ -74,7 +74,6 @@ class MatPropsMaterial:
         if self.YAML_PATH:
             self.loadFile(self.YAML_PATH)
 
-
     def __repr__(self):
         """Provides string representation for MatPropsMaterial class."""
         return f"<MatPropsMaterial {self.name} {str(self.materialType)}>"
--- a/armi/materials/__init__.py
+++ b/armi/materials/__init__.py
@@ -39,6 +39,7 @@ import inspect
 import os
 import pkgutil
 import sysconfig
+from copy import deepcopy
 from typing import List
 
 from armi import runLog
@@ -289,7 +290,7 @@ def createMaterialByName(name: str, namespaceOrder: List[str] = None):
             # grab the global material, and copy it over to a new material to return
             mat0 = _loadedYamlDirs[yDir][name]
             newMat = Material()
-            newMat.__dict__.update(mat0.__dict__)
+            newMat.__dict__.update(deepcopy(mat0).__dict__)
             return newMat
         else:
             # check and see if this is an importable material
--- a/armi/materials/__init__.py
+++ b/armi/materials/__init__.py
@@ -94,7 +94,7 @@ def importYamlMaterialDir(dirPath, overwriteExisting=True, clearFirst=True):
     paths = getYamlPaths(dirPath)
     for yamlPath in paths:
         mat = Material(yamlPath=yamlPath)
-        mat.DATA_SOURCE = dirPath
+        mat.DATA_SOURCE = "venv: " + dirPath.split("site-packages")[1][1:] if "site-packages" in dirPath else dirPath
         # If a class with this name already exists in the package, continue
         _loadedYamlDirs[dirPath][mat.name] = mat
 
--- a/armi/materials/__init__.py
+++ b/armi/materials/__init__.py
@@ -93,7 +93,7 @@ def importYamlMaterialDir(dirPath, overwriteExisting=True, clearFirst=True):
     _loadedYamlDirs[dirPath] = {}
     paths = getYamlPaths(dirPath)
     for yamlPath in paths:
-        mat = Material()
+        mat = Material(yamlPath=yamlPath)
         try:
             mat.loadFile(yamlPath)
         except Exception:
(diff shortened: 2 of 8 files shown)


armi/matProps/material.py
```diff
@@ -200,6 +200,9 @@ def loadFile(self, yamlPath: str):
             sha1.update(materialFile.read())
         self._sha1 = sha1.hexdigest()
 
+        if not self.YAML_PATH:
```
alexhjames: This seems wrong. We're overwriting a class property based on what was handed into this method. That class property is probably also what got passed into the loadFile command upstream. We need to simplify/unify this property across all the different methods.

armi/testing/__init__.py
```diff
@@ -75,6 +75,11 @@ def loadTestReactor(inputFilePath=_ARMI_RUN_DIR, customSettings=None, inputFileN
         # return test reactor from cache
         o, r = pickle.loads(_TEST_REACTORS[reactorHash])
         o.reattach(r, o.cs)
+        if o.cs["materialNamespaceOrder"] != materials.getMaterialNamespaceOrder():
+            # Reload materials if the current global namespace order doesn't match the case settings namespace order
+            _resetBeforeReactorConstructionHook()
+            materials.clear()
+            getPluginManagerOrFail().hook.beforeReactorConstruction(cs=o.cs)
```
alexhjames: why do we need to re-run the whole hook instead of just setting the namespace order? The namespace order is the only global variable.

armi/testing/__init__.py
```diff
@@ -85,6 +90,9 @@ def loadTestReactor(inputFilePath=_ARMI_RUN_DIR, customSettings=None, inputFileN
     if "verbosity" not in customSettings:
         runLog.setVerbosity("error")
 
+    # Always want to reset this hook if the namespace has changed unit tests have varying namespace settings
+    if cs["materialNamespaceOrder"] != materials.getMaterialNamespaceOrder():
+        _resetBeforeReactorConstructionHook()
```
alexhjames, line 95: Since we aren't loading a whole material library every time (only if the yamls got cleared), we should consider just deleting this whole only run once thing so that people don't need to reset it in a bunch of places and also get confusing results if they forget to reset it. It should be pretty quick to just reshuffle the namespace order as needed.

armi/testing/__init__.py
opotowsky: I was thinking rerunning all plugins' beforeReactorConstruction hooks here was the right move. So far, the one other case of this downstream wouldn't need to rerun (it's not changing at this point) but I also figured better safe than sorry?

armi/testing/__init__.py
opotowsky: Yea we can remove the onlyRunOnce logic if we are OK with reloading materials (overwriteExisting/clearFirst defaults to True, so they would be reloaded)

I think the onlyRunOnce logic was meant to protect those running scripts and loading many DB steps. I haven't done performance testing so I don't know the impact. The impact for sims is minimal and is not a concern at all. Unit test: unsure.

alexhjames: I will do some timing after lunch on the namespace order setting

armi/testing/__init__.py
alexhjames: It probably doesn't hurt but also would be overkill if we're worried about test efficiency.

armi/testing/__init__.py
alexhjames: @opotowsky I vote for removing the run once. It only loads the material data into mat_props the first time you include that thing in the directory. After that it is just grabbing those materials from the mat_props loaded list if it exists. Timing wise, I am seeing a sort of slow first call when setting the namespace order, but then after that it's 1e-5 seconds per call.

armi/testing/__init__.py
john-science: Loading materials from YAML files isn't super fast on some clusters, because it is a network AND IO-limited operation.

So we don't want to duplicate such a task if we don't need to.

Having said that, this isn't happening in a loop, so it's still `O(n)` at worst.

armi/testing/__init__.py
john-science: I'm not thrilled with force-reloading the materials every time someone calls `loadTestReactor()`. We JUST made this method faster by caching test reactors. We will lose all that performance bump by reloading all the material YAML files every time this method is called.

alexhjames: Setting the material namespace order doesn't reload the material files from scratch. There is an if statement in `setMaterialNamespace` that guards against reloading `importYamlMaterialDir` if it has already been loaded. So in the case of the directory already being loaded, the `setMaterialNamespace` directory will basically just be updating the global namespace variable which is quite fast. The global material namespace variable though needs to be reset to match this reactor when bringing it back from a cache.

armi/materials/__init__.py
```diff
@@ -131,6 +130,33 @@ def setMaterialNamespaceOrder(order):
             importYamlMaterialDir(yDir)
```
alexhjames: This needs to have `clearFirst=False`. If you have multiple directories or venvs in your namespace order. The current implementation will clear the data in between loading each one.

armi/testing/__init__.py
opotowsky: yea I can just set this to be namespace order. If we were to later add other global things to the beforeReactorConstruction hook then it would be a problem but right now its not.

armi/testing/__init__.py
opotowsky: ah I see. since we arent calling importYaml directly (with overwriteExisting defaulting to true), youre right that there's no materials reloading being done. 

I'd love for that to be a separate ticket though. Removing the onlyRunOnce logic has a few deletions here and in downstream plugins, and it's a small thing to do on its own, but adding more to an already large PR if we do it here. And we could performance test it standalone.

armi/materials/__init__.py
opotowsky: correct! Need to consuder having multiple yaml dirs for sure. adding

opotowsky: https://github.com/terrapower/armi/pull/2568/commits/13de5f8f2a1fcbf455bcde8b5a6573f019cbcfc4

armi/testing/__init__.py
opotowsky: https://github.com/terrapower/armi/pull/2568/commits/dedecb8bbe18cffd49df65c5fe105851b8d4cfad

armi/testing/__init__.py
opotowsky: I did this https://github.com/terrapower/armi/pull/2568/commits/dedecb8bbe18cffd49df65c5fe105851b8d4cfad

And then I got a ton of failures in the unit tests like:

```
  File "/anfhome/team/nala/aopotowsky/nala/framework/armi/reactor/blueprints/componentBlueprint.py", line 340, in _constructMaterial
    mat = materials.createMaterialByName(self.material)
  File "/anfhome/team/nala/aopotowsky/nala/framework/armi/materials/__init__.py", line 300, in createMaterialByName
    raise KeyError(
    ...<2 lines>...
    )
KeyError: 'Cannot find material named `Custom` in any of: []. Please update inputs or plugins. See CONF_MATERIAL_NAMESPACE_ORDER setting.'
```

Reverting the commit solved the issue

alexhjames: @opotowsky the problem is that most if not all of the ARMI test reactors don't define the namespace setting so the current logic is causing the namespace order to be set to `[]` and then it can't find the material. The `beforeReactorConstruction` hook call is handling this because it defaults the namespace order if it's not set. The logic in the image fixes the test failures, or alternatively, the case setting files could be modified to define the namespace. Also, I don't think the `clear` is needed. If materials are loaded but not in the current namespace order then it's not harming anyone and speeds up things in the future if those materials come back into the namespace order.

image

armi/materials/__init__.py
```diff
@@ -90,12 +93,8 @@ def importYamlMaterialDir(dirPath, overwriteExisting=True, clearFirst=True):
...
     paths = getYamlPaths(dirPath)
     for yamlPath in paths:
-        mat = Material()
-        try:
-            mat.loadFile(yamlPath)
-        except Exception:
-            continue
-
+        mat = Material(yamlPath=yamlPath)
+        mat.DATA_SOURCE = "venv: " + dirPath.split("site-packages")[1][1:] if "site-packages" in dirPath else dirPath
```
opotowsky, line 100: I didn't love `materials_data` printing out on its own in our material table printout so now it shows `venv: materials_data` and I like that more

armi/testing/__init__.py
opotowsky: Ah, sure. In downstream plugins I have the namespace order set to a default in the conftest.py files, but of course I didn't / wouldn't do that in ARMI. So the failures here make sense. Ok I'll check this out, thanks!

opotowsky: `setMaterialNamespaceOrder` is what sets the default if the setting is missing though. So odd.

alexhjames: > `setMaterialNamespaceOrder` is what sets the default if the setting is missing though. So odd.

I don't think this is true. `setMaterialNamespaceOrder` just sets the order to be whatever you provide as an input. The initial value of the namespace order global variable includes the armi material library. The `ReactorPlugin` hook only sets the namespace order if the case setting exists. So reactors that are missing the case setting have the default still set from the global variable initialization. In this case, the code is explicitly grabbing the namespace order setting (even if it's not set) and handing to the setter which forcibly sets the namespace order to blank.

armi/testing/__init__.py
opotowsky: oops I continued this convo below. I'll resolve this and follow up there

armi/testing/__init__.py
opotowsky: This works: https://github.com/terrapower/armi/pull/2568/commits/163bc3e3d0ccda0579b18d29c86a16875f0f1d01

Here's what I meant though:
image

where:
image

but I see that global isn't USED it's just updated based on the input, which is why I had the failures I was having

john-science (approved):

armi/materials/__init__.py
john-science: Oh, this is here. Huh.

I was assuming this would go in the code that actually prints out this table. 

huh

opotowsky: the table uses the DATA_SOURCE 

And DATA_SOURCE wasn't getting updated for yaml materials (they said "ARMI") so this was a necessary line of code to add, and I just fiddled with how I wanted it to look

PR #2572, opened 2026-07-07
Adding a plugin hook for defining a material base class

alexhjames:

Add a plugin hook that allows for defining the material base class for materials that are loaded from a YAML. This allows more customization downstream for materials if required.

## SCR Information

Change Type: features

One-Sentence Rationale: Allow ARMI apps to customize material behavior while still using the YAML format without having to define a wrapper class for each material individually.

---

## Checklist


armi/materials/material.py
```diff
@@ -324,7 +325,10 @@ def applyInputParams(self, **kwargs) -> None:
             td = kwargs["TD_frac"]
             if td is not None:
                 if td > 1.0 or td <= 0.0:
-                    runLog.warning(f"Theoretical density frac for {self} is out of range: {td}", single=True)
+                    runLog.warning(
```
john-science: Can you move the formatting back to the one-line version?

ARMI has `ruff` set to do lines of 120 characters, so this was fine as it was.

armi/materials/material.py
```diff
@@ -114,8 +114,9 @@ def __init__(self, yamlPath=None):
             # This material does not have a YAML file to pull the name from.
             self.name = self.__class__.__name__
 
-        # call subclass implementations
-        self.setDefaultMassFracs()
+        if updateMassFracs:
```
john-science, line 122: How is this change related to the material base class change?

armi/materials/__init__.py
```diff
@@ -75,6 +75,9 @@ def importYamlMaterialDir(dirPath, overwriteExisting=True, clearFirst=True):
         A popular safety option is to first clear out the YAML materials loaded into memory before loading new ones.
         This is particularly popular during unit testing.
     """
+    # Needs to be a local import to prevent circular imports
+    from armi import getPluginManager
```
john-science, line 83: Can this not be at the top of the file? Is there a circular import loop problem?

armi/materials/__init__.py
```diff
@@ -286,7 +301,14 @@ def createMaterialByName(name: str, namespaceOrder: List[str] = None):
 
             # grab the global material, and copy it over to a new material to return
             mat0 = _loadedYamlDirs[yDir][name]
-            newMat = Material()
+            pm = getPluginManager()
+            baseClass = Material
+            if pm:
+                baseClassList = getPluginManager().hook.setMaterialBaseClass(materialType=mat0.materialType)
```
john-science, line 314: This probably deserves a one-line comment.

armi/materials/material.py
alexhjames: I ran `ruff --format .` on the repo before committing. Maybe I have a slightly different version number of ruff installed, I will double check.

armi/materials/material.py
alexhjames: When `importYamlMaterialDir` runs, it instantiates a new instance of the material and then copies the `__dict__` of the already loaded material over to that new instance. `setDefaultMassFracs` gets run during the constructor of that new instance but then has issues (ARMI's version doesn't but other downstream ones do) because the composition is blank at that point because the composition has not been copied over yet during the `__dict__` copy. I added this flag so that `importYamlMaterialDir` could call the constructor with the flag set to false, bypass the mass fraction calculation in the constructor, and then the mass fractions will be copied over in the `__dict__` copy.

armi/materials/__init__.py
alexhjames: There is a circular import problem. The base level ARMI init imports `reactors` which imports `materials`.

armi/materials/__init__.py
alexhjames: I'll add some comments

armi/materials/__init__.py
```diff
@@ -94,10 +96,17 @@ def importYamlMaterialDir(dirPath, overwriteExisting=True, clearFirst=True):
     _loadedYamlDirs[dirPath] = {}
     paths = getYamlPaths(dirPath)
     for yamlPath in paths:
+        # This makes us load a given YAML twice per material if they have a custom class
+        # but I can't think of a better way to get the type from the material file.
```
opotowsky, line 104: I'm thinking this could make the unit tests take a hit but maybe not so much simulations

armi/materials/material.py
opotowsky: I'm guessing it's a text editor setting. And since it doesn't violate ruff's settings, it isn't reformatted upon `ruff format .` I'll push up reformats when I'm done reviewing

armi/materials/material.py
opotowsky, line 103: I removed the deepcopy here in the last materials PR because it caused a slowdown.

If `updateMassFracs` is true, then this isn't a problem. But it could be a problem is it's false. The unit tests that shows this issue downstream did not fail when I removed it (since they were using the `createMaterialByName` pathway and the deepcopy that you added  there isn't slowing down code)

opotowsky (commented):
Looking good. we still have a question bleeding over from the last bit of work but otherwise I'm happy with this.

armi/materials/__init__.py
alexhjames: I think the yaml directories only get loaded once? We don't actually ever clear the loaded yaml directories even if they temporarily leave the namespace. Then everything after that is using the cached version of the material, not re-loading from yaml file.

opotowsky: I've confirmed no slowdowns in unit tests

armi/materials/material.py
opotowsky: deepcopy fixed

opotowsky (approved):

Issue #2565, opened 2026-06-25
Inconsistencies in unit testing of `armi/reactor/blocks.py`


yardasol:
The [PR I made recently](https://github.com/terrapower/armi/pull/2564#issuecomment-4801966805) sent me on a deep dive into the unit testing of the `blocks.py` module, and I've discovered some potential issues with it:
1. There are inconsistent test class names in the module:
  -  Some test classes use the `Test` suffix: `TestHexBlock`, `TestValidateSPFSpatialGrids`, et.c
  -  Others don't: `MassConservationTests`, `CartesianBlockTests`, `ThRZBlock_TestCase`, etc.
2. It appears that not all functions are unit tested
  - Example: [coverage claims that `HexBlock.verifyBlockDims()`](https://github.com/terrapower/armi/pull/2564#issuecomment-4801966805) is covered, but the only test of this kind in `test_block.py` is for `ThRZBlock.verifyBlockDims()`

To resolve:
1. The naming convention should be consistent and follow the [unit test](https://terrapower.github.io/armi/developer/standards_and_practices.html#unit-tests) and [naming conventions](https://terrapower.github.io/armi/developer/standards_and_practices.html#unit-tests)
2. Every function `BlockSubclass.x()` should have a unit test `test_Subclass_x()` (the syntax doesn't have to be exactly this of course).

john-science:
Good find!

Yeah, I think direct unit testing of a method is generally preferable to indirect ("don't worry, something else is giving us code coverage for that method") testing.

john-science:
Beware, some people import ARMI tests downstream, so I think we should pick a naming convention that requires the least work. In ARMI right now the county of test classes looks something like this:

* 330 `TestXYZ`
* 55 `XYZTests`
* 51 `XYZ_TestCase`

So, @yardasol and I decided will should standardize on the `TestXZY` naming convention.  (Also, we should document this convention in the Developer docs.)

Be warned, some people have been known to import ARMI's unit tests directly downstream. This is a bad practice, but it's been happening.

drewj-tp:
Two cents: `TestXYZ` and certainly not `XYZ_TestCase` please

yardasol:
superceded by #2570, #2602, and #2603

PR #2579, opened 2026-07-09
Improving code coverage of matProps to 100%

john-science:

The matProps module was at nearly 100% code coverage. This gets the code coverage all the way to 100%. Hurray.

## SCR Information

Change Type: trivial

One-Sentence Rationale: Improved code coverage leads to more reliable code.

---

## Checklist

### DIFF
--- a/armi/matProps/material.py
+++ b/armi/matProps/material.py
@@ -48,10 +48,6 @@ class MatPropsMaterial:
 
     def __init__(self, yamlPath=None):
         """Constructor for MatPropsMaterial class."""
-        # during unpickling, we do not want to reload the YAML file
-        if hasattr(self, "materialType") and self.materialType is not None:
-            return
-
         self._saved = False
         """Boolean denoting whether or not MatPropsMaterial object is saved in materials dict."""
 
--- a/armi/matProps/tests/test_parsing.py
+++ b/armi/matProps/tests/test_parsing.py
@@ -23,6 +23,8 @@ from unittest.mock import patch
 
 import armi.matProps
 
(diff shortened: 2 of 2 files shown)


armi/matProps/material.py
```diff
@@ -48,10 +48,6 @@ class MatPropsMaterial:
 
     def __init__(self, yamlPath=None):
         """Constructor for MatPropsMaterial class."""
-        # during unpickling, we do not want to reload the YAML file
-        if hasattr(self, "materialType") and self.materialType is not None:
-            return
-
```
yardasol, line 54: Can you explain the rationale for deleting this instead of adding a test for it?

yardasol (commented):
Nice test @john-science! I have one question I'd like clarification on before approving. See comment below.

armi/matProps/material.py
john-science: I removed it and then tested and all the tests passed. So, it must not be necessary.

And if it's a performance concern, as my comment suggests, that was easy. I ran a bunch of tests that pickle the reactor with and without this change and found no performance difference.

So, I figure this is no longer necessary. 

Thoughts?

yardasol: > I removed it and then tested and all the tests passed. So, it must not be necessary.

Assuming everything in ARMI is 100% tested, yes, but as I found in #2565 this isn't necessarily the case...

If you checked cases with reactor pickling and found no performance difference, then that's sufficient for me to sign off on this. Great work!

yardasol (approved):

PR #2585, opened 2026-07-14
Moving buildOperatorOfEmptyHexBlocks to armi.testing

yardasol:

This PR moves the functions `buildEmptyOperatorOfHexBlocks()` and `buildEmptyOperatorOfCartesianBlocks()`, previously stored in `armi.reactors.tests.test_reactors` to `armi.testing.buildEmptyOperators`. This is done to allow downstream modules to prevent downstream plugins from importing directly from the ARMI testing suite.

## SCR Information

Change Type: trivial

One-Sentence Rationale: Improving testing suite makes downstream code more resilient.

---

## Checklist


armi/testing/buildEmptyOperators.py
```diff
@@ -0,0 +1,104 @@
...
+whether or not one of the specific methods were called. These should only be used in testing.
+"""
+
+from armi import operators, settings, tests
+from armi.materials import uZr
+from armi.reactor import assemblies, blocks, grids
+from armi.reactor.components import Hexagon, Rectangle
+
+
+def buildOperatorOfEmptyHexBlocks(customSettings=None):
```
john-science: If this is going to be in it's own file, can we import this function (and the one below) into `armi/testing/__init__.py`?

My idea here is that it is infinitely easier for downstream developers to remember to import the testing tool they want directly from `armi.testing`, than to have to remember more path after that.

(It'll shorten people's downstream imports, and they will like that.)

armi/physics/neutronics/fissionProductModel/tests/test_lumpedFissionProduct.py
```diff
@@ -30,8 +30,8 @@
     CONF_LFP_COMPOSITION_FILE_PATH,
 )
 from armi.reactor.flags import Flags
-from armi.reactor.tests.test_reactors import buildOperatorOfEmptyHexBlocks
 from armi.settings import Settings
+from armi.testing.buildEmptyOperators import buildOperatorOfEmptyHexBlocks
```
john-science: The API would be cleaner if this could be written:

```python
from armi.testingimport buildOperatorOfEmptyHexBlocks
```

See my comment below.

armi/testing/buildEmptyOperators.py
john-science: Pro Tip: `ruff` will try to clean up any unused imports, so in that `__init__.py` file, you will have to do this:

```python
from armi.testing.buildEmptyOperators import X, Y # noqa: F401
```

john-science (approved):
Nice!

yardasol:
Downstream testing has been performed, and relevant plugins have been shown to pass the test suite.

PR #2586, opened 2026-07-14
Redesigning our codecov.io config

john-science:

Feel free to double-check me on the [codecov.io docs](https://docs.codecov.com/docs/common-recipe-list).

In this PR I set the new codecov.io custom cofig file (`.github/.codecov.yml) to these settings:

Stop reporting on, or caring about, the code coverage from the PR difference, we only care about the coverage of the entire repo:

```yaml
coverage:
    patch:
      default:
        informational: true
```

Our bare minimum code coverage for ARMI is 90%, but codecov.io is WAY to finicky and will fail the build if the coverage goes down even 0.02%, which is annoying and wrong:

```yaml
coverage:
  status:
    project:
      default:
        target: 90%
        threshold: 99%
```

Stop building in-github view of the coverage (we use the codecov.io website anyway):

```yaml
github_checks:
  annotations: false
```

progress on #2471

**NOTE**: If this doesn't work, I think we can go with the nuclear option of forcing codecov.io to stop complaining about coverage numbers no matter what:

```yaml
coverage:
  status:
    project:
      default:
        informational: true
```

Thoughts?

## SCR Information

Change Type: trivial

One-Sentence Rationale: We need our CI to be stable to fight alarm fatigue

---

## Checklist

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -4,5 +4,11 @@ coverage:
   status:
     project:
       default:
-        target: 1%    # the required coverage value: stop codecov.io from failing the build
+        target: 90%     # the required coverage value: 90%
         threshold: 99%  # allows a 99% drop from the previous base commit coverage
+    patch:
+      default:
+        informational: true  # set non-blocking status check
+github_checks:
+  # https://docs.codecov.com/docs/common-recipe-list#disable-github-check-run-annotations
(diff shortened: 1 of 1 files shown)


opotowsky (approved):
Your logic is sound, so lets give it a shot. I haven't done my own research into this.

Issue #2378, opened 2025-12-09
Remove Nuclide Flags


john-science:
@drewj-tp and @keckler suggest that we remove the nuclide flags in feature in ARMI. They put forward the notion that Nuclide Flags do not add value for them, and frequently slow them down.

There are three options for every nuclide flag:

https://github.com/terrapower/armi/blob/0e830d2c8dece43c57b657caefaed8e8c7ba276b/armi/testing/reactors/smallHexReactor/smallHexReactor-bp.yaml#L2-L6

> The question is: Is it possible these are only used by one downstream plugin and can be safely moved from ARMI to that project?

https://github.com/terrapower/armi/blob/0e830d2c8dece43c57b657caefaed8e8c7ba276b/armi/reactor/blueprints/isotopicOptions.py#L82-L91

> Second Question: Is it possible we could solve the user problem by just auto-defining defaults so the system never complains at you for missing nuclides in the list?

partofthething:
No doubt they can be a pain. I will just say that it's useful to have some way of telling ARMI how to map its compositions into different nuclides for representations in different neutronics systems. The main issue is that different nuclear code data people often build data libraries in weird and unpredictable ways. Notional examples:

* ENDF/B-VII.0 in MC**2-3 has elemental carbon but no C-12, but in MCNP it's the opposite, and in DRAGON it's something else, and in CASMO something else entirely!
* ENDF/B-VIII.1 has isotopic C-12 but no elemental, except in OpenMC and SERPENT
* Some libs have C-12 but no C-13 or elemental
* Do you want to model boron depletion or not?
* Do you want to model tritium production or not?
* Do you want the iron in your steel to activate to literally hundreds of trace nuclides? If so, why are you willing to run your code _so slowly?_

Another thing nuclide flags does is allow you to define the size of your transmutation/decay chain. If you want to model all U-238 going directly to Pu-239 you can do that, but if you have a high power density reactor and want to model U-239 and  Np-239 along the way, then you can do that by including them in your nuclide flags. To know if you need these levels of fidelity, you need to estimate the modeling sensitivity to these kinds of effects, and having a method by which to configure it has been useful.

Fundamentally it's about the tradeoff between computational efficiency, availability of nuclear data, and physical fidelity. If you model all decay chains and activation pathways by default, you'll have hydrogen activating all the way up to uranium justl ike in nature, but your code will run slowly.

[...]

PR #2589, opened 2026-07-16
Updating default nuclide flags

alexhjames:

Update default nuclide flags to make test reactor work with trace elements. 

## SCR Information

Change Type: features

One-Sentence Rationale: New trace elements are causing the test reactor to fail. 

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/isotopicOptions.py
+++ b/armi/reactor/blueprints/isotopicOptions.py
@@ -429,7 +429,7 @@ def getDefaultNuclideFlags():
     for boron in [10, 11]:
         nuclideFlags[f"B{boron}"] = {"burn": False, "xs": True, "expandTo": None}
 
-    for struct in ["ZR", "C", "SI", "V", "CR", "MN", "FE", "NI", "MO", "W", "NA", "HE", "AL", "CO", "NB"]:
+    for struct in ["ZR", "C", "SI", "V", "CR", "MN", "FE", "NI", "MO", "W", "NA", "HE", "AL", "CO", "NB", "CU", "N"]:
         nuclideFlags[struct] = {"burn": False, "xs": True, "expandTo": None}
 
     return nuclideFlags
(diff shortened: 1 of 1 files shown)


keckler:
I will once again chime in to note how tedious the nuclide flags are.

A search through our tickets shows how many times nuclide flags pop up as an annoyance or issue: https://github.com/terrapower/armi/issues?q=nuclide%20flags

john-science:
> I will once again chime in to note how tedious the nuclide flags are.

@keckler  What do you think of me adding in default values to our nuclide flags, because they are nearly always the same thing anyway? Does that solve all the practical problems?

john-science:
@alexhjames I mean, the ARMI test reactors are still working, the ARMI CI is flowing. Why is this change necessary?

Okay, you are using the ARMI test reactor downstream, but with non-ARMI materials, so you have to modify the test reactor?

opotowsky:
> @alexhjames I mean, the ARMI test reactors are still working, the ARMI CI is flowing. Why is this change necessary?
> 
> Okay, you are using the ARMI test reactor downstream, but with non-ARMI materials, so you have to modify the test reactor?

Correct, this is because most downstream plugins rely on the ARMI test reactor, which relies on `getDefaultNuclideFlags`. This has happened in the past where we update some material definition (like adding an impurity to a steel) and the test reactor cant load because those elements are not represented via nuclide flags.

drewj-tp:
> > I will once again chime in to note how tedious the nuclide flags are.
> 
> @keckler What do you think of me adding in default values to our nuclide flags, because they are nearly always the same thing anyway? Does that solve all the practical problems?

I contend they should not be required and/or be handled by applications

#2378

john-science (approved):
Okay, sure. Fine.

But only because I am working on make nuclide flags defaultable.

opotowsky:
> Okay, sure. Fine.
> 
> But only because I am working on make nuclide flags defaultable.

Agreed. this cycle of randomly adding elements gives me the ick. I very much so look forward to next gen nuclide flags!

keckler:
> > I will once again chime in to note how tedious the nuclide flags are.
> 
> @keckler What do you think of me adding in default values to our nuclide flags, because they are nearly always the same thing anyway? Does that solve all the practical problems?

Here is a relevant snippet from our docs. Note the "tip" at the bottom.

image

PR #2598, opened 2026-07-21
Documenting physics parameters retained on shuffling

### LINKED ISSUE #1890: Some parameters after shuffing and rotating don't match reactor's state
During rotation from shuffling and moving assemblies, some parameters will be wrong because they don't match the state of the reactor. And that's acceptable for now.

_Originally posted by @drewj-tp in https://github.com/terrapower/armi/issues/1877#issuecomment-2356575290_

If you grab an assembly and move it to a new location and don't edit any physics parameters, those parameters don't match the state of the reactor. Same for rotating an assembly. Things like `linPowByPin` and `cornerFastFlux` and even `flux` reflect the previous location and/or orientation of that assembly. It's possible to rotate an assembly and then grab physics parameters that _look_ valid (because they have data) but don't reflect the current state of the reactor.

This is related to #1860 in that we need to rotate more parameters when we rotate assemblies and blocks. Both for shuffling and for `growToFullCore` when modeling a third core reactor and then expanding it to full core.

# What to do

How can we signify that some parameters should not be used following a rotation or translation of an assembly? Separate from symmetric reflection and expansion. If we set them to arrays of zero or `None`, that's a way to invalidate the data.

drewj-tp:

Add documentation to explain that physics parameters are not cleared out upon shuffling. This is generally okay as you're likely to run your physics stack again after shuffling.

Closes #1890

## SCR Information

Change Type: docs

One-Sentence Rationale: Document retainment of physics parameters on shuffling

---

## Checklist

- [x] [Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify any new/changed code.~
- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html)~
- [x] The dependencies are still up-to-date in `pyproject.toml`~

### DIFF
--- a/doc/user/spatial_block_data.rst
+++ b/doc/user/spatial_block_data.rst
@@ -329,3 +329,17 @@ Other parameters may be updated to reflect some geometric state. The second posi
 ``Block.p.orientation`` reflects the cumulative rotation around the z-axis and is updated through
 rotation. Displacement parameters like ``Block.p.displacementX`` are updated as the displacement
 vector rotates through space.
+
+Assembly Shuffling
+==================
+
+ARMI can be configured to move assemblies during an outage. See :ref:`fuel-management-input`.
+This can be further configured to rotate hexagonal assemblies via the ``assemblyRotationAlgorithm`` setting.
+During rotation, spatial parameters will be updated as mentioned above to match the current
+orientation.
+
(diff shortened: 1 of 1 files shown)


john-science (approved):
Sure, it's a fine warning to include.

PR #2604, opened 2026-07-23
Covering up a ruamel.yaml bug with zero values

### LINKED ISSUE #2600: `db.loadBlueprints` is converting 0.0 to strings
Blueprint material modifications that are set to 0.0 (e.g. `ZR_wt_frac`) when loaded and then rewritten to a database originate as floats but are written back as strings. This fails when one attempts to load that newly written DB.

We just haven't run into this before because no one was using this workflow on a DB that had mat mods as 0.0.

From Alex:

> It appears to be that the process of converting a string to a dictionary or dictionary back to a string with the yamlize stuff is causing `0.0` to go to `'0.0'` somewhere in the process in the list of weight fractions. The weight fractions are stored as `yamlize.Sequence` objects.

john-science:

There is a bug in one of our dependencies, `ruamel.yaml`, that is causing some of our blueprints to be incorrectly translated. If we have a list with _just_ empty strings and zeros, the zeros will be passed back out as strings `"0.0"`, instead of floats.

This happens when we do `Blueprint.dump()`, which calls a `yamlize` method, which calls a `ruamel.yaml` method:

https://github.com/SimplyKnownAsG/yamlize/blob/fd4a97dc6b20e4422ab5b59658b11b7e7e6f0d1f/yamlize/yamlizable.py#L58

which has a known bug (see the ticket for more information). I tried replacing the `ruamel.yaml.RoundTripDumper()` that `yamlize` uses with the default `ruamel.yaml.Dumper`, but that didn't solve the problem. So this means that the problem exists in all `ruamel.yaml.Dumber` subclasses. There is no way around their bug.

Until `ruamel.yaml` fixes their bug, I am stuck hot-fixing a solution into ARMI. It'll be a little slow and a little ugly, but there aren't any other good options.

**NOTE**: I added a unit test that proves this error exists with nothing more than a `Blueprints.dump()` call: [proof](https://github.com/terrapower/armi/actions/runs/30047323181/job/89341417758?pr=2604#step:5:397).

close #2600

## SCR Information

Change Type: fixes

One-Sentence Rationale: In particular, this bug arose when people were doing material modifications in the blueprints.

---

## Checklist


armi/reactor/blueprints/__init__.py
```diff
@@ -514,14 +513,37 @@ def migrate(cls, inp: typing.TextIO):
...
+    def dump(cls, data, stream=None):
+        """A wrapper around the yamlize.Object.dump.
+
+        With the release of ruamel.yaml 0.19.1, we began to get an error where lists that include only empty and 0.0
+        values incorrectly for the zero values to zero strings: '0.0'.
+        """
+        super().dump(data, stream=stream)
+
+        if stream is None or isinstance(stream, io.TextIOWrapper):
+            return
```
drewj-tp, line 524: Don't we need to return something? When we do `Blueprints.dump(bp)`, where does the dumped object end up? Shouldn't we act on the dumped value and return the cleaned version?

1. If we dump to string, return the cleaned string
2. If we dump to a buffer, ensure the buffer is cleaned

Current behavior now
1. `Blueprints.dump(bp) -> str` containing the exported yaml
2. `Blueprints.dump(bp, buffer:=io.StringIO) -> None` and `buffer` contains the contents

armi/reactor/blueprints/tests/test_blueprints.py
```diff
@@ -158,6 +161,44 @@ def test_traceNuclides(self):
         self.assertLess(fuel.getNumberDensity("AM241"), 1e-5)
 
 
+class TestRuamelYamlBug(unittest.TestCase):
+    def test_ruamelYamlBug(self):
+        """Testing that we are correctly fixing a bug in ruamel.yaml.
+
```
opotowsky, line 166:
```suggestion
```

nitpick

armi/reactor/blueprints/__init__.py
```diff
@@ -514,14 +513,37 @@ def migrate(cls, inp: typing.TextIO):
...
+
+        stream.seek(0)
+        txt = stream.read()
+        txt = txt.replace("'0.0'", "0.0")
+        txt = txt.replace('"0.0"', "0.0")
+        stream.seek(0)
+        stream.truncate(0)
```
drewj-tp: Comments about the seeking / truncation would be helpful but not necessary

armi/reactor/blueprints/__init__.py
```diff
@@ -514,14 +513,37 @@ def migrate(cls, inp: typing.TextIO):
...
+
+        # Loop through the entire text file, to attempt to do the cleaning
```
drewj-tp: Comment occurs after the cleaning is done

opotowsky (approved):

armi/reactor/blueprints/tests/test_blueprints.py
```diff
@@ -158,6 +161,44 @@ def test_traceNuclides(self):
         self.assertLess(fuel.getNumberDensity("AM241"), 1e-5)
 
 
+class TestRuamelYamlBug(unittest.TestCase):
+    def test_ruamelYamlBug(self):
```
drewj-tp, line 164: A test that shows similar behavior if we don't pass a buffer would be appreciated. The buffer approach was how I reproduced the bug, but I don't know if that's the common behavior

armi/reactor/blueprints/tests/test_blueprints.py
```diff
@@ -158,6 +161,44 @@ def test_traceNuclides(self):
...
+            buffer.seek(0)
+
+            # load that string buffer into a dictionary
+            rawy = YAML().load(buffer)
+            lst = rawy["assemblies"]["igniter fuel"]["material modifications"]["ZR_wt_frac"]
+
+            # ensure the zeros are floats, not strings
+            self.assertEqual(lst[1], 0.0)
+            self.assertEqual(lst[2], 0.0)
+            self.assertEqual(lst[3], 0.0)
```
drewj-tp: slightly more informative name. non-blocking
```suggestion
            modFracs = rawy["assemblies"]["igniter fuel"]["material modifications"]["ZR_wt_frac"]

            # ensure the zeros are floats, not strings
            self.assertEqual(modFracs[1], 0.0)
            self.assertEqual(modFracs[2], 0.0)
            self.assertEqual(modFracs[3], 0.0)
```

armi/reactor/blueprints/__init__.py
```diff
@@ -514,14 +513,37 @@ def migrate(cls, inp: typing.TextIO):
                 inp = mig.apply()
         return inp
 
+    @classmethod
+    def dump(cls, data, stream=None):
+        """A wrapper around the yamlize.Object.dump.
+
+        With the release of ruamel.yaml 0.19.1, we began to get an error where lists that include only empty and 0.0
+        values incorrectly for the zero values to zero strings: '0.0'.
```
drewj-tp, line 519: If all our problems were introduced in ruamel.yaml 0.19.1, do we have a strong reason to reduce the dependency? It was not clear from the original PR that pinned the version why this version specifically was chosen

From #2405, it looks like 0.19.1 introduced some bugs that we needed to patch around. Why not stick with 0.19.0?

drewj-tp (changes requested):

armi/reactor/blueprints/__init__.py
john-science: Done

armi/reactor/blueprints/__init__.py
john-science: Done

armi/reactor/blueprints/__init__.py
john-science: We can't use 0.19.0 because it is has a bug that breaks things in an even much larger way.

And we can't roll back too far in versions of ruamel.yaml because at some point it doesn't support python 3.13.

Sorry.

armi/reactor/blueprints/__init__.py
john-science: > Don't we need to return something?

It looks to me like no. We only use this method to dump to a stream.  We aren't using `Blueprints.dump()` to return anything. Only as part of dumping the contents of the blueprints into a text string.

We COULD return something from this method, but that doesn't work in certain circumstances. Like when we are passing the stream of a write-only file handler. For such an example, take a look at the yamlize version:

https://github.com/SimplyKnownAsG/yamlize/blob/fd4a97dc6b20e4422ab5b59658b11b7e7e6f0d1f/yamlize/yamlizable.py#L58

armi/reactor/blueprints/tests/test_blueprints.py
john-science: If we don't pass in a buffer... what does this method do?

I would need to know what else people are using ARMI's `Blueprints.dump()` _for_ in order to support it, you know?

armi/reactor/blueprints/__init__.py
drewj-tp: Okay. Please make stream a required argument and document as such. Otherwise, if people do `Blueprints.dump(bp)` they will still encounter this bug.

armi/reactor/blueprints/tests/test_blueprints.py
drewj-tp: > If we don't pass in a buffer... what does this method do?

It returns the dumped string. Maybe that's not a thing anyone does that we can see. Related to my other comment about `Blueprints.dump(bp)` being how the method _could_ have worked before. It doesn't look like ARMI does this internally. 

If we're only going allow `Blueprints.dump(bp, buffer)`, then we can test that case alone so long as we enforce passing a buffer.

armi/reactor/blueprints/__init__.py
john-science: Done!

armi/reactor/blueprints/tests/test_blueprints.py
john-science: Done.

drewj-tp (approved):

PR #2607, opened 2026-07-27
Fixing Database.genTimeSteps for EOL time steps

### LINKED ISSUE #1960: Further issues found with the EOL being written to the Database
@albeanth and I identified two places where the ARMI codebase does not support the new `EOL` time point being written to the database:

1. `Database3.genTimeSteps()`:

https://github.com/terrapower/armi/blob/8b5163a8b61ebe704419d2d77067bfcf8018f328/armi/bookkeeping/db/database3.py#L612

Returns `(cycle, node)`, but if the `EOL` is present in the DB, it just returns the last time node twice.

2. `Database3.load()` takes in `cycle` and `node` as integers, thus not allowing us to load an EOL timestep:

https://github.com/terrapower/armi/blob/8b5163a8b61ebe704419d2d77067bfcf8018f328/armi/bookkeeping/db/database3.py#L702

john-science:

The method Database.genTimeSteps did not handle well the new case where there can be an "EOL" time step in the ARMI output databases.

close #1960

## SCR Information

Change Type: fixes

One-Sentence Rationale: The Database should completely support the EOL time step.

One-line Impact on Requirements: A small bug fix to I_ARMI_DB_H51.

---

## Checklist


john-science:
@albeanth and @drewj-tp, you both commented on the original bug ticket. So I'm hopeful one of you has time to review this PR.

drewj-tp:
> @albeanth and @drewj-tp, you both commented on the original bug ticket. So I'm hopeful one of you has time to review this PR.

This is on my radar. Ping me if this becomes a blocker

armi/bookkeeping/db/database.py
```diff
@@ -346,7 +346,12 @@ def splitDatabase(self, keepTimeSteps: Sequence[Tuple[int, int]], label: str) ->
             for groupName, _ in dbIn.items():
                 m = self.timeNodeGroupPattern.match(groupName)
                 if m:
-                    timeSteps.add((int(m.group(1)), int(m.group(2))))
+                    try:
+                        timeStep = (int(m.group(1)), int(m.group(2)))
+                    except ValueError:
+                        # handle the EOL time steps like c0102EOL
```
drewj-tp:
```suggestion
                        # handle the EOL time steps like c01n02EOL
```

armi/bookkeeping/db/database.py
```diff
@@ -563,14 +568,18 @@ def getLayout(self, cycle, node):
...
         assert self.h5db is not None, "Must open the database before calling genTimeSteps"
         for groupName in sorted(self.h5db.keys()):
             match = self.timeNodeGroupPattern.match(groupName)
             if match:
                 cycle = int(match.groups()[0])
-                node = int(match.groups()[1])
+                try:
+                    node = int(match.groups()[1])
+                except ValueError:
+                    # handle the EOL time steps like c0102EOL
```
drewj-tp:
```suggestion
                    # handle the EOL time steps like c01n02EOL
```

armi/bookkeeping/db/database.py
```diff
@@ -563,14 +568,18 @@ def getLayout(self, cycle, node):
...
+                    node = match.groups()[1].lstrip("0")
```
drewj-tp: Given the example above, wouldn't this return `2EOL`? [Link to pythex.org example](https://pythex.org/?regex=c%28%5Cd%5Cd%29n%28%5Cd%5Cd%5BA-Z%5D%7B0%2C3%7D%29&test_string=c01n02%0Ac01n02EOL&mode=finditer)

- String :`c01n02EOL`
- Pattern: `^c(\d\d)n(\d\d[A-Z]{0,3})$`
- Match groups: `("01", "02EOL")`
- `match.groups()[1].lstrip("0")` produces `"02EOL".lstrip("0")` -> `"2EOL"`

armi/bookkeeping/db/database.py
```diff
@@ -346,7 +346,12 @@ def splitDatabase(self, keepTimeSteps: Sequence[Tuple[int, int]], label: str) ->
             for groupName, _ in dbIn.items():
                 m = self.timeNodeGroupPattern.match(groupName)
                 if m:
-                    timeSteps.add((int(m.group(1)), int(m.group(2))))
+                    try:
+                        timeStep = (int(m.group(1)), int(m.group(2)))
+                    except ValueError:
+                        # handle the EOL time steps like c0102EOL
+                        timeStep = (int(m.group(1)), m.group(2))
+                    timeSteps.add(timeStep)
```
drewj-tp, line 355: Hmm are we able to reuse `self.genTimeSteps` here?

armi/bookkeeping/db/database.py
```diff
@@ -120,7 +120,7 @@ class Database:
     """
 
     # Allows matching for, e.g., c01n02EOL
-    timeNodeGroupPattern = re.compile(r"^c(\d\d)n(\d\d).*$")
+    timeNodeGroupPattern = re.compile(r"^c(\d\d)n(\d\d[A-Z]{0,3})$")
```
drewj-tp: We also write an error state to database at `cXnYerror`

https://github.com/terrapower/armi/blob/3032f9047c316ee12c823a1ec8990aaece487767/armi/bookkeeping/db/databaseInterface.py#L164-L171

I can see a place where someone wants to load the error state of the reactor for post-mortems.

drewj-tp (changes requested):
I have a question about the usability of this approach.

Say I want to load the EOL state. It lives at `cXnYEOL` in the database. When I do `db.genTimeSteps` to find the EOL point, I get the tuple `(X, "EOL")` but I have no information about the time node, just the cycle.

We need three values to find a unique statepoint in the reactor
1. cycle number
2. node number
3. statepoint name (empty string, `"EOL"`, or even `"error"`)

I think we need to consider returning this full triplet from `genTimeSteps`. If we return a tuple `(cycle, node, statepoint)`, that will be a breaking change for people (like me) who have written stuff like
```python
for cycle, node in db.genTimeSteps():
    # do thing
```
because we can't unpack a three-ple into a two-ple. _maybe_ we can do some clever dataclassing s.th. the `__iter__` method only returns cycle/node information with a deprecation notice? e.g.,
```python
@dataclass(frozen=True, eq=True, slots=True)  # note: slots added py3.10
class DbTimeNode:
    cycle: int
    node: int
    statepointName: str

    def __iter__(self):
        # tbd deprecation notice?
        return iter((self.cycle, self.node))
```

Eh?

armi/bookkeeping/db/database.py
john-science: `cXnYerror` OH RIGHT.

Man, I totally forgot about that.

Okay, I will amend my regex. Thanks!

Good catch!

armi/bookkeeping/db/database.py
john-science: Fixed

armi/bookkeeping/db/database.py
john-science: Naw, I tried. It doesn't save us any lines of code, because we will need to compare every group name in the file against the `timeNodeGroupPattern`.

armi/bookkeeping/db/database.py
john-science: Fixed

armi/bookkeeping/db/tests/test_database.py
```diff
@@ -986,6 +986,53 @@ def test_open(self):
...
+        self.assertListEqual(steps, [(0, 0), (0, 1), (1, 0), (1, 1)])
+
+        # test a case with an EOL time step
+        keys = ["c00n00", "c00n01", "c01n00", "c01n01", "c01n01EOL"]
+        db.h5db = {k: "fake" for k in keys}
+
+        steps = list(db.genTimeSteps())
+        self.assertListEqual(steps, [(0, 0), (0, 1), (1, 0), (1, 1), (1, 1, "EOL")])
+
+        # test a case with an EOL time step
```
drewj-tp:
```suggestion
        # test a case with an error state
```

armi/bookkeeping/db/database.py
```diff
@@ -119,8 +119,8 @@ class Database:
     `doc/user/outputs/database` for more details.
     """
 
-    # Allows matching for, e.g., c01n02EOL
-    timeNodeGroupPattern = re.compile(r"^c(\d\d)n(\d\d).*$")
+    # DB timenode matching. Please see genTimeSteps for details.
+    timeNodeGroupPattern = re.compile(r"^c(\d\d)n(\d\d)([a-zA-Z]{0,9})$")
```
drewj-tp, line 123: Do we care about future us / some plugin adding `cXnYNewExcitingStatePoint`?

```suggestion
    timeNodeGroupPattern = re.compile(r"^c(\d\d)n(\d\d)(.*)$")
```

Non-blocking, resolve w/ or w/o a change

armi/bookkeeping/db/database.py
```diff
@@ -302,7 +302,7 @@ def close(self, completedSuccessfully=False):
             newPath = safeMove(self._fullPath, self._fileName)
             self._fullPath = os.path.abspath(newPath)
 
-    def splitDatabase(self, keepTimeSteps: Sequence[Tuple[int, int]], label: str) -> str:
+    def splitDatabase(self, keepTimeSteps: Sequence[Tuple[int, Any]], label: str) -> str:
```
drewj-tp: It'll either be `Tuple[int, int]` or `Tuple[int, int, str]` or just
```suggestion
    def splitDatabase(self, keepTimeSteps: Sequence[Tuple], label: str) -> str:
```

armi/bookkeeping/db/database.py
```diff
@@ -563,15 +569,32 @@ def getLayout(self, cycle, node):
 
         return Layout(version, self.h5db[timeGroupName])
 
-    def genTimeSteps(self) -> Generator[Tuple[int, int], None, None]:
-        """Returns a generator of (cycle, node) tuples that are present in the DB."""
+    def genTimeSteps(self) -> Generator[Tuple[int, Any], None, None]:
```
drewj-tp: We can be verbose and make this `Generator[Union[Tuple[int, int], Tuple[int, int, str]], None, None]` or `Generator[Tuple, None, None]` or just let the type checkers resolve from the yield type. We're approaching the point of losing the utility of the type checker imo

```suggestion
    def genTimeSteps(self):
```

armi/bookkeeping/db/database.py
```diff
@@ -563,15 +569,32 @@ def getLayout(self, cycle, node):
...
+        """Returns a generator of (cycle, node) tuples that are present in the DB.
```
drewj-tp:
```suggestion
        """Returns a generator of tuples reflecting statepoints that are present in the DB.
```

drewj-tp (changes requested):
Thanks for the response. Soft requests on nothing really technical as type annotations and docs won't impact the functionality of the code.

armi/bookkeeping/db/database.py
john-science: I see what you mean. 

I don't want to ALLOW the statepoint name to have numbers in them though. I think that way lies madness. For instance, I don't to allow people to try things like this:  `c00n001thing`. That's unclear.

And, likewise, I think allowing special characters in the statepoint name would get ugly.

I don't know. It's all so hypothetical...

drewj-tp (approved):

PR #2610, opened 2026-07-28
Supporting Blueprint.dump without streams

john-science:

In a [recent PR](https://github.com/terrapower/armi/pull/2604) I attempted to fix a [bug](https://github.com/terrapower/armi/issues/2600) with `Blueprints.dump()`.

I convinced @drewj-tp  that no one was using the returned string from this method. And so I had a quicker fix.

But @yardasol found a downstream use-case where someone WAS using the retuned string. So I need another version.

Essentially, this is a rewrite of the [same method in yamlizable](https://github.com/SimplyKnownAsG/yamlize/blob/master/yamlize/yamlizable.py#L58), but with a "stream cleaner" added to fix this zeros. I tested downstream where Olek found the problem, and that passes now.

## SCR Information

 Change Type: fixes

One-Sentence Rationale: Blueprints.dump needs to be able to return a string sometimes.

---

## Checklist

### DIFF
--- a/armi/reactor/blueprints/__init__.py
+++ b/armi/reactor/blueprints/__init__.py
@@ -75,7 +75,7 @@ import h5py
 import ordered_set
 import yamlize
 import yamlize.objects
-from ruamel.yaml import RoundTripLoader
+from ruamel.yaml import RoundTripDumper, RoundTripLoader
 
 from armi import (
     context,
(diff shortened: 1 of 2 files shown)


yardasol (approved):
Great work on this @john-science. It's pretty small and seems reasonable. I don't love that we are copying code from `yamlize`, but doesn't seem like there's a way around it. Do you think we've sufficiently attributed `yamlize` in the function docstring? If so, feel free to merge.

armi/reactor/blueprints/__init__.py
```diff
@@ -512,19 +512,50 @@ def migrate(cls, inp: typing.TextIO):
...
+                dumper._emitter.dispose()
+            except AttributeError:
+                raise
+                dumper.dispose()  # cyaml
+
+        try:
+            Blueprints.streamCleaner(stream)
+        except io.UnsupportedOperation:
+            # Not all streams are writable.
+            pass
```
drewj-tp, line 541: This is interesting.

1. we've written to a stream above, right? In what case do we pass a buffer that works with `dump` but we can't write to it? I know some streams are not [seekable](https://docs.python.org/3/library/io.html#io.IOBase.seekable) and we need to do that for the cleanup. Is that what you mean?
2. If we can't do the cleanup, should we do something? Alert the user somehow?

armi/reactor/blueprints/__init__.py
```diff
@@ -512,19 +512,50 @@ def migrate(cls, inp: typing.TextIO):
...
+            dumper._serializer.open()
+            root_node = cls.to_yaml(dumper, data)
+            dumper.serialize(root_node)
+            dumper._serializer.close()
+        finally:
+            try:
+                dumper._emitter.dispose()
+            except AttributeError:
+                raise
+                dumper.dispose()  # cyaml
```
drewj-tp, line 535: I don't _love_ the reliance on non-public methods in a library that we already have frustrations for breaking changes...

What about a pattern like
```python
def dump(cls, data, stream=None):
    output = super().dump(data, stream)
    # no output buffer, user wants the rendered output string
    if stream is None:
        buf = io.StringIO(output)
        cls.streamCleaner(buf)
        return buf.getvalue()
    # output buffer given
    cls.streamCleaner(stream)
```
But then if we get a non-seekable stream, that complicates things. Probably want to dump to a `io.StringIO`, clean that buffer, and then write the cleaned contents to the user requested stream...

😵

armi/reactor/blueprints/__init__.py
john-science: This is necessary because there are places downstream where people pass in a write-only file stream.

And, in that case, we can't read the stream to return it. It's just not possible.

And that is a limitation of the current `yamlize.Object.dump()` method too. There's nothing for it.

armi/reactor/blueprints/__init__.py
john-science: > I don't love the reliance on non-public methods...

Neither do I. But we are trying to hot patch a problem in other libraries.

Functionality over asthetics.

john-science: If it helps, we ARE pinned to a version of `yamlize`, which is no longer under development. So these private methods won't change suddenly on us.

opotowsky (approved):

PR #2612, opened 2026-07-31
Ensuring migrations are not applied when they are not needed

### LINKED ISSUE #2431: Cannot reload from database that includes Nitrogen-14
If you take a blueprints file that includes `N14` (in a material and also in the `nuclide flags`) but does NOT include elemental Nitrogen, run that input, and then try to open the resulting database, ARMI will fail with the following error:

```
ValueError                                Traceback (most recent call last)
Cell In[4], line 2
      1 with db:
----> 2     r = db.load(0,0)

File ~\codes\nala\framework\armi\bookkeeping\db\database.py:753, in Database.load(self, cycle, node, cs, bp, statePointName, allowMissing, handleInvalids, callReactorConstructionHook)
    750     self._assignBlueprintsParams(bp, groupedComps)
    752 # stitch together
--> 753 self._compose(iter(comps), cs)
    755 # also, make sure to update the global serial number so we don't reuse a number
    756 parameterCollections.GLOBAL_SERIAL_NUM = max(parameterCollections.GLOBAL_SERIAL_NUM, layout.serialNum.max())

File ~\codes\nala\framework\armi\bookkeeping\db\database.py:859, in Database._compose(self, comps, cs, parent)
    856 children = []
    858 for _ in range(numChildren):
--> 859     child = self._compose(comps, cs, parent=comp)
    860     children.append(child)
    861     if isinstance(child, Component):

File ~\codes\nala\framework\armi\bookkeeping\db\database.py:871, in Database._compose(self, comps, cs, parent)
    868     comp.add(child)
    870 if isinstance(comp, Core):
--> 871     comp.processLoading(cs, dbLoad=True)
    872 elif isinstance(comp, Assembly):
    873     comp.calculateZCoords()

File ~\codes\nala\framework\armi\reactor\cores.py:2155, in Core.processLoading(self, cs, dbLoad)
   2147     runLog.warning(
   2148         "The system has no heavy metal and therefore is not a nuclear reactor.\n"
   2149         "Please make sure that this is intended and not a input error."
   2150     )
   2152 if dbLoad:
   2153     # reactor.blueprints.assemblies need to be populated this normally happens during
   2154     # blueprint constructAssem. But for DB load, this is not called so it must be here.
-> 2155     self.parent.blueprints._prepConstruction(cs)
   2156 else:
   2157     # set reactor level meshing params
   2158     nonUniformAssems = [Flags.fromStringIgnoreErrors(t) for t in cs[CONF_NON_UNIFORM_ASSEM_FLAGS]]

File ~\codes\nala\framework\armi\reactor\blueprints\__init__.py:302, in Blueprints._prepConstruction(self, cs)
    299 self.assemblies.clear()
    301 for aDesign in self.assemDesigns:
--> 302     a = aDesign.construct(cs, self)
    303     self._assembliesBySpecifier[aDesign.specifier] = a
    304     self.assemblies[aDesign.name] = a

File ~\codes\nala\framework\armi\reactor\blueprints\assemblyBlueprint.py:180, in AssemblyBlueprint.construct(self, cs, blueprint)
    178 runLog.info("Constructing assembly `{}`".format(self.name))
    179 self._checkParamConsistency()
--> 180 a = self._constructAssembly(cs, blueprint)
    181 a.calculateZCoords()
    182 return a

File ~\codes\nala\framework\armi\reactor\blueprints\assemblyBlueprint.py:188, in AssemblyBlueprint._constructAssembly(self, cs, blueprint)
    186 blocks = []
    187 for axialIndex, bDesign in enumerate(self.blocks):
--> 188     b = self._createBlock(cs, blueprint, bDesign, axialIndex)
    189     blocks.append(b)
    191 assemblyClass = self.getAssemClass(blocks)

File ~\codes\nala\framework\armi\reactor\blueprints\assemblyBlueprint.py:259, in AssemblyBlueprint._createBlock(self, cs, blueprint, bDesign, axialIndex)
    249 for key, mod in {
    250     "byBlock": {**self.materialModifications},
    251     **self.materialModifications.byComponent,
    252 }.items():
    253     materialInput[key] = {
    254         modName: modList[axialIndex]
    255         for modName, modList in mod.items()
    256         if self._shouldMaterialModiferBeApplied(modList[axialIndex])
    257     }
--> 259 b = bDesign.construct(cs, blueprint, axialIndex, meshPoints, height, xsType, materialInput)
    261 b.completeInitialLoading()
    263 # set b10 volume cc since its a cold dim param

File ~\codes\nala\framework\armi\reactor\blueprints\blockBlueprint.py:143, in BlockBlueprint.construct(self, cs, blueprint, axialIndex, axialMeshPoints, height, xsType, materialInput)
    141 for componentDesign in self:
    142     filteredMaterialInput, byComponentMatModKeys = self._filterMaterialInput(materialInput, componentDesign)
--> 143     c = componentDesign.construct(
    144         blueprint,
    145         filteredMaterialInput,
    146         cs[CONF_INPUT_HEIGHTS_HOT],
    147     )
    148     components[c.name] = c
    150     # check that the mat mods for this component are valid options
    151     # this will only examine by-component mods, block mods are done later

File ~\codes\nala\framework\armi\reactor\blueprints\componentBlueprint.py:201, in ComponentBlueprint.construct(self, blueprint, matMods, inputHeightsConsideredHot)
    175 """Construct a component or group.
    176
    177 .. impl:: User-defined on material alterations are applied here.
   (...)    198     See the case setting of the same name.
    199 """
    200 runLog.debug(f"Constructing component {self.name}")
--> 201 kwargs = self._conformKwargs(blueprint, matMods)
    202 shape = self.shape.lower().strip()
    203 if shape == COMPONENT_GROUP_SHAPE:

File ~\codes\nala\framework\armi\reactor\blueprints\componentBlueprint.py:316, in ComponentBlueprint._conformKwargs(self, blueprint, matMods)
    313     continue
    314 elif attr.name == "material":
    315     # value is a material instance
--> 316     value = self._constructMaterial(blueprint, matMods)
    317 elif attr.name == "latticeIDs":
    318     # Don't pass latticeIDs on to the component constructor.
    319     # They're applied during block construction.
    320     continue

File ~\codes\nala\framework\armi\reactor\blueprints\componentBlueprint.py:372, in ComponentBlueprint._constructMaterial(self, blueprint, matMods)
    369 missing = set(mat.massFrac.keys()).difference(nucsInProblem)
    371 if missing:
--> 372     raise ValueError(
    373         f"The nuclides {missing} are present in material {mat} by compositions, but are not specified in the "
    374         "`nuclide flags` section of the input file. They need to be added, or custom isotopics need to be "
    375         "applied."
    376     )
    378 return mat

ValueError: The nuclides {'N'} are present in material <Material: Custom> by compositions, but are not specified in the `nuclide flags` section of the input file. They need to be added, or
 custom isotopics need to be applied.
```

This error message appears odd, because the original blueprints do NOT include elemental Nitrogen.

If you extract the blueprints stored on the database, you'll notice that all instances of `N14` in your model have been replaced by `N`, but the nuclide flags have not been also updated. I think this update takes place here: https://github.com/terrapower/armi/blob/e3b7b92160bdfde5914bff90e475ca244c4fab36/armi/migration/m0_1_3.py#L53

If you add elemental Nitrogen to the nuclide flags, the database will be able to load.

This bug makes it difficult to utilize databases from runs when the user didn't include elemental `N` in the blueprints.

john-science:

The original concern was `N14` was disappearing. The problem was that a database migration was wiping out N14 as an input nuclide. This should not have been happening though, because the new database did not need the ancient DB migration applied to it.

Also, per reviewer comments, I have altered the base `Migration` class to use some abstract methods now.

close #2431

## SCR Information

Change Type: fixes

One-Sentence Rationale: Applying redundant migrations can incorrectly modify databases.

---

## Checklist

### DIFF
--- a/armi/migration/base.py
+++ b/armi/migration/base.py
@@ -20,6 +20,7 @@ migration class defined here chooses this behavior based on whether the ``stream
 the constructor.
 """
 
+import io
 import os
 import shutil
 
@@ -58,19 +59,28 @@ class Migration:
         version : str
             Optional DB version string, for example: 1.2.3
         """
-        # TODO: Compare self.toVersion to version of DB or whatever...
+        skip = False
+        # Compare self.toVersion to the version of the DB.
         if version is not None:
-            if versionToNumber(self.toVersion) >= versionToNumber(version):
+            if versionToNumber(version) >= versionToNumber(self.toVersion):
                 # this migration is not necessary, because the DB is newer than that.
-                return
+                skip = True
 
-        runLog.info(f"Applying {self}")
+        if not skip:
+            runLog.info(f"Applying {self}")
+
         if self.path:
             self._loadStreamFromPath()
-        newStream = self._applyToStream()
-        if self.path:
-            self._backupOriginal()
-            self._writeNewFile(newStream)
+
+        if not skip:
+            newStream = self._applyToStream()
+
+            if self.path:
+                self._backupOriginal()
+                self._writeNewFile(newStream)
+        else:
+            newStream = io.StringIO(self.stream.read())
+
         return newStream
 
     def _loadStreamFromPath(self):
--- a/armi/migration/base.py
+++ b/armi/migration/base.py
@@ -14,12 +14,10 @@
 """
 Base migration classes.
 
-A classic migration takes a file name, read the files, migrates the
-data, and re-writes the file. Some migrations need to happen live
-on a stream. For example, if an old/invalid input file is being read
-in from an old database. The migration class defined here
-chooses this behavior based on whether the ``stream`` or ``path``
-variables are given in the constructor.
+A classic migration takes a file name, read the files, migrates the data, and re-writes the file. Some migrations need
+to happen live on a stream. For example, if an old/invalid input file is being read in from an old database. The
+migration class defined here chooses this behavior based on whether the ``stream`` or ``path`` variables are given in
+the constructor.
 """
 
 import os
@@ -27,13 +25,13 @@ import shutil
 
 from armi import runLog
 from armi.settings import caseSettings
+from armi.settings.settingsValidation import versionToNumber
 
 
 class Migration:
     """Generic migration.
 
-    To implement a concrete Migration, one must often only
-    implement the ``_applyToStream`` method.
+    To implement a concrete Migration, one must often only implement the ``_applyToStream`` method.
     """
 
     fromVersion = "x.x.x"
@@ -49,12 +47,23 @@ class Migration:
     def __repr__(self):
         return f"<Migration from {self.fromVersion}: {self.__doc__[:40]}..."
 
-    def apply(self):
+    def apply(self, version: str = None):
         """
         Apply migration.
 
         This is generally called from a subclass.
+
+        Parameters
+        ----------
+        version : str
+            Optional DB version string, for example: 1.2.3
         """
+        # TODO: Compare self.toVersion to version of DB or whatever...
+        if version is not None:
+            if versionToNumber(self.toVersion) >= versionToNumber(version):
+                # this migration is not necessary, because the DB is newer than that.
+                return
+
         runLog.info(f"Applying {self}")
         if self.path:
             self._loadStreamFromPath()
--- a/armi/migration/base.py
+++ b/armi/migration/base.py
@@ -61,7 +61,7 @@ class Migration:
         """
         skip = False
         # Compare self.toVersion to the version of the DB.
-        if version is not None:
+        if version is not None or version != "uncontrolled":
             if versionToNumber(version) >= versionToNumber(self.toVersion):
                 # this migration is not necessary, because the DB is newer than that.
                 skip = True
--- a/armi/migration/base.py
+++ b/armi/migration/base.py
@@ -68,13 +68,11 @@ class Migration:
                 # this migration is not necessary, because the DB is newer than that.
                 skip = True
 
-        if not skip:
-            runLog.info(f"Applying {self}")
-
         if self.path:
             self._loadStreamFromPath()
 
         if not skip:
+            runLog.info(f"Applying {self}")
             newStream = self._applyToStream()
 
             if self.path:
--- a/armi/migration/base.py
+++ b/armi/migration/base.py
@@ -61,7 +61,7 @@ class Migration:
         """
         skip = False
         # Compare self.toVersion to the version of the DB.
-        if version is not None or version != "uncontrolled":
+        if version is not None and version != "uncontrolled":
             if versionToNumber(version) >= versionToNumber(self.toVersion):
                 # this migration is not necessary, because the DB is newer than that.
                 skip = True
--- a/armi/migration/base.py
+++ b/armi/migration/base.py
@@ -32,15 +32,17 @@ from armi.settings.settingsValidation import versionToNumber
 class Migration:
     """Generic migration.
 
-    To implement a concrete Migration, one must often only implement the ``_applyToStream`` method.
+    To implement a concrete Migration, you must implement the ``_applyToStream`` method. You must also over-write the
+    class variables ``fromVersion`` and ``toVersion``. These are the ARMI versions you are migrating from and to. The
+    ``toVersion`` must be higher than the ``fromVersion``.
     """
 
+    # Any subclass MUST over-write these ARMI versions; "toVersion" MUST be higher than "fromVersion".
     fromVersion = "x.x.x"
     toVersion = "x.x.x"
 
     def __init__(self, stream=None, path=None):
-        if not (bool(stream) ^ bool(path)):
-            # XOR
+        if not (bool(stream) ^ bool(path)):  # XOR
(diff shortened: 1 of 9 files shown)


armi/migration/base.py
```diff
@@ -49,19 +48,39 @@ def __init__(self, stream=None, path=None):
...
+
+        if not skip:
+            newStream = self._applyToStream()
+
+            if self.path:
+                self._backupOriginal()
+                self._writeNewFile(newStream)
+        else:
+            newStream = io.StringIO(self.stream.read())
+
```
keckler, line 100: The logic here looks correct, but it looks like there may be some opportunity to consolidate the various conditionals, if you care to. Not blocking.

armi/settings/settingsValidation.py
```diff
@@ -665,4 +665,25 @@ def validateVersion(versionThis: str, versionRequired: str) -> bool:
...
+
+    Parameters
+    ----------
+    versionThis: str
+        ARMI-style version string, typically of the form 1.2.3.
+    """
+    versionBits = version.split(".")
+    num = 1e6 * int(versionBits[0]) + 1e3 * int(versionBits[1])
+    if len(versionBits) >= 3:
+        num += int(versionBits[2])
```
keckler, line 687: Not sure if this check is worth it. I don't know who/what might be passing in the version numbers, so I don't know what kind of opportunities there might be for passing in malformed values.
```suggestion
    if len(versionBits) == 3:
        num += int(versionBits[2])
    elif len(versionBits) > 3:
        raise some kind of error
```

keckler: I guess a user could have put any type of string in the `version` field of their blueprints, so maybe some amount of validation is appropriate.

armi/migration/tests/test_m0_1_6.py
```diff
@@ -38,3 +38,20 @@ def test_locationLabelMigration(self):
         reader = SettingsReader(newCs)
         reader.readFromStream(converter.apply())
         self.assertEqual(newCs["detailAssemLocationsBOL"][0], "011-012")
+
+    def test_noMigrationIfVersion(self):
+        """Ensure that no migration happens if the version of the DB is new than the migration 'toVersion'."""
```
keckler:
```suggestion
        """
        Ensure that no migration happens if the version of the DB is newer than the migration 'toVersion'.
        
        This test checks the specific case of `ConvertAlphanumLocationSettingsToNum`, which should only
        be applied until v0.1.7.
        """
```

armi/migration/base.py
keckler: Slightly tangential to this PR, but we may want to consider making these attributes abstract since we are now more explicitly relying on them to exist and have real values.

armi/migration/base.py
```diff
@@ -14,26 +14,25 @@
...
 from armi.settings import caseSettings
+from armi.settings.settingsValidation import versionToNumber
 
 
 class Migration:
     """Generic migration.
 
-    To implement a concrete Migration, one must often only
-    implement the ``_applyToStream`` method.
+    To implement a concrete Migration, one must often only implement the ``_applyToStream`` method.
```
keckler: This should be updated to note that the `toVersion` attribute also needs to be updated when creating a new `Migration` object.

armi/migration/base.py
keckler: Can we consider adding a docstring here, to explain how these `fromVersion` and `toVersion` attributes will be used? Right now I would not say that it is obvious.

keckler (approved):
A few suggestions. Overall looks like it is doing what its supposed to, so thanks for the fix.

armi/migration/base.py
john-science: hmmm

It seems apparent to me that the "from" and "to" versions are mandatory for a migration to work.

I will make sure that is clearly documented though. Sure.

armi/migration/base.py
john-science: huh

Okay, I can do that.

armi/settings/settingsValidation.py
john-science: Sorry, I don't really WANT to raise an error here.

There is a very rarely used `0.7.0.dev1` version that ARMI only ever uses for Alex, to build him a little test wheel.

But those little dev versions are not real releases, and so don't have their own version of the ARMI Database to compare against. Thus, I don't want this to fail for those rare situations. I want the code to just ignore them.

john-science: I added a comment to this affect.

armi/migration/base.py
john-science: Done

armi/migration/base.py
john-science: Done

armi/migration/base.py
john-science: Done

keckler (approved):

armi/migration/base.py
john-science: I am adding this into the PR. Because it's so easy to do.

john-science: Done

PR #2613, opened 2026-07-31
Redistribute pinNDens and detailedNDens through axial expansion

### LINKED ISSUE #2609: Non-conservation of detailed and pin number densities during axial expansion
I've been tracking this down over the last week and I believe that we do not conserve detailed nuclide number density `component.p.detailedNDens` during axial expansion. This can lead to a loss of nuclides tracked through these "detailed" nuclide vectors. 

When we perform axial expansion, we have a lot of accounting to make sure we retain mass of individual nuclides that are retrievable through `Component.getNumberDensities`. See the `RedistributeMass` helper class

https://github.com/terrapower/armi/blob/3032f9047c316ee12c823a1ec8990aaece487767/armi/reactor/converters/axialExpansionChanger/redistributeMass.py#L30-L31

and the extensive testing in 

https://github.com/terrapower/armi/blob/3032f9047c316ee12c823a1ec8990aaece487767/armi/reactor/converters/tests/test_axialExpansionChanger_MultiPin.py#L426-L446

But the logic for updating the detailed number density filters down, eventually, to multiplying by a scalar based on strictly temperature dependent expansion

https://github.com/terrapower/armi/blob/3032f9047c316ee12c823a1ec8990aaece487767/armi/reactor/components/component.py#L800-L812

https://github.com/terrapower/armi/blob/3032f9047c316ee12c823a1ec8990aaece487767/armi/reactor/components/component.py#L607-L609

## Minimum working example

<details>

<summary>MWE</summary>

```python
import pandas as pd
from armi.testing import loadTestReactor
from armi.reactor.converters.axialExpansionChanger import AxialExpansionChanger

def initDetailedNDens(r: Reactor, /, *, flags):
    foundNucs: set[str] = set()
    comps = r.getComponents(flags)
    for c in comps:
        foundNucs.update(c.getNuclides())
    detailNuclides = sorted(foundNucs)
    for c in comps:
        c.p.detailedNDens = c.getNuclideNumberDensities(detailNuclides)
    return detailNuclides

def getDetailedAtoms(r: Reactor, nuclides: list[str], /):
    index: list[tuple[str, str]] = []
    values: list[float] = []
    for a in r.core:
        for b in a:
            lead = b.getLocation()
            for c in filter(lambda c: c.p.detailedNDens is not None, b):
                cid = f"{lead}-{c.getName()}"
                vol = c.getVolume()
                for nuc, dnd in zip(nuclides, c.p.detailedNDens):
                    if dnd:
                        values.append(dnd * vol)
                        index.append((nuc, cid))
    return pd.Series(
        values,
        index=pd.MultiIndex.from_tuples(index, names=["nuclide", "cid"]),
        name="detailed_atoms",
    )

_, r = loadTestReactor()
detailedNucKeys = initDetailedNDens(r, flags=[Flags.FUEL, Flags.CONTROL, Flags.DUCT])
initialDND = getDetailedAtoms(r, detailedNucKeys)

# Not trying to have physically correct expansion values
# esp. because the test reactor has B4C > 600C already out of the limits
armi.materials.material.FAIL_ON_RANGE = False

expander = AxialExpansionChanger(detailedAxialExpansion=False)
for a in r.core:
    expander.setAssembly(a)
    for c in a.iterComponents():
        expander.expansionData.updateComponentTemp(c, c.temperatureInC * 0.9)
    expander.expansionData.computeThermalExpansionFactors()
    expander.axiallyExpandAssembly(recalculateBurnup=False)

postExpansion = getDetailedAtoms(r, detailedNucKeys)
compare = pd.DataFrame({"initial": initialDND, "post": postExpansion})
totalAtomsByNuclide = compare.groupby("nuclide").sum()
totalAtomsByNuclide["diff"] = totalAtomsByNuclide.eval("post - initial")
totalAtomsByNuclide["rdiff"] = totalAtomsByNuclide.eval("diff / initial")
```

</details>

If you scrape the pre- and post-thermal expansion detailed number of atoms (`c.p.detailedNDens * c.getVolume()`) and compute per-nuclide differences

| nuclide   |     initial |        post |        diff |       rdiff |
|:----------|------------:|------------:|------------:|------------:|
| FE57      |  305.909    |  306.108    | 0.198827    | 0.000649955 |
| W183      |    4.07012  |    4.07277  | 0.0026454   | 0.000649955 |
| NI60      |   23.3343   |   23.3495   | 0.0151663   | 0.000649955 |
| V50       |    0.139807 |    0.139898 | 9.08682e-05 | 0.000649955 |
| MO95      |   15.7587   |   15.769    | 0.0102425   | 0.000649955 |
| W184      |    8.71478  |    8.72044  | 0.00566422  | 0.000649955 |
| CR52      | 1798.01     | 1799.18     | 1.16863     | 0.000649955 |
| SI30      |    2.60937  |    2.61106  | 0.00169597  | 0.000649955 |
| FE54      |  843.812    |  844.36     | 0.54844     | 0.000649955 |
| NI61      |    1.01442  |    1.01508  | 0.000659327 | 0.000649955 |

It's very curious that, for this example, the relative difference seems to be 0.065% almost across the board. Some cases are zero or very near zero like U235. Likely b/c these are likely to reside in the "target" components where we don't need to do this mass transfer asthe axial block boundary lines are defined by these components.

## Further proof

I have updated the axial expansion tests in this branch `drewj/ax-exp-detailed-ndens` to show that component nuclide mass is conserved while detailed nuclide tracking is not conserved

### LINKED ISSUE #2621: Unknown directive "clode-block" building docs
https://github.com/terrapower/armi/actions/runs/31024337991/job/92368796729#step:7:1845

https://github.com/terrapower/armi/blob/321d9466dc476cbaac97fe97e1c1fa3455223e10/doc/user/grids.rst?plain=1#L241

drewj-tp:
### Primary
The `RedistributeMass` object now redistributes
- `Component.p.pinNDens`
- `Component.p.detailedNDens`

through the axial expansion/contraction of two linked components. This change is akin to a volume-weighted average of the densities such that `n_a' = (n_a * v_a + n_b * v_b) / (v_a + v_b)` where:

- `v_a` and `n_a` are the pre-distribution volume and number density from the "to" component
- `n_b` are the pre-distribution volume and number density of the "from" component
- `v_b` is the volume in the "from" component that is being moved to the from component (area * deltaz)
- `n_a'` is the post-distribution number density in the "to" component

Testing is added that ensures conservation of total number of atoms (pin and detailed, separately) before and after the expansion. Tests are added to show expected changes in the "from" component. By which we can infer the "to" component is correctly updated.

Testing is also added on the `RedistributeMass` object directly, on some of the methods. Including the pin and detailed ndens migration.

### Quality-of-Life

Refactored `RedistributeMass.performRedistribution` to make life easier for subclasses. They can override `_performRedistribution` on the assumption the compatability check is complete. Otherwise, you may have patterns like:

```python
def performRedistribution(self):
    super().performRedistribution()
    if self.compatibilityCheck():
        self._doMySpecialRedistribution)
````

Now can be

```python
def _performRedistribution(self):
    super()._performRedistribution()
    self._doMySpecialRedistribution()
```

Cuts down on the warning messages in half b/c there's only one `compatibilityCheck` per invocation of `performRedistribution`

Added tests on "bulk conservation" in the redistribution showing that, before and after, expansion and redistibution:

1. mass of components
2. heavy metal mass
3. heavy metal moles
4. BOL heavy metal moles
5. BOL heavy metal mass

is conserved. These are the most simple checks and assurances we want to have. Now they're explicitly tested.

### TODO

- [x] Downstream testing
- [x] Documentation on handling /structure of `pinNDens` and `detailedNDens`

Closes #2609 

## SCR Information

Change Type: fixes

One-Sentence Rationale: Redistribute pinNDens and detailedNDens through axial expansion

One-line Impact on Requirements: R_ARMI_AXIAL_EXP_THERM, R_ARMI_AXIAL_EXP_PRESC, and R_ARMI_INP_COLD_HEIGHT all involve axial expansion and the redistribution of nuclide data between adjacent components.

---

## Checklist


armi/reactor/converters/axialExpansionChanger/redistributeMass.py
```diff
@@ -233,3 +240,66 @@ def _adjustMassParams(self):
...
+            return False
+        toData: typing.Optional[np.ndarray] = getattr(self.toComp.p, paramName, None)
+        if toData is None:
+            return False
+        # array * scalar gives us a new array, not a copy. So we can do in-place mutations later
+        try:
+            toData = toData * self.toCompVolume
+            fromData = fromData * self.fromCompVolume
+            toData += fromData
+            toData /= self.newVolume
```
albeanth, line 292: Do you have any objections to making this one line? Imo it's easier to read that way
```python
newNDens = (toData * self.toCompVolume + fromData * self.fromCompVolume) / self.newVolume
```

john-science:
> Refactored `RedistributeMass.performRedistribution` to make life easier for subclasses.

TIL: People are subclassing this.

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
```diff
@@ -233,3 +240,66 @@ def _adjustMassParams(self):
...
+        fromData: typing.Optional[np.ndarray] = getattr(self.fromComp.p, paramName, None)
+        if fromData is None:
+            return False
+        toData: typing.Optional[np.ndarray] = getattr(self.toComp.p, paramName, None)
+        if toData is None:
+            return False
+        # array * scalar gives us a new array, not a copy. So we can do in-place mutations later
+        try:
+            toData = toData * self.toCompVolume
+            fromData = fromData * self.fromCompVolume
```
albeanth, line 290: the properties, `self.toCompVolume` and `self.fromCompVolume`, are using the post-temperature update and may therefore have the wrong areas... shown in https://github.com/terrapower/armi/pull/2613/changes#r3707364778

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
```diff
@@ -233,3 +240,66 @@ def _adjustMassParams(self):
...
+            toData += fromData
+            toData /= self.newVolume
+        except Exception:
+            msg = (
+                f"Error updating {paramName} on {self.assemblyName} : toComp={self.toComp} : fromComp={self.fromComp}\n"
+                f"Shape fromData {toData.shape}\nShape toData {toData.shape}"
+            )
+            runLog.error(msg)
+            raise
+        setattr(self.toComp.p, paramName, toData)
```
john-science: Wouldn't this be more clear?

```python
self.toComp.p[paramName] = toData
```

armi/reactor/converters/tests/test_massRedistribution.py
```diff
@@ -0,0 +1,127 @@
...
+        self.toComp.parent = BlockLike(self.toComp.height)
+
+        # Height of fromComp to be shifted to toComp
+        self.dz = 0.6
+        self.distributor = RedistributeMass(
+            fromComp=self.fromComp, toComp=self.toComp, deltaZTop=self.dz, assemName=self._testMethodName, initOnly=True
+        )
+
+    def test_noUpdateMisMatchedPinNDens(self):
+        """If pin number density is only updated if both have similarly shaped data."""
```
albeanth:
```suggestion
        """Pin number density is only updated if both components have `pinNDens` populated."""
```

armi/reactor/converters/tests/test_massRedistribution.py
```diff
@@ -0,0 +1,127 @@
...
+        # neither have pin ndens => no update
+        self.toComp.p.pinNDens = None
+        self.fromComp.p.pinNDens = None
+        self.assertFalse(self.distributor._adjustPinNDens())
+        # only one has pin ndens => update
```
albeanth:
```suggestion
        # generate random sample pNDens data
```

armi/reactor/converters/tests/test_massRedistribution.py
```diff
@@ -0,0 +1,127 @@
...
+        rng = np.random.default_rng()
+        pinDensShape = (7, 3)  # arbitrary
+        sampleData = rng.uniform(low=0, high=1e-2, size=pinDensShape).astype(np.float32)
+
```
albeanth:
```suggestion

    # only fromComp has pin ndens => no update
```

armi/reactor/converters/tests/test_massRedistribution.py
```diff
@@ -0,0 +1,127 @@
...
+        self.fromComp.p.pinNDens = sampleData
+        self.assertFalse(self.distributor._adjustPinNDens())
+
```
albeanth:
```suggestion

    # only toComp has pin ndens => no update
```

armi/reactor/converters/tests/test_axialExpansionChanger_MultiPin.py
```diff
@@ -16,10 +16,11 @@
 import copy
 import io
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, Tuple
```
john-science, line 19: Absolutely not your fault, but there is just too much architecture built up in this unit test file.

I argue that we generally don't want multiple layers of abstraction in unit testing. Unless it makes the tests run faster, it obscures the ideas of the tests from the readers. And it's important that people can at-a-glance understand unit tests.

#end-rant

armi/reactor/converters/tests/test_massRedistribution.py
```diff
@@ -0,0 +1,127 @@
...
+from unittest import TestCase
+
+import numpy as np
+
+from armi.reactor.components import Circle
+from armi.reactor.converters.axialExpansionChanger.redistributeMass import RedistributeMass
+
+
+class BlockLike:
```
john-science, line 29: Good name.

Imma steal this paradigm in the future.

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
```diff
@@ -233,3 +240,66 @@ def _adjustMassParams(self):
...
+        toData: typing.Optional[np.ndarray] = getattr(self.toComp.p, paramName, None)
+        if toData is None:
+            return False
+        # array * scalar gives us a new array, not a copy. So we can do in-place mutations later
+        try:
+            toData = toData * self.toCompVolume
+            fromData = fromData * self.fromCompVolume
+            toData += fromData
+            toData /= self.newVolume
+        except Exception:
```
albeanth, line 293: I also wonder if we should check to make sure that the `detailedPinNucKeys` are the same. Maybe this could be a snapshot corner case.... This may already be caught in a separate internal plugin.

armi/reactor/converters/tests/test_massRedistribution.py
```diff
@@ -0,0 +1,127 @@
...
+        initialFromPinDens = self.fromComp.p.pinNDens.copy()
+        initialToPinDens = self.toComp.p.pinNDens.copy()
+        toVol = self.toComp.height * self.toComp.getArea()
+        fromVol = self.dz * self.fromComp.getArea()
+        expected = (initialToPinDens * toVol + initialFromPinDens * fromVol) / (fromVol + toVol)
+
+        self.assertTrue(self.distributor._adjustPinNDens())
+        # no change in from comp
+        np.testing.assert_allclose(self.fromComp.p.pinNDens, initialFromPinDens)
+        np.testing.assert_allclose(self.toComp.p.pinNDens, expected, rtol=1e-5)
```
albeanth: when I run this locally, we can run with the default rtol of 1e-7 fwiw

armi/reactor/converters/tests/test_massRedistribution.py
```diff
@@ -0,0 +1,127 @@
...
+        initialFromDetailedDens = self.fromComp.p.detailedNDens.copy()
+        initialToDetailedDens = self.toComp.p.detailedNDens.copy()
+        toVol = self.toComp.height * self.toComp.getArea()
+        fromVol = self.dz * self.fromComp.getArea()
+        expected = (initialToDetailedDens * toVol + initialFromDetailedDens * fromVol) / (fromVol + toVol)
+
+        self.assertTrue(self.distributor._adjustDetailedNDens())
+        # no change in from comp
+        np.testing.assert_allclose(self.fromComp.p.detailedNDens, initialFromDetailedDens)
+        np.testing.assert_allclose(self.toComp.p.detailedNDens, expected, rtol=1e-5)
```
albeanth: when I run this locally, we can run with the default rtol of 1e-7 fwiw

armi/reactor/converters/tests/test_axialExpansionChanger_MultiPin.py
```diff
@@ -488,6 +531,36 @@ def _redistributeMassWithTempAssert(self, fromComp: Component, toComp: Component
...
+            # other changes in this test
+            assert_allclose(fromComp.p.pinNDens, fromCompRefData.pinNDens)
+        else:
+            # fromComp ndens should only change due to the thermal expansion in the lower component
+            assert_allclose(fromComp.p.pinNDens, fromCompRefData.pinNDens / growFrac)
+        # conservation of total number of pin atoms through expansion
+        assert_allclose(
+            fromComp.p.pinNDens * fromComp.getVolume() + toComp.p.pinNDens * toComp.getVolume(),
+            fromCompRefData.pinNDens * fromCompRefData.volume + toCompRefData.pinNDens * toCompRefData.volume,
+            rtol=1e-5,
```
albeanth:
```suggestion
```
to have this work, you need to move the temp update to be the last step in the redistribution

armi/reactor/converters/tests/test_axialExpansionChanger_MultiPin.py
```diff
@@ -488,6 +531,36 @@ def _redistributeMassWithTempAssert(self, fromComp: Component, toComp: Component
...
+            # density in the from component is unchanged. it has not undergone any expansion or
+            # other changes in this test
+            assert_allclose(fromComp.p.detailedNDens, fromCompRefData.detailedNDens)
+        else:
+            # fromComp ndens should only change due to the thermal expansion in the lower component
+            assert_allclose(fromComp.p.detailedNDens, fromCompRefData.detailedNDens / growFrac)
+        assert_allclose(
+            fromComp.p.detailedNDens * fromComp.getVolume() + toComp.p.detailedNDens * toComp.getVolume(),
+            fromCompRefData.detailedNDens * fromCompRefData.volume + toCompRefData.detailedNDens * toCompRefData.volume,
+            rtol=1e-5,
```
albeanth:
```suggestion
```
to have this work, you need to move the temp update to be the last step in the redistribution

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
```diff
@@ -56,12 +57,18 @@ def __init__(
...
+            self._performRedistribution()
+
+    def _performRedistribution(self):
+        """Perform the mass redistribution between compatible components."""
+        self.setNewToCompNDens()
+        self.setNewToCompTemperature()
+        if self.fromComp.p.molesHmBOL is not None and self.toComp.p.molesHmBOL is not None:
+            self._adjustMassParams()
+        self._adjustPinNDens()
+        self._adjustDetailedNDens()
```
albeanth, line 74:
```suggestion
        self.setNewToCompNDens()
        if self.fromComp.p.molesHmBOL is not None and self.toComp.p.molesHmBOL is not None:
            self._adjustMassParams()
        self._adjustPinNDens()
        self._adjustDetailedNDens()
        self.setNewToCompTemperature()
```
makes sure that the right areas are used for the redistribution

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
```diff
@@ -233,3 +240,66 @@ def _adjustMassParams(self):
...
+        -------
+        bool
+            If the update was performed.
+        """
+        fromData: typing.Optional[np.ndarray] = getattr(self.fromComp.p, paramName, None)
+        if fromData is None:
+            return False
+        toData: typing.Optional[np.ndarray] = getattr(self.toComp.p, paramName, None)
+        if toData is None:
+            return False
```
albeanth: I wonder what the likelihood is of unintentionally missed redistributions for `pinNDens` or `detailedNDens` that wouldn't get caught with the existing check in `compatabilityCheck`? If we have anything that falls into this category it could be _tough_ to track down.

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
john-science: That link doesn't take me anywhere. Maybe you meant to add a line number to it?

albeanth: Bummer. It should go down to these comments at the bottom of test_axialExpansionChanger_multiPin.py

Image

drewj-tp: will add

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
drewj-tp: I thought about this but it's not an ARMI parameter. I intend to write docs that state how we expect `pinNDens` and `detailedNDens` to be shaped and structured for this to work.

If people have different layouts, these are subclassable

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
drewj-tp: Fewer temporary arrays. The in-place operations means we only need to create two arrays. Rather than the intermediate arrays created

It might be splitting hairs. Axial expansion on large problems with lots of components takes a non-zero amount of time. And this feels like part of "every bit helps"

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
drewj-tp: Will add

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
albeanth: TIL

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
albeanth: https://github.com/terrapower/armi/pull/2613/commits/9bb4fa17bc4cab7b79ba0ae2eb5da0197024aa70

armi/reactor/converters/tests/test_axialExpansionChanger_MultiPin.py
albeanth: https://github.com/terrapower/armi/pull/2613/commits/9bb4fa17bc4cab7b79ba0ae2eb5da0197024aa70 & https://github.com/terrapower/armi/pull/2613/commits/02ca8890dca8f013b59e36030a04ce669a1367e1

armi/reactor/converters/tests/test_axialExpansionChanger_MultiPin.py
albeanth: https://github.com/terrapower/armi/pull/2613/commits/9bb4fa17bc4cab7b79ba0ae2eb5da0197024aa70 & https://github.com/terrapower/armi/pull/2613/commits/02ca8890dca8f013b59e36030a04ce669a1367e1

armi/reactor/converters/tests/test_massRedistribution.py
albeanth: https://github.com/terrapower/armi/pull/2613/commits/9bb4fa17bc4cab7b79ba0ae2eb5da0197024aa70 & https://github.com/terrapower/armi/pull/2613/commits/02ca8890dca8f013b59e36030a04ce669a1367e1

armi/reactor/converters/tests/test_massRedistribution.py
albeanth: https://github.com/terrapower/armi/pull/2613/commits/9bb4fa17bc4cab7b79ba0ae2eb5da0197024aa70 & https://github.com/terrapower/armi/pull/2613/commits/02ca8890dca8f013b59e36030a04ce669a1367e1

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
albeanth: https://github.com/terrapower/armi/pull/2613/commits/9bb4fa17bc4cab7b79ba0ae2eb5da0197024aa70

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
```diff
@@ -233,3 +242,66 @@ def _adjustMassParams(self):
...
+        # array * scalar gives us a new array, not a copy. So we can do in-place mutations later
+        try:
+            toData = toData * self.toCompVolume
+            fromData = fromData * self.fromCompVolume
+            toData += fromData
+            toData /= self.newVolume
+        except Exception:
+            msg = (
+                f"Error updating {paramName} on {self.assemblyName} : toComp={self.toComp} : fromComp={self.fromComp}\n"
+                f"Shape fromData {toData.shape}\nShape toData {toData.shape}"
```
albeanth:
```suggestion
                f"Shape fromData {fromData.shape}\nShape toData {toData.shape}"
```

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
drewj-tp: It would be unlikely. But there's no assurances that two components of identical material's get these parameters set

albeanth: Does it make sense to throw in a warning? If it's unlikely, it might not hurt to do so. If it's annoying/aggravating in downstream testing, we can always walk it back.

drewj-tp: I will add it if you want it added

albeanth: let's do it.

drewj-tp: 04a44d29

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
drewj-tp: f34aee9e

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
```diff
@@ -65,11 +65,17 @@ class AxialExpansionChanger:
...
+    detailedAxialExpansion : bool, optional
+        A boolean to indicate whether or not detailedAxialExpansion is to be utilized.
+
     Attributes
     ----------
-    linked: :py:class:`AssemblyAxialLinkage`
+    linked: :class:`~armi.reactor.converters.axialExpansionChanger.assemblyAxialLinkage.AssemblyAxialLinkage`
         establishes object containing axial linkage information
-    expansionData: :py:class:`ExpansionData <armi.reactor.converters.axialExpansionChanger.expansionData.ExpansionData>`
+    expansionData: :class:`~armi.reactor.converters.axialExpansionChanger.expansionData.ExpansionData`
```
john-science, line 78:
```suggestion
    expansionData: :py:class:`~armi.reactor.converters.axialExpansionChanger.expansionData.ExpansionData`
```

john-science: Everywhere else in ARMI it is ":py:class:" not just ":class":. 

I'm just looking for reasons the docs build might have failed.

armi/reactor/converters/axialExpansionChanger/__init__.py
```diff
@@ -11,7 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Enable component-wise axial expansion for assemblies and/or a reactor."""
+"""Enable component-wise axial expansion for assemblies and/or a reactor.
+
+.. currentmodule:: armi.reactor.converters.axialExpansionChanger
```
john-science: I have never seen `.. currentmodule::` before, and I don't see it in a "grep" of ARMI.

Perhaps this is supported by some Sphinx tooling you use on that other project, but not on this particular set of Sphinx tools and versions?

(I'm just looking for reasons the docs build might have failed.)

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
```diff
@@ -233,3 +255,79 @@ def _adjustMassParams(self):
...
+
+    def _adjustDetailedNDens(self):
+        """Update the detailed number density parameter.
+
+        Returns
+        -------
+        bool
+            If the update was performed.
+
+        :meta public:
```
john-science: Is this necessary?

It's a private method. What's the thought here?

doc/user/grids.rst
```diff
@@ -238,7 +238,7 @@ The other version is the corners up lattice. Notice how the ``-`` placeholders a
 
 and a larger example
 
-.. clode-block:: yaml
+.. code-block:: yaml
```
john-science, line 241: WHOOPS

Thanks.

doc/user/radial_and_axial_expansion.rst
```diff
@@ -110,6 +110,13 @@ This section provides an example thermal expansion calculation for a simple cyli
 from a reference temperature of 20°C to 1000°C with example material properties and dimensions as
 shown in the table below.
 
+.. note::
+
+    Implementation of axial expansion is handled through the :mod:`armi.reactor.converters.axialExpansionChanger`
+    module. Developers desiring to learn more can seek there. If you wish to expose a new class for axial expansion,
+    consider implementing the :meth:`~armi.plugins.ArmiPlugin.getAxialExpansionChanger`
```
john-science, line 117:
```suggestion
    consider implementing the :py:meth:`~armi.plugins.ArmiPlugin.getAxialExpansionChanger`
```

doc/user/radial_and_axial_expansion.rst
```diff
@@ -110,6 +110,13 @@ This section provides an example thermal expansion calculation for a simple cyli
 from a reference temperature of 20°C to 1000°C with example material properties and dimensions as
 shown in the table below.
 
+.. note::
+
+    Implementation of axial expansion is handled through the :mod:`armi.reactor.converters.axialExpansionChanger`
```
john-science, line 115:
```suggestion
    Implementation of axial expansion is handled through the :py:mod:`armi.reactor.converters.axialExpansionChanger`
```

doc/user/radial_and_axial_expansion.rst
```diff
@@ -541,6 +551,50 @@ minimization may fail. In this case, a mass weighted temperature is used instead
...
   \hat{T}_0 = \frac{m_{i,0}T_0 + m_{i,1}T_1}{m_{i,0} + m_{i,1}}.
 
+Component parameter updates
+---------------------------
+
+Pin and detailed number densities
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+During the expansion process and redistribution of mass, some parameters beyond nuclide number densities
+are updated via :class:`~armi.reactor.converters.axialExpansionChanger.redistributeMass.RedistributeMass`.
```
john-science, line 561:
```suggestion
are updated via :py:class:`~armi.reactor.converters.axialExpansionChanger.redistributeMass.RedistributeMass`.
```

doc/conf.py
```diff
@@ -475,7 +475,11 @@ def setup(app):
     message="Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.",
 )
 
-intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
+# Path to an objects inventory file to lookup python objects in other modules
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/3", None),
+    "scipy": ("https://docs.scipy.org/doc/scipy/", None),
+}
```
john-science: Sorry, why are you adding scipy here?

I'm not sure I actually love that. That will slow down our docs build, trying to ingest spinx, and our docs build already takes forever.

Can you make a really strong case for this?

doc/user/radial_and_axial_expansion.rst
```diff
@@ -541,6 +551,50 @@ minimization may fail. In this case, a mass weighted temperature is used instead
 
   \hat{T}_0 = \frac{m_{i,0}T_0 + m_{i,1}T_1}{m_{i,0} + m_{i,1}}.
 
+Component parameter updates
+---------------------------
+
+Pin and detailed number densities
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+During the expansion process and redistribution of mass, some parameters beyond nuclide number densities
```
albeanth:
```suggestion
During the expansion/contraction and redistribution of mass process, some parameters beyond nuclide number densities
```

doc/user/radial_and_axial_expansion.rst
```diff
@@ -541,6 +551,50 @@ minimization may fail. In this case, a mass weighted temperature is used instead
...
+are updated via :class:`~armi.reactor.converters.axialExpansionChanger.redistributeMass.RedistributeMass`.
+
+``Component.p.pinNDens`` and ``Component.p.detailedNDens`` are updated if they are populated for both
+components. They are updated proportional to the volume shifted between the "from" (source) component and the "to"
+(destination) component per
+
+.. math::
+    :name: volumeWeightedUpdate
+
+    \rho_t^\prime = \left(\rho_t * V_t + \rho_f * V_f\right) / \left(V_t + V_f\right)
```
albeanth:
```suggestion
    \rho_t^\prime = \frac{\rho_t * V_t + \rho_f * V_f}{V_t + V_f}.
```

doc/user/radial_and_axial_expansion.rst
```diff
@@ -541,6 +551,50 @@ minimization may fail. In this case, a mass weighted temperature is used instead
...
+
+    hm_t' = hm_t + hm_f * \frac{\delta z}{h_f}
+
+and
+
+.. math::
+
+    hm_f' = hm_f * \left(1 - \frac{\delta z}{h_f}\right)
+
+for a height change :math:`\delta z` and height :math:`h_f` in the source component prior to the redistribution.
```
albeanth:
```suggestion
for a given change in height, :math:`\delta z`, and height of the source component prior to the redistribution, :math:`h_f`.
```

albeanth (commented):
Doc nit picks and personal preferences 🤷 

Once the docs build is working I'll give the compiled docs a gander and approve if all looks good.

armi/reactor/converters/axialExpansionChanger/__init__.py
drewj-tp: This is part of [vanilla sphinx](https://www.sphinx-doc.org/en/master/usage/domains/python.html#directive-py-currentmodule) to make any later class/method/function lookups easier by saying "start the search in this directory"

I can remove and change the one reference below. No problem

armi/reactor/converters/axialExpansionChanger/axialExpansionChanger.py
drewj-tp: For a pure python project where the [default "domain"](https://www.sphinx-doc.org/en/master/usage/domains/index.html#directive-default-domain) is python, `:py:class:` and `:class:` are identical

If we were to mix languages like C++ that also provides a [`:cpp:class:`](https://www.sphinx-doc.org/en/master/usage/domains/cpp.html#role-cpp-class) role, then the value of explicitly using the domains is clear.

The `primary_domain` configuration [defaults to python](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-primary_domain) so the `:py:` is not necessary.

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
drewj-tp: I wanted to expose these in the documentation so other developers could more easily see them in the docs and know what to subclass w/o looking at source.

But looking at my local builds, they still aren't going through. I'll remove

doc/user/grids.rst
drewj-tp: You've got it in #2622 and it's tracked in #2621 FWIW

doc/user/radial_and_axial_expansion.rst
drewj-tp: I would prefer to not as it's not necessary but will include this if it's required.

doc/user/radial_and_axial_expansion.rst
drewj-tp: I would prefer to not as it's not necessary but will include this if it's required.

doc/user/radial_and_axial_expansion.rst
drewj-tp: I would prefer to not as it's not necessary but will include this if it's required.

doc/conf.py
drewj-tp: We have a cross-reference to scipy

https://github.com/terrapower/armi/blob/321d9466dc476cbaac97fe97e1c1fa3455223e10/armi/reactor/converters/axialExpansionChanger/redistributeMass.py#L146

Without adding this, there is no hyper link. If you want this to be removed, I will change the docs to not cross reference scipy

albeanth: If it sufficiently slows down the docs build, we can remove it and just reference it old school with a bibliography-style reference.

john-science: I guess it doesn't slow the build down too much.

I'm seeing a bunch of these kinds of things in the latex build:

```
Overfull \hbox (156.11163pt too wide) in paragraph at lines 121203--121205
[][][]\TU/FreeMono(0)/m/n/10 scipy.sparse.csc_matrix[][] \TU/FreeSerif(0)/m/n/1
0 (https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csc_matrix
.html#scipy.sparse.csc_matrix) 
```

The intersphinx mapping isn't visible during the HTML build, for some reason.

I guess this is fine. I don't love it. The build won't work without internet, or on a cluster without internet access. I guess you should add some sort of documentation somewhere to that affect. Thoughts?

armi/reactor/converters/axialExpansionChanger/__init__.py
john-science: Okay, from the [docs](https://www.sphinx-doc.org/en/master/usage/domains/python.html#directive-py-currentmodule):

```
.. py:currentmodule:: name
This directive tells Sphinx that the classes, functions etc. documented from here are in the given module (like [py:module](https://www.sphinx-doc.org/en/master/usage/domains/python.html#directive-py-module)), but it will not create index entries, an entry in the Global Module Index, or a link target for [py:mod](https://www.sphinx-doc.org/en/master/usage/domains/python.html#role-py-mod). This is helpful in situations where documentation for things in a module is spread over multiple files or sections – one location has the [py:module](https://www.sphinx-doc.org/en/master/usage/domains/python.html#directive-py-module) directive, the others only [py:currentmodule](https://www.sphinx-doc.org/en/master/usage/domains/python.html#directive-py-currentmodule).
```

Okay, so I see this "but it will not create index entries, an entry in the Global Module Index, or a link target for [py:mod](https://www.sphinx-doc.org/en/master/usage/domains/python.html#role-py-mod)..."

So, will this be noticable when we build the API docs?

doc/conf.py
drewj-tp: reverted
4263c2ac

armi/reactor/converters/axialExpansionChanger/__init__.py
drewj-tp: removed
3ebe57b0

john-science (approved):
The important feature is small and clear.

But thank you for all your work adding real testing and documentation.

You rock!

doc/user/radial_and_axial_expansion.rst
```diff
@@ -541,6 +551,50 @@ minimization may fail. In this case, a mass weighted temperature is used instead
...
+are updated via :class:`~armi.reactor.converters.axialExpansionChanger.redistributeMass.RedistributeMass`.
+
+``Component.p.pinNDens`` and ``Component.p.detailedNDens`` are updated if they are populated for both
+components. They are updated proportional to the volume shifted between the "from" (source) component and the "to"
+(destination) component per
+
+.. math::
+    :name: volumeWeightedUpdate
+
+    \rho_t^\prime = \frac{\rho_t * V_t + \rho_f * V_f}{V_t + V_f}.
```
albeanth:
```suggestion
    \rho_t^\prime = \frac{\rho_t V_t + \rho_f V_f}{V_t + V_f}.
```

doc/user/radial_and_axial_expansion.rst
```diff
@@ -541,6 +551,50 @@ minimization may fail. In this case, a mass weighted temperature is used instead
...
+
+    hm_t' = hm_t + hm_f * \frac{\delta z}{h_f}
+
+and
+
+.. math::
+
+    hm_f' = hm_f * \left(1 - \frac{\delta z}{h_f}\right)
+
+for a given change in height, :math:`\delta z`, and height of the source component prior to the redistribution, :math:`h_f`.
```
albeanth:
```suggestion
    hm_t' = hm_t + hm_f \frac{\Delta z}{h_f},

and

.. math::

    hm_f' = hm_f \left(1 - \frac{\Delta z}{h_f}\right),

for a given change in height, :math:`\Delta z`, and height of the source component prior to the redistribution, :math:`h_f`.
```

armi/reactor/converters/axialExpansionChanger/__init__.py
```diff
@@ -11,7 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Enable component-wise axial expansion for assemblies and/or a reactor."""
+"""Enable component-wise axial expansion for assemblies and/or a reactor.
+
+Primary class for performing axial expansion, thermal or otherwise, is
+:class:`~armi.reactor.converters.axialExpansionChanger.AxialExpansionChanger`
```
albeanth, line 17: for w/e reason this isn't compiling into a clickable link for HTML docs

albeanth (approved):

armi/reactor/converters/axialExpansionChanger/__init__.py
drewj-tp: Oh yep b/c the path has one more `axialExpansionChanger`. Coming up

PR #2617, opened 2026-08-03
Removing _Mixture material from ARMI.

### LINKED ISSUE #2614: Parts of ARMI depend on `armi.materials` (which requires all downstream apps to use them in their materials namespace)
After recent work on the `materialNamespaceOrder`, we expected to be able to do this downstream: `setMaterialNamespaceOrder(["myapp.materials", "dir:/path/to/yaml/materials"])`.

However, uniform mesh conversion (UMC) uses both the `_Mixture` and `Void` materials in ARMI. So the above won't work, and when the UMC code is hit, it fails since it can't find one or the other of these materials.

I'm not sure what to do about `Void`, but @mgjarrett did attempt to remove `_Mixture` (it's just a stub class that inherits from ARMI's `Material` class). But that slows down UMC by ....a lot. More than 2x slower. ~So we absolutely cannot just swap it for `Material`.~ (UPDATE: actually we can swap it and that's the solution)

CC @alexhjames

mgjarrett:

Remove the `_Mixture` material from ARMI. This material class existed only to be a dummy material used for temporary composite objects that are created from a mixture of other composites. The class has no special features beyond those of the base `Material` class; the only thing that this does functionally is set the name of the material to `_Mixture`. This material name is not critical to any known downstream use cases.

## SCR Information

Change Type: features

One-Sentence Rationale: Downstream applications that define a materials namespace that doesn't include `armi.materials` would fail because the `_Mixture` material is not present.

One-line Impact on Requirements: Minor implementation change to I_ARMI_BLOCK_HOMOG. Removed T_ARMI_MAT_FRACS1.

---

## Checklist


opotowsky (approved):
@john-science when I made the linked issue I thought this had already been tried. A little miscommunication! This will close the ticket just fine :-)

CC @alexhjames for awareness

john-science:
@mgjarrett Has this been tested downstream?

john-science:
Less is more.

If we don't _need_ this, I'm totally in favor.

john-science (approved):
If all the CI passes in ARMI and your big downstream test jobs, then I am satisfied.

What a fun way to solve problem! Thanks @mgjarrett !

PR #2618, opened 2026-08-04
Fixing Database.mergeHistory for EOL and error time steps

### LINKED ISSUE #1960: Further issues found with the EOL being written to the Database

john-science:

This is a follow up to https://github.com/terrapower/armi/pull/2607. That solution solve the problem, but opened up another problem in `Database.mergeHistory()`, which also needs to support "error" and "EOL" time nodes.

**NOTE**: I also went through all the rest of the uses of `Database.genTimeSteps()` in ARMI, and they are not similarly broken.

close https://github.com/terrapower/armi/issues/1960

## SCR Information

 Change Type: fixes

One-Sentence Rationale: MergeHistory should not fail if there are special error or EOL time steps in the DB.

---

## Checklist

### DIFF
--- a/armi/bookkeeping/db/tests/test_database.py
+++ b/armi/bookkeeping/db/tests/test_database.py
@@ -1055,7 +1055,7 @@ class TestSimplestDatabaseItems(unittest.TestCase):
                 self.timeSteps = timeSteps
 
             def genTimeStepGroups(self):
-                return [FakeHDF5(i) for i in range(len(self.timeSteps))]
+                return [FakeHDF5(t) for t in self.timeSteps]
 
             def genTimeSteps(self):
                 return self.timeSteps
@@ -1063,11 +1063,11 @@ class TestSimplestDatabaseItems(unittest.TestCase):
         db.h5db = FakeHDF5(123)
         db.mergeHistory(MockDB([(0, 0), (0, 1)]), 0, 1)
 
-        self.assertListEqual(results, [0])
+        self.assertListEqual(results, [(0, 0)])
 
         results = []
         db.mergeHistory(MockDB([(0, 0), (0, 0, "error"), (0, 1)]), 0, 1)
-        self.assertListEqual(results, [0])
+        self.assertListEqual(results, [(0, 0), (0, 0, "error")])
 
         # remove the fake H5 file or the DB cleanup in ARMI's context.py will panic
         db.h5db = None
--- a/armi/bookkeeping/db/tests/test_database.py
+++ b/armi/bookkeeping/db/tests/test_database.py
@@ -1033,14 +1033,21 @@ class TestSimplestDatabaseItems(unittest.TestCase):
         db.h5db = None
 
     def test_mergeHistory(self):
+        """Test some edge cases of Database.mergeHistory.
+
+        This test uses some mock database and H5 file tooling, so we can more easily test edge cases in the ........
+        """
         # mock up a test DB
         dbPath = "test_mergeHistory.h5"
         db = Database(dbPath, "w")
         db.close()
 
+        # Optimization: We use this "results" list so we do not need to write to a real H5 file for this tests.
         results = []
 
         class FakeHDF5:
+            """As an optimization, and simplification, a stand-in for a H5 group."""
+
             name = "fake"
 
             def __init__(self, x):
@@ -1050,6 +1057,8 @@ class TestSimplestDatabaseItems(unittest.TestCase):
                 return results.append(a.x)
 
         class MockDB:
+            """As an optimization, and simplification, a stand-in for needing a second ARMI DB file in the tests."""
+
             def __init__(self, timeSteps):
                 self.versionMajor = 3
                 self.timeSteps = timeSteps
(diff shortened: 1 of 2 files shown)


armi/bookkeeping/db/tests/test_database.py
```diff
@@ -1032,15 +1032,72 @@ def test_genTimeStepGroups(self):
         # remove the fake H5 file or the DB cleanup in ARMI's context.py will panic
         db.h5db = None
 
+    def test_mergeHistory(self):
+        """Test some edge cases of Database.mergeHistory.
+
+        This test uses some mock database and H5 file tooling, so we can more easily test edge cases in the ........
```
john-science:
```suggestion
        This test uses some mock database and H5 file tooling, so we can more easily test edge cases.
```

armi/bookkeeping/db/tests/test_database.py
```diff
@@ -1032,15 +1032,72 @@ def test_genTimeStepGroups(self):
...
+
+        # test 2: should grab the first time node and the error time node
+        results = []
+        db.mergeHistory(MockDB([(0, 0), (0, 0, "error"), (0, 1)]), 0, 1)
+        self.assertListEqual(results, [(0, 0), (0, 0, "error")])
+
+        # test 3: should grab all 5 time nodes, including the EOL
+        results = []
+        db.mergeHistory(MockDB([(0, 0), (0, 1), (1, 0), (1, 1), (1, 1, "EOL")]), 2, 4)
+        self.assertListEqual(results, [(0, 0), (0, 1), (1, 0), (1, 1), (1, 1, "EOL")])
```
drewj-tp, line 1090: Odd but I guess it works to show we can handle an EOL state too

armi/bookkeeping/db/tests/test_database.py
```diff
@@ -1032,15 +1032,72 @@ def test_genTimeStepGroups(self):
...
+
+        # test 1: should grab the first 4 of 6 time nodes
+        results = []
+        db.mergeHistory(MockDB([(0, 0), (0, 1), (1, 0), (1, 1), (2, 0), (2, 1)]), 2, 0)
+        self.assertListEqual(results, [(0, 0), (0, 1), (1, 0), (1, 1)])
+
+        # test 2: should grab the first time node and the error time node
+        results = []
+        db.mergeHistory(MockDB([(0, 0), (0, 0, "error"), (0, 1)]), 0, 1)
+        self.assertListEqual(results, [(0, 0), (0, 0, "error")])
```
drewj-tp, line 1085: Would it be possible to have `c00n00error` and `c00n01`? There was an error fatal enough to write to a database but we somehow wrote the next statepoint?

Maybe more realistically is someone had an error and wants to restart prior to that error.

drewj-tp (commented):
The mocking / optimization is almost too much to know if this is really going to do what we want.

I guess the alternative would be for every one of these small steps to use something like `unittest.mock.patch on
- `self.db.copy` to know that we copied the right things
- `inputDB.genTimeSteps` and `inputDB.genTimeStepGroups`

which doesn't feel better. IDK I'm rambling here. 

The changes to `database.py` look promising. Please run this through downstream testing as well

armi/bookkeeping/db/tests/test_database.py
john-science: LOL

Yeah, I'm calling that out in the unit testing because I want to show we are explicitly handling everything: error and EOL.

The other option was to support neither, but the conversation was this was the more conservative option.

#shrug

I bet we don't need either of those. But no one wanted to risk it.

armi/bookkeeping/db/tests/test_database.py
john-science: I mean, yeah, I think dropping error and EOL time steps seems valid.

I think, in reality, people probably don't care about the "error" time steps enough that they want to merge them in. But @opotowsky suggested this was the more conservative option.  The thought goes that no one will complain about having too much data in their output files.

I can flip that and drop EOL and error timesteps.  That's not hard to do, or anything. #shrug

opotowsky: I don't have a strong opinion, we'd just want to make sure it wouldn't hurt anyone's workflow by changing prior behavior.

opotowsky (approved):
I see the other review material but I'm throwing in my 2c approval. 

My take: The unit tests don't make sense bc the behavior doesn't make sense, and John was just trying to retain the nonsensical behavior. We should just make a ticket to clean up the behavior

john-science:
> We should just make a ticket to clean up the behavior

If we can find some real world users of the restart cases, we could get some ground truth on what the best functionality here should be.

PR #2608, opened 2026-07-27
Moving downstream-imported `test_blocks` machinery to `armi.testing`

yardasol:
## What is the change? Why is it being made?

This PR moves and renames the following functions and variables, previously stored in `armi.reactors.tests.test_blocks`, to `armi.testing.singleBlocks`:
* `test_blocks.buildSimpleFuelBlock()` ->
  `testing.buildSimpleFuelHexBlock()`
* `test_blocks.buildLinkedFuelBlock()` ->
  `testing.buildLinkedFuelHexblock()`
* `test_blocks.loadTestBlock()` -> `testing.buildComplexHexBlock()`
* `test_blocks.NUM_PINS_IN_TEST_BLOCK` -> `testing.NUM_PINS_IN_COMPLEX_HEX_BLOCK`

 This is done to prevent downstream plugins from importing directly from the ARMI testing suite.

There are some additional pieces of machinery from around ARMI's test suite moved to `armi.testing` for consistency.
* `test_blocks.applyDummyData()` -> `testing.applyDummyData()`
* `test_assemblies.makeTestAssembly()` -> `testing.buildEmptyHexAssembly()`
* `test_blockConverter.buildSimpleFuelBlockNegativeArea()` -> `testing.buildSimpleFuelHexBlockNegativeArea()`
* `test_blockConverter.buildSimpleFuelBlockNegativeAreaBond()` ->`testing.buildLinkedFuelBlockNegativeArea()`

Any feedback and suggested improvements on this design choice would be appreciated.

## SCR Information

Change Type: features

One-Sentence Rationale: Improving testing suite makes downstream code more resilient.

One-line Impact on Requirements: NA

---

## Checklist

- [x] This PR has only [one purpose or idea](https://terrapower.github.io/armi/developer/tooling.html#one-idea-one-pr).
- [x] [Tests](https://terrapower.github.io/armi/developer/tooling.html#test-it) have been added/updated to verify any new/changed code.
- [x] The [documentation](https://terrapower.github.io/armi/developer/tooling.html#document-it) is still up-to-date in the `doc` folder.
- [x] The code style follows [good practices](https://terrapower.github.io/armi/developer/standards_and_practices.html).
- [x] The dependencies are still up-to-date in `pyproject.toml`.

PR #2622, opened 2026-08-05
Getting ARMI docs building correctly

### LINKED ISSUE #2620: Duplicate test tags for armi.matProps and armi.materials
I'm not sure if this in intended but I noticed a warning building some docs
```
[warn] Need could not be created: A need with ID 'T_ARMI_MAT_FRACS0' already exists. [needs.create_need]
```
You can see it in CI too https://github.com/terrapower/armi/actions/runs/31024337991/job/92368796729#step:7:778

```
armi\matProps\tests\test_constituent.py
43:            :id: T_ARMI_MAT_FRACS0

armi\materials\tests\test_materials.py
614:            :id: T_ARMI_MAT_FRACS0
```

### LINKED ISSUE #2621: Unknown directive "clode-block" building docs

john-science:

There were several things going on in the ARMI docs. I tried to fix them all to get the docs building correctly again.

1. The ARMI docs API is currently not being built due to some pathing changes made in this PR: https://github.com/terrapower/armi/pull/2608 Essentially, this path:

```python
from armi.reactor.tests.test_blocks import buildSimpleFuelBlock
```

became this path:

```python
from armi.testing import buildSimpleFuelHexBlock
```

And that was fixed everywhere but the docs.

2. Also, there was a broken import in the docs (spelling of the word "testing"):

```python
from armi.testng import loadTestReactor
```

3. And, just because I'm looking at the Sphinx log, I cleaned up some other Sphinx warnings I found.

4. And there was a duplicate unit test tag that I fixed.

5. And there was a typo where the `code-block` directive was spelled `clode-block`.

close #2621
close #2620

## SCR Information

Change Type: docs

One-Sentence Rationale: The docs depend on a working API.

---

## Checklist

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -87,6 +87,7 @@ jobs:
           name: html-docs
           path: doc/_build/html
           retention-days: 5
+          include-hidden-files: true
       - name: Archive PDF Docs
         uses: actions/upload-artifact@v7
         with:
--- a/armi/materials/tests/test_materials.py
+++ b/armi/materials/tests/test_materials.py
@@ -611,7 +611,7 @@ class ThoriumTests(AbstractMaterialTest, unittest.TestCase):
         Test default mass fractions.
 
         .. test:: The materials generate nuclide mass fractions.
-            :id: T_ARMI_MAT_FRACS0
+            :id: T_ARMI_MAT_FRACS6
             :tests: R_ARMI_MAT_FRACS
         """
         self.mat.setDefaultMassFracs()
--- a/armi/plugins.py
+++ b/armi/plugins.py
@@ -260,7 +260,7 @@ class ArmiPlugin:
         It is only meant to be run one time. Because ``db.load()`` may be run several times in the same script, it is
         strongly recommended that developers implement the following pattern when adding this hook to any plugin:
 
-        .. code::
+        .. code-block::
             from armi.utils import onlyRunOnce
 
             @staticmethod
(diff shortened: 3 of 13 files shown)


yardasol (approved):
Good changes @john-science!

.github/workflows/docs.yaml
```diff
@@ -87,6 +87,7 @@ jobs:
           name: html-docs
           path: doc/_build/html
           retention-days: 5
+          include-hidden-files: true
```
drewj-tp, line 90: We're talking about this offline. We're adding this so the `.apimodules` directory is added to the archive upload and we can look at the html docs from PRs. We want those api docs.

But I want to propose not doing this because hidden files, directories and files that start with `.` on unix (where actions run) **could** contain secrets in general. Think `~/.ssh/` or an entire `.git/` tree

An alternative would be to change

https://github.com/terrapower/armi/blob/321d9466dc476cbaac97fe97e1c1fa3455223e10/doc/conf.py#L64

https://github.com/terrapower/armi/blob/321d9466dc476cbaac97fe97e1c1fa3455223e10/doc/conf.py#L66

This way they're not considered "hidden" by the archive uploader and we'll get them in the zip.

john-science: I'll make that change.

It'll be slightly bigger than this PR though. Because we also have the `.static` directory, where we store images, and we reference that in a few dozen places. And that `.static/` directory was made hidden because if it's visible it breaks the build in a couple of ways (the API docs auto-finding tooling keeps trying to peak inside, if I remember correctly.)

PR #2624, opened 2026-08-06
Attempting to speed up the docs build on github

john-science:

Here I attempt to speed up the ARMI docs build (specifically on GitHub CI) by:

* letting `sphinx-build` run in parallel
* removing intersphinx entirely
* removing unused sphinx extensions from `conf.py`
* removing unused sphinx dependencies from `pyproject.toml`
* removing some `apt-get` dependencies we don't need

### Notes

* It is tricky to prove this PR speeds up the doc build, because the amount of time the doc build takes is SO variable.
* There are obviously three doc features we could remove to greatly speed up the build, but we aren't doing that: tutorials, gallery, and API docs.

## SCR Information

Change Type: docs

One-Sentence Rationale: Faster CI speeds up developer work.

---

## Checklist


john-science:
@yardasol This isn't a very big speed bump. But that's fine. My only real concern here is the dependencies I removed from `pyprojec.toml` and `conf.py` might break some feature of the docs (HTML or PDF).

So, the annoying part of this PR is paging through the docs to make sure I didn't break any features. Sorry! (I will help with this, but would love a second set of eyes.)

doc/user/inputs.rst
yardasol: If you're gonna remove the `Section` column, you should also remove the reference to this in the docs 
```suggestion
summary of the keywords and their relevant status can be found below.
```

yardasol (changes requested):
I looked at both docs:

HTML:
* Math is rendering correctly in HTML
* Figures seem to be rendering correctly
* Scripts still downloadable

PDF
* Math still rendering
* Figures and blocks are still rendering
* Scripts still render in the document

There's one change that needs to happen based on the change you made to the property components summary table, but I think everything else is working. Are there any particular snags I should keep an eye out for outside of what I already looked at?

yardasol (approved):
Good work on this @john-science! I always like having cleaner build infrastructure!

john-science:
> Are there any particular snags I should keep an eye out for outside of what I already looked at?

The most important things are inside the "QA docs" section, the subsections: SRSD, SDID, SCR, and STR.

But the other stuff I would expect to fail the easiest are the "Tutorial" and "Gallery" sections.

PR #2625, opened 2026-08-06
Fixing Block.getNumPins when there are multiple clads around a pin

### LINKED ISSUE #1946: Block.getNumPins broken for C5G7 model - multiple cladding components
If you load the C5G7 test reactor and grab an assembly, you get a hilariously large number of pins
```python
>>> r = armi.init(fName="armi/testing/reactors/c5g7/c5g7-settings.yaml").r
>>> fuelBlock = r.core[0][0]
>>> fuelBlock.getNumPins()
528
```

In that model, there are two clad components per lattice site

https://github.com/terrapower/armi/blob/c6f46e5d04dc265e7a8e470fdd1977c9a4e16e3e/armi/tests/c5g7/c5g7-blueprints.yaml#L86-L114

So you double count the sites. There are 17x17 total lattice sites, some of which are empty, so the upper bound on "real" pins is 289

_Originally posted by @drewj-tp in https://github.com/terrapower/armi/discussions/1900#discussioncomment-10908081_

john-science:

The method `Block.getNumPins()` was double-counting the number of pins in a block if that pin had two claddings around it.

Fixing this bug requires a significant rewrite of the `Block.getNumPins()` method. We need to be able to group components now, not just by their type (flag), but also by their `.spatialLocator`. This is doable, but it can't be done the easy way, using dictionaries or sets, because not al spatial locators are hashable.

But the end result is a minor change to the conceptual logic of what this method is doing.

For consistency, I had to refactor `Block.getPinNums()` and `Block.getPinLocations()`, so they share the same base logic. Before this PR... strangely... these two methods did not share the same logic and could thus disagree.

close #1946

## SCR Information

Change Type: fixes

One-Sentence Rationale: ARMI should not double count pins if there are multiple claddings around the pins.

One-line Impact on Requirements: Rewrite I_ARMI_BLOCK_NPINS to correct edge case errors for R_ARMI_BLOCK_NPINS.

---

## Checklist


armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,108 @@ def getSortedComponentsInsideOfComponent(self, component):
...
         Notes
         -----
-        If you just want sorted components in this block, use ``sorted(self)``. This will never
-        include any ``DerivedShape`` objects. Since they have a derived area they don't have a well-
-        defined dimension. For now we just ignore them. If they are desired in the future some
-        knowledge of their dimension will be required while they are being derived.
+        If you just want sorted components in this block, use ``sorted(self)``. This will never include any
+        ``DerivedShape`` objects. Since they have a derived area they don't have a well- defined dimension. For now we
+        just ignore them. If they are desired in the future some knowledge of their dimension will be required while
+        they are being derived.
```
john-science:
```suggestion
        just ignore them.
```

armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,107 @@ def getSortedComponentsInsideOfComponent(self, component):
...
+            [c for c in self.iterComponents(compType) if isinstance(c, basicShapes.Circle)]
+            for compType in PIN_COMPONENTS
+        ]
+
+        # Group together components by spatial locator, THEN count the number of pins by group and by type/flag
+        numPinsByType = []
+        for comps in compsByType:
+            compsByLocation = []
+            while len(comps):
+                c = comps.pop()
```
drewj-tp: I usually see a pattern of `while len(list_like)` for some operation where you may be adding to `list_like` through the iteration. Why not use a for loop?

```suggestion
            for c in comps:
```

armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,107 @@ def getSortedComponentsInsideOfComponent(self, component):
...
+                    if c.spatialLocator == cByLoc[0].spatialLocator:
+                        index = i
+                        break
+
+                if index < 0:
+                    # this spatialLocator was NOT seen before
+                    compsByLocation.append([c])
+                else:
+                    # this spatialLocator was seen before, group with the others
+                    compsByLocation[index].append(c)
```
drewj-tp: The logic here is we are making a list of lists, where the sublists contain components that share a spatial locator. And if we go through all the things we've previously seen with no matches, we correctly infer we have not seen a match. Good enough.

But at the point we've found a match, we have the object we'd like to modify directly in hand: `cByLoc`. And we use the [`for/break/else` pattern](https://docs.python.org/3/tutorial/controlflow.html#else-clauses-on-loops) to handle the case where we iterate through the whole list w/o hitting a break to write

```suggestion
                for cByLoc in compsByLocation:
                    if c.spatialLocator == cByLoc[0].spatialLocator:
                        # we have seen this location before
                        cByLoc.append(c)
                        break
                else:
                    # we have not seen this location before
                    compsByLocation.append([c])
                
```
Then we avoid the `index = -9` magic number and the need to do another conditional check on `index < 0` at the end.

drewj-tp: GH does not like syntax highlighting for this. But it's valid python

armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,107 @@ def getSortedComponentsInsideOfComponent(self, component):
...
+
+            # Count pins. If multiple components are at the same location, take the one that has the most pins.
+            numPins = [max([int(c.getDimension("mult")) for c in csByLocs]) for csByLocs in compsByLocation]
```
drewj-tp, line 1012: This implies that components that have identical spatial locators will have different multiplicity. Where `compsByLoc[0][0].getDimension("mult") != compsByLocation[0][1].getDimension("mult")`

I would expect we could grab the first component in our `compsByLocation` and be content. Did you see evidence to the contrary?


```suggestion
            # Count pins. Components that have identical spatial locators should have identical multiplicities
            numPins = [int(csByLocs[0].getDimension("mult")) for csByLocs in compsByLocation]
```

armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,107 @@ def getSortedComponentsInsideOfComponent(self, component):
...
+            numPins = 0 if not numPins else sum(numPins)
```
drewj-tp: Sum of an empty list is valid and returns zero

```suggestion
            numPins = sum(numPins)
```

armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,107 @@ def getSortedComponentsInsideOfComponent(self, component):
...
+        """
+        numPinsByType = self.__getPinLikeCounts()
+
+        # Find which locations of the component type/flag yields the highest pin count
+        maxI = -999
+        pins = []
+        for i, comps in numPinsByType:
+            if i >= maxI:
+                maxI = i
+                pins = comps
```
drewj-tp: Pedantic non-blocking request to not use `i` nor `maxI` here. `i` is not an iteration variable but the number of pins for a given type. So, it may not change loop over loop. Recommend `maxPinsPerType` for `maxI` and `pinsThisType` for `i`

It's more characters but I think the level of descriptiveness helps given we're not doing like an `enumerate` call, where `i` would be different loop to loop.

armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,107 @@ def getSortedComponentsInsideOfComponent(self, component):
...
+        """Produce all the index locations for pins in the block.
+
+        Returns
+        -------
+        list[grids.IndexLocation]
+            Integer locations where pins can be found in the block.
+
+        Notes
+        -----
+        Only components with ``Flags.CLAD`` are considered to define a pin's location.
```
drewj-tp: I don't think this holds anymore since we're looking at `PIN_COMPONENTS`

armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,107 @@ def getSortedComponentsInsideOfComponent(self, component):
...
+            else:
+                pinLocs.append(spatLoc)
+
+        return pinLocs
+
+    def getPinCoordinates(self) -> np.ndarray:
+        """
+        Compute the local centroid coordinates of any pins in this block.
+
+        The pins must have a CLAD-flagged component for this to work.
```
drewj-tp: I don't think this holds anymore since we're looking at `PIN_COMPONENTS`

armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,107 @@ def getSortedComponentsInsideOfComponent(self, component):
...
+        list[grids.IndexLocation]
+            Integer locations where pins can be found in the block.
+
+        Notes
+        -----
+        Only components with ``Flags.CLAD`` are considered to define a pin's location.
+        """
+        numPinsByType = self.__getPinLikeCounts()
+
+        # Find which locations of the component type/flag yields the highest pin count
```
drewj-tp, line 1028: The consequence here is we are assuming that there is one flag that is defined at all pin locations. I imagine in most cases this holds (e.g., even if there are multiple pin types in a block, they probably have cladding)

This will be a problem in the case you have a pin that does not have those flags. 

I'm trying to think if this is a deal breaker or not.

It's easy to think up examples like "What if one pin is a literal steel rod and it's flagged with `Flags.SLUG` or `Flags.TEST` or `Flags.STRUCTURE`" but I'm not sure how practical / feasible / realistic that would be.

Let me think on this / call in some support. It's _probably_ fine but there's a definite edge case lying in wait.

armi/reactor/blocks/block.py
```diff
@@ -986,24 +1077,19 @@ def getNumPins(self):
...
-                    (int(c.getDimension("mult")) if isinstance(c, basicShapes.Circle) else 0)
-                    for c in self.iterComponents(compType)
-                ]
-            )
-            for compType in PIN_COMPONENTS
-        ]
-        return 0 if not nPins else max(nPins)
+        numPinsByType = self.__getPinLikeCounts()
+
+        # Return whichever component type/flag yields the highest pin count
```
drewj-tp, line 1081: Same consequence if you have a pin that does not have any of the flags as any other location. Same uneasiness if this is a likely case or not.

armi/reactor/blocks/block.py
```diff
@@ -986,24 +1077,19 @@ def getNumPins(self):
...
+        return 0 if not numPinsByType else max([c[0] for c in numPinsByType])
```
drewj-tp, line 1082: `max([], default=0)` works on empty sequences

```suggestion
        return max([c[0] for c in numPinsByType], default=0)
```

armi/reactor/tests/test_blocks.py
```diff
@@ -1508,6 +1508,49 @@ def test_getNumPins(self):
...
+            :id: T_ARMI_BLOCK_NPINS2
+            :tests: R_ARMI_BLOCK_NPINS
+        """
+        # create a fuel block and add a single clad component, surrounding a single fuel component, with 8 pins
+        b = blocks.HexBlock("empty")
+
+        randoComp1 = basicShapes.Hexagon("hexagon", "HT9", 100, 100, 1, 0, 0)
+        b.add(randoComp1)
+        randoComp2 = basicShapes.Hexagon("hexagon", "HT9", 100, 100, 1, 0, 0)
+        b.add(randoComp2)
```
drewj-tp: What is the utility of the second overlapping hexagon?

```suggestion
```

armi/reactor/composites.py
```diff
@@ -3109,11 +3109,19 @@ def printContents(self, includeNuclides=True):
...
+        In this, most generic method, there can be multiple children per spatial location. In the version of this method
+        on Core, it is assumed only one child exists per location.
+        """
+        runLog.extra("Generating location-to-children lookup table.")
         self.childrenByLocator = {}
         for child in self:
-            self.childrenByLocator[child.spatialLocator] = child
+            if child.spatialLocator not in self.childrenByLocator:
+                self.childrenByLocator[child.spatialLocator] = []
+            self.childrenByLocator[child.spatialLocator].append(child)
```
drewj-tp: Hm I can't find use of `self._genChildByLocationLookupTable` outside of the core specialization. Can we remove the generic implementation?

Otherwise, please add testing.

drewj-tp: It seems like something we _could_ use for this purpose: we have `Block.childrenByLocator` that already houses all the unique spatial locators for the child components. That's maybe half of what was added in this PR for grouping components by spatial location.

But I don't think we'd want to run this every time we look for pins. Maybe store it on a private attribute, and populate it once when requested? Clearing out if components are added or removed? Like
```python
from types import MappingProxyType  # wrap a dict in a read-only wrapper
class Block(...):
    def __init__(self, ...):
        self._childrenByLocator = None

    @property
    def childrenByLocator(self):
        if self._childrenByLocator is None:
            self._genChildByLocationLookupTable()
        return MappingProxyType(self._childrenByLocator)

    def add(self, ...):
        ...
        self._childrenByLocator = None

    def remove(self, ...):
        ...
        self._childrenByLocator = None
```
Could be scope creep and happy if we don't go down this route. Just thought it could be an alternative way to achieve what we're trying to achieve with `self.__getPinLikeCounts()`

drewj-tp (changes requested):
Thanks @john-science! Thank you for digging into this.

Given the passing of current testing, I have confidence this should be backwards compatible. But we may find some edge cases.

There's one item I'd like to follow up on with some people before resolution.

armi/reactor/blocks/block.py
john-science: You were the one wanted to define pins based entirely on just clad. This list is larger than that, and thus should be less susceptible to this, right?

> This will be a problem in the case you have a pin that does not have those flags.

I mean, if you want to have a pin defined in ARMI, you have to use one of 8 flags. That doesn't seem like a very serious limitation. 

And, please remember, this ticket is about solving a very minor edge case: what if a pin has two clads, not just one. This ticket is NOT "let's redesign all pins in ARMI", which is a different type of [discussion](https://github.com/terrapower/armi/discussions/2234). (Albeit an important one.)

armi/reactor/tests/test_blocks.py
john-science: What are you thinking here?

armi/reactor/blocks/block.py
john-science: Hmm... this seems less safe:

```python
>>> numPinsByType = [[]]
>>> max([c[0] for c in numPinsByType], default=0)
  File "<python-input-3>", line 1, in <module>
    max([c[0] for c in numPinsByType], default=0)
         ~^^^
IndexError: list index out of range
```

But, I suppose, if I trust the logic above can't possibly let that happen... let me look at that.

armi/reactor/tests/test_blocks.py
drewj-tp: The suggested change is to remove this component

armi/reactor/blocks/block.py
drewj-tp: Yeah that's fair. Disregard the `max(..., default=0)`

armi/reactor/blocks/block.py
drewj-tp: > if you want to have a pin defined in ARMI, you have to use one of 8 flags. That doesn't seem like a very serious limitation.

This is okay if we document it.

I'm trying to think through limitations and edge cases of the proposed change and how likely people are to find those cases.

keckler: A couple thoughts:
* EBR-II has many assemblies that contain "dummy" pins. These are solid steel rods that take the place of other pins in the assembly in order to reduce reactivity. Since they are literally named dummy pins, it is not uncommon for ARMI's auto-flagging to give them the DUMMY flag.
* I am personally unsure if there is a good reason for ARMI to even have a method such as `getNumPins()`. As we have seen this is a concept that is very application dependent and fraught with edge cases. I think it would serve most use cases better if developers instead just used flags appropriately for their application, and counted the components with those flags. Giving developers a tool like `getNumPins()` that applies to an ambiguous concept like a "pin" is maybe just an unwise decision. 
    * That being said, I know this already exists and is used, so getting rid of it is probably not an option.

armi/reactor/blocks/block.py
john-science: Actually, I am trying to empty out the `comps` list, so we don't keep an extra reference to these around. Because that is where most of the errors I found in testing came from.

Obviously your way was the first thing I tried.

armi/reactor/blocks/block.py
john-science: I agree `-9` instead of `-1` was cheeky. I can revert that if you like.

I tend to dislike `for` / `else` because it breaks most debugging tools and makes it looks like an error is raised on the `else` even if it is raised on a particular line inside the `for`. 

I agree your version is fewer lines of code.

I'll think on it.

armi/reactor/blocks/block.py
john-science: Oh, sorry, no. Hold on. Please don't merge this change into the branch.

If you merge this change, you'll find one of the unit tests I wrote breaks.

The comment you delete is meaningful and correct. Do you see the problem I'm solving with the more complicated `max()` logic?

armi/reactor/tests/test_blocks.py
john-science: Okay, I added this here because a simpler version of the logic I wrote would fail if there were multiple components of the same time, with zero pins, at the same spatial locator.

So, I REALLY want to keep this here, because it cause a previous error I found if I didn't put enough catches into the new `getPinNum()` logic.

Is that okay?

armi/reactor/composites.py
john-science: I moved a copy of this method to `Core`. 

What I DIDN'T do was remove the method here. The reason is something I should clean up from this PR, just dev crumbs.

I tried to use this method instead of the `self.__getPinLikeCounts()` solution I came up with. But it turns out, putting all the children by spatial locator didn't really change the need to go through and make a new collection by type FIRST, and THEN by spatial locator. This little `_childrenByLocator` is just not what we need.

So I will go ahead and remove this method, sure. That is, I will remove it from Composite, but leave the new one on Core (because that one is used).

john-science: Okay, I removed this method and added a unit test to the new Core version.

armi/reactor/blocks/block.py
john-science: > I am personally unsure if there is a good reason for ARMI to even have a method such as getNumPins()

Well, again, this is a bug ticket that says "fix `getPinNums()`".  If ARMI has a bug, my short term goal is to fix that bug.

drewj-tp:
I have seen changes and pushes to try to add some testing related to pin locations in blocks with multiple pin types. Please pin me when this is ready again. And if it's ready now, great! Just let me know

armi/reactor/blocks/block.py
drewj-tp: > Actually, I am trying to empty out the comps list, so we don't keep an extra reference to these around. Because that is where most of the errors I found in testing came from.

I do not follow. An extra reference to `comps`? or to `compsByType`? Are you attempting to empty `compsByType` at the end of this iteration? What errors? Do those errors still exist now that the testing is more finished?

I ask that you either
1. Use a `for` loop, or
2. Add commentary as to why this approach is taken. It's not clear to this or future readers what problems may lie in wait if a `for` loop is used over this `while` loop.

armi/reactor/blocks/block.py
drewj-tp: > I agree your version is fewer lines of code.
> 
> I'll think on it.

Please consider the above solution, or adding comments as to the magic-ness of the `-9`.

armi/reactor/blocks/block.py
drewj-tp: Please update

armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,107 @@ def getSortedComponentsInsideOfComponent(self, component):
...
+        If you just want sorted components in this block, use ``sorted(self)``. This will never include any
+        ``DerivedShape`` objects. Since they have a derived area they don't have a well- defined dimension. For now we
+        just ignore them.
         """
         sortedComponents = sorted(self)
         componentIndex = sortedComponents.index(component)
         sortedComponents = sortedComponents[:componentIndex]
         return sortedComponents
 
+    def __getPinLikeCounts(self) -> list[Tuple[int, list[grids.IndexLocation]]]:
```
drewj-tp: Type annotation appears to be incorrect and inconsistent with the docstring. Is the second argument a list of components or index locations? Code indicates it will be a list of components

armi/reactor/blocks/block.py
```diff
@@ -967,16 +967,107 @@ def getSortedComponentsInsideOfComponent(self, component):
...
+        """Helper method to find components that look like pins, grouped by PIN_COMPONENT flags.
+
+        This method is not designed to be used by downstream, but is meant to help ``getPinLocations`` and
+        ``getNumPins`` share the same base logic.
+
+        Returns
+        -------
+        list(tuple(int, list(IndexLocation)))
+            A list of tuples: (number of pins, list of components at the same spatialLocator)
```
drewj-tp: Return annotation `list(IndexLocation)` is inconsistent with the line `list of components`

I believe it should be list of components based on my understanding of the code

armi/reactor/tests/test_blocks.py
```diff
@@ -1492,21 +1487,86 @@ def test_getNumPins(self):
...
         hexagon.setType("component", flags=Flags.SHIELD)
         emptyBlock.add(hexagon)
         self.assertEqual(emptyBlock.getNumPins(), 0)
+        self.assertEqual(len(emptyBlock.getPinLocations()), 0)
 
         pins = basicShapes.Circle("circle", "HT9", 100, 100, 1, 0, 8)
         pins.setType("component", flags=Flags.PLENUM)
         emptyBlock.add(pins)
         self.assertEqual(emptyBlock.getNumPins(), 8)
+        self.assertEqual(len(emptyBlock.getPinLocations()), 1)
```
drewj-tp: I think we want to enforce `Block.getNumPins() == len(Block.getPinLocations)`

This particular case has a pin locations of `1` despite a pin multiplicity of `8`. That seems like the component is ill-configured.

armi/reactor/tests/test_blocks.py
```diff
@@ -1492,21 +1487,86 @@ def test_getNumPins(self):
...
+        fuelPins.spatialLocator._j = 2
+        b.add(fuelPins)
+
+        pinClad1 = basicShapes.Circle("circle", "HT9", 100, 100, 1, 0, 8)
+        pinClad1.spatialLocator = fuelPins.spatialLocator
+        pinClad1.setType("component", flags=Flags.CLAD)
+        b.add(pinClad1)
+
+        self.assertEqual(b.getNumPins(), 8)
+        self.assertEqual(len(b.getPinLocations()), 1)
```
drewj-tp: I think we want to enforce Block.getNumPins() == len(Block.getPinLocations)

This particular case has a pin locations of 1 despite a pin multiplicity of 8. That seems like the component is ill-configured.

armi/reactor/tests/test_blocks.py
```diff
@@ -1492,21 +1487,86 @@ def test_getNumPins(self):
...
+
+        # add another clad, at the same position as the original clad, and ensure the pin count does not change
+        pinClad3 = basicShapes.Circle("circle", "HT9", 100, 100, 1, 0, 8)
+        pinClad3.spatialLocator = fuelPins.spatialLocator
+        pinClad3.setType("component", flags=Flags.CLAD)
+        b.add(pinClad3)
+
+        self.assertEqual(b.getNumPins(), 8)
+        self.assertEqual(len(b.getPinLocations()), 1)
```
drewj-tp: I think we want to enforce Block.getNumPins() == len(Block.getPinLocations)

This particular case has a pin locations of 1 despite a pin multiplicity of 8. That seems like the component is ill-configured.

armi/reactor/tests/test_cores.py
```diff
@@ -177,3 +177,8 @@ def test_getFirstBlock(self):
     def test_getFirstAssembly(self):
         a = self.core.getFirstAssembly()
         self.assertIsInstance(a, HexAssembly)
+
+    def test_genChildByLocationLookupTable(self):
+        self.core._genChildByLocationLookupTable()
+        numAssems = len(self.core.getAssemblies())
+        self.assertEqual(len(self.core.childrenByLocator), numAssems)
```
drewj-tp, line 185: Please consider extending this test to show not only is the number of entries identical, but we have all the assemblies in the locator.

armi/reactor/blocks/block.py
drewj-tp: Please update

armi/reactor/blocks/block.py
drewj-tp: >Do you see the problem I'm solving with the more complicated max() logic

No I do not. Please help me understand. 

My understanding is `compsByLocation` is a list of components that all share identical spatial locators. My expectation is that components that have identical spatial locators have an identical multiplicity. If this is not the case, then something is wrong. 

Looking at the unit tests, it seems like they have inconsistent multiplicity and spatial locators. I believe we expect the multiplicity to be
1. One if the component exists at one location,
2. Greater than one if a component exists at multiple locations.

The unit testing added does not satisfy these conditions.

armi/reactor/blocks/block.py
drewj-tp: Please add documentation to the user manual or blueprints guide or _somewhere_ not in an API docs that tells users what flags we consider when determining what a pin is.

drewj-tp (changes requested):
Thanks @john-science!

I believe we want to enforce the expectation that `Block.getNumPins() == len(Block.getPinLocations())`. This is not the case in some of the unit tests you've added. And, with that expectation, I believe some of my comments are applicable and beneficial.

The implementation looks correct. But I believe we can simplify some cases with the expectation `Block.getNumPins() == len(Block.getPinLocations())`. 

If someone was to make a reactor from blueprints, and specify a mult on a component that does not match with the lattice IDs, ARMI will already raise an error

https://github.com/terrapower/armi/blob/af5f9b75826425f6a0c1dfa8186058cde5e4ded9/armi/reactor/blueprints/blockBlueprint.py#L171-L176

Similarly, if the user has a mult and no lattice IDs, the spatial locator is assigned through the `autoCreateSpatialGrids` method

https://github.com/terrapower/armi/blob/af5f9b75826425f6a0c1dfa8186058cde5e4ded9/armi/reactor/blocks/hexBlock.py#L462

I say this to say I believe it is a reasonable expectation that `Block.getNumPins()` will equal `len(Block.getPinLocations())`

armi/reactor/tests/test_cores.py
john-science: Done

armi/reactor/blocks/block.py
john-science: done

armi/reactor/blocks/block.py
john-science: Done

armi/reactor/blocks/block.py
john-science: Done

armi/reactor/tests/test_blocks.py
john-science: Done

armi/reactor/tests/test_blocks.py
john-science: Done

armi/reactor/blocks/block.py
john-science: Okay, done.

armi/reactor/blocks/block.py
john-science: Sorry, this is left over from an older, dev version of the code.

Fixed.

armi/reactor/blocks/block.py
john-science: Sorry, this is left over from an older, dev version of the code.

Fixed.

armi/reactor/blocks/block.py
john-science: done

armi/reactor/blocks/block.py
john-science: > This implies that components that have identical spatial locators will have different multiplicity.

I suppose I see your point. But this is an extremely fast calculation "just go get the mult dimension". And my version is far less likely to break if someone decides to do something insane in the future, (as in many unit tests I have seen downstream, where all the components in a block are just set to the center of the block, regardless of their mult or physical reality).

Granted, the likelihood that people are calling `block.getNumPins()` or `block.getPinLocations()` for those test assemblies probably low. But it seems non-zero it might happen in the future.

I guess I won't die on this hill. But assuming that people will use ARMI in the way we expect has never worked out for me before.

armi/reactor/blocks/block.py
john-science: I have added a section pins to the user docs.

armi/reactor/blocks/block.py
john-science: We have added to the docs that if you don't add one of the 8 flags, ARMI won't know it's a pin.

It sounds like this makes you uneasy, but ARMI has to use _something_ firm to ID a pin.

I actually think we should go the other direction and be like, "Nope, if there is no `PIN` flag, then it's not a pin". It would greatly simplify logic in ARMI and dozens of downstream repos I have seen.

(We will see if we go that direction in the future.)

armi/reactor/tests/test_blocks.py
john-science: Done

armi/reactor/blocks/block.py
drewj-tp: > It sounds like this makes you uneasy, but ARMI has to use something firm to ID a pin

Nope, not uneasy that we need to have a way to find a pin. I just want it to be documented so people are less likely to be surprised. Thanks 😁

doc/user/spatial_block_data.rst
```diff
@@ -205,36 +204,72 @@ sites. Each item in this locator is one location on the underlying grid where we
...
+At the ``Block`` level, there are a few important pin-related methods to know:
+
+- ``Block.getNumPins()`` - return the number of pins in a given Block
+- ``Block.getPinLocations()`` - return the locations of each pin in a given Block
+- ``Block.getPinCoordinates()`` - return the coordinates of the center of each pin in a given Block
+
+These method all share the same underlying logic for identifying pins within a Block, which should help them maintain a
+a high level of agreement.
+
+
```
drewj-tp: This is the body of the real change to this file, other than some automated formatting that your IDE may have enabled.

doc/user/spatial_block_data.rst
```diff
@@ -205,36 +204,72 @@ sites. Each item in this locator is one location on the underlying grid where we
...
+collected into groups inside a physical object called an assembly. ARMI is mostly designed for modeling such reactors.
+
+Finding Pins in an ARMI Model
+-----------------------------
+
+Originally, ARMI was packaged with a ``Flags`` system that include the flag ``PIN``. The assumption was that, in
+blueprints, all pins in an Assembly would be marked with the ``PIN`` flag. Hypothetically, this would yield the most
+optimally simple path to identifying all the pins in a Core, Assembly, or Block. However, in pratice this was not done,
+so the tooling in ARMI has necessarily become more complex to help find pins in assemblies.
+
```
drewj-tp: I understand the frustration about why we don't use `Flags.PIN`. But I don't think this behind-the-scenes commentary is useful for readers.

Please consider removing the last two sentences of the paragraph and joining it with the next paragraph.

If we want to emphasize that users should use `Flags.PIN` in addition to e.g., `Flags.CLAD`, that would be a welcome addition.

doc/user/spatial_block_data.rst
```diff
@@ -205,36 +204,72 @@ sites. Each item in this locator is one location on the underlying grid where we
...
+But in ARMI a Component can have a multiplicity. So if, inside some Block, there is a single circular Component marked
+with the flags ``CLAD`` ``FUEL`` and a multiplicity of 217, then there are 217 pins in that Block. The exact position of
+those pins should be defined on a sub-Block-level lattice.
+
+At the ``Block`` level, there are a few important pin-related methods to know:
+
+- ``Block.getNumPins()`` - return the number of pins in a given Block
+- ``Block.getPinLocations()`` - return the locations of each pin in a given Block
+- ``Block.getPinCoordinates()`` - return the coordinates of the center of each pin in a given Block
```
drewj-tp: Soft, non-blocking recommendation to link to these methods with e.g., 
```rst
:meth:`armi.reactor.blocks.Block.getPinCoordinates`
```

drewj-tp (changes requested):
Thank you for the addition of the docs 😁 

For some reason, GH won't let me make in-line suggestions to the docs 🤷

doc/user/spatial_block_data.rst
john-science: done

doc/user/spatial_block_data.rst
john-science: Updated

drewj-tp (approved):
🚀 🎉

PR #2630, opened 2026-08-10
Improving warnings on negative DPA calculations

### LINKED ISSUE #2629: runLog message for negative DPA rate should be improved
When negative DPA rates are calculated, there is a warning message printed in the runLog:
https://github.com/terrapower/armi/blob/7d2b3b26296558d671a529b9180b9af9226838de/armi/physics/neutronics/globalFlux/globalFluxInterface.py#L901-L906

That message uses `single=True` and specifies a `label` with the message. That means that the message is only ever printed once in a run, even if negative DPA is calculated for every block in the core. This is not great, and does not give the user an appropriate sense for the issue.

Additionally, it would be very helpful if the location of the negative DPA rate was provided in the warning message. As of now, the user has to dig down in the neutronics kernel output file or infer the location from the database.

john-science:

@keckler raised the concern, and @andrewryh agreed, that the warnings for negative DPA calculations were not helpful. They request that the actual `Block` in question be listed in the warning, and we stop trying to silence these warnings if they reappear.

To make this happen, I had to optionally pass in the `Block` in question, for logging. (I can't make this mandatory, because of downstream users.)

Also, I added a bunch of testing to the method `computeDpaRate().

close #2629 

## SCR Information

Change Type: trivial

One-Sentence Rationale: Our users wanted better fidelity on these warnings.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/globalFlux/globalFluxInterface.py
+++ b/armi/physics/neutronics/globalFlux/globalFluxInterface.py
@@ -832,7 +832,7 @@ class GlobalFluxResultMapper(interfaces.OutputReader):
                 a.p.kInf = totalSrc / totalAbs  # assembly average k-inf.
 
 
-def computeDpaRate(mgFlux, dpaXs):
+def computeDpaRate(mgFlux, dpaXs, block=None):
     r"""
     Compute the DPA rate incurred by exposure of a certain flux spectrum.
 
@@ -840,8 +840,8 @@ def computeDpaRate(mgFlux, dpaXs):
         :id: I_ARMI_FLUX_DPA
         :implements: R_ARMI_FLUX_DPA
 
-        This method calculates DPA rates using the inputted multigroup flux and DPA cross sections.
-        Displacements calculated by displacement cross-section:
+        Calculate DPA rates using a provided multi-group flux and DPA cross sections. Displacements are calculated by
+        displacement cross-section:
 
         .. math::
             :nowrap:
@@ -865,48 +865,49 @@ def computeDpaRate(mgFlux, dpaXs):
 
             \frac{\text{dpa}}{s}  = \frac{\phi N \sigma}{N} = \phi * \sigma
 
-        the number density of the structural material cancels out. It's in the macroscopic
-        cross-section and in the original number of atoms.
+        the number density of the structural material cancels out. It is in the macroscopic cross-section and in the
+        original number of atoms.
 
     Parameters
     ----------
     mgFlux : list
-        multigroup neutron flux in #/cm^2/s
-
+        multi-group neutron flux in #/cm^2/s
     dpaXs : list
         DPA cross section in barns to convolute with flux to determine DPA rate
+    block : Block
+        Optional parameter, used purely for logging.
 
     Returns
     -------
     dpaPerSecond : float
-        The dpa/s in this material due to this flux
+        The DPA/s in this material due to this flux
 
     Raises
     ------
     RuntimeError
        Negative dpa rate.
     """
-    displacements = 0.0
     if len(mgFlux) != len(dpaXs):
         runLog.warning(
-            "Multigroup flux of length {} is incompatible with dpa cross section of length {};"
-            "dpa rate will be set do 0.0".format(len(mgFlux), len(dpaXs)),
+            f"Multigroup flux of length {len(mgFlux)} is incompatible with DPA cross section of length {len(dpaXs)}; "
+            "DPA rate will be set do 0.0",
             single=True,
         )
-        return displacements
+        return 0.0
+
+    displacements = 0.0
     for flux, barns in zip(mgFlux, dpaXs):
         displacements += flux * barns
-    dpaPerSecond = displacements * units.CM2_PER_BARN
 
+    dpaPerSecond = displacements * units.CM2_PER_BARN
     if dpaPerSecond < 0:
-        runLog.warning(
-            "Negative DPA rate calculated at {}".format(dpaPerSecond),
-            single=True,
-            label="negativeDpaPerSecond",
-        )
+        forBlock = f" for block {block}" if block else ""
+        runLog.warning(f"Negative DPA rate calculated to be {dpaPerSecond}{forBlock}, setting to zero.")
         # ensure physical meaning of dpaPerSecond, it is likely just slightly negative
         if dpaPerSecond < -1.0e-10:
-            raise RuntimeError("Calculated DPA rate is substantially negative at {}".format(dpaPerSecond))
+            msg = f"Calculated DPA rate is substantially negative at {dpaPerSecond}{forBlock}."
+            runLog.error(msg)
+            raise RuntimeError(msg)
         dpaPerSecond = 0.0
 
     return dpaPerSecond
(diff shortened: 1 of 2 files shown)


john-science:
@keckler Andrei offered to review this PR. But here it is, in case you have strong opinions.

keckler:
@onufer might have some thoughts too

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -865,48 +865,49 @@ def computeDpaRate(mgFlux, dpaXs):
...
 
+    dpaPerSecond = displacements * units.CM2_PER_BARN
     if dpaPerSecond < 0:
-        runLog.warning(
-            "Negative DPA rate calculated at {}".format(dpaPerSecond),
-            single=True,
-            label="negativeDpaPerSecond",
-        )
+        forBlock = f" for block {block}" if block else ""
+        runLog.warning(f"Negative DPA rate calculated to be {dpaPerSecond}{forBlock}, setting to zero.")
```
andrewryh, line 905: This message is exactly what I was looking for.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -865,48 +865,49 @@ def computeDpaRate(mgFlux, dpaXs):
...
         # ensure physical meaning of dpaPerSecond, it is likely just slightly negative
         if dpaPerSecond < -1.0e-10:
-            raise RuntimeError("Calculated DPA rate is substantially negative at {}".format(dpaPerSecond))
+            msg = f"Calculated DPA rate is substantially negative at {dpaPerSecond}{forBlock}."
```
andrewryh, line 908: I think here we also should inform user that DPA is being set to zero, like in message above.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -865,48 +865,49 @@ def computeDpaRate(mgFlux, dpaXs):
...
     if dpaPerSecond < 0:
-        runLog.warning(
-            "Negative DPA rate calculated at {}".format(dpaPerSecond),
-            single=True,
-            label="negativeDpaPerSecond",
-        )
+        forBlock = f" for block {block}" if block else ""
+        runLog.warning(f"Negative DPA rate calculated to be {dpaPerSecond}{forBlock}, setting to zero.")
         # ensure physical meaning of dpaPerSecond, it is likely just slightly negative
         if dpaPerSecond < -1.0e-10:
```
andrewryh, line 907: I think this condition doesn't work properly because we've got:
`[warn] Negative DPA rate calculated at -7.809904265512126e-12`

-7.8e-12 < -1.0e-10 but we don't see
"Calculated DPA rate is substantially negative at ...." message

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -832,16 +832,16 @@ def _updateAssemblyLevelParams(self):
...
     r"""
     Compute the DPA rate incurred by exposure of a certain flux spectrum.
 
     .. impl:: Compute DPA rates.
         :id: I_ARMI_FLUX_DPA
         :implements: R_ARMI_FLUX_DPA
 
-        This method calculates DPA rates using the inputted multigroup flux and DPA cross sections.
-        Displacements calculated by displacement cross-section:
+        Calculate DPA rates using a provided multi-group flux and DPA cross sections. Displacements are calculated by
```
andrewryh: `cross-section` for consistency

armi/physics/neutronics/globalFlux/globalFluxInterface.py
```diff
@@ -832,16 +832,16 @@ def _updateAssemblyLevelParams(self):
                 a.p.kInf = totalSrc / totalAbs  # assembly average k-inf.
 
 
-def computeDpaRate(mgFlux, dpaXs):
+def computeDpaRate(mgFlux, dpaXs, block=None):
```
andrewryh, line 835: @john-science Adding this 3rd positional input parameter to the function would require change all other codes, that use ARMI, accordingly.
Otherwise we will see {None} at every warning.

I guess it is the only way, right?

andrewryh (changes requested):
Looks good but see few in-line comments

armi/physics/neutronics/globalFlux/globalFluxInterface.py
john-science: You wrote above:

> -7.8e-12 < -1.0e-10 but we don't see

But obviously, this is not true: `-7.8e-12 < -1.0e-10`.

`-7.8e-12` is greater than  `-1.0e-10`

john-science: Do you perhaps want me to put an absolute value around both of these numbers?

Or switch `>` to `<`?

armi/physics/neutronics/globalFlux/globalFluxInterface.py
john-science: > Otherwise we will see {None} at every warning.

Not exactly. If you don't pass in the `Block`, then I have the log statement set up to print correctly, just not mentioning the block at all. You won't see an ugly `{None}` in the log statement. Promise.

> I guess it is the only way, right?

Yup. This function `computeDpaRate()` is not part of a class and, before this PR, doesn't know ANYTHING about the Block. So we _have_ to pass the `block` in, if we want to potentially use it in the log message.

We can make downstream changes, to improve the print-outs. But this version of the change in my PR at least doesn't hard break anything downstream.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
john-science: Actually, it is NOT being set to zero here.

And error is being raised and the run is being killed.

armi/physics/neutronics/globalFlux/globalFluxInterface.py
john-science: Done!

armi/physics/neutronics/globalFlux/globalFluxInterface.py
andrewryh: > You wrote above:
> 
> > -7.8e-12 < -1.0e-10 but we don't see
> 
> But obviously, this is not true: `-7.8e-12 < -1.0e-10`.
> 
> `-7.8e-12` is greater than `-1.0e-10`

You are right, it is not. I didn't think in negative direction of the axis

andrewryh: > Do you perhaps want me to put an absolute value around both of these numbers?
> 
> Or switch `>` to `<`?

No, I think it is good as-is.
My bad. Can't do math

armi/physics/neutronics/globalFlux/globalFluxInterface.py
andrewryh: It makes perfect sense, thank you.
We will need to do minor changes in downstream codes.

andrewryh (approved):
All the comments resolved.
Good to merge

PR #2636, opened 2026-08-12
Making Composite.changeNDensByFactor safe for Components

### LINKED ISSUE #2628: RuntimeError waiting to happen in ARMIObject.changeNDensByFactor
In #1990, I added

https://github.com/terrapower/armi/blob/7d2b3b26296558d671a529b9180b9af9226838de/armi/reactor/composites.py#L1511-L1513

to `ArmiObject.changeNDensByFactor()`. This is a `RuntimeError` waiting to happen because `pinNDens` is only defined on the Component level.

https://github.com/terrapower/armi/blob/7d2b3b26296558d671a529b9180b9af9226838de/armi/reactor/components/componentParameters.py#L89-L98

In all of ARMI, with the exception of [one spot](https://github.com/terrapower/armi/blob/7d2b3b26296558d671a529b9180b9af9226838de/armi/reactor/assemblies.py#L757) (which is in a workflow we don't really use anymore), `changeNDensByFactor` is only called on the Component level. It might be a good idea to move that method to the component level.

john-science:

The method `Composite.changeNDensByFactor()` is also called from Components, which do not have the same set of Parameters as Composites.

So, there were two options available to me:

1. Copy the method over to Components, and alter it, or
2. Make the method safe from within.

I dislike code duplication, and if we did (1) we would be heavily relying on docstrings in each method to point to the other method, if there were updates made to one of the methods. So (2) seems like a safer option, not only does it result in there being less code, but the method is not intrinsically safe.

close #2628

## SCR Information

Change Type: trivial

One-Sentence Rationale: While this problem has not occurred yet, it is very easy to safeguard against.

---

## Checklist

### DIFF
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -1505,11 +1505,13 @@ class ArmiObject(metaclass=CompositeModelType):
         """Change the number density of all nuclides within the object by a multiplicative factor."""
         densitiesScaled = {nuc: val * factor for nuc, val in self.getNumberDensities().items()}
         self.setNumberDensities(densitiesScaled)
+
         # Update detailedNDens
-        if self.p.detailedNDens is not None:
+        if self.p.get("detailedNDens", None) is not None:
             self.p.detailedNDens *= factor
+
         # Update pinNDens
-        if self.p.pinNDens is not None:
+        if self.p.get("pinNDens", None) is not None:
             self.p.pinNDens *= factor
 
     def clearNumberDensities(self):
--- a/armi/reactor/composites.py
+++ b/armi/reactor/composites.py
@@ -1506,11 +1506,11 @@ class ArmiObject(metaclass=CompositeModelType):
         densitiesScaled = {nuc: val * factor for nuc, val in self.getNumberDensities().items()}
         self.setNumberDensities(densitiesScaled)
 
-        # Update detailedNDens
+        # Update detailedNDens if it exists (Components only)
         if self.p.get("detailedNDens", None) is not None:
             self.p.detailedNDens *= factor
 
-        # Update pinNDens
+        # Update pinNDens if it exists (Components only)
         if self.p.get("pinNDens", None) is not None:
             self.p.pinNDens *= factor
 


armi/reactor/composites.py
```diff
@@ -1505,11 +1505,13 @@ def changeNDensByFactor(self, factor):
         """Change the number density of all nuclides within the object by a multiplicative factor."""
         densitiesScaled = {nuc: val * factor for nuc, val in self.getNumberDensities().items()}
         self.setNumberDensities(densitiesScaled)
+
         # Update detailedNDens
```
yardasol:
```suggestion
        # Update detailedNDens if it exists (Components only)
```

armi/reactor/composites.py
```diff
@@ -1505,11 +1505,13 @@ def changeNDensByFactor(self, factor):
...
-        if self.p.detailedNDens is not None:
+        if self.p.get("detailedNDens", None) is not None:
             self.p.detailedNDens *= factor
+
         # Update pinNDens
```
yardasol:
```suggestion
        # Update pinNDens if it exists (Components only)
```

yardasol (commented):
Great work here @john-science! Just some comments thinking out loud here, lmk your thoughts.

armi/reactor/composites.py
```diff
@@ -1505,11 +1505,13 @@ def changeNDensByFactor(self, factor):
...
+
+        # Update detailedNDens if it exists (Components only)
+        if self.p.get("detailedNDens", None) is not None:
             self.p.detailedNDens *= factor
-        # Update pinNDens
-        if self.p.pinNDens is not None:
+
+        # Update pinNDens if it exists (Components only)
+        if self.p.get("pinNDens", None) is not None:
             self.p.pinNDens *= factor
```
drewj-tp, line 1515: `Components` override `changeNDensByFactor` already

https://github.com/terrapower/armi/blob/3579a20818190b0269b26bfd9454b982246f8d48/armi/reactor/components/component.py#L819-L831

and it handles scaling these params. 

Would we rather rely on the composite tree structure and do something like
```python
for child in self:
     child.changeNDensByFactor(factor)
```

We also tried to get rid of methods that attempt to change the number density higher up the composite tree than components
- https://github.com/terrapower/armi/issues/1374
- https://github.com/terrapower/armi/pull/2122

drewj-tp: I see the justification for keeping it around in https://github.com/terrapower/armi/blob/7d2b3b26296558d671a529b9180b9af9226838de/armi/reactor/assemblies.py#L754-L757

But that's also compatible with delegating to the child composites `changeNDensByFactor` method

yardasol (approved):
Good work @john-science

PR #2637, opened 2026-08-12
Unpinning PyPDF

### LINKED ISSUE #2633: Unpin or Replace pypdf

john-science:

In this PR I am unpinning the docs-only dependency `pypdf`. The [discussion](https://github.com/terrapower/armi/issues/2633) here leads me to think this is probably a safe option. The `pypdf` team is probably falsely marking every one of their releases as a "security" update, and that's the source of the problem. And that's a real problem, but it's out of our hands.

If, in the future, we have problems with `pypdf` making breaking changes that wreck our docs, then I think we will be left with no other option that to find a replacement library. But, until that happens, this seems like a fine solution.

close #2633

## SCR Information

Change Type: trivial

One-Sentence Rationale: We want to stop the constant stream of Dependabot PRs due to pypdf.

---

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ docs = [
     "nbsphinx==0.9.6", # Parses Jupyter notebooks
     "pandoc", # Must be in the path (to convert file formats)
     "pylint", # Generates UML diagrams
-    "pypdf==6.15.0", # Generating a single PDF file for the Sphinx documentation
+    "pypdf", # Generating a single PDF file for the Sphinx documentation
     "setuptools",  # needed for conf.py tooling
     "sphinx-gallery==0.13.0", # Builds an HTML version of a Python script and puts it into a gallery
     "sphinx-needs==4.2.0", # Requirements traceability matrices for QA


yardasol (approved):
Sad we have to unpin this, but it's better than constantly updating the pinned version.

opotowsky (approved):
We were talking offline about other PDF generation approaches, so maybe this library CAN be replaced with another approach. However, no need to do that work unless `pypdf` becomes too annoying to use.

PR #2642, opened 2026-08-25
Removing inefficiencies in Assembly initialization

mgjarrett:

Remove some inefficiencies in assembly initialization, to speed up reactor creation from blueprints and  uniform mesh conversion. Three inefficiencies are addressed:

1. When performing axial expansion on an `Assembly`, the `RedistributeMass` class would repeatedly call the properties `toCompVolume`, `fromCompVolume`, and `newVolume` on the same components. Each time, a call would be made to `Component.getArea()`, which can be expensive. This routine is changed to call each property once, store the value, and use the stored value in the loop over nuclides. With this update, I observed a speedup of **approximately 20x** (107 s down to 5.4 s) for the axial expansion.
2. When performing axial expansion, the axial linkages need to be determined for each `Component`. The logic was refactored to perform lower-cost checks first (Component multiplicity), which allows the higher-cost check (all spatial locator indices are identical) to be skipped in some cases. In the test case I ran, I observed a speedup of 2.3x (38 s down to 16.3 s).
3. Reduce number of calls to `reestablishBlockOrder` and `calculateZCoords` when building an assembly from the number of blocks down to 1. Each call iterates over all blocks in the assembly, so this reduces the iteration from `O(n^2)` to `O(n)`. In my testing, I observed a drop in time spent adding blocks to assemblies from 7.9 s to 0.4 s -- about 20x speedup, albeit on a small piece of the pie.

## SCR Information

Change Type: features

One-Sentence Rationale: This speeds up initialization of a reactor object and other operations that involve building assemblies, such as uniform mesh conversion.

One-line Impact on Requirements: New implementation tag `I_ARMI_ASSEM_BLOCKS_2` for requirement `R_ARMI_ASSEM_BLOCKS`.

---

## Checklist

### DIFF
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -209,12 +209,9 @@ class Assembly(composites.Composite):
             :id: I_ARMI_ASSEM_BLOCKS_2
             :implements: R_ARMI_ASSEM_BLOCKS
 
-            Builds an assembly from a stack of Blocks. The spatialLocator
-            of the Assembly is updated. In ``reestablishBlockOrder``, the
-            Assembly spatialGrid is reinitialized and Block-wise
-            spatialLocator and name objects are updated. The axial mesh
-            and other Block geometry parameters are updated in
-            ``calculateZCoords``.
+            Builds an Assembly from a stack of Blocks. The spatialLocator of the Assembly is updated. In
+            ``reestablishBlockOrder``, the Assembly spatialGrid is reinitialized and Block-wise spatialLocator and name
+            objects are updated. The axial mesh and other Block geometry parameters are updated in ``calculateZCoords``.
         """
         # replace grid with one that has the right number of locations
         self.removeAll()
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -202,7 +202,20 @@ class Assembly(composites.Composite):
         self.calculateZCoords()
 
     def setBlockStack(self, blockStack: list[blocks.Block]):
-        """Re-initialize this assembly based on a provided stack of blocks."""
+        """
+        Re-initialize this assembly based on a provided stack of blocks.
+
+        .. impl:: Assemblies are made up of type Block.
+            :id: I_ARMI_ASSEM_BLOCKS_2
+            :implements: R_ARMI_ASSEM_BLOCKS
+
+            Builds an assembly from a stack of Blocks. The spatialLocator
+            of the Assembly is updated. In ``reestablishBlockOrder``, the
+            Assembly spatialGrid is reinitialized and Block-wise
+            spatialLocator and name objects are updated. The axial mesh
+            and other Block geometry parameters are updated in
+            ``calculateZCoords``.
+        """
         # replace grid with one that has the right number of locations
         self.removeAll()
         self.spatialGrid = grids.AxialGrid.fromNCells(len(blockStack))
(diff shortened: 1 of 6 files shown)


armi/reactor/assemblies.py
```diff
@@ -201,6 +201,34 @@ def add(self, obj: blocks.Block):
...
+        .. impl:: Assemblies are made up of type Block.
+            :id: I_ARMI_ASSEM_BLOCKS_2
+            :implements: R_ARMI_ASSEM_BLOCKS
+
+            Builds an assembly from a stack of Blocks. The spatialLocator
+            of the Assembly is updated. In ``reestablishBlockOrder``, the
+            Assembly spatialGrid is reinitialized and Block-wise
+            spatialLocator and name objects are updated. The axial mesh
+            and other Block geometry parameters are updated in
+            ``calculateZCoords``.
```
john-science:
```suggestion
            Builds an Assembly from a stack of Blocks. The spatialLocator of the Assembly is updated. In
            ``reestablishBlockOrder``, the Assembly spatialGrid is reinitialized and Block-wise spatialLocator and name
            objects are updated. The axial mesh and other Block geometry parameters are updated in ``calculateZCoords``.
```

john-science: Unimportant. 

Lines can be 120 characters long. So I made the lines longer.

armi/reactor/converters/axialExpansionChanger/redistributeMass.py
```diff
@@ -131,10 +131,13 @@ def setNewToCompNDens(self):
         # calculate the mass of each nuclide and then the ndens for the new mass
         newNDens: dict[str, float] = {}
         nucs = self._getAllNucs(self.toComp.getNuclides(), self.fromComp.getNuclides())
+        toVolume = self.toCompVolume
```
john-science, line 134: This seems like a safe win to me. Nice!

armi/reactor/converters/axialExpansionChanger/assemblyAxialLinkage.py
```diff
@@ -60,29 +60,30 @@ def areAxiallyLinked(componentA: Component, componentB: Component) -> bool:
...
+                    f"Components {componentA} and {componentB} are UnshapedComponents "
+                    "and do not have 'getCircleInnerDiameter' or getBoundingCircleOuterDiameter methods; "
+                    "nor is it physical to do so. Instead of crashing and raising an error, "
+                    "they are going to be assumed to not be linked.",
+                    single=True,
+                )
+            elif isinstance(componentA.spatialLocator, MultiIndexLocation) and isinstance(
+                componentB.spatialLocator, MultiIndexLocation
+            ):
+                ## Case 2
```
john-science, line 79: Okay, so in this version of the logic both components have to have the same multiplicity to reach "case 2". But in the old version of the logic, they could have had different multiplicity and still hit this case.

So there is technically a difference here.

But your thesis here is that, in practice, these components already had the same multiplicity?

mgjarrett: > But your thesis here is that, in practice, these components already had the same multiplicity?

Yes, I think they have to have the same multiplicity to satisfy case 2. When a component has a `spatialLocator` that is a `MultiIndexLocation`, none of the indices should be duplicated. Thus, in order for the set of `componentA.spatialLocator.indices` and `componentB.spatialLocator.indices` to be equivalent, the two sequences must have the same length, and thus the components should have the same multiplicity.

I can't think of a way that this would not be true. I think the purpose of using `set` here is to handle a case where the two components have the same indices but they're in a different order.

john-science (approved):
A simple, effect optimization PR?

And from out of nowhere, with no meetings to discuss it?

You're a hero!

PR #2644, opened 2026-08-27
Fixing Materia.getProperty for matProps properties

### LINKED ISSUE #2643: Material.getProperty not correctly passing Temps in Celcius to matProps
ARMI's `Material.getProperty()` takes in temperature in Celsius or Kelvin:

https://github.com/terrapower/armi/blob/ce01b794b0f128303adcf6faf072cc1e89e34c59/armi/materials/material.py#L517

And does the same thing all the other `Material` property functions do, which is convert things to Kelvin immediately:

https://github.com/terrapower/armi/blob/ce01b794b0f128303adcf6faf072cc1e89e34c59/armi/materials/material.py#L525

But there is an edge case where the material is called directly from `matProps`, and thus should be:

1. Converted to Celcius
2. Called `T`, because `matProps` doesn't care about other units.

This:

https://github.com/terrapower/armi/blob/ce01b794b0f128303adcf6faf072cc1e89e34c59/armi/materials/material.py#L533-L535

should become this:

```python
            Tc = getTk(Tc, Tk)
            val = getattr(self, propName)(T=Tc, **kwargs)
            # cache only one value for each property. Prevents unbounded cache explosion.
            self._setCache(propName, (Tk, val))
```

john-science:

The method `armi.materials.Material.getProperty()` has not been used much. And it was also not unit tested at all. 

And hence there was a bug in the method. If someone asked for a material property from the matProps side of things, this method would fail. (Due to a temperature conversion issue.)

This little PR fixes that issue.

close #2643

## SCR Information

Change Type: fixes

One-Sentence Rationale: Material.getProperty is not heavily used, but if it exists it should work correctly.

---

## Checklist

### DIFF
--- a/armi/materials/material.py
+++ b/armi/materials/material.py
@@ -529,8 +529,16 @@ class Material(MatPropsMaterial):
             # only use cached value if the temperature at which it is cached is the same.
             return cached[1]
         else:
-            # go look it up from material properties.
-            val = getattr(self, propName)(Tk=Tk, **kwargs)
+            # go look it up from material properties
+            try:
+                # If this is an armi.materials.Material property.
+                val = getattr(self, propName)(Tk=Tk, **kwargs)
+            except KeyError:
+                # This is an armi.matProps.MatPropsMaterial property.
+                if Tc is None:
+                    Tc = getTc(Tc, Tk)
+                val = getattr(self, propName)(T=Tc, **kwargs)
+
             # cache only one value for each property. Prevents unbounded cache explosion.
             self._setCache(propName, (Tk, val))
             return val
--- a/armi/materials/tests/test_materials.py
+++ b/armi/materials/tests/test_materials.py
@@ -89,12 +89,24 @@ class AbstractMaterialTest:
         self.assertEqual(val, "Noether")
 
 
-class TestMaterialConstruction(unittest.TestCase):
+class TestMaterialBasics(unittest.TestCase):
     def test_matInit(self):
         """Make sure all materials can be instantiated without error."""
         for matClass in materials.iterAllMaterialClassesInNamespace(materials):
             m = matClass()
             self.assertIsInstance(m, materials.Material)
+
+    def test_getProperty(self):
+        mat = materials.Cs()
+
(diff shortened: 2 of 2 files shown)


armi/materials/material.py
```diff
@@ -529,8 +529,16 @@ def getProperty(self, propName: str, Tk: float = None, Tc: float = None, **kwarg
...
-            val = getattr(self, propName)(Tk=Tk, **kwargs)
+            # go look it up from material properties
+            try:
+                # If this is an armi.materials.Material property.
+                val = getattr(self, propName)(Tk=Tk, **kwargs)
+            except KeyError:
+                # This is an armi.matProps.MatPropsMaterial property.
+                if Tc is None:
+                    Tc = getTc(Tc, Tk)
+                val = getattr(self, propName)(T=Tc, **kwargs)
```
opotowsky, line 540:
```suggestion
                val = getattr(self, propName)(T=Tk, **kwargs)
```
Don't we want Tk only for mat props properties? The only difference between the two is the arg for ARMI materials is `Tc` or `Tk` and the arg for mat props is `T`?

john-science: Actually, for whatever reason, matProps is only C:

https://github.com/terrapower/armi/blob/ce01b794b0f128303adcf6faf072cc1e89e34c59/armi/materials/material.py#L485

https://github.com/terrapower/armi/blob/ce01b794b0f128303adcf6faf072cc1e89e34c59/armi/materials/material.py#L500

https://github.com/terrapower/armi/blob/ce01b794b0f128303adcf6faf072cc1e89e34c59/armi/materials/material.py#L505

opotowsky: OH! I think we said yesterday it was K and I just didn't double check. well la dee da Celsius it is

opotowsky (approved):

armi/materials/material.py
opotowsky: shoot @mgjarrett then we told you wrong yesterday. I'm so sorry, you want to pass in temp in C and not K for the workaround

john-science: Sorry!

PR #2646, opened 2026-08-27
Ignoring negative volume components in mass-weighted averaging

mgjarrett:

**Change**: Ignore any `Component` with negative area when computing a mass-weighted average temperature and composition for a set of components.

**Background**: When building a representative `Block` in the `CrossSectionGroupManager` for a cross section type with the option `averageByComponent: true`, it is necessary to compute an average temperature and composition across all blocks in the `BlockCollection` for each `Component` in the representative `Block`. In normal circumstances, a mass-weighted average is used. However, there are some situations where a `Component` can have negative area/volume and thus negative mass. If some components in the `BlockCollection` have positive mass and others have negative mass, the mass-weighted average temperature of the components can be extremely different than a non-weighted average. In some cases, the resulting temperature could be outside the bounds of defined material properties, which could cause a run to crash.

## SCR Information

Change Type: features

One-Sentence Rationale: This addresses a corner case bug that could cause ARMI apps to crash when there are negative-volume components present.

---

## Checklist

### DIFF
--- a/armi/physics/neutronics/crossSectionGroupManager.py
+++ b/armi/physics/neutronics/crossSectionGroupManager.py
@@ -451,8 +451,9 @@ class AverageBlockCollection(BlockCollection):
         weights = np.array([self.getWeight(b) / b.getHeight() for b in blocks])
         weights /= weights.sum()  # normalize by total weight
         components = [sorted(b.getComponents())[compIndex] for b in blocks]
+        masses = [c.getMass() for c in components]
         weightedAvgComponentMass = sum(
-            w * c.getMass() if c.getMass() > 0.0 else 0.0 for w, c in zip(weights, components)
+            w * mass if mass > 0.0 else 0.0 for w, mass in zip(weights, masses)
         )
         if weightedAvgComponentMass == 0.0:
             # if there is no component mass (e.g., gap), do a regular average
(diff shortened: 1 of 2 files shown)


armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -451,13 +451,18 @@ def _getAverageComponentTemperature(self, compIndex):
         weights = np.array([self.getWeight(b) / b.getHeight() for b in blocks])
         weights /= weights.sum()  # normalize by total weight
         components = [sorted(b.getComponents())[compIndex] for b in blocks]
-        weightedAvgComponentMass = sum(w * c.getMass() for w, c in zip(weights, components))
+        weightedAvgComponentMass = sum(
+            w * c.getMass() if c.getMass() > 0.0 else 0.0 for w, c in zip(weights, components)
+        )
```
john-science:
```suggestion
        weightedAvgComponentMass = [w * c.getMass() for w, c in zip(weights, components)]
        weightedAvgComponentMass = sum(m for m in weightedAvgComponentMass  if m > 0)
```

john-science: This is optional.

But if `.getMass()` here is the computationally most costly thing in this loop, why do it twice?

(Same thing below, if this is interesting to you.)

john-science (approved):
One optional thought.

I imagine it'll be a negligible performance difference, so it's up to you.

armi/physics/neutronics/crossSectionGroupManager.py
mgjarrett: Good suggestion, I implemented it

PR #2647, opened 2026-08-28
Updating Assembly detailed number density when Core pitch changes

drewj-tp:

When the `Core` pitch changes via `Core.setPitchUniform`, the pitch of each `Block` in the `Core` changes to match the uniform pitch. For hex assemblies / hex blocks, this is functionally changes the outer flat-to-flat pitch of the largest hex `Component` in each `Block`. As such, the `Block` and `Assembly` volume can change as a result of this operation.

Changing an `Assembly` volume impacts the `Assembly` detailed number density `Parameter`. We'd like this to be consistent with the volume-weighted average of the child components. See #2613

This PR does two related things.

1. Introduce `Assembly.assignConsistentDetailedNDens` to set the `a.p.detailedNDens` as this volume-weighted average, and
2. Use `Assembly.assignConsistentDetailedNDens` when `Core.setPitchUniform` is called.

## SCR Information

Change Type: fixes

One-Sentence Rationale: Keep assembly detailed number density consistent through changing core pitch

---

## Checklist

### DIFF
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -1250,7 +1250,7 @@ class Assembly(composites.Composite):
 
             N = \frac{\sum_c\left(N_c * V_c\right)}{\sum_cV_c}
 
-        If no detailed number densities are found, sets parameter to ``None``
+        If no detailed number densities are found, sets parameter to ``None``.
         """
         totalVolume = 0.0
         onAssembly = None
@@ -1274,7 +1274,7 @@ class Assembly(composites.Composite):
         Parameters
         ----------
         pitchInCm
-            Requested pitch of this assembly. Means different things for
+           Requested pitch of this assembly. Means different things for
            different assemblies (e.g., hex vs. cartesian). But the child blocks
            also know what to do with this.
 
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -1268,6 +1268,30 @@ class Assembly(composites.Composite):
         else:
             self.p.detailedNDens = None
 
+    def setPitch(self, pitchInCm: float):
+        """Adjust the assembly to alight with a uniform pitch.
+
+        Parameters
+        ----------
+        pitchInCm
+            Requested pitch of this assembly. Means different things for
+           different assemblies (e.g., hex vs. cartesian). But the child blocks
+           also know what to do with this.
+
+        See Also
+        --------
+        :meth:`armi.reactor.cores.Core.setPitchUniform`
+        :meth:`armi.reactor.blocks.Block.setPitch`
+
+        :meth:`assignConsistentDetailedNDens` will be called if ``detailedNDens`` is set
+        because the child volumes may have been adjusted
+        """
+        for block in self:
+            block.setPitch(pitchInCm)
+
+        if self.p.detailedNDens is not None:
+            self.assignConsistentDetailedNDens()
+
 
 class HexAssembly(Assembly):
     """An assembly that is hexagonal in cross-section."""
--- a/armi/reactor/assemblies.py
+++ b/armi/reactor/assemblies.py
@@ -1241,6 +1241,33 @@ class Assembly(composites.Composite):
             if b.spatialGrid is not None:
                 b.assignPinIndices()
 
+    def assignConsistentDetailedNDens(self):
+        r"""Update ``self.p.detailedNDens`` to be consistent with child components.
+
+        Sets the parameter equal to the sum of detailed atoms
+
+        .. math::
+
+            N = \frac{\sum_c\left(N_c * V_c\right)}{\sum_cV_c}
+
+        If no detailed number densities are found, sets parameter to ``None``
+        """
+        totalVolume = 0.0
+        onAssembly = None
+        for c in self.iterComponents():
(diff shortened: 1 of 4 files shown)


armi/reactor/assemblies.py
```diff
@@ -1241,6 +1241,57 @@ def orientBlocks(self, parentSpatialGrid):
...
+            if (onChild := c.p.detailedNDens) is not None:
+                if onAssembly is not None:
+                    onAssembly += onChild * vol
+                else:
+                    onAssembly = onChild * vol
+        if onAssembly is not None:
+            onAssembly /= totalVolume
+            self.p.detailedNDens = onAssembly
+        else:
+            self.p.detailedNDens = None
```
drewj-tp, line 1269: So, I want to talk a bit about the design / potential design here.

This does not adhere to the composite tree model very much, wherein we may want to iterate directly on the children and ask the children to get their detailed number density and volume. That would look like adding something to `Composite` or `ArmiObject` that defines the method, delegating the work to children, like
```python
def getDetailedNumberDensity(self):
    totalVolume = 0.0
    totalDetailedAtoms = None
    for child in self:
        childDetailedNDens = child.getDetailedNumberDensity()
        childVolume = child.getVolume()
        totalVolume += childVolume
        if childDetailedNDens is None:
            continue
        if totalDetailedAtoms is not None:
            totalDetailedAtoms += childDetailedNDens * childVolume
        else:
            totalDetailedAtoms = childDetailedNDens * childVolume
    if totalDetailedAtoms is not None:
        return totalDetailedAtoms / self.getVolume()
    return None
```
with some opportunities for improvement in reducing the `getVolume` calls since we're already aggregating over the children.

Then, on the component level, returning the `self.p.detailedNDens` parameter as the terminating leaf.

Throwing this out there b/c I know @yardasol is interested in more generic composite tree work, and @john-science is constantly interested in more generic reactor support.

armi/reactor/assemblies.py
```diff
@@ -1241,6 +1241,57 @@ def orientBlocks(self, parentSpatialGrid):
...
+
+    def setPitch(self, pitchInCm: float):
+        """Adjust the assembly to alight with a uniform pitch.
+
+        Parameters
+        ----------
+        pitchInCm
+            Requested pitch of this assembly. Means different things for
+           different assemblies (e.g., hex vs. cartesian). But the child blocks
+           also know what to do with this.
```
john-science:
```suggestion
            Requested pitch of this assembly. Means different things for
            different assemblies (e.g., hex vs. cartesian). But the child blocks
            also know what to do with this.
```

john-science: Just RST formatting. NBD.

armi/reactor/assemblies.py
```diff
@@ -1241,6 +1241,57 @@ def orientBlocks(self, parentSpatialGrid):
...
+    def assignConsistentDetailedNDens(self):
+        r"""Update ``self.p.detailedNDens`` to be consistent with child components.
+
+        Sets the parameter equal to the sum of detailed atoms
+
+        .. math::
+
+            N = \frac{\sum_c\left(N_c * V_c\right)}{\sum_cV_c}
+
+        If no detailed number densities are found, sets parameter to ``None``
```
john-science:
```suggestion
        If no detailed number densities are found, sets parameter to ``None``.
```

armi/reactor/tests/test_assemblies.py
```diff
@@ -1045,6 +1045,41 @@ def test_getBIndexFromZIndex(self):
             bIndex = self.assembly.getBIndexFromZIndex(zIndex * 0.5)
             self.assertEqual(bIndex, math.ceil(zIndex / 2) if zIndex < 5 else -1)
 
+    def test_assemblyDetailedNDens(self):
+        """Ensure assembly and component number densities are consistent."""
+        # Set Up
+        numDetailedNucs = 100  # arbitrary
+        didUpdate = False
+        # Use random densities, we just need to show consistency
+        rng = np.random.default_rng()
```
john-science, line 1054: Does this affect other unit tests in ARMI?

If so, we can't have this done inside some random test, it needs to be visible at a very high level. Which worries me, as not the PR I would expect to create high-level differences in ARMI testing infrastructure.

armi/reactor/assemblies.py
john-science: Olek has a multi-step plan for working on the Composite model, and making it recursively (so a Composite can have a child of a Composite).

We can't do all that in this PR, but that's what he's working on today.

armi/reactor/tests/test_assemblies.py
drewj-tp: Can you explain what you mean by

> Does this affect other unit tests in ARMI?

All the existing unit tests have passed so I'm inclined to say no but I don't fully grasp what you're asking about

armi/reactor/assemblies.py
drewj-tp: 5345b013

armi/reactor/assemblies.py
drewj-tp: 5345b013

john-science (approved):

PR #2648, opened 2026-09-02
Removing child-related machinery from `ArmiObject`

### LINKED ISSUE #1449: ArmiObject has too much children logic
The difference between [ArmiObject](https://github.com/terrapower/armi/blob/753660d71f3777680d541ff5d042d2d71100f4a3/armi/reactor/composites.py#L266) and [Composite](https://github.com/terrapower/armi/blob/753660d71f3777680d541ff5d042d2d71100f4a3/armi/reactor/composites.py#L2588) is _supposed_ to be that `Composite`s have children.

But ArmiObject has a ton of children-specific logic, that just doesn't belong there:

https://github.com/terrapower/armi/blob/753660d71f3777680d541ff5d042d2d71100f4a3/armi/reactor/composites.py#L582

https://github.com/terrapower/armi/blob/753660d71f3777680d541ff5d042d2d71100f4a3/armi/reactor/composites.py#L599

https://github.com/terrapower/armi/blob/753660d71f3777680d541ff5d042d2d71100f4a3/armi/reactor/composites.py#L1427

https://github.com/terrapower/armi/blob/753660d71f3777680d541ff5d042d2d71100f4a3/armi/reactor/composites.py#L1671

https://github.com/terrapower/armi/blob/753660d71f3777680d541ff5d042d2d71100f4a3/armi/reactor/composites.py#L1982

It just doesn't seem like _all_ of this logic belongs here.

yardasol:

This PR moves children-related machinery from `ArmiObject` -- which should be a common interface for both `Composite` and `Component` -- to `Composite`, to better reflect the Composite Design Pattern.

closes #1449

## SCR Information

Change Type: features

One-Sentence Rationale: Only `Composite` and derived classes should have children and related machinery.

One-line Impact on Requirements: Moves implementation of I_ARMI_PARAM1, I_ARMI_CMP_GET_MASS, I_ARMI_CMP_NUC0, I_ARMI_CMP_NUC1, I_ARMI_CMP0, I_ARMI_CMP_GET_MASS from the `ArmiObject` class to the `Composite` class.

---

## Checklist


john-science:
The code coverage on this file is 94%, so it's good: https://app.codecov.io/gh/terrapower/armi/blob/main/armi%2Freactor%2Fcomposites.py

I notice only one method has no code coverage at all: `changeNDensByFactor()`.

Do all the methods you moved have explicit unit tests?

armi/reactor/composites.py
```diff
@@ -459,12 +446,6 @@ def duplicate(self):
         """
         raise NotImplementedError
 
-    def clearCache(self):
-        """Clear the cache so all new values are recomputed."""
-        self.cached = {}
```
john-science, line 464: Are you sure we don't want to leave the method `clearCache()` on ArmiObject, but just leave this first line, and not the loop below?

I see `.cached` is used in a couple of methods.

armi/reactor/composites.py
john-science, line 293: This `impl` tag should be mentioned in the PR description, under "One-line Impact on Requirements".

armi/reactor/composites.py
```diff
@@ -895,14 +738,6 @@ def getMass(self, nuclideNames: Union[None, str, list[str]] = None) -> float:
         """
         Determine the mass in grams of nuclide(s) and/or elements in this object.
 
-        .. impl:: Return mass of composite.
-            :id: I_ARMI_CMP_GET_MASS
```
john-science, line 899: This `impl` tag should be mentioned in the PR description, under "One-line Impact on Requirements".

armi/reactor/composites.py
```diff
@@ -1209,64 +1044,30 @@ def adjustMassEnrichment(self, massFraction):
         """
         raise NotImplementedError
 
-    def getNumberDensity(self, nucName):
-        """
-        Return the number density of a nuclide in atoms/barn-cm.
-
-        .. impl:: Get number density for a specific nuclide
-            :id: I_ARMI_CMP_NUC0
```
john-science, line 1217: This `impl` tag should be mentioned in the PR description, under "One-line Impact on Requirements".

armi/reactor/composites.py
```diff
@@ -1209,64 +1044,30 @@ def adjustMassEnrichment(self, massFraction):
...
-        ArmiObject.getNuclideNumberDensities: More efficient for >1 specific nuc density is needed.
-        ArmiObject.getNumberDensities: More efficient for when all nucs in object is needed.
-        """
-        return self.getNuclideNumberDensities([nucName])[0]
-
     def getNuclideNumberDensities(self, nucNames):
         """Return a list of number densities in atoms/barn-cm for the nuc names requested.
 
-        .. impl:: Get number densities for specific nuclides.
-            :id: I_ARMI_CMP_NUC1
```
john-science, line 1242: This `impl` tag should be mentioned in the PR description, under "One-line Impact on Requirements".

armi/reactor/composites.py
```diff
@@ -2083,406 +1550,699 @@ def getSymmetryFactor(self):
...
+    Core and individual 1-D Z-meshes on the assemblies. Then, Assemblies have 2-D
+    spatialLocators (i,j,0) and Blocks have 1-D spatiaLocators (0,0,k). These get added
+    to form the global indices. This way, if an assembly is moved, all the blocks
+    immediately and naturally move with it. Individual children may have
+    coordinate-based spatialLocators mixed with siblings in a grid. This allows mixing
+    grid-representation with explicit representation, often useful in advanced
+    assemblies and thermal reactors.
+
+    .. impl:: Composites are a physical part of the reactor in a hierarchical data model.
+        :id: I_ARMI_CMP0
```
john-science, line 1596: This `impl` tag should be mentioned in the PR description, under "One-line Impact on Requirements".

armi/reactor/composites.py
```diff
@@ -2083,406 +1550,699 @@ def getSymmetryFactor(self):
...
+        fracs = [(ci, nu / denom) for ci, nu in zip(children, numerator)]
+        return fracs
+
+    def getMass(self, nuclideNames: Union[None, str, list[str]] = None) -> float:
         """
-        Return list of components in this block of a particular shape.
+        Determine the mass in grams of nuclide(s) and/or elements in this object.
+
+        .. impl:: Return mass of composite.
+            :id: I_ARMI_CMP_GET_MASS
```
john-science, line 1770: This `impl` tag should be mentioned in the PR description, under "One-line Impact on Requirements".

armi/reactor/composites.py
```diff
@@ -2083,406 +1550,699 @@ def getSymmetryFactor(self):
...
 
-            if j >= maxJ:
-                maxJ = j
-            if j <= minJ:
-                minJ = j
+    def _updateVolume(self):
+        """Recompute and store volume."""
+        children = self.getChildren()
+        # Derived shapes must come last so we temporarily change the order if we
+        # have one.
```
john-science:
```suggestion
        # Derived shapes must come last so we temporarily change the order if we have one.
```

armi/reactor/composites.py
```diff
@@ -2083,406 +1550,699 @@ def getSymmetryFactor(self):
...
-        return ((minI, maxI), (minJ, maxJ), (minK, maxK))
+    def getVolumeFractions(self):
+        """
+        Return volume fractions of each child.
 
-    def getComponentNames(self):
-        r"""
-        Get all unique component names of this Composite.
+        Sets volume or area of missing piece (like coolant) if it exists.  Caching would
+        be nice here.
```
john-science:
```suggestion
        Sets volume or area of missing piece (like coolant) if it exists.
```

armi/reactor/composites.py
```diff
@@ -2657,6 +2417,97 @@ def getChildren(
...
+
+    def copyParamsToChildren(self, paramNames):
+        """
+        Copy param values in paramNames to all children.
+
+        Parameters
+        ----------
+        paramNames : list
+            List of param names to copy to children
+
```
john-science:
```suggestion
```

armi/reactor/composites.py
```diff
@@ -2718,6 +2569,261 @@ def iterComponents(self, typeSpec: TypeSpec = None, exact: bool = False) -> Iter
         """
         return (c for child in self for c in child.iterComponents(typeSpec, exact))
 
+    def getComponentNames(self):
+        r"""
```
john-science:
```suggestion
        """
```

armi/reactor/composites.py
```diff
@@ -2718,6 +2569,261 @@ def iterComponents(self, typeSpec: TypeSpec = None, exact: bool = False) -> Iter
...
+        ----------
+        material : armi.materials.material.Material, optional
+            The material to match
+        materialName : str, optional
+            The material name to match.
+
+        Returns
+        -------
+        componentsWithThisMat : list
+
```
john-science:
```suggestion
```

armi/reactor/composites.py
```diff
@@ -2083,406 +1550,699 @@ def getSymmetryFactor(self):
...
-    def getAverageTempInC(self, typeSpec: TypeSpec = None, exact=False):
-        """Return the average temperature of the ArmiObject in C by averaging all components."""
-        tempNumerator = 0.0
-        totalVol = 0.0
-        for component in self.iterComponents(typeSpec, exact):
-            vol = component.getVolume()
-            tempNumerator += component.temperatureInC * vol
-            totalVol += vol
+            if not indiciesToSet:
+                if dens == 0:
```
john-science, line 1946: We shouldn't change this in this PR. But my gut tells me this should be:

```python
if dens <= 0:
```

You'd be surprised how often people get to negatives.

armi/reactor/composites.py
yardasol: Good catch. Both `Component` and `Composite` have a `clearCache` method. While `Component.clearCache` completely overrides the parent `clearCache` method, it should still be defined in `ArmiObject` so they have the same interface. Thoughts?

armi/reactor/composites.py
yardasol: This is a good catch.

yardasol:
> Do all the methods you moved have explicit unit tests?

I will check

**UPDATE**
Functions with missing unit tests:
- [x] `getChildrenWithFlags`
- [x] `getChildrenWithNuclides`
  - ~~This one is tricky. This function calls `child.getNuclides` on a `Composite` object, assuming that `child.getNuclides` actually returns a list of nuclides. The most straightforward implementation of a unit test for this function calls this function on the test `Composite` which has `DummyLeaf` objects as children. The problem is that `DummyLeaf` inherits from `Composite`, not `Component`, so it gets, both of which subclass the main `Composite` class. This means the the `Composite.getNuclides` function is called, not the terminal `Component.getNuclides` which actually gets the nuclides from the associated parameter object.~~
  - nvm, I was trying to implement this test in `TestCompositePattern` when I needed to do implement it in `TestCompositeTree`
- [x] `getChildrenOfType`
- [x] `getChildParamValues`
- [x] `__add__`
- [x] `clearCache`
- [x] `getArea` -> test exists in `test_blocks.py`. Sufficient? [ ]
- [x] `getVolumeFractions` -> tests exist in `test_blocks.py`. Sufficient? [ ]
- [x] `getVolume`
- [x] `getNuclideNumberDensities` (see test_getNumberDensities`)
- [x] `setNumberDensities`
- [x] `updateNumberDensities`
- [x] `calcAvgParam` -> test exists in `test_assemblies.py`
  - Extended the test 
- [x] `isFuel`
- [x] `getHMMoles`
- [x] `getHMDens` -> test exists in `test_blocks.py`. 
  - [x] Sufficient
- [x] `getMasses` -> test exists in `test_blocks.py`.
  - [x] Sufficient
- [x] `getMgFlux` -> test exists in `test_blocks.py`.
  - [x] Sufficient
- [x] `expandAllElementalsToIsotopics` -> test exists in `test_blocks.py`
  - [x] Sufficient
- [x] `expandElementalToIsotopics` -> test exists in `test_blocks.py`
  - [x] Sufficient
- [x] `getAverageTempInC`
- [x] `resolveLinkedDims` -> removed this function since it is unused in ARMI
- [x] `getDominantMaterial`

No unit tests for `Composite` component functions exist (the functions moved in [this commit](https://github.com/terrapower/armi/pull/2648/changes/c01c18e0b545c741e13c64f6ac586ea57d3799fe)), but they do appear to exist in the `test_blocks.py`. The `Block` class does not override any of those functions, do we think this is sufficient?

armi/reactor/composites.py
john-science: Well, if `ArmiObject` has a `.cached`, we need that to be functional. That's all I'm thinking.

armi/reactor/composites.py
```diff
@@ -395,23 +394,15 @@ def __setstate__(self, state):
 
         Notes
         -----
-        This ArmiObject may have lost a reference to its parent. If the parent was also
-        pickled (serialized), then the parent should update the ``.parent`` attribute
-        during its own ``__setstate__``. That means within the context of
-        ``__setstate__`` one should not rely upon ``self.parent``.
+        This ArmiObject may have lost a reference to its parent. If the parent (a Composite, Assembly, or Block) was
```
john-science, line 397: It might be more accurate to say "(a Composite or subclass)".

I'm just thinking because `Core` and `Reactor` also have this method. Ditto `HexBlock`.

But we could also rewrite this to say something like "(a Composite, Assembly, Block, or similar)"

armi/reactor/composites.py
john-science: We also need to be sure that the requirement still has an impl tag... hopefully moved to the other class?

armi/reactor/composites.py
```diff
@@ -1209,64 +1044,30 @@ def adjustMassEnrichment(self, massFraction):
         """
         raise NotImplementedError
 
-    def getNumberDensity(self, nucName):
```
john-science, line 1212: Wait, why wouldn't an ArmiObject have number densities?

Is that tied to children in some way?

armi/reactor/composites.py
```diff
@@ -1578,391 +1304,136 @@ def getUraniumNumEnrich(self):
...
-        Sums up a parameter throughout the object's children or list of objects.
+    def isFuel(self):
+        """True if this is a fuel block."""
+        return self.hasFlags(Flags.FUEL)
 
-        Parameters
-        ----------
-        param : str
-            Name of the block parameter to sum
+    def containsHeavyMetal(self):
```
john-science, line 1311: It looks to me like this method can be safely deleted.

armi/reactor/composites.py
```diff
@@ -1578,391 +1304,136 @@ def getUraniumNumEnrich(self):
...
-        param,
-        typeSpec: TypeSpec = None,
-        weightingParam=None,
-        volumeAveraged=True,
-        absolute=True,
-        generationNum=1,
-    ):
-        r"""
-        Calculate the child-wide average of a parameter.
+    def constituentReport(self):
```
john-science, line 1378: This appears to me to be unused in the world.

Maybe we can delete it?

john-science: @keckler @mgjarrett @drewj-tp  Does anyone call this report?  It's not called during any of standard runs, or any of your codebases that I see.

armi/reactor/composites.py
```diff
@@ -1578,391 +1304,136 @@ def getUraniumNumEnrich(self):
...
+        """A print out of some pertinent constituent information."""
+        from armi.utils import iterables
```
john-science, line 1380: Does moving this import to the top of the file really cause an infinite import loop?

If not, we should move this to the top. (If the method stays.)

armi/reactor/composites.py
```diff
@@ -2480,201 +1705,1091 @@ def insert(self, index, obj):
...
+        error. This would only happen if some unnatural nuclide like Pu239 built up in
+        fresh UZr. That should be anticipated and dealt with elsewhere.
+
+        Parameters
+        ----------
+        nucName : str
+            Nuclide number density to modify
+        val : float
+            Number density to set in atoms/bn-cm (heterogeneous)
+
```
john-science, line 1889:
```suggestion
```

armi/reactor/composites.py
```diff
@@ -2480,201 +1705,1091 @@ def insert(self, index, obj):
...
+        The complication is that various number densities are distributed among various
+        components. This sets the number density for each nuclide evenly across all
+        components that contain it.
+
+        Parameters
+        ----------
+        numberDensities : dict
+            nucName: ndens pairs.
+
```
john-science, line 1938:
```suggestion
```

armi/reactor/composites.py
```diff
@@ -2480,201 +1705,1091 @@ def insert(self, index, obj):
...
+        allDehomogenizedNDens = collections.defaultdict(dict)
+
+        # compute potentially-different homogenization factors for each child.  evenly
+        # distribute entire number density over the subset of active children.
+        for nuc, dens in numberDensities.items():
+            # get "active" indices, i.e., indices of children containing nuc
+            # NOTE: this is one of the rare instances in which (imo), using explicit
+            # indexing clarifies subsequent code since it's not necessary to zip +
+            # filter + extract individual components (just extract by filtered index).
```
john-science, line 1951:
```suggestion
            # Get "active" indices, i.e. indices of children containing nuclide.
            # NOTE: This uses explicit indexing to clarify subsequent code, since it is not necessary to zip + filter +
            # extract individual components (we just extract by filtered index).
```

john-science: I won't die on this hill. I'm just trying to clean up the language.

armi/reactor/composites.py
```diff
@@ -2480,201 +1705,1091 @@ def insert(self, index, obj):
...
+        generationNum : int, optional
+            Which generation to consider. 1 means direct children, 2 means children of
+            children. Default: Just return direct children.
+
+        calcBasedOnFullObj : bool, optional
+            Some assemblies or blocks, such as the center assembly in a third core
+            model, are not modeled as full assemblies or blocks. In the third core model
+            objects at these positions are modeled as having 1/3 the volume and thus 1/3
+            the power. Setting this argument to True will apply the full value of the
+            parameter as if it was a full block or assembly.
```
john-science, line 2028:
```suggestion

        objs : iterable, optional
            A list of objects to sum over. If None, all children in object will be used
        volumeIntegrated : bool, optional
            Integrate over volume
        addSymmetricPositions : bool, optional
            If True, will multiply by the symmetry factor of the core (3 for 1/3 models, 1 for full core models)
        typeSpec : TypeSpec
            object types to restrict to
        generationNum : int, optional
            Which generation to consider. 1 means direct children, 2 means grandchildren. Default: 1.
        calcBasedOnFullObj : bool, optional
            Some assemblies or blocks, such as the center assembly in a third-core model, are not modeled as full
            assemblies or blocks. In the third-core model, objects at these positions are modeled as having 1/3 the
            volume and thus 1/3 the power. Setting this argument to True will apply the full value of the parameter as
            if it was a full block or assembly.
```

armi/reactor/composites.py
```diff
@@ -2480,201 +1705,1091 @@ def insert(self, index, obj):
...
+
+        volumeAveraged : bool, optional
+            volume (or height, or area) average this param
+
+        absolute : bool, optional
+            Returns the average of the absolute value of param
+
+        generationNum : int, optional
+            Which generation to average over (1 for children, 2 for grandchildren)
+
```
john-science, line 2092:
```suggestion
        param : str
            The ARMI block parameter that you want to average
        typeSpec : TypeSpec
            The child types that should be included in the calculation. Restrict average
            to a certain child type with this parameter.
        weightingParam : None or str, optional
             An optional block param that the average will be weighted against
        volumeAveraged : bool, optional
            volume (or height, or area) average this param
        absolute : bool, optional
            Returns the average of the absolute value of param
        generationNum : int, optional
            Which generation to average over (1 for children, 2 for grandchildren)
```

armi/reactor/composites.py
```diff
@@ -2480,201 +1705,1091 @@ def insert(self, index, obj):
...
+            >>> checker = lambda c: len(c.name) % 3
+            >>> for child in c.getChildren(predicate=checker):
+            ...     pass
+            >>> for child in c.iterChildren(predicate=checker):
+            ...     pass
+            >>> for child in filter(checker, c):
+            ...     pass
+
+        If you're going to be doing traversal beyond the first generation, this method will help you.
+
```
john-science, line 2315:
```suggestion
```

armi/reactor/tests/test_assemblies.py
```diff
@@ -796,12 +790,31 @@ def test_getElevationsMatchingParamValue(self):
...
+            powers.append(random.random())
+            child.p.power = powers[-1]
+        self.assertGreater(len(powers), 2)
+        refAvgPower = sum(powers) / len(powers)
+        self.assertAlmostEqual(self.assembly.calcAvgParam("power"), refAvgPower)
+        ## All the children have the same volume, so turning off volume averaging should yield the same answer
+        self.assertAlmostEqual(self.assembly.calcAvgParam("power", volumeAveraged=False), refAvgPower)
+        ## All powers are positive, so turning off absolute should yield the same answer
+        self.assertAlmostEqual(self.assembly.calcAvgParam("power", absolute=False), refAvgPower)
+        ## All children have the same height, so weighting by the height should yield the same answer
```
john-science, line 804:
```suggestion
        # All the children have the same volume, so turning off volume averaging should yield the same answer
        self.assertAlmostEqual(self.assembly.calcAvgParam("power", volumeAveraged=False), refAvgPower)
        # All powers are positive, so turning off absolute should yield the same answer
        self.assertAlmostEqual(self.assembly.calcAvgParam("power", absolute=False), refAvgPower)
        # All children have the same height, so weighting by the height should yield the same answer
```

john-science: I had to stare at this for a bit to decide if I care about `##` of `#`. On the one hand, no, it's not worth caring about. On the other hand... this isn't done anywhere else in ARMI, so consistency is easier to work with in the long term.

#shrug I'm not dying on this hill.

armi/reactor/tests/test_assemblies.py
```diff
@@ -796,12 +790,31 @@ def test_getElevationsMatchingParamValue(self):
...
-        nums = []
-        for b in self.assembly:
-            nums.append(random.random())
-            b.p.power = nums[-1]
-        self.assertGreater(len(nums), 2)
-        self.assertAlmostEqual(self.assembly.calcAvgParam("power"), sum(nums) / len(nums))
+        powers = []
+        for child in self.assembly:
+            powers.append(random.random())
+            child.p.power = powers[-1]
```
john-science, line 796:
```suggestion
        for block in self.assembly:
            powers.append(random.random())
            block.p.power = powers[-1]
```

armi/reactor/tests/test_assemblies.py
```diff
@@ -796,12 +790,31 @@ def test_getElevationsMatchingParamValue(self):
...
     def test_calcAvgParam(self):
-        nums = []
-        for b in self.assembly:
-            nums.append(random.random())
-            b.p.power = nums[-1]
-        self.assertGreater(len(nums), 2)
-        self.assertAlmostEqual(self.assembly.calcAvgParam("power"), sum(nums) / len(nums))
+        powers = []
+        for child in self.assembly:
+            powers.append(random.random())
```
john-science, line 795: Using random number is interesting here.

If all of 3 of these numbers randomly generated to zero, this test would fail. Though that's unlikely, it would lead to someone doing some tricky debugging.

You could fix that with something like this:

```python
powers.append(random.random() + 1e-10)
```

I know [some people](https://softwareengineering.stackexchange.com/questions/429601/is-the-usage-of-random-values-in-unit-testing-a-good-practice) have strong opinions against random numbers in unit testing. I'm not sure I feel super strongly about it, and I'm just remembering what I've been told.

Thoughts?

armi/reactor/tests/test_blocks.py
```diff
@@ -987,44 +986,50 @@ def test_getTotalNDens(self):
...
 
-        hmDens = 0.0
-        for nuclide in refDict.keys():
+        testHMDens = self.block.getHMDens()
+
+        places = 6
+        self.assertAlmostEqual(testHMDens, refHMDens, places=places)
+
+    def test_getHMMoles(self):
+        from armi.utils import units
```
john-science, line 1018: By preference, I like all the imports to be at the top of the files. It's part of [PEP8](https://peps.python.org/pep-0008/#imports), but more importantly I have found a lot of problematic imports in our various codebases and have to clean them up sometimes. (And thank you `ruff` for making that easier.) Which is hard to do if they are hidden.

armi/reactor/tests/test_composites.py
```diff
@@ -553,13 +560,23 @@ def test_iterChildrenWithFlags(self):
...
     def test_iterChildrenOfType(self):
         clads = self.container.iterChildrenOfType("clad")
         first = next(clads)
         self.assertIs(first, self.cladChild)
         with self.assertRaises(StopIteration):
             next(clads)
 
+    def test_getChildrenOfType(self):
+        clads = self.container.getChildrenOfType("clad")
+        assert len(clads) == 1
```
john-science, line 577: It might be worth checking something like "wire" which isn't here, to show it can return zero.

What happens if you check for the type "bob" or "lavalamp"? Does it fail in a way we like?

armi/reactor/tests/test_composites.py
```diff
@@ -594,6 +611,12 @@ def test_add(self):
         # verify the container's size has increased by one
         self.assertEqual(len(self.container), lenContainer + 1)
 
+    def test___add__(self):
```
john-science, line 614:
```suggestion
    def test__add(self):
```

john-science: This seems to be the ARMI test naming convention. #shrug

armi/reactor/tests/test_composites.py
```diff
@@ -686,7 +709,6 @@ def setUp(self):
             "NA23": 2e-2,
             "ZR": 0.00709003962772,
         }
-        self.block.setNumberDensities(self.refDict)
```
john-science, line 689: Wait, why are we removing this?

armi/reactor/tests/test_composites.py
```diff
@@ -771,11 +804,82 @@ def test_getNuclides(self):
...
+        # try with just U235
+        nuclideList = ["U235"]
+        testChildren = self.block.getChildrenWithNuclides(nuclideList)
+        self.assertListEqual(testChildren, refChildren)
+        # try with just U238
+        nuclideList = ["U238"]
+        testChildren = self.block.getChildrenWithNuclides(nuclideList)
+        self.assertListEqual(testChildren, refChildren)
+
+        ## Multi block: sodium coolant
```
john-science, line 842:
```suggestion
        # Multi block: sodium coolant
```

armi/reactor/tests/test_composites.py
```diff
@@ -771,11 +804,82 @@ def test_getNuclides(self):
...
+
+        ## Empty list input
+        refChildren = []
+        nuclideList = []
+        testChildren = self.block.getChildrenWithNuclides(nuclideList)
+        self.assertListEqual(testChildren, refChildren)
+
+    def test_getChildParamValues(self):
+        """Test getChildParamValues."""
+        import numpy as np
```
john-science, line 857: Can we move this to the top of the file. [PEP8](https://peps.python.org/pep-0008/#imports).

Thanks!

armi/reactor/tests/test_composites.py
```diff
@@ -771,11 +804,82 @@ def test_getNuclides(self):
...
+            "coolant",
+            "duct",
+            "interCoolant",
+        ]
+        testTypes = self.block.getChildParamValues("type")
+        self.assertIsInstance(testTypes, np.ndarray)
+
+        self.assertSequenceEqual(list(testTypes), refTypes)
+
+        ## Check error trips when unknown parameter passed in
```
john-science, line 880:
```suggestion
        # Check error trips when unknown parameter passed in
```

armi/reactor/tests/test_composites.py
```diff
@@ -860,7 +964,7 @@ def test_setMassFrac(self):
         fuelComponent = components.Circle("fuel", "UZr", **fuelDims)
         c.add(fuelComponent)
 
-        # test initial state
+        # test initial state ang getMass, getFPMass, and getHMMass functions
```
john-science, line 967: ang?

armi/reactor/tests/test_composites.py
```diff
@@ -891,6 +995,98 @@ def test_getFissileMass(self):
...
+        # Cleanup
+        self.block.remove(refFuel)
+
+    def test_getVolume(self):
+        """Test getVolume."""
+        import math
+        from copy import copy
+
+        from armi.reactor import blocks
+        from armi.reactor.components import Circle
```
john-science, line 1065: Can we move all these imports to the top of the file?

Thanks!

armi/reactor/tests/test_composites.py
```diff
@@ -891,6 +995,98 @@ def test_getFissileMass(self):
...
+
+    def test_getDominantMaterial(self):
+        """Test getDominantMaterial."""
+        ## Case: fuel flag should just be UZr
+        refFuel = self.block._children[2].material
+        testFuel = self.block.getDominantMaterial(Flags.FUEL)
+        self.assertEqual(refFuel, testFuel)
+
+        ## Case: add a new fuel component that is WAY bigger
+        from armi.reactor.components import Circle
```
john-science, line 1048: Can we move this import to the top of the file?

Thanks!

armi/reactor/tests/test_composites.py
```diff
@@ -1062,9 +1259,9 @@ def test_getNumberDensities(self):
     def test_getNumDensWithExpandedFissProds(self):
         """Get number densities from composite.
 
-        .. test:: Get number densities.
-            :id: T_ARMI_CMP_NUC
-            :tests: R_ARMI_CMP_NUC
+        .. test:: Number density of composite is retrievable.
+            :id: T_ARMI_CMP_GET_NDENS1
```
john-science, line 1263: Not that it matters, but what's going on here?

Does `R_ARMI_CMP_NUC` still have impl tags?

armi/reactor/tests/test_composites.py
```diff
@@ -1103,6 +1300,66 @@ def test_getNumDensWithExpandedFissProds(self):
...
+
+        ## Case: setting a new number density should add it to the nuclide densities
+        testNdens = self.obj.getNumberDensities()
+        self.assertTrue("H1" not in testNdens.keys())
+        self.obj.updateNumberDensities({"H1": 0.0001})
+        testNdens = self.obj.getNumberDensities()
+        self.assertAlmostEqual(testNdens["H1"], 0.0001, 7)
+
+        ## Cleanup
+        import numpy as np
```
john-science, line 1336: Can we move this to the top of the file?

Thanks!

armi/reactor/composites.py
yardasol: It does and was

armi/reactor/composites.py
yardasol: This called `ArmiObject.getNuclideNumberDensities`, which had children-machinery inside of it, so I moved it to the Composite class since the ArmiObject class needs to be the interface for both `Composite` and `Component`.

armi/reactor/tests/test_composites.py
yardasol: This has no effect on the tests and adds in a bunch of nuclides in places they shouldn't be.

armi/reactor/tests/test_composites.py
yardasol: Okay, so I initially thought that this test was incorrectly labeled as testing the R_ARMI_CMP_NUC requirement, as this is by implemented by `getNuclideNumberDensities`, *not* `getNumberDensities` (which this function is testing). However there is a bigger problem here:

Note that `getNumberDensities()` (impementing R_ARMI_CMP_GET_NDENS) returns _all_ nuclide densities, whereas `getNuclideNumberDensities(nucNames)` (implementing R_ARMI_CMP_NUC) returns the densities for nuclides specified by the `nucNames` parameter.

Note also the text for R_ARMI_CMP_NUC:
>  Composites shall be able to return number densities for all their nuclides.

and for R_ARMI_CMP_GET_NDENS:
>The homogenized number densities of specified nuclides in a composite shall be retrievable

**The description of each requirement appears to be mismatched with the linked implementing function.**

armi/reactor/tests/test_assemblies.py
yardasol: I like it!

armi/reactor/tests/test_composites.py
john-science: Great! Open a ticket.

yardasol: https://github.com/terrapower/armi/issues/2656

armi/reactor/composites.py
john-science: Okay, when I look at the code, only `ArmiObject` and NOT `Composite` has a `clearCache()` method.

And, to me, it makes more sense for `clearCache()` to be defined on `ArmiObject` than on `Composite`, because that is the base class.

Thoughts?

armi/reactor/composites.py
```diff
@@ -395,23 +394,16 @@ def __setstate__(self, state):
 
         Notes
         -----
-        This ArmiObject may have lost a reference to its parent. If the parent was also
-        pickled (serialized), then the parent should update the ``.parent`` attribute
-        during its own ``__setstate__``. That means within the context of
-        ``__setstate__`` one should not rely upon ``self.parent``.
+        This ArmiObject may have lost a reference to its parent. If the parent (a Composite or Composite
+        subclass) was
```
john-science:
```suggestion
        This ArmiObject may have lost a reference to its parent. If the parent (a Composite or Composite subclass) was
```

john-science: An entire line with only two words seems odd. That's all.

armi/reactor/composites.py
john-science: Huh.  This is really odd.

Moving number densities out of `ArmiObject` essentially makes `ArmiObject` a dumb base class that can never be used directly. (You can't do anything in a nuclear model with number densities.)

I'm not sure this is bad or wrong, but it's worth thinking about. It's a pretty big side effect.

armi/reactor/tests/test_composites.py
john-science: But we're leaving `self.refDict` in the test?

armi/reactor/composites.py
yardasol: I'm confused, `Composite` definitely has a `clearCache` method, at line 1709:
```python
     def clearCache(self):
         """Clear the cache so all new values are recomputed."""
         ArmiObject.clearCache(self)
         for child in self:
             child.clearCache()
```

armi/reactor/composites.py
john-science: We talked about this. Done.

armi/reactor/composites.py
yardasol: done

armi/reactor/composites.py
yardasol: done

armi/reactor/composites.py
yardasol: done

armi/reactor/tests/test_blocks.py
yardasol: Done

armi/reactor/tests/test_composites.py
yardasol: Done

armi/reactor/tests/test_composites.py
yardasol: Done

armi/reactor/tests/test_composites.py
yardasol: Done

armi/reactor/composites.py
drewj-tp: TIL about this so nope, I do not use this

armi/reactor/tests/test_composites.py
yardasol: typo, was supposed to be "and", should be fixed now

armi/reactor/composites.py
yardasol: This is a different ticket maybe?

armi/reactor/composites.py
yardasol: This is part of the constituent report, we can delete it?

armi/reactor/tests/test_composites.py
yardasol: done

armi/reactor/tests/test_composites.py
yardasol: done

armi/reactor/tests/test_composites.py
yardasol: done

armi/reactor/composites.py
yardasol: nvm i deleted it

armi/reactor/composites.py
yardasol: deleted

armi/reactor/composites.py
yardasol: done

armi/reactor/composites.py
yardasol: done

armi/reactor/tests/test_assemblies.py
yardasol: did this

armi/reactor/tests/test_composites.py
yardasol: Passing "bob" doesn't raise any sort of errors, just returns an empty list.

Internally, `iterChildrenOfType(typeName)` just calls:
```python
self.iterChildren(predicate=lambda o: o.getType() == typeName)
```
 which basically loops through all of the children and adds it to the output if `child.getType() == typeName`. There is no checking whether typeName is valid.

There is a note in the docstring for `ArmiObject.setType()` that seems relevant:

> We are in the process of developing more robust definitions for things like
   "name" and "type". "type" will generally refer to the name of the blueprint that
   created a particular object. When present, a "name" will refer to a specific
   instance of an object of a particular "type". Think unique names for each
   assembly in a core, even if they are all created from the same blueprint and
   therefore have the same "type". When this work is complete, it will be strongly
   discouraged, or even disallowed to change the type of an object after it has
   been created, and ``setType()`` may be removed entirely.

it seems that the ARMI type system may need a bit of a facelift.

armi/reactor/tests/test_composites.py
yardasol: It's used in a couple other tests in this file

armi/reactor/tests/test_composites.py
john-science: Okay, well... 

@opotowsky @yardasol 

I think this shows putting a "TODO" in a comment in the code like this is a bad idea. That comment is over 7 years old, and just disappeared for that whole time.

Doubly so because this is a "TODO" that isn't marked as such, so all my greps over the years never found this.

Should we make this a ticket?

john-science (approved):

armi/reactor/tests/test_composites.py
yardasol: There are a couple other notes like this in the code

yardasol: yes we should make a ticket

PR #2650, opened 2026-09-03
Removing numMoves parameter

### LINKED ISSUE #1590: `numMoves` is different between cases depending on `numProcessors`
Here's an example of running a difference comparison on a case with 24 versus 36 procs. 

Perhaps the fuel shuffling interface is involved. I didn't dig into it too much.

```
                                         /c00n00    /c00n01     /c00n02
       --------------------  ---------------  ---------  ---------  ----------
       HexAssembly/numMoves  max(abs(diff))    0.222222   0.25       0.26087
       HexAssembly/numMoves  mean(abs(diff))              0.011236   0.0117245
       HexAssembly/numMoves  mean(diff)                  -0.011236  -0.0117245
```

john-science:

There are some real concerns about the correctness of `numMoves` in parallel runs. It seems like it is being updated from a class attribute that can change during asynchronous operations.

And that would all be very difficult to fix... Luckily, Arrielle points out that this `Parameter` is not actually used in ARMI or anywhere we can find downstream. So we're just going to remove this `Parameter` that is hard to make correct anyway.

Thanks @opotowsky for the solution!

close #1590

## SCR Information

Change Type: fixes

One-Sentence Rationale: We should not have Parameters around if we cannot validate their correctness.

---

## Checklist

### DIFF
--- a/armi/physics/fuelCycle/fuelHandlers.py
+++ b/armi/physics/fuelCycle/fuelHandlers.py
@@ -249,7 +249,6 @@ class FuelHandler:
         else:
             numMoved = 0
 
-        self.o.r.core.p.numMoves = numMoved
         self.o.r.core.setBlockMassParams()
 
         runLog.important("Fuel handler performed {0} assembly shuffles.".format(numMoved))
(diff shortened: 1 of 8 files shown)


opotowsky (approved):
🧹

PR #2654, opened 2026-09-11
Improving flexibility of CaseSuite.discover

### LINKED ISSUE #2181: Would be great if case discovery didn't require the `yaml` file ending
Case discovery currently requires the `yaml` ending. It would be great if the case discovery mechanism was smart enough to assume this. 

https://github.com/terrapower/armi/blob/a69b905e8a0210c4f396a052141bd96c1d4571b5/armi/cases/suite.py#L88

john-science:

Here I beef up `CaseSuite.discover()` so that it supports `discover(pattern="input")` not just `discover(pattern="input*")`. This required some improvements to `recursivelyLoadSettingsFiles()` as well.

Also, `CaseSuite.discover()` didn't have any unit tests and `recursivelyLoadSettingsFiles()` didn't have enough testing.

close #2181

## SCR Information

Change Type: features

One-Sentence Rationale: The user base requested this feature should be more flexible and easier to use.

---

## Checklist

### DIFF
--- a/armi/cases/suite.py
+++ b/armi/cases/suite.py
@@ -13,15 +13,15 @@
 # limitations under the License.
 
 r"""
-The ``CaseSuite`` object is responsible for running, and executing a set of user inputs.  Many
-entry points redirect into ``CaseSuite`` methods, such as ``clone``, ``compare``, and ``submit``.
+The ``CaseSuite`` object is responsible for running, and executing a set of user inputs. Many entry points redirect into
+``CaseSuite`` methods, such as ``clone``, ``compare``, and ``submit``.
 
-Used in conjunction with the :py:class:`~armi.cases.case.Case` object, ``CaseSuite`` can be used to
-collect a series of cases and submit them to a cluster for execution. Furthermore, a ``CaseSuite``
-can be used to gather executed cases for post-analysis.
+Used in conjunction with the :py:class:`~armi.cases.case.Case` object, ``CaseSuite`` can be used to collect a series of
+cases and submit them to a cluster for execution. Furthermore, a ``CaseSuite`` can be used to gather executed cases for
+post-analysis.
 
-``CaseSuite``\ s should allow ``Cases`` to be added from totally separate directories. This is
-useful for plugin-informed testing as well as other things.
+``CaseSuite``\ s should allow ``Cases`` to be added from totally separate directories. This is useful for plugin-
+informed testing as well as other things.
 
 See Also
 --------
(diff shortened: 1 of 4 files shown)


armi/settings/tests/test_settings.py
```diff
@@ -462,21 +474,31 @@ def setUp(self):
...
+            settings.recursivelyLoadSettingsFiles(".", ["*.yaml"], ignorePatterns="skip*")
 
+        # find files non-recursively via *.yaml with an ignore pattern skip*
         loadedSettings = settings.recursivelyLoadSettingsFiles(
             ".", ["*.yaml"], recursive=False, ignorePatterns=["skip*"]
         )
         names = {cs.caseTitle for cs in loadedSettings}
         self.assertIn("settings1", names)
         self.assertIn("settings2", names)
         self.assertNotIn("settings3", names)
```
opotowsky, line 492:
```suggestion
        self.assertNotIn("notSettings", names)
        self.assertNotIn("settings3", names)
```
Right?

We could also test that one more with a test that asserts `InvalidSettingsFileError` is raised.

armi/cases/suite.py
```diff
@@ -112,6 +109,22 @@ def discover(
         skipInspection : bool, optional
             if True, skip running the check inputs
         """
+        assert not isinstance(patterns, str), "Bare string passed as patterns. Make sure to pass a list"
+        assert not isinstance(ignorePatterns, str), "Bare string passed as ignore patterns. Make sure to pass a list"
+
+        # handle the case where someone gives just a string (input), but they meant to add an asterisk (input*)
```
opotowsky, line 115: so this is interesting. in practice we'd have ppl pass in `blah.yaml` and sometimes they forget and only pass in `blah`, which is where the issue arose. Didn't realize `blah*` was an option. I mean, `*` vs `.yaml`...one is more chars but really it doesn't help some users when the kneejerk is still just `blah`.

Babbling, but just trying to say I'm just surprised the solution is to append `*`.

armi/cases/tests/test_cases.py
```diff
@@ -398,6 +399,77 @@ def test_titleSetterGetter(self):
...
+            f.write("some: other\nyaml: file\n")
+        os.mkdir("subdir")
+        cs.writeToYamlFile("subdir/settings3.yaml")
+        cs.writeToYamlFile("subdir/skipSettings.yaml")
+
+    def tearDown(self):
+        self.dc.__exit__(None, None, None)
+
+    def test_discover(self):
+        """Test CaseSuit.discover().
```
opotowsky, line 434:
```suggestion
        """Test CaseSuite.discover().
```

opotowsky (approved):
I only see f-string edits in `recursivelyLoadSettingsFiles` -- did you mean to include a functional edit?

But this appears to work as intended. I have a few optional suggestions and a babble, so I'm approving.

armi/cases/suite.py
john-science: Well, this seems like a solution to the suggested complaint to me.

But the ticket associated with this PR has a lot of conversation and I could be missing something in the request... if that makes sense. I could be wrong.

Does this seems like a complete solution to you?

opotowsky: Yea  ithink it accomplishes the complaint being taken care of

john-science: Great!

I *thought* this would be an easy fix, but I really needed the feedback that it solved the correct problem.

Thanks for the review.!

PR #2655, opened 2026-09-12
Ensuring `nuclides` and `numberDensities` are numpy arrays

mgjarrett:

Force the component parameters that define the composition (`c.p.nuclides` and `c.p.numberDensities`) to always be a numpy array. If there are no number densities defined for a component, these two parameters will be size-0 numpy arrays instead of `None`. This is being done to avoid errors in the code where ARMI expects these parameters to be iterable.

By default, when special-formatted data such as these parameters go through a round-trip of being packed into the HDF5 database and then unpacked, a size-0 numpy array going in will unpack into a `None` value -- this setter will catch that condition and force the parameter back to a size-0 numpy array.

See this comment: https://github.com/terrapower/armi/pull/2109#issuecomment-5647343852

## SCR Information

Change Type: fixes

One-Sentence Rationale: ARMI expects the `nuclides` and `numberDensities` component parameters to be iterable.

---

## Checklist


mgjarrett:
@john-science @keckler not quite ready for review yet because I haven't added any unit tests. Just tagging you for awareness.

armi/reactor/components/component.py
```diff
@@ -646,7 +644,7 @@ def getNuclideNumberDensities(self, nucNames: list[str]) -> list[float]:
         else:
             byteNucs = [nucName.encode() for nucName in nucNames]
 
-        if self.p.numberDensities is None:
+        if len(self.p.numberDensities) == 0:
             return np.zeros(len(byteNucs), dtype=np.float64)
```
keckler: Is this necessary anymore? Maybe it is useful as a shortcut for speed, but if this was not here, I think the function would return effectively the same thing (a list-like-thing of zeros with `len(byteNucs)`), right?

keckler: Is it right that this chunk would return a numpy array, whereas the return at the bottom of the method would return a list?

armi/physics/neutronics/crossSectionGroupManager.py
```diff
@@ -525,7 +525,7 @@ def getBlockNuclideTemperatureAvgTerms(block, allNucNames):
 
     def getNumberDensitiesWithTrace(component, allNucNames):
         """Needed to make sure temperature of 0-density nuclides in fuel get fuel temperature."""
-        if component.p.nuclides is None:
+        if len(component.p.nuclides) == 0:
             return [0.0 for _nuc in allNucNames]
```
keckler: This is approximately an order of magnitude faster.
```suggestion
            return [0.0] * len(allNucNames)
```

john-science: @yardasol For later discussion.

armi/reactor/components/component.py
mgjarrett: Yes, all the returns should probably be numpy arrays

keckler:
@john-science Can we make sure this makes it into the upcoming release?

armi/reactor/components/component.py
```diff
@@ -622,8 +622,6 @@ def getNuclides(self):
 
         This includes anything that has been specified in here, including trace nuclides.
         """
-        if self.p.nuclides is None:
-            return []
         return [nucName.decode() for nucName in self.p.nuclides]
```
john-science, line 625: So THIS is still a list and not a numpy array?

armi/reactor/components/component.py
```diff
@@ -646,15 +644,23 @@ def getNumberDensity(self, nucName):
         else:
             return 0.0
 
-    def getNuclideNumberDensities(self, nucNames: list[str]) -> list[float]:
-        """Return a list of number densities for the nuc names requested."""
+    def getNuclideNumberDensities(self, nucNames: Iterable[Union[str, np.bytes_]]) -> np.ndarray:
+        """
+        Return a list of number densities for the nuc names requested.
```
john-science:
```suggestion
        Return an array of number densities for the nuclide names requested.
```

john-science (approved):
I have a couple of non-blocking thoughts.

Thanks for the PR!

armi/reactor/components/component.py
mgjarrett: Yes, the nuclide names are stored as a numpy array so they can be easily saved to the HDF5 database. But when we need an iterable of nuclide names somewhere within the code, we a list of plain `str` nuclide names rather than byte strings.

PR #2661, opened 2026-09-21
Preserving full-core Cartesian block volumes on Core symmetry axes

### LINKED ISSUE #2660: Symmetry factor of central assembly in a full-core cartesian model with an odd number of assembly columns incorrectly gives 4 instead of 1
The below code path fires regardless of whether there is symmetry or not. If the domain is full core then this should not happen. 

https://github.com/terrapower/armi/blob/22e6b8d729511c3659d6c9b7cce27cf076ad2da5/armi/reactor/blocks/cartesianBlock.py#L79-L84

The `isThroughCenterAssembly` is automatically set in this condition here:

https://github.com/terrapower/armi/blob/22e6b8d729511c3659d6c9b7cce27cf076ad2da5/armi/reactor/blueprints/gridBlueprint.py#L365-L371

I have a fix, will send a PR.

partofthething:

An additional condition check is added to ensure the symmetry factor of Cartesian blocks is computed correctly in full core cases.

closes #2660

## SCR Information

Change Type: fixes

One-Sentence Rationale: Fixes the symmetry factor for an edge case in Cartesian blocks.

---

## Checklist


john-science:
The CI in this PR is failing... okay, whatever.

MORE INTERESTING @yardasol , is that unit tests didn't run on this PR. So we have reproduced this problem for someone other than you.

I note Nick is working off a forked repo.

john-science (approved):
Well, the unit tests didn't run on this PR. But we think that must be something to do with the fact you are running from a forked repo.

~I don't immediately see HOW that is happening though. I guess I can just run the tests on your branch locally and verify that way. Odd though. Needs fixing.~

Okay, I have a PR up to fix this "forked repo CI" issue: https://github.com/terrapower/armi/pull/2663

PR #2662, opened 2026-09-21
Closing log files safely at the end of run

mgjarrett:

At the end of a run, ARMI attempts to close all stderr files. When a run takes a long time and it is writing log files to a networked drive, it is possible that the network connection will be interrupted at some point during the run, rendering the stored file handles stale. This condition currently would currently cause an `OSError` to be thrown; the proposed change catches this `OSError` with a try/except and just keeps going.

## SCR Information

Change Type: fixes

One-Sentence Rationale: If the file handle for a logger is stale, then there is no open file handle to close and it is acceptable to just do nothing.

---

## Checklist


opotowsky (commented):
I like this, it's so elegant!

mgjarrett:
- closing this for a better solution: https://github.com/terrapower/armi/pull/2669/

Issue #2449, opened 2026-02-05
Blueprints loading technical debt


opotowsky:
When `ruamel.yaml` `0.19.1` was released, our CI began to fail. A temporary version pin was put into place but leaving that there is only going to cost us down the road.

Issue #2404 was created and the pin was able to be unpinned via #2405.

But note the commentary in the ticket:

```python
        # With the release of ruamel.yaml 0.19.1, we began getting the following error:
        # AttributeError: 'RoundTripLoader' object has no attribute 'max_depth'
        # Setting that attribute to `None` solved the issue. However, it would be prudent to rework blueprints loading
        # to side step the issue entirely.
        loader = RoundTripLoader if roundTrip else CLoader
        if roundTrip:
            loader.max_depth = None
        return super().load(stream, Loader=loader)
```

The `max_depth` hack gets us to the goal of unpinning, but the dependence on `yamlize` for blueprints loading is preventing us from removing the hack. This ticket is being created to document the technical debt we are carrying so perhaps we can make a plan to pay it back. 

I wonder what the performance gains are for the current approach. If they are needed, then perhaps we can find more modern tooling to do this for us?

SimplyKnownAsG:
Happy to accept help in yamlize. It needs love. They only thing it is giving is round trip comment retention. Otherwise, there are other options. I recently started using cattrs which has a yaml extension.

john-science:
We are still hoping to solve this tech debt, in yamlize if possible.

john-science:
@yardasol Here is that `yamlize` ticket I was trying to find.

@opotowsky Should this ticket still be assigned to you?

opotowsky:
I am happy to kick it to @yardasol ....I have a bunch of ARMI tickets I haven't been able to get to

SimplyKnownAsG:
I can fully transfer yamlize to terrrapower, if you'd like it!

opotowsky:
That would require we update the code. I'm not sure if it's better to update the code then get it released to PyPI or if it's better to ingest what we need from the code base and be on the hook for maintenance (guess we are on the hook no matter what). What do yall think?

john-science:
It would be helpful if we could make commits to `yamlize`, yeah.

And do releases too, of course. I assume those releases should be to pypi.org.

john-science:
@SimplyKnownAsG How do you feel about us forking `yamlize`, under the `github.com/terrapower/` org?

[...]

Issue #1448, opened 2023-10-23
There are too many (hex) assembly-specific tools in Core


john-science:
The `Core` class in ARMI is meant to be useful to represent all reactor cores:

https://github.com/terrapower/armi/blob/f580e3fe2120f93a0fbb0b5daa8424c9394c4698/armi/reactor/reactors.py#L226

But there are several major items in this class for assembly-based reactors:

https://github.com/terrapower/armi/blob/f580e3fe2120f93a0fbb0b5daa8424c9394c4698/armi/reactor/reactors.py#L398

https://github.com/terrapower/armi/blob/f580e3fe2120f93a0fbb0b5daa8424c9394c4698/armi/reactor/reactors.py#L418

https://github.com/terrapower/armi/blob/f580e3fe2120f93a0fbb0b5daa8424c9394c4698/armi/reactor/reactors.py#L479

https://github.com/terrapower/armi/blob/f580e3fe2120f93a0fbb0b5daa8424c9394c4698/armi/reactor/reactors.py#L489

https://github.com/terrapower/armi/blob/f580e3fe2120f93a0fbb0b5daa8424c9394c4698/armi/reactor/reactors.py#L539

> Wouldn't it better if we had a subclass of `Core` named something like `AssemblyCore` or even `HexAssemblyCore`?

john-science:
ARMI was developed with sodium-cooled fast reactors and molten salt reactors as its primary users. Thus, ARMI has some lingering assumptions that work best for those reactors. We would like to find and eliminate all such assumptions, particularly to improve our support for:

* water-cooled reactors
* TRISO-fuel reactors
* molten chloride fast reactors
* And so many more...

We should also separate out all assumptions that the core is pin-type, or that all assemblies are hex-shaped.

john-science:
@sombrereau says that he worked on a non-pin-type reactor project and wanted to use ARMI but found that ARMI has a baked-in assumption of a structured grid. And he had to spend real time to get around this assumption.

Now, ARMI has a structured grid class that basically everything is based on:

https://github.com/terrapower/armi/blob/969c45c2eac42081373c044eba92516494dfda34/armi/reactor/grids/structuredGrid.py#L35

But the default `Grid` class is, I guess, our unstructured grid type.

[...]

Issue #1008, opened 2022-12-05
Upgrade reactor data model to fully support an arbitrarily-deep composite design pattern


ntouran:
Right now we have a sort-of hybrid [Composite design pattern](https://en.wikipedia.org/wiki/Composite_pattern) for the reactor data model, with the following players:

* ARMI's `ArmiObject` is the official pattern's `Component` (sorry, we have a naming conflict here)
* ARMI's `Composite` is the pattern's `Composite` (yay)
* ARMI's `Component` is the pattern's `Leaf` 

Things that could be improved:

* ARMI's `Leaf` class is the parent of `Component` and `Material`. This should be cleaned up so that there's just `Component` and it' only used for `Component`. (this is #1007)
* It should be possible to add `Component`s and `Composite`s to any parent `Composite` arbitrarily. 
   *  I have a branch going that tried allowing Components to have children but I've decided that this approach will be more sustainable/realistic. 
* `Assembly` and `Block` are special `Composites`. This is sort of ok but shouldn't be forced.

But for this ticket, let's just focus on the 2nd bullet: **Allow deeper nesting of components and composites within composites.**

This will support #477.

john-science:
@drewj-usnctech Just FYI.

john-science:
**Bump.** This is still super important.

Perhaps I need to pin this as High Priority for the next release.

john-science:
And one year later... I still feel like this is really important for the over-all ARMI mission.

yardasol:
@yardasol is intereted in this

john-science:
This would be a GREAT improvement. And it'll have to happen at some point. 

But, be warned, this ticket will have tendrils all over the place. There are all kinds of functions that make assumptions about composites only have children that are components. It'll be a ton of little changes in ARMI, and more downstream.

I imagine the best solution here would be to first build a solid set of little tools to get the children of composites and so on. And then applying those, new, more general tools to solve all the little one-off problems you find.

yardasol:
I've been looking thru the ARMI codebase (and older GH discussions/issues) most of the day and thinking about next steps.

From what I can glean from my reading:
1. Currently, a `Composite` object _can_ nest of other `Composite` objects via the `Composite._children` attribute. `Component` objects are also included in the `Composite._children` attribute, as they subclass `Composite`
    - `Composite.getChildren()` supports retrieval of child `Composite` objects up to a specified depth (or untel the end of the tree is reached)
[...]

PR #2665, opened 2026-09-21
Applying the ruamel max_depth shim to every yamlize class, not just Blueprints

partofthething:

ruamel.yaml 0.19.1 started reading ``max_depth`` off the loader during ``get_single_node()``, but yamlize 0.7.1 builds its loaders in a way that never sets it, so every yamlize load raises:

    AttributeError: 'RoundTripLoader' object has no attribute 'max_depth'

The workaround for this was applied as a side effect of ``Blueprints.load``, which left ``Grids.load`` and the other yamlize classes broken unless blueprints happened to have been loaded first. That fails any new tests in test_gridBlueprints.py on its own.

Supplying the default at module scope covers every yamlize class in ARMI.

Highly related to #2449 but doesn't fix the underlying yamlize issue.

NOTE: Claude helped me find and fix this issue after I ran into it.

## SCR Information

Change Type: fixes

One-Sentence Rationale: Users will want to round-trip all their YAML input files.

---

## Checklist


john-science (approved):

john-science:
@opotowsky and I think that writing a unit tests to show the diff in this code would be tricky, since we already had this shim at a different level in the code.

john-science:
Hey @partofthething !

Nice to see you again! You should meet @yardasol , who just joined the team.

Olek is currently working on a [series of tickets](https://github.com/terrapower/armi/issues/1008#issuecomment-5337446555) to make it so Composites can have children which are Composites. The goal there is to be able to support a wider range of reactor geometries.

I am also trying to find time to separate out all the [hex-specific](https://github.com/terrapower/armi/issues/1448) logic in ARMI, so it is optional for those reactors without hex assemblies.

Since you're independent from TerraPower's specific reactor design, I'll be really interested to hear where the pain points are for your team using ARMI. And what kinds of ARMI design goals might be most helpful to you. (We can talk via GitHub Discussions, Tickets, or offline.)

Cheers!

